PR #6239, opened 2020-09-16
fix: improve error handling of invalid Avro identifier

big-andy-coates:
### Description 

fixes issue where using a column name in ksqlDB that was incompatible with Avro, e.g. one that starts with a numeric, either wasn't detected at the time the statement was issued (in the case of C* statements), or returned a ambiguous error message (in the case of C*AS statements).

### Testing done 

Usual

### Reviewer checklist

### DIFF
--- a/ksqldb-functional-tests/src/test/resources/query-validation-tests/avro.json
+++ b/ksqldb-functional-tests/src/test/resources/query-validation-tests/avro.json
@@ -664,7 +664,7 @@
     {
       "name": "should return error on field names incompatible with avro - C*",
       "statements": [
-        "CREATE STREAM INPUT (1AvroDoesNotLikeFieldsThatStartWithNumbers DOUBLE) WITH (kafka_topic='input_topic', value_format='AVRO');"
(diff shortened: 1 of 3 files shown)


ksqldb-functional-tests/src/test/resources/query-validation-tests/avro.json
```diff
@@ -660,6 +660,33 @@
         "type": "io.confluent.ksql.util.KsqlStatementException",
         "message": "Avro only supports MAPs with STRING keys"
       }
+    },
+    {
+      "name": "should return error on field names incompatible with avro - C*",
+      "statements": [
+        "CREATE STREAM INPUT (1AvroDoesNotLikeFieldsThatStartWithNumbers DOUBLE) WITH (kafka_topic='input_topic', value_format='AVRO');"
```
vcrfxia, line 667: nit: this isn't a valid field name in ksqlDB (even Avro aside) so it'd be more realistic to put this in quotes.

vcrfxia (approved):
LGTM -- one nit inline.

ksqldb-functional-tests/src/test/resources/query-validation-tests/avro.json
big-andy-coates: Good call. Thanks.

PR #6164, opened 2020-09-08
feat: CommandRunner enters degraded state when corruption detected in metastore

stevenpyzhang:
### Description 
Addresses https://github.com/confluentinc/ksql/issues/6104

Some refactoring to use `getCommandRunnerDegradedWarning` to get the error string to return from the API.
Introduced `CommandRunnerDegradedReason` to surface why the command runner is DEGRADED

### Testing done 
Starting with no command topic and backup enabled
```
ksql> create stream test as select * from KSQL_PROCESSING_LOG;
ksql> create stream test1 as select * from test;
```
Delete command topic and start up server again

```
ksql> create stream test as select * from KSQL_PROCESSING_LOG;
ksql> create stream test2 as select * from test;

 Message                            
------------------------------------
 Statement written to command topic 
------------------------------------
WARNING: The server has detected that the command topic may be corrupted. The backup of the command topic does not match the current contents of command topic.
DDL statements will not be processed until either:
1. The current command topic is deleted and the backup file is used to restore the command topic.
2. The current backup file is deleted.
The server must be restarted after performing either operation in order to resume normal functionality
```
note: test2 is read from the command topic consumer but it's not actually executed on the server.
### Reviewer checklist
- [ ] Ensure docs are updated if necessary. (eg. if a user visible feature is being added or changed).
- [ ] Ensure relevant issues are linked (description should include text like "Fixes #<issue number>")

PR #6240, opened 2020-09-16
feat: surface error to user when command topic deleted while server running

stevenpyzhang:
### Description 
Stacked on top of https://github.com/confluentinc/ksql/pull/6164
If the command topic is deleted currently, the command topic consumer in the CommandRunner thread continues to poll the non-existent topic, spamming logs with
```
WARN [Consumer clientId=consumer-null-1, groupId=null] Received unknown topic or partition error in fetch for partition _confluent-ksql-default__command_topic-0
```

PR changes it so that if the command topic doesn't exist, the server enters DEGRADED state and adds warnings to all api requests sent to the `/ksql` that the command topic was deleted.

Also added a new CommandRunnerDegradedReason category to expose this reason in the server metrics.

### Testing done 
Started up the server, deleted the command topic
```
ksql> show streams;

 Stream Name         | Kafka Topic                 | Format 
------------------------------------------------------------
 KSQL_PROCESSING_LOG | default_ksql_processing_log | JSON   
------------------------------------------------------------
WARNING: The server has detected corruption in the command topic due to modifications performed on it.
DDL statements will not be processed any further.
If a backup of the command topic is available, restore the command topic using the backup file.
A server restart is required to restore full functionality
```

Used jconsole to verify the command runner status is DEGRADED and the reason is COMMAND_TOPIC_DELETED

### Reviewer checklist


ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/computation/CommandRunner.java
```diff
@@ -289,6 +302,9 @@ void fetchAndRunCommands() {
     lastPollTime.set(clock.instant());
     final List<QueuedCommand> commands = commandStore.getNewCommands(NEW_CMDS_TIMEOUT);
     if (commands.isEmpty()) {
+      if (!commandTopicExists.get()) {
+        commandTopicDeleted = true;
```
agavra, line 306: we can save this for another PR, but I think it makes sense to have this method return some status then to set a ton of flags and check them in the main loop

ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/computation/CommandRunner.java
```diff
@@ -437,6 +460,13 @@ public void run() {
         if (!closed) {
           throw wue;
         }
+      } catch (final OffsetOutOfRangeException e) {
+        LOG.warn("The command topic offset was reset. CommandRunner thread exiting.");
+        state = new Status(
+            CommandRunnerStatus.DEGRADED,
+            CommandRunnerDegradedReason.COMMAND_TOPIC_DELETED
```
agavra, line 467: does offset reset always mean command topic deleted? it could mean the retention was configured incorrectly perhaps?

ksqldb-rest-app/src/test/java/io/confluent/ksql/rest/server/computation/CommandRunnerTest.java
```diff
@@ -524,6 +549,25 @@ public void shouldCloseEarlyWhenSerializationExceptionInFetch() throws Exception
     inOrder.verify(commandStore).close();
   }
 
+  public void shouldCloseEarlyWhenOffsetOutOfRangeException() throws Exception {
```
agavra, line 552: missing `@Test` annotation?

agavra (approved):
LGTM! Some minor comments inline

ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/computation/CommandRunner.java
```diff
@@ -91,8 +96,9 @@
 
   public enum CommandRunnerDegradedReason {
     NONE(errors -> ""),
-    CORRUPTED(Errors:: commandRunnerDegradedBackupCorruptedErrorMessage),
-    INCOMPATIBLE_COMMAND(Errors:: commandRunnerDegradedIncompatibleCommandsErrorMessage);
+    CORRUPTED(Errors::commandRunnerDegradedBackupCorruptedErrorMessage),
+    INCOMPATIBLE_COMMAND(Errors::commandRunnerDegradedIncompatibleCommandsErrorMessage),
+    COMMAND_TOPIC_DELETED(Errors::commandRunnerDegradedCommandTopicDeletedErrorMessage);
```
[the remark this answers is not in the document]
stevenpyzhang: I think the main reason is right now, backups are optional, so if backups aren't enabled, then it wouldn't really make sense to have the degraded reason be CORRUPTED.

ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/computation/CommandRunner.java
stevenpyzhang: I'll file a follow up issue for that refactoring improvement.

ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/computation/CommandRunner.java
stevenpyzhang: Every time we start up the server, we'd check/override the command topic properties such that the `retention.ms=-1` although I guess it is possible for the user themselves to override the retention config on the command topic which could trigger this. I can enhance the error message so that it mentions the degraded state is a result of `the command topic being deleted or the configurations being modified for it`

The REASON can be updated to `COMMAND_TOPIC_DELETED_OR_MODIFIED`

ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/computation/CommandRunner.java
```diff
@@ -91,8 +96,10 @@
 
   public enum CommandRunnerDegradedReason {
     NONE(errors -> ""),
-    CORRUPTED(Errors:: commandRunnerDegradedBackupCorruptedErrorMessage),
-    INCOMPATIBLE_COMMAND(Errors:: commandRunnerDegradedIncompatibleCommandsErrorMessage);
+    CORRUPTED(Errors::commandRunnerDegradedBackupCorruptedErrorMessage),
+    INCOMPATIBLE_COMMAND(Errors::commandRunnerDegradedIncompatibleCommandsErrorMessage),
+    COMMAND_TOPIC_DELETED_OR_MODIFIED(
```
rodesai, line 101: How is this different than corrupted? We're leaking an implementation detail of how we detect corruption here. The underlying condition is the same - the command topic isn't in the state that the system expects.

stevenpyzhang: I've combined the two now.

ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/computation/CommandRunner.java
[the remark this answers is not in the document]
stevenpyzhang: After talking it over with Rohan a bit more, I've combined CORRUPTED and COMMAND_TOPIC_DELETED into CORRUPTED. The COMMAND_TOPIC_DELETED could show up as CORRUPTED after a restart and having different states for the same external action (deleting the command topic) is strange.

I've made the error message returned from the API generic enough so it address both the backup corruption and the command topic deleted/modified case

```
The server has detected corruption in the command topic due to modifications performed on it. DDL statements will not be processed any further.
If a backup of the command topic is available, restore the command topic using the backup file.
A server restart is required to restore full functionality
```

PR #6241, opened 2020-09-17
docs: Minor improvements to testing tool docs

rmoff:
Add some bash magic to generate input files, and tidy up headings to make the logical flow more obvious


JimGalasyn (approved):
LGTM! It would be great if you could add line breaks at col 80.

PR #5594, opened 2020-06-10
fix: improve join predictability by increasing max task idle

big-andy-coates:
### Description 

improvement for: https://github.com/confluentinc/ksql/issues/5537

This change increases the Stream's `max.task.idle.ms` from its default `0` to `500`ms, which should provide a better out-of-the-box join experience for anyone starting a join with a large backlog of messages, i.e. a backlog large enough to mean the broker may only return records for one side of the join. See #5537 for more info.

### Testing done 

Manual, to ensure this config is being passed down correctly.

### Reviewer checklist
- [ ] Ensure docs are updated if necessary. (eg. if a user visible feature is being added or changed).
- [ ] Ensure relevant issues are linked (description should include text like "Fixes #<issue number>")

PR #5911, opened 2020-07-30
test: fix JoinIntTest by changing generated timestamp for table source

agavra:
### Description 

There's some weird behavior going on in the `JoinIntTest`, but I'm fixing it here to unblock the master branch for now. 

Essentially we have a stream of events that are produced using a timestamp extractor that extracts timestamps equal to the offset within the stream. Since there are 8 events in the stream, the max timestamp is `8L` and the first timestamp is `1L`. The raw time on these events (without the extractor) is `System.currentTimeInMillis()` There is also a table that is produced with `System.currentTimeInMillis() - 500` to ensure that the join events show up "before" the stream events. The problem is that when the timestamp extractor is applied, the table events are no longer "before" the stream events.

The weird part, is that it works if we set the timestamp to anything less than `9L` - meaning that if my stream event has timestamp 1L it can join with a table entry that has timestamp 8L only if there exists an event in the stream that has a timestamp greater than or equal to 8L.

Anyway, until we figure that out, this fixes master.

PS I'm also removing the deprecated suppression and replacing them with other configs.

### Testing done 

Test only change

### Reviewer checklist
- [ ] Ensure docs are updated if necessary. (eg. if a user visible feature is being added or changed).
- [ ] Ensure relevant issues are linked (description should include text like "Fixes #<issue number>")

PR #6251, opened 2020-09-17
revert: "fix: improve join predictability by increasing max task idle …

AlanConfluent:
…(#5594)"

This reverts commit c643d8f059dc9956b7789744b2ee297f72dad6b8.

The reason for reverting this commit is a performance regression with joins.  On the soak workload, we have a three node cluster and we track the messages consumed in aggregate across our queries (unfortunately not broken down by query).  With 70K messages/sec produced to the topic `ksql-ratings-json` we witnessed a steady state aggregate message consumption rate of 280K messages/sec on 0.11.0 with no lag.  If there was built up lag, it was able to drain it by consuming 340K message/sec.  With this commit, it topped out at 277K, with a growing kafka consumer lag of about 2M messages an hour for each streams thread computing the query `RATINGS_WITH_USER_AVERAGE`.  By this rough estimate, the maximum overall performance dropped by around 19%.  Of course, the join query `RATINGS_WITH_USER_AVERAGE` is just one part of the aggregate, so join performance likely fell even more.

Here are the schemas of the streams and table that are used:
```
CREATE STREAM RATINGS (
    rating_id BIGINT,
    user_id INT,
    stars INT,
    route_id INT,
    rating_time BIGINT,
    channel VARCHAR,
    message VARCHAR
) WITH (KAFKA_TOPIC='ksql-ratings-json', VALUE_FORMAT='json');

CREATE STREAM RATINGS_WITH_USER_AVERAGE WITH (KAFKA_TOPIC='RATINGS_WITH_USER_AVERAGE', PARTITIONS=20, REPLICAS=3) AS SELECT
  R.RATING_ID RATING_ID,
  R.USER_ID USER_ID,
  R.STARS STARS,
  RU.AVERAGE_STARS AVERAGE_STARS
FROM RATINGS R
LEFT OUTER JOIN RATINGS_BY_USER RU ON ((R.USER_ID = RU.USER_ID));

CREATE TABLE RATINGS_BY_USER WITH (KAFKA_TOPIC='RATINGS_BY_USER', PARTITIONS=20, REPLICAS=3) AS SELECT
  RATINGS.USER_ID USER_ID,
  (SUM(RATINGS.STARS) / COUNT(*)) AVERAGE_STARS
FROM RATINGS RATINGS
GROUP BY RATINGS.USER_ID;

```

The lower message consumption was fairly consistent across time.  With similar queries to those above, if you can track consumer lag and can produce a high volume of input messages, you should be able to see it start falling behind within minutes.  Probably the best way to measure performance before and after, in the event that a fix is to be tested, is to build up a backlog of messages and test peak message consumption by a join query.


vpapavas (approved):
LGTM! I didn't actually go through it to make sure that all changes are reverted correctly, I trust you did that :)

rodesai:
Are we sure that the "regression" wasn't just an artifact of correctness? Or put another way, before the change this is reverting was performance better because we were doing the wrong thing? A PR description would help here.

AlanConfluent:
> Are we sure that the perf regression wasn't just an artifact of correctness? Or put another way, before the change this is reverting was performance better because we were doing the wrong thing? A PR description would help here.

I don't know offhand how correct or not it was before this change.  The description of the PR implies that maybe it wasn't ideal beforehand (i.e. it doesn't always find records to join with if there's a big backlog?), but I'd have to let @big-andy-coates speak to that.  There's also the fact that we're paying a significant price to do it this way, even if it's more "correct".  I would think that the simplest way to approach this is roll it back and try to create an improved version that has less of a performance penalty (assuming that's fairly straightforward). Maybe people disagree and think it's worth the penalty.

apurvam:
I think reverting is the right thing to do. We will be no more or less correct than we were in 0.11. We should investigate better ways to achieve the objectives outside of blocking the release IMO.

apurvam:
@AlanConfluent i think we should also revert changes to tests which happened in a separate PR by @agavra

agavra:
@apurvam I think the changes made (you're referring to https://github.com/confluentinc/ksql/pull/5911 I suspect) are benefitial independent of this change (it makes the test more correct in either scenario)

apurvam (approved):

rodesai:
Agree we shouldn't block the release - we either ship with out the config change (and maybe miss some results? not sure what the impact is) or ship with it and accept the perf regression. Just saying it would help to (1) understand what the original PR's value/intent was that we're now losing, and (2) call out this choice in this PR's commit msg.

apurvam:
@rodesai that makes sense to me.

big-andy-coates:
Can we document the details of the perf regression we've seen as part of this PR?  How much of an impact was it? To what workloads?   Was it consistent or only seen occasionally? How to recreate, expected vs actual results, etc.

AlanConfluent:
I added a lot of detail @big-andy-coates .  If there's anything else I can tell you, please let me know.

Issue #6130, opened 2020-09-02
Add KSQL processing log message on uncaught exceptions from streams


rodesai:
**Is your feature request related to a problem? Please describe.**
Sometimes Kafka Streams exits due to an uncaught exception. This could be due to a streams bug, or a nontransient error (e.g. no acls, no disk space). 

**Describe the solution you'd like**
When this happens we should log a processing log message so that the error is discoverable. Kafka Streams lets us register an uncaught exception handler. We should register one that logs an error to the processing log.

**Describe alternatives you've considered**
We could also return these errors in the API when querying query status. In fact, we already do this. However, it would also be nice to include them in the processing log as well to increase the likelihood that these errors get bubbled up.

PR #6253, opened 2020-09-18
feat: add KSQL processing log message on uncaught streams exceptions

spena:
### Description 
Fixes https://github.com/confluentinc/ksql/issues/6130

Calls the `ProcessingLogger` on uncaught errors by Kafka streams. The processing log uncaught handler is registered by the `PersistentQueryMetadata` only. Transient queries do not need to log errors to the processing log (they're at sent back to the user instead).

The error sent to the logger uses a new `KafkaStreamsThreadErrorMessage` type that creates an structured error message. It contains the error message, the thread and the exception.

### Testing done 
Added unit tests

```
ksql> select * from KSQL_PROCESSING_LOG emit changes;
+---------------------------------------------+---------------------------------------------+---------------------------------------------+---------------------------------------------+
|LOGGER                                       |LEVEL                                        |TIME                                         |MESSAGE                                      |
+---------------------------------------------+---------------------------------------------+---------------------------------------------+---------------------------------------------+
|processing.CSAS_S1_0.ksql.logger.thread.excep|ERROR                                        |1600872731171                                |{TYPE=4, DESERIALIZATIONERROR=null, RECORDPRO|
|tion.uncaught                                |                                             |                                             |CESSINGERROR=null, PRODUCTIONERROR=null, SERI|
|                                             |                                             |                                             |ALIZATIONERROR=null, KAFKASTREAMSTHREADERROR=|
|                                             |                                             |                                             |{ERRORMESSAGE=Unhandled exception caught in s|
|                                             |                                             |                                             |treams thread, THREADNAME=_confluent-ksql-def|
|                                             |                                             |                                             |ault_query_CSAS_S1_0-22278e08-e1ab-475f-84fe-|
|                                             |                                             |                                             |c4c97815c5d4-StreamThread-4, CAUSE=[One or mo|
|                                             |                                             |                                             |re source topics were missing during rebalanc|
|                                             |                                             |                                             |e]}}                                         |
```

### Reviewer checklist


ksqldb-common/src/main/java/io/confluent/ksql/logging/processing/ProcessingLogMessageSchema.java
```diff
@@ -76,11 +76,23 @@
       .optional()
       .build();
 
+  public static final String KAFKA_STREAMS_THREAD_ERROR_FIELD_MESSAGE = "errorMessage";
+  public static final String KAFKA_STREAMS_THREAD_ERROR_FIELD_NAME = "name";
```
rodesai, line 80: let's use the name "threadName"

ksqldb-engine/src/main/java/io/confluent/ksql/query/QueryExecutor.java
```diff
@@ -229,7 +229,9 @@ public PersistentQueryMetadata buildPersistentQuery(
         ksqlConfig.getLong(KSQL_SHUTDOWN_TIMEOUT_MS_CONFIG),
         classifier,
         physicalPlan,
-        ksqlConfig.getInt(KsqlConfig.KSQL_QUERY_ERROR_MAX_QUEUE_SIZE)
+        ksqlConfig.getInt(KsqlConfig.KSQL_QUERY_ERROR_MAX_QUEUE_SIZE),
+        (ProcessingLogger) streamsProperties.get(
```
rodesai, line 233: We should use a different logger here. The logger is important because its name tells you a lot about the context of the error. In this case, the logger's name is `ksql.logger.production.error`, which tells you this is an error hit while trying to produce a record to Kafka. Other loggers' names include the node in the query topology. Eventually we'll let users adjust the log levels for some loggers at runtime, so you could see very detailed logging from just 1 UDF for example. In this case I'd name this something like `ksql.logger.thread.exception.uncaught`.

To create the logger, you can use the factory in the `processingLogContext` member of this class.

ksqldb-engine/src/main/java/io/confluent/ksql/util/PersistentQueryMetadata.java
```diff
@@ -114,6 +120,15 @@ protected PersistentQueryMetadata(
...
     this.physicalPlan = other.physicalPlan;
     this.materializationProviderBuilder = other.materializationProviderBuilder;
+    this.processingLogger = other.processingLogger;
+  }
+
+  @Override
+  protected void uncaughtHandler(final Thread thread, final Throwable error) {
+    super.uncaughtHandler(thread, error);
+
+    processingLogger.error(KafkaStreamsThreadError.of(
```
rodesai, line 130: consider logging this at the fatal level, since this is a thread-killing error.

rodesai (commented):
Looking good. Why are we only doing this for persistent queries?

ksqldb-engine/src/main/java/io/confluent/ksql/util/PersistentQueryMetadata.java
spena: How do I log at fatal level?

The `ProcessingLogger` has only `error`, an the `inner` logger has only `error, info, debug`.

spena:
@rodesai I didn't do transient queries because seems they catch the uncaught errors differently (not by QueryMetadata::uncaughtHandler), and they send the error back to the client. 

I assumed this is what we want, that users who are running push/pull queries see the error in their side 'cause they can do it. Contrary to persistent queries that they cannot see it unless the look in the processing log. Does it make sense? or should we also configure transient queries to use the internal QueryMetadata::uncaughtHandler() and processing log handler too?

ksqldb-common/src/main/java/io/confluent/ksql/logging/processing/ProcessingLogMessageSchema.java
spena: Done

ksqldb-engine/src/main/java/io/confluent/ksql/query/QueryExecutor.java
spena: Done

rodesai:
> @rodesai I didn't do transient queries because seems they catch the uncaught errors differently (not by QueryMetadata::uncaughtHandler), and they send the error back to the client.

This makes sense. Thanks for the explainer.

JimGalasyn (approved):
LGTM!

spena:
@rodesai I added the queryId to the logger name. This is how it looks now:
```
ksql> select * from KSQL_PROCESSING_LOG emit changes;
+---------------------------------------------+---------------------------------------------+---------------------------------------------+---------------------------------------------+
|LOGGER                                       |LEVEL                                        |TIME                                         |MESSAGE                                      |
+---------------------------------------------+---------------------------------------------+---------------------------------------------+---------------------------------------------+
|processing.CSAS_S1_0.ksql.logger.thread.excep|ERROR                                        |1600872731171                                |{TYPE=4, DESERIALIZATIONERROR=null, RECORDPRO|
|tion.uncaught                                |                                             |                                             |CESSINGERROR=null, PRODUCTIONERROR=null, SERI|
|                                             |                                             |                                             |ALIZATIONERROR=null, KAFKASTREAMSTHREADERROR=|
|                                             |                                             |                                             |{ERRORMESSAGE=Unhandled exception caught in s|
|                                             |                                             |                                             |treams thread, THREADNAME=_confluent-ksql-def|
|                                             |                                             |                                             |ault_query_CSAS_S1_0-22278e08-e1ab-475f-84fe-|
|                                             |                                             |                                             |c4c97815c5d4-StreamThread-4, CAUSE=[One or mo|
|                                             |                                             |                                             |re source topics were missing during rebalanc|
|                                             |                                             |                                             |e]}}                                         |
```

agavra (approved):
LGTM pending Rohan's comments

rodesai (approved):
LGTM

PR #6194, opened 2020-09-13
feat: add `KEY_FORMAT` and `FORMAT` WITH clause properties

vcrfxia:
### Description 

Fixes: https://github.com/confluentinc/ksql/issues/6208

This PR:
- adds two new WITH clause properties: `KEY_FORMAT` and `FORMAT`. An error is thrown if `FORMAT` is provided in addition to either `KEY_FORMAT` or `VALUE_FORMAT`. If one or neither of `KEY_FORMAT` and `VALUE_FORMAT` is provided in CS/CT statement, the missing one(s) will be injected into the statement via `DefaultFormatInjector`.
- adds two server configs to provide default key and value formats: `ksql.persistence.default.format.key` and `ksql.persistence.default.format.value`. The former defaults to "KAFKA" whereas the latter has no default.
- adds a feature flag for the primitive key work (`ksql.key.format.enabled`) that defaults to false. If false, the current behavior of requiring `VALUE_FORMAT` is preserved, and an error is thrown if either `KEY_FORMAT` or `FORMAT` is provided.

As of this PR, the only supported key format is "KAFKA". An error is thrown if other formats are provided (in both C* statements and also C*AS statements).

Separate PR for docs: https://github.com/confluentinc/ksql/pull/6252

### Testing done 

Unit + QTT. Not all tests are passing which is why this is a draft PR -- see comments inline.

### Reviewer checklist
- [ ] Ensure docs are updated if necessary. (eg. if a user visible feature is being added or changed).
- [ ] Ensure relevant issues are linked (description should include text like "Fixes #<issue number>")

PR #6254, opened 2020-09-18
chore: example wiring up of KEY_FORMAT

big-andy-coates:
Just raising as an example of how we can wire in KEY_FORMAT and default key format config without an injector. 

Very rough code only.  Doesn't handle default value format, doesn't include any tests, doesn't add `FORMAT` etc etc.

Basically, we need to worry about:

1. What formats the user supplied in C* statements and the defaults in the config:  this is handled by `TopicFactory`.
2. What formats the user supplied in C*AS statements: this is handled by `Analyzer`.

These changes are inline with the general design of the code (as its always been), and don't require an injector or reformatting of the SQL

### DIFF
--- a/ksqldb-common/src/main/java/io/confluent/ksql/serde/SerdeOption.java
+++ b/ksqldb-common/src/main/java/io/confluent/ksql/serde/SerdeOption.java
@@ -28,7 +28,7 @@ public enum SerdeOption {
    *
    * @see SerdeOption#UNWRAP_SINGLE_VALUES
    */
-  WRAP_SINGLE_VALUES(SerdeFeature.WRAP_SINGLES),
+  WRAP_SINGLE_VALUES(SerdeFeature.WRAP_SINGLES, false),
 
   /**
    * If the value schema contains only a single field, persist it as an anonymous value.
@@ -38,17 +38,23 @@ public enum SerdeOption {
    *
    * @see SerdeOption#WRAP_SINGLE_VALUES
    */
-  UNWRAP_SINGLE_VALUES(SerdeFeature.UNWRAP_SINGLES),
+  UNWRAP_SINGLE_VALUES(SerdeFeature.UNWRAP_SINGLES, false),
 
   /**
    * Key version of {@link #UNWRAP_SINGLE_VALUES}.
    */
-  UNWRAP_SINGLE_KEYS(SerdeFeature.UNWRAP_SINGLES);
+  UNWRAP_SINGLE_KEYS(SerdeFeature.UNWRAP_SINGLES, true);
 
+  private final boolean isKey;
   private final SerdeFeature requiredFeature;
 
-  SerdeOption(final SerdeFeature requiredFeature) {
+  SerdeOption(final SerdeFeature requiredFeature, final boolean isKey) {
+    this.isKey = isKey;
     this.requiredFeature = Objects.requireNonNull(requiredFeature, "requiredFeature");
+  }
+
+  public boolean isKeyOption() {
+    return isKey;
   }
 
   public SerdeFeature requiredFeature() {
--- a/ksqldb-common/src/main/java/io/confluent/ksql/serde/SerdeOptions.java
+++ b/ksqldb-common/src/main/java/io/confluent/ksql/serde/SerdeOptions.java
@@ -60,17 +60,12 @@ public final class SerdeOptions {
     validate(this.options);
   }
 
-  @SuppressWarnings("MethodMayBeStatic")
   public EnabledSerdeFeatures keyFeatures() {
-    // Currently there are no key features:
-    return EnabledSerdeFeatures.of();
+    return features(true);
   }
 
   public EnabledSerdeFeatures valueFeatures() {
-    // Currently there are no key features, so all are value features:
-    return EnabledSerdeFeatures.from(options.stream()
-        .map(SerdeOption::requiredFeature)
-        .collect(Collectors.toSet()));
+    return features(false);
   }
 
   public Set<SerdeOption> all() {
@@ -109,6 +104,13 @@ public final class SerdeOptions {
   @Override
   public String toString() {
     return options.toString();
+  }
+
+  private EnabledSerdeFeatures features(final boolean key) {
+    return EnabledSerdeFeatures.from(options.stream()
+        .filter(option -> option.isKeyOption() == key)
+        .map(SerdeOption::requiredFeature)
+        .collect(Collectors.toSet()));
   }
 
   private static void validate(final Set<SerdeOption> options) {
--- a/ksqldb-engine/src/main/java/io/confluent/ksql/analyzer/Analyzer.java
+++ b/ksqldb-engine/src/main/java/io/confluent/ksql/analyzer/Analyzer.java
@@ -162,7 +162,7 @@ class Analyzer {
       final String topicName = sink.getProperties().getKafkaTopic()
           .orElseGet(() -> topicPrefix + sink.getName().text());
 
-      final KeyFormat keyFormat = buildKeyFormat();
+      final KeyFormat keyFormat = buildKeyFormat(sink);
       final Format format = getValueFormat(sink);
 
       final Map<String, String> sourceProperties = new HashMap<>();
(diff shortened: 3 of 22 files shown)


ksqldb-common/src/main/java/io/confluent/ksql/serde/SerdeOption.java
```diff
@@ -28,7 +28,7 @@
    *
    * @see SerdeOption#UNWRAP_SINGLE_VALUES
    */
-  WRAP_SINGLE_VALUES(SerdeFeature.WRAP_SINGLES),
+  WRAP_SINGLE_VALUES(SerdeFeature.WRAP_SINGLES, false),
```
big-andy-coates, line 31: Ignore the changes in this file - this is work I need to do, but was needed to get the examples working.

ksqldb-common/src/main/java/io/confluent/ksql/serde/SerdeOptions.java
```diff
@@ -60,17 +60,12 @@ private SerdeOptions(final ImmutableSet<SerdeOption> options) {
     validate(this.options);
   }
 
-  @SuppressWarnings("MethodMayBeStatic")
   public EnabledSerdeFeatures keyFeatures() {
-    // Currently there are no key features:
-    return EnabledSerdeFeatures.of();
+    return features(true);
```
big-andy-coates, line 64: Ignore the changes in this file - this is work I need to do, but was needed to get the examples working.

ksqldb-engine/src/main/java/io/confluent/ksql/analyzer/Analyzer.java
```diff
@@ -196,22 +196,30 @@ private void analyzeNonStdOutSink(final Sink sink) {
...
+          .getDataSource()
+          .getKsqlTopic()
+          .getKeyFormat()
+          .getFormatInfo();
+
+      final FormatInfo keyFormat = sink.getProperties().getKeyFormatInfo()
+          .orElse(sourceKeyFormat);
+
+      // Todo: need to handle inheritable vs non-inheritable key properties
+
```
big-andy-coates, line 214: This is the bit of code that handles getting the key format from the WITH properties for C*AS statements.

ksqldb-engine/src/main/java/io/confluent/ksql/schema/ksql/inference/SchemaRegisterInjector.java
```diff
@@ -76,10 +78,15 @@ private void registerForCreateSource(final ConfiguredStatement<? extends CreateS
 
     final LogicalSchema schema = cs.getStatement().getElements().toLogicalSchema();
 
+    final KsqlConfig ksqlConfig = cs.getConfig()
+        .cloneWithPropertyOverwrite(cs.getConfigOverrides());
+
+    final KsqlTopic ksqlTopic = TopicFactory.create(cs.getStatement().getProperties(), ksqlConfig);
```
big-andy-coates, line 84: SchemaRegisterInjector should delegate to `TopicFactory` to get key and value formats, so its not duplicating code.

ksqldb-engine/src/main/java/io/confluent/ksql/topic/TopicFactory.java
```diff
@@ -31,17 +32,21 @@
...
+      final CreateSourceProperties properties,
+      final KsqlConfig config
+  ) {
     final String kafkaTopicName = properties.getKafkaTopic();
 
     final Optional<WindowType> windowType = properties.getWindowType();
     final Optional<Duration> windowSize = properties.getWindowSize();
 
+    final FormatInfo keyFormatInfo = properties.getKeyFormatInfo()
+        .orElseGet(() -> getDefaultKeyFormat(config));
```
big-andy-coates, line 45: This is the code that is responsible for handling `KEY_FORMAT` for C* statements.  You'd need to extend it to handle `FORMAT` too.

ksqldb-parser/src/main/java/io/confluent/ksql/parser/properties/with/CreateSourceAsProperties.java
```diff
@@ -105,6 +99,11 @@ public SerdeOptions getSerdeOptions() {
     return SerdeOptions.of(builder.build());
   }
 
+  public Optional<FormatInfo> getKeyFormatInfo() {
+    return Optional.ofNullable(props.getString(CommonCreateConfigs.KEY_FORMAT_PROPERTY))
+        .map(format -> FormatInfo.of(format, ImmutableMap.of()));
+  }
```
big-andy-coates, line 105: This should also return non-empty if `FORMATS` is set, (same for value version).

ksqldb-parser/src/main/java/io/confluent/ksql/parser/properties/with/CreateSourceAsProperties.java
```diff
@@ -69,10 +67,6 @@ private CreateSourceAsProperties(final Map<String, Literal> originals) {
     props.validateDateTimeFormat(CommonCreateConfigs.TIMESTAMP_FORMAT_PROPERTY);
   }
 
-  public Optional<Format> getValueFormat() {
-    return getFormatInfo().map(FormatFactory::of);
-  }
```
big-andy-coates, line 74: Got rid!

ksqldb-parser/src/main/java/io/confluent/ksql/parser/properties/with/CreateSourceProperties.java
```diff
@@ -78,10 +76,6 @@ public static CreateSourceProperties from(final Map<String, Literal> literals) {
     validateWindowInfo();
   }
 
-  public Format getValueFormat() {
-    return FormatFactory.of(getFormatInfo());
-  }
```
big-andy-coates, line 83: got rid!

ksqldb-parser/src/main/java/io/confluent/ksql/parser/properties/with/CreateSourceProperties.java
```diff
@@ -131,6 +125,11 @@ public String getKafkaTopic() {
     return Optional.ofNullable(props.getInt(CreateConfigs.SCHEMA_ID));
   }
 
+  public Optional<FormatInfo> getKeyFormatInfo() {
+    return Optional.ofNullable(props.getString(CommonCreateConfigs.KEY_FORMAT_PROPERTY))
+        .map(format -> FormatInfo.of(format, ImmutableMap.of()));
+  }
```
big-andy-coates, line 131: This should also return non-empty if `FORMATS` is set, (same for value version).

ksqldb-engine/src/main/java/io/confluent/ksql/topic/TopicFactory.java
```diff
@@ -51,4 +56,14 @@ public static KsqlTopic create(final CreateSourceProperties properties) {
...
   }
+
+  private static FormatInfo getDefaultKeyFormat(final KsqlConfig config) {
+    final String defaultKeyFormat = config.getString(KsqlConfig.KSQL_DEFAULT_KEY_FORMAT_CONFIG);
+    if (defaultKeyFormat == null) {
+      throw new KsqlException("Blah blha");
+    }
+
+    // Would need to wire in key properties obvs
+    return FormatInfo.of(defaultKeyFormat);
```
big-andy-coates, line 67: This is the code that handles the default key format config for C* statements - you'd need to extend for default value format.

ksqldb-common/src/main/java/io/confluent/ksql/serde/SerdeOption.java
big-andy-coates: see:  https://github.com/confluentinc/ksql/pull/6256

ksqldb-common/src/main/java/io/confluent/ksql/serde/SerdeOptions.java
big-andy-coates: see:  https://github.com/confluentinc/ksql/pull/6256

ksqldb-engine/src/main/java/io/confluent/ksql/schema/ksql/inference/DefaultSchemaInjector.java
```diff
@@ -121,6 +122,15 @@ private SchemaAndId getValueSchema(
     return result.schemaAndId.get();
   }
 
+  private static boolean valueFormatDoesNotSupportSchemaInference(
+      final ConfiguredStatement<CreateSource> statement
+  ) {
+    final Format valueFormat = FormatFactory
+        .of(statement.getStatement().getProperties().getFormatInfo());
```
vcrfxia, line 129: This might not be present at this step. Should this also delegate to TopicFactory? Feels a bit odd since there's no topic being created.

vcrfxia: Also, we still need to throw on unsupported key formats somewhere. For C*AS statements we could do it in the Analyzer, but I wonder where it makes sense to do it for C* statements. Should we do that here?

In https://github.com/confluentinc/ksql/pull/6194 this is happening for both C* and C*AS statements in EngineExecutor#plan(). If we want to do it there we'll need to add logic for finding the default formats there as well, which doesn't feel right.

vcrfxia: Another argument in favor of the injector is that the statement shown to the user currently has everything else filled in explicitly: schemas (in the case of schema inference), resolved types (in the case of custom types), and number of topic replicas/partitions. The statement reflects exactly what the engine executed. Not populating default key/value formats would break pattern here.

In terms of arguments against the injector:
- one less injector, definitely a plus
- ??

Both approaches involve a decent amount of headache around the feature flag. In this PR we'll have to add checks somewhere to validate that default value formats (and probably explicit key formats also) can't be used unless the feature flag is enabled. Looks like we'll have to do that separately for `C*` and `C*AS` statements under this design.

Issue #6209, opened 2020-09-16
[JSON Support]: All query plans require WRAP_SINGLE_KEYS set to `false`


big-andy-coates:
To ensure future versions of ksqlDB, which support wrapped single keys, are backwards compatible with plans created with this version of ksqlDB, all generated plans will need to be built with the `UNWRAP_SINGLE_KEYS` `SerdeOption` set.  This will ensure such plans will continue to unwrap the key, even once the default may be to wrap it.

Note: `WRAP_SINGLE_KEYS` does not exist as a `WITH` clause property yet. However, it will be added later.

big-andy-coates:
Re-opened, as realised this also needs to happen for internal topics, not just sink and source topics, and the way the code currently works is that SerdeOptions are _not_ inherited. This means an internal topic won't inherit the unwrapped flag from the source.

The right fix here is likely to make them inherited.

PR #6256, opened 2020-09-18
chore: finish wiring up unwrapped keys

big-andy-coates:
### Description 

fixes: https://github.com/confluentinc/ksql/issues/6209

This finishes the wiring of the `WRAP_SINGLE_KEYS` option up. (Some bits were missed in the first PR)

### Testing done 

Usual

### Reviewer checklist

### DIFF
--- a/ksqldb-common/src/main/java/io/confluent/ksql/serde/SerdeOption.java
+++ b/ksqldb-common/src/main/java/io/confluent/ksql/serde/SerdeOption.java
@@ -28,7 +28,7 @@ public enum SerdeOption {
    *
    * @see SerdeOption#UNWRAP_SINGLE_VALUES
    */
-  WRAP_SINGLE_VALUES(SerdeFeature.WRAP_SINGLES),
+  WRAP_SINGLE_VALUES(SerdeFeature.WRAP_SINGLES, false),
 
   /**
    * If the value schema contains only a single field, persist it as an anonymous value.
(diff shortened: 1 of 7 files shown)


ksqldb-common/src/main/java/io/confluent/ksql/serde/SerdeOptions.java
```diff
@@ -60,33 +59,22 @@ private SerdeOptions(final ImmutableSet<SerdeOption> options) {
...
-  public Optional<SerdeOption> keyWrapping() {
-    return Optional.ofNullable(
-        Iterables.getFirst(Sets.intersection(options, KEY_WRAPPING_OPTIONS), null)
-    );
-  }
-
-  public Optional<SerdeOption> valueWrapping() {
-    return Optional.ofNullable(
-        Iterables.getFirst(Sets.intersection(options, VALUE_WRAPPING_OPTIONS), null)
-    );
```
big-andy-coates, line 89: Dropped these in favour of `findAny`, which is more generic.

ksqldb-common/src/main/java/io/confluent/ksql/serde/SerdeOptions.java
```diff
@@ -111,6 +99,13 @@ public String toString() {
     return options.toString();
   }
 
+  private EnabledSerdeFeatures features(final boolean key) {
+    return EnabledSerdeFeatures.from(options.stream()
+        .filter(option -> option.isKeyOption() == key)
+        .map(SerdeOption::requiredFeature)
+        .collect(Collectors.toSet()));
+  }
```
big-andy-coates, line 107: key features are now returned as key features, not value features!

ksqldb-common/src/main/java/io/confluent/ksql/serde/SerdeOption.java
```diff
@@ -38,19 +38,25 @@
...
    * Key version of {@link #UNWRAP_SINGLE_VALUES}.
    */
-  UNWRAP_SINGLE_KEYS(SerdeFeature.UNWRAP_SINGLES);
+  UNWRAP_SINGLE_KEYS(SerdeFeature.UNWRAP_SINGLES, true);
 
+  private final boolean isKey;
   private final SerdeFeature requiredFeature;
 
-  SerdeOption(final SerdeFeature requiredFeature) {
+  SerdeOption(final SerdeFeature requiredFeature, final boolean isKey) {
```
vcrfxia, line 51: Does it make sense to assume all serde options will always be either key- or value-specific, never both nor neither? Can't think of why this wouldn't be true in the future but seemed worth double-checking.

vcrfxia (approved):
LGTM. Thanks!

ksqldb-common/src/main/java/io/confluent/ksql/serde/SerdeOption.java
big-andy-coates: That's the assumption at the moment.  If that changes... we change the code! ;)

Ideally, we'd ditch `SerdeOption` and just keep `SerdeFeature`, but that's a bigger change.

PR #6257, opened 2020-09-18
chore: drop support for executing legacy DDL/DML

vcrfxia:
### Description 

After query execution plans were introduced by [KLIP 6](https://github.com/confluentinc/ksql/blob/master/design-proposals/klip-6-execution-plans.md), we required users to migrate existing statements rather than performing in-place upgrades when upgrading from ksqlDB 0.7.0 to ksqlDB 0.8.0, and also when migrating from Confluent Platform KSQL 5.4 to ksqlDB 5.5. 

As a result, the engine no longer needs to support legacy `C*`, `C*AS`, `INSERT INTO`, and `DROP` source commands that do not have execution plans in the command topic.

### Testing done 

New unit tests + existing tests pass.

### Reviewer checklist


rodesai (approved):
LGTM

PR #6259, opened 2020-09-18
docs: KLIP-38 for Variable Substitution

spena:
### Description 
Add KLIP-38 to support variable substitution in ksqlDB.

### Testing done 

### Reviewer checklist

### DIFF
--- a/design-proposals/klip-38-variable-substitution.md
+++ b/design-proposals/klip-38-variable-substitution.md
@@ -1,4 +1,4 @@
-# KLIP 34 - Variable Substitution
+# KLIP 38 - Variable Substitution
 
 **Author**: Sergio Peña (@spena) |
 **Release Target**: 0.14 |
@@ -30,10 +30,27 @@ Issue: https://github.com/confluentinc/ksql/issues/6199
 
 ## What is not in scope
 
-* Provide support to prompt users to type a variable value if it wasn't defined (Other DBs support this)
-* Support variable substitution of environment and other non-user variables
-* Provide a set of pre-defined variables to use in the current session
-* Support variable substitution in the REST API (cannot store variables after consecutive requests)
+* `Prompt users to type values`
+
+  Oracle SQL*Plus prompts users to type the variable if not defined. Useful, but not necessary, and it will be a limited
+  support only (i.e. CLI and UI, but not Java API or other clients).
+
+* `Provide list of pre-defined variables`
+
+   Some DBs provide a list of pre-defined variables users can use out-of-box. We don't have any variables we'd like to add at this time.
+
+* `Support variable substitution in the REST API`
+
+  More complex to do to achieve the same functionality. We want variable substitution in clients only.
+
+* `Support UDF when defining variables`
+
+  It would be nice to use functions, like `set var = now();`, `set var = uppercase('a');`, etc. But this requires clients (CLI, UI, Java API, ...) to have
+  better parsing mechanism to detect UDF, make a request to the server, then assign the value to the variable. Also, there
+  are functions that might be executed in the CLI context, such as NOW() which should return the current time of the CLI.
+
+  We could define a set of functions that can be used instead, and just execute them in the local context; but this will require
+  every client supports that, which will make things more complicated (i.e. the ksqlDB UI).
 
 ## Value/Return
 
@@ -95,7 +112,7 @@ Also, the use of single-quotes in `SET` server settings is tedious to type. So a
 This looks confusing too. So we have to decide for a new syntax for substitution variables.
 
 Other DBs use different syntax to create user variables (DECLARE, DEFINE, \set, SET). For ksqlDB, I'll choose to use `DEFINE` and `UNDEFINE` which
-is what Oracle SQL*plus uses. Any user coming from this DB will immediately understand the command.  
+is what [Oracle SQL*plus uses](https://blogs.oracle.com/opal/sqlplus-101-substitution-variables). Any user coming from this DB will immediately understand the command.  
 This is opened to discussion, but will continue with this for now.
 
 #### Creating and printing substitution variables
@@ -264,7 +281,7 @@ Error: Fail because ${topic_${env}} topic name is invalid.
 The error message is just an example of what ksqlDB will display.
 
 When using SQL scripts in headless mode, we need to provide a server setting to enable/disable variable substitution. The setting will be
-`ksql.enable.headless.variable.substitution` as Boolean (default = true). The reason is that `SET CLI` is a CLI-specific command not available
+`ksql.headless.variable.substitution.enable` as Boolean (default = true). The reason is that `SET CLI` is a CLI-specific command not available
 in headless execution. Also, CLI should not request the Server is variable substitution is enabled as that should be a user decision to substitute or not.
 
 ### 6. Implementation
--- a/design-proposals/klip-38-variable-substitution.md
+++ b/design-proposals/klip-38-variable-substitution.md
@@ -150,10 +150,17 @@ UNDEFINE name;
 
 ### 2. Scope
 
-All variables will live during the session of a ksqlDB connection, then destroyed when the connection is closed. Only the user
+This variable has a session scope.
+
+Variables will live during the session of a ksqlDB connection, then destroyed when the connection is closed. Only the user
 in that session can interact with the defined substitution variables. No other users can interact with other users' variables.
 
 In CLI, the session is opened when starting the CLI with the `ksql` command, and destroyed when typing `exit`.
+
+In Java API client, the session is opened when `Client.create()` is called, and closed when `Client.close()` is called. Variables
+may be defined using a client method instead of a new syntax. This is to ensure users know the scope starts after the `create()` call.
+
+TBD: Define a Java API method for variable substitution.
 
 In headless, the session is opened when starting the ksqlDB server with a SQL script. The variables stay alive until the server is
 killed.
@@ -183,39 +190,46 @@ executed, thus to avoid limiting users from using DDL/DML commands while definin
 
 ### 3. Referencing substitution variables
 
-Variables will be referenced by wrapping the variable between `{}` characters (i.e. `{replicas}`). These characters are also used by
-QTT (or ksqlDB functional tests) to replace variables. It's better to keep the same format between testing and production to avoid
-confusion.
+Variables will be referenced by wrapping the variable between `${}` characters (i.e. `${replicas}`).
+
+Noe: We need to make changes in QTT tests to use `${}` references instead of `{}` for consistency.
 
 Example:
 ```
 ksql> DEFINE format = 'AVRO';
 ksql> DEFINE replicas = '3';
-ksql> CREATE STREAM stream1 (id INT) WITH (kafka_topic='stream1', value_format='{format}', replicas={replicas});
+ksql> CREATE STREAM stream1 (id INT) WITH (kafka_topic='stream1', value_format='${format}', replicas=${replicas});
 ```
 
 Substitution will not attempt to add single-quotes to the values. Users will need to know the data type to use when using a variable. The ksqlDB
 server will be in charge of parsing and failing if the type is not allowed.
 
-Note: Variables are not case sensitive. A reference to `{replicas}` is the same as `{REPLICAS}`.
+Note: Variables are case-insensitive. A reference to `${replicas}` is the same as `${REPLICAS}`.
 
 ### 4. Context for substitution variables
 
-Variable substitution will be allowed only in specific SQL statements, and only in places where literals or hard-coded values can be used.
-Variables cannot be used as column-names, reserved keywords, etc.
+Variable substitution will be allowed in specific SQL statements. They can be used to replace text and non-text literals, and identifiers such as
+column names and stream/table names. Variables cannot be used as reserved keywords.
 
-For instance (`{VAR}` is used a reference where variables are permitted):
+For instance:
+
+Note: `{VAR}` is used a reference where variables are permitted.
 ```
-ksql> CREATE STREAM <name> (col1 INT, col2 STRING) WITH (kafka_topic='{VAR}', format='{VAR}', replicas={VAR}, ...);
-ksql> INSERT INTO <stream> (col1, col2) VALUES ({VAR}, '{VAR}');
-ksql> SELECT * FROM <stream> WHERE col1 == {VAR} and col2 == '{VAR}' EMIT CHANGES; 
+ksql> CREATE STREAM ${streamName} (${colName1} INT, ${colName2} STRING) \
+      WITH (kafka_topic='${topicName}', format='${format}', replicas=${replicas}, ...);
+      
+ksql> INSERT INTO ${streamName} (${colName1}, ${colName2}) \
+      VALUES (${val1}, '${val2}');
+
+ksql> SELECT * FROM ${streamName} \
+      WHERE ${colName1} == ${val1} and ${colName2} == '${val2}' EMIT CHANGES; 
 ```
 Any attempt of using variables on non-permitted places will fail with the current SQL parsing error found when parsing the variable string.
 
 Variables can also be assigned to other variables:
 ```
 ksql> DEFINE var1 = 'topic';
-ksql> DEFINE var2 = 'other_{var1}';
+ksql> DEFINE var2 = 'other_${var1}';
 ksql> DEFINE var2;
 DEFINE var2    = 'other_topic'
 ```
@@ -223,20 +237,20 @@ DEFINE var2    = 'other_topic'
 Variables can also be assigned to set server or CLI settings:
 ```
 ksql> DEFINE offset = 'earliest'
-ksql> SET 'auto.offset.reset' = '{offset}';
+ksql> SET 'auto.offset.reset' = '${offset}';
 
 ksql> DEFINE wrap_toggle = 'ON'
-ksql> SET CLI WRAP {WRAP_TOGGLE}
+ksql> SET CLI WRAP ${WRAP_TOGGLE}
 ```
 
 Variables can also be used as nested variables:
 ```
 ksql> DEFINE env = 'prod';
 
-ksql> DEFINE topic_prod = 'topic_prod';
-ksql> DEFINE topic_qa = 'topic_qa';
+ksql> DEFINE topic_prod = 'my_prod_topic';
+ksql> DEFINE topic_qa = 'my_qa_topic';
 
-ksql> CREATE STREAM ... WITH (kafka_topic='{topic_{env}}'); 
+ksql> CREATE STREAM ... WITH (kafka_topic='${topic_${env}}'); 
 ```
 
 ### 5. Enable/disable substitution variables
@@ -244,8 +258,8 @@ ksql> CREATE STREAM ... WITH (kafka_topic='{topic_{env}}');
 When using CLI, we'll use the `SET CLI VARIABLE-SUBSTITUTION (ON|OFF)` command to enable/disable subsitution. Default will be `ON`.
 ```
 ksql> SET CLI VARIABLE-SUBSTITUTION OFF;
-ksql> CREATE STREAM ... WITH (kafka_topic='{topic_{env}}');
-Error: Fail because {topic_{env}} topic name is invalid. 
+ksql> CREATE STREAM ... WITH (kafka_topic='${topic_${env}}');
+Error: Fail because ${topic_${env}} topic name is invalid. 
 ```
 The error message is just an example of what ksqlDB will display.
 
--- a/design-proposals/klip-38-variable-substitution.md
+++ b/design-proposals/klip-38-variable-substitution.md
@@ -1,0 +1,288 @@
+# KLIP 34 - Variable Substitution
+
+**Author**: Sergio Peña (@spena) |
+**Release Target**: 0.14 |
+**Status**: _In Discussion_ |
+**Discussion**: TBD
+
+**tl;dr:** _Allow users to use variable substitution in SQL statements. Variable substitution enable users to write SQL scripts with customized output based on their environment needs._
+
+## Motivation and background
+
+The current plan to provide tooling and syntax for ksqlDB migrations enable users to write SQL scripts
+that can be run in CI/CD environments to verify their schema will work as expected. Also, it is very common for users
+to run their SQL scripts in different environments (QA, Devel, Prod, ...) to validate ksqlDB functionality. In order
+to give more flexibility on customizing their SQL output, we can support variable substitution to help separate
+environment-specific configuration variables from SQL code.
+
+This KLIP proposes a new syntax to define and interact with user variables, and implementation details to allow
+variable substitution in SQL commands.
+
+Issue: https://github.com/confluentinc/ksql/issues/6199
+
+## What is in scope
+
+* Define a new syntax to declare and interact with user variables
+* Allow variable substitution in SQL commands
+* Discuss the context and scope for variable substitution
+* Provide a configuration to enable/disable variable substitution
+* Support variable substitution in interactive and headless mode
+
+## What is not in scope
+
+* Provide support to prompt users to type a variable value if it wasn't defined (Other DBs support this)
+* Support variable substitution of environment and other non-user variables
+* Provide a set of pre-defined variables to use in the current session
+* Support variable substitution in the REST API (cannot store variables after consecutive requests)
+
+## Value/Return
+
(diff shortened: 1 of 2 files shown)


colinhicks (commented):
Very exciting @spena! I had some initial questions mostly to check my understanding in my first pass.

design-proposals/klip-38-variable-substitution.md
```diff
@@ -0,0 +1,288 @@
...
+```
+
+The `SOURCE` command will work like a `RUN SCRIPT`. If other SQL statements are in the file, then those statements will also be
+executed, thus to avoid limiting users from using DDL/DML commands while defining variables at the same time from one single script.
+
+`SOURCE` will be used only on the CLI and headless mode. REST API will not accept it because files may be out of the scope of the server filesystem.
+
+### 3. Referencing substitution variables
+
+Variables will be referenced by wrapping the variable between `{}` characters (i.e. `{replicas}`). These characters are also used by
```
spena, line 186: I might change the prefix/suffix for '${}' to avoid conflicting with JSON literal strings.

design-proposals/klip-38-variable-substitution.md
```diff
@@ -0,0 +1,288 @@
...
+The error message is just an example of what ksqlDB will display.
+
+When using SQL scripts in headless mode, we need to provide a server setting to enable/disable variable substitution. The setting will be
+`ksql.enable.headless.variable.substitution` as Boolean (default = true). The reason is that `SET CLI` is a CLI-specific command not available
+in headless execution. Also, CLI should not request the Server is variable substitution is enabled as that should be a user decision to substitute or not.
+
+### 6. Implementation
+
+For CLI, variable substitution will happen in the CLI side before making a ksqlDB request to the server. This allows less changes in ksqlDB code as we have  
+the `SET CLI` to enable/disable, and variables will be living in the CLI server memory, so we have quickly access to them.
```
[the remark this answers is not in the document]
spena: @colinhicks do we need to support Cloud? If we don't support REST API then we can't support it. Only way to support REST and Cloud is to open a session in the server, send all variables to the server so they're stored in that temporal session, and replace variables in the server side. It's more involved, but we'll allow Cloud users to use it.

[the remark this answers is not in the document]
colinhicks: We can think of this as a feature of clients, where each client that supports string substitution applies any transformations before sending the SQL to the server.

The Confluent Cloud UI is a ksqlDB client that can implement support for substitution at some point. It does not need to support it initially.

design-proposals/klip-38-variable-substitution.md
```diff
@@ -0,0 +1,288 @@
...
+
+Substitution will not attempt to add single-quotes to the values. Users will need to know the data type to use when using a variable. The ksqlDB
+server will be in charge of parsing and failing if the type is not allowed.
+
+Note: Variables are not case sensitive. A reference to `{replicas}` is the same as `{REPLICAS}`.
+
+### 4. Context for substitution variables
+
+Variable substitution will be allowed only in specific SQL statements, and only in places where literals or hard-coded values can be used.
+Variables cannot be used as column-names, reserved keywords, etc.
```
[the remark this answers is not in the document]
colinhicks: I think supporting identifiers like source name, column name, in addition to literals would be best.

I we should probably not support arbitrary statement fragments.

design-proposals/klip-38-variable-substitution.md
```diff
@@ -0,0 +1,288 @@
...
+```
+ksql> DEFINE format = 'AVRO';
+ksql> DEFINE replicas = '3';
+ksql> CREATE STREAM stream1 (id INT) WITH (kafka_topic='stream1', value_format='{format}', replicas={replicas});
+```
+
+Substitution will not attempt to add single-quotes to the values. Users will need to know the data type to use when using a variable. The ksqlDB
+server will be in charge of parsing and failing if the type is not allowed.
+
+Note: Variables are not case sensitive. A reference to `{replicas}` is the same as `{REPLICAS}`.
```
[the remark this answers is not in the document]
colinhicks: I think that makes sense. For reference MySQL's variables are also case-insensitive.

design-proposals/klip-38-variable-substitution.md
colinhicks: +1 for `${}`.

design-proposals/klip-38-variable-substitution.md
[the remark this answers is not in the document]
spena: Done. I made a reference to support identifiers for column and source names.

design-proposals/klip-38-variable-substitution.md
spena: Done

design-proposals/klip-38-variable-substitution.md
```diff
@@ -0,0 +1,302 @@
...
+```
+
+Also, the use of single-quotes in `SET` server settings is tedious to type. So after removing them, we'll have two different sets:
+* `SET 'server_property' = 'server_value'`
+* `SET VAR variable = `variable_value'
+
+This looks confusing too. So we have to decide for a new syntax for substitution variables.
+
+Other DBs use different syntax to create user variables (DECLARE, DEFINE, \set, SET). For ksqlDB, I'll choose to use `DEFINE` and `UNDEFINE` which
+is what Oracle SQL*plus uses. Any user coming from this DB will immediately understand the command.
```
agavra, line 98: can you link to the Oracle docs here for future reference (is it https://docs.oracle.com/cd/E18283_01/server.112/e16604/ch_twelve017.htm ?)

Though if I had to choose, I think the SQL+ syntax here is a little unintuitive. MySQL: https://docs.oracle.com/cd/E17952_01/mysql-5.7-en/user-variables.html uses something that to me is more intuitive (and is similar to SqlServer)

design-proposals/klip-38-variable-substitution.md
```diff
@@ -0,0 +1,302 @@
...
+```
+
+The `SOURCE` command will work like a `RUN SCRIPT`. If other SQL statements are in the file, then those statements will also be
+executed, thus to avoid limiting users from using DDL/DML commands while defining variables at the same time from one single script.
+
+`SOURCE` will be used only on the CLI and headless mode. REST API will not accept it because files may be out of the scope of the server filesystem.
+
+### 3. Referencing substitution variables
+
+Variables will be referenced by wrapping the variable between `${}` characters (i.e. `${replicas}`).
```
agavra, line 193: If we're following Oracle SQL+ why not use `&` or `&&` to reference the variables?

design-proposals/klip-38-variable-substitution.md
```diff
@@ -0,0 +1,302 @@
...
+When using CLI, we'll use the `SET CLI VARIABLE-SUBSTITUTION (ON|OFF)` command to enable/disable subsitution. Default will be `ON`.
+```
+ksql> SET CLI VARIABLE-SUBSTITUTION OFF;
+ksql> CREATE STREAM ... WITH (kafka_topic='${topic_${env}}');
+Error: Fail because ${topic_${env}} topic name is invalid. 
+```
+The error message is just an example of what ksqlDB will display.
+
+When using SQL scripts in headless mode, we need to provide a server setting to enable/disable variable substitution. The setting will be
+`ksql.enable.headless.variable.substitution` as Boolean (default = true). The reason is that `SET CLI` is a CLI-specific command not available
```
agavra, line 267:
```suggestion
`ksql.headless.variable.substitution.enable` as Boolean (default = true). The reason is that `SET CLI` is a CLI-specific command not available
```

design-proposals/klip-38-variable-substitution.md
```diff
@@ -0,0 +1,302 @@
...
+server will be in charge of parsing and failing if the type is not allowed.
+
+Note: Variables are case-insensitive. A reference to `${replicas}` is the same as `${REPLICAS}`.
+
+### 4. Context for substitution variables
+
+Variable substitution will be allowed in specific SQL statements. They can be used to replace text and non-text literals, and identifiers such as
+column names and stream/table names. Variables cannot be used as reserved keywords.
+
+For instance:
```
agavra, line 214: what if I don't want a literal `${foo}` to be replaced? e.g. I actually want `${foo}` to be the value in the statement (`SELECT '${foo}' as dolla_foo FROM bar`)? do i need to turn substitution off and then on after I issue the statement?

agavra (commented):
This LGTM! Some questions in line before green check

design-proposals/klip-38-variable-substitution.md
spena: Oracle SQL*Plus
https://blogs.oracle.com/opal/sqlplus-101-substitution-variables

I wanted to use a similar character like Mysql. I also like the `@`, but the `StringSubstitutor` library doesn't allow me to us only the @. I have to add a suffix to the variable, so a `@var@` or `:var:` would work.

spena: And regarding the `SET`. I didn't want to mix the current SET syntax which uses single-quotes for keynames, and I wasn't sure what other keyword to add to make reference to user-variables. Any proposal?

```
ksql> SET 'auto.offset.reset' = 'earliest';
ksql> SET VAR var1 = 'hola';
```

design-proposals/klip-38-variable-substitution.md
spena: In Oracle, the `&` prompts users to type the variable value when not defined, and `&&` to reuse the variable after the user has typed a value. That behavior is out of scope, so we either use `&` or something else. And I had the problem with the `StringSubstitutor` that requires a suffix too, so `&var&`? 

Either we write our own string parsing or get in love with some characters to wrap the variables names :).

Btw, the `${var}` is used in Hive, Impala, Spark and Presto. We're not alone. They use `SET`, though, for setting variables, but they do not override server settings or they use a different name for user variables. For instance, Hive uses `SET hiveconf:var1 = 'value'` and it is referenced as `${hiveconf:var1}` just to differentiate between user variables and system overrides `${system:varSys1}`.

design-proposals/klip-38-variable-substitution.md
spena: Done.

design-proposals/klip-38-variable-substitution.md
```diff
@@ -0,0 +1,295 @@
+# KLIP 34 - Variable Substitution
```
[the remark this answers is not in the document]
spena: Done

design-proposals/klip-38-variable-substitution.md
spena: Yes,
```
ksql> set variable-substitution off;
ksql> SELECT '${foo}' as dolla_foo FROM bar;
ksql> set variable-substitution on;
```

I was playing with escape characters, but the `\${foo}` fails, and the `\\${foo}` results in `SELECT '\${foo}'`. 

I wanted to be fancier and behave like Mysql and Posgtres. They do not substitute variables inside quotes. For instance:
```
ksql> CREATE ... WITH (kafka_topic=@topicName);
or
ksql> CREATE ... WITH (kafka_topic=:topicName);
```
I would have to trick the `StringSubstitutor` a little bit to reject replacing variables inside quotes, it's not a problem. The problem I had was how to replace this:
```
ksql> define colName = 'name';
ksql> SELECT id, ${colName} FROM stream;
```
would become:
```
ksql> SELECT id, "name" FROM stream;   // valid statement 
or
ksql> SELECT id, name FROM stream;     // valid statement
```
Hard to know.  You can do prepared statements in Mysql, but that's very hacky.
https://stackoverflow.com/questions/4428761/mysql-field-name-from-variable

design-proposals/klip-38-variable-substitution.md
```diff
@@ -0,0 +1,302 @@
...
+Variables will live during the session of a ksqlDB connection, then destroyed when the connection is closed. Only the user
+in that session can interact with the defined substitution variables. No other users can interact with other users' variables.
+
+In CLI, the session is opened when starting the CLI with the `ksql` command, and destroyed when typing `exit`.
+
+In Java API client, the session is opened when `Client.create()` is called, and closed when `Client.close()` is called. Variables
+may be defined using a client method instead of a new syntax. This is to ensure users know the scope starts after the `create()` call.
+
+TBD: Define a Java API method for variable substitution.
+
```
vcrfxia, line 164: Do we have plans to support variable substitution in the UI? The UI handles setting properties differently from the CLI: rather than the SET/UNSET syntax for session properties supported by the CLI, query properties instead have to be set using the dropdown in the UI editor. This divergence from the CLI has confused users in the past and I worry that having a different mechanism for defining variables in the UI will lead to similar confusion.

design-proposals/klip-38-variable-substitution.md
```diff
@@ -0,0 +1,302 @@
...
+Variables will live during the session of a ksqlDB connection, then destroyed when the connection is closed. Only the user
+in that session can interact with the defined substitution variables. No other users can interact with other users' variables.
+
+In CLI, the session is opened when starting the CLI with the `ksql` command, and destroyed when typing `exit`.
+
+In Java API client, the session is opened when `Client.create()` is called, and closed when `Client.close()` is called. Variables
+may be defined using a client method instead of a new syntax. This is to ensure users know the scope starts after the `create()` call.
+
+TBD: Define a Java API method for variable substitution.
```
vcrfxia, line 163: +1 to defining a new method for this rather than trying to piggyback off existing methods. If we don't support variable substitution in the REST API, the Java client will have to maintain variables in its own code, similar to the CLI. If we updated the CLI to use the Java client (https://github.com/confluentinc/ksql/issues/6269) we'd have fewer issues like these in the future but that's out of scope for this work for sure.

design-proposals/klip-38-variable-substitution.md
```diff
@@ -0,0 +1,302 @@
...
+
+To allow users load variables and keep their scope in the session, then the new `SOURCE` command will be added. This will override
+any previously defined variables, and will be destroyed until the user leaves the session.
+
+Syntax:
+```
+SOURCE '<script_file>';
+```
+
+The `SOURCE` command will work like a `RUN SCRIPT`. If other SQL statements are in the file, then those statements will also be
```
vcrfxia, line 186: What's the benefit in having `SOURCE` allow users to execute statements similar to `RUN SCRIPT`? It seems confusing to have two ways to do this with only a very subtle distinction (variables set in `SOURCE` persist into the sessions outside the scope of the command while those for `RUN SCRIPT` don't). Can we have `SOURCE` _only_ load variables from a file, and then ask users to use `RUN SCRIPT` afterwards if they want to also load statements?

design-proposals/klip-38-variable-substitution.md
```diff
@@ -0,0 +1,302 @@
...
+
+Substitution will not attempt to add single-quotes to the values. Users will need to know the data type to use when using a variable. The ksqlDB
+server will be in charge of parsing and failing if the type is not allowed.
+
+Note: Variables are case-insensitive. A reference to `${replicas}` is the same as `${REPLICAS}`.
+
+### 4. Context for substitution variables
+
+Variable substitution will be allowed in specific SQL statements. They can be used to replace text and non-text literals, and identifiers such as
+column names and stream/table names. Variables cannot be used as reserved keywords.
```
vcrfxia, line 212: Is it normal for variables to not be allowed as keywords? I don't mind but if we go with this we need to make sure we have good error messages since a natural extension (unless it's just my naivete speaking?) from the example below:
```
ksql> CREATE STREAM ${streamName} (${colName1} INT, ${colName2} STRING) \
      WITH (kafka_topic='${topicName}', format='${format}', replicas=${replicas}, ...);
```
is to wonder about
```
ksql> CREATE STREAM ${streamName} (${colName1} ${colType1}, ${colName2} ${colType2}) \
      WITH (kafka_topic='${topicName}', format='${format}', replicas=${replicas}, ...);
```
which won't be accepted based on this proposal. Hopefully the error message makes it very clear why.

design-proposals/klip-38-variable-substitution.md
```diff
@@ -0,0 +1,302 @@
...
+
+For instance:
+
+Note: `{VAR}` is used a reference where variables are permitted.
+```
+ksql> CREATE STREAM ${streamName} (${colName1} INT, ${colName2} STRING) \
+      WITH (kafka_topic='${topicName}', format='${format}', replicas=${replicas}, ...);
+      
+ksql> INSERT INTO ${streamName} (${colName1}, ${colName2}) \
+      VALUES (${val1}, '${val2}');
```
vcrfxia, line 222: It's interesting to me that the contents of variables are substituted into the statement directly, and it's left to the user to get the quoting of the variables correct. This makes sense but it'd be great to have helpful error messages when this goes wrong. For example, if the user mistakenly submits
```
ksql> CREATE STREAM ${streamName} (${colName1} INT, ${colName2} STRING) \
      WITH (kafka_topic=${topicName}, format='${format}', replicas=${replicas}, ...);
```
instead of
```
ksql> CREATE STREAM ${streamName} (${colName1} INT, ${colName2} STRING) \
      WITH (kafka_topic='${topicName}', format='${format}', replicas=${replicas}, ...);
```
where the difference is that they're missing quotes around the topic name, it'd be great to have a useful error message so they aren't left guessing in the dark.

design-proposals/klip-38-variable-substitution.md
```diff
@@ -0,0 +1,302 @@
...
+
+ksql> SELECT * FROM ${streamName} \
+      WHERE ${colName1} == ${val1} and ${colName2} == '${val2}' EMIT CHANGES; 
+```
+Any attempt of using variables on non-permitted places will fail with the current SQL parsing error found when parsing the variable string.
```
vcrfxia, line 227: > fail with the current SQL parsing error found when parsing the variable string

As mentioned above, I think we'll want better error messages here. The SQL parsing errors we see today are awful in terms of debuggability and I worry that variable substitution will make this all the more confusing. Can we make improving these error messages in scope for this work, or is that too much?

vcrfxia (commented):
Cool stuff @spena ! Comments/questions inline.

design-proposals/klip-38-variable-substitution.md
agavra: If there's precedent (Hive, Impala, Spark and Presto) I think we can stick with your proposal as users will be accustomed to this already.

design-proposals/klip-38-variable-substitution.md
agavra: I don't think implementation should drive the design. We should define what design we want here in the KLIP, and then if it's tough to implement we can issue a known bug report and fix it later.

cc @colinhicks do you think we need to hammer this issue out before we approve the KLIP?

design-proposals/klip-38-variable-substitution.md
spena: Yes, it's gonna be different in the UI than the CLI. I raised this question to @colinhicks too, and seems we'll have this feature separately from the UI. The reason? We'll have to work on opening temporary sessions between the client/server connections, then send variables to the server and stored them temporary there, then do the replacement in the server. That's more involved work.

We might need to do these changes in the future to make better UX for users.

design-proposals/klip-38-variable-substitution.md
spena: We could that too. I'm not sure how users will use their scripts, though. Are they willing to split their scripts between vars and commands?

I just though of using the bash behavior:
```
$ cat /tmp/s1
export ID=1
echo id=$ID

$ cat /tmp/s2
export ID=2
echo id=$ID

$ source /tmp/s1
id=1
$ bash /tmp/s2
id=2

// Uses the ID set in by the source command
$ echo $ID
1
```

Mysql uses SOURCE too, and it allows variables and commands.
https://dev.mysql.com/doc/refman/5.7/en/mysql-batch-commands.html

design-proposals/klip-38-variable-substitution.md
spena: That really depends. Some big-data DBs allow any substitution, but other DBs (mysql, oracle, ...) do not allow replacing keywords. They have a different way to do that, though. 

Here's the discussion with @colinhicks about it.
https://github.com/confluentinc/ksql/pull/6259#discussion_r491181382

design-proposals/klip-38-variable-substitution.md
spena: Right. I'll try to add a customized error. I was thinking on leverage the parsing errors, which displays an error when using `${topicName}`. I'll see if the second parser can detect those undesired places.

design-proposals/klip-38-variable-substitution.md
colinhicks:
```
ksql> set variable-substitution off;
ksql> SELECT '${foo}' as dolla_foo FROM bar;
ksql> set variable-substitution on;
```

This seems like a reasonable limitation to me.

design-proposals/klip-38-variable-substitution.md
agavra: Does `MySQL` also have a `RUN SCRIPT`? I tend to agree with Victoria that this seems a little confusing.

design-proposals/klip-38-variable-substitution.md
agavra: I think given the goal (running these scripts in different environments) it's probably OK to not allow the types

agavra (approved):
LGTM! This is an exciting feature in the CI/CD realm

design-proposals/klip-38-variable-substitution.md
vcrfxia: Agreed it's fine to not allow keyword substitutions -- as long as we have good error messages. The error messages on syntax errors we have today are pretty abysmal... :)

derekjn:
Overall I think this is a good approach here. The only thing I'd like to raise is that I think that if we're using syntax for this, then substitution should happen server side. I think that processing syntax on the client could lead to confusion for users who are using HTTP or their own client implementation. But fundamentally, I feel that syntax should be completely independent of what client is being used.

spena:
@derekjn I was thinking on the scenario of users using raw HTTP requests. If we support variable substitution in the server-side, then they should work as the `SET` statements, which work only per request.
```
{ 
   statement: "
        DEFINE x = 'stream';

        SET 'auto.offset.reset' = 'earliest';
        SELECT * FROM ${x} EMIT CHANGES;
    "
}
```

This is for simple cases. Users can even read all statements from a file and make one single HTTP request. I don't know if that's something users are used to do, though. I would make one HTTP request per line if it was me.

Anyway, we should still support variable substitution in the CLI side. The CLI sends one request at a time. Otherwise, we'll have to send all the current variables on every request made by the user.

One advantage supporting server-side substitution might be when using the UI. I need to confirm it works like this, but I assume that the UI sends all statements in the UI textbox in one request. Like:
```
DEFINE x = 'stream';
SET 'auto.offset.reset' = 'earliest';
SELECT * FROM ${x} EMIT CHANGES;
```
If the above is sent as one request by the UI, then we'll have automatic support for the UI :). But need to confirm if that's the case, otherwise the UI will still have the same problem.

Btw, we need to add a configuration to override variable substitution in the HTTP request too. Like:
```
{ 
   statement: "
        DEFINE x = 'stream';

        SET 'ksql.variable.substitution.enable' = 'false'
        SELECT * FROM ${x} EMIT CHANGES;  // This fill fail 'cause ${x} is no substituted
    "
}
```

@colinhicks what do you think?

colinhicks:
> Anyway, we should still support variable substitution in the CLI side. The CLI sends one request at a time. Otherwise, we'll have to send all the current variables on every request made by the user.

Does this mean both both client-side substitution and server-side substitution apply to the CLI? Or would the CLI always use client-side substitution and not pass variables to the server?

spena:
@derekjn I added more details to how variable substitution will happen in the server-side. You were right about the new syntax introduction. We should allow variables in SQL scripts, and let users execute the script from any client they want, such as load the file in Java and send one HTTP request with all statements, or copy/paste a script file on the UI.

This is what I changed if you want to review it:

* Use of triple-quotes to escape single-quotes (see Creating variables)
* Add printing variables section only for CLI (see Printing variables)
* Added a context/scope for variable substitution in raw HTTP requests
* Replace `set cli variable-substutition` for `set 'ksql.variable.substitution.enable'` now that server-side substitution
  is the way to go.
* Edited the implementation section to explain how CLI will send requests for substitution

derekjn:
@spena LGTM 👍

colinhicks (approved):

spena:
Tests failure unrelated to the doc

PR #6261, opened 2020-09-18
refactor: introduce SessionConfig

big-andy-coates:
### Description 

There's a fair amount of code that passes around `KsqlConfig`, holding the system config, paired with a `Map<String, Object>`, containing the overrides the user supplied with their request.  Recipients of these parameters can apply the overrides to the config to get a new `KsqlConfig` instance with the overrides applied by calling `cloneWithPropertyOverwrite`.  However, I've fixed a few bugs previously where this was missing, or simplified it where was done multiple times.

To address these, I've bitten the bullet and spent and hour introducing a `SessionConfig` class, which holds the system config and overrides.  This type can be passed where ever the two configs were previously being passed.

The `getConfig` method takes a boolean control whether overrides are applied or not.  I chose this over two separate methods, e.g. `getSystemConfig` and `getConfigWithOverrides`, as my hope is that the boolean will force anyone writing new code to think about whether they want the config with or without overrides.

This change should change any functionality.

### Testing done 

Usual

### Reviewer checklist


spena (approved):
LGTM. Any plans to add a check inside the `SessionConfig.of()` all overrides are not part of the denied list of overrides?

big-andy-coates:
> LGTM. Any plans to add a check inside the `SessionConfig.of()` all overrides are not part of the denied list of overrides?

No at the mo, but it's adds a class where this could be added later.

Issue #4981, opened 2020-04-02
Add CLI parameter to execute commands without KSQL interative


spena:
**Is your feature request related to a problem? Please describe.**

I sometimes want to execute a single statement in KSQL, i.e list current streams, without going into the KSQL command line,  but can't find how. The only option is to open the CLI, and wait until it initializes, then run the command.

**Describe the solution you'd like**

I'd like to use a CLI parameter, like Mysql and other databases do, to execute a command without going into the KSQL command line. This will allow me to write scripts to get info from KSQL.

Example:
```
$ ksql -e "show streams;" | grep -v KSQL_PROCESSING_LOG | awk '{print $1}' | sed '/^\s*$/d' | tail +2
PS1
s1
```

**Describe alternatives you've considered**

There are no alternatives.

**Additional context**

blueedgenick:
big +1 from me. In fact, this used to work, not sure where along the line it got regressed.

PR #6267, opened 2020-09-21
feat: new CLI parameter to execute a command and quit (without CLI interaction)

spena:
### Description 
Fixes https://github.com/confluentinc/ksql/issues/4981

Adds a new paramter to the CLI to execute a command and quit. This without having to interact with the CLI console.
```
sergio@pop-os:~/Development/ksql$ time ./bin/ksql -e "show streams;"

 Stream Name         | Kafka Topic                 | Format 
------------------------------------------------------------
 KSQL_PROCESSING_LOG | default_ksql_processing_log | JSON   
 TOPIC1              | topic1                      | JSON   
------------------------------------------------------------

real	0m2.250s
user	0m4.604s
sys	0m0.185s
```

```
sergio@pop-os:~/Development/ksql$ ./bin/ksql -e "set 'auto.offset.reset'='earliest'; select * from topic1 emit changes;"
Successfully changed local property 'auto.offset.reset' to 'earliest'. Use the UNSET command to revert your change.
+---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+
|ID                                                                                                                                                                                                                               |
+---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+
|1                                                                                                                                                                                                                                |
^CQuery terminated
```

You can use a different output format to easily parse the command result. For instance, JSON:
```
$ ./bin/ksql -e "show streams" --output JSON | jq '.[].streams' | jq '.[].name'
"TOPIC1"
"KSQL_PROCESSING_LOG"
"S1"
```

### Testing done 
Added unit tests
Verified manually

### Reviewer checklist


colinhicks:
> Adds a new paramter to the CLI to execute a command and quit. This without having to interact with the CLI console.

Just to clarify: For streaming results like push queries without a limit clause, you need to also issue ctrl+c to interrupt the console, correct?

colinhicks:
> * Add new parameter to remove headers and columns from the command output
> * Add new parameter to change command output format (i.e. comma-separated and tab-separated)

It might be worth considering a flag to send the raw JSON to stdout. This way it could be piped to other scripts like `jq` for such formatting.

spena:
@colinhicks Ctrl+C is part of the stream. The example of the Description shows the `^Query terminated` line that was caused by the Ctrl+C.

ksqldb-cli/src/main/java/io/confluent/ksql/cli/Options.java
```diff
@@ -111,6 +113,12 @@
           + "defaults to TABULAR)")
   private String outputFormat = OutputFormat.TABULAR.name();
 
+  @SuppressWarnings("unused") // Accessed via reflection
+  @Option(
+      name = {EXECUTE_OPTION, EXECUTE_SHORT_OPTION},
+      description = "Execute a command and quit.")
```
rodesai, line 119: nit: the user may not understand "command" - maybe "run one or more ksql statements and quit."?

ksqldb-cli/src/main/java/io/confluent/ksql/cli/Cli.java
```diff
@@ -179,13 +182,31 @@ public void makeKsqlRequest(final String statements) {
...
+    RemoteServerSpecificCommand.validateClient(terminal.writer(), restClient);
+    try {
+      // Handles the RUN SCRIPT command if found
+      if (!terminal.maybeHandleCliSpecificCommands(command)) {
+        handleLine(command);
+      }
+    } catch (final EndOfFileException exception) {
+      // Ignore - only used by runInteractively() to exit the CLI
+    } catch (final Exception exception) {
+      LOGGER.error("", exception);
```
rodesai, line 195: why the empty log message?

rodesai (commented):
LGTM other than a couple minor comments inline. Also, is there a place in the docs we can add this?

ksqldb-cli/src/main/java/io/confluent/ksql/cli/Options.java
spena: Done.

ksqldb-cli/src/main/java/io/confluent/ksql/cli/Cli.java
spena: Done.

agavra (approved):
LGTM, I think @rodesai meant to check the green tick

PR #6268, opened 2020-09-22
docs: Update README.md

gAmUssA:
### Description 

fix a broken link in KLIP 18

### Testing done 

previewed markdown and link looks ok

### Reviewer checklist


spena (approved):
LGTM. Thanks for the fix

spena:
I don't know why the PR semantic check failed. Everything looks good. I will merge it.

vcrfxia:
> I don't know why the PR semantic check failed.

The reason the check failed is explained on the check: "PR has only one non-merge commit and it's not semantic". The reason this fails the build is because when there is only one commit in a PR, the default commit message that is used when merging the PR is the message of the single commit, not the PR title (as is the case for PRs with multiple commits). So, if the single commit is not semantic, then the default commit message is not semantic. If you don't want to add another commit or update the message on the existing commit, you can still merge the PR but make sure to override the default commit message with the PR title. That was not done in this case so the final merge commit was not semantic: https://github.com/confluentinc/ksql/commit/7455bd766a25c687b2bbc27e1a22399a521109de

No harm done in this case since docs commits don't show up in our changelog anyway. Just thought I'd explain what happened and the reasoning for future reference.

spena:
Oh, thanks @vcrfxia for the explanation. 
I didn't notice I didn't add the `docs:` when I merge the PR. I got used to it uses the PR title.

PR #6270, opened 2020-09-22
refactor: Format interface

big-andy-coates:
### Description 

In prep for user-defined-formats and just to tidy up the interface and fix some bugs while I had a spare moment...

## Refactor

Refactored the `Format` interface so that the functions used during schema inference, i.e. `toParsedSchema` and `toColumns`, are moved onto a new interface `SchemaTranslator`. The `SchemaTranslator` is obtained from the `Format` by calling `getSchemaTranslator`, passing in the format properties.

This removes the strange requirement of having to pass a `FormatInfo`, which contains the name of the format, to the `toParsedSchema` method.

And ensures consistent properties are used for both `toParsedSchema` and `toColumns`. (Though at the moment properties are only needed in `toParsedSchema`, that may not be the case once user-definfed-formats are supported.

Switched the `supportsSchemaInference` method on `Format` to a `SerdeFeature`, reducing the surface of the interface.

## Fixes / Enhancements
Fixed a bug / ux issue, where a C* statement that was using schema inference would not report any issues if the supplied `VALUE_FORMAT` did not match the type of the schema in the schema registry. For example, if `VALUE_FORMAT` was `AVRO`, but the schema returned by the Schema Registry was `JsonSchema`, then the statement would succeed... and would later generate lots of deserialisation errors if used. The statement will now fail.

### Testing done 

Usual

### Reviewer checklist


ksqldb-common/src/main/java/io/confluent/ksql/serde/SerdeFeature.java
```diff
@@ -24,6 +24,15 @@
...
 public enum SerdeFeature {
 
+  /**
+   * The format supports interaction with the Confluent Schema Registry.
+   *
+   * <p>Indicates whether or not a format can support CREATE statements that
+   * omit the table elements and instead determine the schema from a Confluent Schema Registry
+   * query.
+   */
+  SCHEMA_INFERENCE,
```
agavra, line 34: 💯

ksqldb-serde/src/main/java/io/confluent/ksql/serde/FormatProperties.java
```diff
@@ -0,0 +1,51 @@
...
+  private FormatProperties() {
+  }
+
+  public static void validateProperties(
+      final String formatName,
+      final Map<String, String> properties,
+      final Set<String> supportedProperties
+  ) {
+    // by default, this method ensures that there are no property names
+    // (case-insensitive) that are not in the getSupportedProperties()
```
agavra, line 38: nit: comment needs to be updated `getSupportedProperties()` is no longer a method call here

agavra (approved):
I was worried when I saw a 1k line refactor, but the fact that it was so easy to review is a testament to a good refactor 😂 LGTM

ksqldb-serde/src/main/java/io/confluent/ksql/serde/FormatProperties.java
big-andy-coates: updated.

ksqldb-common/src/main/java/io/confluent/ksql/serde/SerdeFeature.java
big-andy-coates: We could even use this to get rid of `JSON_SR`, or the very least the `JsonSchemaFormat` class.

We could choose to get rid of `JSON_SR` by adding a `WITH` property that can be used with the `JSON` format that indicates it should use the SR, e.g. 

```sql
CREATE STREAM FOO WITH (format='json', use_schema_registry=true);
```

Even if we don't do that, we can internally translate any use of `JSON_SR` through to using the `JsonFormat` class with the `SCHEMA_INFERENCE` feature set.

Thoughts?

big-andy-coates:
Thanks for the review @agavra

ksqldb-common/src/main/java/io/confluent/ksql/serde/SerdeFeature.java
agavra: We had discussed something like that with @derekjn and @MichaelDrogalis and decided that it would be the lesser of two evils (w.r.t. user confusion) if we just kept them as two separate formats to make it extra "clear" what the difference is. I'm always happy reducing the code overhead though if we can use it to get rid of `JsonSchemaFormat` class

PR #6273, opened 2020-09-22
test: support source value format in QTT post conditions (MINOR)

vcrfxia:
### Description 

Addresses https://github.com/confluentinc/ksql/pull/6194#discussion_r490670303 by adding value format into the sources post-condition check of QTT, and updating the new tests to use the sources post-condition check rather than the topics post-condition check.

### Testing done 

Test-only change. Verified that tests fail if the specified source value format does not match what's expected.

### Reviewer checklist


big-andy-coates (approved):
LGTM!

Issue #6216, opened 2020-09-16
[JSON Support]: Add RQTT tests to test JSON key format


big-andy-coates:
1. Enhance RQTT to support JSON key formats. 
1. Add  RQTT tests for different JSON key format (which shouldn't be affected, but let's ensure by testing):
  - `INSERT VALUES` in `insert-values.json`
  - Pull queries in `pull-queries-against-materialized-aggregates.json`
   - Push queries in `push-query.json`

PR #6274, opened 2020-09-22
test: add RQTT for non-KAFKA key format (MINOR)

vcrfxia:
### Description 

Fixes https://github.com/confluentinc/ksql/issues/6216

Add RQTT tests with non-Kafka formats. No changes were needed to the RQTT framework. This PR just adds a few tests as sanity checks going forward.

The tests aren't super useful as sanity checks since RQTT doesn't support the post-conditions check that QTT does which would allow us to verify the key and value formats of the created sources. This means an (unlikely) bug could be introduced into the RQTT executor which always hardcodes a particular key format and these tests would still pass. I don't think it's worth introducing a post-conditions check to RQTT, though, since that's the job of QTT not RQTT.

### Testing done 

Tests pass.

### Reviewer checklist

### DIFF
--- a/ksqldb-functional-tests/src/test/resources/rest-query-validation-tests/insert-values.json
+++ b/ksqldb-functional-tests/src/test/resources/rest-query-validation-tests/insert-values.json
@@ -383,6 +383,21 @@
       "responses": [
         {"admin": {"@type": "currentStatus"}}
       ]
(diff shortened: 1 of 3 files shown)


big-andy-coates (approved):
LGTM!

PR #6278, opened 2020-09-22
fix: don't use queryId of last terminate command after restore

### LINKED ISSUE #6010: Query Id doesn't use correct offset if the last query in command topic has an associated terminate command
**Describe the bug**
If the last query in the command topic is terminated and then the server is restarted, we don't correctly set the queryIdGenerator to the correct offset for the next query that's created.

**To Reproduce**
Start a ksqlDB with a new command topic
```
ksql> create stream test as select * from KSQL_PROCESSING_LOG;

 Message                           
 Created query with ID CSAS_TEST_0 
ksql> create stream test1 as select * from test;

 Message                            
------------------------------------
 Created query with ID CSAS_TEST1_3 
------------------------------------
ksql> TERMINATE CSAS_TEST1_3;

 Message           
-------------------
 Query terminated. 
-------------------
```
Restart server
```
ksql> create stream test2 as select * from test1;

 Message                            
------------------------------------
 Created query with ID CSAS_TEST2_3 
------------------------------------
```

**Expected behavior**
The queryId should be greater than 3 since it was already used for the CSAS for `test1`

**Actual behaviour**
The queryId 3 is used again. This violates the guarantee that query id is monotonically increasing as commands are put into the command topic.

**Additional context**
https://github.com/confluentinc/ksql/pull/5002 removed the query plans from `CSAS/CTAS/INSERT` statements that were terminated later in the command topic. That means when we execute these command, we don't return a `QueryMetadata` in the `ExecuteResult` from the `KsqlEngine` since there's no QueryPlan.

https://github.com/confluentinc/ksql/blob/master/ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/computation/InteractiveStatementExecutor.java#L243
We only set the queryID to the current offset if the ExecuteResult has a query in it. Therefore, if the last command that created a query is terminated, when the server restarts, the next QueryId wouldn't be set to the correct value.

This only affects the next query created after the restart, after it's created, we'll set the QueryIdGenerator to the correct offset.

Having duplicate query id's could lead to data corruption as we use the query id for both internal topics and schema names.

vpapavas:
### Description 
Fixes #6010 
The problem is that if the last command before a restart was a terminate, then the first command after the restart gets the same queryId as the terminate. Moreover, if the command topic contains only terminate commands, the next queryId will be 0. To avoid this, we always increase the queryId for every command, whether it has a plan or not

### Testing done 
Unit tests.

Manual testing
```
ksql> CREATE STREAM A (COLUMN STRING) WITH (KAFKA_TOPIC='A', VALUE_FORMAT='JSON', partitions=1);

 Message
----------------
 Stream created
----------------
ksql> CREATE STREAM B AS SELECT * FROM A;

 Message
--------------------------------
 Created query with ID CSAS_B_3
--------------------------------
ksql> CREATE STREAM C AS SELECT * FROM A;

 Message
--------------------------------
 Created query with ID CSAS_C_5
--------------------------------
ksql> terminate CSAS_B_3;

 Message
-------------------
 Query terminated.
-------------------
ksql> terminate CSAS_C_5;

 Message
-------------------
 Query terminated.
-------------------
```
Restart
```
ksql> CREATE STREAM D AS SELECT * FROM A;

 Message
--------------------------------
 Created query with ID CSAS_D_7
--------------------------------
```
### Reviewer checklist


stevenpyzhang:
@vpapavas can you try this test scenario? I'm not sure this is fixing the issue.

Delete any existing command topic
Startup server
```
create stream test as select * from KSQL_PROCESSING_LOG;
terminate CSAS_TEST_0;
drop stream test;
```
Stop the server. Start the server up.
```
create stream test as select level from KSQL_PROCESSING_LOG;
```
I think even with this change, we'd duplicate the `CSAS_TEST_0` as the QueryId

ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/computation/RestoreCommandsCompactor.java
```diff
@@ -24,12 +24,17 @@
...
 import javax.annotation.Nullable;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 /**
  * Util for compacting the restore commands
  */
 public final class RestoreCommandsCompactor {
 
+  static QueryId greatestQueryId;
```
agavra, line 35: making this static is a design smell. if we need to access information from within this class we have two options (1) we can make `compact` return an object that is a pair of `greatestQueryId` and the `List<QueuedCommand>` (2) we can make `RestoreCommandsCompactor` non-static, and create the instance where we need it

you can read along https://stackoverflow.com/questions/7026507/why-are-static-variables-considered-evil for more reasons why

ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/computation/RestoreCommandsCompactor.java
```diff
@@ -86,6 +91,8 @@ public static CompactedNode maybeAppend(
       if (queued.getAndDeserializeCommandId().getType() == Type.TERMINATE) {
         final QueryId queryId = new QueryId(queued.getAndDeserializeCommandId().getEntity());
         markShouldSkip(queryId, latestNodeWithId);
+        //keep track of the last terminate command
+        //lastTerminateQueryId = queryId;
```
agavra, line 95: nit: not used

ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/computation/InteractiveStatementExecutor.java
```diff
@@ -241,6 +254,13 @@ private void executePlan(
         new CommandStatus(CommandStatus.Status.EXECUTING, "Executing statement")
     );
     final ExecuteResult result = ksqlEngine.execute(serviceContext, configured);
+    long queryID = Long.MIN_VALUE;
+    if (greatestQueryId.isPresent() && mode == Mode.RESTORE) {
+      final String ltq = greatestQueryId.get().toString();
+      final int lastIndex = ltq.lastIndexOf("_");
+      queryID = Long.parseLong(ltq.substring(lastIndex + 1));
```
agavra, line 261: this is breaking an abstraction barrier - we're relying on internal implementations of the string representation of the `QueryId` which doesn't guarantee that it will match the format that you assume here

agavra (changes requested):
Thanks for the fix Vicky! The test coverage gives me more confidence that this works.

The change, however, I think is fixing the issue with a band-aid instead of fixing it at the root. We're adding more edge cases to the code and making the queryID generation more difficult to reason about. I think we should look for another solution that simplifies queryID generation instead of adding more conditions. Let's chat offline if you need some help coming up with solutions.

ksqldb-rest-app/src/test/java/io/confluent/ksql/rest/server/computation/CommandRunnerTest.java
```diff
@@ -160,9 +159,9 @@ public void shouldRunThePriorCommandsCorrectly() {
 
     // Then:
     final InOrder inOrder = inOrder(statementExecutor);
-    inOrder.verify(statementExecutor).handleRestore(eq(queuedCommand1));
-    inOrder.verify(statementExecutor).handleRestore(eq(queuedCommand2));
-    inOrder.verify(statementExecutor).handleRestore(eq(queuedCommand3));
+    inOrder.verify(statementExecutor).handleRestore(eq(queuedCommand1), any());
```
agavra, line 162: I'm guessing you didn't mean to keep all of these? This PR doesn't change the `handleRestore` syntax anymore (same throughout the PR)

ksqldb-rest-app/src/test/java/io/confluent/ksql/rest/server/computation/RecoveryTest.java
```diff
@@ -707,9 +704,42 @@ public void shouldRecoverQueryIDs() {
...
+    server.recover();
+    server.submitCommands("CREATE STREAM C AS SELECT * FROM A;");
+    final Set<QueryId> queryIdNames = queriesById(server.ksqlEngine.getPersistentQueries())
+        .keySet();
+
+    assertThat(queryIdNames, contains(new QueryId("CSAS_C_2")));
+  }
+
+  @Test
+  public void shouldIncrementQueryIDsWithPlan() {
```
agavra, line 727: I feel like I'm blind, what's the difference between these two tests?

agavra (approved):
LGTM! It looks like you accidentally left some of the previous commit's work (`handleRestore`). Just for anyone reading this discussion, I have thought about whether this is backwards compatible and i think it is because any old commands that used the old numbering (e.g. the first query was always `_0`) will have their queryIds persisted. It might break some people's automated tests in the worst case

ksqldb-rest-app/src/test/java/io/confluent/ksql/rest/server/computation/RecoveryTest.java
vpapavas: No you are not blind, I messes it up somehow. Fixed it though

agavra (approved):

PR #6279, opened 2020-09-22
test: fix error handling in QTT

big-andy-coates:
### Description 

I've noticed that the QTT tests aren't failing on the statement that caused the error. Rather it was executing the next statement, which generally failed with some 'unknown source' error. Very misleading.

The issue was that if a statement fails, the code calls `hasNext` on the iterator of test statements to ensure its the last statement that failed. (This is needed to ensure we don't have misleading tests with statements after an invalid one that are never executed).  However, the `hasNext` call was evaluating the next statement, which would likely fail due to the previous statement failing.

The fix is to make `hasNext` _not_ parse and prepare the statement.  This is delayed until `next` is called. This required moving the INSERT VALUES execution code, but that's a good thing anyway - that should never had been executed as part of `hasNext`!

### Testing done 

Usual

### Reviewer checklist


vcrfxia (approved):
Thanks for the fix -- LGTM.

PR #6285, opened 2020-09-23
test: correct bug in QTT that meant some tests were ignored

big-andy-coates:
### Description 

Some QTT test cases used `input` rather than `inputs` and `output` rather than `outputs`. Previously, QTT would have detected this mistake, but the check has been removed.

This change adds back in the check to ensure a test case provides at least one of:

- an expected exception
- input messages
- output messages

The change also corrects the bad tests and their historic plans

### Testing done 

Test only change.

### Reviewer checklist

### DIFF
--- a/ksqldb-functional-tests/src/main/java/io/confluent/ksql/test/model/TestCaseNode.java
+++ b/ksqldb-functional-tests/src/main/java/io/confluent/ksql/test/model/TestCaseNode.java
@@ -23,6 +23,7 @@ import com.fasterxml.jackson.annotation.JsonIgnoreProperties;
 import com.fasterxml.jackson.annotation.JsonProperty;
 import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableMap;
+import io.confluent.ksql.test.TestFrameworkException;
 import io.confluent.ksql.test.tools.exceptions.InvalidFieldException;
 import io.confluent.ksql.test.tools.exceptions.MissingFieldException;
 import java.util.List;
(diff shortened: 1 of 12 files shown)


ksqldb-functional-tests/src/main/java/io/confluent/ksql/test/model/TestCaseNode.java
```diff
@@ -133,12 +134,22 @@ public VersionBoundsNode versionBounds() {
...
     }
 
-    if (this.statements.isEmpty()) {
-      throw new InvalidFieldException("statements", "was empty");
+    try {
+      if (statements.isEmpty()) {
+        throw new InvalidFieldException("statements", "was empty");
+      }
+
+      if (!expectedException.isPresent() && inputs.isEmpty() && outputs.isEmpty()) {
```
vcrfxia, line 146: How come we only require one of inputs or outputs to be present? Under what scenario would it make sense to provide one and not the other?

vcrfxia (approved):
LGTM -- nice catch!

ksqldb-functional-tests/src/main/java/io/confluent/ksql/test/model/TestCaseNode.java
vcrfxia: I now understand why the input file may be null: we might have tests that start with no inputs and instead use `INSERT VALUES` to insert data, but I still don't understand why it'd be useful to have a test where the output file is null. FWIW it looks like the testing tool requires the output file to be present: https://github.com/confluentinc/ksql/blob/c5d6b56a3188f100ea473ef8f13959da79d1c79e/ksqldb-functional-tests/src/main/java/io/confluent/ksql/test/tools/KsqlTestingTool.java#L55

big-andy-coates: A test where a filter filters everything out.  Weird... yes... do we have such tests... yes.

vcrfxia: But shouldn't a test like that require an empty outputs node rather than a missing outputs node? Having a missing outputs node is counterintuitive since missing nodes in other parts of the test specification indicate that nothing should be checked, whereas in a test like that we would still want to verify that no outputs were produced.

PR #6260, opened 2020-09-18
docs: docs for new KEY_FORMAT property and associated server configs

vcrfxia:
### Description 

Porting over https://github.com/confluentinc/ksql/pull/6252 to a branch on the main repo rather than my personal fork, in preparation for working on this jointly with @big-andy-coates .

Opening a separate docs PR for the changes in https://github.com/confluentinc/ksql/pull/6194 as these docs cannot go live before the feature flag guarding https://github.com/confluentinc/ksql/pull/6194 is enabled by default.

### Testing done 

Docs-only change.

### Reviewer checklist
- [ ] Ensure docs are updated if necessary. (eg. if a user visible feature is being added or changed).
- [ ] Ensure relevant issues are linked (description should include text like "Fixes #<issue number>")

PR #6287, opened 2020-09-23
chore: expose default format configs from behind feature flag

vcrfxia:
### Description 

The `ksql.key.format.enabled` feature flag introduced in https://github.com/confluentinc/ksql/pull/6194 currently guards two features:
- the use of default key and value formats specified by the `ksql.persistence.default.format.key` and `ksql.persistence.default.format.value` configs
- whether certain key formats are enabled for general use or not

This PR updates the feature flag to only guard the latter and not the former. In other words, `VALUE_FORMAT` is not longer a required property. An error is thrown if `VALUE_FORMAT` is not provided and `ksql.persistence.default.format.value` is not set.

This PR also cleans up a few unnecessary usages of the feature flag, in light of the decision to allow the use of `KEY_FORMAT` and `FORMAT` even when the feature flag is not enabled.

Docs for this change are in https://github.com/confluentinc/ksql/pull/6260. However, we aren't planning on merging https://github.com/confluentinc/ksql/pull/6260 until 0.14.0 when JSON key format support is released. If we feel strongly about keeping the docs in sync, we can either hold on this PR until 0.14.0 or split the docs PR into one for 0.13.0 (for this change) and a separate one for 0.13.0. @big-andy-coates do you have a preference?

### Testing done 

Updated QTT.

### Reviewer checklist


agavra (approved):
LGTM! I think this makes sense, not sure why the config and defaulting need to be coupled in the current state as it's completely backwards compatible. I think the docs should be in sync to prevent people from getting confused (shouldn't be too much work to document the new default config?)

big-andy-coates (approved):
Well that simplifies things! LGTM!

Regarding the docs, I'm with Almog. If this is going out in 0.13, then the docs _and the release notes_ for 0.13 should cover the new configs, the addition of `KEY_FORMAT` and `FORMAT` (even if only `KAFKA` is supported).

NB: If we want this in the release notes, we shouldn't commit it as a `chore` commit - I'd merge as a `feat` commit.

vcrfxia:
@agavra 0.13 has already been cut from master, right? If so I'm going to merge this into master for 0.14 and not worry about splitting the docs PR.

agavra:
@vcrfxia yeah 0.13 has been cut 2753ccd7295d978e967ec309aa062334c16b36a1 was the last commit to make it in without cherry-picking

vcrfxia:
Thanks for confirming! Merging into master for 0.14. The linked docs PR will be merged for 0.14 as well.

Issue #6214, opened 2020-09-16
[JSON Support]: Support `DECIMAL` and `BOOLEAN` key types


big-andy-coates:
JSON supports `DECIMAL` and `BOOLEAN` types. 
1. Add support for these types are key columns.
1. Add suitable QTT tests, (`json.json` or `elements.json`)
1. Ensure `AS_VALUE` works for these new types.

apurvam:
I'm thinking that maybe we can leave this out of the initial releases. I am not sure if we know that people have DECIMAL / BOOLEAN in keys. if they do, we can add it as a followup feature.

apurvam:
cc @MichaelDrogalis

Issue #6215, opened 2020-09-16
[JSON Support]: add QTT tests for JSON key format


big-andy-coates:
Ensure full support for the JSON key format in all supported streaming SQL syntax by adding exhaustive QTT tests:

* Test `PARTITION BY` in `partition-by.json`, especially by newly supported `DECIMAL` and `BOOLEAN` types.
* Test `GROUP BY` in `group-by.json`, especially by newly supported `DECIMAL` and `BOOLEAN` types.
* Test declaring sources with JSON key types in `elements.json`
* Test windowed keys work with non-KAFKA key in `session-windows.json`, `tumbling-windows.json`, and `hopping-windows.json`
* `WHERE` and `HAVING` shouldn't be affected, but add a test with non-KAFKA key format to `having.json` and `where.json`.

PR #6292, opened 2020-09-24
Json keys

big-andy-coates:
### Description 

part of: #6215
fixes: #6214

First pass of adding QTT tests around the JSON key format, including adding tests for `DECIMAL` and `BOOLEAN` types, which is the first time they've been used as a key column type.

Commit includes a fix in `SerdeOptionsFactory` to ensure no key unwrapping is set if there is no key, plus refactored `SerdeOptionsFactory`, combining the two public methods.

### Testing done 

usual

### Reviewer checklist


ksqldb-functional-tests/src/test/resources/query-validation-tests/json.json
```diff
@@ -1,31 +1,426 @@
...
+        {"topic": "input_topic", "value": {"V": 1234567890}},
+        {"topic": "input_topic", "value": null}
+      ],
+      "outputs": [
+        {"topic": "OUTPUT", "value": 1234567890},
+        {"topic": "OUTPUT", "value": null}
+      ]
+    },
+    {
+      "name": "DOUBLE - key",
```
vcrfxia, line 154: The PR description says this test is currently disabled. How so?

vcrfxia (approved):
LGTM though I think it's worth adding a few QTTs for converting between different key formats, both positive and negative such as validating what happens if I start with a JSON boolean key and try converting the key format to KAFKA.

ksqldb-functional-tests/src/test/resources/query-validation-tests/json.json
big-andy-coates: Ah, I'll update the title.  I pushed another commit after creating the PR to enable the double testing.

big-andy-coates:
> LGTM though I think it's worth adding a few QTTs for converting between different key formats, both positive and negative such as validating what happens if I start with a JSON boolean key and try converting the key format to KAFKA.

See https://github.com/confluentinc/ksql/pull/6345

PR #6293, opened 2020-09-24
test: move QTT coercion into Kafka test serde

big-andy-coates:
### Description 

Moves the coercion required by the Kafka Serde _into_ the Kafka Serde used by QTT.

Previously, this was handled by `TopicInfoCache`. However, it's only the `KAFKA` format that needs this, (as it has specific serde for different primitive types).  Moving it into `KafkaSerdeSupplier` simplifies the code path for other formats.

### Testing done 

Usual

### Reviewer checklist


big-andy-coates:
Test only change... committing

PR #6294, opened 2020-09-24
test: drop internal topics from QTT test cases

big-andy-coates:
### Description 

The historical plans now capture all records send to internal topics, so there is no need for the QTT test cases to detail the internal messages. Removing them makes the tests easier to read.

### Testing done 

Test only change.

### Reviewer checklist


big-andy-coates:
Test only change - committing.

PR #6295, opened 2020-09-24
fix: JSON format to set correct scale of decimals

big-andy-coates:
### Description 

It is the responsibility of the format to ensure the data returned matches the required schema. This includes the scale of decimals. The JSON format was not correctly setting the scale of decimals when deserializing. For example, give:

```sql
CREATE STREAM S (ID INT KEY, PRICE DECIMAL(10,2) WITH (kafka_topic='S', formats='JSON');
```

The above creates a stream with a single value column `PRICE` which should have a scale of `2`, i.e. two decimal places.

If the data in tbe Kafka record's value was to have too small a scale, e.g.

```json
{
   "price": 12
}
```

or

```json
{
   "price": 12.1
}
```

Then the deserializer was returning the decimal as provided, i.e. `12` or `12.1`. However, this is incorrect as the schema of the column states is has a scale of two. So all values for the column should have the scale set to two, i.e. the above examples should deserialize to `12.00` and `12.10`. With this change they now do.

### Testing done 

Usual

### Reviewer checklist

### DIFF
--- a/ksqldb-serde/src/main/java/io/confluent/ksql/serde/json/KsqlJsonDeserializer.java
+++ b/ksqldb-serde/src/main/java/io/confluent/ksql/serde/json/KsqlJsonDeserializer.java
@@ -126,7 +126,7 @@ public class KsqlJsonDeserializer<T> implements Deserializer<T> {
       return SerdeUtils.castToTargetType(coerced, targetType);
     } catch (final Exception e) {
       throw new SerializationException(
-          "Failed to deserialize " + target + " from topic: " + topic, e);
+          "Failed to deserialize " + target + " from topic: " + topic + ". " + e.getMessage(), e);
     }
   }
 
@@ -173,17 +173,17 @@ public class KsqlJsonDeserializer<T> implements Deserializer<T> {
   }
 
   private static Object enforceValidBytes(final JsonValueContext context) {
-    final BigDecimal decimal;
     final boolean isDecimal = DecimalUtil.isDecimal(context.schema);
-    if (isDecimal && context.val instanceof NumericNode) {
-      decimal = context.val.decimalValue();
-      DecimalUtil.ensureFit(decimal, context.schema);
-      return decimal;
-    } else if (isDecimal && context.val instanceof TextNode) {
-      decimal = new BigDecimal(context.val.textValue());
-      DecimalUtil.ensureFit(decimal, context.schema);
-      return decimal;
+    if (isDecimal) {
+      if (context.val instanceof NumericNode) {
+        return DecimalUtil.ensureFit(context.val.decimalValue(), context.schema);
+      }
+
+      if (context.val instanceof TextNode) {
+        return DecimalUtil.ensureFit(new BigDecimal(context.val.textValue()), context.schema);
+      }
     }
+
     throw invalidConversionException(context.val, context.schema);
   }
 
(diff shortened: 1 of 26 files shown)


ksqldb-serde/src/test/java/io/confluent/ksql/serde/json/KsqlJsonDeserializerTest.java
```diff
@@ -538,6 +538,40 @@ public void shouldDeserializeDecimalsWithoutStrippingTrailingZeros() {
...
+
+    // When:
+    final Object result = deserializer.deserialize(SOME_TOPIC, bytes);
+
+    // Then:
+    assertThat(result, is(new BigDecimal("1.100")));
+  }
+
+  @Test
+  public void shouldThrowIfDecimalHasLargerScale() {
```
vcrfxia, line 557: Does ksqlDB always throw when deserializing decimals that don't fit into the specified precision/scale? What about when casting from other numeric types to decimals? There's no way for the user to specify rounding behavior? (Asking in case you know off the top of your head. If not I can poke around the code, not seeing anything in the docs one way or the other.)

ksqldb-serde/src/main/java/io/confluent/ksql/serde/json/KsqlJsonDeserializer.java
```diff
@@ -126,7 +126,7 @@ public T deserialize(final String topic, final byte[] bytes) {
       return SerdeUtils.castToTargetType(coerced, targetType);
     } catch (final Exception e) {
       throw new SerializationException(
-          "Failed to deserialize " + target + " from topic: " + topic, e);
+          "Failed to deserialize " + target + " from topic: " + topic + ". " + e.getMessage(), e);
```
vcrfxia, line 129: What's the philosophy behind when it makes sense to copy the underlying error message into the new error message?

vcrfxia (approved):
LGTM!

ksqldb-serde/src/main/java/io/confluent/ksql/serde/json/KsqlJsonDeserializer.java
big-andy-coates: Humm.. hard to give a hard and fast rule, but I generally make this change if otherwise the user would be presented with a less than useful error message.  In this case, I added it to make debugging easier... both for us and for users interacting with the processing logger.  Having the cause's message as part of the top level message means you can quickly understand what's happening without having to scroll through exception call stacks.

ksqldb-serde/src/test/java/io/confluent/ksql/serde/json/KsqlJsonDeserializerTest.java
big-andy-coates: @agavra can probably answer more definitely than me, as he did the initial work. However, I believe a deserializer _should_ throw if the deserialized value doesn't fit in the ksql decimal scale and precision.  (See `DecimalUtil.ensureFit`). This isn't necessary for some formats, e.g. `AVRO`, as the Avro schema dictates scale and precision, and the ksql schema is driven from the Avro one. 

Casting is supported, and I'm assuming `ensureFit` is being called appropriately.

I don't think there's any way for a user to specify rounding behaviour, though I'm fairly certain it was discussed during the design, but punted to a later feature.

agavra: it's been a while since i've looked at the code but I agree with what Andy is saying. As far as casting is concerned, we hardcode the scale/precision for casts:
```java
  /**
   * Schema of an INT up-cast to a DECIMAL
   */
  public static final SqlDecimal INT_UPCAST_TO_DECIMAL = SqlDecimal.of(10, 0);

  /**
   * Schema of an BIGINT up-cast to a DECIMAL
   */
  public static final SqlDecimal BIGINT_UPCAST_TO_DECIMAL = SqlDecimal.of(19, 0);
```
And then for `double` we do some other magic

PR #6302, opened 2020-09-25
chore: unwrap single keys on internal topics

big-andy-coates:
### Description 

fixes: https://github.com/confluentinc/ksql/issues/6209

Ensure internal changelog and repartition topics set `SerdeFeature.UNWRAP_SINGLES` if the key format supports both unwrapping and wrapping.  This is done to avoid ambiguity, which may cause compatibility issues once wrapped keys are supported.

### Testing done 

Usual

### Reviewer checklist

### DIFF
--- a/ksqldb-engine/src/test/java/io/confluent/ksql/structured/SchemaKGroupedStreamTest.java
+++ b/ksqldb-engine/src/test/java/io/confluent/ksql/structured/SchemaKGroupedStreamTest.java
@@ -42,6 +42,7 @@ import io.confluent.ksql.schema.ksql.types.SqlTypes;
 import io.confluent.ksql.serde.FormatFactory;
 import io.confluent.ksql.serde.FormatInfo;
 import io.confluent.ksql.serde.KeyFormat;
+import io.confluent.ksql.serde.SerdeOption;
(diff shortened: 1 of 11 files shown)


ksqldb-engine/src/test/java/io/confluent/ksql/structured/SchemaKGroupedStreamTest.java
```diff
@@ -107,6 +108,7 @@
 
   @Before
   public void setUp() {
+    when(keyFormat.getFormat()).thenReturn(FormatFactory.KAFKA);
```
vcrfxia, line 111: Any reason we don't remove `getFormat` and inline using `getFormatInfo` here, similar to the changes we recently made to the source properties classes?

vcrfxia (approved):
LGTM!

ksqldb-engine/src/test/java/io/confluent/ksql/structured/SchemaKGroupedStreamTest.java
big-andy-coates: That's coming in the next PR :)

PR #6307, opened 2020-09-28
Refactor: remove `SerdeOption`

big-andy-coates:
### Description 

Stacked on top of https://github.com/confluentinc/ksql/pull/6302.

First off....

# DON'T PANIC!

This PR is not a massive as it looks.. honest!

`SerdeOption` and `SerdeFeature` serve pretty much the same function. Formats define what features they support using the `SerdeFeature` enum, i.e. it operates at the _format_ level. Conversely, `SerdeOption` operates at the _topic_ level. It holds the serde features required for both the key and value formats. It's just code bloat, in there for historical reasons.

Removing `SerdeOptions` now will clean up the code and remove confusion. Doing it later will be much much harder... i.e it'll never happen. So... biting that bullet hard...

This commit removes `SerdeOption` from the code base, replacing it use with direct use of `SerdeFeature`.

Changes:

1. Removed `SerdeOption` enum and `SerdeOptions` class.
1. Renamed `EnabledSerdeFeatures` -> `SerdeFeatures`.  This class is just a validated set of features. You can't build an instance with conflicting features, e.g. both `WRAP_SINGLES` and `UNWRAPE_SINGLES`.
1. Split the `SerdeOptions` attribute of `DataSource` into `SerdeFeatures` attributes of the `KeyFormat` and `ValueFormat` classes, which are attributes of the `KsqlTopic` class, which itself is an attribute of the `DataSource`.
  ```
  // Previously:
  dataSource->serdeOptions   <-- stored SerdeOptions for both key and value

  // Now:
  dataSource->ksqlTopic->keyFormat->features  <-- stores SerdeFeatures for the key
                       ->valueFormat->features  <-- stores SerdeFeatures for the value
  ```
1. C* statements simply construct the `KsqlTopic` with the serde features set on the key and value format.
1. C*AS are slightly more complex. Previously the `Analyser` was constructing the `KsqlTopic`, and the Logical Plan was constructing the `SerdeOption`.  Now the `Analsyser` can't construct the `KsqlTopic` as the `SerdeFeatures` are unknown at this point.  Hence it creates a `NewTopic`, which the Logical Plan uses to build the `KsqlTopic` once its constructed the `SerdeFeatures`.
1. `SchemaKStream.into` and `SchemaKTable.into` now just take the `KsqlTopic`, rather than lots of individual params.
1. `SerdeOptionsFactory`, the class responsible for building the `SerdeOptions` has been renamed to `SerdeFeaturesFactory` and now has separate methods for building key and value features.
1. `PhysicalSchema` no longer stores `SerdeOptions`, as the key and value `PersistenceSchema`s already track the `SerdeFeatures`.
1. Each `SerdeFeature` now defines any other features it is not compatible with - making this more data driven.
1. *Command Topic change*:
   1. The `Formats` type is used to serialize format information to the command topic. Previously, it stored the key and value `FormatInfo` and the `SerdeOptions`.  Now it stores key and value `FormatInfo` and key and value `SerdeFeatures`.  To enable backwards compatibility with older plans the factory method used by Jackson when deserializing the plans accepts both old and new attributes and converts the old into the new format.
   1. New plans will be written with `valueFeatures` and `keyFeatures`, but old plans can continue to have `options`.  This is necessary to maintain backwards compatibility.
1. *QTT changes*
   1. Not historic _query plans_ need to change because of this change!
   1. Historic query _specs_ have been changed: places that stored the serde options have been changed to store the serde features.  This is fine, as its just the test data that's changing.
   1. `ValueFormat` was being serialized as a value type, but now has a `SerdeFeatures` attribute, so the type now exposes the format name and properties of its inner `FormatInfo` manually. (So the JSON is clean).
   1. `KeyFormat` was being serialized with an inner `formatInfo` field. This was out of whack with `ValueFormat`, so this has been changed to follow the same pattern as `ValueFormat`.
   1. A few QTT changes have been updated to use the new layout.

### Reviewing notes:

A LOT of files needed to change for this work, but not much functionality. Focus your review on the functionality!

To make things easier, I've broken the PR down into the following commits:

1. (ignore): https://github.com/confluentinc/ksql/commit/c95f9c0658badb42ef144d5d281a0d4dbaa7734d: This is the commit for https://github.com/confluentinc/ksql/pull/6302, upon which this PR is stacked.
1. [Production code: Main functionality changes](https://github.com/confluentinc/ksql/commit/2b58bed63f1200aa2cfe3325c59f6411d437d003)
1. [Test code: Main functionality changes](https://github.com/confluentinc/ksql/commit/0e30cdbb399a56da0723ef5dedcdfc261c44e2f3)
1. [Production code: Minor stuff](https://github.com/confluentinc/ksql/commit/acabcfc0ba7c0b9973fc90638970b0e395793c58), e.g. changes due to renames, switch `SerdeOption` to `SerdeFeature`, or other slight API changes - not interesting.
1. [Test code: Minor stuff](https://github.com/confluentinc/ksql/commit/59d619fa938f73d211934f497b18a62adf364448)
1. [QTT test changes](https://github.com/confluentinc/ksql/commit/22848de7e0b7e54e162ee65d969580298e87da1b)
1. [Historical Plans changes](https://github.com/confluentinc/ksql/commit/0c2062a499e9c6e562ecc6caa8461a89b7c917c7) - definitely don't review all of this!  Take a look at how one or two files have changed to get an idea of the change if you want.

### Testing done 

usual

### Reviewer checklist


ksqldb-common/src/main/java/io/confluent/ksql/serde/SerdeFeatures.java
```diff
@@ -0,0 +1,126 @@
...
+import java.util.Set;
+
+
+/**
+ * Validated set of enabled features
+ *
+ * <p>Known to not have conflicting features enabled
+ */
+@Immutable
+public final class SerdeFeatures {
```
big-andy-coates, line 36: Actually a rename from `EnabledSerdeFeatures`, but git's lost track.

ksqldb-engine/src/main/java/io/confluent/ksql/serde/SerdeFeaturesFactory.java
```diff
@@ -0,0 +1,165 @@
...
+package io.confluent.ksql.serde;
+
+import com.google.common.collect.ImmutableSet;
+import io.confluent.ksql.properties.with.CommonCreateConfigs;
+import io.confluent.ksql.schema.ksql.LogicalSchema;
+import io.confluent.ksql.util.KsqlConfig;
+import io.confluent.ksql.util.KsqlException;
+import java.util.Optional;
+
+public final class SerdeFeaturesFactory {
```
big-andy-coates, line 25: Actually a rename from `SerdeOptionsFactory`...

vcrfxia (approved):
LGTM! Left comments separately since the complete diff was too large to load which means I couldn't add comments into a review...

vcrfxia:
Not sure why Github isn't showing merge conflicts with https://github.com/confluentinc/ksql/pull/6273. Would be good to sanity check that we aren't losing functionality from that PR in this one.

big-andy-coates:
> Not sure why Github isn't showing merge conflicts with #6273. Would be good to sanity check that we aren't losing functionality from that PR in this one.

I've already merged that in - think we're good...

PR #6315, opened 2020-09-29
docs: add javadoc artifacts for API client

vcrfxia:
### Description 

Adds API docs to the release branch so that they can be served on https://docs.ksqldb.io.

> This PR should NOT be cherry-picked to other branches. We'd like to keep the generated API docs only on docs branches, as the source of truth for these are the Javadocs on the interfaces.

### Testing done 

Docs-only change.

### Reviewer checklist


JimGalasyn (approved):
LGTM!

vcrfxia:
Merging without build as builds against docs branches are expected to fail + this is a docs-only change.

PR #6317, opened 2020-09-29
refactor: move transactional command topic inside CommandStore

spena:
### Description 
Note: This refactor helps continue working on the KSQL restore metadata command.

The refactor moves the transactional operations when writing commands to the command topic inside the `CommandStore`. This guarantees the `enqueueCommand` is always done in a transactional operation, and not depending on the caller to create the transactional context.

The main class to review is the `TransactionManager`, which is in charge of executing code that uses the `KafkaProducer` using a single transaction. i.e.
```
TransactionManager transactionManager = ...;
ProducerRecord record = ...;

transactionManager.executeTransaction(producer -> {
   return producer.send(record).get();
});
```
`TransactionManager` initializes and begins each transaction before calling the transaction block. It commits the transaction afterewards, or aborts it in case the transaction block throws an exception.

I made `TransactionManager` generic so it can be used in future code (i.e. passing multiple values in insert/value statement).

The rest of the code does:
- `CommandStore` calls the `TransactionManager` to execute the command topic producer in a single transaction.
- `DistributingExecutor` just calls `enqueueCommand` expecting the transaction happens internally.
- Tests

### Testing done 
Added unit tests
Verified manually in the CLI by executing some CREATE/DROP commands.

### Reviewer checklist

### DIFF
--- a/ksqldb-benchmark/src/main/java/io/confluent/ksql/benchmark/SerdeBenchmark.java
+++ b/ksqldb-benchmark/src/main/java/io/confluent/ksql/benchmark/SerdeBenchmark.java
@@ -16,6 +16,7 @@
 package io.confluent.ksql.benchmark;
 
 import com.google.common.collect.ImmutableMap;
+import edu.umd.cs.findbugs.annotations.SuppressFBWarnings;
 import io.confluent.avro.random.generator.Generator;
 import io.confluent.kafka.schemaregistry.client.MockSchemaRegistryClient;
 import io.confluent.kafka.schemaregistry.client.SchemaRegistryClient;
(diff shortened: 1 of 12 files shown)


ksqldb-benchmark/src/main/java/io/confluent/ksql/benchmark/SerdeBenchmark.java
```diff
@@ -183,6 +184,7 @@ private static RowGenerator getRowGenerator(final String schemaName) throws IOEx
       return new RowGenerator(generator, keyField, Optional.empty());
     }
 
+    @SuppressFBWarnings("RCN_REDUNDANT_NULLCHECK_OF_NONNULL_VALUE")
```
spena, line 187: My local build fails without this. I don't know why ksql Jenkins passes, but even cleaning my MVN repo was failing.

spena:
Closing this PR without merge. Will try using a simple KafkaProducer in the restore command without transactions as it's not necessary.

PR #6321, opened 2020-09-30
Historic serde

big-andy-coates:
### Description 

[As noted in a previous PR](2b58bed#r42793874), not have the `keyFeatures` and `valueFeatures` in the post condition check for sources in the metastore means QTT won't check the key and value features.

While this isn't an issue at the moment, (as the `topics` nodes _will_ check these), it's likely safer to add in these checks, so that future changes don't inadvertently remove the checks from the `topics`, leaving nothing checking the serde features.

This commit also fixes a bug where the `valueFormat` was being incorrectly set to the key format, and adds in that `valueFormat` to the `spec.json` files.

Note: using the `PlannedTestRewriter` to regenerate the historic plans can result in the order of nodes within the `post` node's `sources` to be re-ordered, making the diff larger than it need to be.  To stop this happening next time the nodes are now sorted alphabetically.

### Testing done 

usual

### Reviewer checklist


vcrfxia (approved):
LGTM -- thanks for the fix (and additional test coverage)!

PR #6323, opened 2020-09-30
docs: fix javadocs artifact version

vcrfxia:
### Description 

Follow-up to https://github.com/confluentinc/ksql/pull/6315. The 0.12.0-ksqldb docs branch was cut from the release branch rather than the release tag, which means the current version on the API docs shows as `6.1.0-SNAPSHOT` rather than `0.12.0`. This PR fixes the issue (for the API docs, not where the docs branch is cut from -- there should be no other implications of the discrepancy).

### Testing done 

Docs-only change.

### Reviewer checklist


JimGalasyn (approved):
Good catch!

vcrfxia:
Merging despite the build failure as builds against this branch aren't expected to pass, and this is a docs-only change.

PR #6324, opened 2020-09-30
chore: add ksqldb staging repo to release maven settings (MINOR)

vcrfxia:
### Description 

This is needed in order to generate HTML API docs from javadocs on Java client interfaces from release tags, rather than release branches. The former is preferred so that the version displayed on the HTML docs will be the release version (e.g., 0.12.0) rather than the master version (e.g., 6.1.0-SNAPSHOT). See https://github.com/confluentinc/ksql/pull/6323 for an example of the difference.

### Testing done 

Non-functional change.

### Reviewer checklist

### DIFF
--- a/maven-settings-template.xml
+++ b/maven-settings-template.xml
@@ -16,12 +16,19 @@
           <name>Confluent</name>
           <url>PACKAGES_MAVEN_URL</url>
         </repository>
+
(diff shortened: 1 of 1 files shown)


agavra (approved):
lgtm :) if it builds I'm happy with it

Issue #6218, opened 2020-09-16
[JSON Support]: Test CLI with non-KAFKA keys


big-andy-coates:
Test DESCRIBE and anywhere else that the key format is shown to ensure it all works.

Issue #6269, opened 2020-09-22
Update ksqlDB CLI to use the Java client


vcrfxia:
We should update the ksqlDB CLI to use the Java client rather than having two codepaths to maintain, and since it'll be a great opportunity to dogfood the Java client. This would also allow us to deprecate the `/query` endpoint used by the CLI for push queries and remove another codepath there as well.

A prerequisite for doing this would be to enhance the Java client to support all the different types of statements that the REST API and CLI support, including connector management, the extended versions of admin commands, and other admin commands such as explaining queries. 

We'd also have to decide how to handle the fact that the new `/inserts-stream` endpoint used by the Java client does not support expressions in `INSERT VALUES` statements as the `/ksql` endpoint currently used by the CLI for `INSERT VALUES` statements does. If we don't want to break backwards compatibility, we'd have to either enhance the `/inserts-stream` endpoint, introduce separate Java client methods for the different endpoints, or introduce logic into the CLI / Java client to determine where the requests should be sent (not recommended).

apurvam:
@vcrfxia to do a rough scoping with @colinhicks . We can prioritize after that.

vcrfxia:
Recording the remaining open questions after syncing with @colinhicks below.

**Should we migrate `INSERT VALUES` to use the Java client?**

The new HTTP/2 endpoint for inserting values differs from the old endpoint in two main ways:
- the new endpoint only allows insertion into streams, not tables.
- the new endpoint only supports inserting literals, not expressions.
The first point is easy to change if we want to, but it's unclear that we do since inserting into a table today only inserts into the underlying Kafka topic and not the corresponding state store (if materialized) which is confusing.

The second point is more work to reconcile. Options include:
- enhancing the HTTP/2 endpoint to support expressions. Open question about what the interface on the Java client should look like if we do this
- drop support for inserting expressions from the CLI
- do not migrate the CLI to the new HTTP/2 endpoint

**Should we maintain backwards compatibility between the new CLI (post-migration) and older server versions?**

The HTTP/2 endpoints used by the Java client were only introduced in recent ksqlDB releases, and we'll likely need to make additional server-side changes to enable the CLI migration as well. Maintaining compatibility between the new CLI post-migration and older server versions would be significantly more work and likely not worth it.

**How should we expose command sequence number or equivalent functionality in the Java client?**

[...]

Issue #6042, opened 2020-08-18
Refactor Java client to use Jackson for parsing server responses


vcrfxia:
As discussed in https://github.com/confluentinc/ksql/pull/5944#issuecomment-675022279, we should:
- refactor the Java client interfaces into the ksqldb-rest-model module
- have the server response classes implement the interfaces
- switch to using Jackson to parse the server responses
- shade the Jackson dependency in the client pom to avoid conflicts with other user app dependencies.

PR #6327, opened 2020-09-30
chore: ensure CLI works with other key formats

big-andy-coates:
### Description 

fixes: https://github.com/confluentinc/ksql/issues/6218

Updates a few places in the rest api, cli and the new Java client to specify the key format as well as the value format.

Unfortunately, the value format was called `format`.  This change introduces a `keyFormat` and `valueFormat` to hold the two formats, but both the CLI and new Java client are backwards compatible with the old JSON schema.

Also in this commit, streams also indicate if there key is windowed.  This means the key is serialized using the windowed serializer. It is possible to have a stream over a topic with a windowed key. Now the CLI and Java client expose this info.

Main change was to `show tables` and `show streams` response.  Previously, these returned

```
ksql> show streams;

 Stream Name | Kafka Topic | Format
---------------------------------------
 PAGEVIEWS   | pageviews   | DELIMITED
---------------------------------------

ksql> show tables;

 Table Name | Kafka Topic | Format | Windowed
----------------------------------------------
 USERS      | users       | JSON   | false
----------------------------------------------
```

And now return:

```
ksql> show streams;

 Stream Name | Kafka Topic | Key Format | Value Format | Windowed
-----------------------------------------------------------------
 PAGEVIEWS   | pageviews   | KAFKA      | DELIMITED    | false
-----------------------------------------------------------------

ksql> show tables;

 Table Name | Kafka Topic | Key Format | Value Format | Windowed
----------------------------------------------------------------
 USERS      | users       | KAFKA      | JSON          | false
----------------------------------------------------------------
```

### Testing done 

Usual + manually created streams and tables with non-KAFKA keys and ensured it all worked.

### Reviewer checklist

### DIFF
--- a/ksqldb-api-client/src/main/java/io/confluent/ksql/api/client/impl/StreamInfoImpl.java
+++ b/ksqldb-api-client/src/main/java/io/confluent/ksql/api/client/impl/StreamInfoImpl.java
@@ -22,12 +22,22 @@ public class StreamInfoImpl implements StreamInfo {
 
   private final String name;
   private final String topicName;
-  private final String format;
+  private final String keyFormat;
+  private final String valueFormat;
+  private final boolean windowed;
 
-  StreamInfoImpl(final String name, final String topicName, final String format) {
+  StreamInfoImpl(
+      final String name,
+      final String topicName,
+      final String keyFormat,
+      final String valueFormat,
+      final boolean windowed
+  ) {
     this.name = Objects.requireNonNull(name);
     this.topicName = Objects.requireNonNull(topicName);
-    this.format = Objects.requireNonNull(format);
+    this.keyFormat = Objects.requireNonNull(keyFormat);
+    this.valueFormat = Objects.requireNonNull(valueFormat);
+    this.windowed = windowed;
   }
 
   @Override
@@ -41,8 +51,18 @@ public class StreamInfoImpl implements StreamInfo {
   }
 
   @Override
-  public String getFormat() {
-    return format;
+  public String getKeyFormat() {
+    return keyFormat;
+  }
+
+  @Override
+  public String getValueFormat() {
+    return valueFormat;
+  }
+
+  @Override
+  public boolean isWindowed() {
+    return windowed;
   }
 
   @Override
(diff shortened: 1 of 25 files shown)


big-andy-coates:
@vcrfxia, following on from our previous conversation about using the existing rest model pojos and Jackson, rather than new classes and hand crafted JSON deserialization code... 

This PR is a good example of why we should share the model between the server and the java client. Because we don't I needed to update two parallel sets of code that handle the JSON<>Java conversion for the `SourceInfo`, and both bits of code needed to be enhanced and tested to ensure they were backwards compatible.

Over time, this shadow model and deserialization code will be expensive to maintain and invariably get out of sync.

JimGalasyn (approved):
Excellent!

ksqldb-api-client/src/main/java/io/confluent/ksql/api/client/impl/StreamInfoImpl.java
```diff
@@ -41,8 +51,18 @@ public String getTopic() {
   }
 
   @Override
-  public String getFormat() {
```
vcrfxia, line 44: This is a public API. Should we leave it here and mark as deprecated, to be removed in a future release?

ksqldb-api-client/src/main/java/io/confluent/ksql/api/client/impl/TableInfoImpl.java
```diff
@@ -48,13 +51,18 @@ public String getTopic() {
   }
 
   @Override
-  public String getFormat() {
```
vcrfxia, line 51: As above.

vcrfxia (approved):
LGTM -- thanks! One question inline.

> @vcrfxia, following on from our previous conversation about using the existing rest model pojos and Jackson, rather than new classes and hand crafted JSON deserialization code...
>
> This PR is a good example of why we should share the model between the server and the java client.

Agreed. There's already an issue for this (https://github.com/confluentinc/ksql/issues/6042) and it's marked as a blocker before any additional features are added (or other major changes are made) to the client, such as https://github.com/confluentinc/ksql/issues/6269.

ksqldb-api-client/src/main/java/io/confluent/ksql/api/client/impl/StreamInfoImpl.java
big-andy-coates: My personal view is no, I think the Java client is new enough we can get away with just changing this. People will get a compilation error and will just switch to the new method. But... we don't actually have a defined policy for this yet.

big-andy-coates: Meh... done.

Issue #6213, opened 2020-09-16
[JSON Support]: Have joins fail if key formats differ


big-andy-coates:
1. Add code to ensure joins fail if the key formats do not match... including key serde features.
1. Enhance QTT test to be able to handle different key formats.
1. Add negative tests to `join.json` and `multi-join.json` to ensure joins will fail with sensible error message for now.

big-andy-coates:
Re-opened as we can do better on how we handle joins with mismatched key formats.  Currently code fails even if one or both sides will be repartitioned.

PR #6336, opened 2020-09-30
chore: fail joins where the key format differs

big-andy-coates:
### Description 

fixes: https://github.com/confluentinc/ksql/issues/6213

Add code to throw a useful error message if a join has incompatible key formats. QTT test added too.

### Testing done 

Usual

### Reviewer checklist

### DIFF
--- a/ksqldb-engine/src/main/java/io/confluent/ksql/planner/plan/JoinNode.java
+++ b/ksqldb-engine/src/main/java/io/confluent/ksql/planner/plan/JoinNode.java
@@ -73,7 +73,7 @@ public class JoinNode extends PlanNode {
   }
 
   private static final ImmutableList<KeyProperty> KEY_PROPERTIES = ImmutableList.of(
-      new KeyProperty("format", kf -> kf.getFormatInfo().getFormat()),
+      new KeyProperty("formats", kf -> kf.getFormatInfo().getFormat()),
       new KeyProperty("properties", kf -> kf.getFormatInfo().getProperties()),
       new KeyProperty("features", KeyFormat::getFeatures),
       new KeyProperty("widowing", kf -> kf.getWindowInfo()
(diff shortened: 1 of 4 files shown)


ksqldb-engine/src/main/java/io/confluent/ksql/planner/plan/JoinNode.java
```diff
@@ -67,6 +72,20 @@
...
+      new KeyProperty("properties", kf -> kf.getFormatInfo().getProperties()),
+      new KeyProperty("features", KeyFormat::getFeatures),
+      new KeyProperty("widowing", kf -> kf.getWindowInfo()
+          .map(WindowInfo::getType)
+          .map(t -> t == WindowType.SESSION
+              ? "session windowing" : "non session windowing")
+          .orElse("no windowing"))
+  // Note: we deliberately do NOT fail if one side is HOPPING and the other is TUMBLING, or fail
+  // on different window sizes. This is because we only fail if the _schema_ of the key is
+  // different. Neither of these differences make the schemas different.
```
vcrfxia, line 86: Why are the schemas for session and non-session windows different? Is this because non-session windows don't store window end times?

ksqldb-engine/src/main/java/io/confluent/ksql/planner/plan/JoinNode.java
```diff
@@ -97,6 +116,8 @@ public JoinNode(
     this.left = requireNonNull(left, "left");
     this.right = requireNonNull(right, "right");
     this.withinExpression = requireNonNull(withinExpression, "withinExpression");
+
+    throwOnKeyMismatch(left, right);
```
vcrfxia, line 120: How come this validation happens here rather than in buildStream() where other validation (matching partition counts) occurs? Is it because we actually prefer to validate in the constructor when possible, but the partition count validation requires a topic client which isn't available until buildStream() is called?

vcrfxia (approved):
LGTM!

ksqldb-engine/src/main/java/io/confluent/ksql/planner/plan/JoinNode.java
big-andy-coates: Correct

ksqldb-engine/src/main/java/io/confluent/ksql/planner/plan/JoinNode.java
big-andy-coates: Where validation happens currently is... a little undefined.  A lot happens in the Analyzer, but personally I'm like to kill off the Analyzer and use the building of the LogicalPlan to validate thing.  I don't think the Analyzer adds anything.

Failing in the constructor is better than `buildStream` as the former means you can't build an invalid logical plan, where as the latter only means you can't build an invalid physical plan.

Issue #6308, opened 2020-09-28
Add new detail to healthcheck response to expose CommandRunnner health


stevenpyzhang:
**Is your feature request related to a problem? Please describe.**
As a follow up to https://github.com/confluentinc/ksql/pull/6169 and https://github.com/confluentinc/ksql/pull/6240 , we should still expose some information in the healthcheck that indicates that the CommandRunner is in a bad state so the user has all the statuses they need in one place to determine the health of a server.

**Describe the solution you'd like**
Add a `commandRunnerHealthy` to the details list in the healthcheck response

**Additional context**
Any healthchecks services that want to use the healthcheck endpoint to alert on an unhealthy server should only use the `metastore` and `kafka` detail in the response.

PR #6346, opened 2020-10-01
feat: add commandRunnerCheck to healthcheck detail

stevenpyzhang:
### Description 
https://github.com/confluentinc/ksql/issues/6308
Exposes command runner health in the healthcheck endpoint

### Testing done 
Issued a request to healthcheck endpoint
Unit test

### Reviewer checklist

### DIFF
--- a/ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/healthcheck/HealthCheckAgent.java
+++ b/ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/healthcheck/HealthCheckAgent.java
@@ -25,6 +25,7 @@ import io.confluent.ksql.rest.entity.HealthCheckResponseDetail;
 import io.confluent.ksql.rest.entity.KsqlEntityList;
 import io.confluent.ksql.rest.server.KsqlRestConfig;
 import io.confluent.ksql.rest.server.ServerUtil;
+import io.confluent.ksql.rest.server.computation.CommandRunner;
 import io.confluent.ksql.services.ServiceContext;
 import io.confluent.ksql.services.SimpleKsqlClient;
 import io.confluent.ksql.util.KsqlConfig;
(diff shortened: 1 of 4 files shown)


ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/healthcheck/HealthCheckAgent.java
```diff
@@ -156,4 +162,24 @@ public HealthCheckResponseDetail check(final HealthCheckAgent healthCheckAgent)
...
+    @Override
+    public String getName() {
+      return name;
+    }
+
+    @Override
+    public HealthCheckResponseDetail check(final HealthCheckAgent healthCheckAgent) {
+      return new HealthCheckResponseDetail(
+          healthCheckAgent.commandRunner.checkCommandRunnerStatus()
+              == CommandRunner.CommandRunnerStatus.RUNNING);
```
vcrfxia, line 182: Do we understand what causes the command runner to transition into ERROR state in practice? From looking at the code, it looks like this happens if the command runner has been unable to poll for 15 seconds (by default). Do we have a sense of how often this happens in practice, and why? It'd be good to have some assurance that this is rare, and that when it happens it singles an actual issue rather than something spurious, else failing the healthcheck because of it would not provide useful signal to users.

BTW, how was the 15 second default time limit chosen?

vcrfxia (approved):
Thanks @stevenpyzhang -- LGTM with a question inline!

ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/healthcheck/HealthCheckAgent.java
stevenpyzhang: The 15 seconds is kind of arbitrary. A normal poll of the command topic by default is 5 seconds so just 3x that was the rationale.

So there's two conditions that can lead to returning an ERROR, if there's currently a command being processed or not. 

If there's no command being processed, we check to see if the command runner has been unable to poll for 15 seconds. If it hasn't, it means that the CommandRunner thread is most likely dead so the server would still be running without a CommandRunner thread, which is an unhealthy state (this is if the UncaughtExceptionHandler isn't enabled).

If there's a command being processed, and it's been processing for longer than `ksql.server.command.blocked.threshold.error.ms` that's also an unhealthy state for the server since it's probably stuck executing a command (this would most likely be a bug in the code and this wouldn't eventually finish executing on its own) and isn't making progress on the command topic

PR #6348, opened 2020-10-01
fix: backup files are re-created on every restart

spena:
### Description 
This PR fixes an issue with backups re-created over and over again on every server restart. 

The cause was that sometimes the command topic comes with command properties (originalProperties) with nulls and empty values. When the backup serializes those properties to write them as JSON in the file, the nulls and empty are missing. Later, when restarting the server, the command properties read from command topic are not equals from the ones found in the backup file; so a new backup file is created.

The fix makes sure to include nulls and empty properties in the backup file. It also adds a header to include the version of the backup file. When the backup file do not contain a version header (during an upgrade to 0.13), then the comparison will only use the commandId and the command statement (without the properties), so it the server start does not fail with degradation state.

### Testing done 
Modified test cases
Verified manually

### Reviewer checklist

### DIFF
--- a/ksqldb-benchmark/src/main/java/io/confluent/ksql/benchmark/SerdeBenchmark.java
+++ b/ksqldb-benchmark/src/main/java/io/confluent/ksql/benchmark/SerdeBenchmark.java
@@ -16,6 +16,7 @@
 package io.confluent.ksql.benchmark;
 
 import com.google.common.collect.ImmutableMap;
+import edu.umd.cs.findbugs.annotations.SuppressFBWarnings;
 import io.confluent.avro.random.generator.Generator;
 import io.confluent.kafka.schemaregistry.client.MockSchemaRegistryClient;
 import io.confluent.kafka.schemaregistry.client.SchemaRegistryClient;
@@ -183,6 +184,7 @@ public class SerdeBenchmark {
       return new RowGenerator(generator, keyField, Optional.empty());
     }
 
+    @SuppressFBWarnings("RCN_REDUNDANT_NULLCHECK_OF_NONNULL_VALUE")
     private static Generator getGenerator(final String schemaName) throws IOException {
       final Path schemaPath = SCHEMA_DIR.resolve(schemaName + SCHEMA_FILE_SUFFIX);
 
--- a/ksqldb-engine/src/main/java/io/confluent/ksql/execution/json/PlanJsonMapper.java
+++ b/ksqldb-engine/src/main/java/io/confluent/ksql/execution/json/PlanJsonMapper.java
@@ -32,23 +32,19 @@ public enum PlanJsonMapper {
 
   INSTANCE;
 
-  private final ObjectMapper mapper = newInstance();
-
-  public static ObjectMapper newInstance() {
-    return new ObjectMapper()
-        .registerModules(
-            new Jdk8Module(),
-            new JavaTimeModule(),
-            new KsqlParserSerializationModule(),
-            new KsqlTypesSerializationModule(),
-            new KsqlTypesDeserializationModule()
-        )
-        .enable(DeserializationFeature.FAIL_ON_UNKNOWN_PROPERTIES)
-        .enable(DeserializationFeature.FAIL_ON_NULL_FOR_PRIMITIVES)
-        .enable(DeserializationFeature.FAIL_ON_NULL_CREATOR_PROPERTIES)
-        .enable(DeserializationFeature.FAIL_ON_INVALID_SUBTYPE)
-        .setSerializationInclusion(Include.NON_EMPTY);
-  }
+  private final ObjectMapper mapper = new ObjectMapper()
+      .registerModules(
+          new Jdk8Module(),
+          new JavaTimeModule(),
+          new KsqlParserSerializationModule(),
+          new KsqlTypesSerializationModule(),
+          new KsqlTypesDeserializationModule()
+      )
+      .enable(DeserializationFeature.FAIL_ON_UNKNOWN_PROPERTIES)
+      .enable(DeserializationFeature.FAIL_ON_NULL_FOR_PRIMITIVES)
+      .enable(DeserializationFeature.FAIL_ON_NULL_CREATOR_PROPERTIES)
+      .enable(DeserializationFeature.FAIL_ON_INVALID_SUBTYPE)
+      .setSerializationInclusion(Include.NON_EMPTY);
 
   public ObjectMapper get() {
     return mapper;
--- a/ksqldb-engine/src/main/java/io/confluent/ksql/execution/json/PlanJsonMapper.java
+++ b/ksqldb-engine/src/main/java/io/confluent/ksql/execution/json/PlanJsonMapper.java
@@ -32,19 +32,23 @@ public enum PlanJsonMapper {
 
   INSTANCE;
 
-  private final ObjectMapper mapper = new ObjectMapper()
-      .registerModules(
-          new Jdk8Module(),
-          new JavaTimeModule(),
-          new KsqlParserSerializationModule(),
-          new KsqlTypesSerializationModule(),
-          new KsqlTypesDeserializationModule()
-      )
-      .enable(DeserializationFeature.FAIL_ON_UNKNOWN_PROPERTIES)
-      .enable(DeserializationFeature.FAIL_ON_NULL_FOR_PRIMITIVES)
-      .enable(DeserializationFeature.FAIL_ON_NULL_CREATOR_PROPERTIES)
-      .enable(DeserializationFeature.FAIL_ON_INVALID_SUBTYPE)
-      .setSerializationInclusion(Include.NON_EMPTY);
+  private final ObjectMapper mapper = newInstance();
+
+  public static ObjectMapper newInstance() {
+    return new ObjectMapper()
+        .registerModules(
+            new Jdk8Module(),
+            new JavaTimeModule(),
+            new KsqlParserSerializationModule(),
+            new KsqlTypesSerializationModule(),
+            new KsqlTypesDeserializationModule()
+        )
+        .enable(DeserializationFeature.FAIL_ON_UNKNOWN_PROPERTIES)
+        .enable(DeserializationFeature.FAIL_ON_NULL_FOR_PRIMITIVES)
+        .enable(DeserializationFeature.FAIL_ON_NULL_CREATOR_PROPERTIES)
+        .enable(DeserializationFeature.FAIL_ON_INVALID_SUBTYPE)
+        .setSerializationInclusion(Include.NON_EMPTY);
+  }
 
   public ObjectMapper get() {
     return mapper;
--- a/ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/BackupReplayFile.java
+++ b/ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/BackupReplayFile.java
@@ -15,152 +15,70 @@
 
 package io.confluent.ksql.rest.server;
 
-import com.fasterxml.jackson.annotation.JsonInclude;
-import com.fasterxml.jackson.databind.ObjectMapper;
-import io.confluent.ksql.execution.json.PlanJsonMapper;
-import io.confluent.ksql.rest.entity.CommandId;
-import io.confluent.ksql.rest.server.computation.Command;
 import io.confluent.ksql.util.KsqlException;
 import io.confluent.ksql.util.Pair;
-import java.io.BufferedWriter;
 import java.io.Closeable;
 import java.io.File;
 import java.io.FileNotFoundException;
 import java.io.FileOutputStream;
 import java.io.IOException;
-import java.io.OutputStreamWriter;
 import java.nio.charset.StandardCharsets;
 import java.nio.file.Files;
-import java.nio.file.Path;
-import java.nio.file.attribute.PosixFilePermissions;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.Objects;
-import java.util.Optional;
+import org.apache.kafka.clients.consumer.ConsumerRecord;
 
 /**
  * A file that is used by the backup service to replay command_topic commands.
  */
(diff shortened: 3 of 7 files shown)


ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/BackupReplayFile.java
```diff
@@ -30,21 +31,84 @@
...
+      .setSerializationInclusion(JsonInclude.Include.ALWAYS);
+
   private static final String KEY_VALUE_SEPARATOR = ":";
 
   private final File file;
   private final BufferedWriter writer;
+  private final Versions version;
+
+  public static BackupReplayFile newFile(final Path path) {
+    final String versionHeader =
```
agavra, line 66: let's use JSON for the header so that we can add more things to it in the future if we want to (and we can even make it a POJO to make it easy to serde)

ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/BackupReplayFile.java
```diff
@@ -30,21 +31,84 @@
...
+
+  // Current version of the Backup file
+  private static final String BACKUP_VERSION_HEADER = "BACKUP_VERSION";
+  private static final Versions DEFAULT_BACKUP_VERSION = Versions.V1;
+
+  // Include.ALWAYS is necessary to serialize nulls and empty values. This is required to allow
+  // the CommandTopicBackupImpl to compare against all properties from a record read from
+  // the command topic, which come with nulls and empty values.
+  private static final ObjectMapper MAPPER = PlanJsonMapper.INSTANCE.get()
+      .setSerializationInclusion(JsonInclude.Include.ALWAYS);
```
agavra, line 57: I think this will set it on the underlying instance. Instead we should create a new instance and set the serialization on the new instance (we'll need to change `PlanJsonMapper` to expose a newInsatcne() call)

ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/BackupReplayFile.java
```diff
@@ -61,6 +125,10 @@ private static BufferedWriter createWriter(final File file) {
     }
   }
 
+  public Versions getVersion() {
```
agavra, line 128: nit: javadoc what this version means and what it's used for

ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/CommandTopicBackupImpl.java
```diff
@@ -230,7 +245,7 @@ private BackupReplayFile newReplayFile() {
     }
 
     return (latestBakFile != null)
-        ? Optional.of(new BackupReplayFile(latestBakFile))
+        ? Optional.of(BackupReplayFile.openFile(latestBakFile))
```
agavra, line 248: I think things get a little complicated here - if the replay file is `NO_VERSION`, then we need to create a new one with the newer version but restore using the existing one. I think it might just be simpler if we don't compare across versions - we didn't have this functionality until 0.13, it's probably OK to go one more deploy without it.

thoughts @stevenpyzhang @rodesai?

agavra (commented):
Overall I think this makes sense, I have a few comments inline that I think we need to address (and as you mentioned, more testing)

ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/CommandTopicBackupImpl.java
rodesai: why do we have multiple backup files?

ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/CommandTopicBackupImpl.java
```diff
@@ -120,8 +120,23 @@ private boolean isRestoring() {
 
   private boolean isRecordInLatestReplay(final ConsumerRecord<CommandId, Command> record) {
     final Pair<CommandId, Command> latestReplayRecord = latestReplay.get(latestReplayIdx);
```
rodesai, line 122: why are we dealing with the deserialized command and commandid? `CommandTopic` reads the records as byte[]. Why can't we just back up exactly what we read. I understand deserializing before backing up as a sanity check, but we can still write the raw byte[] from the actual backup operation.

ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/CommandTopicBackupImpl.java
agavra: Before 0.13, if the file was ever corrupted we would simply start a new one and keep the old one. This behavior is changing to fail hard if it's corrupted. After this change, if the old file was created before 0.13 we want to write it in a format that doesn't have the problems described in this PR, but we want to leave the old one in place just in case (at least that was my thoughts).

Does that answer your question?

ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/CommandTopicBackupImpl.java
agavra: Are the raw bytes UTF8 encoded? What you suggest might be a really good solution. The only thing that's important to me is to make sure the backup file is human readable (which it might be if we just write the bytes of what we read anyway)

ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/CommandTopicBackupImpl.java
rodesai: I'm fine with dropping the older backup files for this release. We can do this by configuring a new location in cc-spec-ksql. We would be kicking the problem down the road, so at least file an issue that we need to add something to rewrite to the current version if the backup is old.

agavra: > so at least file an issue that we need to add something to rewrite to the current version if the backup is old.
I don't think we need to do this, the backup file would be written from the beginning of the command topic so we could just ignore the old ones altogether.

rodesai: We still need something to detect that the current version is old and then write out the new file right? (even if we don't delete the file)

ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/CommandTopicBackupImpl.java
rodesai: Yep, it should just be a json string.

agavra: 💯 I like your suggestion best then

ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/CommandTopicBackupImpl.java
agavra: We have two options: (1) we configure a new location in cc-spec and then the old files are as good as having never existed or (2) we use the header approach, which the very first time it starts up on 0.13 it'll write a new file, and from then on it ignores the old one. Perhaps I don't understand what you meant - what we're kicking down the road?

rodesai: This patch adds versioning to the backup. That means we need to be able to deal with the older versions. I think what you're proposing is that if the current backup file's version is old, we simply throw it away? I'm ok with this to keep things simple. These bumps should be very infrequent.

ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/CommandTopicBackupImpl.java
spena: Using this deserialization in the past was unnecessary, but I just detected while fixing this PR. I initially was going with this approach of writing the string directly, but I wanted to keep things compatible with old backups. However, if the new proposal is to change the location in cc-spec-ksql to ignore them, then I'll remove the unnecessary code and write bytes directly.

ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/CommandTopicBackupImpl.java
spena: @almog Are we still doing the versioning now we'll ignore the old backups?

agavra: @spena - feel free to make the call on that. The benefit of adding a header is that in the future if we ever make changes to the functionality we'll have a non-disruptive way forward. If we don't think we'll be making those changes, we can just make it simple.

ksqldb-benchmark/src/main/java/io/confluent/ksql/benchmark/SerdeBenchmark.java
```diff
@@ -183,6 +184,7 @@ private static RowGenerator getRowGenerator(final String schemaName) throws IOEx
       return new RowGenerator(generator, keyField, Optional.empty());
     }
 
+    @SuppressFBWarnings("RCN_REDUNDANT_NULLCHECK_OF_NONNULL_VALUE")
```
spena, line 187: I had to add this because my build fails.

ksqldb-engine/src/main/java/io/confluent/ksql/execution/json/PlanJsonMapper.java
```diff
@@ -32,19 +32,23 @@
...
-          new KsqlTypesDeserializationModule()
-      )
-      .enable(DeserializationFeature.FAIL_ON_UNKNOWN_PROPERTIES)
-      .enable(DeserializationFeature.FAIL_ON_NULL_FOR_PRIMITIVES)
-      .enable(DeserializationFeature.FAIL_ON_NULL_CREATOR_PROPERTIES)
-      .enable(DeserializationFeature.FAIL_ON_INVALID_SUBTYPE)
-      .setSerializationInclusion(Include.NON_EMPTY);
+  private final ObjectMapper mapper = newInstance();
+
+  public static ObjectMapper newInstance() {
```
spena, line 37: Oops, I forgot to remove this. Could this be useful at some point?

ksqldb-engine/src/main/java/io/confluent/ksql/execution/json/PlanJsonMapper.java
```diff
@@ -32,19 +32,23 @@
...
-          new KsqlParserSerializationModule(),
-          new KsqlTypesSerializationModule(),
-          new KsqlTypesDeserializationModule()
-      )
-      .enable(DeserializationFeature.FAIL_ON_UNKNOWN_PROPERTIES)
-      .enable(DeserializationFeature.FAIL_ON_NULL_FOR_PRIMITIVES)
-      .enable(DeserializationFeature.FAIL_ON_NULL_CREATOR_PROPERTIES)
-      .enable(DeserializationFeature.FAIL_ON_INVALID_SUBTYPE)
-      .setSerializationInclusion(Include.NON_EMPTY);
+  private final ObjectMapper mapper = newInstance();
```
agavra, line 35: we don't need this change anymore, right?

ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/CommandTopicBackupImpl.java
```diff
@@ -118,44 +121,47 @@ private boolean isRestoring() {
...
-              .deserialize(record.topic(), record.value())
-      );
-    } catch (Exception e) {
-      LOG.error("Failed to deserialize command topic record when backing it up: {}:{}",
-          record.key(), record.value());
-      return;
+      InternalTopicSerdes.deserializer(CommandId.class).deserialize(record.topic(), record.key());
+    } catch (final Exception e) {
+      throw new KsqlException(String.format(
+          "Cannot deserialize record key: %s",
```
agavra, line 141: let's make these error messages a little bit more meaningful (similar to what you head before, mentioning "while backing it up from the replay file") and also include the exception (so that we know why we can't de-serialize it)

ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/CommandTopicBackupImpl.java
```diff
@@ -209,7 +215,7 @@ private BackupReplayFile newReplayFile() {
 
     File latestBakFile = null;
     if (files != null) {
-      long latestTs = 0;
+      long latestTs = -1;
```
agavra, line 218: is `latestTs` used anywhere?

agavra (approved):
LGTM! Thanks @spena - I can backport this to 0.13 after you've merged it

ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/CommandTopicBackupImpl.java
spena: Changed. It was a change I did in the previous commit when changing the unit tests. It's not needed anymore.

ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/CommandTopicBackupImpl.java
spena: Done

ksqldb-engine/src/main/java/io/confluent/ksql/execution/json/PlanJsonMapper.java
spena: Removed.

Issue #6221, opened 2020-09-16
[NONE Support]: Add a `NONE` format to key-less streams.


big-andy-coates:
To support key-less streams we need a `NONE` format.
1. Add `NONE` format and serde. Serde should same pattern as `KafkaFormat` does on no columns.
1. Reject `NONE` as `VALUE_FORMAT` for now - add QTT test to cover this.
1. `PARTITION BY null` should set key format to `NONE`.
1. `GROUP BY something` and `PARTITION BY something` where the source key format is `NONE` and the user hasn't supplied an explicit key format, should get the key format to use from the system config. 
1. Ensure negative testing:
   - Declaring a table with a key format of `NONE` should fail
   - Declaring a stream or table with key columns and key format `NONE` should fail.
   - CSAS with `GROUP|PARTITION BY something`, where source key format is NONE and default not set in config, should fail
1. Positive testing:
    - CS without key columns and key format NONE.
    - CSAS with `PARTITION BY null`, where source key format is KAFKA, should set sink key format to NONE.
    - CSAS with `PARTITION BY null`, where source key format is NONE, should work and leave key format as-is.
    - CSAS with `GROUP|PARTITION BY something`, where source key format is NONE
       - should key format from WITH clause if supplied (either KEY_FORMAT or FORMAT), and not use config
       -  should pick up from config if not explicitly supplied.
1. Doc updates
    - Document using NONE to override default key format set in config for key-less streams.
     - Update docs on the default key format system config to mention how it affects C*AS statements as well now.
     - Document GROUP BY / PARTITION BY null/something and how this interacts with NONE key formats and config
     - Document NONE not supported as value format yet.

PR #6349, opened 2020-10-02
feat: `NONE` format for key-less streams

big-andy-coates:
### Description 

fixes: #6221

Introduce a `NONE` format, used to indicate the key data, in key-less streams, should not be deserialized.

There's a piece of follow up work to make joins involving key-less streams with `NONE` key format work.  This will be done under https://github.com/confluentinc/ksql/issues/6213.

### Testing done 

usual

### Reviewer checklist

### DIFF
--- a/docs/developer-guide/serialization.md
+++ b/docs/developer-guide/serialization.md
@@ -32,15 +32,74 @@ Serialization Formats
 
 ksqlDB supports these serialization formats:
 
--   `DELIMITED` supports comma separated values. See [DELIMITED](#delimited) below.
--   `JSON` and `JSON_SR` support JSON values. See [JSON](#json) below.
--   `AVRO` supports AVRO serialized values. See [AVRO](#avro) below.
--   `KAFKA` supports primitives serialized using the standard Kafka
-    serializers. See [KAFKA](#kafka) below.
--   `PROTOBUF` supports Protocol Buffers. See [Protobuf](#protobuf) below.
+-   [`NONE`](#none) used to indicate the data should not be derialized.
+-   [`DELIMITED`](#delimited) supports comma separated values.
+-   [`JSON`](#json) and [`JSON_SR`](#json) support JSON values, with and within schema registry integration 
+-   [`AVRO`](#avro) supports AVRO serialized values. 
+-   [`KAFKA`](#kafka) supports primitives serialized using the standard Kafka serializers. 
+-   [`PROTOBUF`](#protobuf) supports Protocol Buffers.
 
 All formats are supported as value formats. Only a subset of formats are
 currently supported as key formats. See individual formats for details.
+
+### NONE
+
+| Feature                      | Supported |
+|------------------------------|-----------|
+| As value format              | No        |
+| As key format                | Yes       |
+| [Schema Registry required][0]| No        |
+| [Schema inference][1]        | No        |
+| [Single field wrapping][2]   | No        |
+| [Single field unwrapping][2] | No        | 
+
+The `NONE` format is a special marker format that is used to indicate ksqlDB should not attempt to 
+deserialize that part of the  {{ site.ak }} record.
+
+It's main use is as the `KEY_FORMAT` of key-less streams, especially where a default key format 
+has been set, via `ksql.persistence.default.format.key`, that supports Schema inference. If the
+key format was not overridden, the server would attempt to load the key schema from the {{ site.sr }}.
+If the schema existed, the key columns would be inferred from the schema, which may not be the intent.
+If the schema did no exist, the statment would be rejected.  In such situations, the key format can
+be set to `NONE`: 
+
+```sql
+CREATE STREAM KEY_LESS_STREAM (
+    VAL STRING
+  ) WITH (
+    KEY_FORMAT='NONE',
+    VALUE_FORMAT='JSON',
+    KAFKA_TOPIC='foo'
+  );
+```
+
+Any statement using format `NONE` that defines columns will result in an error.
+
+If a `CREATE TABLE AS` or `CREATE STREAM AS` statement has a source with a key format of `NONE`, but
+the newly created table or stream has key column, then you may either explicitly define the key 
+format to use in the `WITH` clause, or the default key format, as set in `ksql.persistence.default.format.key`
+will be used.
+
+```sql
+-- keyless stream with NONE key format:
+CREATE STREAM KEY_LESS_STREAM (
+    VAL STRING
+  ) WITH (
+    KEY_FORMAT='NONE',
+    VALUE_FORMAT='JSON',
+    KAFKA_TOPIC='foo'
+  );
+
+-- Table created from stream with explicit key format declared in WITH clause:
+CREATE TABLE T WITH (KEY_FORMAT='KAFKA') AS 
+  SELECT VAL, COUNT() FROM KEY_LESS_STREAM
+  GROUP BY VAL;
+
+-- or, using the default key format set in the ksql.persistence.default.format.key config:
+CREATE TABLE T AS 
+  SELECT VAL, COUNT() FROM KEY_LESS_STREAM
+  GROUP BY VAL;
+```
 
 ### DELIMITED
 
--- a/docs/developer-guide/serialization.md
+++ b/docs/developer-guide/serialization.md
@@ -11,10 +11,8 @@ at runtime. ksqlDB offers several mechanisms for controlling serialization
 and deserialization.
 
 The primary mechanism is by choosing the serialization format when you
-create a stream or table and specify the `VALUE_FORMAT` in the `WITH`
+create a stream or table and specify `FORMAT`, `KEY_FORMAT` or `VALUE_FORMAT` in the `WITH`
 clause.
-
-While ksqlDB supports different value formats, it requires keys to be `KAFKA` format.
 
 ```sql
 -- create table with JSON value format:
@@ -22,6 +20,7 @@ CREATE TABLE ORDERS (
     F0 INT PRIMARY KEY, 
     F1 STRING
   ) WITH (
+    KEY_FORMAT='KAFKA',
     VALUE_FORMAT='JSON', 
     ...
   );
@@ -39,8 +38,8 @@ ksqlDB supports these serialization formats:
 -   [`KAFKA`](#kafka) supports primitives serialized using the standard Kafka serializers. 
 -   [`PROTOBUF`](#protobuf) supports Protocol Buffers.
 
-All formats are supported as value formats. Only a subset of formats are
-currently supported as key formats. See individual formats for details.
+
+Not all formats can be used as both key and value formats. See individual formats for details.
 
 ### NONE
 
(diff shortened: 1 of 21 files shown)


docs/developer-guide/serialization.md
```diff
@@ -32,16 +32,75 @@ Serialization Formats
 
 ksqlDB supports these serialization formats:
 
--   `DELIMITED` supports comma separated values. See [DELIMITED](#delimited) below.
--   `JSON` and `JSON_SR` support JSON values. See [JSON](#json) below.
--   `AVRO` supports AVRO serialized values. See [AVRO](#avro) below.
--   `KAFKA` supports primitives serialized using the standard Kafka
-    serializers. See [KAFKA](#kafka) below.
--   `PROTOBUF` supports Protocol Buffers. See [Protobuf](#protobuf) below.
+-   [`NONE`](#none) used to indicate the data should not be derialized.
```
vcrfxia, line 35:
```suggestion
-   [`NONE`](#none) used to indicate the data should not be deserialized.
```

vcrfxia: Also, can we move this to the end of the list? Seems confusing to have it as the first item (before the actual formats).

docs/developer-guide/serialization.md
```diff
@@ -32,16 +32,75 @@ Serialization Formats
...
+-   [`DELIMITED`](#delimited) supports comma separated values.
+-   [`JSON`](#json) and [`JSON_SR`](#json) support JSON values, with and within schema registry integration 
+-   [`AVRO`](#avro) supports AVRO serialized values. 
+-   [`KAFKA`](#kafka) supports primitives serialized using the standard Kafka serializers. 
+-   [`PROTOBUF`](#protobuf) supports Protocol Buffers.
 
 All formats are supported as value formats. Only a subset of formats are
```
vcrfxia, line 42: This is no longer true if the `NONE` format is not supported as a value format.

docs/developer-guide/serialization.md
```diff
@@ -32,16 +32,75 @@ Serialization Formats
...
+| [Schema Registry required][0]| No        |
+| [Schema inference][1]        | No        |
+| [Single field wrapping][2]   | No        |
+| [Single field unwrapping][2] | No        | 
+
+The `NONE` format is a special marker format that is used to indicate ksqlDB should not attempt to 
+deserialize that part of the  {{ site.ak }} record.
+
+It's main use is as the `KEY_FORMAT` of key-less streams, especially where a default key format 
+has been set, via `ksql.persistence.default.format.key`, that supports Schema inference. If the
```
vcrfxia, line 60: Add link to config?

docs/developer-guide/serialization.md
```diff
@@ -32,16 +32,75 @@ Serialization Formats
...
+key format was not overridden, the server would attempt to load the key schema from the {{ site.sr }}.
+If the schema existed, the key columns would be inferred from the schema, which may not be the intent.
+If the schema did no exist, the statment would be rejected.  In such situations, the key format can
```
vcrfxia, line 63:
```suggestion
If the schema did not exist, the statement would be rejected.  In such situations, the key format can
```

docs/developer-guide/serialization.md
```diff
@@ -32,16 +32,75 @@ Serialization Formats
...
+CREATE STREAM KEY_LESS_STREAM (
+    VAL STRING
+  ) WITH (
+    KEY_FORMAT='NONE',
+    VALUE_FORMAT='JSON',
+    KAFKA_TOPIC='foo'
+  );
+```
+
+Any statement using format `NONE` that defines columns will result in an error.
```
vcrfxia, line 76: Clarify that it's only columns for the relevant portion of the record that fails the statement? I know it's clear after a moment's thought but on first read this sentence as written is confusing, especially given the immediately preceding example that defines columns.

There's also room for misinterpretation in light of the discussion about schema inference above, which is a different scenario under which columns might not be defined.

docs/developer-guide/serialization.md
```diff
@@ -32,16 +32,75 @@ Serialization Formats
...
+
+If a `CREATE TABLE AS` or `CREATE STREAM AS` statement has a source with a key format of `NONE`, but
+the newly created table or stream has key column, then you may either explicitly define the key 
+format to use in the `WITH` clause, or the default key format, as set in `ksql.persistence.default.format.key`
+will be used.
```
vcrfxia, line 81: What about the reverse scenario: the source stream has a key format defined and the CSAS removes the key (via `PARTITION BY NULL`). Does the user need to explicitly set the key format in the CSAS to `NONE`? I assume not, could be good to clarify in the docs.

docs/developer-guide/serialization.md
```diff
@@ -32,16 +32,75 @@ Serialization Formats
...
+
+It's main use is as the `KEY_FORMAT` of key-less streams, especially where a default key format 
+has been set, via `ksql.persistence.default.format.key`, that supports Schema inference. If the
+key format was not overridden, the server would attempt to load the key schema from the {{ site.sr }}.
+If the schema existed, the key columns would be inferred from the schema, which may not be the intent.
+If the schema did no exist, the statment would be rejected.  In such situations, the key format can
+be set to `NONE`: 
+
+```sql
+CREATE STREAM KEY_LESS_STREAM (
```
vcrfxia, line 67: nit: "keyless" is usually one word?

ksqldb-engine/src/main/java/io/confluent/ksql/serde/InternalFormats.java
```diff
@@ -50,10 +52,15 @@ private InternalFormats() {
    * @see SerdeFeaturesFactory#buildInternal
    */
   public static Formats of(final KeyFormat keyFormat, final ValueFormat valueFormat) {
+    // Do not use NONE format for internal topics:
```
vcrfxia, line 55: Is the thinking here that JSON works even if no key columns are specified, so we might as well always use JSON so that if a stream with no key columns is the source for a another stream/table down the line, then the sink stream/table will automatically get JSON as the key format, regardless of what the config default key format is?

ksqldb-functional-tests/src/test/resources/query-validation-tests/none.json
```diff
@@ -0,0 +1,181 @@
...
+    },
+    {
+      "name": "explicitly set in CTAS",
+      "statements": [
+        "CREATE STREAM INPUT (foo INT, bar INT) WITH (kafka_topic='input_topic', key_format='KAFKA', value_format='JSON');",
+        "CREATE TABLE OUTPUT WITH (key_format='NONE') AS SELECT BAR AS K, COUNT() FROM INPUT GROUP BY BAR;"
+      ],
+      "expectedException": {
+        "type": "io.confluent.ksql.util.KsqlStatementException",
+        "message": "The 'NONE' format can only be used when no columns are defined. Got: [`K` INTEGER KEY]"
```
vcrfxia, line 102: Not directly related to this PR: I notice the printing of the schema doesn't distinguish between `KEY` and `PRIMARY KEY`. What was the reasoning for that? Is it just easier to not worry about passing in information about source type to the schema, or is the current behavior preferred outside of the logistical concern too?

ksqldb-functional-tests/src/test/resources/query-validation-tests/none.json
```diff
@@ -0,0 +1,181 @@
...
+      "statements": [
+        "CREATE TABLE INPUT (K STRING PRIMARY KEY, foo INT) WITH (kafka_topic='input_topic', key_format='NONE', value_format='JSON');"
+      ],
+      "expectedException": {
+        "type": "io.confluent.ksql.util.KsqlStatementException",
+        "message": "The 'NONE' format can only be used when no columns are defined. Got: [`K` STRING KEY]"
+      }
+    },
+    {
+      "name": "as value format with value columns",
```
vcrfxia, line 44: Can we add a test for what happens if someone tries to set NONE as value format with no value columns? I assume this is not supported. It'd be good to check that the error message is sensible.

vcrfxia (approved):
Thanks @big-andy-coates -- LGTM! Questions inline as usual.

ksqldb-engine/src/main/java/io/confluent/ksql/serde/InternalFormats.java
vcrfxia: Perhaps I've missed it but I don't see any QTT for testing this. Can we add one in order to prevent regressions here?

docs/developer-guide/serialization.md
big-andy-coates: I deliberately put it at the top of the list as it's a 'special'  format.

ksqldb-engine/src/main/java/io/confluent/ksql/serde/InternalFormats.java
big-andy-coates: >  if no key columns are specified

Internal topics _always_ have a key column.

This is just stop-gap functionality. I'm working on better handling of internal key format in another PR.

ksqldb-functional-tests/src/test/resources/query-validation-tests/none.json
big-andy-coates: The text is coming from `LogicalSchema` / `Column`, which do not differentiate between primary key and key. 

We could just strip off the KEY all together, but I don't really see it as a big thing.  The error message has enough information that the user can correct the error.

ksqldb-engine/src/main/java/io/confluent/ksql/serde/InternalFormats.java
vcrfxia: If that's the case, why would we ever arrive in this situation (of trying to build an internal topic with key format NONE)? Can we just fail if that happens, instead of converting the format to JSON?

ksqldb-functional-tests/src/test/resources/query-validation-tests/none.json
big-andy-coates: You'd think no value columns would not work right?  Nope... it works, and just fails if you try and use the source.

(See https://github.com/confluentinc/ksql/issues/6352)

docs/developer-guide/serialization.md
vcrfxia: I realize but it still feels a bit jarring to have it listed first since it's usage will be relatively uncommon. If the goal is to call attention to the format as being different from the others, maybe we can split this list into two: one for "regular formats" and one for "special formats" or something similalr.

big-andy-coates:
Any follow up comments or issues @vcrfxia, just let me know and I can always pick them up in the next PR.

ksqldb-functional-tests/src/test/resources/query-validation-tests/none.json
vcrfxia: That's unfortunate... is it worth slipping a line into the docs for the new `NONE` format to clarify that ksqlDB doesn't yet support empty value columns, so the `NONE` format should only be used as key format for now? Could also fail up front if users try to use the `NONE` format as a value format but that might not be worth the effort since it's code that will eventually be removed.

PR #6353, opened 2020-10-02
feat: Add an endpoint for returning the query limit configuration

### LINKED ISSUE #6275: feat: Add a ksqlDB API for returning the query limit configuration (and other configs)
It would be good for the ksqlDB server to have an endpoint API that allows clients to request configs from the server. For instance, some cliens in CCloud require to know what's the max. number of persistent queries the server can handle.

An endpoint that returns only specific and non-sensitive configs.

For example, we can start with the requested config for CCloud for now. We can add more later, or we can use a query parameter to request a specific config.

Example:
```
GET /v1/configs
{
  "ksql.query.persistent.active.limit": "20"
}
```
or
```
GET /v1/configs&name=ksql.query.persistent.active.limit
{
  "ksql.query.persistent.active.limit": "20"
}
```

An alternative is to use the `SET` command to fetch the config value.
i.e.
```
ksql> set 'ksql.query.persistent.active.limit';
20
```

The above is not implemented yet. It currently fails. I'm not  sure if implemented will cause other problems with users not knowing if they have set or not set a config. For instance:
```
ksql> set 'auto.offset.reset';
latest
ksql> set 'auto.offset.reset' = 'earliest';
ksql> set 'auto.offset.reset';
earliest
ksql> unset 'auto.offset.reset'
ksql> set 'auto.offset.reset';
latest
```
^ If I call the last command, how do I know I did set the config manually or is the one fetched from the server?

jzaralim:
### Description 
This PR adds a new ksqlDB API endpoint for requesting specific configs. Currently, only the persistent query limit can be requested. 

Fixes #6275 

### Testing done 
Manually tested with:

```
 curl "http://localhost:8088/v1/configs"
{"configs":{"ksql.query.persistent.active.limit":20}}

 curl "http://localhost:8088/v1/configs?name=foo"
{"configs":{}}

 curl "http://localhost:8088/v1/configs?name=ksql.query.persistent.active.limit&name=foo"
{"configs":{"ksql.query.persistent.active.limit":20}}
```

### Reviewer checklist


ksqldb-rest-model/src/main/java/io/confluent/ksql/rest/entity/ConfigResponse.java
```diff
@@ -0,0 +1,65 @@
...
+import com.fasterxml.jackson.annotation.JsonProperty;
+import com.fasterxml.jackson.annotation.JsonSubTypes;
+import com.google.common.collect.ImmutableMap;
+import java.util.Map;
+import java.util.Objects;
+
+@JsonIgnoreProperties(ignoreUnknown = true)
+@JsonSubTypes({})
+public class ConfigResponse {
+  private final ImmutableMap<String, Object> configs;
```
spena, line 31: Even if this is immutable, it's good practice to use the interface `Map` here when declaring the variable.

spena (approved):
LGTM

ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/resources/ConfigResource.java
```diff
@@ -0,0 +1,52 @@
...
+
+import io.confluent.ksql.rest.EndpointResponse;
+import io.confluent.ksql.rest.entity.ConfigResponse;
+import io.confluent.ksql.util.KsqlConfig;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+
+public class ConfigResource {
+  private final Map<String, Object> allowedConfigs = new HashMap<>();
```
spena, line 26: Perhaps use `visibleConfigs` instead. Just to provide more meaning to what these configs are.

ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/resources/ConfigResource.java
```diff
@@ -0,0 +1,52 @@
...
+        KsqlConfig.KSQL_ACTIVE_PERSISTENT_QUERY_LIMIT_CONFIG,
+        ksqlConfig.getInt(KsqlConfig.KSQL_ACTIVE_PERSISTENT_QUERY_LIMIT_CONFIG)
+    );
+  }
+
+  public EndpointResponse getConfigs(final List<String> requestedConfigs) {
+    final Map<String, Object> configs = new HashMap<>();
+    for (String config : requestedConfigs) {
+      if (visibleConfigs.containsKey(config)) {
+        configs.put(config, visibleConfigs.get(config));
```
agavra, line 43: nit: generally good practice to avoid a double lookup. Obviously not so important here as it's not a perf critical area, but figured I'd drop the comment so it's top of mind next time :) 
```java
Object value = visibleConfigs.get(config);
if (value != null) {
  configs.put(config, value)
}
```

ksqldb-rest-app/src/test/java/io/confluent/ksql/rest/integration/RestApiTest.java
```diff
@@ -289,6 +291,28 @@ public void shouldExecuteServerMetadataIdRequest() {
     assertThat(response, is(notNullValue()));
   }
 
+  @Test
+  public void shouldExecuteAllConfigsRequest() {
```
agavra, line 295: we should add tests to make sure that we don't accidentally leak any config that isn't in the allowed list of configs

agavra (approved):
LGTM! I think there's probably corresponding documentation updates that need to be made though... 😉

JimGalasyn (approved):
LGTM!

Issue #6286, opened 2020-09-23
feat: new command to restore the ksqlDB command topic


spena:
Currently, ksqlDB backup the command topic onto a backup file. When users need to restore this file, they require some manual steps, including downloading Kafka tools, to achieve that.

It would be good to have a command that allows users easily restore their command topic. from a backup file.
i.e.
```
$ ksql-command-topic-restore <server-properties> <backup-file>
```

The new command should have enough information from the <server-properties> to know:
- The Kafka host and credentials
- The command topic name
- The internal topic replicas/partitions
- Serialization of the command topic
- Etc.

**Other parameters**
We should add a flag that filters out commands from the backup-file that are not compatible with the current ksqlDB version. This will allow us use backups files on older versions of ksqlDB. Each command in the command topic and backup have a `version` which specifies which version that command was created.

PR #6361, opened 2020-10-05
feat: new command to restore ksqlDB command topic backups

spena:
### Description 
Fixes https://github.com/confluentinc/ksql/issues/6286

This is a new command that helps to restore command topic backups:
```
NAME
        ksql-restore-metadata - KSQL Restore Metadata

SYNOPSIS
        ksql-restore-metadata --config-file <configFile> [ {-h | --help} ]
                [ {--yes | -y} ] [--] <backup-file>

OPTIONS
        --config-file <configFile>
            A file specifying configs for the KSQL Server, KSQL, and its
            underlying Kafka Streams instance(s). Refer to KSQL documentation
            for a list of available configs.

        -h, --help
            Display help information

        --yes, -y
            Automatic "yes" as answer to prompt and run non-interactively.

        --
            This option can be used to separate command-line options from the
            list of arguments (useful when arguments might be mistaken for
            command-line options)

        <backup-file>
            A file specifying the file that contains the metadata backup.
```

With default options, the command prompts the user to answer if continuing with the restore process or not. This is required to avoid an accidental restore process (the option -y will assume "yes" as the answer).
```
./bin/ksql-restore-command-topic --config-file <config-file> <backup-file>
Restoring the command topic will DELETE your actual metadata.
Continue [yes or no] (default: no)? 
```

The restoration process displays timing information in the console to let the user know how much time it took to restore the command topic:
```
./bin/ksql-restore-command-topic --config-file <config-file> <backup-file> --yes
Loading backup file ...
Backup (1 records) loaded in memory in 347 ms.

Restoring command topic ...
Restore process completed in 1164 ms.

You need to restart the ksqlDB server to re-load the command topic.
```

### Testing done 
Added unit tests
Verified manually

### Reviewer checklist


stevenpyzhang (commented):
What should we do in the case that the restore is interrupted/fails in the middle of the process? We'll have a command topic with only a partial set of commands. I'm wondering if we should just delete this incomplete command topic.

stevenpyzhang (changes requested):
I think we might actually need to use transactions to restore the command topic.
When the command topic is produced to normally, there are transactional messages produced to the command topic that take up offsets. If we restore without using transactions, this could mean the restored command topic will have a lower offset than the original command topic, that could result in query id's conflicting since we generate query id based on the offset.

For example, there are 10 command records in command topic (offset would be 15-20 potentially)
After restoring 10 command records using this tool, the offset would be 10.

We don't have to go through validation/wait for the CommandRunner like we do in DistributingExecutor so the protocol here can be much simpler.

```
init transaction
for each command in the backupfile
   begin transaction
   enqueue command
   commit transaction
close producer
```

ksqldb-streams/src/test/java/io/confluent/ksql/execution/streams/StreamAggregateBuilderTest.java
```diff
@@ -431,6 +431,7 @@ public void shouldBuildAggregatorParamsCorrectlyForUnwindowedAggregate() {
   }
 
   @Test
+  @SuppressWarnings("RV_RETURN_VALUE_IGNORED_NO_SIDE_EFFECT")
```
stevenpyzhang, line 434: nit: we shouldn't check these in if this is a specific checkstyle error to your local environment

ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/BackupReplayFile.java
```diff
@@ -16,36 +16,21 @@
...
-  private static final byte[] KEY_VALUE_SEPARATOR_BYTES =
-      KEY_VALUE_SEPARATOR_STR.getBytes(StandardCharsets.UTF_8);
-  private static final byte[] NEW_LINE_SEPARATOR_BYTES =
-      NEW_LINE_SEPARATOR_STR.getBytes(StandardCharsets.UTF_8);
-
-  private final File file;
+public final class BackupReplayFile extends BackupInputFile implements Closeable {
   private final FileOutputStream writer;
 
   public BackupReplayFile(final File file) {
```
stevenpyzhang, line 32: Do we need to actually split this into two classes? `BackupReplayFile` and `BackupInputFile`

It doesn't really look like there's much difference between the two so it's not clear to me why having both is necessary.

stevenpyzhang (approved):
LGTM, would it be possible to also add an integration test as part of this pr? We could start up a server with backups enabled, issue a bunch of DDL statements, stop server, delete command topic, run the executable to restore, start up new server and verify the metastore matches what it was previously.

ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/BackupReplayFile.java
spena: There's a slightly difference. The `BackupReplayFile` opens the file for writing. It fails if there are no permissions for writing. I created the `BackupInputFile` just for that, so it can open it if only read permissions are set.

ksqldb-rest-app/src/test/java/io/confluent/ksql/rest/server/restore/KsqlRestoreCommandTopicTest.java
```diff
@@ -0,0 +1,308 @@
...
+  @Test
+  public void shouldCreateAndRestoreCommandTopic() throws ExecutionException, InterruptedException {
+    // Given:
+    when(topicClient.isTopicExists(COMMAND_TOPIC_NAME)).thenReturn(false);
+
+    // When:
+    restoreCommandTopic.restore(BACKUP_COMMANDS);
+
+    // Then:
+    verifyCreateTopic(COMMAND_TOPIC_NAME);
```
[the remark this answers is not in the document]
spena: Done

ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/BackupReplayFile.java
spena: I reverted the BackupInputFile. I added two static methods to the BackupReplayFile:
```
BackupReplayFile.readOnly(file);
BackupReplayFile.writetable(file);
```

spena:
Added an integration tests

devdev606 (approved):

Issue #6222, opened 2020-09-16
[SR Support]: Add support for reading & writing key schemas from & to SR


big-andy-coates:
1. Add support for reading and writing schemas to and from the schema registry.
 * Pick up https://github.com/confluentinc/ksql/issues/6133 as part of this work.
  * Register key schema on CREATE and CREATE AS statements in the same way as is done for value schemas.
  * Delete key subject on delete (and value subject for non-Avro, which may not currently be happening)
  * Delete key subject on internal topics too when query terminated (and value subject for non-Avro, which may not currently be happening)
2. Doc updates
 - Document the subject name for key schemas (and value if not already done) and when schemas are created / deleted

agavra:
The above two PRs should get the bulk of this functionality done. We'll track follow-ups separately.

PR #6363, opened 2020-10-05
chore: wire in auto-register for key schemas

agavra:
### Description 
Partial work for #6222 - this PR supports _writing_ to schema registry (and ensuring compatibility) for formats that support schema inference. This PR also includes most of the piping we need to test it in QTT, but I haven't actually written any of those tests until we have `AVRO` key formats wired up.

### Testing done 
Unit testing

### Reviewer checklist

### DIFF
--- a/ksqldb-engine/src/main/java/io/confluent/ksql/schema/ksql/inference/SchemaRegisterInjector.java
+++ b/ksqldb-engine/src/main/java/io/confluent/ksql/schema/ksql/inference/SchemaRegisterInjector.java
@@ -86,16 +86,6 @@ public class SchemaRegisterInjector implements Injector {
         schema,
         FormatFactory.of(keyFormat)
     );
-    registerSchema(
-        schema.key(),
-        statement.getProperties().getKafkaTopic(),
-        keyFormat,
-        keyFeatures,
-        cs.getSessionConfig().getConfig(false),
-        cs.getStatementText(),
-        false,
-        KsqlConstants.SCHEMA_REGISTRY_KEY_SUFFIX
-    );
 
     final FormatInfo valueFormat = SourcePropertiesUtil.getValueFormat(statement.getProperties());
     final SerdeFeatures valFeatures = SerdeFeaturesFactory.buildValueFeatures(
@@ -105,15 +95,16 @@ public class SchemaRegisterInjector implements Injector {
         cs.getSessionConfig().getConfig(false)
     );
 
-    registerSchema(
-        schema.withoutPseudoAndKeyColsInValue().value(),
+    registerSchemas(
+        schema,
         statement.getProperties().getKafkaTopic(),
+        keyFormat,
+        keyFeatures,
         valueFormat,
         valFeatures,
         cs.getSessionConfig().getConfig(false),
         cs.getStatementText(),
-        false,
-        KsqlConstants.SCHEMA_REGISTRY_VALUE_SUFFIX
+        false
     );
   }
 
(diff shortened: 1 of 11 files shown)


ksqldb-functional-tests/src/main/java/io/confluent/ksql/test/model/TopicNode.java
```diff
@@ -39,20 +38,26 @@
...
   private final JsonNode schema;
   private final int numPartitions;
   private final int replicas;
+  private final String keyFormat;
   private final String format;
 
   public TopicNode(
       @JsonProperty("name") final String name,
+      @JsonProperty("keySchema") final JsonNode keySchema,
       @JsonProperty("schema") final JsonNode schema,
```
vcrfxia, line 51: Can we rename `schema` and `format` to `valueSchema` and `valueFormat`? I realize there's a little extra effort needed to maintain backwards compatibility for users of the testing tool but I think not doing the rename will be confusing down the line, especially since `format` is used throughout the rest of the codebase to mean both key and value format (e.g., `CREATE STREAM ... WITH (FORMAT='JSON', ...);` is valid now).

ksqldb-functional-tests/src/main/java/io/confluent/ksql/test/tools/TestCaseBuilderUtil.java
```diff
@@ -147,12 +148,27 @@ private static Topic createTopicFromStatement(
...
+      final LogicalSchema logicalSchema = statement.getElements().toLogicalSchema();
+
+      final FormatInfo keyFormatInfo = SourcePropertiesUtil.getKeyFormat(props);
+      final Format keyFormat = FormatFactory.fromName(keyFormatInfo.getFormat());
+      final SerdeFeatures keySerdeFeats = buildKeyFeatures(
+          keyFormat, logicalSchema);
+
+      final Optional<ParsedSchema> keySchema =
+          keyFormat.supportsFeature(SerdeFeature.SCHEMA_INFERENCE)
+              ? buildValueSchema(sql, logicalSchema.key(), keyFormatInfo, keyFormat, keySerdeFeats)
```
vcrfxia, line 160: Rename method from `buildValueSchema` to `buildSchema` to avoid confusion? (And rename the params in the method as well)

ksqldb-functional-tests/src/main/java/io/confluent/ksql/test/tools/TestCaseBuilderUtil.java
```diff
@@ -228,6 +238,21 @@ private static Topic createTopicFromStatement(
     }
   }
 
+  private static SerdeFeatures buildKeyFeatures(
+      final Format valueFormat,
```
vcrfxia, line 242: Rename from `valueFormat` to `keyFormat` to avoid confusion?

ksqldb-functional-tests/src/main/java/io/confluent/ksql/test/tools/TestExecutorUtil.java
```diff
@@ -177,28 +177,37 @@ private static Topic buildSinkTopic(
...
+        .fromName(format);
 
     if (!valueFormat.supportsFeature(SerdeFeature.SCHEMA_INFERENCE)) {
       return Optional.empty();
     }
 
     try {
-      final String subject =
-          dataSource.getKafkaTopicName() + KsqlConstants.SCHEMA_REGISTRY_VALUE_SUFFIX;
+      final String subject = topicName + KsqlConstants.SCHEMA_REGISTRY_VALUE_SUFFIX;
```
vcrfxia, line 210: This method is called is called to retrieve the key schema above (line 180) even though the method hardcodes the value suffix. Is key schema inference meant to be supported in QTT / the testing tool in this PR or not? I see in `initializeTopics()` below that only the value schema is checked as well.

ksqldb-engine/src/main/java/io/confluent/ksql/schema/ksql/inference/SchemaRegisterInjector.java
```diff
@@ -76,27 +78,42 @@ private void registerForCreateSource(final ConfiguredStatement<? extends CreateS
...
         FormatFactory.of(valueFormat),
-        cs.getStatement().getProperties().getValueSerdeFeatures(),
+        statement.getProperties().getValueSerdeFeatures(),
         cs.getSessionConfig().getConfig(false)
     );
 
     registerSchema(
-        schema,
-        cs.getStatement().getProperties().getKafkaTopic(),
+        schema.withoutPseudoAndKeyColsInValue().value(),
```
vcrfxia, line 109: nit: not your code but can we add a quick unit test to ensure that `withoutPseudoAndKeyColsInValue()` here isn't lost? There doesn't appear to be one right now.

vcrfxia (approved):
Thanks @agavra -- LGTM with the usual questions inline!

Also, I've re-titled the PR from `feat` to `chore` since I think the feature(s) that will be of interest to users are when AVRO and JSON_SR are available, rather than the individual steps we need in order to get there.

ksqldb-engine/src/main/java/io/confluent/ksql/schema/ksql/inference/SchemaRegisterInjector.java
big-andy-coates: This might actually not be needed. Be interesting to see if this can be removed.

ksqldb-engine/src/main/java/io/confluent/ksql/schema/ksql/inference/SchemaRegisterInjector.java
```diff
@@ -115,24 +132,36 @@ private void registerForCreateAs(final ConfiguredStatement<? extends CreateAsSel
...
+        queryMetadata.getResultTopic().getKafkaTopicName(),
+        queryMetadata.getResultTopic().getKeyFormat().getFormatInfo(),
+        queryMetadata.getPhysicalSchema().keySchema().features(),
+        cas.getSessionConfig().getConfig(false),
+        cas.getStatementText(),
+        true,
+        KsqlConstants.SCHEMA_REGISTRY_KEY_SUFFIX
+    );
+    registerSchema(
+        queryMetadata.getLogicalSchema().withoutPseudoAndKeyColsInValue().value(),
```
big-andy-coates, line 145: minor: consider adding a `registerSchemas` method, which takes the logical schema and both formats. Internally, it justs called `registerSchema` for key then value.

This would make the code easier to read and remove a bit of duplication.

ksqldb-functional-tests/src/main/java/io/confluent/ksql/test/model/TopicNode.java
agavra: sure, I'll do this in a follow-up PR since it requires changing tons of test files

ksqldb-functional-tests/src/main/java/io/confluent/ksql/test/tools/TestExecutorUtil.java
agavra: I started going down the path of supporting QTT tests, but because AVRO key formats aren't yet supported I couldn't get too far with testing. I figured the changes I've made here will be useful in a bit so I just left them in - but it is (as you can tell) a little raw...

I'll clean it up.

ksqldb-engine/src/main/java/io/confluent/ksql/schema/ksql/inference/SchemaRegisterInjector.java
agavra: I confirmed anecdotally and also by following the code path that it isn't necessary. The statement as passed in as a statement (as written) and the logical schema is derived from the `getElements` of the statement. 

Also `withPseudoAndKeyColsInValue` is only called from `LogicalPlanner`, `Analyzer` and `SourceBuilder`, all of which happen after the injector phase.

PR #6365, opened 2020-10-06
Query schemas

big-andy-coates:
### Description 

When building a query, the `KsqlQueryBuilder` captures the schema and serde features used when creating each value serde.  This is later used by QTT to build historical plans that capture this information, giving us confidence that any changes we make aren't changing the schema and serde features being used by historical queries.

This commit extends this to capture details from _key_ serde, not just _value_ serde, and to capture full `KeyFormat` and `ValueFormat` info.

The benefit is two fold:  increased test coverage of historical plans and its a necessary improvement to enable testing of non-KAFKA key formats. A follow up PR will improve the `TopicInfoCache` to use this more detailed information.

The first commit is the code changes, the second the updated historical plans.  

### Testing done 

usual

### Reviewer checklist
- [ ] Ensure docs are updated if necessary. (eg. if a user visible feature is being added or changed).
- [ ] Ensure relevant issues are linked (description should include text like "Fixes #<issue number>")

PR #6366, opened 2020-10-06
Topic info cache

big-andy-coates:
### Description 

The `TopicInfoCache` is used by QTT to retrieve info about the topics being used by the test case. Specifically, its used to get at the schema, key and value formats of the topic, and potentially use this information to build appropriate serde.

The implementation wasn't correctly handling internal topics. When building information about internal topics it was using the key and value formats of the sink topic. This is often incorrect.  For example, consider a windowed group by where the source needs to be repartitioned, i.e. the `GROUP BY` is not on the key. In such a situation the repartition topic doesn't have a windowed key, which the sink topic does.

To get the correct key format, the `TopicInfoCache` now gets the information from the query itself. Mainly the `QuerySchemas`, which holds the logical schema plus the key and value formats used when creating the serde passed to Kafka Streams.

### Testing done 

usual

### Reviewer checklist


ksqldb-execution/src/test/java/io/confluent/ksql/execution/builder/KsqlQueryBuilderTest.java
```diff
@@ -281,4 +282,60 @@ public void shouldTrackValueSchemasUsed() {
             SerdeFeatures.of(SerdeFeature.WRAP_SINGLES)))
     )));
   }
+
+  @Test
```
big-andy-coates, line 286: Changes in this file are related to review comment in https://github.com/confluentinc/ksql/pull/6365

ksqldb-functional-tests/src/test/java/io/confluent/ksql/test/model/SchemaNodeTest.java
```diff
@@ -49,12 +49,12 @@ public void shouldRoundTrip() {
   }
 
   @Test
-  public void shouldRoundNoKeyFormat() {
+  public void shouldRoundTripNoKeyFormat() {
```
big-andy-coates, line 52: Changes in this file are related to review comment in https://github.com/confluentinc/ksql/pull/6365

ksqldb-functional-tests/src/main/java/io/confluent/ksql/test/tools/TopicInfoCache.java
```diff
@@ -87,28 +155,34 @@ public void clear() {
...
-            ? OptionalLong.of(DurationParser.parse(windowedJoinMatcher.group(1)).toMillis())
+            ? OptionalLong
+            .of(DurationParser.parse(windowedJoinMatcher.group("duration")).toMillis())
             : OptionalLong.empty();
 
         return new TopicInfo(
             topicName,
             query.getLogicalSchema(),
-            query.getResultTopic().getKeyFormat(),
-            query.getResultTopic().getValueFormat(),
```
big-andy-coates, line 111: This is the bad code. It was getting the internal topic's key and value format from the sink topic... which just ain't right ;)

ksqldb-functional-tests/src/main/java/io/confluent/ksql/test/tools/TopicInfoCache.java
```diff
@@ -26,36 +26,104 @@
...
+ * against the topic name to determine the correct {@code loggerNamePrefix} to look up and any
+ * additional logic needded.
  */
 public class TopicInfoCache {
 
-  private static final Pattern INTERNAL_TOPIC_PATTERN = Pattern
-      .compile("_confluent.*query_(.*_\\d+)-.*-(changelog|repartition)");
+  private static final String TOPIC_PATTERN_PREFIX = "_confluent.*query_(?<queryId>.*_\\d+)-";
+
+  private static final List<InternalTopicPattern> INTERNAL_TOPIC_PATTERNS = ImmutableList.of(
```
agavra, line 74: this class seems super fragile... we're extracting information from internal topic naming conventions, mapping that to logger names, which we use to extract the serde info that we pass to Kafka Streams, and then magically deduce from that what the actual serde should be based on what we know Kafka Streams does. I'm sure that coding this up you had a bad feeling in your stomach, this just doesn't sit right! Can we do something else?

What we need is a mapping from topic name to logger prefix. We could have the `LoggingSerializer` aware of the prefix that it's using, and register this mapping whenever it (de)serializes anything (and it's easy to pass in the prefix there, we have it in `GenericSerdeFactory`). Since we probably don't want to maintain this mapping in PROD as it might have real perf implications to do this per-message, we could do the evil thing of having a static map that is populated and cleared just for tests.

I don't love that solution either, but maybe it'll get the brain juices flowing and I think it's a lot more robust (even if it breaks some encapsulation) than what's here.

agavra (commented):
TBH, I didn't love this class when we introduced it, but it's grown to become a little monster 😨 I'd like to try out the suggestion I have below, and hopefully it'll work out, but if after that we decide there isn't anything better than we can just merge this to unblock us - at least it only makes the testing framework brittle and not production code

ksqldb-functional-tests/src/main/java/io/confluent/ksql/test/tools/TopicInfoCache.java
big-andy-coates: You're right, of course. Was just trying to get this done so could move on, but looking,...

big-andy-coates: Your wish is my command...

ksqldb-execution/src/main/java/io/confluent/ksql/schema/query/QuerySchemas.java
```diff
@@ -17,39 +17,147 @@
...
  */
-@Immutable
 public final class QuerySchemas {
 
-  private final ImmutableMap<String, SchemaInfo> schemas;
+  // Maps logger name prefixes -> Schema Info
+  private final LinkedHashMap<String, SchemaInfo> loggerToSchemas = new LinkedHashMap<>();
+
+  // Maps topic name -> (map of key/value flag -> set of logger name prefixes)
+  private final Map<String, Map<Boolean, Set<String>>> topicsToLoggers = new HashMap<>();
```
agavra, line 57: 😂 seems like overkill for a `Map` but we can refactor that another time

ksqldb-execution/src/main/java/io/confluent/ksql/execution/builder/KsqlQueryBuilder.java
```diff
@@ -207,7 +217,18 @@ public KsqlQueryBuilder withKsqlConfig(final KsqlConfig newConfig) {
...
         serviceContext.getSchemaRegistryClientFactory(),
         loggerNamePrefix,
-        processingLogContext
+        processingLogContext,
+        getSerdeTracker(loggerNamePrefix)
+    );
+  }
+
+  private Optional<TrackedCallback> getSerdeTracker(final String loggerNamePrefix) {
+    if (System.getProperty(KSQL_TEST_TRACK_SERDE_TOPICS) == null) {
```
agavra, line 226: nit: worth checking equality to `true` in case people think they can just flip it from "true"/"false"

ksqldb-functional-tests/src/main/java/io/confluent/ksql/test/tools/TopicInfoCache.java
```diff
@@ -138,6 +168,151 @@ private TopicInfo load(final String topicName) {
...
+    /**
+     * Gives the pattern a chance to adjust the key format
+     */
+    KeyFormat keyFormat(final KeyFormat baseFormat, final PersistentQueryMetadata query) {
+      return baseFormat;
+    }
+
+    /**
+     * Used by stream-stream join changelogs of windowed stream, where the statestore key is
+     * double-wrapped in {@link org.apache.kafka.streams.kstream.Windowed}. This can be represented
```
agavra, line 192:
```suggestion
     * double-wrapped in {@link org.apache.kafka.streams.kstream.Windowed}. This can't be represented
```

agavra (approved):
😂 thanks @big-andy-coates

ksqldb-execution/src/main/java/io/confluent/ksql/execution/builder/KsqlQueryBuilder.java
big-andy-coates: Meh, it's not something anyone should be setting every, except in QTT, so I don't see it as an issue.

PR #6375, opened 2020-10-07
feat: Implement a physical plan builder and physical plan for pull queries

vpapavas:
### Description 
This aims to replace the `PullQueryExecutor`. I essentially moved all the code from the `PullQueryExecutor` into physical plan operators. The physical plan execution is pipelined, each operator works on a row that is fed from the data store. There is no new functionality, the physical plan supports everything the `PullQueryExecutor` supports namely,  single key lookups on windowed and non-windowed tables and IN queries.

### Testing done 
Make sure all prior tests pass.
Checked all benchmarks and performance is comparable (a little bit slower 3K vs 2.8K qps for single node non-windowed queries) which is expected since we do the logical planning which was not done in the `PullQueryExecutor`.

### Reviewer checklist


ksqldb-engine/src/main/java/io/confluent/ksql/engine/EngineExecutor.java
```diff
@@ -221,6 +314,41 @@ private ExecutorPlans planQuery(
     return new ExecutorPlans(logicalPlan, physicalPlan);
   }
 
+  private LogicalPlanNode buildAndValidateLogicalPlan(
+      final ConfiguredStatement<?> statement,
+      final ImmutableAnalysis analysis,
+      final KsqlConfig config
+  ) {
+    final OutputNode outputNode = new LogicalPlanner(config, analysis, engineContext.getMetaStore())
```
AlanConfluent, line 322: So we're just using the existing, unchanged `LogicalPlanner` to begin with?  Looking through it, it probably handles the basic logical nodes we would want.  I can see us doing more of our analysis and error checking there (e.g. not more than one key comparison for pull queries, etc. rather than doing it in the physical planner since that seems too far downstream).   If we handle more complex operations in the future, we'd obviously want to have a logical component.

ksqldb-engine/src/main/java/io/confluent/ksql/engine/EngineExecutor.java
```diff
@@ -403,4 +531,66 @@ private String buildPlanSummary(final QueryId queryId, final ExecutionStep<?> pl
...
+    @Override
+    public Optional<Expression> visitQualifiedColumnReference(
+        final QualifiedColumnReferenceExp node,
+        final Context<Void> ctx
+    ) {
+      return Optional.of(new UnqualifiedColumnReferenceExp(node.getColumnName()));
+    }
+  }
+
+  private static final class ConfigRoutingOptions implements RoutingOptions {
```
AlanConfluent, line 551: Maybe pull this out to its own class?

ksqldb-engine/src/main/java/io/confluent/ksql/engine/EngineExecutor.java
```diff
@@ -118,6 +139,78 @@ ExecuteResult execute(final KsqlPlan plan) {
...
+      checkRateLimit(rateLimiter);
+    }
+
+
+    try {
+      final QueryAnalyzer queryAnalyzer = new QueryAnalyzer(engineContext.getMetaStore(), "");
+      final ImmutableAnalysis analysis = new RewrittenAnalysis(
+          queryAnalyzer.analyze(statement.getStatement(), Optional.empty()),
+          new ColumnReferenceRewriter()::process
+      );
```
AlanConfluent, line 189: Is this not the type of logic that would live in the logical planner? Or at least in `buildAndValidateLogicalPlan`

ksqldb-engine/src/main/java/io/confluent/ksql/engine/EngineExecutor.java
```diff
@@ -221,6 +314,41 @@ private ExecutorPlans planQuery(
...
+        statement.getStatementText(),
+        Optional.of(outputNode)
+    );
+  }
+
+  private PullPhysicalPlan buildPullPhysicalPlan(
+      final KsqlExecutionContext ksqlEngine,
+      final LogicalPlanNode logicalPlan,
+      final KsqlConfig config,
+      final ImmutableAnalysis analysis,
```
AlanConfluent, line 334: This analysis exposes all of the parser expressions.  Isn't that something that should be at the logical planner layer?  I would assume that the physical planner would take some logical nodes which have all of the relevant data extracted from the expressions.

Maybe this is just a first step, and there's still some intermixing of layers, but I just thought I would ask.

ksqldb-engine/src/main/java/io/confluent/ksql/physical/pull/PullPhysicalPlanBuilder.java
```diff
@@ -0,0 +1,342 @@
...
+    whereInfo = WhereInfo.extractWhereInfo(logicalNode, analysis, persistentQueryMetadata);
+    return new SelectOperator(logicalNode);
+  }
+
+  private AbstractPhysicalOperator translateDataSourceNode(
+      final DataSourceNode logicalNode,
+      final PersistentQueryMetadata persistentQueryMetadata
+  ) {
+
+    key = asKeyStruct(
```
AlanConfluent, line 215: Is there anything that sanity checks the order in which these methods will be called?  For example, will `whereInfo` be set at this point?

I guess we know this will be processed last because it's the source node, but it might make sense to assert some of the invariants you're assuming to begin with (e.g. whereInfo != null)

ksqldb-engine/src/main/java/io/confluent/ksql/physical/pull/operators/SelectOperator.java
```diff
@@ -0,0 +1,57 @@
...
+  }
+
+  @Override
+  public void open() {
+    child.open();
+  }
+
+  @Override
+  public Object next() {
+    return child.next();
```
AlanConfluent, line 37: I assume this doesn't currently do anything because this is done at the rocksdb layer?

ksqldb-engine/src/main/java/io/confluent/ksql/physical/pull/operators/AbstractPhysicalOperator.java
```diff
@@ -0,0 +1,29 @@
...
+ * specific language governing permissions and limitations under the License.
+ */
+
+package io.confluent.ksql.physical.pull.operators;
+
+public abstract class AbstractPhysicalOperator {
+
+  public abstract void open();
+
+  // Scan returns TableRow, Project returns List<List<?>>
```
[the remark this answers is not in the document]
vpapavas: Unfortunately, I haven't found a better way since two different operators need to return each `TableRow` and the other `List<List<?>>`

ksqldb-engine/src/main/java/io/confluent/ksql/engine/EngineExecutor.java
vpapavas: Yes. The logical plan nodes are there but they all have AST expressions and miss the analysis that is done for instance in the `PhysicalPlanBuilder`. A lot of that code in there should be moved to the logical plan

ksqldb-engine/src/main/java/io/confluent/ksql/engine/EngineExecutor.java
vpapavas: The analyzer performs some initial validation and categorization and is a different step happening before the logical plan. The logical planner exposes as well parser expressions but inside logical nodes. That's what I want to change. So that at the physical planner we don't deal with any AST expressions anymore, only logical entities.

ksqldb-engine/src/main/java/io/confluent/ksql/physical/pull/PullPhysicalPlanBuilder.java
vpapavas: I can add some sanity checks

ksqldb-engine/src/main/java/io/confluent/ksql/physical/pull/operators/SelectOperator.java
vpapavas: This doesn't do anything because currently we only support single key equality and nothing else in the WHERE clause. When we make it more general and allow others stuff as well, (like functions) we will need to add logic here.

ksqldb-engine/src/main/java/io/confluent/ksql/engine/EngineExecutor.java
vpapavas: The analyzer is a different step from the logical plan.

guozhangwang: I thought the plan is to have a single code path for logical plans across all queries. But this new function is only for pull queries, whereas for other queries we are still using `QueryEngine#buildQueryLogicalPlan`.

ksqldb-engine/src/main/java/io/confluent/ksql/engine/EngineExecutor.java
guozhangwang: +1

ksqldb-engine/src/main/java/io/confluent/ksql/engine/KsqlEngine.java
```diff
@@ -246,6 +250,30 @@ public TransientQueryMetadata executeQuery(
     }
   }
 
+  @Override
+  public PullQueryResult executePullQuery(
```
guozhangwang, line 254: Another question for my own education: what's the difference between this class and `QueryEngine`? It seems the latter is only for persistent and transient push queries?

ksqldb-engine/src/main/java/io/confluent/ksql/physical/pull/operators/AbstractPhysicalOperator.java
```diff
@@ -0,0 +1,29 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OF ANY KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations under the License.
+ */
+
+package io.confluent.ksql.physical.pull.operators;
+
+public abstract class AbstractPhysicalOperator {
```
guozhangwang, line 18: Is this going to be used beyond pull queries in the future?

guozhangwang (commented):
Made a pass over the PR and also related code base. Honestly I had a hard time trying to understand the code base and make meaning reviews since the class hierarchy / function call trace are different. I'd suggest we try to improve the code readability and cleanness first in order to spread the knowledge and reviewer base before large refactoring. A few things on top of my head:

* There are a bunch of functions that have very similar names like `execute`, `executeQuery`, `plan`, `planQuery` etc, some of them shares some callees while some aren't. We should add a one-liner javadoc on each of these, and also consider renaming the function names for better distinguishment.

* I think some of those call paths are also duplicated, and potentially can be consolidated towards our end goal: the logical plan is shared among all queries, and for each query type (persistent, transient, pull) we would have a separate physical plan generator and set of physical operators.

* Class names are probably define with only persistent query in mind originally, and now with more query types their names are no longer feasible, such like `QueryMetadata`, `KsqlPlan` etc seems only for a small subset of those queries. We should consider renaming them (since they are not public APIs anyways).

ksqldb-engine/src/main/java/io/confluent/ksql/engine/EngineExecutor.java
vpapavas: Yes but they are doing the same thing with respect to the logical planner i.e. calling the same methods to build the logical plan. The difference is in the Analyzer where for pull queries there is something different done, that's why I had to create a new method. The analyzer part will be cleaned up in a later PR once we get more clarity about what we need from push queries as well.

ksqldb-engine/src/main/java/io/confluent/ksql/physical/pull/operators/AbstractPhysicalOperator.java
vpapavas: Yes, for push queries if the use the same physical plan builder

ksqldb-engine/src/main/java/io/confluent/ksql/KsqlExecutionContext.java
```diff
@@ -114,6 +118,15 @@ TransientQueryMetadata executeQuery(
       ConfiguredStatement<Query> statement
   );
 
+  PullQueryResult executePullQuery(
```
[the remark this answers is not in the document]
vpapavas: Yeah, I don't know what to tell you :)

vpapavas:
@guozhangwang There definitely needs to be some cleanup done in the query processor but that should be part of a different refactoring PR once we are done with all our new functionality (push and pull plans). First, we are not yet knowledgeable of this specific code base, second a lot of code is historic that remained and never got refactored.  You ask good questions but I don't know why things are the way they are (yet). Third, I would like to minimize the scope of the PRs for pull queries to make sure that if there is a regression or error, it is easy to know where to look.

ksqldb-engine/src/main/java/io/confluent/ksql/engine/EngineExecutor.java
```diff
@@ -118,6 +139,78 @@ ExecuteResult execute(final KsqlPlan plan) {
...
+    final RateLimiter rateLimiter = RateLimiter.create(sessionConfig.getConfig(true).getInt(
+        KsqlConfig.KSQL_QUERY_PULL_MAX_QPS_CONFIG));
+    // If internal listeners are in use, we require the request to come from that listener to
+    // treat it as having been forwarded.
+    final boolean isAlreadyForwarded = routingOptions.skipForwardRequest()
+        // Trust the forward request option if isInternalRequest isn't available.
+        && isInternalRequest.orElse(true);
+
+    // Only check the rate limit at the forwarding host
+    if (!isAlreadyForwarded) {
```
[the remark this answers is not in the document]
vpapavas: You are right, I fixed it to be created only once.

ksqldb-engine/src/main/java/io/confluent/ksql/physical/pull/PullPhysicalPlanBuilder.java
```diff
@@ -0,0 +1,345 @@
...
+      final PersistentQueryMetadata persistentQueryMetadata
+  ) {
+    if (whereInfo == null) {
+      throw new KsqlException("Pull queries must have a WHERE clause");
+    }
+    keys = whereInfo.getKeysBound().stream()
+        .map(keyBound -> asKeyStruct(keyBound, persistentQueryMetadata.getPhysicalSchema()))
+        .collect(ImmutableList.toImmutableList());
+
+    if (!whereInfo.getWindowBounds().isPresent()) {
```
AlanConfluent, line 221: `getWindowBounds` is always present if it's windowed, regardless of what the bounds on the window are, right?  This is effectively an isWindowed check.

ksqldb-engine/src/main/java/io/confluent/ksql/physical/pull/operators/ProjectOperator.java
```diff
@@ -0,0 +1,231 @@
...
+        final Struct key = row.key();
+        final GenericRow value = row.value();
+
+        final List<Object> keyFields = key.schema().fields().stream()
+            .map(key::get)
+            .collect(Collectors.toList());
+
+        value.ensureAdditionalCapacity(
+            1 // ROWTIME
+                + keyFields.size()
```
AlanConfluent, line 119: This wrapping looks a little funny

ksqldb-engine/src/main/java/io/confluent/ksql/physical/pull/operators/ProjectOperator.java
```diff
@@ -0,0 +1,231 @@
...
+        .getProcessingLogContext()
+        .getLoggerFactory()
+        .getLogger(
+            QueryLoggerUtil.queryLoggerName(
+                QueryType.PULL_QUERY, contextStacker.push("PROJECT").getQueryContext())
+        );
+
+    transformer = select.getTransformer(logger);
+
+    return;
```
AlanConfluent, line 152: Unnecessary return?

ksqldb-engine/src/main/java/io/confluent/ksql/physical/pull/operators/ProjectOperator.java
```diff
@@ -0,0 +1,231 @@
...
+    final GenericRow intermediate = preSelectTransform.apply(row);
+
+    final GenericRow mapped = transformer.transform(
+        row.key(),
+        intermediate,
+        new PullProcessingContext(row.rowTime())
+    );
+    validateProjection(mapped, outputSchema);
+
+    // return List<?>
```
AlanConfluent, line 174: Remove comment?

ksqldb-engine/src/main/java/io/confluent/ksql/physical/pull/operators/ProjectOperator.java
```diff
@@ -0,0 +1,231 @@
...
+
+    rowList.addAll(row.value().values());
+
+    return rowList;
+  }
+
+  private Stream<?> keyFields(final Struct key) {
+    return key.schema().fields().stream().map(key::get);
+  }
+
```
AlanConfluent, line 230: Extra line

ksqldb-engine/src/main/java/io/confluent/ksql/physical/pull/operators/KeyedTableLookupOperator.java
```diff
@@ -0,0 +1,102 @@
...
+  }
+
+  @Override
+  public void open() {
+    final List<Row> result = new ArrayList<>();
+    for (KsqlPartitionLocation location : partitionLocations) {
+      if (!location.getKeys().isPresent()) {
+        throw new IllegalStateException("Table lookup queries should be done with keys");
+      }
+      for (Struct key : location.getKeys().get()) {
```
AlanConfluent, line 58: It's slightly funny to do the real work in open and then just dish out the cached rows on each next.  Since the interface is row oriented, we could also keep track of the key we're on and then do the actual single key lookup on each call to next.  In theory, this allows for "back pressure" and avoiding a burst a work since we could do it incrementally and on demand.

ksqldb-engine/src/main/java/io/confluent/ksql/physical/pull/operators/KeyedTableLookupOperator.java
```diff
@@ -0,0 +1,102 @@
...
+    return null;
+  }
+
+  @Override
+  public void close() {
+
+  }
+
+  @Override
+  public void addChild(final AbstractPhysicalOperator child) {
```
AlanConfluent, line 84: Throw an exception since this should be a leaf?

ksqldb-engine/src/main/java/io/confluent/ksql/physical/pull/operators/KeyedWindowedTableLookupOperator.java
```diff
@@ -0,0 +1,105 @@
...
+  }
+
+  @Override
+  public void open() {
+    final List<WindowedRow> result = new ArrayList<>();
+    for (KsqlPartitionLocation location : partitionLocations) {
+      if (!location.getKeys().isPresent()) {
+        throw new IllegalStateException("Window queries should be done with keys");
+      }
+      for (Struct key : location.getKeys().get()) {
```
AlanConfluent, line 64: Similar question as for non windowed.

ksqldb-engine/src/main/java/io/confluent/ksql/physical/pull/operators/KeyedWindowedTableLookupOperator.java
```diff
@@ -0,0 +1,105 @@
...
+    return null;
+  }
+
+  @Override
+  public void close() {
+
+  }
+
+  @Override
+  public void addChild(final AbstractPhysicalOperator child) {
```
AlanConfluent, line 87: Throw an exception if this is meant to be a leaf?

ksqldb-engine/src/main/java/io/confluent/ksql/internal/PullQueryExecutorMetrics.java
```diff
@@ -108,11 +108,11 @@ public void recordResponseSize(final double value) {
     this.responseSizeSensor.record(value);
   }
 
-  List<Sensor> getSensors() {
+  public List<Sensor> getSensors() {
```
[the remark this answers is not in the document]
vpapavas: No, they actually need to be public.

ksqldb-engine/src/main/java/io/confluent/ksql/physical/pull/operators/KeyedTableLookupOperator.java
vpapavas: There are multiple locations and keys, so we need to keep track of both in next and exhaust both iterators. I made the change, lmk how it looks

ksqldb-engine/src/main/java/io/confluent/ksql/physical/pull/PullPhysicalPlanBuilder.java
vpapavas: If there are no window-bounds in the query, the query is not windowed. So, yes this is a `isWindowed` check

ksqldb-engine/src/main/java/io/confluent/ksql/physical/pull/PullPhysicalPlanBuilder.java
```diff
@@ -0,0 +1,345 @@
...
+      }
+
+      if (prevPhysicalOp == null) {
+        rootPhysicalOp = currentPhysicalOp;
+      } else {
+        prevPhysicalOp.addChild(currentPhysicalOp);
+      }
+      prevPhysicalOp = currentPhysicalOp;
+      // For now assume only single source which is the case for pull queries
+      if (currentLogicalNode.getSources().isEmpty()) {
```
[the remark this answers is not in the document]
vpapavas: No, it's not an error. This is just to check that we have reached the leaf of the LogicalPlan and we can stop the translation to physical

ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/resources/streaming/StreamedQueryResource.java
```diff
@@ -153,6 +163,7 @@ public void configure(final KsqlConfig config) {
     }
 
     ksqlConfig = config;
+    rateLimiter = RateLimiter.create(ksqlConfig.getInt(KsqlConfig.KSQL_QUERY_PULL_MAX_QPS_CONFIG));
```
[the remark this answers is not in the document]
vpapavas: If I put in the `EngineExecutor` it will be per pull query (that's where I had it initially). I moved it to the `KsqlRestApplication`

ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/resources/streaming/StreamedQueryResource.java
```diff
@@ -266,9 +277,24 @@ private EndpointResponse handlePullQuery(
...
-    final PullQueryResult result = pullQueryExecutor.execute(
+    final PullQueryResult result = ksqlEngine.executePullQuery(
+        serviceContext,
+        routingFilterFactory,
+        configured,
+        requestProperties,
+        isInternalRequest,
+        pullQueryMetrics,
+        rateLimiter);
+    final TableRows tableRows = new TableRows(
```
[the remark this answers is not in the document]
vpapavas: The `PullQueryResult` of the pull physical plan is a different class than the `PullQueryResult` of `PullQueryExecutor`. The first doesn't have visibility to `TableRows` that's why I have to create them in `StreamedQueryResource`

ksqldb-engine/src/main/java/io/confluent/ksql/physical/pull/operators/WhereInfo.java
```diff
@@ -0,0 +1,542 @@
...
+  public List<Object> getKeysBound() {
+    return keysBound;
+  }
+
+  private static List<Object> extractKeysFromInPredicate(
+      final List<InPredicate> inPredicates,
+      final boolean windowed,
+      final LogicalSchema schema
+  ) {
+    final InPredicate inPredicate = Iterables.getLast(inPredicates);
```
agavra, line 136: why are we using `getLast` here? did we intend to use `getOnlyElement`?

ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/resources/streaming/PullQueryPublisher.java
```diff
@@ -42,32 +45,50 @@
...
+    this.routingFilterFactory = requireNonNull(routingFilterFactory, "routingFilterFactory");
+    this.rateLimiter = requireNonNull(rateLimiter, "rateLimiter");
   }
 
   @Override
   public synchronized void subscribe(final Subscriber<Collection<StreamedRow>> subscriber) {
     final PullQuerySubscription subscription = new PullQuerySubscription(
         subscriber,
         () -> {
+          ksqlEngine.executePullQuery(
```
agavra, line 83: uh - shouldn't we be using the result of this? otherwise we're just executing the pull query twice, right? once here and once on line 92?

also now that this refactor is done, can we get rid of `PullQueryExecutor`?

ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/resources/streaming/StreamedQueryResource.java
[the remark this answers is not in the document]
agavra: @vpapavas did you mean to push a more recent udpate? some of your comments, like this one, don't seem to be resolved

agavra (commented):
Thanks @vpapavas - I think this structure will let us iterate much faster and with more clarity than the previous code. Most of my comments inline are code-level comments/notes. 

I do have one serious concern though: We're missing unit tests for all of the new classes added in this PR! Either (1) we should add unit tests or (2) these classes are too difficult to unit test, which IMO is a sign of a bad design and we should take a deeper look about how we can simplify the design to support better unit testing.

Now that we have the "big picture". I think it makes sense to split this PR into smaller ones, where each one introduces a single part (e.g. one PhysicalOperator, the HARouting, the WhereClauseInfo) along with unit tests. Then the PR that introduces the PhysicalPlanBuilder can be really small and just tie all of them together.

ksqldb-engine/src/main/java/io/confluent/ksql/physical/pull/operators/KeyedTableLookupOperator.java
vpapavas: I first commented and then waited for the tests to pass before pushing. Unfortunately, it took longer than expected :(

ksqldb-engine/src/main/java/io/confluent/ksql/KsqlExecutionContext.java
```diff
@@ -125,6 +129,16 @@ TransientQueryMetadata executeQuery(
...
+      ServiceContext serviceContext,
+      RoutingFilterFactory routingFilterFactory,
+      ConfiguredStatement<Query> statement,
+      Map<String, Object> requestProperties,
+      Optional<Boolean> isInternalRequest,
+      Optional<PullQueryExecutorMetrics> pullQueryMetrics,
+      RateLimiter rateLimiter
```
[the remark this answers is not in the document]
vpapavas: I am not sure I am following what you mean.
The `rateLimiter` (after my last changes) is created once in the `KsqlRestApplication` and then used in `StreamedQueryResource` and `WsQueryEndpoint`.  After that, the logic of how it is applied hasn't changed from the `PullQueryExecutor`: it is used for every pull query. The same `rateLimiter` is being used for every pull query since it is created once.
I don't understand what you are saying that the design isn't safe. Alan's comment was about the `StreamedQueryResource` and `WsQueryEndpoint` having a different `rateLimiter` whereas we wanted to limit on their combined pull queries.

vpapavas:
Thank you @agavra  for your comments! I agree that we need unit tests and will add for the physical operators and physical plan builder. All the rest of the code has been refactored (without changes) from the `PullQueryExecutor`.  And though I agree we want to make the code better with every iteration, some of it will change in the next weeks so it seems throw-away work. 

The `HARouting` has been extensively tested in `PullQueryRoutingFuntionalTests`, also there is no new functionality from what the routing was doing in `PullQueryExecutor`. @AlanConfluent do you feel there are unit tests missing?

The `WhereInfo` is a temporary class which will be replaced by the logical plan work so it will be throw away work to implement unit tests for code that will change in the next 2-3 weeks.

ksqldb-engine/src/main/java/io/confluent/ksql/KsqlExecutionContext.java
```diff
@@ -125,6 +129,16 @@ TransientQueryMetadata executeQuery(
       ConfiguredStatement<Query> statement
   );
 
+  PullQueryResult executePullQuery(
+      ServiceContext serviceContext,
+      RoutingFilterFactory routingFilterFactory,
+      ConfiguredStatement<Query> statement,
+      Map<String, Object> requestProperties,
```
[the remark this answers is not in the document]
vpapavas: The `SessionConfig` contains the overrides not the request properties, right?

ksqldb-engine/src/main/java/io/confluent/ksql/KsqlExecutionContext.java
```diff
@@ -125,6 +129,16 @@ TransientQueryMetadata executeQuery(
...
+      Optional<Boolean> isInternalRequest,
```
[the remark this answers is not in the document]
vpapavas: I am not sure I understand why it is clunky but @AlanConfluent wrote this code: Alan do you have any background information here of why this needs to be `Optional`? Can it be replaced with boolean? I see in `ServerVerticle:isInternalRequest` that it could just as easily return boolean (true if is internal, false otherwise), right?
```
  private static Optional<Boolean> isInternalRequest(final RoutingContext routingContext) {
    return Optional.ofNullable(routingContext.get(CONTEXT_DATA_IS_INTERNAL));
  }
```

ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/resources/streaming/PullQueryPublisher.java
vpapavas: Yes, I still need to remove all the `PullQueryExecutor` code. I am leaving it for last because it helps with debugging to verify what the correct behavior should be.

ksqldb-engine/src/main/java/io/confluent/ksql/physical/pull/operators/ProjectOperator.java
```diff
@@ -0,0 +1,227 @@
...
+      };
+    }
+
+    final SelectValueMapper<Object> select = SelectValueMapperFactory.create(
+        logicalNode.getSelectExpressions(),
+        intermediateSchema,
+        ksqlConfig,
+        metaStore
+    );
+
```
vpapavas, line 141: Does anyone know what the code below is doing and why it is needed?

ksqldb-engine/src/main/java/io/confluent/ksql/engine/KsqlEngine.java
```diff
@@ -250,6 +254,32 @@ public TransientQueryMetadata executeQuery(
...
+      final RateLimiter rateLimiter
+  ) {
+    return EngineExecutor
+        .create(
+            primaryContext,
+            serviceContext,
+            statement.getSessionConfig()
+        )
+        .executePullQuery(
+            this,
```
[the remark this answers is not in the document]
vpapavas: Refactored to not pass `this`

ksqldb-engine/src/main/java/io/confluent/ksql/engine/PullQueryConfigRoutingOptions.java
```diff
@@ -0,0 +1,95 @@
...
+  }
+
+  private long getLong(final String key) {
+    if (configOverrides.containsKey(key)) {
+      return (Long) configOverrides.get(key);
+    }
+    return ksqlConfig.getLong(key);
+  }
+
+  private boolean getForwardedFlag(final String key) {
```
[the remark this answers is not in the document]
vpapavas: Fixed

ksqldb-engine/src/main/java/io/confluent/ksql/physical/pull/PullPhysicalPlanBuilder.java
```diff
@@ -0,0 +1,345 @@
...
+  /**
+   * Visits the logical plan top-down to build the physical plan.
+   * @param logicalPlanNode the logical plan root node
+   * @return the root node of the tree of physical operators
+   */
+  public PullPhysicalPlan buildPullPhysicalPlan(final LogicalPlanNode logicalPlanNode) {
+    DataSourceOperator dataSourceOperator = null;
+    persistentQueryMetadata = findMaterializingQuery(executionContext, analysis);
+    queryId = uniqueQueryId();
+    mat = persistentQueryMetadata
```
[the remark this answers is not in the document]
vpapavas: Fixed

ksqldb-engine/src/main/java/io/confluent/ksql/physical/pull/PullPhysicalPlanBuilder.java
```diff
@@ -0,0 +1,345 @@
...
+        mat,
+        dataSourceOperator);
+  }
+
+  private ProjectOperator translateProjectNode(final ProjectNode logicalNode) {
+    LogicalSchema outputSchema = null;
+    boolean isStar = false;
+    if (isSelectStar(statement.getStatement().getSelect())) {
+      isStar = true;
+      outputSchema = buildSchema(mat.schema(), mat.windowType().isPresent());
```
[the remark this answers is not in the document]
vpapavas: Yes! All this code is going to move to the logical plan with the refactoring I will do after this PR is merged

ksqldb-engine/src/main/java/io/confluent/ksql/physical/pull/PullPhysicalPlanBuilder.java
```diff
@@ -0,0 +1,345 @@
...
+        rootPhysicalOp,
+        ((ProjectOperator)rootPhysicalOp).getOutputSchema(),
+        queryId,
+        keys,
+        mat,
+        dataSourceOperator);
+  }
+
+  private ProjectOperator translateProjectNode(final ProjectNode logicalNode) {
+    LogicalSchema outputSchema = null;
```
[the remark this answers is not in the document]
vpapavas: It can't be final

ksqldb-engine/src/main/java/io/confluent/ksql/physical/pull/operators/HARouting.java
```diff
@@ -0,0 +1,318 @@
...
+    }
+
+    return rows;
+  }
+
+  private void validateSchemas(final List<LogicalSchema> schemas) {
+    final LogicalSchema schema = Iterables.getLast(schemas);
+    for (LogicalSchema s : schemas) {
+      if (!schema.equals(s)) {
+        throw new KsqlException("Schemas from different hosts should be identical");
```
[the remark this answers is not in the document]
vpapavas: Tracked in #6571

ksqldb-engine/src/main/java/io/confluent/ksql/physical/pull/PullPhysicalPlanBuilder.java
```diff
@@ -0,0 +1,348 @@
...
+    }
+
+    final QueryId queryId = new QueryId(Iterables.get(queries, 0));
+
+    final PersistentQueryMetadata query = executionContext
+        .getPersistentQuery(queryId)
+        .orElseThrow(() -> new KsqlException("Materializing query has been stopped"));
+
+    if (query.getDataSourceType() != DataSourceType.KTABLE) {
+      throw new KsqlException("Pull queries are not supported on streams.");
```
guozhangwang, line 258: Could we also include the actual `getDataSourceType` as well.

Also, for other thrown exceptions, I find that we tend to not include some key information into the exception message that may make debugging harder. I'd suggest as a rule of thumb to always include all variable values if they participate in the condition that leads to exception.

guozhangwang (commented):
Made another pass over the code, and here are some comments that do not necessarily required to be addressed for this PR, but are kind of meta thoughts about the current code base:

1) It was not clear to me why have both a `KsqlEngine` and an `EngineExecutor`, especially when many of the method calls in the former are directly triggering the same named methods of the latter. I realized the differences only later, and hence would suggest we make the distinction more clear, as:

a) `KsqlEngine` takes in the query statement String, and do all the query analyzing / query compilation, the output of it should be the physical query plan.

b) `EngineExecutor` takes only physical query plan as the input and execute, and since we will have three different query plans maybe we can just have three executor classes as well.

2) I'd suggest we have a better class boundary for pull query routing-related logic: i.e. after the query is compiled, if it can be executed locally, then the execution happens in `EngineExecutor` that returns `PullQueryResult`; if it needs to be executed remotely, then send to the Routing object which ONLY does the forwarding, not execution. The forwarding is primarily for read/write on the socket.

guozhangwang:
> Thank you @agavra for your comments! I agree that we need unit tests and will add for the physical operators and physical plan builder. All the rest of the code has been refactored (without changes) from the `PullQueryExecutor`. And though I agree we want to make the code better with every iteration, some of it will change in the next weeks so it seems throw-away work.
> 
> The `HARouting` has been extensively tested in `PullQueryRoutingFuntionalTests`, also there is no new functionality from what the routing was doing in `PullQueryExecutor`. @AlanConfluent do you feel there are unit tests missing?
> 
> The `WhereInfo` is a temporary class which will be replaced by the logical plan work so it will be throw away work to implement unit tests for code that will change in the next 2-3 weeks.

I'd suggest we still have dedicated unit tests for the new classes, even if they are covered in the coarsen grained functional tests. Generally speaking we should have a one-to-one mapping of classes / testClasses, and the testClasses focus on the "contracts" of the public APIs, on edge input parameters etc.

ksqldb-engine/src/main/java/io/confluent/ksql/physical/pull/PullPhysicalPlan.java
```diff
@@ -0,0 +1,102 @@
...
+import io.confluent.ksql.physical.pull.operators.DataSourceOperator;
+import io.confluent.ksql.query.QueryId;
+import io.confluent.ksql.schema.ksql.LogicalSchema;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Objects;
+import java.util.Optional;
+import org.apache.kafka.connect.data.Struct;
+
+public class PullPhysicalPlan {
```
guozhangwang, line 31: Javadoc for this class regarding 1) what metadata it contains, 2) what assumption it makes (e.g.  what `root` operator should be`) would be highly appreciated :)

ksqldb-engine/src/main/java/io/confluent/ksql/physical/pull/PullPhysicalPlanBuilder.java
```diff
@@ -0,0 +1,348 @@
...
+  private final RoutingOptions routingOptions;
+  private final ConfiguredStatement<Query> statement;
+
+  private WhereInfo whereInfo;
+  private PersistentQueryMetadata persistentQueryMetadata;
+  private  QueryId queryId;
+  private Materialization mat;
+  private List<Struct> keys;
+
+  public PullPhysicalPlanBuilder(
```
guozhangwang, line 93: Javadoc for this class regarding the common pattern of the generated physical plan (e.g. should it always be a scan -> filter), what validation it does, what metadata (e.g. materialization) it relies on would be highly appreciated :)

ksqldb-engine/src/main/java/io/confluent/ksql/physical/pull/operators/AbstractPhysicalOperator.java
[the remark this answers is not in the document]
guozhangwang: BTW Why declare it as abstract class not interface??

ksqldb-engine/src/main/java/io/confluent/ksql/physical/pull/operators/WhereInfo.java
```diff
@@ -0,0 +1,542 @@
...
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Map;
+import java.util.Map.Entry;
+import java.util.Objects;
+import java.util.Optional;
+import java.util.Set;
+import java.util.stream.Collectors;
+
+public final class WhereInfo {
```
guozhangwang, line 58: Javadocs regarding which metadata it wraps in is appreciated.

ksqldb-engine/src/main/java/io/confluent/ksql/KsqlExecutionContext.java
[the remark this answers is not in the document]
agavra: ah yes, you are right.

ksqldb-engine/src/main/java/io/confluent/ksql/KsqlExecutionContext.java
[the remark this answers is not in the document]
AlanConfluent: In some areas of code, it was necessary to see whether the internal interface was enabled at all.  Effectively, it's a 3 state flag: Internal interface not in use at all, internal interface in use and this request is internal, and internal interface in use and this request is not internal.  If you look at how it's being used in the code it's this:

```
final boolean isAlreadyForwarded = routingOptions.skipForwardRequest()
          // Trust the forward request option if isInternalRequest isn't available.
          && isInternalRequest.orElse(true);
```

So, if the interface isn't in use, the value defaults to true, not false.  If you want to do this at the calling layer, that's fine.  Also, if this three state thing isn's clear, it could be wrapped in a class that makes it clearer.

AlanConfluent:
> The `HARouting` has been extensively tested in `PullQueryRoutingFuntionalTests`, also there is no new functionality from what the routing was doing in `PullQueryExecutor`. @AlanConfluent do you feel there are unit tests missing?
> 
> The `WhereInfo` is a temporary class which will be replaced by the logical plan work so it will be throw away work to implement unit tests for code that will change in the next 2-3 weeks.

We discussed this offline a bit.  I don't think you should waste time testing temporary classes if the future testing will look every different.

I agree that `PullQueryRoutingFuntionalTests` does do a good job of testing much of `HARouting`.  It would be nice to actually introduce some unit testing if possible though.  I actually did this in `PullQueryExecutorTest` to test falling back on standbys.  Presumably that should go in `HARoutingTest` now.  That might actually cover a lot of `HARouting` already.

ksqldb-engine/src/main/java/io/confluent/ksql/physical/pull/operators/KeyedWindowedTableLookupOperator.java
```diff
@@ -0,0 +1,137 @@
...
+      keyIterator = nextLocation.getKeys().get().iterator();
+    }
+    nextKey = keyIterator.next();
+    resultIterator = mat.windowed().get(
+        nextKey,
+        nextLocation.getPartition(),
+        windowBounds.getStart(),
+        windowBounds.getEnd())
+        .iterator();
+    if (resultIterator.hasNext()) {
```
AlanConfluent, line 105: What if `resultIterator.hasNext()` is false, but `keyIterator.hasNext()` is true?  You would return null below when there might be more.  It seems like you could recursively call `next()` after setting up the `nextKey` and `resultIterator`.

ksqldb-engine/src/main/java/io/confluent/ksql/physical/pull/operators/KeyedTableLookupOperator.java
```diff
@@ -0,0 +1,128 @@
...
+        throw new IllegalStateException("Table lookup queries should be done with keys");
+      }
+      keyIterator = nextLocation.getKeys().get().iterator();
+    }
+    nextKey = keyIterator.next();
+    resultIterator = mat.nonWindowed()
+        .get(nextKey, nextLocation.getPartition())
+        .map(ImmutableList::of)
+        .orElse(ImmutableList.of()).iterator();
+    if (resultIterator.hasNext()) {
```
AlanConfluent, line 96: This has the same issue as in `KeyedWindowedTableLookupOperator`

ksqldb-engine/src/main/java/io/confluent/ksql/physical/pull/operators/ProjectOperator.java
AlanConfluent: I think because `SelectValueMapper` is shared with persistent queries and it expects this so it can write errors to the processing topic.  I'm not sure if we really would want that for pull queries or if it's just shoehorned in here because the code is shared.

ksqldb-engine/src/main/java/io/confluent/ksql/KsqlExecutionContext.java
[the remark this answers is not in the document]
agavra: that makes sense @AlanConfluent  thanks for the clarification, I think it would be nice to wrap it in an enum representing these three states (or in a class like you suggest) but this can be done in a follow-up (or now, if you're eager!)

ksqldb-engine/src/main/java/io/confluent/ksql/physical/pull/operators/HARouting.java
```diff
@@ -0,0 +1,318 @@
...
+        routingOptions.isDebugRequest()
+            ? Collections.nCopies(rows.size(), node) : null);
+    return new PullQueryResult(
+        rows,
+        debugNodes,
+        outputSchema,
+        queryId);
+  }
+
+  private static List<List<?>> forwardTo(
```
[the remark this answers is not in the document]
AlanConfluent: Currently, we do multiple requests in parallel using the executorService.  We should maybe make this logic use async rather than a threadpool and futures, but it accomplishes the same thing at the moment.

ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/execution/PullQueryExecutor.java
```diff
@@ -65,6 +65,7 @@
 import io.confluent.ksql.execution.transform.select.SelectValueMapper;
```
AlanConfluent, line 65: Do you want to delete this file and it's test file?  (Of couse, it would be great to migrate all of those tests to their updated places).

ksqldb-engine/src/main/java/io/confluent/ksql/physical/pull/PullPhysicalPlanBuilder.java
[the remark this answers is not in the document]
AlanConfluent: I see.  Will the logic `while (currentLogicalNode.getSources() != null) {` ever be false?  Walking through the code, it looks like it wouldn't be and this is the main exit from the loop.  If so, maybe it makes sense to either make it `while (true)` so it's obvious this is the way the iteration ends.

ksqldb-engine/src/main/java/io/confluent/ksql/physical/pull/operators/HARouting.java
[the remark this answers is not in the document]
guozhangwang: Yeah I agree the current code achieves the same thing, but using a threadpool of 100 threads by default could be costly than using a single thread (or at most, we only need one thread per destination socket). Anyways, as I mentioned in the meta comment I feel this is not necessarily the scope of the PR and we can just refactor it later if people agree to the approach.

guozhangwang:
> We discussed this offline a bit. I don't think you should waste time testing temporary classes if the future testing will look every different.

Ah okay, if the HARouting class is soon going to be replaced or refactored then I agree with do not need to gear up testing coverage. I'd just like to make sure that, at the release cut, for the classes that we have in the repo, each of them would have a corresponding unit test covering its public API contracts :)

guozhangwang:
Just to clarify on my previous comments regarding unit test coverage and javadocs: I think that over time I'd like to push our bar higher on those aspects for better readability and debuggbility, but I did not intend to push all of them on this single PR :) So please don't feel pressured @vpapavas it's not really meant to get to your shoulder at all. It would be a long procedure with series of small improvements moving forward.

As for this PR I think we can merge it once the rebasing is done since Vicky has confirmed from tests that all behaviors stays at the same.

ksqldb-engine/src/main/java/io/confluent/ksql/KsqlExecutionContext.java
```diff
@@ -125,6 +129,16 @@ TransientQueryMetadata executeQuery(
       ConfiguredStatement<Query> statement
   );
 
+  PullQueryResult executePullQuery(
```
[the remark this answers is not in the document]
vpapavas: Yes, I thought about this as well. It's a tradeoff between sending larger messages (the physical plan vs the query) and computation. If we do want to make this change, this would be part of a larger refactoring, and not part of this PR. 

The routing already works as you describe, in the sense that server B executes the request locally. If it fails, server A will try another server.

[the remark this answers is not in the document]
guozhangwang: Yeah this is totally a meta thing which is not to be addressed in this PR. I left it here just to dump my wild thoughts here for future references.

Regarding the trade-off, I think with some encoding the number of bytes that we need to transmit between servers with the physical plan would not be orders of magnitude larger (my bold claim, of course :P) but the benefit that only one single server makes the compilation decision to achieve consistent behavior could be huge.

ksqldb-engine/src/main/java/io/confluent/ksql/KsqlExecutionContext.java
```diff
@@ -125,6 +129,16 @@ TransientQueryMetadata executeQuery(
...
+      ServiceContext serviceContext,
+      RoutingFilterFactory routingFilterFactory,
+      ConfiguredStatement<Query> statement,
+      Map<String, Object> requestProperties,
+      Optional<Boolean> isInternalRequest,
+      Optional<PullQueryExecutorMetrics> pullQueryMetrics,
```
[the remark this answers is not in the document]
vpapavas: `pullQueryMetrics` is actually used in many classes and not only `KsqlEngine`. The most important one is `OldApiUtils` where we measures the size of the request/response and the only access to this is through the `KsqlRestApplication`. The same for the other endpoints where we want to measure latency. The only access point is `KsqlRestApplication`.

ksqldb-engine/src/main/java/io/confluent/ksql/physical/pull/PullPhysicalPlanBuilder.java
```diff
@@ -0,0 +1,348 @@
...
+      // For now assume only single source which is the case for pull queries
+      currentLogicalNode = currentLogicalNode.getSources().get(0);
+    }
+
+    if (dataSourceOperator == null) {
+      throw new IllegalStateException("DataSourceOperator cannot be null in Pull physical plan");
+    }
+    return new PullPhysicalPlan(
+        rootPhysicalOp,
+        ((ProjectOperator)rootPhysicalOp).getOutputSchema(),
```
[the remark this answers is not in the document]
vpapavas: I am not following here. I need a pointer to the root operator of the physical plan.

ksqldb-engine/src/main/java/io/confluent/ksql/physical/pull/PullPhysicalPlan.java
```diff
@@ -0,0 +1,102 @@
...
+
+  private Object next() {
+    return root.next();
+  }
+
+  private void close() {
+    root.close();
+  }
+
+  public Materialization getMaterialization() {
```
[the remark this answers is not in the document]
vpapavas: I don't understand how figuring out the hosts/partitions can be done at compile time. First we compile the query into a physical plan and then we try to route it. Depending on which node is down, the location changes at runtime.

ksqldb-engine/src/main/java/io/confluent/ksql/physical/pull/PullPhysicalPlanBuilder.java
vpapavas: The metadata/validation information is going to change with the next PR so I'd rather add it when the code is done. I don't want to end up with wrong javadocs.

ksqldb-engine/src/main/java/io/confluent/ksql/physical/pull/operators/KeyedWindowedTableLookupOperator.java
vpapavas: `resultIterator.hasNext()` is false, but `keyIterator.hasNext()` is true it will go to line 91 and get the next keyIterator `keyIterator.next()`. Am I missing something?

ksqldb-engine/src/main/java/io/confluent/ksql/engine/EngineExecutor.java
```diff
@@ -116,6 +137,82 @@ ExecuteResult execute(final KsqlPlan plan) {
...
+    }
+
+    final RoutingOptions routingOptions = new PullQueryConfigRoutingOptions(
+        sessionConfig.getConfig(false),
+        statement.getSessionConfig().getOverrides(),
+        requestProperties
+    );
+
+    // If internal listeners are in use, we require the request to come from that listener to
+    // treat it as having been forwarded.
```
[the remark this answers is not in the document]
vpapavas: It true that we forward only once but this variable is not used to achieve this. It only checks if the request is a forwarded one or not to apply the rate limiting. So, isAlreadyForwarded == don't apply rate limiting

ksqldb-engine/src/main/java/io/confluent/ksql/KsqlExecutionContext.java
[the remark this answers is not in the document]
vpapavas: I don't think it can be applied at the `KsqlRestApplication` as we don't know there whether a request is a pull query or not.

ksqldb-engine/src/main/java/io/confluent/ksql/physical/pull/operators/KeyedWindowedTableLookupOperator.java
AlanConfluent: If you were to call `this.next()` again, it would do what you're saying and see that there are more keys, but from what I can see, it doesn't do that by default (and hence my recursive comment).

Specifically, imagine you're looking up two keys, `1` and `2`.  If you do `resultIterator = mat.windowed().get(` .. for `1` and get nothing, `resultIterator.hasNext()` will be false, and you'll fall through to `return null` below. 

I guess maybe that's the correct if you want to return nulls for empty lookups, but then you have to do a null check with the caller.  I was effectively imagining that you would just continue on to the next non empty iterator.

ksqldb-engine/src/main/java/io/confluent/ksql/physical/pull/PullPhysicalPlan.java
```diff
@@ -0,0 +1,109 @@
...
+      final List<KsqlPartitionLocation> locations) {
+
+    // We only know at runtime which partitions to get from which node.
+    // That's why we need to set this explicitly for the dataSource operators
+    dataSourceOperator.setPartitionLocations(locations);
+
+    open();
+    final List<List<?>> localResult = new ArrayList<>();
+    List<?> row = null;
+    while ((row = (List<?>)next()) != null) {
```
AlanConfluent, line 70: Here is where you do the null check.  My original thinking was mostly that if you cannot reference which key this is returning null for, there's not really any point in returning null at all.  Maybe you feel differently.

AlanConfluent (approved):

ksqldb-engine/src/main/java/io/confluent/ksql/physical/pull/PullPhysicalPlan.java
vpapavas: You were right, I wasn't handling correctly the case where a key does not exist in the sate store. Fixed it now

ksqldb-engine/src/test/java/io/confluent/ksql/physical/pull/operators/KeyedTableLookupOperatorTest.java
```diff
@@ -0,0 +1,130 @@
...
+  private Struct KEY4;
+  @Mock
+  private Row ROW1;
+  @Mock
+  private Row ROW3;
+  @Mock
+  private Row ROW4;
+
+
+  @Before
```
agavra, line 77: `@Before` runs before every test, not once before all the tests. If you want these to be populated once, which I think you do, it should be in the constructor

ksqldb-engine/src/test/java/io/confluent/ksql/physical/pull/operators/ProjectOperatorTest.java
```diff
@@ -0,0 +1,356 @@
...
+        A_ROWTIME
+    );
+    when(child.next()).thenReturn(row);
+    projectOperator.open();
+
+    // Then:
+    final List<Object> rowList = new ArrayList<>();
+    row.key().schema().fields().stream().map(row.key()::get).forEach(rowList::add);
+    rowList.addAll(row.value().values());
+    assertThat(projectOperator.next(), is(rowList));
```
agavra, line 123: I think it makes sense to add an assert that `selectValueMapper` was never called (`Mockito.verifyNoInteractions(selectValueMapper)`). This will (1) make sure we're not calling unnecessary code, and (2) makes the test easier to understand (I asked myself why we didn't need to mock out the mapper for this method)

ksqldb-engine/src/test/java/io/confluent/ksql/physical/pull/operators/ProjectOperatorTest.java
```diff
@@ -0,0 +1,356 @@
...
+        A_ROWTIME
+    );
+    when(child.next()).thenReturn(row);
+    when(selectValueMapperFactorySupplier.create(any(), any(), any(), any()))
+        .thenReturn(selectValueMapper);
+    when(selectValueMapper.getTransformer(logger)).thenReturn(transformer);
+    when(transformer.transform(any(), any(), any())).thenReturn(GenericRow.genericRow("k", "a", "b"));
+    when(mat.schema()).thenReturn(SCHEMA);
+    projectOperator.open();
+    projectOperator.next();
```
agavra, line 187: nit: this applies to all tests, but this is what belongs in the `//When:` part (it makes it easy to identify what is setup and what the test is actually doing. I know this PR has gone on for a while so I won't ask you to refactor all the tests, but going forward it makes tests really easy to read when then `//When` portion contains the code that you're testing.

For example:
```java
// When:
List<?> result = projectOperator.next();

// Then:
final List<Object> expected = new ArrayList<>();
windowedRow.key().schema().fields().stream().map(windowedRow.key()::get).forEach(expected::add);
expected.add(windowedRow.window().get().start().toEpochMilli());
expected.add(windowedRow.window().get().end().toEpochMilli());
expected.addAll(windowedRow.value().values());
assertThat(result, is(expected)); 
```

Compare that with:
```java
// Then:
final List<Object> expected = new ArrayList<>();
windowedRow.key().schema().fields().stream().map(windowedRow.key()::get).forEach(expected::add);
expected.add(windowedRow.window().get().start().toEpochMilli());
expected.add(windowedRow.window().get().end().toEpochMilli());
expected.addAll(windowedRow.value().values());
assertThat(result, is(rowList)); 
assertThat(projectOperator.next(), is(expected)); 
```

For the second one I need to dig in to figure out what is the production code being called.

agavra (approved):
LGTM! I briefly took a look at the new tests and trusted @AlanConfluent's review for everything else (plus what we've discussed I assume you've addressed it). Looking forward to the improved `WhereInfo` and `HARouting` code 😉

PR #6389, opened 2020-10-08
chore: rename schema to valueSchema in QTT tests

agavra:
### Description 

Simple change as listed in the description... took a long time till I figured out a good way to rewrite the historical plans. Follow up from #6363 (comment)

### Testing done 

Testing-only change

### Reviewer checklist

### DIFF
--- a/ksqldb-functional-tests/src/main/java/io/confluent/ksql/test/model/TopicNode.java
+++ b/ksqldb-functional-tests/src/main/java/io/confluent/ksql/test/model/TopicNode.java
@@ -39,26 +39,26 @@ public final class TopicNode {
 
   private final String name;
   private final JsonNode keySchema;
-  private final JsonNode schema;
+  private final JsonNode valueSchema;
   private final int numPartitions;
   private final int replicas;
   private final String keyFormat;
-  private final String format;
+  private final String valueFormat;
 
   public TopicNode(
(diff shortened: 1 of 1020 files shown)


vcrfxia (approved):
LGTM

> Simple change as listed in the description... took a long time till I figured out a good way to rewrite the historical plans.

Did you not use the PlannedTestRewriter?

agavra:
@vcrfxia - the `PlannedTestRewriter` couldn't rewrite them because my change meant it can't read the old specs. What I ended up doing was hacking the `TopicNode` to have a constructor that accepted both `valueSchema` and `schema` but in either way it only serialized `valueSchema`, then I hacked `PlannedTestRewriter` to only rewrite the specs, not the plans, because it was messing up with a ton of the plans as well. Then it worked 😂

vcrfxia:
> the PlannedTestRewriter couldn't rewrite them because my change meant it can't read the old specs. What I ended up doing was hacking the TopicNode to have a constructor that accepted both valueSchema and schema but in either way it only serialized valueSchema, then I hacked PlannedTestRewriter to only rewrite the specs, not the plans, because it was messing up with a ton of the plans as well. Then it worked 😂

Ah yeah, that's exactly the same thing I did last time I changed the format of the schemas node. Hopefully this use case is rare, else we should reconsider the way we handle rewrites.

PR #6397, opened 2020-10-09
fix: recovery hangs when using TERMINATE ALL

### LINKED ISSUE #6390: `terminate all` does not apply terminates when the service is restarted
**Describe the bug**
When a user issues `terminate all`, all queries are terminated, as expected. However, when ksql restarts, terminates get compacted. The compaction code does not understand the "all" keyword, interprets it as a query id, and does nothing because no such query exists. So the terminates are not applied. If the user subsequently dropped the sink of a terminated query, then the command runner cannot make progress.

**To Reproduce**
1. start ksqlDB master
1. create stream ...
2. create stream ... as select * from ...
3. terminate all
4. drop stream
5. restart

**Expected behavior**
The server restarts with no queries and the sink stream dropped.

**Actual behaviour**
The server starts the terminated query, and blocks trying to drop the sink stream.

agavra:
### Description 

fixes #6390

this PR now properly handles `TERMINATE ALL`. I checked to see if I could do this without "hacking" the fact that the `QueryId` is `ALL` but it would require parsing the statement, which would require quite a few changes (like passing the engine into the compactor, which doesn't sound right). Anything else (changing the `CommandId`) is not backwards compatible.

### Testing done 

Unit test

### Reviewer checklist

### DIFF
--- a/ksqldb-parser/src/main/java/io/confluent/ksql/parser/SqlFormatter.java
+++ b/ksqldb-parser/src/main/java/io/confluent/ksql/parser/SqlFormatter.java
@@ -399,7 +399,7 @@ public final class SqlFormatter {
     @Override
     protected Void visitTerminateQuery(final TerminateQuery node, final Integer context) {
       builder.append("TERMINATE ");
-      builder.append(node.getQueryId().map(QueryId::toString).orElse("ALL"));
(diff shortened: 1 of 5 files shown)


rodesai (approved):
LGTM!

PR #6396, opened 2020-10-09
chore: fix build by pinning AK to 6.1.0-23-ccs

agavra:
### Description 

The following AK commit: https://github.com/apache/kafka/pull/9156 seems to have broken the `KsMaterializationFunctionalTest#shouldIgnoreHavingClause` test. I'm not entirely sure why, but while we're looking into that I'm pinning AK to the earlier microversion to unblock our master

### Testing done 

Ran the test

### Reviewer checklist
- [ ] Ensure docs are updated if necessary. (eg. if a user visible feature is being added or changed).
- [ ] Ensure relevant issues are linked (description should include text like "Fixes #<issue number>")

PR #6399, opened 2020-10-10
chore: revert "fix build by pinning AK to 6.1.0-23-ccs"

agavra:
Reverts confluentinc/ksql#6396 because it breaks the main confluent packaging build.


agavra:
Merging without waiting for a green build because, well it won't pass...

PR #6400, opened 2020-10-10
feat: Add support for ALTER STREAM|TABLE

### LINKED ISSUE #6038: Support ALTER STREAM|TABLE ... ADD COLUMN statement

https://github.com/confluentinc/ksql/issues/2440 is adding support for `CREATE OR REPLACE` streams/tables which allows users to add new columns to a stream/table schema in a compatible way. However, `CREATE OR REPLACE` requires the user to specify the full original schema including the new added column, which will be tedious if the stream/table contains a very wide schema (10, 30, 100 cols?) and the user just wants to add one column.

`ALTER TABLE/STREAM` should allow users to easily add one or more columns without having to copy/paste the full schema and replace the stream/table. It is also useful for ksqlDB migrations + upgrades as users can only add an `ALTER TABLE` to the new SQL scripts.

Add support for `ALTER` commands to allow users change the schema of streams and tables. The `ALTER` command should be compatible with what `CREATE OR REPLACE` allows which is to only add new columns at the end of the schema.

ALTER syntax:
```
ALTER [TABLE | STREAM] <stream_or_table_name> <alter_option [, <alter_option>] ...];

alter_option:
    ADD [COLUMN] <col_name> <col_type>
```

The `ALTER` should be supported only on streams/tables created using DDL statements (`CREATE ... WITH`). Streams/tables created with `CREATE ... AS SELECT` cannot be altered as the schema is defined from the `SELECT` clause.

Example:
```
ksql> CREATE STREAM topic1(c1 INT, c2 INT) WITH (kafka_topic='t1', value_format='avro');
ksql> ALTER STREAM topic1 \
               ADD COLUMN c3 INT,
               ADD COLUMN c4 INT;
```

Continue using `CREATE OR REPLACE`. It is useful, but more tedious to use when using a very wide schema.

**Additional context**

User request for ALTER statements: https://github.com/confluentinc/ksql/issues/1725
Create or replace support: https://github.com/confluentinc/ksql/issues/2440

jzaralim:
### Description 

Add new syntax, `ALTER STREAM|TABLE`.  It works as follows:

```
ksql> alter stream riderlocations add column name string;

 Message
-------------------------------
 Stream RIDERLOCATIONS altered
-------------------------------

ksql> describe extended riderlocations;

...
Statement            : CREATE STREAM riderLocations (profileId VARCHAR, latitude DOUBLE, longitude DOUBLE)
  WITH (kafka_topic='quickstart-locations', value_format='json', partitions=1);alter stream riderLocations add column a string;alter stream riderlocations add column name string;

 Field     | Type
-----------------------------
 PROFILEID | VARCHAR(STRING)
 LATITUDE  | DOUBLE
 LONGITUDE | DOUBLE
 NAME      | VARCHAR(STRING)
-----------------------------
```

It only works on data sources that are created using DDL statements.
 
Fixes #6038 

### Testing done 
Unit and QTT tests.

### Reviewer checklist


JimGalasyn (approved):
LGTM!

ksqldb-engine/src/main/java/io/confluent/ksql/ddl/commands/DdlCommandExec.java
```diff
@@ -127,6 +131,81 @@ public DdlCommandResult executeDropType(final DropTypeCommand dropType) {
...
+      } else {
+        return alterTable(dataSource, newSchema);
+      }
+    }
+
+    private DdlCommandResult alterStream(
+        final DataSource dataSource,
+        final LogicalSchema newSchema
+    ) {
+      final KsqlStream<?> ksqlStream = new KsqlStream<>(
```
agavra, line 176: instead of constructing this here, we should add a method to `KsqlStream` labeled `withSchema`. This ensures that if anyone ever modifies `KsqlStream` and adds new arguments it won't accidentally be missed when we copy it here (same for `KsqlTable`).

also, it could make it cleaner if you just added it to `DataSource` API, then you don't even need to check which type it is :)

ksqldb-engine/src/main/java/io/confluent/ksql/ddl/commands/DdlCommandExec.java
```diff
@@ -127,6 +131,81 @@ public DdlCommandResult executeDropType(final DropTypeCommand dropType) {
...
+          dataSource.getSqlExpression().concat(sql),
```
agavra, line 177: how does this end up looking like to the user when you describe a source? is this the UX we want? (make sure to add test coverage for this in `ConsoleTest`

ksqldb-execution/src/main/java/io/confluent/ksql/execution/ddl/commands/AlterSourceCommand.java
```diff
@@ -0,0 +1,54 @@
...
+public class AlterSourceCommand implements DdlCommand {
+  private final SourceName sourceName;
+  private final String ksqlType;
+  private final LogicalSchema newColumns;
+
+
+  public AlterSourceCommand(
+      @JsonProperty(value = "sourceName", required = true) final SourceName sourceName,
+      @JsonProperty(value = "ksqlType", required = true) final String ksqlType,
+      @JsonProperty(value = "newColumns", required = true) final LogicalSchema newColumns
```
agavra, line 31: passing in a `LogicalSchema` to represent the new columns seems... odd? we should probably be passing down just the new columns - users shouldn't be, for example, creating new `KEY` columns (at the moment at least), and the new columns shouldn't include any of the psuedocolumns.

we can see this in the way that you use it (you only get the columns):
```java
alterSource.getNewColumns().columns()
```

ksqldb-functional-tests/src/test/resources/query-validation-tests/alter.json
```diff
@@ -0,0 +1,47 @@
+{
```
agavra, line 1: somewhere we should add a test that adds a column with the same name as a column that already exists

ksqldb-parser/src/main/java/io/confluent/ksql/parser/SqlFormatter.java
```diff
@@ -493,6 +495,22 @@ protected Void visitGroupBy(final GroupBy node, final Integer indent) {
       return null;
     }
 
+    @Override
+    public Void visitAlterSource(final AlterSource node, final Integer indent) {
```
agavra, line 499: I think we're missing test coverage for this in `SqlFormatterTest`

ksqldb-rest-model/src/main/java/io/confluent/ksql/rest/entity/CommandId.java
```diff
@@ -42,7 +42,8 @@
     CREATE,
     DROP,
     EXECUTE,
-    TERMINATE
+    TERMINATE,
+    ALTER
```
agavra, line 46: hmm, I'm wondering what this means for compatibility - I'm not sure what adding a value to an enum will do to the JSON serde. I think it works but we should test issuing some commands on an earlier version of ksqlDB then upgrading to a version that has this patch and seeing that it works. Similarly, we should test what happens if we downgrade (submit something with `ALTER` in it, and then roll back). We should make sure we gracefully handle the failure and enter `DEGRADED` state)

FYI @stevenpyzhang

ksqldb-parser/src/main/java/io/confluent/ksql/parser/tree/AlterSource.java
```diff
@@ -0,0 +1,98 @@
...
+    }
+    final AlterSource that = (AlterSource) o;
+    return Objects.equals(name, that.name)
+        && Objects.equals(dataSourceType, that.dataSourceType)
+        && Objects.equals(alterOptions, that.alterOptions);
+  }
+
+  @Override
+  public String toString() {
+    return toStringHelper(this)
```
agavra, line 92: nit - please use the string concatenation method for `toString` (that's what the rest of the code base uses)

agavra (commented):
Thanks @jzaralim! This approach is pretty clean and straightforward. Devil is in the details, so I've added a bunch of comments around that.

ksqldb-execution/src/main/java/io/confluent/ksql/execution/ddl/commands/AlterSourceCommand.java
jzaralim: This was done because the Column class could not be deserialized:

```
ERROR Did not write the command to the command topic as it could not be deserialized. This is a bug! Please raise a Github issue containing the series of commands you ran to get to this point.
Cannot construct instance of `io.confluent.ksql.schema.ksql.Column` (no Creators, like default construct, exist): cannot deserialize from Object value (no delegate- or property-based Creator)
```

But yeah, using LogicalSchema here is pretty odd. Perhaps creating a new class for this purpose would make more sense.

agavra: That makes sense, I think all you'd need to do is add a serde module. You can see `LogicalSchemaDeserializer` and `LogicalSchemaSerializer` (and they're hooked up in `KsqlTypesDeserializationModule`)

ksqldb-rest-model/src/main/java/io/confluent/ksql/rest/entity/CommandId.java
jzaralim: I don't think there's any good way to gracefully handle the failure without changing the earlier versions, which kinda goes against the point. And on top of that, the ddl command subtype `alterSourceV1` also causes backwards compatibility issues. So given that this is supposed to just be an alternative to `CREATE OR REPLACE`, I'll go ahead and  reuse the commands for `createStream/createTable` for this instead.

ksqldb-engine/src/main/java/io/confluent/ksql/ddl/commands/DdlCommandExec.java
```diff
@@ -127,6 +131,81 @@ public DdlCommandResult executeDropType(final DropTypeCommand dropType) {
...
+
+      if (!dataSource.getDataSourceType().getKsqlType().equals(alterSource.getKsqlType())) {
+        throw new KsqlException(String.format(
+            "Incompatible data source type is %s, but statement was ALTER %s",
+            dataSource.getDataSourceType().getKsqlType(),
+            alterSource.getKsqlType()
+        ));
+      }
+
+      if (dataSource.isCasTarget()) {
```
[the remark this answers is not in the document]
jzaralim: Same thing as `CREATE OR REPLACE` - the new rows are populated with `NULL`.

ksqldb-rest-model/src/main/java/io/confluent/ksql/rest/entity/CommandId.java
jzaralim: Sorry, ignore what I said - I tested this on a really old version that does not have the degraded state change. This actually does enter `DEGRADED` state, and switching between versions works.

jzaralim:
I did not write a very good commit message, so here are the main changes in the most recent commit:
 - switching to YATT tests
 - adding a Column serializer/deserializer
 - added a new KsqlException type for duplicate columns

agavra (approved):
LGTM! Awesome work @jzaralim

ksqldb-parser/src/main/java/io/confluent/ksql/parser/json/ColumnDeserializor.java
```diff
@@ -0,0 +1,60 @@
...
+      final JsonParser p,
+      final DeserializationContext ctxt
+  ) throws IOException {
+    final String text = p.readValueAs(String.class);
+    final SqlBaseLexer lexer = new SqlBaseLexer(
+        new CaseInsensitiveStream(CharStreams.fromString(text))
+    );
+    final CommonTokenStream tokStream = new CommonTokenStream(lexer);
+    final SqlBaseParser parser = new SqlBaseParser(tokStream);
+    final TableElementContext context = parser.tableElement();
```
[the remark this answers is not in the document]
jzaralim: Ended up using SchemaParser to parse the string instead

PR #6401, opened 2020-10-10
fix: Revert PR 6261 that introduced performance regression for pull queries

vpapavas:
### Description 
Revert PR #6261 as it introduced a 50% slowdown in pull query throughput.

### Testing done 
Ensure tests pass locally

### Reviewer checklist


big-andy-coates (dismissed):
Do we understand why this is causing the perf issue? It's only meant to be a refactor.

Rather than reverting the whole thing, it would be better to understand the cause and fix that.

AlanConfluent:
> Do we understand why this is causing the perf issue? It's only meant to be a refactor.
> 
> Rather than reverting the whole thing, it would be better to understand the cause and fix that.

I agree that it would be good to understand what the issue is, but if we don't roll it back, we'll be compounding the effects of this change and any additional commits that might affect pull query performance. Rather than block that other work, it's easier to roll this back so it can be understood on its own time.

agavra (approved):
+1 to what @AlanConfluent said (assuming we're certain that this is the cause). At the moment, this is blocking an already delayed release. I think it makes sense to revert first so long as master is "blocked" and then (when we are not pressured on time) revisit the refactor.

I think about it this way - the perf regression should have been caught in a per-commit test and this change would've never been merged to master. Just because we lack the per-commit testing for this doesn't mean it gets a free ride 😉

vpapavas:
@big-andy-coates We did not investigate what specifically in the refactor is causing the perf regression, just that this PR is the reason for it. As @agavra said, the investigation and fix should be on the committer and if we had automatic tests per commit, the PR would never have been merged. I can send you the flamegraphs and investigation I did to determine the performance drop and you can take it from there if you want.

big-andy-coates:
I love this new idea. Each time we find a bug we just revert the code that introduced it. Fantastic!

apurvam:
@agavra is this blocking 0.13? My understanding is that this was reverted on 0.13, since this patch just targets master. If it is just master, then we can also fix forward on 0.14, ie. file a blocker and assign to @big-andy-coates to fix the perf issue in lieu of reverting the entire patch. All the details of the perf regression (how to run the test, the existing metrics and flamegraphs, etc) can be shared here or in the new blocker ticket. 

In general terms, I think it is actually fine to undo a refactor if the team doesn't have the time to investigate and fix the perf implications of the said refactor.

agavra:
@apurvam your understanding is correct

apurvam:
@big-andy-coates the blocker issue is filed and assigned to you https://github.com/confluentinc/ksql/issues/6407. Please either fix it in before 0.14 is cut or unblock the merge of this PR. 

Also, 

> What do we do if the original committer is no longer involved in the project? Or is an external committer that we can't get in touch with?

Neither is true here. We have to take it on a case by case basis. We rarely revert commits which introduced bugs. You can count the number of reverts in the repo versus bugfixes to verify for yourself. We have to balance the importance of the commit with the impact of the bug.

In this case, the commit itself doesn't seem to be especially critical to ongoing work, but the negative impact on perf is provable and severe. And that negative impact clouds the performance testing of ongoing work on pull queries. On the balance, reverting it makes sense to me. And it is also absolutely right to require original patch author to address the issue, especially since it looks like it may be a bit involved to understand the causality and fix it.

vpapavas:
Fixed-forward with https://github.com/confluentinc/ksql/pull/6451

apurvam:
Thanks @vpapavas

Issue #3558, opened 2019-10-11
Avoid output `null` values when HAVING clauses not met


big-andy-coates:
e.g. given this test case:

```json
{
      "name": "with having expression (stream->table)",
      "statements": [
        "CREATE STREAM TEST (f1 INT) WITH (kafka_topic='test_topic', KEY='f1', value_format='DELIMITED');",
        "CREATE TABLE OUTPUT AS SELECT f1, COUNT(*) FROM TEST GROUP BY f1 HAVING SUM(f1) > 1;"
      ],
      "inputs": [
        {"topic": "test_topic", "key": "1", "value": "1"},
        {"topic": "test_topic", "key": "2", "value": "2"},
        {"topic": "test_topic", "key": "1", "value": "1"},
        {"topic": "test_topic", "key": "2", "value": "2"},
        {"topic": "test_topic", "key": "3", "value": "3"}
      ],
      "outputs": [
        {"topic": "OUTPUT", "key": "1", "value": null},
        {"topic": "OUTPUT", "key": "2", "value": "2,1"},
        {"topic": "OUTPUT", "key": "1", "value": "1,2"},
        {"topic": "OUTPUT", "key": "2", "value": "2,2"},
        {"topic": "OUTPUT", "key": "3", "value": "3,1"}
      ]
    }
```

Notice the `null` value output for the first input row.  This is because the sum of the aggregate after the first row isn't high enough to pass the HAVING criteria.

The issue here is that the HAVING criteria is implemented using a `KTable::filter` call, which is stateless, and hence does not _know_ there hasn't been a previous value that did pass the filter, and so it must emit a tombstone to clear any previous output row.

The fix is actually a fix in KS, which is to turn on `KTableImpl:: enableSendingOldValues` for any filter that is downstream of a statestore.  This will mean the filter receives a CDC event, (old, new), rather than just (new).  Meaning it can be more intelligent and only send the tombstone if the change transitions from passing the filter for the old value to not passing for the new.

vector1983:
@big-andy-coates Any work around for this? I keep see the ERROR log in KSQL:
ERROR Invalid format: 10.121.8.11 : null (io.confluent.ksql.structured.SqlPredicate)

big-andy-coates:
> @big-andy-coates Any work around for this? I keep see the ERROR log in KSQL:
> ERROR Invalid format: 10.121.8.11 : null (io.confluent.ksql.structured.SqlPredicate)

[...]

PR #6405, opened 2020-10-12
feat: avoid spurious tombstones in table output

big-andy-coates:
### Description 

fixes: https://github.com/confluentinc/ksql/issues/3558

AK commit https://github.com/apache/kafka/pull/9156 enhances Kafka Streams so that filters on tables now avoid emitting spurious tombstones. ksqlDB now benefits from this.  Tombstones are no longer emitted to the sink topic when a HAVING clause excludes a row from the result _that has never been in the result table_.

BREAKING CHANGE: This change fixes a _bug_ where unnecessary tombstones where being emitted when a `HAVING` clause filtered out a row from the source that is not in the output table

For example, given:

```sql
-- source stream:
CREATE STREAM FOO (ID INT KEY, VAL INT) WITH (...);

-- aggregate into a table:
CREATE TABLE BAR AS
    SELECT ID, SUM(VAL) AS SUM
    FROM FOO
    GROUP BY ID
    HAVING SUM(VAL) > 0;


-- insert some values into the stream:
INSERT INTO FOO VALUES(1, -5); 
INSERT INTO FOO VALUES(1, 6); 
INSERT INTO FOO VALUES(1, -2); 
INSERT INTO FOO VALUES(1, -1); 
```

Where previously the contents of the sink topic `BAR` would have contained records:

| Key | Value | Notes |
|-----|-------|------|
| 1.     | null.   | Spurious tombstone: the table does not contain a row with key `1`, so no tombstone is required. |
| 1.     | {sum=1} | Row added as HAVING criteria now met |
| 1.     | null.   | Row deleted as HAVING criteria now not met |
| 1.     | null.   | Spurious tombstone: the table does not contain a row with key `1`, so no tombstone is required. |

Note: the first record in the tom

The topic will now contain:

| Key | Value |
|-----|-------|
| 1.     | {sum=1} |
| 1.     | null.   |

Note: two historical tests are currently disabled. These need an upstream fix. See https://github.com/confluentinc/ksql/issues/6406 for details.

### Testing done 

usual

### Reviewer checklist


vcrfxia (approved):
LGTM. Thanks for the fix!

big-andy-coates:
More needed. Doing it now...

ksqldb-engine/src/test/java/io/confluent/ksql/materialization/ks/KsMaterializationFunctionalTest.java
```diff
@@ -619,8 +619,9 @@ public void shouldQueryMaterializedTableWithMultipleAggregationColumns() {
   }
 
   @Test
-  public void shouldIgnoreHavingClause() {
-    // Note: HAVING clause are handled centrally by KsqlMaterialization
+  public void shouldHandleHavingClause() {
+    // Note: HAVING clause are handled centrally by KsqlMaterialization. This logic will have been
+    // installed as part of building the below statement:
```
big-andy-coates, line 624: Note: test name and comments were misleading as the extra steps `KsqlMaterialization` adds to handle the HAVING clause are installed as part of this test.

ksqldb-engine/src/test/java/io/confluent/ksql/materialization/ks/KsMaterializationFunctionalTest.java
```diff
@@ -632,7 +633,11 @@ public void shouldIgnoreHavingClause() {
 
     final LogicalSchema schema = schema("COUNT", SqlTypes.BIGINT);
 
-    final Map<String, GenericRow> rows = waitForUniqueUserRows(STRING_DESERIALIZER, schema);
+    final int matches = (int) USER_DATA_PROVIDER.data().values().stream()
+        .filter(row -> ((Long) row.get(0)) > 2)
+        .count();
+
+    final Map<String, GenericRow> rows = waitForUniqueUserRows(matches, STRING_DESERIALIZER, schema);
```
big-andy-coates, line 640: The number of expected rows is now reduced as we no longer produce spurious tombstones.

ksqldb-engine/src/test/java/io/confluent/ksql/materialization/ks/KsMaterializationFunctionalTest.java
```diff
@@ -641,16 +646,22 @@ public void shouldIgnoreHavingClause() {
...
+      assertThat(row.map(Row::value), is(Optional.of(value)));
     });
+
+    USER_DATA_PROVIDER.data().entries().stream()
+        .filter(e -> !rows.containsKey(e.getKey().getString("USERID")))
+        .forEach(e -> {
+          // Rows filtered by the HAVING clause:
+          final Optional<Row> row = withRetry(() -> table.get(e.getKey()));
+          assertThat(row, is(Optional.empty()));
+        });
```
big-andy-coates, line 664: Get's against the table for filtered out rows should return nothing.

ksqldb-execution/src/main/java/io/confluent/ksql/execution/function/udaf/KudafAggregator.java
```diff
@@ -52,9 +52,11 @@ public KudafAggregator(
 
   @Override
   public GenericRow apply(final K k, final GenericRow rowValue, final GenericRow aggRowValue) {
+    final GenericRow result = GenericRow.fromList(aggRowValue.values());
```
big-andy-coates, line 55: Kafka Streams does not expect the aggregator to mutate its parameters. The streams code is passing in the "old value", which ksqlDB was then mutating and returning as the "new value". This meant, when then function returned, the old and new values matched.  This is obviously bad!

Code now takes a copy and mutates that.  There is a perf hit, obviously, but it's unavoidable.

ksqldb-execution/src/main/java/io/confluent/ksql/execution/function/udaf/KudafUndoAggregator.java
```diff
@@ -51,17 +51,19 @@ public KudafUndoAggregator(
   @SuppressWarnings({"unchecked", "rawtypes"})
   @Override
   public GenericRow apply(final Struct k, final GenericRow rowValue, final GenericRow aggRowValue) {
+    final GenericRow result = GenericRow.fromList(aggRowValue.values());
```
big-andy-coates, line 54: As above.

ksqldb-execution/src/main/java/io/confluent/ksql/execution/function/udaf/KudafAggregator.java
vcrfxia: Not sure I understand -- why did the old code work, in that case? Or did something change on the Streams side recently?

vcrfxia (approved):
LGTM besides the question inline, and the test failures in the build. Thanks!

ksqldb-execution/src/main/java/io/confluent/ksql/execution/function/udaf/KudafAggregator.java
big-andy-coates: The old code works because we were never enabling the sending of old values. We now do, to avoid the spurious tombstones.

vcrfxia: Sorry, still not understanding. What was being sent before, if not the old values? Was this method even being called, previously?

big-andy-coates: The processing nodes in the streams topology can optionally include the old/previous value, as well as the new/current value, to child nodes.  This is not on by default.  An upstream change to how table filters is handled means this is now turned on.

The streams code for aggregation looks something like:

```
V process(K key, Change<V> change) {
   // Get old value from store:
   final V oldAgg = store.get(key);

   // Undo any previous value:
  final T intermediateAgg = value.oldValue != null && oldAgg != null
    ? remove.apply(key, value.oldValue, oldAgg)
    : oldAgg;

   // Then add the new value
   final T newAgg;
   if (value.newValue != null) {
       final T initializedAgg = intermediateAgg == null
         ?  initializer.apply();
          : intermediateAgg;

       newAgg = add.apply(key, value.newValue, initializedAgg);
   } else {
       newAgg = intermediateAgg;
   }

   // update the store with the new value. & forard
   store.put(key, newAgg);
   tupleForwarder.maybeForward(key, newAgg, sendOldValues ? oldAgg : null);
}
```

The two calls: `remove.apply(key, value.oldValue, oldAgg)` and `add.apply(key, value.newValue, initializedAgg)` are calling out to ksqlDB code.  If these calls directly mutate the `oldAgg` or `initializedAgg` parameters passed, rather than creating copies, then the old and new values forwarded to child nodes will match. i.e. in `   tupleForwarder.maybeForward(key, newAgg, sendOldValues ? oldAgg : null)`, the parameters `newAgg` and `oldAgg` will have the same updated value, rather than `oldAgg` holding the previous value.  This breaks downstream processes, which expect the old and new value.

Previously the nodes weren't configured to send old values, so where just sending `null` for the old value and downstream could handle this correctly.

PR #6409, opened 2020-10-12
feat: Add support for IN clause to pull queries

AlanConfluent:
### Description 
Adds support for the `IN` clause to pull queries such that multiple keys can be fetched at once.  This can be used in conjunction with all of the existing window comparisions (e.g. windowstart, windowend).

An example of such a query is:
```
CREATE STREAM INPUT (ID STRING KEY, IGNORED INT) WITH (kafka_topic='test_topic', value_format='JSON');
CREATE TABLE AGGREGATE AS SELECT ID, COUNT(1) AS COUNT FROM INPUT GROUP BY ID;

SELECT * FROM AGGREGATE WHERE ID IN ('12345', '39874', '45450');
```

And you might get a response like this:
```
+----------------------------------+----------------+
|ID                                |COUNT           |
+----------------------------------+----------------+
|12345                             |8               |
|45450                             |10              |
```

### Testing done 
Ran unit tests and rest query validation tests.

### Reviewer checklist


ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/execution/PullQueryExecutor.java
```diff
@@ -224,26 +230,48 @@ public PullQueryResult execute(
...
+      final List<List<?>> tableRows = new ArrayList<>();
+      final List<LogicalSchema> schemas = new ArrayList<>();
+      final List<Future<PullQueryResult>> futures = new ArrayList<>();
+      final List<List<Struct>> keysByLocation = mat.locator().groupByLocation(
+          whereInfo.keysBound.stream()
+              .map(keyBound -> asKeyStruct(keyBound, query.getPhysicalSchema()))
+              .collect(Collectors.toList()));
+
+      for (List<Struct> keys : keysByLocation) {
+        final PullQueryContext pullQueryContext = new PullQueryContext(
```
vpapavas, line 243: The `pullQueryContext` is the same, the only thing that changes are the keys right?You could avoid creating a new object and just set the keys in each iteration.

ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/execution/PullQueryExecutor.java
```diff
@@ -224,26 +230,48 @@ public PullQueryResult execute(
           .getMaterialization(queryId, contextStacker)
           .orElseThrow(() -> notMaterializedException(getSourceName(analysis)));
 
-      final Struct key = asKeyStruct(whereInfo.keyBound, query.getPhysicalSchema());
+      final List<Optional<KsqlNode>> sourceNodes = new ArrayList<>();
+      final List<List<?>> tableRows = new ArrayList<>();
+      final List<LogicalSchema> schemas = new ArrayList<>();
+      final List<Future<PullQueryResult>> futures = new ArrayList<>();
+      final List<List<Struct>> keysByLocation = mat.locator().groupByLocation(
```
vpapavas, line 237: Why don't you return besides they keys also the active, standby per group? This way you wouldn't need to do `locate` twice basically.

ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/execution/PullQueryExecutor.java
```diff
@@ -271,8 +308,9 @@ private PullQueryResult handlePullQuery(
   ) {
     // Get active and standby nodes for this key
     final Locator locator = pullQueryContext.mat.locator();
+    final Struct key = Iterables.getLast(pullQueryContext.keys);
     final List<KsqlNode> filteredAndOrderedNodes = locator.locate(
-        pullQueryContext.key,
+        key,
```
vpapavas, line 313: You already did the locate when splitting the keys into groups

ksqldb-streams/src/main/java/io/confluent/ksql/execution/streams/materialization/ks/KsLocator.java
```diff
@@ -118,6 +123,21 @@
     return filteredHosts;
   }
 
+  @Override
+  public List<List<Struct>> groupByLocation(final List<Struct> keys) {
+    final Map<String, List<Struct>> groups = new HashMap<>();
+    for (Struct key : keys) {
```
vpapavas, line 129: Here you group keys based on the node that hosts the active partition. However, the keys may belong to different partitions and hence the standby may be different for keys in the same group. I think you should group them per <active, standby> combination else when we do the routing, we cannot assume that all keys go to the same standby  (when active dead)

ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/execution/PullQueryExecutor.java
vpapavas: Actually, I think currently the routing is wrong and you have to return the <active,standby> per group

ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/execution/PullQueryExecutor.java
```diff
@@ -761,19 +827,78 @@ private static Instant asInstant(final Expression other) {
   }
 
   private enum ComparisonTarget {
-    KEYCOL,
     WINDOWSTART,
     WINDOWEND
   }
 
-  private static Map<ComparisonTarget, List<ComparisonExpression>> extractComparisons(
+  private static class KeyAndWindowBounds {
```
vpapavas, line 834: What's the benefit of having this class versus adding one more entry in the `ComparisonTarget`? Seems a lot of work for achieving the same functionality?

ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/execution/PullQueryExecutor.java
```diff
@@ -224,26 +230,48 @@ public PullQueryResult execute(
...
-          key,
-          mat,
-          analysis,
-          whereInfo,
-          queryId,
-          contextStacker,
-          pullQueryMetrics);
+      }
+      for (Future<PullQueryResult> future : futures) {
+        final PullQueryResult result = future.get();
```
vpapavas, line 262: Shouldn't there be some error handling here? What happens if a thread dies due to an uncaught exception? We would like to fail the entire query and not return partial results, right? Maybe handle something like this:
```
        try {
            future.get();
        } catch (InterruptedException e) {
        } catch (ExecutionException e) {
            // Extract the actual exception from its wrapper
            Throwable t = e.getCause();
            System.err.println("Uncaught exception is detected! " + t
                    + " st: " + Arrays.toString(t.getStackTrace()));
            // ... Handle the exception
        }
```

vpapavas:
We are missing testing for the case where keys belong to different partitions, where there are partitions that are on the same active host but different standbys. I think this must be done manually, I don't think there is any automatic test currently for multiple partitions per host, right?

ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/execution/PullQueryExecutor.java
AlanConfluent: Yeah, I had thought that a given set of partitions were grouped together at each active and standby, but I think you're right this isn't the case.  I can fairly easily change this to be `keysByPartition` rather than host, though this isn't much better than just forwarding by individual key, unless the number of keys fetched is very large.  I'm going to think about if this can be easily reworked.

ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/execution/PullQueryExecutor.java
AlanConfluent: I had thought that a given set of partitions were grouped together at each active and standby, but I think you're right this isn't the case.  I'll rework this.

ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/execution/PullQueryExecutor.java
AlanConfluent: InPredicate isn't a ComparisonExpression.  Here, I can have a different type for each target type.

Alternatively, I could have change the map to point to an expression and casted back, but that's not a great practice.

ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/execution/PullQueryExecutor.java
AlanConfluent: `PullQueryContext` is threadsafe because it's immutable.  Dealing with small immutable objects is fairly efficient in the JVM and is considered worth doing to make things threadsafe and easy to understand.

ksqldb-streams/src/main/java/io/confluent/ksql/execution/streams/materialization/ks/KsLocator.java
AlanConfluent: Yeah, I think either grouping by partition or <active, standby> pair would be best.  I think in the normal case where you don't have many standbys, the latter is a good method.

ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/execution/PullQueryExecutor.java
AlanConfluent: Yeah, I agree that we don't want partial results, so I wouldn't want to catch each individual future.

Also, calling interrupt is what causes the `InterruptedException` and it sends that exception to other threads.  To have the current thread have it thrown, all we have to do is not catch it.

I pretty much just let the existing handler catch Exception in the wider scope.  I added a small special case to unwrap `ExecutionException` so we get the same error message.

ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/execution/PullQueryExecutor.java
AlanConfluent: I reworked this so that it now groups key together only if they share the same list of nodes, including active and all standbys.  Most of the time if there's 1 or 2 standbys and lots of keys fetched, this will hopefully reduce unnecessary calls.

AlanConfluent:
> We are missing testing for the case where keys belong to different partitions, where there are partitions that are on the same active host but different standbys. I think this must be done manually, I don't think there is any automatic test currently for multiple partitions per host, right?

@vpapavas Yeah, there's currently no test of this sort with multiple partitions where we also have failures (since those are required to resort to different standbys).  Muckrake tests actually might do the job, but I'll also try to test this manually.

I addressed all of your above comments and reworked the code to not only find the location once, but also do it based on the whole list of active and standbys, so this should cover the different partitions case.

ksqldb-engine/src/main/java/io/confluent/ksql/engine/rewrite/PullQueryKeyUpdater.java
```diff
@@ -0,0 +1,106 @@
...
+        case INT16:
+        case INT32:
+          return new IntegerLiteral((int) value);
+        case INT64:
+          return new LongLiteral((long) value);
+        case FLOAT32:
+        case FLOAT64:
+          return new DoubleLiteral((double) value);
+        default:
+          throw new KsqlException("Unknown key type " + schema.type());
```
agavra, line 102: very soon we're going to support all expression types as keys (including structs) and we already support `DECIMAL` keys. I'm not entirely sure that this approach will work then since it's not always possible to take a java object and convert it into an expression as the inverse conversion is lossy. Is it possible to extract the original expression when we construct the `List<Struct> keys` instead of mapping it to Java objects?

ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/execution/PullQueryExecutor.java
```diff
@@ -224,36 +234,73 @@ public PullQueryResult execute(
...
 
-      final PullQueryResult result = handlePullQuery(
-          statement,
-          executionContext,
-          serviceContext,
-          pullQueryContext,
-          routingOptions
-      );
+      }
+      for (Future<PullQueryResult> future : futures) {
```
agavra, line 270: is there any requirement that the pull query result is reproducible? it seems like we add rows to the result based on the order of the nodes that we request, but those nodes can host different partitions at different times. I think it's reasonable to not guarantee ordering, but food for thought (and probably should be documented)

ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/execution/PullQueryExecutor.java
agavra: > calling interrupt is what causes the `InterruptedException`

I think you probably meant this, but just a small clarification here - `Thread.currentThread().interrupt()` does not directly cause an `InterruptedException`. It just sets an interrupt flag on the thread's state, and then any blocking code is expected to check that flag and raise an `InterruptedException` when it notices that flag. There is no guarantee that calling `interupt` will actually do anything (especially if there's misbehaving client code). It's a cooperative strategy.

ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/execution/PullQueryExecutor.java
```diff
@@ -289,11 +329,14 @@ private PullQueryResult handlePullQuery(
...
     for (KsqlNode node : filteredAndOrderedNodes) {
       try {
+        final TableRows rows
+            = routeQuery(node, statement, executionContext, serviceContext, pullQueryContext);
         final Optional<KsqlNode> debugNode = Optional.ofNullable(
             routingOptions.isDebugRequest() ? node : null);
-        return new PullQueryResult(
-            routeQuery(node, statement, executionContext, serviceContext, pullQueryContext),
-            debugNode);
+        final List<Optional<KsqlNode>> debugNodes = rows.getRows().stream()
```
agavra, line 336: nit: probably makes sense to make this `Optional<List<KsqlNode>>` instead of `List<Optional<KsqlNode>>` to prevent creating a list of empty optionals in the non-debug case. And then you can also use [`Collections#nCopies`](https://docs.oracle.com/javase/7/docs/api/java/util/Collections.html#nCopies(int,%20T)) to make it a little more readable

ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/execution/PullQueryExecutor.java
```diff
@@ -389,8 +438,14 @@ private static TableRows queryRowsLocally(
   private static TableRows forwardTo(
       final KsqlNode owner,
       final ConfiguredStatement<Query> statement,
-      final ServiceContext serviceContext
+      final ServiceContext serviceContext,
+      final PullQueryContext pullQueryContext
   ) {
+    // Rewrite the expression to only query for the particular keys we care about for this node.
+    // Otherwise, we'll risk reading standby data for other partitions.
```
agavra, line 445: I'm worried this approach doesn't scale well for things other than `IN` queries (not to mention that it feels hacky). Instead, it probably makes sense to have a way to specify that a pull query (internally routed only) should only read from certain partitions. Otherwise, how would we handle things like range queries? I can't "rewrite" the range query and avoid the risk of reading standby data. 

Generally, I think it might make sense to think about communicating pull queries internally using something other than just the SQL statement.

ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/execution/PullQueryExecutor.java
```diff
@@ -535,44 +590,65 @@ private static WhereInfo extractWhereInfo(
...
       final boolean windowed,
       final LogicalSchema schema
   ) {
-    if (comparisons.size() != 1) {
-      throw invalidWhereClauseException("Multiple bounds on key column", windowed);
+    final InPredicate inPredicate = Iterables.getLast(inPredicates);
+    final List<Object> result = new ArrayList<>();
+    for (Expression expression : inPredicate.getValueList().getValues()) {
+      if (!(expression instanceof Literal)) {
+        throw new KsqlException("Ony comparison to literals is currently supported: "
```
agavra, line 634:
```suggestion
        throw new KsqlException("Only comparison to literals is currently supported: "
```

ksqldb-streams/src/main/java/io/confluent/ksql/execution/streams/materialization/ks/KsLocator.java
```diff
@@ -68,54 +74,61 @@
...
+      // is applied.
+      final ImmutableList<KsqlNode> filteredHosts = allHosts.stream()
+          .filter(routingFilter::filter)
+          .map(this::asNode)
+          .collect(ImmutableList.toImmutableList());
+
+      LOG.debug("Filtered and ordered hosts: {}", filteredHosts);
+
+      groups.computeIfAbsent(filteredHosts, nl -> new ArrayList<>());
+      groups.get(filteredHosts).add(key);
```
agavra, line 127: nit: `computeIfAbsent` returns the result of the compute, so you could just 
```suggestion
      groups.computeIfAbsent(filteredHosts, nl -> new ArrayList<>()).add(key);
```

agavra (commented):
Thanks @AlanConfluent! This feature is going to be a huge improvement over the existing functionality. I'm mostly aligned with the approach, but some implementation concerns inline.

ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/execution/PullQueryExecutor.java
AlanConfluent: Yeah, that's fair.  We actually were discussing range queries as well and it's probably easier to use a unified approach.  I think that it should be sufficient to specify partition when reading from the table (to ensure you're only reading from the actives and standbys you intend) and also on the second hop, avoid reading keys that are not hosted on this machine or that don't agree with those specified partitions.  That same approach will work on range queries as well.

ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/execution/PullQueryExecutor.java
AlanConfluent: Yes, you're right that there are no guarantees about ordering.  I'll document that in the code and consider documenting elsewhere as well.

ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/execution/PullQueryExecutor.java
AlanConfluent: Yeah, you're right that a thread may have to explicitly check for being interrupted if it's doing a lot of computation.  If it's sleeping in some manner, there's a good chance that it will be thrown for the thread, e.g. `Thread.sleep`.

ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/execution/PullQueryExecutor.java
AlanConfluent: I had been thinking that maybe this was necessary because different machines can have different configs, but I now remember that this is a request config, so it should always work.  Ok, made it `Optional<List<KsqlNode>>`

ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/execution/PullQueryExecutor.java
AlanConfluent: Ok, implemented that other approach where partition is specified.  No rewriting necessary.

big-andy-coates (commented):
Haven't looked in depth at the code - just scanned the test cases. They LGTM!

ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/execution/PullQueryExecutor.java
```diff
@@ -265,45 +294,104 @@ void checkRateLimit() {
...
+  private static Map<KsqlNode, List<KsqlLocation>> groupByHost(
+      final ConfiguredStatement<Query> statement,
+      final List<KsqlLocation> locations,
+      final int round) {
+    final Map<KsqlNode, List<KsqlLocation>> groupedByHost = new LinkedHashMap<>();
+    for (KsqlLocation location : locations) {
+      // If one of the partitions required is out of nodes, then we cannot continue.
+      if (round >= location.getNodes().size()) {
+        throw new MaterializationException(String.format(
+            "Unable to execute pull query: %s", statement.getStatementText()));
```
vpapavas, line 389: Consider adding a more descriptive error message so that we know at which point in the code the query failed and why.

ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/execution/PullQueryExecutor.java
```diff
@@ -265,45 +294,104 @@ void checkRateLimit() {
...
+            new TableRows(statement.getStatementText(), queryId, Iterables.getLast(schemas),
+                tableRows),
+            sourceNodes.isEmpty() ? Optional.empty() : Optional.of(sourceNodes));
       }
     }
-    throw new MaterializationException(String.format(
-        "Unable to execute pull query: %s", statement.getStatementText()));
+  }
+
+  private static Map<KsqlNode, List<KsqlLocation>> groupByHost(
```
vpapavas, line 380: What this method does, is group all locations per the same host, which if round=0, will be the active. So, all locations (all keys) that have the same host as active will we grouped together. Then, in the second round, for any keys that the active failed, we will get the standby that is second in ordering.

ksqldb-streams/src/main/java/io/confluent/ksql/execution/streams/materialization/ks/KsLocator.java
```diff
@@ -68,28 +73,64 @@
...
-      LOG.debug("KeyQueryMetadata not available for state store {} and key {}",
-                stateStoreName, key);
-      throw new MaterializationException(String.format(
-          "KeyQueryMetadata not available for state store %s and key %s", stateStoreName, key));
-    }
+    // Maintain request order for reproducibility by using a LinkedHashMap, even though it's
+    // not a guarantee of the API.
+    final Map<Integer, List<KsqlNode>> locationsByPartition = new LinkedHashMap<>();
+    final Map<Integer, Set<Struct>> keysByPartition = new HashMap<>();
+    final Set<Integer> filterPartitions = routingOptions.getPartitions();
```
vpapavas, line 85: 1. Why do you need to know the partitions here? Is it because you don't know which partitions are active and which standbys? If so, we can get this information with `KafkaStreams.allMetadata()`.  If we know what each partition is, then for a forwarded request, we know we need to access the standby partitions whereas for a non-forwarded only the active ones. This could avoid sending the partitions over the network causing overhead.

2. If you include the partitions in the request (and the host per list<partition>), I think we can avoid all this lookup and grouping per keys for forwarded requests. I know we were already doing this for a single key but that was cheap. Now, I feel we do unnecessary work for forwarded requests. What is the benefit of doing the grouping again for forwarded requests?

ksqldb-streams/src/main/java/io/confluent/ksql/execution/streams/materialization/ks/KsLocator.java
```diff
@@ -68,28 +73,64 @@
...
+    for (Struct key : keys) {
```
vpapavas, line 86: Can we make this loop cheaper for forwarded requests if we do send the partitions in the request?

AlanConfluent: The only issue is that we still need to find which keys belong to which partitions to figure out which we should query for locally so we don't try them unnecessarily on rocksdb, which would also be expensive.  I could try to expose the info necessary to create a `DefaultStreamPartitioner`, but this seems a bit complex.  This may be a little wasteful, but it's not too bad.

ksqldb-functional-tests/src/test/resources/rest-query-validation-tests/pull-queries-against-materialized-aggregates.json
```diff
@@ -1540,6 +1540,404 @@
...
+        "CREATE STREAM INPUT (ID INT KEY, IGNORED INT) WITH (kafka_topic='test_topic', value_format='JSON');",
+        "CREATE TABLE AGGREGATE AS SELECT ID, COUNT(1) AS COUNT FROM INPUT GROUP BY ID;",
+        "SELECT * FROM AGGREGATE WHERE ID IN ('10', 8);"
+      ],
+      "expectedError": {
+        "type": "io.confluent.ksql.rest.entity.KsqlStatementErrorMessage",
+        "message": "'10' can not be converted to the type of the key column: ID INTEGER KEY",
+        "status": 400
+      }
+    },
```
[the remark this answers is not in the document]
AlanConfluent: Interesting.  I find it a bit funny that `a IN (v1)` isn't just syntactic sugar for `a = v1`.  I may add that as a followup to try to keep from adding much more to this PR.

ksqldb-streams/src/main/java/io/confluent/ksql/execution/streams/materialization/ks/KsLocator.java
vpapavas: I was thinking that you can be sure if it is a forwarded request, you need to access the standby partitions. And if you can know which partitions are active and which standby, then you can answer the question which partitions should be used to lookup keys. A request that is not forwarded always looksup active partitions whereas a forwarded request always looks up standby partitions.

ksqldb-streams/src/main/java/io/confluent/ksql/execution/streams/materialization/ks/KsLocator.java
vpapavas: No it's ok. I also realized that there is not much to do that is a low hanging fruit. Just wanted to pick your brain if you had anything in mind. But it's not necessary to do now.

ksqldb-streams/src/main/java/io/confluent/ksql/execution/streams/materialization/ks/KsLocator.java
vpapavas: Discussed offline. The reason we need the partitions is that we need to know also which standby we want since they are ordered so we cannot rely only on the round number

vpapavas (approved):
Awesome job @AlanConfluent ! LGTM!

PR #6432, opened 2020-10-13
chore: qtt coverage for avro primitive keys

agavra:
### Description 

Adds coverage for AVRO in QTT tests. In order to support decimal (and in future when we want to compare structs) I needed to change the test framework to compare the `JsonNode` instead of the java-ified value, piggy-backing on the existing value comparison framework.

### Testing done 

Ran the QTT tests :) 

### Reviewer checklist

### DIFF
--- a/ksqldb-functional-tests/src/test/java/io/confluent/ksql/test/tools/RecordTest.java
+++ b/ksqldb-functional-tests/src/test/java/io/confluent/ksql/test/tools/RecordTest.java
@@ -19,7 +19,6 @@ import static org.hamcrest.CoreMatchers.equalTo;
 import static org.hamcrest.CoreMatchers.instanceOf;
 import static org.hamcrest.MatcherAssert.assertThat;
 
-import com.fasterxml.jackson.databind.node.JsonNodeFactory;
 import io.confluent.ksql.test.model.WindowData;
 import java.util.Optional;
 import org.apache.kafka.streams.kstream.Windowed;
--- a/ksqldb-functional-tests/src/test/java/io/confluent/ksql/test/tools/RecordTest.java
+++ b/ksqldb-functional-tests/src/test/java/io/confluent/ksql/test/tools/RecordTest.java
@@ -19,6 +19,7 @@ import static org.hamcrest.CoreMatchers.equalTo;
 import static org.hamcrest.CoreMatchers.instanceOf;
 import static org.hamcrest.MatcherAssert.assertThat;
 
+import com.fasterxml.jackson.databind.node.JsonNodeFactory;
 import io.confluent.ksql.test.model.WindowData;
 import java.util.Optional;
 import org.apache.kafka.streams.kstream.Windowed;
@@ -36,6 +37,7 @@ public class RecordTest {
     final Record record = new Record(
         TOPIC_NAME,
         10,
+        null,
         "bar",
         null,
         Optional.of(1000L),
@@ -55,6 +57,7 @@ public class RecordTest {
     final Record record = new Record(
         TOPIC_NAME,
         "foo",
+        null,
         "bar",
         null,
         Optional.of(1000L),
@@ -78,6 +81,7 @@ public class RecordTest {
     final Record record = new Record(
         TOPIC_NAME,
         "foo",
+        null,
         "bar",
         null,
         Optional.of(1000L),
--- a/ksqldb-functional-tests/src/test/java/io/confluent/ksql/test/tools/TestExecutorTest.java
+++ b/ksqldb-functional-tests/src/test/java/io/confluent/ksql/test/tools/TestExecutorTest.java
@@ -342,8 +342,8 @@ public class TestExecutorTest {
     final ProducerRecord<byte[], byte[]> rec0 = producerRecord(sinkTopic, 123456719L, "k1", "v1");
     when(kafkaService.readRecords(SINK_TOPIC_NAME)).thenReturn(ImmutableList.of(rec0));
 
-    final Record expected_0 = new Record(SINK_TOPIC_NAME, "k1", "v1", null, Optional.of(1L), null);
-    final Record expected_1 = new Record(SINK_TOPIC_NAME, "k1", "v1", null, Optional.of(1L), null);
+    final Record expected_0 = new Record(SINK_TOPIC_NAME, "k1", null, "v1", null, Optional.of(1L), null);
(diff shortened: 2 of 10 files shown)


ksqldb-functional-tests/src/test/resources/query-validation-tests/avro.json
```diff
@@ -667,6 +667,33 @@
...
+      "statements": [
+        "CREATE STREAM INPUT (foo INT) WITH (kafka_topic='input_topic', format='AVRO');",
+        "CREATE STREAM OUTPUT AS SELECT * FROM INPUT;"
+      ],
+      "topics": [
+        {
+          "name": "input_topic",
+          "keySchema": {"type": "boolean"},
+          "keyFormat": "AVRO",
+          "valueSchema": {"name": "ignored", "type": "record", "fields": [{"name": "FOO", "type": "int"}]},
```
vcrfxia, line 684: Is there a purpose to including this value schema here, given that the test does not use value inference? If we want to test that this value schema is ignored in favor of the schema that's explicitly supplied in the statement, we should update this schema to be different. If not, can we remove this in order to clean up the tests?

Or does the QTT framework require that it's present for Avro types, even if the schema inference isn't used...?

ksqldb-functional-tests/src/test/resources/query-validation-tests/avro.json
```diff
@@ -696,6 +723,340 @@
...
+        "ksql.key.format.enabled": true
+      },
+      "statements": [
+        "CREATE STREAM INPUT (K BIGINT KEY, foo INT) WITH (kafka_topic='input_topic', format='AVRO');",
+        "CREATE STREAM OUTPUT AS SELECT * FROM INPUT;"
+      ],
+      "topics": [
+        {
+          "name": "input_topic",
+          "keySchema": {"type": "long"},
```
vcrfxia, line 767: Same question about whether this key schema needs to be provided, given that there is no key inference in this test.

ksqldb-functional-tests/src/test/java/io/confluent/ksql/test/tools/RecordTest.java
```diff
@@ -19,6 +19,7 @@
 import static org.hamcrest.CoreMatchers.instanceOf;
 import static org.hamcrest.MatcherAssert.assertThat;
 
+import com.fasterxml.jackson.databind.node.JsonNodeFactory;
```
vcrfxia, line 22: nit: unused import.

ksqldb-functional-tests/src/test/java/io/confluent/ksql/test/tools/TestExecutorTest.java
```diff
@@ -414,9 +415,9 @@ public void shouldPassOnExpectedOutput() {
     final ProducerRecord<byte[], byte[]> rec1 = producerRecord(sinkTopic, 123456789L, "k2", "v2");
     when(kafkaService.readRecords(SINK_TOPIC_NAME)).thenReturn(ImmutableList.of(rec0, rec1));
 
-    final Record expected_0 = new Record(SINK_TOPIC_NAME, "k1", "v1", TextNode.valueOf("v1"),
+    final Record expected_0 = new Record(SINK_TOPIC_NAME, "k1", JsonNodeFactory.instance.textNode("k1"), "v1", TextNode.valueOf("v1"),
```
vcrfxia, line 418: Is this different from `TextNode.valueOf("k1")`? Curious about the inconsistency between the key and value nodes.

ksqldb-functional-tests/src/test/resources/query-validation-tests/avro.json
```diff
@@ -696,6 +723,340 @@
...
+            "items": "string"
+          },
+          "keyFormat": "AVRO",
+          "valueSchema": {"name": "ignored", "type": "record", "fields": [{"name": "FOO", "type": "int"}]},
+          "valueFormat": "AVRO"
+        }
+      ],
+      "expectedException": {
+        "type": "io.confluent.ksql.util.KsqlStatementException",
+        "message": "Unsupported key schema: [`ROWVAL` ARRAY<STRING> KEY]"
```
vcrfxia, line 1028: Whoops, this should be renamed to `ROWKEY` with the change in https://github.com/confluentinc/ksql/pull/6436. You should merge first and I'll update the test in my PR accordingly :)

vcrfxia (approved):
LGTM -- exciting stuff!

ksqldb-functional-tests/src/test/resources/query-validation-tests/avro.json
agavra: 💯 that's good haha

ksqldb-functional-tests/src/test/java/io/confluent/ksql/test/tools/TestExecutorTest.java
agavra: 🙄 to be honest, I've just always used `JsonNodeFactory` to create the nodes (I didn't even notice that they were using `TextNode` here). I don't think there's a difference:
```java
    // From JsonNodeFactory
    public TextNode textNode(String text) { return TextNode.valueOf(text); }
```

ksqldb-functional-tests/src/test/resources/query-validation-tests/avro.json
agavra: I don't think it's necessary - I can add separate tests to make sure that it's respected if the schema differs.

agavra: Actually, it looks like QTT framework does require this even if it's present for Avro types. We can look at removing that later

agavra: OK - I figured out what's going on and I'll fix it in a follow-up PR. We generate the topic information from the CT/CS statement if there is no topic node in the QTT. The problem is that if we have either the key or the value schema we need both because otherwise it won't generate the topic from the CT statement and use the one supplied (and we get a 404 when looking up the schema in SR). The fix isn't necessarily straightforward, so for now whenever we are using inference I'll just supply the key and value schema.

agavra: https://github.com/confluentinc/ksql/issues/6441#issue-722459755

vcrfxia: Thanks for investigating! Agreed we shouldn't block this PR.

PR #6437, opened 2020-10-14
fix: Properly clean up state when executing a command fails

### LINKED ISSUE #6340: ksqlDB does not clear command futures map if a command transaction fails to commit
**Describe the bug**
ksqlDB does not clear command futures map if a command transaction fails to commit.

**Expected behavior**
A ksqlDB command may fail to commit. This might happen for a variety of not unexpected reasons: a transient issue between ksql and the broker, a conflicting command, etc. When this happens, a subsequent retry of the command should work fine (as long as it passes validation).

**Actual behaviour**
Instead, if the command fails to commit a subsequent retry will return the following error:
```
Could not write the statement '<statement>' into the command topic.
Caused by: Another command with the same id (<command id>) is being executed
```

**Proposed solution**
To fix this, we need to clean up the command from the futures map on a commit failure.

AlanConfluent:
### Description 
Attempts to cleanup and state associated with a command id when a transaction for that command fails to commit properly.

Fixes #6340 

### Testing done 
Ran unit tests

### Reviewer checklist

### DIFF
--- a/ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/computation/CommandStore.java
+++ b/ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/computation/CommandStore.java
@@ -304,6 +304,21 @@ public class CommandStore implements CommandQueue, Closeable {
   }
 
   @Override
+  public void abortCommand(final CommandId commandId) {
+    commandStatusMap.compute(
+        commandId,
+        (k, v) -> {
+          if (v == null) {
+            throw new IllegalStateException(
+                String.format("Trying to abort an unknown commandId (%s)", commandId)
+            );
+          }
+          return null;
+        }
+    );
+  }
+
+  @Override
   public void waitForCommandConsumer() {
     try {
       final long endOffset = getCommandTopicOffset();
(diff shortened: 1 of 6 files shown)


ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/computation/CommandStore.java
```diff
@@ -303,6 +303,21 @@ public void ensureConsumedPast(final long seqNum, final Duration timeout)
     );
   }
 
+  @Override
+  public void abortCommand(final CommandId commandId) {
+    commandStatusMap.compute(
+        commandId,
+        (k, v) -> {
+          if (v == null) {
+            throw new IllegalStateException(
```
rodesai, line 312: We could actually call this before the command is registered right? If something throws between generating the command id and enqueueing?

AlanConfluent: That's true, it is possible.

Ok, now I just log when something is being aborted, but throw no error either way.

rodesai (approved):
LGTM

Issue #6444, opened 2020-10-16
Schema inference does not support unwrapped structs


vcrfxia:
**Describe the bug**

The following fails:
```
  {
      "name": "deserialize anonymous struct - value - inference",
      "statements": [
        "CREATE STREAM INPUT WITH (kafka_topic='input_topic', value_format='AVRO', wrap_single_value=false);",
        "CREATE STREAM OUTPUT AS SELECT * FROM INPUT;"
      ],
      "topics": [
        {
          "name": "input_topic",
          "valueSchema": {"type": "record", "name": "ignored", "fields": [{"name": "F0", "type": ["null", "int"]}]}
        }
      ],
      "inputs": [
        {"topic": "input_topic", "value": {"F0": 1}},
        {"topic": "input_topic", "value": {"F0": null}},
        {"topic": "input_topic", "value": null}
      ],
      "outputs": [
        {"topic": "OUTPUT", "value": {"ROWVAL": {"F0": 1}}},
        {"topic": "OUTPUT", "value": {"ROWVAL": {"F0": null}}},
        {"topic": "OUTPUT", "value": null}
      ]
    }
```
with
```
org.apache.kafka.common.errors.SerializationException: Error deserializing message from topic: input_topic
Caused by: org.apache.kafka.connect.errors.DataException: Cannot deserialize type struct as type int32 for path: 
	at io.confluent.ksql.serde.connect.ConnectDataTranslator.throwTypeMismatchException(ConnectDataTranslator.java:71)
	at io.confluent.ksql.serde.connect.ConnectDataTranslator.validateType(ConnectDataTranslator.java:90)
	at io.confluent.ksql.serde.connect.ConnectDataTranslator.validateSchema(ConnectDataTranslator.java:165)
	at io.confluent.ksql.serde.connect.ConnectDataTranslator.toKsqlValue(ConnectDataTranslator.java:215)
	at io.confluent.ksql.serde.connect.ConnectDataTranslator.toKsqlRow(ConnectDataTranslator.java:58)
	at io.confluent.ksql.serde.avro.AvroDataTranslator.toKsqlRow(AvroDataTranslator.java:65)
	at io.confluent.ksql.serde.connect.KsqlConnectDeserializer.deserialize(KsqlConnectDeserializer.java:51)
	at io.confluent.ksql.serde.tls.ThreadLocalDeserializer.deserialize(ThreadLocalDeserializer.java:37)
[...]

PR #6446, opened 2020-10-16
fix: support unwrapped struct value inference

vcrfxia:
### Description 

Fixes https://github.com/confluentinc/ksql/issues/6444

The cause of the bug is that the DefaultSchemaInjector always unwraps struct schemas (assuming that they're wrapped), ignoring any serde features that should cause it to not do so (specifically, `wrap_single_value=false`). This PR fixes the bug by passing serde features into the schema translator, which is admittedly hacky and ugly but I couldn't think of a way around this. 

### Testing done 

QTT

### Reviewer checklist

### DIFF
--- a/ksqldb-engine/src/main/java/io/confluent/ksql/schema/ksql/inference/DefaultSchemaInjector.java
+++ b/ksqldb-engine/src/main/java/io/confluent/ksql/schema/ksql/inference/DefaultSchemaInjector.java
@@ -33,6 +33,7 @@ import io.confluent.ksql.serde.FormatFactory;
 import io.confluent.ksql.serde.FormatInfo;
 import io.confluent.ksql.serde.SerdeFeature;
 import io.confluent.ksql.serde.SerdeFeatures;
+import io.confluent.ksql.serde.SerdeFeaturesFactory;
 import io.confluent.ksql.statement.ConfiguredStatement;
 import io.confluent.ksql.statement.Injector;
 import io.confluent.ksql.util.ErrorMessageUtil;
@@ -123,7 +124,7 @@ public class DefaultSchemaInjector implements Injector {
         props.getKafkaTopic(),
         props.getKeySchemaId(),
         keyFormat,
-        SerdeFeatures.of(SerdeFeature.UNWRAP_SINGLES),
+        SerdeFeaturesFactory.buildInternal(FormatFactory.of(keyFormat)),
         statement.getStatementText(),
         true
     ));
--- a/ksqldb-engine/src/main/java/io/confluent/ksql/schema/ksql/inference/DefaultSchemaInjector.java
+++ b/ksqldb-engine/src/main/java/io/confluent/ksql/schema/ksql/inference/DefaultSchemaInjector.java
@@ -32,6 +32,7 @@ import io.confluent.ksql.schema.ksql.inference.TopicSchemaSupplier.SchemaResult;
 import io.confluent.ksql.serde.FormatFactory;
 import io.confluent.ksql.serde.FormatInfo;
 import io.confluent.ksql.serde.SerdeFeature;
+import io.confluent.ksql.serde.SerdeFeatures;
 import io.confluent.ksql.statement.ConfiguredStatement;
 import io.confluent.ksql.statement.Injector;
 import io.confluent.ksql.util.ErrorMessageUtil;
@@ -122,6 +123,7 @@ public class DefaultSchemaInjector implements Injector {
         props.getKafkaTopic(),
         props.getKeySchemaId(),
         keyFormat,
+        SerdeFeatures.of(SerdeFeature.UNWRAP_SINGLES),
         statement.getStatementText(),
         true
     ));
@@ -141,6 +143,7 @@ public class DefaultSchemaInjector implements Injector {
         props.getKafkaTopic(),
         props.getValueSchemaId(),
         valueFormat,
+        props.getValueSerdeFeatures(),
         statement.getStatementText(),
         false
     ));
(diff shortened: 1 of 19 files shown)


ksqldb-engine/src/main/java/io/confluent/ksql/schema/ksql/inference/SchemaRegistryTopicSchemaSupplier.java
```diff
@@ -127,7 +133,11 @@ private SchemaResult fromParsedSchema(
 
     final List<SimpleColumn> columns;
     try {
-      columns = translator.toColumns(parsedSchema, isKey);
+      columns = translator.toColumns(
+          parsedSchema,
+          isKey,
+          serdeFeatures.enabled(SerdeFeature.UNWRAP_SINGLES)
```
big-andy-coates, line 139: I'd probably just pass all the features down.  There could be other features later that also affect the schema.

ksqldb-serde/src/main/java/io/confluent/ksql/serde/connect/ConnectFormatSchemaTranslator.java
```diff
@@ -59,16 +59,29 @@ public String name() {
...
   @Override
-  public List<SimpleColumn> toColumns(final ParsedSchema schema, final boolean isKey) {
+  public List<SimpleColumn> toColumns(
+      final ParsedSchema schema,
+      final boolean isKey,
+      final boolean unwrapSingle) {
     Schema connectSchema = connectSrTranslator.toConnectSchema(schema);
 
-    if (connectSchema.type() != Type.STRUCT) {
+    if (connectSchema.type() != Type.STRUCT || unwrapSingle) {
```
big-andy-coates, line 68: I don't think we need to check both.   Just checking the serde features is enough.  So I think this should become:

```java
 @Override
  public List<SimpleColumn> toColumns(
      final ParsedSchema schema,
      final SerdeFeatures serdeFeatures,
      final boolean isKey
  ) {
    SerdeUtils.throwOnUnsupportedFeatures(serdeFeatures, format.supportedFeatures());

    Schema connectSchema = connectSrTranslator.toConnectSchema(schema);

    if (serdeFeatures.enabled(SerdeFeature.UNWRAP_SINGLES)) {
      connectSchema = SerdeUtils.wrapSingle(connectSchema, isKey);
    }

    if (connectSchema.type() != Type.STRUCT) {
      if (isKey) {
        throw new IllegalStateException("Key schemas are always unwrapped.");
      }
      
      throw new KsqlException("Schema returned from schema registry is anonymous type. "
          + "To use this schema with ksqlDB, set '" + CommonCreateConfigs.WRAP_SINGLE_VALUE
          + "=false' in the WITH clause properties.");
    }

    final Schema rowSchema = connectKsqlTranslator.apply(connectSchema);

    return rowSchema.fields().stream()
        .map(ConnectFormatSchemaTranslator::toColumn)
  }
```

ksqldb-engine/src/main/java/io/confluent/ksql/schema/ksql/inference/DefaultSchemaInjector.java
```diff
@@ -122,6 +123,7 @@ public DefaultSchemaInjector(final TopicSchemaSupplier schemaSupplier) {
         props.getKafkaTopic(),
         props.getKeySchemaId(),
         keyFormat,
+        SerdeFeatures.of(SerdeFeature.UNWRAP_SINGLES),
```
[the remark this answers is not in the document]
big-andy-coates: First off, you should only be setting this flag if the format supports unwrapping!  Or more correctly, if the format supports both unwrapping and wrapping.

As to hard coding it... humm, yes, so far it's been hard-coded in one place: `SerdeFeaturesFactory`.  

We could do something like:

```suggestion
        SerdeFeaturesFactory.buildInternal(FormatFactory.of(keyFormat),
```

But, to be honest, this is only a temporary thing.  At the moment, we only support unwrapped keys. Once we support something else we'll wire it up and this hardcoding will go.

big-andy-coates (approved):
LGTM

ksqldb-engine/src/main/java/io/confluent/ksql/schema/ksql/inference/DefaultSchemaInjector.java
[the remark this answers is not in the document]
vcrfxia: Good call on only setting the feature if it's supported.

I worry the naming of `buildInternal()` will be confusing here since it's not clear what internal topics have to do with schema registry. Is it worth introducing another alias for the method such as `setUnwrappingIfSupported()` or would that only make things worse?

ksqldb-serde/src/main/java/io/confluent/ksql/serde/connect/ConnectFormatSchemaTranslator.java
vcrfxia: Much cleaner! Thanks for the suggestion.

Wanted to check that we're OK removing the exception that's thrown if an unwrapped schema is encountered for a format that does not support unwrapping ([link](https://github.com/confluentinc/ksql/blob/a510ceef8dbad9be79f80bc9dff6e541ac580c5d/ksqldb-serde/src/main/java/io/confluent/ksql/serde/connect/ConnectFormatSchemaTranslator.java#L66-L70)) -- it makes sense to me that we're removing this since the only way we'd end up in this situation is if ksqlDB designated a format the supports unwrapping as not supporting it, but maybe it's better to be defensive and leave the check in?

vcrfxia:
Merging in order to unblock the other PR behind this one -- will address anything that comes out of the open discussions in a follow-up PR.

Issue #6133, opened 2020-09-02
Integrate code paths for schema compatibility checks for AVRO and other formats


big-andy-coates:
ksqlDB supports two new formats that integrate with the schema register: `JSON_SR` and `PROTOBUF`, yet it only checks that a new schema abides by the schema compatibility rules for `AVRO` formats.

Given QTT test cases added to `serdes.json`:

```json
[
  {
      "name": "serialization should fail if primitive schema not compatible with existing value schema - JSON_SR",
      "statements": [
        "CREATE STREAM INPUT (K STRING KEY, foo STRING) WITH (kafka_topic='input_topic', value_format='JSON_SR');",
        "CREATE STREAM OUTPUT WITH (wrap_single_value=false) AS SELECT * FROM INPUT;"
      ],
      "topics": [
        {"name": "OUTPUT", "schema": {"name": "KsqlRecord", "type": "record", "fields": [{"name": "FOO", "type": "string"}]}}
      ],
      "expectedException": {
        "type": "io.confluent.ksql.util.KsqlStatementException",
        "message": "Cannot register JSON schema for OUTPUT as the schema is incompatible with the current schema version registered for the topic."
      }
    },
    {
      "name": "serialization should fail if primitive schema not compatible with existing value schema - PROTOBUF",
      "statements": [
        "CREATE STREAM INPUT (K STRING KEY, foo STRING) WITH (kafka_topic='input_topic', value_format='PROTOBUF');",
        "CREATE STREAM OUTPUT WITH (wrap_single_value=false) AS SELECT * FROM INPUT;"
      ],
      "topics": [
        {"name": "OUTPUT", "schema": {"name": "KsqlRecord", "type": "record", "fields": [{"name": "FOO", "type": "string"}]}}
      ],
      "expectedException": {
        "type": "io.confluent.ksql.util.KsqlStatementException",
        "message": "Cannot register protobuf schema for OUTPUT as the schema is incompatible with the current schema version registered for the topic."
      }
    }
]
```

### Expected result:

Tests pass, i.e. invalid schema evolution is detected.,

### Actual result:

Tests fail, as no exception is thrown.

The implications of this would be that we could start producing data to a topic with a different schema to other producers, breaking downstream processes.

big-andy-coates:
This feels like a P0 to me, but leaving as `needs-triage` so it can be discussed.

agavra:
[...]

PR #6452, opened 2020-10-16
chore: remove unnecessary AvroUtil 🪓

### LINKED ISSUE #6394: [SR Support]: Support clean up of key schemas when topics are deleted
^ including for both `DROP STREAM/TABLE ... DELETE TOPIC` statements and during cluster termination, similar to what we do today for value schemas.

### LINKED ISSUE #6395: [SR Support]: Finish refactor to make SCHEMA_REGISTRY_VALUE_SUFFIX private
Context in https://github.com/confluentinc/ksql/pull/6380#discussion_r501392031: once the other bits of key schema support are in, we should finish refactoring uses of `SCHEMA_REGISTRY_VALUE_SUFFIX` to instead use the new util method (`KsqlConstants#getSRSubject()`). As part of this refactor, we should audit the remaining uses of `SCHEMA_REGISTRY_VALUE_SUFFIX` to ensure we haven't missed places where support for key schemas is not up to parity with that for value schemas.

agavra:
### Description 

Fixes #6394
Fixes #6395 
Fixes https://github.com/confluentinc/ksql/issues/6133  by removing `AvroUtil` 🪓 The class isn't necessary anymore because it was only used `EngineExecutor#buildPlan` when building a plan for `CREATE AS` statements. This is no longer necessary, because we always go through the `SchemaRegisterInjector` which ensures that it can register the schema for the `CREATE AS` statement.

### Testing done 

I tried creating a new unit test to prove that it works end-to-end, but there's a "bug" in `MockSchemaRegistryClient` in that it'll let you register incompatible schemas 😢 so I can't get it to work. Perhaps it's worth submitting a ticket to SR or just fixing it ourselves, but we have other unit test coverage (such as `SchemaRegistryInjectorTest`) and I confirmed it manually:
```
ksql> CREATE STREAM a (col1 INT, col2 INT) WITH (kafka_topic='a', value_format='avro', partitions=1);

 Message
----------------
 Stream created
----------------

ksql> CREATE STREAM b (col1 VARCHAR, col2 VARCHAR) WITH (kafka_topic='b', value_format='avro', partitions=1);

 Message
----------------
 Stream created
----------------

ksql> CREATE STREAM c WITH (kafka_topic='b') AS SELECT * FROM a;
Could not register schema for topic: Schema being registered is incompatible with an earlier schema; error code: 409
```

### Reviewer checklist


vcrfxia (approved):
LGTM!

PR #6466, opened 2020-10-20
test: more avro key testing & schema verification

agavra:
### Description 

And some more tests to AVRO keys and the ability to check the registered schema in the post-conditions check. Shockingly, nothing was broken...

### Testing done 

Test only change

### Reviewer checklist

### DIFF
--- a/ksqldb-functional-tests/src/main/java/io/confluent/ksql/test/model/PostConditionsNode.java
+++ b/ksqldb-functional-tests/src/main/java/io/confluent/ksql/test/model/PostConditionsNode.java
@@ -135,10 +135,8 @@ public final class PostConditionsNode {
       }
     }
 
-    // CHECKSTYLE_RULES.OFF: BooleanExpressionComplexity
     @SuppressWarnings("unchecked")
     Matcher<Iterable<PostTopicNode>> buildTopics() {
-      // CHECKSTYLE_RULES.ON: BooleanExpressionComplexity
       final Matcher<PostTopicNode>[] matchers = topics.stream()
           .map(topic -> new BaseMatcher<PostTopicNode>() {
             @Override
@@ -150,6 +148,7 @@ public final class PostConditionsNode {
               }
             }
 
+            // CHECKSTYLE_RULES.OFF: BooleanExpressionComplexity
             @Override
             public boolean matches(final Object item) {
               if (!(item instanceof PostTopicNode)) {
@@ -165,6 +164,7 @@ public final class PostConditionsNode {
                   && (topic.valueSchema instanceof NullNode
                           || topic.valueSchema.equals(that.valueSchema));
             }
+            // CHECKSTYLE_RULES.ON: BooleanExpessionComplexity
           })
           .toArray(Matcher[]::new);
 
--- a/ksqldb-functional-tests/src/main/java/io/confluent/ksql/test/model/PostConditionsNode.java
+++ b/ksqldb-functional-tests/src/main/java/io/confluent/ksql/test/model/PostConditionsNode.java
@@ -20,10 +20,14 @@ import static org.hamcrest.Matchers.hasItems;
 
 import com.fasterxml.jackson.annotation.JsonIgnoreProperties;
 import com.fasterxml.jackson.annotation.JsonProperty;
+import com.fasterxml.jackson.core.JsonProcessingException;
+import com.fasterxml.jackson.databind.JsonNode;
+import com.fasterxml.jackson.databind.node.NullNode;
 import com.google.common.collect.ImmutableList;
 import io.confluent.ksql.metastore.model.DataSource;
 import io.confluent.ksql.serde.KeyFormat;
 import io.confluent.ksql.serde.ValueFormat;
+import io.confluent.ksql.test.tools.TestJsonMapper;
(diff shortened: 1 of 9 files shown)


ksqldb-functional-tests/src/main/java/io/confluent/ksql/test/model/PostConditionsNode.java
```diff
@@ -133,7 +138,34 @@ Pattern buildBlackList() {
     @SuppressWarnings("unchecked")
     Matcher<Iterable<PostTopicNode>> buildTopics() {
       final Matcher<PostTopicNode>[] matchers = topics.stream()
-          .map(Matchers::is)
+          .map(topic -> new BaseMatcher<PostTopicNode>() {
```
vcrfxia, line 141: nit: can we move this matcher logic into `PostTopicNode`, so someone updating `PostTopicNode` is less likely to forget to update the matcher? The other QTT classes all follow a particular pattern with this, for example: https://github.com/confluentinc/ksql/blob/master/ksqldb-functional-tests/src/main/java/io/confluent/ksql/test/model/SourceNode.java#L114

ksqldb-functional-tests/src/main/java/io/confluent/ksql/test/tools/TestExecutor.java
```diff
@@ -229,6 +241,22 @@ public void buildAndExecuteQuery(
     }
   }
 
+  private static JsonNode fromSchemaMetadata(final Optional<SchemaMetadata> metadata) {
+    if (!metadata.isPresent()) {
+      return NullNode.getInstance();
+    }
+
+    if (ProtobufFormat.NAME.equals(metadata.get().getSchemaType())) {
```
vcrfxia, line 249: Can we add a comment explaining that we need this special case because protobuf schemas aren't valid JSON? If I hadn't coincidentally encountered this in SerdeUtil ([link](https://github.com/confluentinc/ksql/blob/807549c1b979d393137244c2586c575f59d6df1a/ksqldb-functional-tests/src/main/java/io/confluent/ksql/test/utils/SerdeUtil.java#L100)) the other day, I would've been confused by this.

Relatedly, it sucks that we have to duplicate this special-casing in multiple places... can't think of a good alternative at the moment though.

ksqldb-functional-tests/src/test/resources/query-validation-tests/as_value.json
```diff
@@ -149,6 +149,27 @@
           {"name": "OUTPUT", "type": "stream", "schema": "ID DECIMAL(4,2) KEY, ID_COPY DECIMAL(4,2), V1 INT"}
         ]
       }
+    },
+    {
+      "name": "STRUCT",
```
vcrfxia, line 154: Can we also add array and map for completeness? Same for the other tests (group by, partition by, etc). I don't think the coverage needs to be as thorough as what you've added for struct but I still think it'd be good to cover our bases.

ksqldb-functional-tests/src/test/resources/query-validation-tests/group-by.json
```diff
@@ -2272,6 +2313,277 @@
...
+      ],
+      "outputs": [
+        {"topic": "OUTPUT", "key": true, "value": {"COUNT": 1}},
+        {"topic": "OUTPUT", "key": false, "value": {"COUNT": 1}},
+        {"topic": "OUTPUT", "key": true, "value": {"COUNT": 2}}
+      ],
+      "post": {
+        "sources": [
+          {
+            "comment": "unwrap singles is implicitly set on AVRO key formats",
```
vcrfxia, line 2348: nit: unwrap singles is implicitly set on all key formats that support both wrapping and unwrapping, including AVRO. (I was confused for a moment when I read first this since I thought Avro was somehow special.)

ksqldb-functional-tests/src/test/resources/query-validation-tests/group-by.json
```diff
@@ -2272,6 +2313,277 @@
...
+      "outputs": [
+        {"topic": "OUTPUT", "key": {"F1": 1}, "value": {"SUM": 1}},
+        {"topic": "OUTPUT", "key": {"F1": 2}, "value": {"SUM": 3}}
+      ]
+    },
+    {
+      "properties": {
+        "ksql.key.format.enabled": "true"
+      },
+      "name": "Struct key used in aggregate expression",
```
vcrfxia, line 2574: This test looks identical to the one above it?

vcrfxia (approved):
LGTM! 🎉 Minor comments inline.

ksqldb-functional-tests/src/main/java/io/confluent/ksql/test/tools/TestExecutor.java
agavra: I was originally going to refactor all of it to just be a string, but then that made the QTT tests really annoying to write (escaped JSON inside JSON strings really sucks)

ksqldb-functional-tests/src/test/resources/query-validation-tests/group-by.json
agavra: eh, I forgot to remove the one above it (which actually fails - the data is different)

PR #6468, opened 2020-10-20
chore: add more aggregated rocksdb metrics

rodesai:
### Description 
adds more rocksdb metrics. they are free to add, we only pay the cost of collecting them if we query
jmx. Plus, they are pretty aggressively throttled, so collection should be very infrequent.

### Testing done 
tested on a local ksql server


spena (approved):
lgtm

PR #6474, opened 2020-10-20
feat: new syntax to interact with session variables (define/undefine/show variables)

spena:
### Description 
This is part of [KLIP-38 - Variable Substitution](https://github.com/confluentinc/ksql/blob/master/design-proposals/klip-38-variable-substitution.md
)
Variable substitution will have a follow-up PR.

It adds a syntax to interact with session variables:
- DEFINE <name> = '<value>';
- UNDEFINE <name>;
- (LIST | SHOW) VARIABLES
^ The KLIP proposed the use of `DEFINE` to list values. But I saw the `SHOW` more appropriate as we have SHOW and LIST syntax for other entities. I will change the KLIP after variable substitution is merged.

The syntax is added on the server-side and CLI-side. 

For the server-side, the variable scope is seen per-request. i.e. 
```
Request 1 will return the variables `x` and `y`.
{ 
  statement: "
    DEFINE x = '1';
    DEFINE y = '2';
    SHOW VARIABLES;
  "
}

Request 2 will return the variables `z`.
{ 
  statement: "
    DEFINE z = '3';
    SHOW VARIABLES;
  "
}
```

For the CLI-side, the variable scope is during the session of the CLI. i.e.
```
ksql> DEFINE x = '1';
ksql> DEFINE y = '2';
ksql> SHOW VARIABLES;
 Variable Name | Value 
-----------------------
 x             | 1     
 y             | 2     
-----------------------
ksql> DEFINE z = '3';
ksql> SHOW VARIABLES;
 Variable Name | Value 
-----------------------
 x             | 1     
 y             | 2     
 z             | 3     
-----------------------
```

### Testing done 
Added unit tests
Verified manually (see above examples)

### Reviewer checklist

### DIFF
--- a/ksqldb-parser/src/main/antlr4/io/confluent/ksql/parser/SqlBase.g4
+++ b/ksqldb-parser/src/main/antlr4/io/confluent/ksql/parser/SqlBase.g4
@@ -53,6 +53,7 @@ statement
     | (LIST | SHOW) FUNCTIONS                                               #listFunctions
     | (LIST | SHOW) (SOURCE | SINK)? CONNECTORS                             #listConnectors
     | (LIST | SHOW) TYPES                                                   #listTypes
+    | (LIST | SHOW) VARIABLES                                               #listVariables
     | DESCRIBE EXTENDED? sourceName                                         #showColumns
     | DESCRIBE FUNCTION identifier                                          #describeFunction
     | DESCRIBE CONNECTOR identifier                                         #describeConnector
@@ -517,6 +518,7 @@ REPLACE: 'REPLACE';
 ASSERT: 'ASSERT';
 ADD: 'ADD';
 ALTER: 'ALTER';
+VARIABLES: 'VARIABLES';
 
 IF: 'IF';
 
--- a/ksqldb-parser/src/main/antlr4/io/confluent/ksql/parser/SqlBase.g4
+++ b/ksqldb-parser/src/main/antlr4/io/confluent/ksql/parser/SqlBase.g4
@@ -62,6 +62,8 @@ statement
     | TERMINATE ALL                                                         #terminateQuery
     | SET STRING EQ STRING                                                  #setProperty
     | UNSET STRING                                                          #unsetProperty
+    | DEFINE variableName EQ variableValue                                  #defineVariable
+    | UNDEFINE variableName                                                 #undefineVariable
     | CREATE (OR REPLACE)? STREAM (IF NOT EXISTS)? sourceName
                 (tableElements)?
                 (WITH tableProperties)?                                     #createStream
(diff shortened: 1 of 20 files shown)


ksqldb-parser/src/main/antlr4/io/confluent/ksql/parser/SqlBase.g4
```diff
@@ -566,6 +581,10 @@ TIMESTAMP_WITH_TIME_ZONE
     : 'TIMESTAMP' WS 'WITH' WS 'TIME' WS 'ZONE'
     ;
 
+VARIABLE
```
agavra, line 584: is this used anywhere for this PR?

ksqldb-rest-app/src/test/java/io/confluent/ksql/rest/server/execution/ListVariablesExecutorTest.java
```diff
@@ -0,0 +1,85 @@
...
+import org.junit.Before;
+import org.junit.Rule;
+import org.junit.Test;
+import org.junit.runner.RunWith;
+import org.mockito.junit.MockitoJUnitRunner;
+
+@RunWith(MockitoJUnitRunner.class)
+public class ListVariablesExecutorTest {
+  @Rule
+  public final TemporaryEngine engine = new TemporaryEngine();
```
agavra, line 40: do you need to actually run through configure etc... or can you mock out the configured statements?

ksqldb-rest-app/src/test/java/io/confluent/ksql/rest/server/execution/VariableExecutorTest.java
```diff
@@ -0,0 +1,168 @@
...
+    // When:
+    final KsqlEntity response = executeUndefineVariable("UNDEFINE var1;").get();
+
+    // Then:
+    assertThat(((WarningEntity)response).getMessage(),
+        containsString("Cannot undefine variable 'var1' which was never defined"));
+  }
+
+  @Test
+  public void shouldThrowOnInvalidValues() {
```
agavra, line 148: should we add a test on adding duplicates? we should also check to make sure that case insensitive duplicates are stilled rejected

agavra (approved):
LGTM!

ksqldb-rest-app/src/test/java/io/confluent/ksql/rest/server/execution/ListVariablesExecutorTest.java
spena: I copied most of the code from ListTopicsExecutorTest and ListPropertiesExecutorTest. But you're right, we don't need this engine. I'll change it with a mock.

ksqldb-rest-app/src/test/java/io/confluent/ksql/rest/server/execution/VariableExecutorTest.java
spena: Variables will be overridden (no rejected) if an existing case insensitive variable exists. I'll add a test case that verifies that. 

It will allow us to do this:
```
DEFINE var1 = '1';
DEFINE VAR1 = '2';
DEFINE vAr1 = '3'; // latest update
```

Issue #6479, opened 2020-10-20
Internal Server Error for /healthcheck endpoint in RBAC-enabled ksqlDB cluster


spena:
**Describe the bug**
Internal server error (500) occurs when accessing /healthcheck endpoint in RBAC-enabled ksqlDB 6.0.0 cluster. Error occurs with or without valid credentials (this is supposed to be an open, non-authorized endpoint). Other endpoints like /info work fine with the same credentials.

Exception in ksqlDB server looks like:
```
[2020-10-20 14:56:03,341] ERROR Failed to handle request -1 /healthcheck (io.confluent.ksql.api.server.FailureHandler:38)
io.confluent.ksql.util.KsqlException: Invalid user principal found. Got: null Expected: JwtPrincipal.
	at io.confluent.ksql.security.utils.KsqlSecurityUtils.toJwtPrincipal(KsqlSecurityUtils.java:27)
	at io.confluent.ksql.security.KsqlUserContextProviderImpl.getKafkaClientSupplier(KsqlUserContextProviderImpl.java:41)
	at io.confluent.ksql.api.impl.DefaultKsqlSecurityContextProvider.lambda$provide$0(DefaultKsqlSecurityContextProvider.java:74)
	at java.base/java.util.Optional.map(Optional.java:265)
	at io.confluent.ksql.api.impl.DefaultKsqlSecurityContextProvider.provide(DefaultKsqlSecurityContextProvider.java:69)
	at io.confluent.ksql.rest.server.KsqlServerEndpoints.executeOldApiEndpoint(KsqlServerEndpoints.java:295)
	at io.confluent.ksql.rest.server.KsqlServerEndpoints.executeCheckHealth(KsqlServerEndpoints.java:230)
	at io.confluent.ksql.api.server.ServerVerticle.lambda$handleHealthcheckRequest$12(ServerVerticle.java:289)
	at io.confluent.ksql.api.server.OldApiUtils.handleOldApiRequest(OldApiUtils.java:69)
	at io.confluent.ksql.api.server.ServerVerticle.handleHealthcheckRequest(ServerVerticle.java:287)
```

**To Reproduce**
Issue can be reproduced in cp-demo (6.0.0-post branch), executing the following curl command:
```
curl -u ksqlDBAdmin:ksqlDBAdmin http://localhost:8088/healthcheck
```

**Expected behavior**
Curl should return a response from the healthcheck endpoint without errors.

**Actual behaviour**
Curl is failing with the output:
```
[...]

PR #6482, opened 2020-10-21
fix: Internal Server Error for /healthcheck endpoint in RBAC-enabled

spena:
### Description 
Fixes https://github.com/confluentinc/ksql/issues/6479

Context:
There are endpoints in `KsqlServerEndpoints` that do not require a security context initialized. Some of these endpoints do not require authentication, so a security context should not be initialized because there is no `Principal` authenticated. 

The current Vert.x code in `KsqlServerEndpoints` attempted to initialize the security context on endpoints where the user was not authenticated. This caused an error when calling these endpoints even with valid credentials. The fix was just to prevent initializing the security contexts, which is unnecessary.

### Testing done 
Added unit tests.
Verified manually with Confluent RBAC library.

* `User:ksql` has valid credentials and has authorization to the server.
* `User:user1` has valid credentials but not authorization to the server.

1. Endpoints that required authentication only 
```
curl   http://localhost:8088/info | jq .
{
  "@type": "generic_error",
  "error_code": 40100,
  "message": "Failed authentication"
}

curl -u user1:user1  http://localhost:8088/info | jq .
{
  "KsqlServerInfo": {
    "version": "6.1.0-0",
    "kafkaClusterId": "Bj_7MIRSQ36T97mtEimh-A",
    "ksqlServiceId": "default_",
    "serverStatus": "RUNNING"
  }
}
```

2. Endpoints that do not require authentication
```
 curl http://localhost:8088/v1/metadata/id 2>/dev/null | jq .
{
  "scope": {
    "path": [],
    "clusters": {
      "kafka-cluster": "Bj_7MIRSQ36T97mtEimh-A",
      "ksql-cluster": "default_"
    }
  },
  "id": ""
}
```

3. Endpoints that require authentication and authorization
```
$ curl http://localhost:8088/info 2>/dev/null | jq .
{
  "@type": "generic_error",
  "error_code": 40100,
  "message": "Failed authentication"
}

$ curl -u user1:user1 http://localhost:8088/status 2>/dev/null | jq .
{
  "@type": "generic_error",
  "error_code": 40300,
  "message": "You are forbidden from using this cluster."
}

$ curl -u ksql:ksql http://localhost:8088/status 2>/dev/null | jq .
{
  "commandStatuses": {
    "stream/`KSQL_PROCESSING_LOG`/create": "SUCCESS"
  }
}
```

### Reviewer checklist

### DIFF
--- a/ksqldb-rest-app/src/main/java/io/confluent/ksql/api/auth/KsqlAuthorizationProviderHandler.java
+++ b/ksqldb-rest-app/src/main/java/io/confluent/ksql/api/auth/KsqlAuthorizationProviderHandler.java
@@ -15,9 +15,9 @@
 
 package io.confluent.ksql.api.auth;
 
+import static io.confluent.ksql.api.auth.AuthenticationPluginHandler.KSQL_AUTHENTICATION_SKIP_PATHS;
 import static io.netty.handler.codec.http.HttpResponseStatus.FORBIDDEN;
 
-import com.google.common.collect.ImmutableSet;
 import io.confluent.ksql.security.KsqlAuthorizationProvider;
 import io.vertx.core.AsyncResult;
 import io.vertx.core.Handler;
@@ -25,15 +25,11 @@ import io.vertx.core.Promise;
 import io.vertx.core.WorkerExecutor;
 import io.vertx.ext.auth.User;
 import io.vertx.ext.web.RoutingContext;
-import java.util.Set;
 
 /**
  * Handler that calls a KsqlAuthorizationProvider plugin that can be used for custom authorization
  */
 public class KsqlAuthorizationProviderHandler implements Handler<RoutingContext> {
-
-  public static final Set<String> KSQL_AUTHENTICATION_SKIP_PATHS = ImmutableSet
-      .of("/v1/metadata", "/v1/metadata/id", "/healthcheck");
 
   private final WorkerExecutor workerExecutor;
   private final KsqlAuthorizationProvider ksqlAuthorizationProvider;
--- a/ksqldb-rest-app/src/main/java/io/confluent/ksql/api/auth/KsqlAuthorizationProviderHandler.java
+++ b/ksqldb-rest-app/src/main/java/io/confluent/ksql/api/auth/KsqlAuthorizationProviderHandler.java
@@ -32,7 +32,7 @@ import java.util.Set;
  */
 public class KsqlAuthorizationProviderHandler implements Handler<RoutingContext> {
 
-  public static final Set<String> PATHS_WITHOUT_AUTHORIZATION = ImmutableSet
+  public static final Set<String> KSQL_AUTHENTICATION_SKIP_PATHS = ImmutableSet
       .of("/v1/metadata", "/v1/metadata/id", "/healthcheck");
 
   private final WorkerExecutor workerExecutor;
@@ -49,7 +49,7 @@ public class KsqlAuthorizationProviderHandler implements Handler<RoutingContext>
 
     final String path = routingContext.normalisedPath();
 
-    if (PATHS_WITHOUT_AUTHORIZATION.contains(path)) {
+    if (KSQL_AUTHENTICATION_SKIP_PATHS.contains(path)) {
       routingContext.next();
       return;
     }
--- a/ksqldb-rest-app/src/main/java/io/confluent/ksql/api/impl/DefaultKsqlSecurityContextProvider.java
+++ b/ksqldb-rest-app/src/main/java/io/confluent/ksql/api/impl/DefaultKsqlSecurityContextProvider.java
@@ -57,11 +57,12 @@ public class DefaultKsqlSecurityContextProvider implements KsqlSecurityContextPr
     final Optional<Principal> principal = apiSecurityContext.getPrincipal();
     final Optional<String> authHeader = apiSecurityContext.getAuthToken();
 
-    // A user context is not necessary if a user context provider is not present or
-    // the user principal is missing. The missing principal was already validated by the
-    // Authentication and Authorization plugins before calling this method, which means the
-    // missing principal is a valid connection. For those cases, we create a default service
-    // context that the missing user can use.
+    // A user context is not necessary if a user context provider is not present or the user
+    // principal is missing. If a failed authentication attempt results in a missing principle,
+    // then the authentication plugin will have already failed the connection before calling
+    // this method. Therefore, if we've reached this method with a missing principle, then this
+    // must be a valid connection that does not require authentication.
+    // For these cases, we create a default service context that the missing user can use.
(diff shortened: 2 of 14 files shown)


ksqldb-rest-app/src/test/java/io/confluent/ksql/api/AuthTest.java
```diff
@@ -539,7 +612,7 @@ private void shouldNotAllowAccessIfPermissionCheckThrowsException(
     void run() throws Exception;
   }
 
-  private static class StringPrincipal implements Principal {
+  public static class StringPrincipal implements Principal {
```
vcrfxia, line 615: Why can't this still be private?

ksqldb-rest-app/src/main/java/io/confluent/ksql/api/auth/KsqlAuthorizationProviderHandler.java
```diff
@@ -32,7 +32,7 @@
  */
 public class KsqlAuthorizationProviderHandler implements Handler<RoutingContext> {
 
-  public static final Set<String> PATHS_WITHOUT_AUTHORIZATION = ImmutableSet
+  public static final Set<String> KSQL_AUTHENTICATION_SKIP_PATHS = ImmutableSet
```
vcrfxia, line 35: Can we move this into `AuthenticationPluginHandler` if we're going to rename it? It looks out of place here.

ksqldb-rest-app/src/test/java/io/confluent/ksql/api/AuthTest.java
```diff
@@ -137,6 +142,11 @@ public void close() {
     server.start();
   }
 
+  @Override
+  protected HttpResponse<Buffer> sendGetRequest(final String uri) throws Exception {
+    return sendGetRequestWithCreds(client, uri, USER_WITH_ACCESS, USER_WITH_ACCESS_PWD);
+  }
+
   @Override
   protected HttpResponse<Buffer> sendRequest(final WebClient client, final String uri,
```
vcrfxia, line 151: nit: can we rename this to `sendPostRequest()` in light of the new method above?

ksqldb-rest-app/src/test/java/io/confluent/ksql/api/AuthTest.java
```diff
@@ -415,6 +436,22 @@ private void shouldFailInsertRequest(final String username, final String passwor
...
+    VertxCompletableFuture<HttpResponse<Buffer>> requestFuture = new VertxCompletableFuture<>();
+    HttpRequest<Buffer> request = client.get(uri);
+    if (username != null || password != null) {
+      request = request.basicAuthentication(username, password);
+    }
+    request.send(requestFuture);
+    return requestFuture.get();
+  }
+
   private HttpResponse<Buffer> sendRequestWithCreds(
```
vcrfxia, line 455: nit: can we rename this to `sendPostRequestWithCreds()` in light of the new method above?

vcrfxia (commented):
Hey @spena , I don't think this is the right way to address this bug. This PR fixes the issue for the default unauthenticated paths, but it doesn't solve the problem more generally. For example, if a user chooses to make the `/ksql` endpoint unauthenticated (not sure why anyone would do that but bear with me 😆), then they would again hit the bug in the Github issue. I'd need to dig into the code more in order to figure out the proper way to fix this. It might involve changes around ApiSecurityContext. Ping me tomorrow if you want to brainstorm together offline?

ksqldb-rest-app/src/test/java/io/confluent/ksql/api/AuthTest.java
spena: Done

ksqldb-rest-app/src/test/java/io/confluent/ksql/api/AuthTest.java
spena: done

ksqldb-rest-app/src/main/java/io/confluent/ksql/api/auth/KsqlAuthorizationProviderHandler.java
spena: done

ksqldb-rest-app/src/test/java/io/confluent/ksql/api/AuthTest.java
spena: Left from another approach I was taking. Fixed.

spena:
@vcrfxia You're right. I was expecting that whenever an endpoint required a service context, then the authentication would require a user to create one. It's just lucky that we don't have an endpoint that requires a service context with no authentication.

I fixed the issue, and now I create a default service context when there is a missing user. I left a comment in the code about why there's no problem about it. 

I added tests to the DefaultKsqlServiceContextProvider, I copied the tests we had for this class (https://github.com/confluentinc/ksql/blob/5.5.x/ksqldb-rest-app/src/test/java/io/confluent/ksql/rest/server/context/KsqlSecurityContextBinderFactoryTest.java), and added just one when the user missing. 

I also noticed that the `AuthTest` just tests Authentication and Authorization. It does not test the path when building the user context. So I removed the unnecessary code, and rely now On the `DefaultKsqlServiceContextProviderTest` to return the right service context without failing.

I verified the endpoints work manually again.

ksqldb-rest-app/src/main/java/io/confluent/ksql/api/impl/DefaultKsqlSecurityContextProvider.java
```diff
@@ -57,7 +57,17 @@ public KsqlSecurityContext provide(final ApiSecurityContext apiSecurityContext)
     final Optional<Principal> principal = apiSecurityContext.getPrincipal();
     final Optional<String> authHeader = apiSecurityContext.getAuthToken();
 
-    if (securityExtension == null || !securityExtension.getUserContextProvider().isPresent()) {
+    // A user context is not necessary if a user context provider is not present or
+    // the user principal is missing. The missing principal was already validated by the
+    // Authentication and Authorization plugins before calling this method, which means the
+    // missing principal is a valid connection. For those cases, we create a default service
+    // context that the missing user can use.
```
vcrfxia, line 64:
```suggestion
    // A user context is not necessary if a user context provider is not present or
    // the user principal is missing. If a failed authentication attempt results in a missing principle,
    // then the authentication plugin will have already failed the connection before calling this method.
    // Therefore, if we've reached this method with a missing principle, then this must be a valid connection that does not require authentication. 
    // For these cases, we create a default service
    // context that the missing user can use.
```

vcrfxia (approved):
Thanks @spena -- LGTM with a suggested update to clarify the missing user case.

PR #6485, opened 2020-10-22
refactor: add engine-common module + move SqlValueCoercer

big-andy-coates:
### Description 

Adds a `ksqldb-engine-common` module for... well, common engine/server side stuff.  Historically, too much server side stuff gets put in `ksqldb-common`, which is also used by the CLI. This blurred the server/client lines.

Moved `DefaultSqlValuesCoercer` from `ksqldb-parser` to `ksqldb-engine-common` so that its accessible within `ksqldb-execution`.

Split `DefaultSqlValuesCoercer` into two impls of `SqlValueCoercer`. The default one that's used internally and a new API specific one.

### Testing done 

Usual

### Reviewer checklist


vcrfxia (commented):
The changes LGTM though I'm curious to hear more about why it's preferable to split DefaultSqlValueCoercer into two classes and duplicate code as a result. I've been going back and forth on whether I think we're better off separating them or not.

It's nice that ApiSqlValueCoercer defers a bunch of functionality to DefaultSqlValueCoercer. This means that if a new datatype (that doesn't require special API handling) is added, then the change still only needs to be made in one place. OTOH, there's nothing to validate that target datatypes that are supported by one coercer are also supported by the other, so someone could unwittingly add support for a new datatype in  DefaultSqlValueCoercer and not realize it's broken in ApiSqlValueCoercer since they wouldn't know to copy the new tests from DefaultSqlValueCoercer to ApiSqlValueCoercer.

A related issue is that there's a good chunk of duplicated code between the two. If someone finds a bug in one implementation (e.g., in struct coercion), the bug is almost certainly also present in the other implementation, and they won't know to fix it in the other location. 

All said, I'm not convinced this refactor is an improvement. What am I missing?

big-andy-coates:
>  I'm curious to hear more about why it's preferable to split DefaultSqlValueCoercer into two classes

`DefaultSqlValueCoercer` previously didn't need to worry about vertex types.  Where it was being used it _should never encounter them_.   Internally, we use `Struct`, `List` and `Map` for `STRUCT`, `ARRAY` and `MAP`.

We then needed a coercer that can be used for the API, where `STRUCT`, `ARRAY` and `MAP` use `JsonObject`, `JsonArray` and `JsonObject`.  In the api case it _should never encounter_ `Struct, `List` and `Map`.

So we do not need a class that can handle both.  Handling both in one class is more complex.  More than that, it may allow funky unintentional uses, e.g. a case that mixes the two sets of types. It makes the code harder to read. People reading the code may incorrectly assume it needs to handle the mixed case, which is doesn't.

It's better to have more targeted implementations.  

> duplicate code as a result.

There is no code duplication. The api coercer delegates any non-complex type coercion to the default one. The only thing that is duplicated, if you like, is the pattern of iterating over fields/entries/elements when converting one complex type to another - and I see no issue with that.

> someone could unwittingly add support for a new datatype in DefaultSqlValueCoercer and not realize it's broken in ApiSqlValueCoercer since they wouldn't know to copy the new tests from DefaultSqlValueCoercer to ApiSqlValueCoercer.

When I copied `DefaultSqlValueCoercerTest` to `ApiSqlValueCoercerTest` I was being lazy. Ideally, the api test would inject the default coercer, and the test would only test the specific types the api coercer handles and test everything else is delegated to the default coercer.

If someone comes along an adds a new type, then there are many places in the code that need to be updated.  If you miss just one place then things won't work.  It's the job of the person adding the new type to ensure all the necessary places are updated. That's not a valid argument for lumping these two use-cases into a single class, IMHO.

I tend to add meta tests: tests that fail if the test doesn't cover a specific type, to handle the case of 'what happens if someone adds a new type'.

> A related issue is that there's a good chunk of duplicated code between the two. If someone finds a bug in one implementation

I'd disagree there's a bunch of duplication of code.  There's a _similar_ looping around all the fields/elements/entries.  But they're different.  Lots of code does iteration, we don't say they're all duplicates!

Sure, we could move the iteration pattern into a helper that's shared.  That's valid. But... what's the ROI? Are we really expecting a bug in the iteration pattern?  Much more likely there's a bug elsewhere.

vcrfxia (approved):
As discussed offline, I'm still not totally sold on this, particularly because of the duplicate tests and the fact that ApiSqlValueCoercer isn't discoverable from DefualtSqlValueCoercer. However, I might be leaning towards your side of the fence and you seem to feel strongly about this so let's go for it.

PR #6486, opened 2020-10-22
test: fix QTT tests after upstream change

big-andy-coates:
### Description 

There seems to have been an upstream change that means intermediate output is no longer being suppressed.  This is only affecting a single QTT test, (and its historic versions) The test was previously:

```json
{
      "name": "collect_list string table",
      "format": ["AVRO", "JSON", "PROTOBUF"],
      "statements": [
        "CREATE TABLE TEST (ID BIGINT PRIMARY KEY, VALUE varchar) WITH (kafka_topic='test_topic', value_format='{FORMAT}');",
        "CREATE TABLE S2 as SELECT ID, collect_list(value) as collected FROM test group by id;"
      ],
      "inputs": [
        {"topic": "test_topic", "key": 0,"value": {"VALUE": "foo"}},
        {"topic": "test_topic", "key": 100,"value": {"VALUE": "baz"}},
        {"topic": "test_topic", "key": 0,"value": {"VALUE": "bar"}},
        {"topic": "test_topic", "key": 100,"value": {"VALUE": "baz"}},
        {"topic": "test_topic", "key": 100,"value": {"VALUE": "foo"}}
      ],
      "outputs": [
        {"topic": "S2", "key": 0,"value": {"COLLECTED": ["foo"]}},
        {"topic": "S2", "key": 100,"value": {"COLLECTED": ["baz"]}},
        {"topic": "S2", "key": 0,"value": {"COLLECTED": []}},
        {"topic": "S2", "key": 0,"value": {"COLLECTED": ["bar"]}},
        {"topic": "S2", "key": 100,"value": {"COLLECTED": []}},
        {"topic": "S2", "key": 100,"value": {"COLLECTED": ["foo"]}}
      ]
    }
```

Where as now it needs to be:

```json
{
      "name": "collect_list string table",
      "format": ["AVRO", "JSON", "PROTOBUF"],
      "statements": [
        "CREATE TABLE TEST (ID BIGINT PRIMARY KEY, VALUE varchar) WITH (kafka_topic='test_topic', value_format='{FORMAT}');",
        "CREATE TABLE S2 as SELECT ID, collect_list(value) as collected FROM test group by id;"
      ],
      "inputs": [
        {"topic": "test_topic", "key": 0,"value": {"VALUE": "foo"}},
        {"topic": "test_topic", "key": 100,"value": {"VALUE": "baz"}},
        {"topic": "test_topic", "key": 0,"value": {"VALUE": "bar"}},
        {"topic": "test_topic", "key": 100,"value": {"VALUE": "baz"}},
        {"topic": "test_topic", "key": 100,"value": {"VALUE": "foo"}}
      ],
      "outputs": [
        {"topic": "S2", "key": 0,"value": {"COLLECTED": ["foo"]}},
        {"topic": "S2", "key": 100,"value": {"COLLECTED": ["baz"]}},
        {"topic": "S2", "key": 0,"value": {"COLLECTED": []}},
        {"topic": "S2", "key": 0,"value": {"COLLECTED": ["bar"]}},
        {"topic": "S2", "key": 100,"value": {"COLLECTED": []}},            <-- new output
        {"topic": "S2", "key": 100,"value": {"COLLECTED": ["baz"]}},  <-- new output
        {"topic": "S2", "key": 100,"value": {"COLLECTED": []}},
        {"topic": "S2", "key": 100,"value": {"COLLECTED": ["foo"]}}
      ]
    }
```

Question is... is this additional output expected?   cc @abbccdda

### DIFF
--- a/ksqldb-functional-tests/src/test/resources/historical_plans/collect-list_-_collect_list_string_table_-_AVRO/6.0.0_1594233263741/spec.json
+++ b/ksqldb-functional-tests/src/test/resources/historical_plans/collect-list_-_collect_list_string_table_-_AVRO/6.0.0_1594233263741/spec.json
@@ -107,6 +107,18 @@
       "topic" : "S2",
       "key" : 100,
       "value" : {
+        "COLLECTED" : [ "baz" ]
+      }
+    }, {
+      "topic" : "S2",
+      "key" : 100,
+      "value" : {
+        "COLLECTED" : [ ]
+      }
+    }, {
+      "topic" : "S2",
+      "key" : 100,
+      "value" : {
         "COLLECTED" : [ "foo" ]
       }
     } ],
--- a/ksqldb-functional-tests/src/test/resources/historical_plans/collect-list_-_collect_list_string_table_-_AVRO/6.1.0_1594164260142/spec.json
+++ b/ksqldb-functional-tests/src/test/resources/historical_plans/collect-list_-_collect_list_string_table_-_AVRO/6.1.0_1594164260142/spec.json
@@ -107,6 +107,18 @@
       "topic" : "S2",
       "key" : 100,
       "value" : {
+        "COLLECTED" : [ "baz" ]
+      }
+    }, {
+      "topic" : "S2",
+      "key" : 100,
+      "value" : {
+        "COLLECTED" : [ ]
+      }
+    }, {
+      "topic" : "S2",
+      "key" : 100,
+      "value" : {
         "COLLECTED" : [ "foo" ]
       }
     } ],
--- a/ksqldb-functional-tests/src/test/resources/historical_plans/collect-list_-_collect_list_string_table_-_JSON/6.0.0_1594233263917/spec.json
+++ b/ksqldb-functional-tests/src/test/resources/historical_plans/collect-list_-_collect_list_string_table_-_JSON/6.0.0_1594233263917/spec.json
@@ -107,6 +107,18 @@
       "topic" : "S2",
       "key" : 100,
       "value" : {
+        "COLLECTED" : [ "baz" ]
+      }
+    }, {
+      "topic" : "S2",
+      "key" : 100,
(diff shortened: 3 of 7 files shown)


spena (approved):
lgtm

big-andy-coates:
Merging in an effort to get the build green. Question around whether this is intentional or not moved to https://github.com/confluentinc/ksql/issues/6496.

mjsax:
I am actually wondering about the original output. Maybe something with caching changes? The new output seems to be the expected one if caching is disabled? Or could it be related to avoid "idempotent updates" \cc @vvcephei -- not sure to what extend this feature is implemented already?

vvcephei:
Hmm, yeah, that new output doesn't look expected to me. I don't think it's a result of the indempotent thing, which has been in the codebase for months already.

It's probably worth actually debugging it. Especially considering it was worth @big-andy-coates 's time to avoid spurious intermediate updates for filters, it seems like just accepting these apparently spurious intermediate results seems suspicious.

mjsax:
@vvcephei -- Thanks. Just wanted to double check.

However, with caching disabled (I assume that is the case for this test) the new output seems to be correct. Not sure why the old output was "missing" those two records. Might be worth to follow up for sure.

@big-andy-coates Do the tests run with caching enabled or disabled? I would assume disabled? With caching the output should be just 2 records (one per key)?

big-andy-coates:
No caching. Tests run with `cache.max.bytes.buffering` set to `0`.

I'd agree the new output seems more correct than the old output. 

But I'm left wondering what changed!

blueedgenick:
On a slightly different note, it seems we only caught this behavior change "by accident" due to the way the input data for this particular udaf test was constructed. Presumably the are many other udaf tests which pass just fine. Given that the main point of such tests is generally the "functional" shifts if each udaf that may be ok but perhaps we should have a test suite to test this kind of behavior for aggregations in the general sense?

PR #6487, opened 2020-10-22
test: topology change: state store dropped

big-andy-coates:
### Description 

The upstream AK commit https://github.com/apache/kafka/pull/9156 seems to have had an unexpected side affect, which is causing QTT to fail because the topology has changed.

With the new streams snapshot the topology is no longer materializing the `KTABLE-SOURCE-000000000X` nodes.  This looks to be a good thing, as the next node is also materialized, making the first materialization redundant.

I don't think this will cause any backwards compatibility issues, as I don't believe the dropped state store was adding any values, i.e. it was redundant.  

I don't believe there is any chance of dropped messages when users upgrade, because, as I understand it, offsets would not be committed until all the nodes in the sub-topology have processed a specific input record. Hence its not possible for only the first statestore to have a message and for the offsets to have been committed, meaning the message would be missed on an upgrade. Either both statestores have the message and offsets are committed, or the message is only partially processed and offsets are not committed.

However, asking for confirmation that this change is backwards compatible and no messages can get dropped cc @abbccdda, @mjsax ?


spena (approved):
LGTM from the ksql side. I see you just removed the `STATE-STORE-` references in all topology files. If the streams team is good with this, then merge it.

big-andy-coates:
Merging to fix the build. Open questions moved to https://github.com/confluentinc/ksql/issues/6497.

mjsax:
@big-andy-coates Your reasoning sounds correct to me.

Issue #6406, opened 2020-10-12
Blocker: compatibility break in joins


big-andy-coates:
## Description

AK commit apache/kafka#9156  introduced a compatibility breaking change: it's causing table source nodes to be materialized into a state store, even through the later mapValues node is already materialized.  This MUST be fixed before we release off of master.

The following historic tests have been disabled:

ksqldb-functional-tests/src/test/resources/historical_plans/joins_-_table_table_join_with_where_clause/6.0.0_1594233291874
ksqldb-functional-tests/src/test/resources/historical_plans/joins_-_table_table_join_with_where_clause/6.1.0_1594164287151

The topology has changed due to the Apache Kafka change:

old topology
```
Topologies:
   Sub-topology: 0
    Source: KSTREAM-SOURCE-0000000001 (topics: [left_topic])
      --> KTABLE-SOURCE-0000000002
    Source: KSTREAM-SOURCE-0000000007 (topics: [right_topic])
      --> KTABLE-SOURCE-0000000008
    Processor: KTABLE-SOURCE-0000000002 (stores: [])
      --> KTABLE-MAPVALUES-0000000003
      <-- KSTREAM-SOURCE-0000000001
    Processor: KTABLE-SOURCE-0000000008 (stores: [])
      --> KTABLE-MAPVALUES-0000000009
      <-- KSTREAM-SOURCE-0000000007
    Processor: KTABLE-MAPVALUES-0000000003 (stores: [KafkaTopic_Left-Reduce])
      --> KTABLE-TRANSFORMVALUES-0000000004
      <-- KTABLE-SOURCE-0000000002
    Processor: KTABLE-MAPVALUES-0000000009 (stores: [KafkaTopic_Right-Reduce])
      --> KTABLE-TRANSFORMVALUES-0000000010
      <-- KTABLE-SOURCE-0000000008
    Processor: KTABLE-TRANSFORMVALUES-0000000004 (stores: [])
      --> PrependAliasLeft
      <-- KTABLE-MAPVALUES-0000000003
    Processor: KTABLE-TRANSFORMVALUES-0000000010 (stores: [])
      --> PrependAliasRight
      <-- KTABLE-MAPVALUES-0000000009
    Processor: PrependAliasLeft (stores: [])
      --> KTABLE-JOINTHIS-0000000013
      <-- KTABLE-TRANSFORMVALUES-0000000004
    Processor: PrependAliasRight (stores: [])
      --> KTABLE-JOINOTHER-0000000014
[...]

PR #6488, opened 2020-10-22
test: re-enable historic qtt tests

big-andy-coates:
### Description 

fixes: https://github.com/confluentinc/ksql/issues/6406

The upstream fix is now available, so these tests can be re-enabled.


spena (approved):
lgtm

PR #6490, opened 2020-10-22
test: stop (r)qtt locking up

big-andy-coates:
### Description 

Fixes an issue where running QTT can cause IntelliJ and even MacOS to lockup for some period at the end of a test run.

I've found that IntelliJ can lock up for over a minute after running all QTT tests.  This can sometimes cause MacOS to also become unresponsive.

Well, I've finally tracked down the cause! The Confluent `TestUtils` class the code used to create temporary directories registers a shutdown hook for each temp directory created to ensure the directory is deleted on shutdown. (Simply calling `deleteOnExist` is not enough, as this fails if the directory is not empty).

A new directory is created for each QTT/RQTT test.  This means thousands of hooks are registered, (there's currently over 5K QTT tests).  When the JVM is shutting down it creates THOUSANDS of threads, each trying to delete one directly.

Fixed by having a single parent directory, with a single shutdown hook.


agavra (approved):
Nice catch

ksqldb-functional-tests/src/main/java/io/confluent/ksql/test/utils/TestUtils.java
```diff
@@ -0,0 +1,74 @@
+/*
+ * Copyright 2019 Confluent Inc.
```
vcrfxia, line 2: Living in the past, I see ;P

vcrfxia (approved):
Awesome find @big-andy-coates ! Out of curiosity, how'd you manage to track this down?

big-andy-coates:
Watched the build in VisualVM and saw a massive spike in threads at the end.  Ran it again with sampling turned on in VisualVm and can see the stack of those threads.

PR #6491, opened 2020-10-22
test: qtt rewriting should not fail on output mismatch

big-andy-coates:
### Description 

Running `PlannedTestRewriterTest.rewritePlans` can fail because the output is not as expected, e.g. the topology may have changed, or the output may not be as expected, or the post conditions may not be met.

However... you're generally running `rewritePlans` _because_ something has changed and you want to rewrite the plans!  The upshot of this is that you have to comment out the validation, run the re-write, then revert the commenting-out.

This change disables the validation during rewriting, removing some of the tribal knowledge and the need to jump through hoops.


big-andy-coates:
Test only change... merging without review

Issue #6493, opened 2020-10-22
Query upgrades: Changing the filter on a table can result in missed tombstones


big-andy-coates:
If a user uses `CREATE OR REPLACE` to add a filter to a table then existing rows that don't match the new filter are not deleted, even if they are later updated with another value that doesn't pass the filter.

Take, as an example, the following test case from filters.sql:

```sql
----------------------------------------------------------------------------------------------------
--@test: add filter to basic TABLE without filter
----------------------------------------------------------------------------------------------------
SET 'ksql.create.or.replace.enabled' = 'true';

CREATE TABLE a (id INT PRIMARY KEY, col1 INT) WITH (kafka_topic='a', value_format='JSON');
CREATE TABLE b AS SELECT * FROM a;

-- INSERT row [1 -> 0]
INSERT INTO a (id, col1) VALUES (1, 0);

ASSERT VALUES b (id, col1) VALUES (1, 0);

-- At this point the table contains a single row: [1 -> 0]
-- Note, this row would not pass the new filter being added.
-- No tombstone is emitted
CREATE OR REPLACE TABLE b AS SELECT * FROM a WHERE col1 > 0;

-- Even though the following insert does not pass the new filter
-- and the table contains a row with key `1`
-- No tombstone is emitted
INSERT INTO a (id, col1) VALUES (1, -1);

-- This assert fails!
ASSERT NULL VALUES b (id) KEY (1);
```

There are two points at which we could emit a tombstone, neither of which are:

1. From the `CREATE OR REPLACE` statement.  This would involve running the new filer against all entries currently in the materialized table.  What to do if the table is not materialized?
1. From the `INSERT VALUES` following the `CREATE OR REPLACE` statement. The issue here is _probably_ that when the`Change` is passed through the streams topology and it hits the new filter, neither the old value nor the new value pass the filter, so no tombstone is emitted. (A tombstone is only emitted if the old value passes the filter and the new value does not).

To be correct, the `CREATE OR REPLACE` statement should cause a tombstone to be emitted.   Though not sure how practical / obtainable this is, especially for none-materialized tables, e.g. 

```
-- Neither FOO, nor BAR, need to be materialized.
CREATE TABLE FOO AS SELECT * FROM BAR;
```

It may be that we choose to live with this. That being the case we need to clearly document this limitations in whatever 'query upgrade' docs we have.

big-andy-coates:
[...]

PR #6494, opened 2020-10-22
test: fix SQL tester tests

big-andy-coates:
### Description 

Fixes some test failures due to upstream changes.

With the upstream fix https://github.com/apache/kafka/pull/9156 in, the streams topology is no longer emitting tombstones for rows that were never in the table to start with when an input doesn't pass a filter.

This change was causing some SQL Tester tests to fail.  This change fixes those tests.

The downside of the upstream change is that query upgrades can now miss sending tombstones for any rows that where in the table _before_ the upgrade, but which don't belong in the table _after_ the upgrade, e.g. because a filter has changed.  See https://github.com/confluentinc/ksql/issues/6493 for more details.


agavra (approved):
Thanks @big-andy-coates!

Issue #6506, opened 2020-10-23
`null` value with `PARTITION BY IS NULL` causes NPE


big-andy-coates:
**Describe the bug**

Seeing an NPE during stream processing, which causes the query to terminate.

Any `PARTITION BY exp`, where `exp` is not a column reference, will expose the query to this bug.

**To Reproduce**

Enable test in `null.json` that tracks this: "stream PARTITION BY IS NULL with null value".

```json
{
      "name": "stream PARTITION BY IS NULL with null value",
      "enabled": true,
      "statements": [
        "CREATE STREAM INPUT (ID INT KEY, NAME STRING) WITH (kafka_topic='test_topic', format='JSON');",
        "CREATE STREAM OUTPUT AS SELECT * FROM INPUT PARTITION BY (NAME IS NULL);"
      ],
      "inputs": [
        {"topic": "test_topic", "timestamp": 1, "key": 1, "value": null},
        {"topic": "test_topic", "timestamp": 2, "key": 2, "value": {"NAME": null}},
        {"topic": "test_topic", "timestamp": 3, "key": 3, "value": {"NAME": "n"}}
      ],
      "outputs": [
        {"topic": "OUTPUT", "timestamp": 1, "key": null, "value": null},
        {"topic": "OUTPUT", "timestamp": 2, "key": null, "value": {"ID": 2}},
        {"topic": "OUTPUT", "timestamp": 3, "key": "n", "value": {"ID": 3}}
      ]
    }
```

**Expected behavior**

test passes, maybe with a little tweaking of output

**Actual behaviour**

Test fails with:

```
failed test: null - stream PARTITION BY IS NULL with null value
in file:///Users/andy/dev/github.com/ksql/ksqldb-functional-tests/src/test/resources/query-validation-tests/null.json:53

	at io.confluent.ksql.test.EndToEndEngineTestUtil.shouldBuildAndExecuteQuery(EndToEndEngineTestUtil.java:48)
	at io.confluent.ksql.test.QueryTranslationTest.shouldBuildAndExecuteQueries(QueryTranslationTest.java:91)
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.lang.reflect.Method.invoke(Method.java:498)
[...]

PR #6507, opened 2020-10-23
test: add explicit QTT tests for null handling

big-andy-coates:
### Description 

Add more QTT tests around null handling - which we know to be inconsistent.  This is more about documenting current behaviour.

Raised https://github.com/confluentinc/ksql/issues/6506 for NPE bug found.


big-andy-coates:
Merging without review as its only adding tests.

PR #6508, opened 2020-10-23
fix: NPE in PARTITON BY on null value

big-andy-coates:
### Description 

fixes: #6506

Fixes a NPE during stream processing where the value in the source stream is `null` and the `PARTITION BY` is on an expression other than a column reference, e.g.

```sql
SELECT * FROM FOO
  PARTITION BY a + b;
```

The NPE would terminate the query.

### Testing done 

usual

### Reviewer checklist


big-andy-coates:
Note: I don't particularly agree with _how_ the code is currently handling `null` values.  But that's a separate matter.  

the purpose of this change is to avoid the NPE only.

agavra (approved):

agavra:
I'm merging this change on @big-andy-coates's behalf since he's on time off this week.

PR #6510, opened 2020-10-23
test: allow RQTT to 'run until failure' (MINOR)

big-andy-coates:
### Description 

Fixes the `TestKsqlRestApp` so that you can shut it down and restart it and submit new requests without it failing because it 'timed out waiting for previous command to be handled', because it's not resetting the `lastCommandSequenceNumber` when the server is restarted.

This means you can configure intellj to run RQTT until failure.  Which is useful!


big-andy-coates:
Merging as test only change.

PR #6511, opened 2020-10-23
chore: fix leaked executor

big-andy-coates:
### Description 

Ensures the `ExecutorService` in `PullQueryExecutor` is shutdown with the app.

This omission shouldn't cause issues in prod, but does in tests. Each test that starts the `PullQueryExecutor` was creating 100 threads and leaving them running.  Over the entire build that's a LOT of threads!

### Testing done 

Usual

### Reviewer checklist


big-andy-coates:
Merging without review as I believe it will help stabilise the build and I think it's pretty non controversial. Ping me if there's anything you think needs changing!

PR #6512, opened 2020-10-23
chore: minor RQTT debug improvements

big-andy-coates:
### Description 

Some minor tweaks that make it easier to debug RQTT failures and production issues.

### DIFF
--- a/ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/execution/PullQueryExecutor.java
+++ b/ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/execution/PullQueryExecutor.java
@@ -381,7 +381,7 @@ public final class PullQueryExecutor {
           schemas.add(result.getTableRows().getSchema());
           tableRows.addAll(result.getTableRows().getRows());
         } catch (ExecutionException e) {
-          LOG.debug("Error routing query {} to host {} at timestamp {} with exception {}",
+          LOG.warn("Error routing query {} to host {} at timestamp {} with exception {}",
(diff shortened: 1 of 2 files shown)


ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/execution/PullQueryExecutor.java
```diff
@@ -381,7 +381,7 @@ static PullQueryResult handlePullQuery(
           schemas.add(result.getTableRows().getSchema());
           tableRows.addAll(result.getTableRows().getRows());
         } catch (ExecutionException e) {
-          LOG.debug("Error routing query {} to host {} at timestamp {} with exception {}",
+          LOG.warn("Error routing query {} to host {} at timestamp {} with exception {}",
```
big-andy-coates, line 384: I'd argue that this should be a `WARN` because in normal production steady state operation this shouldn't happen, right?

If it is happening, e.g. because the consumer group is unstable, then people will likely be seeing poor pull query performance and may be wondering why.   Logging this will help highlight the cause.

During a cluster resize or bounce this can be logged, but again, if someone is investigating spikes in pull query latency, such a log message may be useful.

Thoughts? cc @AlanConfluent @vpapavas

AlanConfluent: Yes, I agree.  During normal operation, you wouldn't expect to see this, so `warn` is appropriate.

AlanConfluent (approved):

PR #6518, opened 2020-10-23
chore: json_sr tests and minor fixes

### LINKED ISSUE #6225: [JSON_SR Support]: Add support for `JSON_SR` as a key format
**JSON_SR support** Adds support for the `JSON_SR` key format, inc. schema registry integration. (5)
1. Allow it as key format
1. Can key serialiser be configured to NOT publish the schema to the SR? Or fail if the schema is different?  We really do not want to publish a new schema, which changes the schema id, breaking partitioning.
1. Enhancements to QTT and the ksqlDB testing tool to support JSON_SR key format.
   - Support populating key schemas, like can be done with `$.tests[*]->topics->schema`.  
   - Rename ’topics.schema’ -> ‘topics.valueSchema’
       - Support ’schema’ as value schema for legacy compatibility, in case users are using this in the testing tool
       - Release note to state use of `schema` is deprecated in the testing tool: use `valueSchema`.
       - Use `@deprecated` annotation in code for `schema` to increase likelyhood this will be removed later.
    - Add a few test cases.
        - `elements.json`
        - CSAS that takes key-less stream with NONE key format and repartitions to set key column: 
        - If default key format set in config, this should be used.
        - If default key format not set in config, error should be generated.
        - What happens if CSAS key format supports registering schemas, but statement includes PARTITION BY null?
     - Negative test that fails C* statements where no key columns provided, key format supports inference, and there is no key schema.
      - Test `INSERT INTO` statement does not change key schema.
      - Test `C*AS` statement fails if schema already registered and not the same: we don’t want a new schema version published.
1. Doc updates
       - update docs to reflect that `JSON_SR` can be used as a key schema.

agavra:
### Description 

mostly fixes #6225

- add `JSON_SR` key tests
- fix wiring issue with `isKey` and json_sr

### Testing done 

this change is nearly all testing :D 

### Reviewer checklist


ksqldb-functional-tests/src/main/java/io/confluent/ksql/test/model/PostConditionsNode.java
```diff
@@ -205,7 +205,7 @@ public PostTopicNode(
         @Override
         public void describeTo(final Description description) {
           try {
-            description.appendText(TestJsonMapper.INSTANCE.get().writeValueAsString(this));
+            description.appendText(TestJsonMapper.INSTANCE.get().writeValueAsString(PostTopicNode.this));
```
vcrfxia, line 208: Is this just for clarity? `this` should already refer to `PostTopicNode.this` here, right?

ksqldb-serde/src/main/java/io/confluent/ksql/serde/KeyFormatUtils.java
```diff
@@ -25,7 +25,8 @@
       FormatFactory.NONE,
       FormatFactory.KAFKA,
       FormatFactory.DELIMITED,
-      FormatFactory.JSON
+      FormatFactory.JSON,
+      FormatFactory.JSON_SR
```
vcrfxia, line 29: How come we're making JSON_SR available by default, given that AVRO still isn't enabled by default? Aren't both blocked on the same additional testing for joins?

Also, what's the point of having `"ksql.key.format.enabled": true` in the new QTTs (with primitive keys) if we're enabling this by default?

ksqldb-serde/src/test/java/io/confluent/ksql/serde/json/KsqlJsonSerializerTest.java
```diff
@@ -233,7 +236,7 @@ public void shouldHandleNestedStruct() throws IOException {
   }
 
   @Test
-  public void shouldSerializeBoolean() {
+  public void shouldSerializeBoolean() throws IOException, RestClientException {
```
vcrfxia, line 239: nit: I'm with @big-andy-coates on preferring tests simply have `throws Exception` so we don't need to worry about keeping lists of exceptions up to date. I don't think seeing the specific exceptions is insightful either.

vcrfxia (approved):
LGTM with a question inline! Exciting stuff :)

ksqldb-functional-tests/src/main/java/io/confluent/ksql/test/model/PostConditionsNode.java
agavra: no, this is within the `Matcher` anonymous class. I need to use `PostTopicNode.this` to refer to the outer wrapper (non-anon) class

ksqldb-serde/src/main/java/io/confluent/ksql/serde/KeyFormatUtils.java
agavra: 🤦 this is me not being careful. Thank you for checking this

Issue #6305, opened 2020-09-26
Head truncation on incompatible commands in ksql-restore command


stevenpyzhang:
**Is your feature request related to a problem? Please describe.**
As a follow up to https://github.com/confluentinc/ksql/issues/6286
We can extend it to head truncate certain commands.

**Describe the solution you'd like**
We can expose a parameter to head truncate commands from the command topic that are greater than some Command version number. This would help restore server functionality in the event of a rollback.

For the commands that are truncated, we'll need to search through them and check if there are any queries. For any queries, we'll need to take the application id/queryId and clean up internal topic and state stores.

**Describe alternatives you've considered**
Using special skip records in the command topic to skip certain commands.

PR #6524, opened 2020-10-26
feat: update ksql restore command to skip incompatible commands if flag set

stevenpyzhang:
### Description 
https://github.com/confluentinc/ksql/issues/6305

Added a `-s` flag for skipping commands when restoring. If the flag isn't present, the restore command program will exit if it encounters errors deserializing commands from the command topic backup file. If it is present and true, it'll skip over the incompatible command instead. The program then prints out the number of skipped commands and tries to clean up any left over KafkaStreams internal topics and state stores if the skipped command was a query.

### Testing done 
Added integration test
Local manual test

### Reviewer checklist

### DIFF
--- a/ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/CommandTopic.java
+++ b/ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/CommandTopic.java
@@ -89,8 +89,6 @@ public class CommandTopic {
       for (ConsumerRecord<byte[], byte[]> record : iterable) {
         try {
           backupRecord(record);
-        } catch (final KsqlException e) {
-          log.warn("Error encountered while backing up command.", e);
         } catch (final KsqlServerException e) {
           log.warn("Backup is out of sync with the current command topic. "
               + "Backups will not work until the previous command topic is "
@@ -118,8 +116,6 @@ public class CommandTopic {
       for (final ConsumerRecord<byte[], byte[]> record : records) {
         try {
           backupRecord(record);
-        } catch (final KsqlException e) {
-          log.warn("Error encountered while backing up command.", e);
         } catch (final KsqlServerException e) {
           log.warn("Backup is out of sync with the current command topic. "
               + "Backups will not work until the previous command topic is "
--- a/ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/CommandTopic.java
+++ b/ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/CommandTopic.java
@@ -17,6 +17,8 @@ package io.confluent.ksql.rest.server;
 
 import com.google.common.collect.Lists;
 import io.confluent.ksql.rest.server.computation.QueuedCommand;
+import io.confluent.ksql.util.KsqlException;
+import io.confluent.ksql.util.KsqlServerException;
 import java.time.Duration;
 import java.util.ArrayList;
 import java.util.Collections;
@@ -24,6 +26,7 @@ import java.util.List;
 import java.util.Map;
 import java.util.Objects;
 import java.util.Optional;
+
 import org.apache.kafka.clients.consumer.Consumer;
 import org.apache.kafka.clients.consumer.ConsumerRecord;
 import org.apache.kafka.clients.consumer.ConsumerRecords;
@@ -86,10 +89,13 @@ public class CommandTopic {
       for (ConsumerRecord<byte[], byte[]> record : iterable) {
         try {
           backupRecord(record);
-        } catch (final Exception e) {
+        } catch (final KsqlException e) {
+          log.warn("Error encountered while backing up command.", e);
+        } catch (final KsqlServerException e) {
           log.warn("Backup is out of sync with the current command topic. "
               + "Backups will not work until the previous command topic is "
               + "restored or all backup files are deleted.", e);
+          return records;
         }
         records.add(record);
       }
@@ -112,10 +118,13 @@ public class CommandTopic {
       for (final ConsumerRecord<byte[], byte[]> record : records) {
         try {
           backupRecord(record);
-        } catch (final Exception e) {
+        } catch (final KsqlException e) {
+          log.warn("Error encountered while backing up command.", e);
+        } catch (final KsqlServerException e) {
           log.warn("Backup is out of sync with the current command topic. "
               + "Backups will not work until the previous command topic is "
               + "restored or all backup files are deleted.", e);
+          return restoreCommands;
         }
 
         if (record.value() == null) {
--- a/ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/CommandTopic.java
+++ b/ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/CommandTopic.java
@@ -17,8 +17,7 @@ package io.confluent.ksql.rest.server;
 
 import com.google.common.collect.Lists;
 import io.confluent.ksql.rest.server.computation.QueuedCommand;
-import io.confluent.ksql.util.KsqlException;
-import io.confluent.ksql.util.KsqlServerException;
+import io.confluent.ksql.rest.server.resources.CommandTopicCorruptionException;
 import java.time.Duration;
 import java.util.ArrayList;
 import java.util.Collections;
@@ -89,7 +88,7 @@ public class CommandTopic {
       for (ConsumerRecord<byte[], byte[]> record : iterable) {
         try {
           backupRecord(record);
-        } catch (final KsqlServerException e) {
+        } catch (final CommandTopicCorruptionException e) {
           log.warn("Backup is out of sync with the current command topic. "
               + "Backups will not work until the previous command topic is "
               + "restored or all backup files are deleted.", e);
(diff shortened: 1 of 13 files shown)


ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/CommandTopic.java
```diff
@@ -90,7 +90,6 @@ public void start() {
           log.warn("Backup is out of sync with the current command topic. "
               + "Backups will not work until the previous command topic is "
               + "restored or all backup files are deleted.", e);
-          return records;
```
stevenpyzhang, line 93: We should be returning the record even if it isn't backed up. The CommandRunner has logic to detect when the command topic and backup are out of sync/the command is incompatible.

ksqldb-rest-app/src/test/java/io/confluent/ksql/api/integration/RestoreCommandTopicIntegrationTest.java
```diff
@@ -173,4 +256,27 @@ private boolean isDegradedState() {
...
+    }
+
+    final byte[] keyValueSeparator = ":".getBytes(StandardCharsets.UTF_8);
+    final byte[] newLine = "/n".getBytes(StandardCharsets.UTF_8);
+
+    writer.write(InternalTopicSerdes.serializer().serialize("", commandId));
+    writer.write(keyValueSeparator);
+    writer.write(InternalTopicSerdes.serializer().serialize("", command));
+    writer.write(newLine);
+    writer.flush();
```
spena, line 280: This looks like what the `BackupReplayFile` does. Can it be reused?
```
 // Just create another write() that accept the key and value instead of a consumer
BackupReplayFile.writable(backupFileLocation.toFile())
    .write(commandIdInBytes, commandInBytes);
```

stevenpyzhang: I reused it now

ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/restore/RestoreOptions.java
```diff
@@ -42,8 +42,16 @@
...
-      name = {"--yes", "-y"},
-      description = "Automatic \"yes\" as answer to prompt and run non-interactively.")
+      name = {"--skip-incompatible-commands", "-s"},
+      description = "This restore command can restore command topic commands that "
+          + "are of version (" + io.confluent.ksql.rest.server.computation.Command.VERSION + ") "
+          + "or lower. If true, the restore command will skip all incompatible commands."
+          + "If false, the restore command will restore the backup file as is.")
+  private boolean skipIncompatibleCommands = false;
+
+  @Option(
```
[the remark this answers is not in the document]
stevenpyzhang: I removed the header

ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/restore/KsqlRestoreCommandTopic.java
```diff
@@ -67,17 +70,26 @@ private static KsqlConfig loadServerConfig(final File configFile) {
...
+      final RestoreOptions options
+  ) throws IOException {
     final BackupReplayFile commandTopicBackupFile = BackupReplayFile.readOnly(file);
+    List<Pair<byte[], byte[]>> records = commandTopicBackupFile.readRecords();
 
-    final List<Pair<byte[], byte[]>> records = commandTopicBackupFile.readRecords();
-    throwOnInvalidRecords(records);
+    if (options.isSkipIncompatibleCommands()) {
+      records = removeIncompatibleCommands(records);
+    }
```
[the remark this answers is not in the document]
stevenpyzhang: Ok I changed it so that we always check that the commands are valid. The only different with the skip option is that if it's set to true, instead of throwing an exception and exiting the program when failing to deserialize the command due to a SerializationException/IncomaptibleKsqlCommandVersionException (which indicates an incompatible command) the method continues by skipping the command.

ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/CommandTopic.java
```diff
@@ -117,7 +116,6 @@ public void start() {
           log.warn("Backup is out of sync with the current command topic. "
               + "Backups will not work until the previous command topic is "
               + "restored or all backup files are deleted.", e);
-          return restoreCommands;
```
rodesai, line 120: We need to make the above exception more specific. We shouldn't just ignore any exception. Ditto for the other place this is used. I know this isn't part of your patch, but might as well do this now.

ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/CommandTopic.java
rodesai: this check detects the case where the backup and the command topic are inconsistent (have different contents). in that case we shouldn't return the command here, otherwise the command runner will try to execute it.

rodesai (changes requested):

spena (approved):
lgmt

ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/CommandTopic.java
stevenpyzhang: We should still return the record in some cases. For example, the command topic backup will fail to back up incompatible commands (it tries to deserialize the command before backing it up). I think these incompatible commands should still be returned so the CommandRunner can attempt to deserialize it

stevenpyzhang (commented):
This PR still needs some work actually, we need to check if the commands that are skipped have queries in them, and then clean up internal topics and state stores for them.

ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/CommandTopic.java
stevenpyzhang: I changed it to only catch a KsqlException or KsqlServerException

ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/CommandTopic.java
stevenpyzhang: @rodesai I changed it so that we still return the command in the case of a KsqlException since this is thrown by the backup when there's an issue deserializing either the commandId or command. The command isn't returned if it's a KsqlServerException since that means the corruption check has triggered.

ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/CommandTopic.java
```diff
@@ -86,7 +89,9 @@ public void start() {
       for (ConsumerRecord<byte[], byte[]> record : iterable) {
         try {
           backupRecord(record);
-        } catch (final Exception e) {
+        } catch (final KsqlException e) {
+          log.warn("Error encountered while backing up command.", e);
+        } catch (final KsqlServerException e) {
```
rodesai, line 94: we should define a new exception type for this case. this is a very specific error condition, and it should have its own exception class.

ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/CommandTopic.java
```diff
@@ -86,7 +89,9 @@ public void start() {
       for (ConsumerRecord<byte[], byte[]> record : iterable) {
         try {
           backupRecord(record);
-        } catch (final Exception e) {
+        } catch (final KsqlException e) {
```
rodesai, line 92: again - this should be a very specific exception indicating a deserialization error. Also, why are we even trying to deserialize the command from the backup logic (cc @spena)? It doesn't look like we're using any of the fields.

ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/restore/KsqlRestoreCommandTopic.java
```diff
@@ -65,17 +79,42 @@ private static KsqlConfig loadServerConfig(final File configFile) {
     return new KsqlConfig(serverProps);
   }
 
-  public static List<Pair<byte[], byte[]>> loadBackup(final File file) throws IOException {
+  public static List<Pair<byte[], byte[]>> loadBackup(
+      final File file,
+      final RestoreOptions options,
```
rodesai, line 84: nit: try and use the same variable name for the same thing everywhere (so either `options` or `restoreOptions`

rodesai (commented):
Couple more nits inline around naming/exception types

ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/CommandTopic.java
stevenpyzhang: Ok after talking it over with @spena , we don't need this deserialization logic in the backup impl, it should just backup whatever is found in the command topic, deserialization should be handled after it's backed up. So I'll get rid of this catch block and we'll only have the KsqlServerException one.

rodesai (approved):
LGTM!

Issue #6484, opened 2020-10-21
KSQL N-Way Join for the same table  is not working 


vrudenskyi:
**Describe the bug**
N-Way Join for the same table does not work 

**To Reproduce**
Steps to reproduce the behavior, include:
1. The version of KSQL: `Server v0.12.0`
2. Sample source data: `n/a`
3. Any SQL statements you ran
```
CREATE STREAM joined AS 
  SELECT * 
  FROM A
    LEFT JOIN B b1 ON A.b1_id = b1.id
    LEFT JOIN B b2 ON A.b2_id = b2.id;
```
**Expected behavior**
Stream created

**Actual behaviour**

 CLI output/Error messages:
`Can not join 'B' to 'B': self joins are not yet supported.`

**Additional context**
`n/a`

vcrfxia:
Hi @vrudenskyi , the behavior you're requesting is not possible in ksqlDB today as Kafka Streams (the underlying runtime) does not support it, and workarounds would require duplicating data in wasteful ways. Here are some related issues:
- Kafka Streams does not support reading the same topic multiple times: https://issues.apache.org/jira/browse/KAFKA-6687
- ksqlDB does not support self-joins for the same reason: https://github.com/confluentinc/ksql/issues/2030

(cc @mjsax in case something has changed since the filing of those issues)

I'm going to update this issue from `bug` to `enhancement` accordingly.

agavra:
@vcrfxia - I think the one thing that's interesting is that this isn't technically a self-join since there is no loop (though the limitation in KS probably still applies here):
```
  A
  |
  |--B
  |
  AB
  |
  |--B
  |
  ABB
```
We might want to change the error message to `"multi-way joins do not support reading the same topic multiple times"`

mjsax:
As the statement is `CSAS` it seems that `A` is a stream and `B` is a table. For this case, if should be possible to actually compile the query? It's not necessary to read `B` twice.

```
KTable b = builder.table("bTopic");
KStream A = builder.stream("aTopic");
a.selectKey(b1_id).join(b).select(b2_id).join(b);
```

I guess our check if a stream or table is referenced twice is rather "primitive" (I guess count based?).

vcrfxia:
> We might want to change the error message to "n-way joins do not support reading the same topic multiple times"

Sure, I've opened a PR to clarify the error message: https://github.com/confluentinc/ksql/pull/6527

rmoff:
[...]

PR #6527, opened 2020-10-27
chore: clarify self-join error message for n-way joins (MINOR)

vcrfxia:
### Description 

Relates to https://github.com/confluentinc/ksql/issues/6484.

Currently, the error message for a query such as 
```
CREATE STREAM joined AS 
  SELECT * 
  FROM A
    LEFT JOIN B b1 ON A.b1_id = b1.id
    LEFT JOIN B b2 ON A.b2_id = b2.id;
```
is `Can not join 'B' to 'B': self joins are not yet supported.` which is misleading since the query is not actually a self-join. This PR updates the error message to `N-way joins do not support multiple occurrences of the same source. Source: 'B'`

### Testing done 

Unit + QTT.

### Reviewer checklist


rodesai (approved):
LGTM

PR #6528, opened 2020-10-27
docs: remove references to EMIT FINAL

### LINKED ISSUE #6517: Remove references to SUPPRESS in the doc
Since suppression isn't yet support, we should remove these references from the docs for now: 

https://docs.ksqldb.io/en/0.12.0-ksqldb/developer-guide/ksqldb-reference/select-push-query/#final
https://docs.ksqldb.io/en/0.12.0-ksqldb/operate-and-deploy/installation/server-config/config-reference/#ksqlsuppressbuffersizebytes

agavra:
fixes #6517 - we can revert this PR when we're ready to re-introduce suppress after finishing the AK side of the change

### DIFF
--- a/docs/concepts/time-and-windows-in-ksqldb-queries.md
+++ b/docs/concepts/time-and-windows-in-ksqldb-queries.md
@@ -426,21 +426,3 @@ CREATE TABLE pageviews_per_region AS
 Note that the specified retention period should be larger than the sum of window size and any grace
 period.
 
-### Window Final Results
-
-In ksqlDB, windowed aggregations update their results continuously. As new data arrives for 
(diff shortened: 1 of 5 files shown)


JimGalasyn (approved):
LGTM, I'll cherry-pick to the 0.12.0-ksqldb, 0.12.x-ksqldb, and 0.13.x-ksqldb branches.

sreuland1:
Hello, is there a ticket for the AK side changes you referenced as blocker and an expectation on when that may be resolved? Does this mean we shouldn't be using the suppress/EMIT FINAL that's available on 0.12 or accept the known risks of buffer mentioned?

Issue #6367, opened 2020-10-06
ksqlDB leaks admin client threads for terminated http2 push queries JOIN


agavra:
**Describe the bug**
ksqlDB leaks threads when issued from the command line and then `^C`

**To Reproduce**

1. Run a ksqlDB server (happens on 0.13)
1. Create a stream (`CREATE STREAM a (id int key, col1 int) with (kafka_topic='a', value_format='json', partitions=1);`)
1. Create another stream (`CREATE STREAM c (id int key, col1 int) with (kafka_topic='c', value_format='json', partitions=1);`)
1. Record the number of running threads 
```
jstack `ps ax | grep ksql-server | awk '{print $1}' | head -1` | grep -oE '".*"' | wc -l
```
5. Issue a query 
```
curl --http2 -m 240 -d '{"sql":"SELECT * FROM a JOIN c WITHIN 10 second ON a.id = c.id EMIT CHANGES;"}' http://0.0.0.0:8088/query-stream
```
6. Exit the query with `ctrl+C`
1. Check the number of running threads 

**Expected behavior**
The number of threads should eventually be the same

**Actual behaviour**
The number of threads is increased by 1

**Additional context**
It seems that the threads which are being leaked are admin client threads:
```
~ ❯ jstack `ps ax | grep ksql-server | awk '{print $1}' | head -1` | grep "admin-client" | wc -l                                                                  16:55:34
      61
~ ❯ jstack `ps ax | grep ksql-server | awk '{print $1}' | head -1` | grep "admin-client" | wc -l                                                                  16:56:26
      62
```

Here is the stack for one such thread:
```
"kafka-admin-client-thread | adminclient-66" #1198 daemon prio=5 os_prio=31 tid=0x00007f9c75780000 nid=0x28607 runnable [0x000070001a71b000]
   java.lang.Thread.State: RUNNABLE
        at sun.nio.ch.KQueueArrayWrapper.kevent0(Native Method)
        at sun.nio.ch.KQueueArrayWrapper.poll(KQueueArrayWrapper.java:198)
        at sun.nio.ch.KQueueSelectorImpl.doSelect(KQueueSelectorImpl.java:117)
        at sun.nio.ch.SelectorImpl.lockAndDoSelect(SelectorImpl.java:86)
        - locked <0x0000000706797ce0> (a sun.nio.ch.Util$3)
        - locked <0x0000000706797cf0> (a java.util.Collections$UnmodifiableSet)
[...]

PR #6532, opened 2020-10-27
fix: clean up leaked admin client threads when issuing join query to query-stream endpoint

stevenpyzhang:
### Description 
Fixes https://github.com/confluentinc/ksql/issues/6367

Whenever a request is sent to either query endpoint, `DefaultApiSecurityContext.create(routingContext)` is called and a new service context is created. We're not closing the service context created in `createQueryPublisher` which causes the admin client created as part of the join query to never get closed, thus leaking the admin client thread.

### Testing done 
Manual testing with debugger, the number of admin client threads don't go up after repeatedly issuing join queries to the `query-stream` endpoint

### Reviewer checklist

### DIFF
--- a/ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/KsqlServerEndpoints.java
+++ b/ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/KsqlServerEndpoints.java
@@ -124,11 +124,22 @@ public class KsqlServerEndpoints implements Endpoints {
       final Context context,
       final WorkerExecutor workerExecutor,
       final ApiSecurityContext apiSecurityContext) {
-    return executeOnWorker(
-        () -> new QueryEndpoint(ksqlEngine, ksqlConfig, pullQueryExecutor, pullQueryMetrics)
-            .createQueryPublisher(sql, properties, context, workerExecutor,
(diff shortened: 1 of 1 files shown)


agavra (approved):
Nice catch @stevenpyzhang! Is it straightforward to add a test to make sure close is called, even on failure?

stevenpyzhang:
> Nice catch @stevenpyzhang! Is it straightforward to add a test to make sure close is called, even on failure?

@agavra unfortunately I don't think so. There are actually no unit tests at all for the `KsqlServerEndpoints` class currently. I played around with writing tests for the class, but it's hard to do since we're creating new objects inside the `createQueryPublisher` such as
```
        return new QueryEndpoint(ksqlEngine, ksqlConfig, pullQueryExecutor, pullQueryMetrics)
            .createQueryPublisher(
                sql,
                properties,
                context,
                workerExecutor,
                ksqlSecurityContext.getServiceContext());
```
The constructor was hanging when I ran the test. It might be doable if we can figure out how to set up the test environment properly, I'm not sure how to right now.

Alternatively, if we had PowerMock capabilities to mock this constructor that'd also work. 
https://www.gyanblog.com/java/21-how-mock-constructor-junit-test-case-development-issues/

agavra:
🤔 sounds good, we'll probably want to add tests at some point but I think it's fine to merge this for now

PR #6537, opened 2020-10-28
fix: '-e/-f' CLI parameters are not handling session variables correctly

spena:
### Description 

There are three fixes in this PR:
* fix: allow '-e' parameter to use variable substitution
* fix: SHOW VARIABLES is not working with RUN SCRIPT and '-f' parameter
* fix: RUN SCRIPT and '-f' parameter do not handle PRINT/SELECT queries

### Testing done 
Added unit tests and verified manually

### Reviewer checklist

### DIFF
--- a/ksqldb-cli/src/main/java/io/confluent/ksql/cli/Cli.java
+++ b/ksqldb-cli/src/main/java/io/confluent/ksql/cli/Cli.java
@@ -220,10 +220,9 @@ public class Cli implements KsqlRequestExecutor, Closeable {
   public void runCommand(final String command) {
     RemoteServerSpecificCommand.validateClient(terminal.writer(), restClient);
     try {
-      // Handles the RUN SCRIPT command if found
-      if (!terminal.maybeHandleCliSpecificCommands(command)) {
-        handleLine(command);
-      }
+      // Commands executed by the '-e' parameter do not need to execute specific CLI
+      // commands. For RUN SCRIPT commands, users can use the '-f' command parameter.
+      handleLine(command);
     } catch (final EndOfFileException exception) {
       // Ignore - only used by runInteractively() to exit the CLI
     } catch (final Exception exception) {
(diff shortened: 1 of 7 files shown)


ksqldb-cli/src/main/java/io/confluent/ksql/cli/Cli.java
```diff
@@ -354,22 +352,21 @@ private boolean isVariableSubstitutionEnabled() {
...
 
-  private List<ParsedStatement> substituteVariables(final List<ParsedStatement> statements) {
+  private ParsedStatement substituteVariables(final ParsedStatement statement) {
     if (isVariableSubstitutionEnabled()) {
-      return statements.stream()
-          .map(stmt -> VariableSubstitutor.substitute(stmt, sessionVariables))
-          .flatMap(replacedSql -> KSQL_PARSER.parse(replacedSql).stream())
-          .collect(Collectors.toList());
+      final String replacedStmt = VariableSubstitutor.substitute(statement, sessionVariables);
+      return KSQL_PARSER.parse(replacedStmt).get(0);
```
agavra, line 358: should we use `Iterables.getOnlyElement` here to make sure it fails if there's more than 1?

ksqldb-rest-model/src/main/java/io/confluent/ksql/rest/entity/VariablesList.java
```diff
@@ -80,7 +80,7 @@ public String toString() {
   @JsonCreator
   public VariablesList(
       @JsonProperty("statementText") final String statementText,
-      @JsonProperty("properties") final List<Variable> variables
+      @JsonProperty("variables") final List<Variable> variables
```
agavra, line 83: this has never been released, right? should be safe to change

agavra (approved):
LGTM

ksqldb-rest-model/src/main/java/io/confluent/ksql/rest/entity/VariablesList.java
spena: Yeap, never released, so we're good here.

PR #6544, opened 2020-10-29
feat: support multi-column key declarations

agavra:
### Description 

Well @big-andy-coates does it again. I dug into this and hit some walls the first times around making me think it would be really tough, but after a few scrap-my-work-and-try-again loops I have proved Andy right and myself wrong - it was actually much easier to support multi-column keys than I suspected! 

There are still imitations (you cannot issue pull queries against a multi-column key table and you cannot join a multi-column source that cannot be repartition) and the existing limitations still hold (group by still causes an ugly concatenation of the key) but will be fixed in future releases.

### Testing done 

Mostly fixed unit tests, added `multi-key-cols.json` which tests most of the basic scenarios.

### TODOs 

After this PR we will still need to:
- split up the mega `ksql.key.format.enabled` config into multiple that target smaller diffs
- schema inference work
- make sure avro/pb work (all tests were done in JSON)
- fix semantics
- joins, pull query support
- partition by multiple column support

### Reviewer checklist


ksqldb-engine/src/main/java/io/confluent/ksql/planner/plan/FinalProjectNode.java
```diff
@@ -104,6 +108,7 @@ public void validateKeyPresent(final SourceName sinkName) {
 
     if (into.isPresent()) {
       // Persistent queries have key columns as key columns - so final projection can exclude them:
```
vcrfxia, line 110: nit (not your code):
```suggestion
      // Persistent queries have key columns as value columns - so final projection can exclude them:
```

ksqldb-engine/src/main/java/io/confluent/ksql/planner/plan/FinalProjectNode.java
```diff
@@ -114,10 +119,28 @@ public void validateKeyPresent(final SourceName sinkName) {
...
+          }
         }
         return false;
       });
+
+      for (final Entry<ColumnName, Set<ColumnName>> seenKey : seenKeyColumns.entrySet()) {
+        if (seenKey.getValue().size() > 1) {
+          final String keys = GrammaticalJoiner.and().join(
+              seenKey.getValue().stream().map(Name::text).sorted());
+          throw new KsqlException("The projection contains a key column (" + seenKey.getKey()
```
vcrfxia, line 134: Not directly related to your PR, but can you help me understand why the validation that key columns are not selected more than once is performed in `FinalProjectNode` while the validation that key columns are selected is performed in `DataSourceNode`, rather than performing both checks in the same place?

ksqldb-engine/src/main/java/io/confluent/ksql/planner/plan/FinalProjectNode.java
```diff
@@ -142,16 +165,6 @@ public void validateKeyPresent(final SourceName sinkName) {
   private void validate() {
```
vcrfxia, line 165: nit: it's a bit odd that this method is called `validate` now that there's validation being performed in `build()` as well. Can we either move the validation back into this method, or rename this method to something more specific (`throwOnZeroValueColumns()` or similar)?

ksqldb-execution/src/main/java/io/confluent/ksql/execution/util/StructKeyUtil.java
```diff
@@ -58,16 +57,14 @@ public static KeyBuilder keyBuilder(final ColumnName name, final SqlType type) {
...
-    private final org.apache.kafka.connect.data.Field keyField;
 
-    private KeyBuilder(final Schema keySchema) {
+    public KeyBuilder(final Schema keySchema) {
       this.keySchema = Objects.requireNonNull(keySchema, "keySchema");
-      this.keyField = Iterables.getOnlyElement(keySchema.fields());
     }
 
-    public Struct build(final Object keyValue) {
+    public Struct build(final Object keyValue, final int fieldIndex) {
```
vcrfxia, line 65: To check my understanding: this works today because we only ever end up with a single key after partition by or group by, but this method signature will have to change to support building structs with multiple key columns in a future PR, yes?

ksqldb-functional-tests/src/test/resources/query-validation-tests/multi-col-keys.json
```diff
@@ -0,0 +1,371 @@
...
+        {"topic": "OUTPUT", "key": 0, "value": {"S1_K": 1, "S1_K2": 2, "S2_K": 1, "S2_K2": 2, "S2_V": 0}}
+      ],
+      "post": {
+        "sources": [
+          {"name": "OUTPUT", "type": "stream", "schema": "S1_V INTEGER KEY, S1_K INTEGER, S1_K2 INTEGER, S2_K INTEGER, S2_K2 INTEGER, S2_V INTEGER"}
+        ]
+      }
+    },
+    {
+      "name": "nested struct key",
```
vcrfxia, line 306: How is this is a nested struct key?

ksqldb-functional-tests/src/test/resources/query-validation-tests/multi-col-keys.json
```diff
@@ -0,0 +1,371 @@
...
+        "CREATE TABLE NO_KEY (K STRING PRIMARY KEY, K2 STRING PRIMARY KEY, NAME string) WITH (kafka_topic='NO_KEY', format='JSON');",
+        "CREATE STREAM OUTPUT as SELECT s.k, name FROM S JOIN NO_KEY t ON s.k = t.k;"
+      ],
+      "expectedException": {
+        "type": "io.confluent.ksql.util.KsqlStatementException",
+        "message": "Cannot repartition a TABLE source. If this is a join, joins on tables with multiple columns is not yet supported."
+      }
+    }
+  ]
+}
```
vcrfxia, line 371: Can we also add a negative test for partition by on multiple columns?

ksqldb-streams/src/test/java/io/confluent/ksql/execution/streams/SourceBuilderTest.java
```diff
@@ -555,6 +560,70 @@ public void shouldHandleNullKey() {
   }
 
   @Test
+  public void shouldHandleMultiField() {
```
vcrfxia, line 563: nit: 
```suggestion
  public void shouldHandleMultiKeyField() {
```
and similarly for the other new tests below.

ksqldb-streams/src/test/java/io/confluent/ksql/execution/streams/SourceBuilderTest.java
```diff
@@ -768,6 +849,20 @@ private void givenWindowedSourceTable() {
     );
   }
 
+  private void givenWindowedMultiKeySourceTable() {
```
vcrfxia, line 852: It's not possible for a user to have a windowed, multi-key table today, right? Since group by on multiple columns still creates a single key, and we don't allow importing windowed tables?

Also, this method looks unused.

vcrfxia (approved):
Awesome stuff @agavra ! LGTM with the usual questions + nits inline.

ksqldb-execution/src/main/java/io/confluent/ksql/execution/util/StructKeyUtil.java
agavra: yes - I tried to change this to make it more generic but it causes a shocking amount of code jitter, so I'll do that in a future PR

ksqldb-engine/src/main/java/io/confluent/ksql/planner/plan/FinalProjectNode.java
agavra: Every node "validates" that the key is selected, they just all delegate to the child node that has that information:
```java
// PlanNode.java:L145
    getSources().forEach(s -> s.validateKeyPresent(sinkName, projection));
  }
```

`DataSourceNode`, `JoinNode` and `UserRepartitionNode` all override this method because those are the nodes that truly "know" whether the key was selected. Imagine we projected the key column with a different name, the schema in the `FinalProjectNode` would consider that the key - but of course we didn't select that! We selected the original key. 

To illustrate, in the example below the project node has a schema `[col1 INT PRIMARY KEY]` but we didn't select `col1` - so it would fail! Of course we could do some magic and "remember" that `b.col1` is actually `a.id`, but why not just delegate that to the node that already tracks that?
```sql
CREATE TABLE a (id INT PRIMARY KEY, col1 INT);
CREATE TABLE b AS SELECT id AS col1; 
```

(Note that it's a little confusing, because there is also `VerifiableNode`, which `AggregateNode`, `FinalProjectNode` and `SuppressNode` all override - and it is `VerifiableNode#validateKeyPresent` is what is called at the top level, but these just delegate down to the three source nodes eventually). 

> why the validation that key columns are not selected more than once is performed in `FinalProjectNode`

I think this one is clearer why it can't be in `DataSourceNode`, it's the other way around that might need some justification.

---

All that being said, I think this can definitely be made clearer going forward. The logical nodes don't follow a very strict abstraction model unfortunately.

ksqldb-functional-tests/src/test/resources/query-validation-tests/multi-col-keys.json
agavra: physically, it's a struct within a struct. I'll rename this "struct as column in multi-column key"

ksqldb-functional-tests/src/test/resources/query-validation-tests/multi-col-keys.json
agavra: I originally did not add this because our parser doesn't even except `PARTITION BY x,y` but i'll add one in and make sure it doesn't parse

ksqldb-streams/src/test/java/io/confluent/ksql/execution/streams/SourceBuilderTest.java
agavra: yes you are right, that was from a change i had backed out and forgot to udpate! thanks for catching this

ksqldb-engine/src/main/java/io/confluent/ksql/planner/plan/FinalProjectNode.java
vcrfxia: Thanks for the explanation -- super helpful!

PR #6143, opened 2020-09-02
feat: terminate persistent query on DROP command

spena:
### Description 
Fixes https://github.com/confluentinc/ksql/issues/2177

This PR automatically terminates the persistent query of a stream/table when executing the DROP command. If the stream/table to drop has another query writing/reading on it, then the DROP will fail. This prevents leaking zombie queries on dropped streams.

The query termination is done by the `MetaStoreImpl.deleteSource` in an atomic operation. This code is the one checking for referential integrity so it is the best place to do the job. 

To check if the query is good to terminate, the metastore stores the stream/table source with a reference to the queryId running. This way, when attempting to delete the source, it checks if the referenced query matches the query in the `ReferentialIntegrityTableEntity`. The query is terminated only if the query is a Sink query and no other queries exist, either source or sink. 

I added a few tests in the `KsqlEngine` to verify the drop command works (negative & positively).

### Testing done 
Added unit tests
Verified manually
```
# terminate query allowed
create stream s1 as select * from t1;
drop stream s1;

# terminate query not allowed
create stream s1 as select * from t1;
create stream s2 as select * from s1;
drop stream s1; // fail
drop stream s2; // success
drop stream s1; // success

# terminate query not allowed
create stream s1 as select * from t1;
insert into s1 select * from t1;
drop stream s1; // fail
terminate insertquery_1;
drop stream s1; // success
```
terminate_drop_stream

### Reviewer checklist
- [ ] Ensure docs are updated if necessary. (eg. if a user visible feature is being added or changed).
- [ ] Ensure relevant issues are linked (description should include text like "Fixes #<issue number>")

PR #6545, opened 2020-10-29
feat: add constraint to deny drop sources referenced by other CREATE_AS sources

spena:
### Description 
This is partial implementation of [feat: terminate persistent query on DROP command](https://github.com/confluentinc/ksql/pull/6143)

The first commit of this PR was already reviewed in the above PR (https://github.com/confluentinc/ksql/pull/6143). This PR only adds extra work to:
* Return DROP source constraints from DESCRIBE EXTENDED
* Drop sources in order for Rest QTT cleanup methods

The change for this PR adds a constraint to a source to not delete it if it is referenced by another source. For instance:
```
ksql> create stream t1(id int) with (kafka_topic='t1');
ksql> create stream s1 as select * from t1;
ksql> terminate CSAS_S1_9;

ksql> drop stream t1;
Cannot drop T1.
The following streams and/or tables read from this source: [S1].
You need to drop them before dropping T1.

ksql> describe extended t1;
...
Sources that have a DROP constraint on this source
--------------------------------------------------
S1
...
```

Even if queries are terminated, the stream or table cannot be deleted until all sources that reference to it are deleted. This is a common approach in other DBs, which could keep constraints and references on tables thus avoiding deleting them leaving tables with an invalid link.

This feature was asked by https://github.com/confluentinc/ksql/pull/6143. The main reason of splitting the PR is for easy review the rest of the commits.

Commit #1 is already reviewed and approved. If you want you can review again, but you can opt-out, and review only Commit #2 and Commit #3, which add the list to the DESCRIBE EXTENDED and fixes the RQTT tests. The rest of the commits are just fixing a couple of tests.

### Testing done 

### Reviewer checklist

### DIFF
--- a/ksqldb-api-client/src/main/java/io/confluent/ksql/api/client/SourceDescription.java
+++ b/ksqldb-api-client/src/main/java/io/confluent/ksql/api/client/SourceDescription.java
@@ -86,8 +86,8 @@ public interface SourceDescription {
   String sqlStatement();
 
   /**
-   * Returns a list of sources that have a DROP constraint reference on this source. This DROP
-   * constraint does not allow to delete this source after all referenced sources are deleted.
+   * Returns a list of sources that have a DROP constraint reference on this source. This source
+   * cannot be dropped until all sources returned by this method are deleted.
    *
    * @return a list of sources
    */
--- a/ksqldb-api-client/src/main/java/io/confluent/ksql/api/client/SourceDescription.java
+++ b/ksqldb-api-client/src/main/java/io/confluent/ksql/api/client/SourceDescription.java
@@ -85,4 +85,11 @@ public interface SourceDescription {
    */
   String sqlStatement();
 
+  /**
+   * Returns a list of sources that have a DROP constraint reference on this source. This DROP
+   * constraint does not allow to delete this source after all referenced sources are deleted.
+   *
+   * @return a list of sources
+   */
+  List<String> getSourceConstraints();
 }
--- a/ksqldb-rest-model/src/main/java/io/confluent/ksql/rest/entity/SourceDescription.java
+++ b/ksqldb-rest-model/src/main/java/io/confluent/ksql/rest/entity/SourceDescription.java
@@ -48,6 +48,7 @@ public class SourceDescription {
   private final int replication;
   private final String statement;
   private final List<QueryOffsetSummary> queryOffsetSummaries;
+  private final List<String> sourceConstraints;
 
   // CHECKSTYLE_RULES.OFF: ParameterNumberCheck
   @JsonCreator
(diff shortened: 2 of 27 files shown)


ksqldb-rest-model/src/main/java/io/confluent/ksql/rest/entity/SourceDescription.java
```diff
@@ -68,7 +69,9 @@ public SourceDescription(
       @JsonProperty("partitions") final int partitions,
       @JsonProperty("replication") final int replication,
       @JsonProperty("statement") final String statement,
-      @JsonProperty("queryOffsetSummaries") final List<QueryOffsetSummary> queryOffsetSummaries) {
+      @JsonProperty("queryOffsetSummaries") final List<QueryOffsetSummary> queryOffsetSummaries,
+      @JsonProperty("sourceConstraints") final List<String> sourceConstraints
```
agavra, line 73: what happens if we get old response with new CLI and vice versa?

ksqldb-api-client/src/main/java/io/confluent/ksql/api/client/SourceDescription.java
```diff
@@ -85,4 +85,11 @@
    */
   String sqlStatement();
 
+  /**
+   * Returns a list of sources that have a DROP constraint reference on this source. This DROP
+   * constraint does not allow to delete this source after all referenced sources are deleted.
```
agavra, line 90:
```suggestion
   * Returns a list of sources that have a DROP constraint reference on this source. This source
   * cannot be dropped until all sources returned by this method are deleted.
```

agavra (approved):
LGTM!

ksqldb-rest-model/src/main/java/io/confluent/ksql/rest/entity/SourceDescription.java
spena: From an old CLI to the new Server works.

From the new CLI to the old Server doesn't work. But, is this a valid support? And, how do others maintain compatibility with an old Server? I was looking at the code for other properties and I don't see how this is supported. Any idea?

spena:
Btw, I added a new test in RecoverTest to validate the scenario caused by an upgrade, where the DROP stream could be in an order where it has DROP constraints. There was nothing to fix. The compactor did the thing to remove the query, thus removing the DROP constraint. I'll make sure to re-validate this before merging the DROP+TERMINATE command.

ksqldb-rest-model/src/main/java/io/confluent/ksql/rest/entity/SourceDescription.java
agavra: > From an old CLI to the new Server works.

This one is more important. The other is a nice-to-have. It's not expected to work, but it would be nice if it behaved nicely (either ignoring it or telling you there's an issue). You don't need to implement that for this pr

PR #6549, opened 2020-10-30
build: st-4518 Fix release script to work with changes made from nano versioning. still using beta artifacts.

elismaga:
### Description 
The ksqldb release build broke due to the nano versioning changes. This does not change the ksqldb release build to use nano versioned artifacts so it will continue to use beta artifacts. That is a separate project coming soon. 

The build was failing because when it runs the maven plugins it loads the pom file first and that was failing because it couldn't find the nano versioned artifacts which the parent pom is set to use. Since this repo is pointing at the maven repo with the beta artifacts and not the repo with the nano versioned artifacts it does not find them.

I fixed this by doing a find and replace of the parent version specified in the parent pom file instead of using the maven plugin. Once that is done the pom file should be able to load successfully and the other plugins we run should work. I removed the call to the plugin we had which was setting the parent version before.

There are also two new properties introduced with nano versioning that we also need to set now, and I added two calls the maven set-property plugin to update those. One is the schema registry version property which tells the project which version of that to use, which is set to the beta version. The other property is similar to the project.version so we set that to the ksqldb version we are building.

**I also had to update the beta version that was being used because the version that is currently is used is old and does not contain the nano versioning changes which the latest version of the ksql code expect since it has the nano versioning changes. I updated it to the latest version of 6.1.0 instead of jumping to 6.2.0**

### Testing done 
I tested the regex locally by running it from a groovy script and then doing a diff of the file to make sure it only changed what it was supposed to. All of the changes I made to this script get run through during the PR build so it should be all set. The only additional step that is done when it is not a PR build is to tag the repo and upload the artifacts. I don't think any of the changes made from the nano versioning roll out will affect those steps though.

**I am going to be on vacation starting tomorrow 10/30 and all of next week so I will not be around to support this, but some one else on the team should be available to merge this for and help out with any bugs. We can also test this out by doing a replay of the Jenkins build that was being done when we encountered this bug.**

### Reviewer checklist


Jenkinsfile
```diff
@@ -208,12 +210,22 @@ def job = {
                                 sh "docker pull ${config.dockerRegistry}${dockerRepo}:${config.cp_version}-latest"
                             }
 
+                            // We need to replace the parent version range before we can run any maven commands
+                            def pomFile = readFile('pom.xml')
+                            def parentVersionPattern = Pattern.compile(/(.*<parent>.*<groupId>io.confluent\S*<\/groupId>.*<version>)\[\d+\.\d+\.\d+-\d+\,\s+\d+\.\d+\.\d+-\d+\)(<\/version>.*<\/parent>.*)/, Pattern.DOTALL)
+                            // Groovy regex replaces the groups we didn't match, so we print the beginning of the file, our version, and the rest of the file.
+                            def newPomFile = pomFile.replaceFirst(parentVersionPattern, "\$1${config.cp_version}\$2")
```
andrewegel, line 217:
```
                            def newPomFile = pomFile.replaceFirst(parentVersionPattern, "\$1${config.cp_version}\$2")
```

Is the `\$1` groovy syntax to reference matched groups?

Jenkinsfile
```diff
@@ -208,12 +210,22 @@ def job = {
...
+                            writeFile(file: 'pom.xml', text: newPomFile)
```
andrewegel, line 218: I think somewhere else in this Jenkinsfile we mark the root pom as an artifact, since we're re-writing it here, should we mark that as the artifact?

Jenkinsfile
```diff
@@ -208,12 +210,22 @@ def job = {
                                 sh "docker pull ${config.dockerRegistry}${dockerRepo}:${config.cp_version}-latest"
                             }
 
+                            // We need to replace the parent version range before we can run any maven commands
+                            def pomFile = readFile('pom.xml')
+                            def parentVersionPattern = Pattern.compile(/(.*<parent>.*<groupId>io.confluent\S*<\/groupId>.*<version>)\[\d+\.\d+\.\d+-\d+\,\s+\d+\.\d+\.\d+-\d+\)(<\/version>.*<\/parent>.*)/, Pattern.DOTALL)
```
andrewegel, line 215: I wonder if we should be factoring this out into function jenkins-common?

andrewegel (approved):
Just nitpicks and questions.

Jenkinsfile
```diff
@@ -1,8 +1,10 @@
+import java.util.regex.Pattern
+
 def baseConfig = {
     owner = 'ksql'
     slackChannel = '#ksql-alerts'
     ksql_db_version = "0.14.0"  // next version to be released
-    cp_version = "6.1.0-beta201006024150"  // must be a beta version from the packaging build
+    cp_version = "6.1.0-beta201020194508"  // must be a beta version from the packaging build
```
vcrfxia, line 7: Hey @elismaga (and the rest of the tools team since Eli is OOO), this version bump is a problem for us since the previous version `6.1.0-beta201006024150` has been qualified for our release, but not this new one. Do we have options to release with the old version?

gracechensd: I'm not familiar with the ksql release process so apologies in advance, but would it be a possibility if we instead cherrypick the nanoversioning rollout changes on top of the old 6.1.0-beta201006024150? Or would qualifying that take just as long as qualifying the newer version?

vcrfxia: 6.1.0-beta201006024150 is a beta packaging build produced by the tools team. Someone from your team can probably comment more authoritatively than I can on how difficult it is to cherry-pick and reproduce beta artifacts?

xli1996: I do not image that would be easy to cherrypick nanoversion change to the old beta version, I guess we could try just add the regex part and removed the version bump part.

xli1996:
@vcrfxia @vpapavas After reviewing the code I think we have few options:

1. Temporary revert the nano version change to cut the release (We could create a new branch with revert changes and use that branch)
2. Cherry pick all nano version change to the identical beta tag and redo the beta packaging

I think first option might be more easier, any concern to that?

vcrfxia:
Hi @xli1996 , did you confirm it's not possible to back out the portion of this PR that depends on nano-versioning, in order to leave the rest of the PR that uses the old beta packaging dependencies (without nano-versioning)? This sounds easier to me than reverting the nano-versioning change on the branch (the option you proposed above), but if this is not possible then reverting nano-versioning on the branch sounds good to me. Thanks.

Jenkinsfile
elismaga: yes

Issue #6547, opened 2020-10-29
Joins fail on key formats with different default serde features


vcrfxia:
**Describe the bug**

Issuing a join where the key formats of the two sources differ fails if the default serde features for the two sources also differ. For example, joining a source with KAFKA key format (default serde features `[ ]`) and a source with JSON key format (default serde features `[ UNWRAP_SINGLES ]`) results in the following error:
```
io.confluent.ksql.util.KsqlStatementException: Key format features mismatch in join. left: [], right: [UNWRAP_SINGLES]
```

**To Reproduce**

Here's a QTT to reproduce:
```
    {
      "name": "stream-steam key-to-key - key format and default feature mismatch",
      "statements": [
        "CREATE STREAM S1 (ID INT KEY, FOO INT) WITH (kafka_topic='s1', key_format='DELIMITED', value_format='JSON');",
        "CREATE STREAM S2 (ID INT KEY, VAL STRING) WITH (kafka_topic='s2', key_format='JSON', value_format='JSON');",
        "CREATE STREAM OUTPUT AS SELECT S1.ID, S2.VAL FROM S1 JOIN S2 WITHIN 10 SECONDS ON S1.ID = S2.ID;"
      ],
      "inputs": [
        {
          "topic": "s2",
          "key": 10,
          "value": {
            "VAL": "hello"
          }
        },
        {
          "topic": "s1",
          "key": 10,
          "value": {
            "foo": 22
          }
        }
      ],
      "outputs": [
        {
          "topic": "OUTPUT",
          "key": 10,
          "value": {
            "VAL": "hello"
          }
        }
      ],
      "post": {
        "sources": [
          {
            "name": "OUTPUT",
            "type": "stream",
            "keyFormat": {
              "format": "DELIMITED"
            }
          }
        ],
        "topics": {
          "topics": [
            {
              "name": "_confluent-ksql-some.ksql.service.idquery_CSAS_OUTPUT_0-Join-right-repartition",
              "keyFormat": {
                "format": "DELIMITED"
              },
              "valueFormat": {
                "format": "JSON"
              }
            },
            {
              "name": "_confluent-ksql-some.ksql.service.idquery_CSAS_OUTPUT_0-KSTREAM-JOINTHIS-0000000012-store-changelog",
              "keyFormat": {
                "format": "DELIMITED"
              },
              "valueFormat": {
                "format": "JSON"
              }
            },
            {
[...]

PR #6550, opened 2020-10-30
fix: support joins on key formats with different default serde features

vcrfxia:
### Description 

Fixes https://github.com/confluentinc/ksql/issues/6547

### Testing done 

Added QTT. Looked into adding a unit test in SchemaKStreamTest but couldn't find a good way to do it (existing coverage for the changed block of code is missing as well, probably for the same reason).

### Reviewer checklist


agavra (approved):
LGTM! Soon we'll need to revisit `SerdeFeaturesFactory#buildInternal` to handle multi-column keys, but I think we don't need that yet for #6544 because it only supports re partitioning by a single key

Issue #6533, opened 2020-10-27
feat: new WITH(id = '<query-id>') clause to INSERT/SELECT 


spena:
**Is your feature request related to a problem? Please describe.**
There is a discussion about removing the `INSERT/SELECT` command from ksqlDB (see [KLIP-17](https://github.com/confluentinc/ksql/pull/4125)). The discussion does not have an agreement yet. However, users are still using the `INSERT/SELECT` in their environments, and there is not good experience when using this command in SQL scripts, especially when terminating the `INSERT/SELECT` queries.

When a user executes an `INSERT/SELECT` command, ksqlDB registers the query with the `INSERT_{randomN}` ID. To terminate the query, the user must know this `{randomN}`. ksqlDB uses an algorithm to get this number based on the command topic offset. Users might able to guess this and use it in their SQL scripts, but this is not a good user experience when working with SQL scripts and migrations.

**Describe the solution you'd like**
To help users terminate `INSERT/SELECT` queries from their SQL scripts without guessing the `{randomN}`, we should let them  choose an appropriate query ID they want to use, and let them terminate the query with such ID.

For example:
```
INSERT INTO s1 WITH(ID='my-insert-1') SELECT * FROM s0;
TERMINATE 'my-insert-1';
```

With this approach, users could have the following benefits:
* Give a meaning to the query ID, such as `insert_missing_rows`, `insert_valid_metrics`, etc.
* Terminate queries without knowing the `{randomN}` (very useful when working with SQL scripts)

It is desired that the new ID is displayed by `SHOW QUERIES` too. I don't know if this new ID should replace the `INSERT_{randomN}` ID or just use as an alias.

PR #6553, opened 2020-10-30
feat: add WITH clause and QUERY_ID property to INSERT INTO statements

spena:
### Description 
Implements https://github.com/confluentinc/ksql/issues/6533

Add a `WITH` clause to the `INSERT/SELECT` statement to allow users specify a custom query QUERY_ID. 
For intance:
```
ksql> insert into s1 with (query_id='my_insert_id') select * from s1;

 Message                            
------------------------------------
 Created query with ID MY_INSERT_ID 
------------------------------------

ksql> show queries;

 Query ID     | Query Type | Status    | Sink Name | Sink Kafka Topic | Query String                                              
----------------------------------------------------------------------------------------------------------------------------------
 MY_INSERT_ID | PERSISTENT | RUNNING:1 | S1        | t1               | insert into s1 with (query_id='my_insert_id') select * from s1; 
----------------------------------------------------------------------------------------------------------------------------------
For detailed information on a Query run: EXPLAIN <Query ID>;

ksql> terminate my_insert_id;

 Message           
-------------------
 Query terminated. 
-------------------
```

### Testing done 

### Reviewer checklist

### DIFF
--- a/ksqldb-common/src/main/java/io/confluent/ksql/properties/with/InsertIntoConfigs.java
+++ b/ksqldb-common/src/main/java/io/confluent/ksql/properties/with/InsertIntoConfigs.java
@@ -21,7 +21,7 @@ import org.apache.kafka.common.config.ConfigDef;
  * 'With Clause' properties for 'INSERT INTO' statements.
  */
 public final class InsertIntoConfigs {
-  public static final String QUERY_ID_PROPERTY = "ID";
+  public static final String QUERY_ID_PROPERTY = "QUERY_ID";
 
   private static final ConfigDef CONFIG_DEF = new ConfigDef()
       .define(
@@ -29,7 +29,11 @@ public final class InsertIntoConfigs {
           ConfigDef.Type.STRING,
           null,
           ConfigDef.Importance.LOW,
-          "Custom query ID to use for INSERT INTO queries"
+          "Optional query ID to use for INSERT INTO queries. This query ID will be "
+              + "displayed by `SHOW QUERIES`, and will also be used to terminate the query. "
+              + "If empty, the default 'INSERTQUERY_' prefix plus a number will be used. "
+              + "Queries IDs are case-insensitive. Queries IDs will be displayed in uppercase "
+              + "with `SHOW QUERIES`."
       );
 
   public static final ConfigMetaData CONFIG_METADATA = ConfigMetaData.of(CONFIG_DEF);
--- a/ksqldb-common/src/main/java/io/confluent/ksql/properties/with/InsertIntoConfigs.java
+++ b/ksqldb-common/src/main/java/io/confluent/ksql/properties/with/InsertIntoConfigs.java
@@ -1,0 +1,39 @@
+/*
+ * Copyright 2020 Confluent Inc.
+ *
+ * Licensed under the Confluent Community License (the "License"); you may not use
+ * this file except in compliance with the License.  You may obtain a copy of the
+ * License at
+ *
+ * http://www.confluent.io/confluent-community-license
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OF ANY KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations under the License.
+ */
+
+package io.confluent.ksql.properties.with;
+
+import org.apache.kafka.common.config.ConfigDef;
+
+/**
+ * 'With Clause' properties for 'INSERT INTO' statements.
+ */
+public final class InsertIntoConfigs {
+  public static final String QUERY_ID_PROPERTY = "ID";
+
+  private static final ConfigDef CONFIG_DEF = new ConfigDef()
+      .define(
+          QUERY_ID_PROPERTY,
+          ConfigDef.Type.STRING,
+          null,
+          ConfigDef.Importance.LOW,
+          "Custom query ID to use for INSERT INTO queries"
+      );
+
+  public static final ConfigMetaData CONFIG_METADATA = ConfigMetaData.of(CONFIG_DEF);
+
+  private InsertIntoConfigs() {
+  }
+}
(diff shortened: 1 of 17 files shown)


ksqldb-common/src/main/java/io/confluent/ksql/properties/with/InsertIntoConfigs.java
```diff
@@ -0,0 +1,39 @@
...
+
+package io.confluent.ksql.properties.with;
+
+import org.apache.kafka.common.config.ConfigDef;
+
+/**
+ * 'With Clause' properties for 'INSERT INTO' statements.
+ */
+public final class InsertIntoConfigs {
+  public static final String QUERY_ID_PROPERTY = "ID";
```
spena, line 24: I chose `ID` as the best name for this property. However, if it is confusing with the `ID` keyword in the schema (i.e. `ID id int`), then we can consider these options.

```
with (id = 'my_id')
with (query_id = 'my_id')
with (name = 'my_id')
with (alias = 'my_id')
```

agavra: I think originally I liked `id`, but now my preference is `query_id` 😂 it gives a little bit more clarity into what exactly it is. cc @colinhicks

ksqldb-common/src/main/java/io/confluent/ksql/properties/with/InsertIntoConfigs.java
```diff
@@ -0,0 +1,39 @@
...
+
+  private static final ConfigDef CONFIG_DEF = new ConfigDef()
+      .define(
+          QUERY_ID_PROPERTY,
+          ConfigDef.Type.STRING,
+          null,
+          ConfigDef.Importance.LOW,
+          "Custom query ID to use for INSERT INTO queries"
```
agavra, line 32: let's have some more detailed documentation here. What is it used for? what happens if it isn't supplied? what are valid values? is it case insensitive?

ksqldb-engine/src/main/java/io/confluent/ksql/engine/QueryIdUtil.java
```diff
@@ -24,18 +24,62 @@
...
+    @Override
+    public String toString() {
+      return prefix;
+    }
+  }
 
   private QueryIdUtil() {
   }
 
+  private static void validateWithQueryId(final String queryId) {
```
agavra, line 62: we should make this a `ConfigDef.Validator` and validate it as part of the config def

ksqldb-engine/src/main/java/io/confluent/ksql/engine/QueryIdUtil.java
```diff
@@ -49,14 +93,21 @@ static QueryId buildId(
       final MetaStore metaStore,
       final QueryIdGenerator idGenerator,
       final OutputNode outputNode,
-      final boolean createOrReplaceEnabled) {
+      final boolean createOrReplaceEnabled,
+      final Optional<String> withQueryId) {
+    if (withQueryId.isPresent()) {
```
agavra, line 98: should we also validate to make sure they don't register the same queryID twice? (please add a test for this)

ksqldb-parser/src/main/java/io/confluent/ksql/parser/tree/QueryContainer.java
```diff
@@ -33,4 +35,11 @@
    * @return the sink info.
    */
   Sink getSink();
+
+  /**
+   * Return an optional query ID if specified by the user
```
agavra, line 40: let's specify here that this is only intended to be used by `INSERT INTO` - that way we can remove it when we remove `INSERT INTO`

ksqldb-engine/src/test/java/io/confluent/ksql/engine/KsqlEngineTest.java
```diff
@@ -347,6 +347,31 @@ public void shouldThrowWhenInsertIntoSchemaDoesNotMatch() {
         is("insert into bar select orderTime, itemid from orders;")));
   }
 
+  @Test
+  public void shouldExecuteInsertIntoWithCustomQueryId() {
```
agavra, line 351: Can we add a test in `RecoveryTest` to make sure we can recover create/terminate with custom query IDs?

agavra (approved):
LGTM! Some minor comments inline

ksqldb-engine/src/main/java/io/confluent/ksql/engine/QueryIdUtil.java
spena: I initially thought of this, but I decided to do the validation in the `QueryIdUtil` to make sure that whoever wants to use that class, then they pass the right QueryID string. I don't think nobody else will use it, but I think is a better design as the `buildId` is the one will check IDs are correct.

Unless you have another reason to use the Validator?

ksqldb-engine/src/main/java/io/confluent/ksql/engine/QueryIdUtil.java
spena: Done. I added the validation in the `EngineExecutor` 'cause is the one that knows about IDs by calling the `engineContext.getPersistentQuery`.

ksqldb-engine/src/test/java/io/confluent/ksql/engine/KsqlEngineTest.java
spena: Done

ksqldb-parser/src/main/java/io/confluent/ksql/parser/tree/QueryContainer.java
spena: Done

ksqldb-common/src/main/java/io/confluent/ksql/properties/with/InsertIntoConfigs.java
spena: Done

spena:
Thanks @agavra. I changed the `ID` to `QUERY_ID`. I liked it better too.

ksqldb-engine/src/main/java/io/confluent/ksql/engine/QueryIdUtil.java
agavra: yeah that makes sense

PR #6570, opened 2020-11-03
chore: improve logging around query error type

agavra:
### Description 
We should always log the error type in our logs so that if we know what type the error was if metrics don't properly get emitted.

### Testing done 
Logging-only change

### Reviewer checklist


vcrfxia (approved):
LGTM. Thanks!

PR #6572, opened 2020-11-04
fix: allow reserved keywords on variables names

spena:
### Description 
Issue: The `DEFINE/UNDEFINE` statements do not allow variables names that matches the symbolic names found in `SqlBase.g4`. For instance `define topic = 't1';` fails because `topic` is a reserved keyword. However, these should not be a problem for variables names. We  should allow any variable name.

The `SqlBase.g4` has a rule for `nonReservedKeyword` which I used initially, but this is not scalable. There are several missing keywords in that rule, and not all statements allow the same keywords as non-reserved. To scale this rule, I added an extra validation to the parser error listener to check if a syntax error is caused by a keyword being used as variable name. If so, then it returns silently without throwing any error, thus allowing the `define` and `undefine` statements to continue.

i.e.
```
# BEFORE
ksql> define topic = 'a';
line 1:8: mismatched input 'topic' expecting {'EMIT', 'CHANGES', 'FINAL', 'ESCAPE', 'INTEGER', 'DATE', 'TIME', 'TIMESTAMP', 'INTERVAL', 'YEAR', 'MONTH', 'DAY', 'HOUR', 'MINUTE', 'SECOND', 'ZONE', 'PARTITION', 'STRUCT', 'EXPLAIN', 'ANALYZE', 'TYPE', 'TYPES', 'SHOW', 'TABLES', 'COLUMNS', 'COLUMN', 'PARTITIONS', 'FUNCTIONS', 'FUNCTION', 'ARRAY', 'MAP', 'SET', 'RESET', 'SESSION', 'KEY', 'SINK', 'SOURCE', 'PRIMARY', 'REPLACE', 'ASSERT', 'ADD', 'ALTER', 'IF', IDENTIFIER}
Statement: define topic = 'a';
Caused by: line 1:8: mismatched input 'topic' expecting {'EMIT', 'CHANGES',
	'FINAL', 'ESCAPE', 'INTEGER', 'DATE', 'TIME', 'TIMESTAMP', 'INTERVAL', 'YEAR',
	'MONTH', 'DAY', 'HOUR', 'MINUTE', 'SECOND', 'ZONE', 'PARTITION', 'STRUCT',
	'EXPLAIN', 'ANALYZE', 'TYPE', 'TYPES', 'SHOW', 'TABLES', 'COLUMNS', 'COLUMN',
	'PARTITIONS', 'FUNCTIONS', 'FUNCTION', 'ARRAY', 'MAP', 'SET', 'RESET',
	'SESSION', 'KEY', 'SINK', 'SOURCE', 'PRIMARY', 'REPLACE', 'ASSERT', 'ADD',
	'ALTER', 'IF', IDENTIFIER}
Caused by: org.antlr.v4.runtime.InputMismatchException

# AFTER 
ksql> define topic = 'a';

ksql> show variables;

 Variable Name | Value 
-----------------------
 topic         | a     
-----------------------

ksql> undefine topic;
```

This fix allowed me to removed the undesired keywords from the error message.
```
sql> define 1 = 'a';
line 1:8: mismatched input '1' expecting IDENTIFIER
Statement: define 1 = 'a';
Caused by: line 1:8: mismatched input '1' expecting IDENTIFIER
Caused by: org.antlr.v4.runtime.InputMismatchException
```

The  validator is on `SyntaxErrorValidator`, which is used as an error listener for the parsers. I just copied the `syntaxError` implementation from `DefaultKsqlParser` to this class. We can add more validators in the future if we find another error that should be valid and is difficult to add as rule in `SqlBase.g4`. We can also add validators to format the error messages for a better user experience.

### Testing done 
Added unit tests
Verified manually

### Reviewer checklist

### DIFF
--- a/ksqldb-parser/src/main/antlr4/io/confluent/ksql/parser/SqlBase.g4
+++ b/ksqldb-parser/src/main/antlr4/io/confluent/ksql/parser/SqlBase.g4
@@ -343,7 +343,6 @@ identifier
 
 variableName
     : IDENTIFIER
-    | nonReserved
     ;
 
 variableValue
--- a/ksqldb-parser/src/main/java/io/confluent/ksql/parser/DefaultKsqlParser.java
+++ b/ksqldb-parser/src/main/java/io/confluent/ksql/parser/DefaultKsqlParser.java
@@ -19,19 +19,14 @@ import io.confluent.ksql.metastore.TypeRegistry;
 import io.confluent.ksql.parser.SqlBaseParser.SingleStatementContext;
 import io.confluent.ksql.parser.exception.ParseFailedException;
 import io.confluent.ksql.parser.tree.Statement;
-import io.confluent.ksql.util.ParserUtil;
 import java.util.List;
 import java.util.function.Function;
-import java.util.regex.Matcher;
 import java.util.stream.Collectors;
 import org.antlr.v4.runtime.BaseErrorListener;
 import org.antlr.v4.runtime.CharStream;
 import org.antlr.v4.runtime.CharStreams;
 import org.antlr.v4.runtime.CommonTokenStream;
 import org.antlr.v4.runtime.ParserRuleContext;
-import org.antlr.v4.runtime.RecognitionException;
-import org.antlr.v4.runtime.Recognizer;
-import org.antlr.v4.runtime.Token;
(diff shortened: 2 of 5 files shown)


agavra (approved):
lgtm!

PR #6574, opened 2020-11-04
feat: add CLI parameter to define variables

spena:
### Description 
Add a new CLI parmeter `--define | -d` that allows users to pre-define variables when executing the CLI. This is useful when combining variables with the `-e` or `-f`.

Example:
```
$ ./bin/ksql -d env=qa -e 'show variables;'

 Variable Name | Value 
-----------------------
 env           | qa    
-----------------------
```

Users can define multiple variables using multiple `--define | -d` parameters:
```
$ ./bin/ksql -d env=qa -d streamName=s1 -e 'show variables;'

 Variable Name | Value 
-----------------------
 env           | qa    
 streamName    | s1    
-----------------------
```

### Testing done 
Added unit tests
Verified manually

### Reviewer checklist


ksqldb-cli/src/main/java/io/confluent/ksql/cli/Options.java
```diff
@@ -127,6 +133,12 @@
       description = "Execute commands from a file and exit.")
   private String scriptFile = null;
 
+  @SuppressWarnings("unused") // Accessed via reflection
+  @Option(
+      name = {DEFINE_OPTION, DEFINE_SHORT_OPTION},
+      description = "Define variables.")
```
agavra, line 139: should we add a little bit more description here? Maybe something like "equivalent to the `DEFINE` CLI command"

agavra (approved):
lgtm

Issue #6441, opened 2020-10-15
QTT should support partial schema inference


agavra:
We generate the topic information from the CT/CS statement if there is no topic node in the QTT. The problem is, in the case when both key and value are AVRO, that if we have either the key or the value schema we need both in the topic node because otherwise it won't generate the topic from the CT statement and use the one supplied (and we get a 404 when looking up the schema in SR). 

The test below fails with 404 from SR unless we add in the valueSchema into the topic node.
```json
    {
      "name": "BOOLEAN - key - inference",
      "properties": {
        "ksql.key.format.enabled": true
      },
      "statements": [
        "CREATE STREAM INPUT (foo INT) WITH (kafka_topic='input_topic', format='AVRO');",
        "CREATE STREAM OUTPUT AS SELECT * FROM INPUT;"
      ],
      "topics": [
        {
          "name": "input_topic",
          "keySchema": {"type": "boolean"},
          "keyFormat": "AVRO"
        }
      ],
      "inputs": [
        {"topic": "input_topic", "key": true, "value": {"FOO": 10}},
        {"topic": "input_topic", "value": null}
      ],
      "outputs": [
        {"topic": "OUTPUT", "key": true, "value": {"FOO": 10}},
        {"topic": "OUTPUT", "value": null}
      ]
    },
```

_Originally posted by @agavra in https://github.com/confluentinc/ksql/pull/6432#discussion_r505647275_

fjbecerra:
Hi @agavra , I just raised a PR on this issue, Im happy to be assigned. thanks

agavra:
Thanks @fjbecerra - appreciate it! I'll take a look at your PR today

PR #6600, opened 2020-11-12
chore: support partial schema inference in QTT (#6441)

fjbecerra:
### Description
Implements #6441

Registering schema defined at CS/CT statement in SR when inferring schema in topic, so that avoid the need to set a dummy key/value schema.

Current code:

As the comments advise, there three places from which `allTopics` map get populated:

1. add all topics from topic nodes to the map
2. Infer topics if not added already
3. Get topics from inputs and outputs fields

This concept is also supported [schema-inference](https://docs.ksqldb.io/en/latest/concepts/schemas/#schema-inference)

As per **CR with/without key**, the doc mentions, when sending **message-values**, `The schema must be registered in Schema Registry under the subject pageviews-avro-topic-value`, therefore these scenarios must declare **schema-values** in `topic-node` to replicate SR, however in `avro.json`, **schema-value** name has mistakenly been named as `ignored` in these cases.

At Point 2, key/value schema is available, if any,  yet the logic unintentionally loses them.

Proposed solution:

To keep it simple, at point 2,  the key/value schema of the topic if already present in `allTopic` is updated with the info given by `createTopicFromStatement`, as long as the key or value schema in topic-node isn't present.

### Testing done
Existing tests in `avro.json` all tests are for CS
Unaware of any test covering CT for this, please advise.

### Reviewer checklist

### DIFF
--- a/ksqldb-functional-tests/src/main/java/io/confluent/ksql/test/tools/TestCaseBuilderUtil.java
+++ b/ksqldb-functional-tests/src/main/java/io/confluent/ksql/test/tools/TestCaseBuilderUtil.java
@@ -125,6 +125,18 @@ public final class TestCaseBuilderUtil {
     for (String sql : statements) {
       final Topic topicFromStatement = createTopicFromStatement(sql, metaStore, ksqlConfig);
       if (topicFromStatement != null) {
+        allTopics.computeIfPresent(topicFromStatement.getName(), (key, topic) -> {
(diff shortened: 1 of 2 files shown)


fjbecerra:
[clabot:check]

agavra (approved):
Thanks @fjbecerra! Simple and elegant fix for the issue.

PR #6601, opened 2020-11-12
chore: refuse INSERT VALUES that would modify key schema

agavra:
### Description 

Prevents `INSERT VALUES` from overwriting the schema in schema registry.

### Testing done 

- unit tests
- manual testing to make sure that we can still insert values if ksqlDB created the schema via `CREATE STREAM/TABLE`

### Reviewer checklist

### DIFF
--- a/ksqldb-engine/src/main/java/io/confluent/ksql/engine/InsertValuesExecutor.java
+++ b/ksqldb-engine/src/main/java/io/confluent/ksql/engine/InsertValuesExecutor.java
@@ -16,6 +16,8 @@
 package io.confluent.ksql.engine;
 
 import com.google.common.annotations.VisibleForTesting;
+import io.confluent.kafka.schemaregistry.ParsedSchema;
+import io.confluent.kafka.schemaregistry.client.SchemaMetadata;
 import io.confluent.kafka.schemaregistry.client.rest.exceptions.RestClientException;
 import io.confluent.ksql.GenericRow;
 import io.confluent.ksql.KsqlExecutionContext;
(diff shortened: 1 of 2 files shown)


spena (approved):
lgtm

ksqldb-engine/src/main/java/io/confluent/ksql/engine/InsertValuesExecutor.java
```diff
@@ -266,12 +274,52 @@ private void throwIfDisabled(final KsqlConfig config) {
...
+      return;
+    }
+
+    final ParsedSchema schema = format
+        .getSchemaTranslator(keyFormat.getFormatInfo().getProperties())
+        .toParsedSchema(keySchema);
+
+    try {
+      final SchemaMetadata latest = serviceContext.getSchemaRegistryClient()
+          .getLatestSchemaMetadata(
```
vcrfxia, line 308: How come we're not using `SchemaRegistryUtil#getLatestSchema()` here?

ksqldb-engine/src/main/java/io/confluent/ksql/engine/InsertValuesExecutor.java
```diff
@@ -266,12 +274,52 @@ private void throwIfDisabled(final KsqlConfig config) {
...
+              KsqlConstants.getSRSubject(dataSource.getKafkaTopicName(), true));
+      if (!latest.getSchema().equals(schema.canonicalString())) {
```
vcrfxia, line 310: Are we handling the case where no schema is present in Schema Registry? I know this doesn't happen under normal circumstances since ksqlDB registers schemas at source creation time, but we should handle potential subsequent deletion by the user gracefully rather than throwing an NPE.

agavra: I thought it throws an exception if the schema does not exist with a 404 but I will double check that (so it would be handled in the catch block below)

ksqldb-engine/src/main/java/io/confluent/ksql/engine/InsertValuesExecutor.java
agavra: umm, I wasn't aware that existed!

ksqldb-engine/src/main/java/io/confluent/ksql/engine/InsertValuesExecutor.java
agavra: It appears that your comment above handles this case 😂

ksqldb-engine/src/main/java/io/confluent/ksql/engine/InsertValuesExecutor.java
agavra: also 🤦 seems like I created it...

agavra: https://github.com/confluentinc/ksql/pull/6632

Issue #6341, opened 2020-10-01
CollectListUdaf doesn't undo values correctly due to limit


AlanConfluent:
**Describe the bug**
Suppose we have two tables:

```
CREATE TABLE PAGE_TO_LATEST_CLICK (
    page_id VARCHAR PRIMARY KEY,
    user_id VARCHAR,
    url VARCHAR
  ) WITH (
    KAFKA_TOPIC = 'page_clicks',
    VALUE_FORMAT='JSON'
);

CREATE TABLE URLS_BY_USER  AS
SELECT 
    user_id,
    COLLECT_LIST(url) AS urls
FROM PAGE_TO_LATEST_CLICK
GROUP BY user_id;
```

Suppose I already have 1000 values for user_id "B" in an aggregated list already in `URLS_BY_USER`, and it's therefore at its limit.  Incoming `PAGE_TO_LATEST_CLICK` records `{page_id: "p1", user_id: "B", url: "foo"}` and `{page_id: "p1", user_id: "B", value:"bar"}` then arrive and the "B" aggregate must be updated in `URLS_BY_USER`.  It tries to add `{user_id: "B", url: "foo"}`, but since it's at the limit, it gets dropped.  Now it tries to add `{user_id: "B", url:"bar"}`, which results in an undo of `{user_id: "B", url: "foo"}` since they are for the same page.  Since "foo" was never added the following snippet of code results in an exception because `aggregateValue.lastIndexOf(valueToUndo)` is -1:

```
      @Override
      public List<T> undo(final T valueToUndo, final List<T> aggregateValue) {
        aggregateValue.remove(aggregateValue.lastIndexOf(valueToUndo));
        return aggregateValue;
      }
```

**To Reproduce**

Input data for the same user and different page ids until the limit is reached, 1000 times.  Then add two new entries for both that same user and a new page.   It should result in an `ArrayIndexOutOfBoundsException`

**Expected behavior**
No exception.

I'm assuming a solution might be as simple as updating `CollectListUdaf`, given my understanding of `TableUdaf`:
```
@Override
public List<T> undo(final T valueToUndo, final List<T> aggregateValue) {
   int index = aggregateValue.lastIndexOf(valueToUndo);
   if (index < 0) {
     return aggregateValue;
   }
   aggregateValue.remove(index);
   return aggregateValue;
}
```

[...]

PR #6603, opened 2020-11-12
fix: Check for index before removing value in undo of COLLECT_LIST

AlanConfluent:
### Description 
Fixes https://github.com/confluentinc/ksql/issues/6341

Adds a check for the index of the last value and does nothing if it's not found.

### Testing done 
Ran unit tests.

### Reviewer checklist


stevenpyzhang (approved):
lgtm

stevenpyzhang:
Let's target this to 5.5.x (or maybe even older depending on when the function was introduced).

Issue #5299, opened 2020-05-06
Add datetime/timestamp type support


mikebin:
It would be useful to add support for a datetime/timestamp type in ksqlDB which can be represented as a timestamp logical type (`timestamp-millis`) in Avro, similar to how ksqlDB `DECIMAL` can be represented as a logical decimal type in Avro.

Currently, a timestamp can be represented in ksqlDB as a `BIGINT`, but when using Avro, if the source field is of logical type `timestamp-millis`, the logical type information is lost when propagated to the sink through a query, and it also doesn't seem possible to construct a return type manually in a UDF which contains the logical type information.

agavra:
closing as duplicate of #4148

Issue #4148, opened 2019-12-16
Introduce TIMESTAMP or DATETIME types


big-andy-coates:
Potentially breaking change...

We should look to switch ROWTIME to be a `TIMESTAMP` or `DATETIME` type, rather than a `BIGINT`.

agavra:
related to #5299

agavra:
Adding this as a p0 to help enable improved connect integration

big-andy-coates:
@agavra may help to add details of _why_ this would help improve connect integration :D

agavra:
Fair enough! Many DBs have columns that use `TIMESTAMP`/`DATETIME` and we convert it usually to just long, which causes issues when we then sink back into a database. It doesn't necessarily mean we need to change `ROWTIME` to those formats, but I was piggy-backing on this ticket as a "let's introduce these new formats".

jzaralim:
Closed with:

https://github.com/confluentinc/ksql/pull/6649
https://github.com/confluentinc/ksql/pull/6806
https://github.com/confluentinc/ksql/pull/6852
https://github.com/confluentinc/ksql/pull/6853
https://github.com/confluentinc/ksql/pull/6901
https://github.com/confluentinc/ksql/pull/6927
https://github.com/confluentinc/ksql/pull/6948

PR #6605, opened 2020-11-12
New UDF for epoch time conversion to different time zones

k1th:
New UDF for epoch time conversion to different time zones to enable windowing by non UTC aligned days.

### Description 
One new UDF class 
io.confluent.ksql.function.udf.datetime.EpochTimeManipulator

### Testing done 
Unit test in io.confluent.ksql.function.udf.datetime.EpochTimeManipulatorTest

### Reviewer checklist

### DIFF
--- a/ksqldb-common/src/main/java/io/confluent/ksql/schema/Operator.java
+++ b/ksqldb-common/src/main/java/io/confluent/ksql/schema/Operator.java
@@ -15,6 +15,7 @@
 
 package io.confluent.ksql.schema;
 
+import static io.confluent.ksql.util.DecimalUtil.toSqlDecimal;
 import static java.util.Objects.requireNonNull;
 
 import io.confluent.ksql.schema.ksql.types.SqlBaseType;
@@ -66,7 +67,7 @@ public enum Operator {
           return right;
         }
 
-        return binaryResolver.apply(toDecimal(left), (SqlDecimal) right);
+        return binaryResolver.apply(toSqlDecimal(left), (SqlDecimal) right);
       }
 
       if (right.baseType().canImplicitlyCast(left.baseType())) {
@@ -74,25 +75,11 @@ public enum Operator {
           return left;
         }
 
-        return binaryResolver.apply((SqlDecimal) left, toDecimal(right));
+        return binaryResolver.apply((SqlDecimal) left, toSqlDecimal(right));
       }
     }
 
     throw new KsqlException(
         "Unsupported arithmetic types. " + left.baseType() + " " + right.baseType());
-  }
-
-  private static SqlDecimal toDecimal(final SqlType type) {
-    switch (type.baseType()) {
-      case DECIMAL:
-        return (SqlDecimal) type;
-      case INTEGER:
-        return SqlTypes.INT_UPCAST_TO_DECIMAL;
-      case BIGINT:
-        return SqlTypes.BIGINT_UPCAST_TO_DECIMAL;
-      default:
-        throw new KsqlException(
-            "Cannot convert " + type.baseType() + " to " + SqlBaseType.DECIMAL + ".");
-    }
   }
 }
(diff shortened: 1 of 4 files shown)


agavra:
Thanks for the pull request @k1th! In the upcoming release (0.15) we're planning on adding a `DATETIME` type to ksql (see #5299 and #4148) - it might make sense to hold off on this PR in order to align with the more standard sql syntax for timestamp time zone conversion: https://docs.oracle.com/cd/B19306_01/server.102/b14225/ch4datetime.htm#i1007699 and https://docs.microsoft.com/en-us/sql/t-sql/queries/at-time-zone-transact-sql?view=sql-server-ver15

```sql
-- example taken from SQL server docs
SELECT CONVERT(DATETIME2(0), '2015-03-29T01:01:00', 126)     
AT TIME ZONE 'Central European Standard Time';  
```

In the meantime, if you need this to unblock some product development, you can always roll this as a custom UDF! Let me know if I'm not understanding the functionality of this UDF correctly.

k1th:
We are already using this in our deployments as custom UDF, so we are not really waiting for the PR. 

Regarding the why:
KSQL window operations are aligned to the epoch which is always in UTC. This means it’s impossible to create windows in other timezones.
Because we need windows in Berlin time, we have to modify the rowtime to be a „corrected“ epoch that is offset by the timezone difference at the rowtime point in time. 
The UDF does this in a daylight saving safe way. 
Introducing the DATETIME type won‘t change the window alignment issue.

agavra:
I'm not sure I understand why having the `DATETIME` format along side `AT TIME ZONE` is any different than this UDF? They both allow us to convert one timestamp to another in a different time zone (daylight savings aware)

k1th:
Will KSQL windowing then align to UTC (which is probably the format the datetime will be stored in) or will it align to the time in the selected timezone?

agavra:
The process would be the same as you have with the UDF - you would need to create the new timestamps "corrected" under epoch time and window using those. So it still windows based on UTC time (but if I understand correctly, that's the case with the UDF as well).

k1th:
It would, depending very much on the implementation of rowtime as `DATETIME` with TZs.

- `ROWTIME` implemented as as timestamp with usual DB like semantics: Internally this would be stored as UTC and only upon query be transformed in the user‘s selected TZ. This would need additional syntax for the `WINDOW` clause to allow for selection of TZ as well. 
- `ROWTIME` implemented as a timestamp already converted with the TZ offset. This would do what we want, but not be very DB like. When, as a user, choosing to use another TZ this would mean changing the underlying data by copying into another topic, probably. 

For me, the former would be a better choice when introduced a new `DATETIME` type, but would require the change to the `WINDOW` clause...

k1th:
KSQL 0.18 just came out providing similar functionality, although with the added step of converting to `TIMESTAMP` and back to `BIGINT` like this:

```
SELECT
      TRACKING_DATA[1]->EVENT->TS `event_tsstr`,
      PARSE_TIMESTAMP(TRACKING_DATA[1]->EVENT->TS, 'yyyy-MM-dd''T''HH:mm:ss.SSSX') `event_time`,
      CONVERT_TZ(PARSE_TIMESTAMP(TRACKING_DATA[1]->EVENT->TS, 'yyyy-MM-dd''T''HH:mm:ss.SSSX'), 'UTC', 'Europe/Berlin') `ts_with_offset`,
      UNIX_TIMESTAMP(CONVERT_TZ(PARSE_TIMESTAMP(TRACKING_DATA[1]->EVENT->TS, 'yyyy-MM-dd''T''HH:mm:ss.SSSX'), 'UTC', 'Europe/Berlin')) `epoch_with_offset`,
      TRACKING_DATA[1]->IDENTIFIERS[1]->ID `usage_id`
    FROM MY_EVENTS emit changes limit 1;
```

Do you think we should discard the PR in favor of that?

k1th:
closing because this can be done with built-in methods in 0.18

Issue #6332, opened 2020-09-30
[SR Support]: Figure out how to join on keys where the key contains the schema id


big-andy-coates:
Any format that supports schema inference will be adding the magic byte and the schema id as the first few bytes of the key, i.e. the serialized key is `<magic_byte><schema-id><serialized-key>`.

Without doing anything special, for a join to work we'd need the serialized-key _and_ the schema-id to match!!!!

ksqlDB only supports topic-level subjects.   The two topics being joined, let's call them `left` and `right`, will have their schema's registered under the subjects `left-key` and `right-key`.   If two keys have the same schema, do they share the same id?  If they don't, then we've got a problem!

big-andy-coates:
Knocked up some quick test code to register the same key schema against multiple topics... they are get the same id. This is good!

However, there's still scope for joins not to work where the schema differs only slightly, e.g. the name of the AVRO record, which doesn't affect the serialized bytes, but would change the schema id.

Hence, we should probably have some kind of check when executing a join to ensure there is at least some overlap between the schema ids of the key schemas of each side. 

I say some overlap, as its possible for each subject to have multiple schema versions associated with it, each with its own id.  It could be that some ids match and some don't.  This should probably be enough to allow the join, as the mismatched ids may not be in use, e.g. they could be old data or maybe there from when a mismatch in schemas was corrected, etc.

Ideally, we could return:
 * An error, if there is no overlap of schema ids between the two sides
 * A warning, if there is some overlap, but some differences.
 * Success, if ids on both sides match.

vcrfxia:
Spent some time looking into this. Turns out it's rather challenging to check whether the two sides of the join have the same schema ID or not. It's not just a matter of looking up the topics on both sides of the join in Schema Registry since in the case of pre-join repartition topics (or other internal topics prior to the join), there will be no schema in SR since ksqlDB doesn't create these subjects until after the join is validated (and started) which makes sense. Inferring what the key schema will be before it's actually registered is also non-trivial since it depends on details of the physical plan, such as serde options, which the logical plan does not (and should not) have. As a result, we've decided to go the route of forcing repartitions prior to the join in order to ensure matching schemas as this is preferable to either disallowing joins on SR-enabled key formats or allowing them with potential silent join misses. This approach is implemented in https://github.com/confluentinc/ksql/pull/6635.

[...]

PR #6609, opened 2020-11-13
chore: set key schema name to avoid conflicts

### LINKED ISSUE #6465: [AVRO SUPPORT] Avoid record name clashes
The value schema generated for Avro values has a constant default record name of `io.confluent.ksql.avro_schemas.KsqlDataSourceSchema`.

This is lame. 

The record name becomes the Java class name if anyone downstream generated Java types from the schema. (The same is similar for other programming languages which can generate types from Avro schemas too).

So if a downstream consumer of ksqlDB created topics wants to generate java types and consume multiple topics they run into issues because all the schemas generate a class called `KsqlDataSourceSchema`.

To get around this issue, way back when, support for overriding the default name was added via the `WITH(VALUE_AVRO_SCHEMA_FULL_NAME='some.full.name')`.

This is slightly less lame.

With the support for AVRO keys, the default behaviour will be for both the key and value schemas to generate types called `KsqlDataSourceSchema`, meaning a downstream app can't even generate types so it can deserialize both the key and the value from a single topic without running into a name clash!  

We could introduce `KEY_AVRO_SCHEMA_FULL_NAME`.  Quickest solution... and lame.

I propose we default the name of the record in the key schema to `<sourceNameInCamelCase>Key`, e.g. for a source named `BOB`, the record in the key schema would be called `BobKey`.

Note: we want camel casing here!

**Be use to add test cases with nests STRUCT and MAP types as these generate nested records - which also need decent and unique names.**

Given value schema properties are part of the query plan, we can also change the value schemas so they too benefit from this lame-removing change!!!!

agavra:
### Description 

First two commits fixes #6465 by always setting the key format info to have an avro schema name, and set that avro schema name to something unique to the source. The approach in this PR is to simply add the `fullSchemaName` property into the `FormatInfo` for all AVRO keys. In the future, if we want to introduce functionality that allows users to set the name for their schema, we can have it only override the `fullSchemaName` if it's not present.

This commit will fail until https://github.com/confluentinc/ksql/pull/6708 is reviewed and accepted. When it is accepted, I will merge them as two separate commits simultaneously. 

### Testing done 

Updated unit tests, checked that e2e flow was working by check schema registry:
```
~/dev/releases/confluent-6.0.1-SNAPSHOT ❯ curl -XGET 'http://localhost:8081/subjects/e-key/versions/1'                            12:48:06
{"subject":"e-key","version":1,"id":4,"schema":"{\"type\":\"record\",\"name\":\"avro_schemas.E_key\",\"namespace\":\"io.confluent.ksql\",\"fields\":[{\"name\":\"F1\",\"type\":[\"null\",\"int\"],\"default\":null}],\"connect.name\":\"io.confluent.ksql.avro_schemasE_key\"}"}%
```

### Reviewer checklist


vcrfxia:
Hey @agavra what's your plan for how this fits in with joins on Avro sources? Currently we disallow joins if the format properties don't match: https://github.com/confluentinc/ksql/blob/9262cf47518b1a5378d8b3b9a6b15510809d6680/ksqldb-engine/src/main/java/io/confluent/ksql/planner/plan/JoiningNode.java#L86 IIUC, the change in this PR would mean all table-table joins on Avro keys would be disallowed. With the work I'm doing for https://github.com/confluentinc/ksql/issues/6332 we could make an exception for this particular format property since we're planning to force repartitions on sources with SR-enabled key formats anyway (assuming we ensure that the internal repartition uses the same Avro schema name on both sides of the join), but that feels hacky.

agavra:
@vcrfxia - I think we may need to establish a way for us to make exceptions (or perhaps replace the current deny-all logic with a denylist of format properties) anyway. 

Imagine we want to introduce an equivalent to `VALUE_AVRO_SCHEMA_FULL_NAME` for keys (e.g. `KEY_AVRO_SCHEMA_FULL_NAME`) to allow users to specify the key schema name. That definitely belongs in the format properties as `fullSchemaName`, but would we not want to allow joins for it? In general, I think the current logic made sense as a stop-gap safety measure which allowed us to punt thinking about it to another time, not as a rule that we necessarily want to follow.

ksqldb-engine/src/main/java/io/confluent/ksql/analyzer/Analyzer.java
```diff
@@ -169,14 +169,18 @@ private void analyzeNonStdOutSink(final Sink sink) {
           .getDataSource()
           .getKsqlTopic();
 
-      final FormatInfo keyFmtInfo = buildKeyFormatInfo(
```
vcrfxia, line 172: Are the changes in this file purely a refactor to have separate helpers for getting format name and info, or are there functional differences here?

ksqldb-parser/src/main/java/io/confluent/ksql/parser/properties/with/CreateSourceProperties.java
```diff
@@ -131,10 +132,22 @@ public String getKafkaTopic() {
...
     final String keyFormat = getFormatName()
         .orElse(props.getString(CommonCreateConfigs.KEY_FORMAT_PROPERTY));
-    return Optional.ofNullable(keyFormat).map(format -> FormatInfo.of(format, ImmutableMap.of()));
+    return Optional.ofNullable(keyFormat)
+        .map(format -> FormatInfo.of(format, getKeyFormatProperties(keyFormat, name.text())));
+  }
+
+  private Map<String, String> getKeyFormatProperties(final String keyFormat, final String name) {
+    if (AvroFormat.NAME.equalsIgnoreCase(keyFormat)) {
+      // ensure that the schema name for the key is unique to the sink - this allows
```
vcrfxia, line 144:
```suggestion
      // ensure that the schema name for the key is unique to the source - this allows
```

ksqldb-parser/src/test/java/io/confluent/ksql/parser/properties/with/CreateSourceAsPropertiesTest.java
```diff
@@ -113,6 +113,21 @@ public void shouldThrowOnInvalidTimestampFormat() {
...
+    );
+
+    // When:
+    final String avroSchemaName = properties.getKeyFormatProperties("name", AvroFormat.NAME)
+        .get(AvroFormat.FULL_SCHEMA_NAME);
+
+    // Then:
+    assertThat(avroSchemaName, is("io.confluent.ksql.avro_schemas.name_key"));
+  }
+
```
vcrfxia, line 130: Add a quick test to ensure the property isn't set for non-Avro sinks?

ksqldb-parser/src/test/java/io/confluent/ksql/parser/properties/with/CreateSourcePropertiesTest.java
```diff
@@ -490,10 +492,26 @@ public void shouldGetKeyAndValueFormatFromFormat() {
             .build());
 
     // When / Then:
-    assertThat(props.getKeyFormat().get().getFormat(), is("KAFKA"));
+    assertThat(props.getKeyFormat(SourceName.of("foo")).get().getFormat(), is("KAFKA"));
     assertThat(props.getValueFormat().get().getFormat(), is("AVRO"));
   }
 
+  @Test
+  public void shouldGetKeyFormatFromAvroFormatWithAvroName() {
```
vcrfxia, line 500: nit:
```suggestion
  public void shouldSetAvroNameOnAvroKeyFormat() {
```
unless there's a separate test to verify the same behavior with `KEY_FORMAT='AVRO'` rather than `FORMAT='AVRO'`, I think that bit makes the name harder to grok rather than easier.

ksqldb-serde/src/main/java/io/confluent/ksql/serde/avro/AvroFormat.java
```diff
@@ -83,4 +83,8 @@ protected ConnectSchemaTranslator getConnectSchemaTranslator(
     return new KsqlAvroSerdeFactory(schemaFullName)
         .createSerde(connectSchema, config, srFactory, targetType, isKey);
   }
+
+  public static String getKeySchemaName(final String name) {
+    return AvroProperties.AVRO_SCHEMA_NAMESPACE + "." + name + "_key";
```
vcrfxia, line 88: The linked issue (https://github.com/confluentinc/ksql/issues/6465) suggests camel casing -- any reason we're not doing that?

vcrfxia (approved):
LGTM with one point of clarification: does the name specified by `AVRO_SCHEMA_FULL_NAME` get embedded into nested schema names (e.g., in the case of structs)? I assume so but want to check. Ideally we'd have a test case with a struct type to ensure the behavior here is sane.

ksqldb-engine/src/main/java/io/confluent/ksql/analyzer/Analyzer.java
agavra: yes, just a refactor because it made it easier to share the `buildFormatInfo` method across the key/value

ksqldb-serde/src/main/java/io/confluent/ksql/serde/avro/AvroFormat.java
agavra: Good call - at first I was worried that it would be hard to determine the original casing and convert it to Camel case  (e.g. `FOO_BAR` vs `FooBar` vs `FOOBAR` as the input) but if we're generating java classes we probably want to err on the side of aggressively turning it into java-friendly class names. I've added a test to demonstrate this

agavra:
> LGTM with one point of clarification: does the name specified by AVRO_SCHEMA_FULL_NAME get embedded into nested schema names (e.g., in the case of structs)?

I will add the following test to the next PR:
```json
{
      "name": "partition by nested struct",
      "statements": [
        "CREATE STREAM INPUT (ID INT KEY, VAL STRUCT<F1 INT, F2 INT>) with (kafka_topic='input', key_format='AVRO', value_format='JSON');",
        "CREATE STREAM OUTPUT AS select STRUCT(F1:=VAL), ID from INPUT partition by STRUCT(f1:=VAL);"
      ],
      "inputs": [{"topic": "input", "key": 10, "value": {"VAL": {"F1": 1, "F2": 2}}}],
      "outputs": [{"topic": "OUTPUT", "key": {"F1": {"F1": 1, "F2": 2}}, "value": {"ID": 10}}],
      "post": {
        "topics": {
          "topics": [
            {
              "name": "OUTPUT",
              "keyFormat": { "format": "AVRO", "features": [ "UNWRAP_SINGLES" ], "properties": {"fullSchemaName": "io.confluent.ksql.avro_schemas.OutputKey"}},
              "valueFormat": { "format": "JSON" },
              "keySchema": {
                "type": "record",
                "name": "OutputKey",
                "namespace": "io.confluent.ksql.avro_schemas",
                "fields": [
                  {
                    "name": "F1",
                    "type": [
                      "null",
                      { "type": "record",
                        "name": "OutputKey_F1",
                        "fields": [
                          { "name": "F1", "type": [ "null", "int" ], "default": null },
                          { "name": "F2", "type": [ "null", "int" ], "default": null }
                        ]
                      }
                    ],
                    "default": null
                  }
                ]
              }
            }
          ]
        }
      }
    },
```

agavra:
Closing this pull request. I'll just merge it as part of #6708

PR #6613, opened 2020-11-13
docs: klip-41: ksqlDB .Net Client

alex-basiuk:
### Description 
A proposal for a .Net client

### Testing done 
N/A

### Reviewer checklist


agavra:
Thanks @alex-basiuk - this is really exciting stuff. We haven't really come up with a good way to structure development around ksqlDB clients. Did you have any idea about where you were planning on having the code live? Do you have a prototype or code hosted somewhere publicly that you want us to review? A good way forward on this (the KLIP LGTM) would be to start development as an independent client and then when it gains traction popularity, we can bring it under the ksqlDB "official" umbrella of clients and in the meantime, we can point to it from our README to help get it some traction.

alex-basiuk:
> 
> 
> Thanks @alex-basiuk - this is really exciting stuff. We haven't really come up with a good way to structure development around ksqlDB clients. Did you have any idea about where you were planning on having the code live? Do you have a prototype or code hosted somewhere publicly that you want us to review? A good way forward on this (the KLIP LGTM) would be to start development as an independent client and then when it gains traction popularity, we can bring it under the ksqlDB "official" umbrella of clients and in the meantime, we can point to it from our README to help get it some traction.

Thanks @agavra, it's not publicly available yet, but will be soon. I'll post updates here and will add a link to the README.
One thing which is not entirely clear to me is how are you going to evolve your REST API. You have a notion of version in your schema e.g. `application/vnd.ksql.v1+json`. I guess you'll create new versions in future. But you also support `application/json` which is not versioned at all.

design-proposals/klip-41-ksqldb-.net-client.md
```diff
@@ -0,0 +1,72 @@
...
+## Motivation and background
+
+Basically, it's similar to [KLIP 15](./klip-15-new-api-and-client.md).
+It should lower the barrier for .NET developers and increase adoption of ksqlDB by the community.
+The library should offer the same feature set as the Java library and should follow common .NET conventions and patterns.  
+
+## What is in scope
+
+* A new client implemented in C#.
+* Initially, the client will be targeted at .NET Core 3 and .NET Framework 5. Support of older versions will be added later.
```
MichaelDrogalis, line 21: I'm not versed in .NET, but one thing I'm wondering is how support for multiple versions will look over time. Any ideas about how to manage that workflow?

design-proposals/klip-41-ksqldb-.net-client.md
```diff
@@ -0,0 +1,72 @@
...
+* The client will support execution and streaming of queries (both push and pull), inserting of rows into streams, DDL operations and admin operations
+such as list and describe.
+* The client will utilise C# [async streams](https://docs.microsoft.com/en-us/dotnet/csharp/tutorials/generate-consume-asynchronous-stream) feature.
+* The client will also support a direct / synchronous interaction style.
+* High-quality documentation and examples.
```
MichaelDrogalis, line 26: Super. :) If this is going to live inside of a repo owned by you for a while, we can put the documentation in a README there so it is self-contained.

design-proposals/klip-41-ksqldb-.net-client.md
```diff
@@ -0,0 +1,72 @@
...
+* The client will expose convenience builder extension methods for ASP.NET Core. It should be easy to add and configure the client in ASP.NET Core idiomatic way. 
+  But the mechanism will be optional. Users won't be required to use it or another DI. 
+* The client will have minimal dependencies on third party libraries.
+* The client will follow performance best practices.
+* The client will be thread-safe.
+
+## Test plan
+
+Unit and integration test will be mandatory. Performance benchmarks are desirable but unlikely be the first priority.
+Ideally, tests should be executed on both Windows and Linux platforms.
```
MichaelDrogalis, line 56: Do any CI providers make dual Windows tests easy? Been a while since I've tested much on Windows.

MichaelDrogalis (approved):

design-proposals/klip-41-ksqldb-.net-client.md
alex-basiuk: @MichaelDrogalis .NET 5 and .NET Core 3 are very close in terms of APIs. I don't think .NET 5 has any breaking changes in the libraries required by this project. It won't be hard to support both [until end of support of .NET Core 3.1](https://dotnet.microsoft.com/platform/support/policy/dotnet-core). Basically, you can have one project which [targets multiple framework versions](https://docs.microsoft.com/en-us/dotnet/standard/frameworks).
Support of older versions is probably more involved.

design-proposals/klip-41-ksqldb-.net-client.md
alex-basiuk: @MichaelDrogalis  Travis, CircleCI, Azure CI. Folks from [confluent-kafka-dotnet](https://github.com/confluentinc/confluent-kafka-dotnet) use Travis.

design-proposals/klip-41-ksqldb-.net-client.md
MichaelDrogalis: Cool, that makes sense. Mostly just asking since Im unfamiliar with the .NET stack.

agavra (approved):

agavra:
> One thing which is not entirely clear to me is how are you going to evolve your REST API. You have a notion of version in your schema e.g. application/vnd.ksql.v1+json. I guess you'll create new versions in future. But you also support application/json which is not versioned at all.

cc @vcrfxia might have some input on this, I'm not totally sure what our versioning strategy for the client is going forward

vcrfxia:
> > One thing which is not entirely clear to me is how are you going to evolve your REST API. You have a notion of version in your schema e.g. application/vnd.ksql.v1+json. I guess you'll create new versions in future. But you also support application/json which is not versioned at all.
> 
> cc @vcrfxia might have some input on this, I'm not totally sure what our versioning strategy for the client is going forward

`application/json` refers to the latest version at any given point in time:
https://github.com/confluentinc/ksql/blob/e6ba436536a10678eb9cbafa150c3fd0501d31ab/ksqldb-rest-app/src/main/java/io/confluent/ksql/api/server/ServerVerticle.java#L344-L346
Currently the latest (and only) version is `application/vnd.ksql.v1+json`: https://github.com/confluentinc/ksql/blob/master/ksqldb-rest-model/src/main/java/io/confluent/ksql/rest/entity/KsqlMediaType.java#L31

alex-basiuk:
Thanks @vcrfxia 
Basically, it means that if I use `application/json` in my application, it can break at any time when you upgrade services on your side to a newer version. Why do you need it at all?

Also, I noticed that the `/ksq` endpoint return @type property which is not documented. It would be cleaner to remove it.

vcrfxia:
> Basically, it means that if I use `application/json` in my application, it can break at any time when you upgrade services on your side to a newer version. Why do you need it at all?

The decision to include this option predates me. Perhaps @rodesai has context?

> Also, I noticed that the `/ksql` endpoint return @type property which is not documented. It would be cleaner to remove it.

Thanks for the feedback. The field comes from ksqlDB's use of Jackson subtypes for server response types. ksqlDB 1.0 will likely include pretty significant changes to the server API endpoints. That would be the time to reconsider this, as removing an existing field would be a breaking change.

Issue #6621, opened 2020-11-16
MAP keys serialize intederministically


agavra:
Different implementations of Java's `Map` serialize differently. This means that we could have the same map serialize to two different bytes representations, and therefore get sent to different partitions. This was discovered with the following pull query test:

_Note: this test will fail in master because we have not yet implemented non-literal keys. It was run on a test branch where I am implementing that feature_

```json
    {
      "name": "non-windowed - map keys",
      "properties": {
        "ksql.key.format.enabled": true
      },
      "statements": [
        "CREATE STREAM INPUT (ID MAP<STRING, STRING> KEY, IGNORED INT) WITH (kafka_topic='test_topic', format='JSON');",
        "CREATE TABLE AGGREGATE AS SELECT ID, COUNT(1) AS COUNT FROM INPUT GROUP BY ID;",
        "SELECT ID, COUNT, ROWTIME FROM AGGREGATE WHERE ID=MAP('F2':='0', 'F1':='1');"
      ],
      "inputs": [
        {"topic": "test_topic", "timestamp": 12345, "key": {"F1": "1", "F2": "1"}, "value": {}},
        {"topic": "test_topic", "timestamp": 12365, "key": {"F1": "1", "F2": "0"}, "value": {}}
      ],
      "responses": [
        {"admin": {"@type": "currentStatus"}},
        {"admin": {"@type": "currentStatus"}},
        {"query": [
          {"header":{"schema":"`ID` MAP<STRING, STRING> KEY, `COUNT` BIGINT, `ROWTIME` BIGINT"}},
          {"row":{"columns":[{"F1": "1", "F2": "0"}, 1, 12365]}}
        ]}
      ]
    },
```

The lookup could not succeed because the serialized key did not match the serialized where clause value.

PR #6628, opened 2020-11-17
chore: support complex key pull queries

### LINKED ISSUE #6602: Support "literal expressions" for pull queries on advanced keys
We will soon enable support for structured and other advanced key types (map/array) (#824) but we will not be able to issue pull queries against materialized views that have those as keys. There are two tasks that need to be done in that bucket:

1. support the ability to create structs/maps/arrays in the pull query's where clause (e.g. `WHERE rowkey = STRUCT(foo:=bar)`) so long as every element within the struct/array/map declaration is a literal (or resolvable to a literal value)
1. support the ability to issue pull queries against multi-column keys (e.g. `WHERE colA = 'foo' AND colB = 'bar'`).

This ticket tracks the former of the two.

agavra:
### Description 

fixes #6602 

Supports key-lookups for keys that are not just literals. This is necessary for the generic key work being done because we will expose keys that are not just primitives, but can also be arrays and structs.

Note that maps are not tested because of #6621, we may consider just prohibiting maps from being keys altogether. I will port this over after #6375 is merged.

### Testing done 

- new unit tests
- new RQTT tests

### Reviewer checklist

### DIFF
--- a/ksqldb-engine/src/main/java/io/confluent/ksql/engine/generic/GenericExpressionResolver.java
+++ b/ksqldb-engine/src/main/java/io/confluent/ksql/engine/generic/GenericExpressionResolver.java
@@ -40,7 +40,11 @@ import java.util.function.Supplier;
  * Builds a Java object, coerced to the desired type, from an arbitrary SQL
  * expression that does not reference any source data.
  */
-class GenericExpressionResolver {
+public class GenericExpressionResolver {
+
+  // GenericExpressionResolver doesn't accept any column references, so we don't
+  // actually need the schema, but the CodeGenRunner expects on to be passed in
+  private static final LogicalSchema NO_COLUMNS = LogicalSchema.builder().build();
 
   private static final Supplier<String> IGNORED_MSG = () -> "";
   private static final ProcessingLogger THROWING_LOGGER = errorMessage -> {
@@ -49,21 +53,18 @@ class GenericExpressionResolver {
 
   private final SqlType fieldType;
   private final ColumnName fieldName;
-  private final LogicalSchema schema;
   private final SqlValueCoercer sqlValueCoercer = DefaultSqlValueCoercer.STRICT;
   private final FunctionRegistry functionRegistry;
   private final KsqlConfig config;
 
-  GenericExpressionResolver(
+  public GenericExpressionResolver(
       final SqlType fieldType,
       final ColumnName fieldName,
-      final LogicalSchema schema,
       final FunctionRegistry functionRegistry,
       final KsqlConfig config
   ) {
     this.fieldType = Objects.requireNonNull(fieldType, "fieldType");
     this.fieldName = Objects.requireNonNull(fieldName, "fieldName");
-    this.schema = Objects.requireNonNull(schema, "schema");
     this.functionRegistry = Objects.requireNonNull(functionRegistry, "functionRegistry");
     this.config = Objects.requireNonNull(config, "config");
   }
@@ -80,7 +81,7 @@ class GenericExpressionResolver {
           CodeGenRunner.compileExpression(
               expression,
               "insert value",
-              schema,
+              NO_COLUMNS,
               config,
               functionRegistry
           );
--- a/ksqldb-engine/src/main/java/io/confluent/ksql/engine/generic/GenericExpressionResolver.java
+++ b/ksqldb-engine/src/main/java/io/confluent/ksql/engine/generic/GenericExpressionResolver.java
@@ -20,6 +20,9 @@ import io.confluent.ksql.execution.codegen.CodeGenRunner;
 import io.confluent.ksql.execution.codegen.ExpressionMetadata;
 import io.confluent.ksql.execution.expression.tree.Expression;
 import io.confluent.ksql.execution.expression.tree.NullLiteral;
+import io.confluent.ksql.execution.expression.tree.QualifiedColumnReferenceExp;
+import io.confluent.ksql.execution.expression.tree.TraversalExpressionVisitor;
+import io.confluent.ksql.execution.expression.tree.UnqualifiedColumnReferenceExp;
 import io.confluent.ksql.execution.expression.tree.VisitParentExpressionVisitor;
 import io.confluent.ksql.function.FunctionRegistry;
 import io.confluent.ksql.logging.processing.ProcessingLogger;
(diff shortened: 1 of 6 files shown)


ksqldb-functional-tests/src/test/resources/rest-query-validation-tests/pull-queries-against-materialized-aggregates.json
```diff
@@ -1927,17 +1914,90 @@
       }
     },
     {
-      "name": "IN: fail on non-literal key",
+      "name": "non-windowed - function in where clause",
```
vcrfxia, line 1917: Do we want to support generic expressions in pull query WHERE clauses at this time? Last I spoke to @AlanConfluent it sounded like we wanted to limit support to literal expressions (i.e., those that create arrays and structs with literal values) for now.

agavra: it would be more work to not support this than to support this. Is there any reason why we don't want to?

ksqldb-functional-tests/src/test/resources/rest-query-validation-tests/pull-queries-against-materialized-aggregates.json
```diff
@@ -1927,17 +1914,103 @@
...
       "statements": [
         "CREATE STREAM INPUT (ID INT KEY, IGNORED INT) WITH (kafka_topic='test_topic', value_format='JSON');",
         "CREATE TABLE AGGREGATE AS SELECT ID, COUNT(1) AS COUNT FROM INPUT GROUP BY ID;",
-        "SELECT * FROM AGGREGATE WHERE ID IN (CAST(1 AS INTEGER));"
+        "SELECT * FROM AGGREGATE WHERE ID IN (COUNT + 1);"
       ],
       "expectedError": {
         "type": "io.confluent.ksql.rest.entity.KsqlStatementErrorMessage",
-        "message": "Only comparison to literals is currently supported: (ID IN (CAST(1 AS INTEGER)))",
+        "message": "Unsupported column reference in pull query: (COUNT + 1)",
```
vpapavas, line 1925: What should be supported is functions on literals, maps and arrays of literals but not functions on columns?

agavra: yes, essentially anything static (column references can't be used anywhere)

vpapavas: Discussed offline. In this PR we only want to support non-literal keys (maps and structs) and expressions that use only literals.

ksqldb-functional-tests/src/test/resources/rest-query-validation-tests/pull-queries-against-materialized-aggregates.json
```diff
@@ -1927,17 +1914,103 @@
...
+          {"row":{"columns":[{"F1": "1", "F2": "0"}, 1, 12365]}}
+        ]}
+      ]
+    },
+    {
+      "name": "IN: support non-literal key",
+      "statements": [
+        "CREATE STREAM INPUT (ID INT KEY, IGNORED INT) WITH (kafka_topic='test_topic', value_format='JSON');",
+        "CREATE TABLE AGGREGATE AS SELECT ID, COUNT(1) AS COUNT FROM INPUT GROUP BY ID;",
+        "SELECT * FROM AGGREGATE WHERE ID IN (CAST(10 AS INTEGER));"
```
vpapavas, line 2000: Could we maybe make the IN take more than one key since that is its primary use case? Something like:
```
SELECT * FROM AGGREGATE WHERE ID IN (CAST(10 AS INTEGER), CAST("11" AS INTEGER));
```

vpapavas (approved):
LGTM with minor comment about a test case

ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/execution/PullQueryExecutor.java
```diff
@@ -778,31 +798,40 @@ private static Object extractKeyWhereClause(
...
+      final KsqlConfig config,
+      final Expression errorMessageHint
   ) {
-    if (schema.key().size() != 1) {
-      throw invalidWhereClauseException("Only single KEY column supported", windowed);
+    final Object obj;
+    if (exp instanceof NullLiteral) {
+      obj = null;
+    } else if (exp instanceof Literal) {
+      // skip the GenericExpressionResolver because this is
```
AlanConfluent, line 816: This is just done because `GenericExpressionResolver` isn't as low overhead right?

ksqldb-engine/src/main/java/io/confluent/ksql/engine/generic/GenericExpressionResolver.java
```diff
@@ -76,11 +83,12 @@ public Object resolve(final Expression expression) {
 
     @Override
     protected Object visitExpression(final Expression expression, final Void context) {
+      new EnsureNoColReferences(expression).process(expression, context);
       final ExpressionMetadata metadata =
           CodeGenRunner.compileExpression(
```
AlanConfluent, line 88: Just curious, how much overhead is it to compile the java into bytecode, load the bytecode, and then run it.  It makes a lot of sense when you compile once and then run many times, but in this case, there's a 1:1 relationship.  In general, it seems like interpreting the expression would be lower overhead and possibly faster.  Do we do it this way because this is the main method that currently exists for "evaluating a sql expression"?

I'd be curious to run a benchmark doing just expression lookups (rather than literals).

AlanConfluent (approved):

ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/execution/PullQueryExecutor.java
agavra: yup, that was exactly the idea

ksqldb-engine/src/main/java/io/confluent/ksql/engine/generic/GenericExpressionResolver.java
agavra: Talked offline about this, we'll look into running specialized benchmarks to see how much the overhead is. Given that this is just "new" functionality, we won't block the PR on this and we'll run those benchmarks going forward.

PR #6632, opened 2020-11-17
chore: improve sr key schema check in insert values (MINOR)

agavra:
### Description 
Follow-up from https://github.com/confluentinc/ksql/pull/6601#discussion_r524878013

### Testing done 
Ran the unit tests

### Reviewer checklist

### DIFF
--- a/ksqldb-engine/src/main/java/io/confluent/ksql/engine/InsertValuesExecutor.java
+++ b/ksqldb-engine/src/main/java/io/confluent/ksql/engine/InsertValuesExecutor.java
@@ -31,6 +31,7 @@ import io.confluent.ksql.parser.tree.InsertValues;
 import io.confluent.ksql.rest.SessionProperties;
 import io.confluent.ksql.schema.ksql.PersistenceSchema;
 import io.confluent.ksql.schema.ksql.PhysicalSchema;
+import io.confluent.ksql.schema.registry.SchemaRegistryUtil;
 import io.confluent.ksql.serde.Format;
 import io.confluent.ksql.serde.FormatFactory;
 import io.confluent.ksql.serde.GenericKeySerDe;
@@ -46,7 +47,6 @@ import io.confluent.ksql.util.KsqlConstants;
 import io.confluent.ksql.util.KsqlException;
 import io.confluent.ksql.util.KsqlStatementException;
 import io.confluent.ksql.util.ReservedInternalTopics;
-import java.io.IOException;
 import java.time.Duration;
 import java.util.Map;
 import java.util.Objects;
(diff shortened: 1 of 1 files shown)


ksqldb-engine/src/main/java/io/confluent/ksql/engine/InsertValuesExecutor.java
```diff
@@ -303,21 +303,25 @@ private void ensureKeySchemasMatch(
...
-            + "\n\tksqlDB Generated: " + schema.canonicalString());
-      }
-    } catch (IOException | RestClientException e) {
+      latest = SchemaRegistryUtil.getLatestSchema(
+          serviceContext.getSchemaRegistryClient(),
+          dataSource.getKafkaTopicName(),
+          true);
+
+    } catch (final Exception e) {
       maybeThrowSchemaRegistryAuthError(format, dataSource.getKafkaTopicName(), true, "read", e);
```
vcrfxia, line 314: What's the purpose of separately handling potential auth errors here rather than unifying in SchemaRegistryUtil where missing schemas are handled?

ksqldb-engine/src/main/java/io/confluent/ksql/engine/InsertValuesExecutor.java
```diff
@@ -303,21 +303,25 @@ private void ensureKeySchemasMatch(
...
-            + "\n\tExisting Schema: " + latest.getSchema()
-            + "\n\tksqlDB Generated: " + schema.canonicalString());
-      }
-    } catch (IOException | RestClientException e) {
+      latest = SchemaRegistryUtil.getLatestSchema(
+          serviceContext.getSchemaRegistryClient(),
+          dataSource.getKafkaTopicName(),
+          true);
+
+    } catch (final Exception e) {
```
vcrfxia, line 313: Do we ever expect exceptions other than KsqlException here?

vcrfxia (approved):
LGTM with a couple minor questions. Thanks @agavra !

ksqldb-engine/src/main/java/io/confluent/ksql/engine/InsertValuesExecutor.java
agavra: No, we shouldn't have any other exception type here. I'll consolidate it

ksqldb-engine/src/main/java/io/confluent/ksql/engine/InsertValuesExecutor.java
agavra: It makes it a little easier to handle the error message. If it's an auth error, I want to expose that directly. If it's another error, then I want to throw our generic shim: "Could not determine that insert values..."

If I handled this in the SchemaRegistryUtil, then I would have to throw a different exception type or parse the exception message to decide whether to propagate it or to wrap it.

Issue #6249, opened 2020-09-17
KeyQueryMetadata not available for state store Aggregate-Aggregate-Materialize and key Struct{X=X}


rmoff:
ksqlDB v0.11

This error happens if you run a pull query against a table too quickly after creation. 

If you paste the following into the CLI in one go and hit enter, you see the error

```
➜ docker exec -it ksqldb ksql http://ksqldb:8088

OpenJDK 64-Bit Server VM warning: Option UseConcMarkSweepGC was deprecated in version 9.0 and will likely be removed in a future release.

                  ===========================================
                  =       _              _ ____  ____       =
                  =      | | _____  __ _| |  _ \| __ )      =
                  =      | |/ / __|/ _` | | | | |  _ \      =
                  =      |   <\__ \ (_| | | |_| | |_) |     =
                  =      |_|\_\___/\__, |_|____/|____/      =
                  =                   |_|                   =
                  =  Event Streaming Database purpose-built =
                  =        for stream processing apps       =
                  ===========================================

Copyright 2017-2020 Confluent Inc.

CLI v0.11.0, Server v0.11.0 located at http://ksqldb:8088

Having trouble? Type 'help' (case-insensitive) for a rundown of how things work!

ksql>
ksql> CREATE STREAM pageviews (msg VARCHAR)
>  WITH (KAFKA_TOPIC ='pageviews',
>        VALUE_FORMAT='JSON');
>
>SET 'auto.offset.reset' = 'earliest';
>
>CREATE TABLE MSG_COUNT AS
>    SELECT 'X' AS X,
>        COUNT(*) AS MSG_CT
>    FROM PAGEVIEWS
>    GROUP BY 'X'
>    EMIT CHANGES;
>
>SELECT * FROM MSG_COUNT WHERE X='X';
>

 Message
----------------
 Stream created
----------------
Successfully changed local property 'auto.offset.reset' to 'earliest'. Use the UNSET command to revert your change.

 Message
----------------------------------------
 Created query with ID CTAS_MSG_COUNT_0
----------------------------------------
KeyQueryMetadata not available for state store Aggregate-Aggregate-Materialize and key Struct{X=X}
```

If you then run the `SELECT` again after a few moments, it works

```
ksql> SELECT * FROM MSG_COUNT WHERE X='X';
+-----------------------------------------------------------------------------+-----------------------------------------------------------------------------+
|X                                                                            |MSG_CT                                                                       |
[...]

PR #6633, opened 2020-11-17
fix: KeyQueryMetadata not available on Pull query after stream is created

spena:
### Description 
Fixes https://github.com/confluentinc/ksql/issues/6249

A KeyQueryMetadata.NOT_AVAILABLE is usually caused when resources (hosts/streams) are not ready to access the state store. You can see this error when executing a pull query right after the stream is created. It takes some time for the streams to create the materialization in all the hosts.

To fix this issue, the clients must retry the pull query after some time in order to access the state store. This PR allows clients to retry requests that are marked as `ERROR_CODE_RESOURCE_NOT_READY` for a maximum number of retries and for some waiting time between retries.

By default, a request will wait for 500ms between retries and up to 10 retries. Users can configure this with the following config:
```
ksql> set 'ksql.retriable.requests.max.retries'='5';
ksql> set 'ksql.retriable.requests.sleep.ms'='1000';
```

### Testing done 

### Reviewer checklist

### DIFF
--- a/ksqldb-common/src/main/java/io/confluent/ksql/util/KsqlConfig.java
+++ b/ksqldb-common/src/main/java/io/confluent/ksql/util/KsqlConfig.java
@@ -366,6 +366,18 @@ public class KsqlConfig extends AbstractConfig {
   public static final String KSQL_VARIABLE_SUBSTITUTION_ENABLE_DOC
       = "Enable variable substitution on SQL statements.";
 
+  public static final String KSQL_RETRIABLE_REQUESTS_MAX_RETRIES
+      = "ksql.retriable.requests.max.retries";
+  public static final int KSQL_RETRIABLE_REQUESTS_MAX_RETRIES_DEFAULT = 10;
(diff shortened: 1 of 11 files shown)


ksqldb-common/src/main/java/io/confluent/ksql/util/KsqlConfig.java
```diff
@@ -366,6 +366,18 @@
   public static final String KSQL_VARIABLE_SUBSTITUTION_ENABLE_DOC
       = "Enable variable substitution on SQL statements.";
 
+  public static final String KSQL_RETRIABLE_REQUESTS_MAX_RETRIES
```
agavra, line 369: I haven't looked at the code yet, but I wanted to comment that I don't think we should set the default behavior to retry or sleep. For pull queries, I suspect a better default behavior is to fail fast than to hang - so that the user can decide whether they want to retry or not. What's important is to make sure that we have a good error message returned to the user.

spena:
Closing this now the issue is fixed here: https://github.com/confluentinc/ksql/pull/6709

Issue #6639, opened 2020-11-18
Remove Deprecation Suppression for uncaught exception handling


agavra:
There are a few places that are currently suppressing deprecation for `KafkaStreams#setUncaughtExceptionHandler`:

- `QueryMetadata` (constructor) 
- `QueryMetadata` (`setUncaughtHanlder`)
- `PersistentQueryMetadataTest`
- `StreamedQueryResourceTest`

We should remove these suppressions before shipping 0.15. We are currently suppressing them to unblock the build while Kafka Streams works on a clearer migration path or un-deprecates the old behavior.

PR #6640, opened 2020-11-18
chore: suppress deprecation UncaughtExceptionHandler

agavra:
### Description 

Fixes the build. Streams is not yet ready for migrating to the new API, but since we're building off master we're picking up this deprecation. AK might revert the `@Deprecated` for this method and we can revert this PR after that build is picked up in kafka-ccs.

https://github.com/confluentinc/ksql/issues/6639 tracks reverting it

### Testing done 

Built it

### Reviewer checklist


vcrfxia (approved):
LGTM -- thanks!

Issue #6646, opened 2020-11-19
ksqlDB does not propagate null-valued records when repartitioning


vcrfxia:
**Describe the bug**

ksqlDB currently drops null-valued records from repartition topics, which makes sense when repartitioning on an expression dependent on a non-key column but not when the expression depends only on key column(s). In the latter cases it makes more sense to compute the new key and propagate the null-value. 

Here's an example QTT that should pass:
```
    {
      "name": "key expression - with null value",
      "statements": [
        "CREATE STREAM TEST (K BIGINT KEY, ID bigint, NAME varchar, VALUE bigint) with (kafka_topic='test_topic', value_format = 'delimited');",
        "CREATE STREAM REPARTITIONED AS select K + 2, ID, NAME, VALUE from TEST partition by K + 2;"
      ],
      "inputs": [
        {"topic": "test_topic", "key": 0, "value": null},
        {"topic": "test_topic", "key": 0, "value": "0,zero,50"}
      ],
      "outputs": [
        {"topic": "REPARTITIONED", "key": 2, "value": null},
        {"topic": "REPARTITIONED", "key": 2, "value": "0,zero,50"}
      ]
    }
```
However, the QTT currently fails because the first output message has key `null` instead of `2`.

PR #6647, opened 2020-11-19
fix: propagate null-valued records in repartition

vcrfxia:
### Description 

Fixes https://github.com/confluentinc/ksql/issues/6646

The current mapper used in repartitions builds the new key by evaluating the generated expression on the value. If the value is null, the mapper returns an empty key rather than the original key. More intuitive behavior is to detect when the partition by expression depends only on key column(s) and evaluate the new key on the key columns in these cases, allowing null values to be propagated. This is needed to enable the propagation of tombstones in https://github.com/confluentinc/ksql/pull/6635

### Testing done 

Unit + QTT.

### Reviewer checklist

### DIFF
--- a/ksqldb-execution/src/main/java/io/confluent/ksql/execution/util/ColumnExtractor.java
+++ b/ksqldb-execution/src/main/java/io/confluent/ksql/execution/util/ColumnExtractor.java
@@ -13,7 +13,7 @@
  * specific language governing permissions and limitations under the License.
  */
 
-package io.confluent.ksql.analyzer;
+package io.confluent.ksql.execution.util;
 
 import io.confluent.ksql.execution.expression.tree.ColumnReferenceExp;
 import io.confluent.ksql.execution.expression.tree.Expression;
--- a/ksqldb-execution/src/main/java/io/confluent/ksql/execution/util/StructKeyUtil.java
+++ b/ksqldb-execution/src/main/java/io/confluent/ksql/execution/util/StructKeyUtil.java
@@ -58,7 +58,7 @@ public final class StructKeyUtil {
   }
 
   @SuppressWarnings("unchecked")
-  public static List<?> asList(final Object key) {
+  public static List<Object> asList(final Object key) {
     final Optional<Windowed<Object>> windowed = key instanceof Windowed
         ? Optional.of((Windowed<Object>) key)
         : Optional.empty();
--- a/ksqldb-streams/src/main/java/io/confluent/ksql/execution/streams/PartitionByParamsFactory.java
+++ b/ksqldb-streams/src/main/java/io/confluent/ksql/execution/streams/PartitionByParamsFactory.java
@@ -110,20 +110,6 @@ public final class PartitionByParamsFactory {
     return new PartitionByParams(resultSchema, mapper);
   }
 
-  private static class PartitionByExpressionEvaluator {
-
-    private final Function<GenericRow, Object> evaluator;
-    private final boolean acceptsKey;
-
-    PartitionByExpressionEvaluator(
-        final Function<GenericRow, Object> evaluator,
-        final boolean acceptsKey
-    ) {
-      this.evaluator = Objects.requireNonNull(evaluator, "evaluator");
-      this.acceptsKey = acceptsKey;
-    }
-  }
-
   public static LogicalSchema buildSchema(
       final LogicalSchema sourceSchema,
       final Expression partitionBy,
@@ -235,5 +221,19 @@ public final class PartitionByParamsFactory {
         row -> expressionMetadata.evaluate(row, null, logger, () -> errorMsg),
         isKeyExpression
     );
+  }
+
+  private static class PartitionByExpressionEvaluator {
+
+    private final Function<GenericRow, Object> evaluator;
+    private final boolean acceptsKey;
+
+    PartitionByExpressionEvaluator(
+        final Function<GenericRow, Object> evaluator,
+        final boolean acceptsKey
+    ) {
+      this.evaluator = Objects.requireNonNull(evaluator, "evaluator");
+      this.acceptsKey = acceptsKey;
+    }
   }
 }
--- a/ksqldb-streams/src/main/java/io/confluent/ksql/execution/streams/PartitionByParamsFactory.java
+++ b/ksqldb-streams/src/main/java/io/confluent/ksql/execution/streams/PartitionByParamsFactory.java
@@ -38,7 +38,7 @@ import java.util.Objects;
 import java.util.Optional;
 import java.util.Set;
 import java.util.function.BiFunction;
-import java.util.function.Function;
+import java.util.function.Supplier;
 import org.apache.kafka.connect.data.Struct;
 import org.apache.kafka.streams.KeyValue;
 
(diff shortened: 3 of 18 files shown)


ksqldb-execution/src/main/java/io/confluent/ksql/execution/util/StructKeyUtil.java
```diff
@@ -58,7 +58,7 @@ public static KeyBuilder keyBuilder(final ColumnName name, final SqlType type) {
   }
 
   @SuppressWarnings("unchecked")
-  public static List<?> asList(final Object key) {
+  public static List<Object> asList(final Object key) {
```
vcrfxia, line 61: Is there a reason this was previously `List<?>` rather than `List<Object>`? I don't fully understand the meaning of this change, but it lets us avoid an unchecked cast downstream.

ksqldb-execution/src/main/java/io/confluent/ksql/execution/util/ColumnExtractor.java
```diff
@@ -13,7 +13,7 @@
  * specific language governing permissions and limitations under the License.
  */
 
-package io.confluent.ksql.analyzer;
+package io.confluent.ksql.execution.util;
```
vcrfxia, line 16: I don't love the package this was moved to but couldn't think of anywhere better. Suggestions welcome!

ksqldb-streams/src/main/java/io/confluent/ksql/execution/streams/PartitionByParamsFactory.java
```diff
@@ -202,6 +217,23 @@ private static LogicalSchema buildSchema(
...
+    return new PartitionByExpressionEvaluator(
+        row -> expressionMetadata.evaluate(row, null, logger, () -> errorMsg),
+        isKeyExpression
+    );
+  }
+
+  private static class PartitionByExpressionEvaluator {
+
+    private final Function<GenericRow, Object> evaluator;
+    private final boolean acceptsKey;
```
agavra, line 229:
```suggestion
    private final boolean operatesOnKeyOnly;
```
A little hard for me to understand it inline without context on what this PR does:
```java
      final Object newKey = evaluator.evaluator.apply(
          evaluator.acceptsKey
          ? GenericRow.fromList(StructKeyUtil.asList(k))
          : v
      );
```

ksqldb-streams/src/main/java/io/confluent/ksql/execution/streams/PartitionByParamsFactory.java
```diff
@@ -202,6 +217,23 @@ private static LogicalSchema buildSchema(
     final String errorMsg = "Error computing new key from expression "
         + expressionMetadata.getExpression();
 
-    return row -> expressionMetadata.evaluate(row, null, logger, () -> errorMsg);
+    return new PartitionByExpressionEvaluator(
+        row -> expressionMetadata.evaluate(row, null, logger, () -> errorMsg),
```
agavra, line 221: I suspect this isn't the only place where this bug happens (a null value doesn't get properly passed into an `ExpressionEvaluator` 🤔). I had a PR at one point going to try to have `ExpressionMetadata#evaluate` take in both a key and a value instead of just a `GenericRow` which would have prevented this issue. Unfortunately, I couldn't get joins to work because the joiner doesn't actually have access to the key 😭 

tldr, I think it would be good if we encapsulate as much of this "put the keys and only the keys into the generic row if the row was null" into the `PartitionByExpressionEvaluator` and have that implement `BiFunction<Struct, GenericRow, Object>` so at least the part that creates the key is all in the same place that calls the evaluator:
```java
      final Object newKey = evaluator.evaluator.apply(
          evaluator.acceptsKey
          ? GenericRow.fromList(StructKeyUtil.asList(k))
          : v
      );
```

I would also just pass in `expressionMetadata` into the `PartitionByExpressionEvaluator` instead of passing it in as a lambda so that it's centralized and we don't have to track down where it's being called and what's getting passed into it

agavra (approved):
Superb sleuthing 🕵️ some minor comments inline but I think we can sneak this by without worrying about backwards compatibility too much (famous last words)

ksqldb-streams/src/main/java/io/confluent/ksql/execution/streams/PartitionByParamsFactory.java
```diff
@@ -87,12 +90,19 @@ public static PartitionByParams build(
       // is already present in the current value
       mapper = (k, v) -> new KeyValue<>(null, v);
     } else {
-      final Function<GenericRow, Object> evaluator = buildExpressionEvaluator(
+      final Set<? extends ColumnReferenceExp> partitionByCols =
+          ColumnExtractor.extractColumns(partitionBy);
+      final boolean isKeyExpression = partitionByCols.stream()
+          .map(ColumnReferenceExp::getColumnName)
+          .allMatch(sourceSchema::isKeyColumn);
```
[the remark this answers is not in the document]
vcrfxia: Not sure I understand. The validation that all key columns are selected happens elsewhere. This piece of code just determines whether only the old key is needed to compute the new key (partition by expression) or if the value is needed as well. If the former, we can propagate null-valued records, else we can't.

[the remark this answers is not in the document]
vcrfxia: Renamed `isKeyColumn` -> `partitionByInvolvesKeyColsOnly` to clarify.

ksqldb-streams/src/main/java/io/confluent/ksql/execution/streams/PartitionByParamsFactory.java
vcrfxia: Ah, thanks for the suggestion -- much cleaner!

What's the benefit of implementing `BiFunction<Object, GenericRow, Object>` rather than simply have a method? I've currently got
```
  Object evaluate(final Object key, final GenericRow value) {
      final GenericRow row = evaluateOnKeyOnly
          ? GenericRow.fromList(StructKeyUtil.asList(key))
          : value;
      return expressionMetadata.evaluate(row, null, logger, errorMsg);
    }
```

ksqldb-streams/src/main/java/io/confluent/ksql/execution/streams/PartitionByParamsFactory.java
vcrfxia: I think this is addressed by your suggestion below? I went with `evaluateOnKeyOnly` though I don't feel strongly.

ksqldb-streams/src/main/java/io/confluent/ksql/execution/streams/PartitionByParamsFactory.java
agavra: there is none, I think that's great (especially if nowhere else expects the BiFunction)

ksqldb-streams/src/main/java/io/confluent/ksql/execution/streams/PartitionByParamsFactory.java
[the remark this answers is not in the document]
agavra: ah! that makes more sense. I had misread what this code was doing.

PR #6649, opened 2020-11-20
docs: klip-43: timestamp data type support

jzaralim:
### Description 
KLIP for adding timestamp data type support to KSQL.

### Reviewer checklist

### DIFF
--- a/design-proposals/klip-43-timestamp-data-type-support.md
+++ b/design-proposals/klip-43-timestamp-data-type-support.md
@@ -1,7 +1,7 @@
 # KLIP-43: TIMESTAMP Data Type Support
 
 **Author**: @jzaralim | 
-**Release Target**: 0.15 | 
+**Release Target**: 0.15, 0.16 | 
 **Status**: In Discussion | 
 **Discussion**: https://github.com/confluentinc/ksql/pull/6649
 
@@ -25,10 +25,11 @@ lot of math. Having time data in a dedicated data type allows for a lot of new U
 
 * Add TIMESTAMP type to KSQL
 * Support TIMESTAMP arithmetic and comparisons
+* Allow window units (HOUR, DAY etc) to be used in timestamp functions
 * Support TIMESTAMP usage in STRUCT, MAP and ARRAY
 * Serialization and de-serialization of TIMESTAMPs to Avro, JSON, Protobuf and Delimited formats
 * Update existing built-in functions to use the TIMESTAMP data type
-* Casting between TIMESTAMP and BIGINT
+* Casting TIMESTAMP to and from BIGINT and STRING
 
 ## What is not in scope
 * Changing the ROWTIME data type. We will eventually want this to happen, but that is a separate
@@ -38,7 +39,6 @@ for negative timestamps, but until that is implemented, KSQL can only support po
 * DATE and TIME types - because TIMESTAMPs represent a point in time, DATE and TIME types
 would be useful if a user wants to represent a less specific time. These would be added after
 TIMESTAMP is implemented though.
-* Allow window units (HOUR, DAY etc) to be used in timestamp functions.
 
 ## Public APIS
 
@@ -50,7 +50,7 @@ CREATE STREAM stream_name (time TIMESTAMP, COL2 STRING) AS ...
 CREATE TABLE table_name (col1 STRUCT<field TIMESTAMP>) AS ...
 ```
 
-TIMESTAMPS will be displayed in console as strings in ISO-8601 format:
+TIMESTAMPS will be displayed in console as strings in ODBC canonical format with millisecond precision:
 
 ```roomsql
 > SELECT time FROM stream_name EMIT CHANGES;
@@ -58,21 +58,21 @@ TIMESTAMPS will be displayed in console as strings in ISO-8601 format:
 +------------------------+
 |time                    |
 +------------------------+
-|1994-11-05T13:15:30Z    |
+|1994-11-05 13:15:30:112 |
 ```
 
-TIMESTAMPS can be represented by milliseconds from Unix epoch or by using conversion functions:
+TIMESTAMPS can be represented by milliseconds from Unix epoch or by date strings:
 
 ```roomsql
 INSERT INTO stream_name VALUES (1605927509166);
-INSERT INTO stream_name VALUES (STRINGTOTIMESTAMP("2020-11-20", "YYYY-MM-DD"));
+INSERT INTO stream_name VALUES ("1994-11-05 13:15:30");
 ```
 
 ## Design
 
 ### Serialization/Deserialization
 
-TIMESTAMPs will be handled by the `java.time.Instant` class within KSQL. The corresponding Kafka Connect type is
+TIMESTAMPs will be handled by the `java.time.Instant` class in UTC within KSQL. The corresponding Kafka Connect type is
 [org.apache.kafka.connect.data.Timestamp](https://kafka.apache.org/0100/javadoc/org/apache/kafka/connect/data/Timestamp.html).
 They are represented as long types in Schema Registry, but also come with a tag indicating that it
 is a timestamp, so they should be distinguishable from long types when handling serialized values in KSQL.
@@ -106,7 +106,10 @@ The following UDFs should be updated to use the TIMESTAMP type instead of BIGINT
 Because BIGINTs will be implicitly cast into TIMESTAMPs and vice versa, queries using these functions
 with BIGINT will still work when TIMESTAMP is introduced.
 
-There will also be a new function, `NOW()` that returns the current timestamp.
+The following functions should also be added:
+
+* `NOW()` - returns the time after the issuing query is done executing
+* `CONVERT_TZ(timestamp, from_tz ,to_tz)` - converts a timestamp from one timezone to another
 
 There are a few existing UDFs that deal with dates. These should be left as is until a DATE type is implemented:
 
@@ -123,6 +126,10 @@ from Unix epoch.
 If a user attempts to cast a negative number or a timestamp before Unix epoch, the CAST will throw
 an error.
 
+Casting from TIMESTAMP to STRING will return the timestamp in ODBC canonical form with millisecond
+precision (yyyy-mm-dd HH:mm:ss:fff), and  casting from STRING to TIMESTAMP will attempt to parse the
+string into a TIMESTAMP.
+
 ### Arithmetic operations and comparisons
 
 At the very least, we would want KSQL to be able to compare, add and subtract TIMESTAMPS.
@@ -134,12 +141,13 @@ defines behaviors for arithmetic operators.
 Because functions are more well-defined, we should opt for built-in arithmetic functions
 instead of using operators. The following functions are necessary:
 
-* `TIMESTAMPADD(time_stamp, big_int)` - adds `big_int` milliseconds to `time_stamp` and returns the result as a
+* `TIMESTAMP_ADD(time_stamp, duration)` - adds a duration to `time_stamp` and returns the result as a
 TIMESTAMP
-* `TIMESTAMPDIFF(time_stamp1, time_stamp2)` - returns a BIGINT representing the number of milliseconds
-between `timestamp1` and `timestamp2`
-* `TIMESTAMPSUB(time_stamp1, big_int)` - subtracts `big_int` milliseconds from `time_stamp` and returns the
+* `TIMESTAMP_SUB(time_stamp1, duration)` - subtracts a duration from `time_stamp` and returns the
 result as a TIMESTAMP
+
+Durations will be expressed in the form, `<integer_value> <unit>`. This is discussed further in the
+next section.
 
 As for comparisons, the following expressions should be supported:
 ```
@@ -151,13 +159,12 @@ time_stamp1 BETWEEN time_stamp2 AND time_stamp3
 
 Comparisons between TIMESTAMPs and other data types should not be allowed.
 
-### Window units (not in scope)
+### Durations / Time units
 
-It might make sense to give window units a time representation so that they could be used as
-time intervals in TIMESTAMP arithmetic. For example,
+Giving window units a time representation enables timestamp arithmetic. For example,
 
 ```roomsql
-SELECT ADDTIME(TIME, 1 DAY) FROM FOO;
+SELECT TIMESTAMP_ADD(TIME, 1 DAY) FROM FOO;
 ```
 
 Because windowing is handled separately from expression evaluation, doing this will not have any
@@ -172,8 +179,12 @@ The following will need to be tested:
 
 ## LOEs and Delivery Milestones
 
-This feature can be broken down into two milestones: implementing (including testing and documentation)
-the TIMESTAMP type (1-2 weeks) and adding all of the supporting UDFs (~1 week).
+This feature can be broken down into three milestones:
+
+1. Implementing (including testing and documentation)
+the TIMESTAMP type (1-2 weeks)
+2. Adding and updating supporting UDFs (~1 week)
+3. Implementing duration and arithmetic (~1-2 weeks) - this will be in 0.16
 
 ## Documentation Updates
 
--- a/design-proposals/klip-43-timestamp-data-type-support.md
+++ b/design-proposals/klip-43-timestamp-data-type-support.md
@@ -52,7 +52,7 @@ CREATE STREAM stream_name (time TIMESTAMP, COL2 STRING) AS ...
 CREATE TABLE table_name (col1 STRUCT<field TIMESTAMP>) AS ...
 ```
 
-TIMESTAMPS will be displayed in console as strings in ODBC canonical format with millisecond precision:
+TIMESTAMPS will be displayed in console as strings in ISO-8601 form with millisecond precision:
 
 ```roomsql
 > SELECT time FROM stream_name EMIT CHANGES;
@@ -111,7 +111,7 @@ TIMESTAMP type and the new functions.
 The following functions will be added:
 * `PARSE_TIMESTAMP(format, timestamp_string)` - converts a string into a TIMESTAMP 
 * `FORMAT_TIMESTAMP(format, timestamp)` - returns a string in the specified format
-* `NOW()` - returns the time after the logical plan is built. All calls of now() within the same query return the same value.
+* `NOW()` - returns the timestamp after the logical plan is built. All calls of now() within the same query return the same value.
 * `CONVERT_TZ(timestamp, from_tz ,to_tz)` - converts a timestamp from one timezone to another
 
 There are a few existing UDFs that deal with dates. These should be left as is until a DATE type is implemented:
@@ -122,8 +122,8 @@ There are a few existing UDFs that deal with dates. These should be left as is u
 
 ### Casting
 
-Casting from TIMESTAMP to STRING will return the timestamp in ODBC canonical form with millisecond
-precision (yyyy-mm-dd HH:mm:ss:fff), and  casting from STRING to TIMESTAMP will attempt to parse the
+Casting from TIMESTAMP to STRING will return the timestamp in ISO-8601 form with millisecond
+precision (yyyy-mm-ddTHH:mm:ss:fff), and  casting from STRING to TIMESTAMP will attempt to parse the
 string into a TIMESTAMP.
 
 ```roomsql
--- a/design-proposals/klip-43-timestamp-data-type-support.md
+++ b/design-proposals/klip-43-timestamp-data-type-support.md
@@ -60,13 +60,13 @@ TIMESTAMPS will be displayed in console as strings in ODBC canonical format with
 +------------------------+
 |time                    |
 +------------------------+
-|1994-11-05 13:15:30:112 |
+|1994-11-05T13:15:30:112 |
 ```
 
 TIMESTAMPS can be represented by date strings:
 
 ```roomsql
-INSERT INTO stream_name VALUES ("1994-11-05 13:15:30");
+INSERT INTO stream_name VALUES ("1994-11-05T13:15:30");
 ```
 
 ## Design
@@ -94,7 +94,7 @@ Protobuf 3 has a Timestamp type. The Protouf deserializer KSQL uses supports thi
 
 #### JSON/Delimited
 
-Timestamps will get stored in JSON and CSV files as long values. The KSQL JSON and delimited deserializers
+Timestamps will get stored in JSON and CSV files in milliseconds as long values. The KSQL JSON and delimited deserializers
 will be updated to parse timestamps.
 
 ### UDFs
@@ -111,7 +111,7 @@ TIMESTAMP type and the new functions.
 The following functions will be added:
 * `PARSE_TIMESTAMP(format, timestamp_string)` - converts a string into a TIMESTAMP 
 * `FORMAT_TIMESTAMP(format, timestamp)` - returns a string in the specified format
-* `NOW()` - returns the time after the issuing query is done executing
+* `NOW()` - returns the time after the logical plan is built. All calls of now() within the same query return the same value.
 * `CONVERT_TZ(timestamp, from_tz ,to_tz)` - converts a timestamp from one timezone to another
 
 There are a few existing UDFs that deal with dates. These should be left as is until a DATE type is implemented:
@@ -125,6 +125,13 @@ There are a few existing UDFs that deal with dates. These should be left as is u
 Casting from TIMESTAMP to STRING will return the timestamp in ODBC canonical form with millisecond
 precision (yyyy-mm-dd HH:mm:ss:fff), and  casting from STRING to TIMESTAMP will attempt to parse the
 string into a TIMESTAMP.
+
+```roomsql
+ > SELECT CAST("1994-11-05T13:15:30" AS TIMESTAMP) FROM stream_name EMIT CHANGES;
+"1994-11-05T13:15:30"
+ > SELECT CAST(time AS STRING) FROM stream_name EMIT CHANGES;
+"1994-11-05T13:15:30"
+```
 
 ### Arithmetic operations and comparisons
 
@@ -157,7 +164,14 @@ Comparisons between TIMESTAMPs and other data types should not be allowed.
 
 ### Durations / Time units
 
-Giving window units a time representation enables timestamp arithmetic. For example,
+KSQL currently recognizes the following time units for defining windows:
+* DAY, DAYS
+* HOUR, HOURS
+* MINUTE, MINUTES
+* SECOND, SECONDS
+* MILLISECOND, MILLISECONDS
+
+These could be repurposed to enable timestamp arithmetic. For example,
 
 ```roomsql
 SELECT TIMESTAMP_ADD(TIME, 1 DAY) FROM FOO;
--- a/design-proposals/klip-43-timestamp-data-type-support.md
+++ b/design-proposals/klip-43-timestamp-data-type-support.md
@@ -5,39 +5,67 @@
 **Status**: In Discussion | 
 **Discussion**: https://github.com/confluentinc/ksql/pull/6649
 
-**tl;dr:** _KSQL currently converts time data into long values. This KLIP introduces a TIMESTAMP
-data type to handle time data instead._
+**tl;dr:** _Add support for TIMESTAMP column types in ksqlDB. This will allow users to easily migrate
+time data from other databases without having to convert column types, as well as simplify time data
+manipulation_
 
 ## Motivation and background
 
-With Connect integration, KSQL processes data from various sources and sinks. Most other databases
-have a TIMESTAMP or DATETIME type, which KSQL currently converts into long values. This sometimes
-causes problems when sinking into a database. Supporting TIMESTAMP types would make moving time data
-between KSQL and other data sources/sinks smoother and less error-prone.
+With Connect integration, KSQL processes data from various databases. Most other databases
+have a TIMESTAMP or DATETIME type, which KSQL currently converts to BIGINT. This makes sinking to
+databases that don't implicitly cast long values to timestamps much more complicated - the user would
+have to set up their connector to convert these columns back to TIMESTAMP. Supporting TIMESTAMP types
+would make moving time data between KSQL and other data sources/sinks smoother and less error-prone.
+
+Adding a TIMESTAMP type also simplifies time data manipulation. For example, currently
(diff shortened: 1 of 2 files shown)


design-proposals/klip-43-timestamp-data-type-support.md
```diff
@@ -0,0 +1,159 @@
...
+
+**tl;dr:** _KSQL currently converts time data into long values. This KLIP introduces a TIMESTAMP
+data type to handle time data instead._
+
+## Motivation and background
+
+With Connect integration, KSQL processes data from various sources and sinks. Most other databases
+have a TIMESTAMP or DATETIME type, which KSQL currently converts into long values. This sometimes
+causes problems when sinking into a database. Supporting TIMESTAMP types would make moving time data
+between KSQL and other data sources/sinks smoother and less error-prone.
```
spena, line 16: Let's extend a little bit what the current problems exist and the benefits users would get when TIMESTAMP column types exist. Seems useful to differentiate between TIMESTAMP column types vs data types. As you mentioned at some point, delimited and JSON will use long values as the data format type, and AVRO will use its own data format type (with ending being a long value too). 

A few comments from above:

`With Connect integration, KSQL processes data from various sources and sinks.`. This seems not in the right context. `sources and sinks` are also a concept for CREATE_AS queries, not related to Connect. What's the real problem when users use Connect and are migrating tables with timestamps? I don't know about them, so it would be good to know what's wrong. Errors during the Connect migration? Users require UDF to migrate them? What's wrong?

`which KSQL currently converts into long values.` I'm not sure about this one. As I mentioned in the TLDR section, long values are used in the final format written to the topic anyway, right? Is the intention here that ksqlDB always use long column types to reference to timestamps?

`This sometimes causes problems when sinking into a database.` What's the problem when sinking? If I store a timestamp as long value, then use UDF to read from that sink, then it should be read fine, right? Or is there an actual problem trying to store the timestamp as long value?

`Supporting TIMESTAMP types would make moving time data between KSQL and other data sources/sinks smoother  and less error-prone.` This is good. Are there any other reasons? Look at the problems I mentioned in the TLDR section.

design-proposals/klip-43-timestamp-data-type-support.md
```diff
@@ -0,0 +1,159 @@
...
+
+## What is in scope
+### Required
+
+* Add TIMESTAMP type to KSQL
+* Serialization and de-serialization of TIMESTAMPs to Avro, JSON and Delimited formats
```
spena, line 22: What about Protobuf? We support that serde too.

design-proposals/klip-43-timestamp-data-type-support.md
```diff
@@ -0,0 +1,159 @@
...
+* Built-in functions to support TIMESTAMP usage
```
spena, line 23: Will this include new built-in functions or modify the existing ones to support TIMESTAMP?

design-proposals/klip-43-timestamp-data-type-support.md
```diff
@@ -0,0 +1,159 @@
...
+
+### Future enhancements
+
+* Give window units (HOUR, DAY etc) a TIMESTAMP value
+* Add DATE and TIME types
+* UDFs such as DAY and HOUR
```
spena, line 29: Aren't this out of scope for this KLIP? Or are there plans to work on them? Like DATE and TIME, those seem different data types than TIMESTAMP, so they're not in this scope, right?

design-proposals/klip-43-timestamp-data-type-support.md
```diff
@@ -0,0 +1,159 @@
...
+
+## Motivation and background
+
+With Connect integration, KSQL processes data from various sources and sinks. Most other databases
+have a TIMESTAMP or DATETIME type, which KSQL currently converts into long values. This sometimes
+causes problems when sinking into a database. Supporting TIMESTAMP types would make moving time data
+between KSQL and other data sources/sinks smoother and less error-prone.
+
+## What is in scope
+### Required
```
spena, line 19: What about:
  - Should we support arithmetic functions too? casting? any other basic feature should be listed here.
  - Should you support STRUCT, ARRAY, MAP as well?

design-proposals/klip-43-timestamp-data-type-support.md
```diff
@@ -0,0 +1,159 @@
...
+* Add DATE and TIME types
+* UDFs such as DAY and HOUR
+
+## What is not in scope
+Changing the ROWTIME data type. We will eventually want this to happen, but that is a separate discussion.
+
+## Public APIS
+
+The TIMESTAMP data type will store a time without timezone information. It will be useable in any
+place STRING or INTEGER could be used:
```
spena, line 37: `It will be useable in any place STRING or INTEGER could be used:` Will you able to concatenate or do arithmetic operations with timestamps?

design-proposals/klip-43-timestamp-data-type-support.md
```diff
@@ -0,0 +1,159 @@
...
+```roomsql
+CREATE STREAM stream_name (col1 TIMESTAMP, COL2 STRING) AS ...
+CREATE TABLE table_name (col1 STRUCT<field TIMESTAMP>) AS ...
+```
+
+## Design
```
spena, line 43: I know this is trivial, but could you add the syntax how this would work? It's part of the design.

Also, what is a timestamp value? Is it a long value? So far, I read that ksqldB treats timestamps as long values, and that we don't what that (even if we store long values), so is there a different value to refer to timestamps?

Some SQL use this:
```
create or replace table time (ltz timestamp);
insert into time values ('2016-05-01 00:00:00.000');
```

We should clarify how a timestamp value will be defined.

design-proposals/klip-43-timestamp-data-type-support.md
```diff
@@ -0,0 +1,159 @@
...
+
+### Serialization/Deserialization
```
spena, line 45: What about Protobuf serde?

design-proposals/klip-43-timestamp-data-type-support.md
```diff
@@ -0,0 +1,159 @@
...
+The Avro deserializer KSQL uses supports this.
+
+#### JSON/Delimited
+
+Timestamps will get stored in JSON and CSV files as long values. The KSQL JSON and delimited deserializers
+will be updated to parse timestamps.
+
+### UDFs
+
+The following UDFs should be updated to use the TIMESTAMP type instead of BIGINT:
```
spena, line 69: Will BIGINT be discontinued? If so, that might break compatibility, right? Or will there an implicity casting of BIGINT -> TIMESTAMP?

For instance, I wonder what would happen if these scenarios:
- TIMESTAMPTOSTRING(timestamp_col, 'yyyy-MM-dd HH:mm:ss zzz','America/Los_Angeles')
- TIMESTAMPTOSTRING(bigint_col, 'yyyy-MM-dd HH:mm:ss zzz','America/Los_Angeles')
- TIMESTAMPTOSTRING(cast(int_col as bigint), 'yyyy-MM-dd HH:mm:ss zzz','America/Los_Angeles')
^ Will this fail because we're casting to BIGINT?

design-proposals/klip-43-timestamp-data-type-support.md
```diff
@@ -0,0 +1,159 @@
...
+
+* TIMESTAMPTOSTRING
+* STRINGTOTIMESTAMP
```
spena, line 72: Can't this be an implicit cast? Like:
```
insert into (ts) values ('2020-11-20 14:34:00');
```

That above doesn't require a UDF to convert to timestamps. 

Or is the function used for string columns instead of literals? For this, other SQL systems use `TO_TIMESTAMP`, which converts a String to a Timestamp.

design-proposals/klip-43-timestamp-data-type-support.md
```diff
@@ -0,0 +1,159 @@
...
+* UNIX_TIMESTAMP
```
spena, line 73: What does this do? How is it used?

`FROM_UNIXTIME` is a good UDF to support too. Though, this converts the unix in seconds to a String format (no a timestamp column).

design-proposals/klip-43-timestamp-data-type-support.md
```diff
@@ -0,0 +1,159 @@
...
+
+The following UDFs should also be added to support conversions between TIMESTAMP and BIGINT representations of time:
+
+* TIMESTAMPTOBIGINT
```
spena, line 77: This can be done with `unix_timestamp`, can't it?

design-proposals/klip-43-timestamp-data-type-support.md
```diff
@@ -0,0 +1,159 @@
...
+#### JSON/Delimited
+
+Timestamps will get stored in JSON and CSV files as long values. The KSQL JSON and delimited deserializers
+will be updated to parse timestamps.
+
+### UDFs
+
+The following UDFs should be updated to use the TIMESTAMP type instead of BIGINT:
+
+* TIMESTAMPTOSTRING
```
spena, line 71: Isn't this done as a cast? Like `cast(ts as string)` ?

design-proposals/klip-43-timestamp-data-type-support.md
```diff
@@ -0,0 +1,159 @@
...
+Since the above UDFs already handle conversions, casting between TIMESTAMP and STRING or BIGINT is
+unnecessary. If DATE and TIME types are implemented, then the conversions between them and TIMESTAMP
+will be as follows:
+
+| From      | To        | Result                                                |
+|---------- |:----------|:------------------------------------------------------|
+| TIMESTAMP | DATE      | DATE type representing date portion of the TIMESTAMP. |
+| TIMESTAMP | TIME      | TIME type representing time portion of the TIMESTAMP. |
+| DATE      | TIMESTAMP | TIMESTAMP with time portion set to 00:00:00.          |
+| TIME      | TIMESTAMP | TIMESTAMP with date portion set to Unix epoch.        |
```
spena, line 97: If DATE and TIME are out of scope, should we remove these casting for the doc?

design-proposals/klip-43-timestamp-data-type-support.md
```diff
@@ -0,0 +1,159 @@
...
+
+* UNIX_DATE
+* DATETOSTRING
+* STRINGTODATE
+
+### Casting/Converting
+
+Since the above UDFs already handle conversions, casting between TIMESTAMP and STRING or BIGINT is
+unnecessary. If DATE and TIME types are implemented, then the conversions between them and TIMESTAMP
+will be as follows:
```
spena, line 90: Casting to/from timestamp is a valid casting in other DBs. Any reason why we're not supporting them?

design-proposals/klip-43-timestamp-data-type-support.md
```diff
@@ -0,0 +1,159 @@
...
+```roomsql
+SELECT ADDTIME(TIME, 1 DAY) FROM FOO;
+```
+
+Because windowing is handled separately from expression evaluation, doing this will not have any
+impact on windows.
+
+## Test plan
+
+There should be an integration test with Kafka Connect and Schema Registry.
```
spena, line 136: What about QTT tests with different serdes? and the supported UDFs?

Do we need a performance test plan for TIMESTAMP serdes? We have one for some serde formats, but I don't know if we need one for Timestamps (your call).

design-proposals/klip-43-timestamp-data-type-support.md
```diff
@@ -0,0 +1,159 @@
...
+
+## LOEs and Delivery Milestones
+
+This feature can be broken down into two milestones: implementing the TIMESTAMP type (1-2 weeks) and
+adding all of the supporting UDFs (~1 week).
```
spena, line 141: What about the milestones for docs? We need to add this as a new syntax.

design-proposals/klip-43-timestamp-data-type-support.md
```diff
@@ -0,0 +1,159 @@
...
+
+* Description of the TIMESTAMP data type
+* TIMESTAMP usage in WHERE/GROUP/PARTITION clauses
+* New and updated UDFs
+* We might want to add this into one of the quick-starts
+
+## Compatibility Implications
+
+If a user creates a stream or table with a TIMESTAMP typed column, then that column will not be
+(de)serializable in earlier versions. Besides that, there should be no other compatibility issues.
```
spena, line 155: Does it mean a value written as TIMESTAMP in 0.15 won't be deserialized in 0.14? Why not? If timestamps will be stored as long values in Delimited, Json, and Avro too; why will they fail?

jzaralim: The situation I had in mind was if there was a command like `CREATE STREAM stream_name (col TIMESTAMP) AS ...`. KSQL would not recognize `TIMESTAMP` as a type. That was worded very inaccurately, I'll rewrite that part.

Timestamps in Avro/JSON etc will be stored and deserialized as long values in older versions.

design-proposals/klip-43-timestamp-data-type-support.md
jzaralim: The reason for this instead of a cast is to support different string formats. Perhaps having a date format property in KSQL could make supporting TIMESTAMP/STRING casting possible?

design-proposals/klip-43-timestamp-data-type-support.md
jzaralim: UNIX_TIMESTAMP returns the current millis in BIGINT. KSQL doesn't use this for conversions.

design-proposals/klip-43-timestamp-data-type-support.md
jzaralim: Casting to/from BIGINT makes sense for KSQL. I have some reservations about casting to strings, as explained above.

design-proposals/klip-43-timestamp-data-type-support.md
jzaralim: BIGINT/TIMESTAMP implicit casting makes sense. In that case, all three of those commands would work

design-proposals/klip-43-timestamp-data-type-support.md
jzaralim: I don't think having the TIMESTAMP type in a schema would have any effect on performance, and `java.time.Instant` does not seem to have any performance issues. It might be good to check at some point, but I don't think we will need to benchmark it.

design-proposals/klip-43-timestamp-data-type-support.md
```diff
@@ -0,0 +1,194 @@
...
+
+The following UDFs should be updated to use the TIMESTAMP type instead of BIGINT:
+
+* TIMESTAMPTOSTRING
+* STRINGTOTIMESTAMP
+
+Because BIGINTs will be implicitly cast into TIMESTAMPs and vice versa, queries using these functions
+with BIGINT will still work when TIMESTAMP is introduced.
+
+There will also be a new function, `NOW()` that returns the current timestamp.
```
PeterLindner, line 109: Time is tricky in Kafka, may be beneficial to clarify which 'now' this would be. Is it the servers time or stream-time? If it is stream-time, will each partition have a different time? Will it be evaluated when the query is issued or when the function is called (important for persistent queries)?
All of those variants are useful in different scenarios and unfortunately all of them could be called `NOW`

big-andy-coates:
IMHO NOW should work as it would in a transitional db. It should be wall
clock time and should be only evaluated once when the statement is executed
and the results balked in.

It may be possible to add another STREAMTIME function that tracks stream
time per partition, but this should not be what MIW does.

design-proposals/klip-43-timestamp-data-type-support.md
```diff
@@ -0,0 +1,194 @@
...
+```roomsql
+INSERT INTO stream_name VALUES (1605927509166);
+INSERT INTO stream_name VALUES (STRINGTOTIMESTAMP("2020-11-20", "YYYY-MM-DD"));
+```
+
+## Design
+
+### Serialization/Deserialization
+
+TIMESTAMPs will be handled by the `java.time.Instant` class within KSQL. The corresponding Kafka Connect type is
```
agavra, line 75: I think it's important to have a discussion on time zone. Will all timestamps be treated by default in UTC? how could a user convert timestamps between zones?

design-proposals/klip-43-timestamp-data-type-support.md
agavra: we could copy what mySQL does here: https://dev.mysql.com/doc/refman/8.0/en/datetime.html but it would require implementing a local time zone to do the conversion as well

agavra (commented):
Some minor comments, but LGTM

design-proposals/klip-43-timestamp-data-type-support.md
```diff
@@ -0,0 +1,194 @@
...
++------------------------+
+|time                    |
++------------------------+
+|1994-11-05T13:15:30Z    |
+```
+
+TIMESTAMPS can be represented by milliseconds from Unix epoch or by using conversion functions:
+
+```roomsql
+INSERT INTO stream_name VALUES (1605927509166);
```
spena, line 67: There's a concern I have about the implicit cast here. Some systems (including Mysql, Hive, and Impala) use timestamp as seconds, but Java and other internal formats (i.e. Avro) use timestamp as milliseconds. Why are we choosing milliseconds instead of the common seconds approach? Could users get confused about the value here? Should we prevent this implicit cast instead?
```
mysql> select unix_timestamp('2020-11-30 13:25:00');
+---------------------------------------+
| unix_timestamp('2020-11-30 13:25:00') |
+---------------------------------------+
|                            1606764300 |
+---------------------------------------+
1 row in set (0.00 sec)
```

design-proposals/klip-43-timestamp-data-type-support.md
```diff
@@ -0,0 +1,194 @@
...
+At the very least, we would want KSQL to be able to compare, add and subtract TIMESTAMPS.
+
+[MySQL](https://dev.mysql.com/doc/refman/8.0/en/date-and-time-functions.html) and [MariaDB](https://mariadb.com/kb/en/date-time-functions/)
+have very complete lists of built-in arithmetic functions, while [PostgreSQL](https://www.postgresql.org/docs/9.0/functions-datetime.html)
+defines behaviors for arithmetic operators.
+
+Because functions are more well-defined, we should opt for built-in arithmetic functions
+instead of using operators. The following functions are necessary:
+
+* `TIMESTAMPADD(time_stamp, big_int)` - adds `big_int` milliseconds to `time_stamp` and returns the result as a
```
spena, line 137: Mysql and Snowflake uses the `TIMESTAMPADD(unit, interval, datetime_expr)` format. I don't know about other DBs yet, but should we align with those systems? If so, that means we cannot handle this `timestampadd` and `timestampsub` in this klip because time units are out of scope, right?

jzaralim: Yeah those syntaxes make a lot more sense. I do think that supporting timestamp arithmetic is valuable, so I might include this and time units in scope as another milestone for a later release.

design-proposals/klip-43-timestamp-data-type-support.md
jzaralim: I agree with Andy's suggestion on this. I would also like to note that there is an existing function in KSQL called `UNIX_TIMESTAMP` that returns the time in ms when the function is called, so the distinction should be clearly defined in the docs.

design-proposals/klip-43-timestamp-data-type-support.md
jzaralim: I was thinking that the timestamps here would reflect the timestamps in Kafka, which assumes UTC. MySQL and Snowflake have the `convert_tz` function, so we could do something similar.

design-proposals/klip-43-timestamp-data-type-support.md
jzaralim: The main reason is because Kafka represents timestamps as milliseconds, and a lot of what we have now is based off that (ROWTIME representation as BIGINT, timestamps being converted to long types) so implicitly casting to milliseconds would help with compatibility. However, it seems that most other DBs use a seconds representation, so a new user could get confused by this. We should explain this clearly in the docs if we decide to implicitly cast.

design-proposals/klip-43-timestamp-data-type-support.md
jzaralim: Yeah supporting string/timestamp casting should work. I think we should keep these functions though, in case a user wants to convert strings in a format that isn't recognized by standard formatters. .

design-proposals/klip-43-timestamp-data-type-support.md
```diff
@@ -0,0 +1,206 @@
...
+```
+The Avro deserializer KSQL uses supports this.
+
+#### Protobuf
+Protobuf 3 has a Timestamp type. The Protouf deserializer KSQL uses supports this.
+
+#### JSON/Delimited
+
+Timestamps will get stored in JSON and CSV files as long values. The KSQL JSON and delimited deserializers
+will be updated to parse timestamps.
```
spena, line 97: Will this be stored as milliseconds or seconds? Same concern I had with implicit casting.

Btw, there's another thing to discuss here. What if users have a different timestamp format? Say a String format `2020-12-07 10:02:05`? There are no rules of how JSON stores timestamps, so users may have a different format. The may have CSV topics (comma delimited), which might have this string format as well. But others may long values.

Is there a way we can define the timestamp format for the text (delimited/json) serdes? Should we define a syntax for this?

design-proposals/klip-43-timestamp-data-type-support.md
```diff
@@ -0,0 +1,206 @@
...
+
+### UDFs
+
+The following UDFs should be updated to use the TIMESTAMP type instead of BIGINT:
+
+* TIMESTAMPTOSTRING
+* STRINGTOTIMESTAMP
```
spena, line 104: This is gonna cause some confusion between millis and seconds. We really need to discuss this with the rest of the team, how should we handle timestamps, as millis or seconds? If seconds, then we don't want to break compatibility with TIMESTAMPTOSTRING and STRINGTOTIMESTAMP, which currently support milliseconds.

If we do milliseconds, then we're good with the above UDF.
If we do seconds, we perhaps want to add a `to_timestamp(string, format)` function that returns the UNIX epoch in seconds. `to_timestamp` is used in other DB too in the same format. Postgres adds `to_timestamp(double)` which converts seconds to timestamp.

spena (commented):
Hey @jzaralim, I left a few more comments regarding the use of milliseconds vs seconds. I'm still worried about the implicit casting and other UDF implicit casting because of how other DBs use timestamps. Now that ksql is moving as a DB, I think we should design timestamps in a same way as they do (in seconds), but our history in Kafka and Ksql has been in millis.

@derekjn Could you take a look at this KLIP and let us know your opinion, especially on the use of millis vs seconds?

design-proposals/klip-43-timestamp-data-type-support.md
spena: Here's an example of a JSON format I saw in some logs
```
"modified_at": "2020-12-09T01:11:45.194487+00:00"
```

If users want to analyze JSON logs with the above timestamp, then the current proposal for Long values won't work. I've seen that a common thing in logs.

jzaralim: This is supported by implicit casting. If the user is writing these into a timestamp valued column, then they'll be automatically cast into a timestamp.

design-proposals/klip-43-timestamp-data-type-support.md
```diff
@@ -0,0 +1,206 @@
...
+
+TIMESTAMPS will be displayed in console as strings in ODBC canonical format with millisecond precision:
+
+```roomsql
+> SELECT time FROM stream_name EMIT CHANGES;
+
++------------------------+
+|time                    |
++------------------------+
+|1994-11-05 13:15:30:112 |
```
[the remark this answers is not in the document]
derekjn: > Now that ksql is moving as a DB, I think we should design timestamps in a same way as they do (in seconds)

I think we should decouple the presentation of timestamps from how they're actually stored internally. We can store timestamps as milliseconds internally and then present them with whatever default precision we feel is best. For example, PostgreSQL stores timestamps internally as 8-byte integers in nanoseconds, but will present their default epoch representation as fractional seconds.

design-proposals/klip-43-timestamp-data-type-support.md
agavra: We should follow the lead of connect serializers for how they store JSON timestamp data by default

derekjn:
@agavra @spena @jzaralim summarizing our discussion around the remaining open items:

* **Time zones** - `TIMESTAMP` type will not include support for time zones. All values will be stored as UTC timestamps. Adding support for time zones will be future work.
* **Implicit casting** - Disallow implicit casting of integers as timestamps. Users must convert integral timestamps to `TIMESTAMPs` using the appropriate builtin functions.
* **Arithmetic** - Timestamp arithmetic will be accomplished via builtin functions such as `TIMESTAMP_ADD`.
* **Storage precision** - Store timestamps using millisecond-level precision. Our storage format should be compatible with Connect's serialization format for timestamps.
* **Presentation** - `TIMESTAMP` values should be presented as standard timestamp strings (e.g. ISO 8601).

design-proposals/klip-43-timestamp-data-type-support.md
```diff
@@ -0,0 +1,204 @@
...
+There are a few existing UDFs that deal with dates. These should be left as is until a DATE type is implemented:
+
+* UNIX_DATE
+* DATETOSTRING
+* STRINGTODATE
+
+### Casting
+
+Casting from TIMESTAMP to STRING will return the timestamp in ODBC canonical form with millisecond
+precision (yyyy-mm-dd HH:mm:ss:fff), and  casting from STRING to TIMESTAMP will attempt to parse the
```
spena, line 126:
```suggestion
precision (yyyy-mm-ddTHH:mm:ss:fff), and  casting from STRING to TIMESTAMP will attempt to parse the
```

Based ISO 8601

design-proposals/klip-43-timestamp-data-type-support.md
```diff
@@ -0,0 +1,204 @@
...
+* `NOW()` - returns the time after the issuing query is done executing
+* `CONVERT_TZ(timestamp, from_tz ,to_tz)` - converts a timestamp from one timezone to another
+
+There are a few existing UDFs that deal with dates. These should be left as is until a DATE type is implemented:
+
+* UNIX_DATE
+* DATETOSTRING
+* STRINGTODATE
+
+### Casting
```
spena, line 123: Could you add the example for casting? Just to avoid  confusion between implicit and explicit casting. I assume this is explicit casting, right?

design-proposals/klip-43-timestamp-data-type-support.md
```diff
@@ -0,0 +1,204 @@
...
+```
+time_stamp1 < time_stamp2
+time_stamp1 > time_stamp2
+time_stamp1 = time_stamp2
+time_stamp1 BETWEEN time_stamp2 AND time_stamp3
+```
+
+Comparisons between TIMESTAMPs and other data types should not be allowed.
+
+### Durations / Time units
```
spena, line 158: What units will we support? Only DAY?

Could you add a list of units to support?

design-proposals/klip-43-timestamp-data-type-support.md
```diff
@@ -0,0 +1,204 @@
...
+* STRINGTOTIMESTAMP
+
+These functions will still be available so that existing queries using these functions won't break,
+but the documentation will state that they are deprecated and will direct users towards using the
+TIMESTAMP type and the new functions.
+
+The following functions will be added:
+* `PARSE_TIMESTAMP(format, timestamp_string)` - converts a string into a TIMESTAMP 
+* `FORMAT_TIMESTAMP(format, timestamp)` - returns a string in the specified format
+* `NOW()` - returns the time after the issuing query is done executing
```
spena, line 114:
```suggestion
* `NOW()` - returns the timestamp after the issuing query is done executing. All calls of now() within the same query return the same value.
```

Btw, I'm not sure about getting the time after the query is done executing. What if a user wants to use `NOW()` in the where clause to get the current day? 
I.e.
```
select * from stream where completed_date = format_timestamp(now(), 'YYYY-mm-dd');
```

Perhaps the now() needs to be obtained after the query is evaluated or planned? During the logical or physical plan? What if we use a push query (execution does not end until manually terminate it)?

jzaralim: Updated to clarify this. Obtaining the `NOW` value seems to make the most sense in the logical plan.

design-proposals/klip-43-timestamp-data-type-support.md
```diff
@@ -52,21 +52,21 @@ CREATE STREAM stream_name (time TIMESTAMP, COL2 STRING) AS ...
...
+TIMESTAMPS will be displayed in console as strings in ISO-8601 form with millisecond precision:
 
 ```roomsql
 > SELECT time FROM stream_name EMIT CHANGES;
 
 +------------------------+
 |time                    |
 +------------------------+
-|1994-11-05 13:15:30:112 |
+|1994-11-05T13:15:30:112 |
```
spena, line 63:
```suggestion
|1994-11-05T13:15:30.112 |
```

There is a fractional part separated by a dot mentioned in the standard: `"yyyy-mm-ddThh:mm:ss[.mmm]`.

design-proposals/klip-43-timestamp-data-type-support.md
```diff
@@ -122,10 +122,17 @@ There are a few existing UDFs that deal with dates. These should be left as is u
 
 ### Casting
 
-Casting from TIMESTAMP to STRING will return the timestamp in ODBC canonical form with millisecond
-precision (yyyy-mm-dd HH:mm:ss:fff), and  casting from STRING to TIMESTAMP will attempt to parse the
+Casting from TIMESTAMP to STRING will return the timestamp in ISO-8601 form with millisecond
+precision (yyyy-mm-ddTHH:mm:ss:fff), and  casting from STRING to TIMESTAMP will attempt to parse the
```
spena, line 126:
```suggestion
precision (yyyy-mm-ddTHH:mm:ss.fff), and  casting from STRING to TIMESTAMP will attempt to parse the
```

spena (approved):
Looks good. I just made a couple of changes to use a fractional part in the timestamp format.

Thanks @jzaralim for the KLIP. It looks pretty good!.

Issue #3624, opened 2019-10-18
UDF code does not check return type is compatible with defined schema of return value


big-andy-coates:
**Describe the bug**

If the UDF provides the schema of its return type using the `schema` field of the `Udf` annotation, then there does not seem to be any checking that the methods return value is compatible/matches.

**To Reproduce**

Add this UDF:

```java
@UdfDescription(name = "badudf", description = "UDF that returns a type that does not match the advertised schema")
public class BadUdf {

  @Udf(schema = "ARRAY<VARCHAR>")
  public List<Integer> apply(final String value) {
    return ImmutableList.of(1);
  }
}
```

**Expected behavior**

For the UDF to be rejected when the server starts up.

**Actual behaviour**

UDF is loaded by KSQL and can be used in a statement. The result type will be `ARRAY<VARCHAR>`.  Only at runtime, when processing each row, will it fail due to the difference in types.  How it fails depends on how it is used.  Either in the serializer or in any UDF that the result is passed to.

**Additional context**

The UDF loading code should check that the java return type of the method is compatible with the provided schema.

big-andy-coates:
Checking the `retVal.getClass()` matches what is expected would be cheap and could catch potential errors, especially when other users are developing / using UDFs.

Of course, for structured and decimal types, it's also possible the data within the returned value is incorrect, e.g. a decimal with the wrong scale, or an array with the wrong element type.  Checking such, especially for a heavily nested data type, can be very expensive.

I would therefore suggest we do not, by default, check anything (as the code currently does).  Instead, we add a `--debug` flag (or similar) that allows ksqlDB to be started up with extra checks in place.  These checks should:

 - Perform _deep_ checks on the return values from functions.
 - Perform _deep_ checks on the return values from deserializers.

We can run QTT with these checks enabled to ensure our own serde and functions are 'doing the right thing'. Users can also use them if they are writing their own serdes and functions.

Outside of this, we should rely on functions and serdes being correctly tested and avoid incurring the repetitive overhead of of type checking everything.

PR #6653, opened 2020-11-20
refactor: remove unused validation code

big-andy-coates:
### Description 

`KsqlStruct` is not yet used. However, the code calls `validateValue` when setting the value of a field. This check is not free and is on the critical path.  ksqlDB is strongly typed, so the only way the type would be wrong would be if a function or deserializer where returning incorrect data, i.e. a bug.  We should not pay the production runtime cost of checking _every_ field we set during _every_ record we process.

Rather than validate here, we should add a `-debug` option to ksqlDB that validates the returned data from deserializers and functions against the expected schema, as suggested here: https://github.com/confluentinc/ksql/issues/3624#issuecomment-729532558

### Testing done 

not production code.

### Reviewer checklist


agavra (approved):
`shouldNotThrowOnValueWrongTypeAsTypeCheckingEverySetIsTooExpensive` 😆

PR #6655, opened 2020-11-21
fix: fix error categorization on NPE from streams

rodesai:
Fixes error categorization when streams exits due to an internal NPE
- fixes the regex categorizer to correctly handle NPEs. NPEs have null
  descriptions, so this patch changes the categorizer to ignore the
  description if its null.
- fixes the uncaught handler to categorize errors as UNKNOWN if the
  categorizer throws


agavra (approved):
🙀 nice catch. Thanks for the quick fix

PR #6659, opened 2020-11-23
chore: switch to schema-less KsqlStruct

big-andy-coates:
### Description 

Note: built on top of https://github.com/confluentinc/ksql/pull/6653, review that first.

No other sql instance types, (e.g. BigDecimal, String, Double, Map, List, etc), require the schema during instantiation. Only the Connect `Struct` type and the ksqlDB specific `KsqlStruct` type, which is to be `Struct`'s replacement, require the schema during Construction.  This can make working with structs more tricky than other types.

This change removes the requirement to provide the schema when instantiating `KsqlStruct`. The aim is to make this type easier to use.

The `Field` type has also been moved into the `SqlStruct` class. The code base has many `Field` types. `Field` is specifically a struct field. Making this an inner class of `SqlStruct` helps avoid confusion as to which field type we are talking about.

### Testing done 

Usual.

### Reviewer checklist


ksqldb-udf/src/main/java/io/confluent/ksql/types/KsqlStruct.java
```diff
@@ -35,36 +39,57 @@
...
   public static Builder builder(final SqlStruct schema) {
     return new Builder(schema);
   }
 
-  private KsqlStruct(
-      final SqlStruct schema,
-      final List<Optional<?>> values
-  ) {
-    this.schema = Objects.requireNonNull(schema, "schema");
+  private KsqlStruct(final List<Optional<?>> values) {
```
agavra, line 49: I feel like if we're touting efficiency in this class, we should strive not to use optionals to avoid unnecessary wrapping/unwrapping - but that can be change for another time :)

agavra (approved):
LGTM! We'll see how it plays when we actually replace the connect data types 🤞 it should be faster and easier to use

ksqldb-udf/src/main/java/io/confluent/ksql/types/KsqlStruct.java
big-andy-coates: I'd be surprised if it made a significant difference.  We'd need to profile to work that out.

PR #6663, opened 2020-11-23
Pull query thread leak

big-andy-coates:
### Description 

This fixes a leaked executor service with 100 threads that is being created once-per-request for pull queries.

1. The first commit fixes the leaked executor.
2. The second enhances RQTT to (hopefully) catch such errors in the future (this is the second such bug in quick succession).

This bug was stopping RQTT running locally for me, because the JVM was running out of resources.

**IMPORTANT**:  It looks to my untrained eye that the current pull query code is creating an executor with 100 threads **_on each request_**. (HARouting is created per-request).  Even with the code now closing down the executor, surely this isn't what we want?

### Testing done 

**Minimal**

### Reviewer checklist

### DIFF
--- a/ksqldb-engine/src/main/java/io/confluent/ksql/engine/EngineExecutor.java
+++ b/ksqldb-engine/src/main/java/io/confluent/ksql/engine/EngineExecutor.java
@@ -164,10 +164,12 @@ final class EngineExecutor {
           ksqlConfig,
           analysis,
           statement);
-      final HARouting routing = new HARouting(
+
+      try (HARouting routing = new HARouting(
           ksqlConfig, physicalPlan, routingFilterFactory, routingOptions, statement, serviceContext,
-          physicalPlan.getOutputSchema(), physicalPlan.getQueryId(), pullQueryMetrics);
-      return routing.handlePullQuery();
+          physicalPlan.getOutputSchema(), physicalPlan.getQueryId(), pullQueryMetrics)) {
+        return routing.handlePullQuery();
+      }
     } catch (final Exception e) {
       pullQueryMetrics.ifPresent(metrics -> metrics.recordErrorRate(1));
       throw new KsqlStatementException(
--- a/ksqldb-functional-tests/src/test/java/io/confluent/ksql/test/rest/RestQueryTranslationTest.java
+++ b/ksqldb-functional-tests/src/test/java/io/confluent/ksql/test/rest/RestQueryTranslationTest.java
@@ -16,6 +16,7 @@
 package io.confluent.ksql.test.rest;
 
 
+import static io.confluent.ksql.test.util.ThreadTestUtil.filterBuilder;
 import static java.util.Objects.requireNonNull;
 
 import com.fasterxml.jackson.annotation.JsonIgnoreProperties;
@@ -27,6 +28,8 @@ import io.confluent.ksql.rest.server.TestKsqlRestApp;
 import io.confluent.ksql.test.loader.JsonTestLoader;
 import io.confluent.ksql.test.loader.TestFile;
 import io.confluent.ksql.test.model.TestFileContext;
+import io.confluent.ksql.test.util.ThreadTestUtil;
+import io.confluent.ksql.test.util.ThreadTestUtil.ThreadSnapshot;
 import io.confluent.ksql.test.utils.TestUtils;
 import io.confluent.ksql.util.KsqlConfig;
 import java.nio.file.Path;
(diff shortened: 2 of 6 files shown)


ksqldb-engine/src/main/java/io/confluent/ksql/engine/EngineExecutor.java
```diff
@@ -164,10 +164,12 @@ PullQueryResult executePullQuery(
           ksqlConfig,
           analysis,
           statement);
-      final HARouting routing = new HARouting(
+
+      try (HARouting routing = new HARouting(
```
AlanConfluent, line 168: This must have accidentally been introduced in the refactor.  I'm surprised it didn't show up in benchmarking if the threads were being leaked.  I think that technically, it would have been eventually garbage collected once the executor was GCed, but it is certainly preferable to shut it down immediately.

ksqldb-functional-tests/src/test/java/io/confluent/ksql/test/rest/RestQueryTranslationTest.java
```diff
@@ -121,6 +127,17 @@ public void tearDown() {
     REST_APP.closePersistentQueries();
     REST_APP.dropSourcesExcept();
     TEST_HARNESS.getKafkaCluster().deleteAllTopics(TestKsqlRestApp.getCommandTopicName());
+
+    if (STARTING_THREADS.get() == null) {
```
AlanConfluent, line 131: Very useful.  How did you discover the leak without this kind of test in place to begin with?

AlanConfluent (approved):

AlanConfluent:
> IMPORTANT: It looks to my untrained eye that the current pull query code is creating an executor with 100 threads on each request. (HARouting is created per-request). Even with the code now closing down the executor, surely this isn't what we want?
I agree that what we should really do is factor out the thread pool to be a singleton per application.  I think that's how it originally was.

ksqldb-engine/src/main/java/io/confluent/ksql/engine/EngineExecutor.java
agavra: Are we sure the change was properly picked up when benchmarked? It looks like it used to be created once per `PullQueryExecutor` (which I think was once per request). It could be that the benchmarks didn't catch it because the `newFixedThreadPool` is good about not allocating 100 threads unless it needs them (so we were really only leaking one thread each time). I still suspect that this would have very quickly caused OOM (cannot allocate native thread) after we issue a few thousand pull queries.

ksqldb-functional-tests/src/test/java/io/confluent/ksql/test/rest/RestQueryTranslationTest.java
big-andy-coates: I knew there was an issue because RQTT wouldn't run locally in the IDE for me.  Then investigated  with VisualVM :D

ksqldb-engine/src/main/java/io/confluent/ksql/engine/EngineExecutor.java
big-andy-coates: It is GC'd, but leaving threads lying around is not good practice as they are an OS resource and hold a chunk of memory too.  A bug such as this can result in requests failing due to insufficient resources, especially in something like pull queries that can see high call volumes.

AlanConfluent: PullQueryExecutor was created once per application before the refactor.  @vpapavas did benchmark this with the new physical plan code, but it must not have shown up only doing high qps for ~30 minutes at a time.  Well, good to catch before hitting the soak cluster.

AlanConfluent: If it's not too big of a change for this PR, it would be great to make the threadpool one per application.  Otherwise we can do it after this PR.  We should make sure we do that before 0.15 regardless.

apurvam:
>  It looks to my untrained eye that the current pull query code is creating an executor with 100 threads on each request. (HARouting is created per-request). Even with the code now closing down the executor, surely this isn't what we want?

If this is true, it would be good to understand if this is something we could have caught in the review. Also, did it show up in any of our benchmarks? It seems like it should have.

big-andy-coates:
I think the thread pools are pretty clever these days and don't create threads till needed, which may explain why its not showing up in benchmarking.  Of course, if the benchmarking is also making calls that will need the HA routing to other nodes, then I'd of expected it to show up.

The old code used a single shared executor. I don't think we want to move to a per-request one, as it won't be as efficient.

ksqldb-engine/src/main/java/io/confluent/ksql/engine/EngineExecutor.java
big-andy-coates: https://github.com/confluentinc/ksql/issues/6666

PR #6664, opened 2020-11-24
test: fix JaasAuthProviderTest validation

vcrfxia:
### Description 

Functionality was lost from the JaasAuthProviderTest over the course of a couple refactors. Specifically, the test was no longer verifying that the authentication roles config was properly used in validating user authorization. This PR fixes the tests to check the behavior again, and also performs some minor cleanup in the file.

Credits to @big-andy-coates for discovering the test gap!

### Testing done 

Tests pass now, and they also fail if conditions are changed to be erroneous.

### Reviewer checklist


big-andy-coates (approved):
LGTM

PR #6667, opened 2020-11-24
swap `Struct` key to `GenericKey`

big-andy-coates:
### Description 

Part of the work to remove Connect `Struct` from the code base.  This commit changes the type used to represent the key from Connect's `Struct` to a new `GenericKey` class, to match our existing `GenericRow` class for the value.

Lot of files changed, but not much functionality.

`GenericKey`:
* is a simple wrapper around a `List` of values, indexed by the position of the key column in the key schema.
* is immutable, (keys should be immutable, but `Struct` is mutable)
* does not perform type validation. (`Struct` does, which is expensive and unnecessary).

### Reviewing notes.

Three main commits to make it easier to review:
1. (88 files) First contains any files that simply change `Struct` with `GenericKey`. **Not worth reviewing IMHO**.
2. (45 files) Second contains non-production files, mainly test, where there are slightly more changes, e.g. changing tests to create and expect `GenericKey` rather than `Struct`. **Not worth reviewing IMHO**.
3. (27 files) Third contains the meat of the change, including associated test code changes.  Review this!

Main changes of interest:
1. `GenericKey`.
1. `GenericKeySerDe` - changed to work with `GenericKey` rather than `Struct`.
1. `GenericSerializer` & `GenericDeserializer` -> code pulled out from `GenericRowSerde` as its now common across key & value serde
1. Changes in the `ksqldb-streams` module, which is where the key is manipulated.

### Testing done 

usual

### Reviewer checklist


big-andy-coates:
Merging without review as the PR touches many files and will be a PITA to keep up to date with merges.  Plus there's not actually much functionality change. Mainly a find-and-replace.

PR #6669, opened 2020-11-24
docs: development in single Slack channel

MichaelDrogalis:
We've just been using the community channel for all discussion, so let's recommend that one alone.


JimGalasyn (approved):
Cool, I'll cherry-pick to the live branches.

vcrfxia (approved):

PR #6680, opened 2020-11-25
fix: throw error message on create source with no value columns

### LINKED ISSUE #6352: It's possible to create a source with no value columns... but its not usable.
**Describe the bug**

It's possible to create a stream or table with no value columns using a `CREATE TABLE` or `CREATE STREAM` statement.  No error is returned.

However, if you try and use the stream or table in a downstream `CREATE TABLE AS` or `CREATE STREAM AS` statement you get a error message.

i.e. the table/stream created with no columns isn't useful.

**To Reproduce**

```sql
CREATE STREAM WHAT_NO_VAL_COLS (ID INT KEY)
  WITH (kafka_topic='blah', format='json', partitions=1);

CREATE STREAM FOO AS SELECT * FROM FROM WHAT_NO_VAL_COLS;
```

**Expected behavior**

Either:

1. The first statement fails with an error about no value columns, or 
1. [we fix ksql so that the second statement works](https://github.com/confluentinc/ksql/issues/5564).

**Actual behaviour**

The first statement succeeds and the second statement fails with an error about no value columns in the projection, which ain't great UX!

agavra:
### Description 

Fixes #6352 by throwing an error message if anyone creates a source with no value columns.

### Testing done 

Unit tests and QTT tests

### Reviewer checklist

### DIFF
--- a/ksqldb-engine/src/test/java/io/confluent/ksql/ddl/commands/CreateSourceFactoryTest.java
+++ b/ksqldb-engine/src/test/java/io/confluent/ksql/ddl/commands/CreateSourceFactoryTest.java
@@ -928,7 +928,10 @@ public class CreateSourceFactoryTest {
     // Given:
     final CreateStream statement = new CreateStream(
         SOME_NAME,
-        TableElements.of(tableElement(KEY, "k", new Type(SqlTypes.STRING))),
+        TableElements.of(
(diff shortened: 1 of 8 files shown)


vcrfxia (approved):
LGTM!

PR #6684, opened 2020-11-26
feat: Move code from physical to logical plan for Projection

vpapavas:
### Description 

No changes made to the persistent query code path. Ideally, I wanted to re-use the existing code without special casing for pull queries but that was not possible since the schemas are handled so differently. Below I describe with examples their differences.

Assume pull query `SELECT * from agg WHERE STR=foo;` where table is created as 
```
CREATE TABLE agg AS SELECT STR, LATEST_BY_OFFSET(LONG) AS LONG FROM stream GROUP BY STR;"
```

#### Logical schema handling in pull query project code. 
There are 3 schemas we handling, the input schema, the intermediate schema used for codegen and the output schema. 

- The input is the materialization schema. For example, `mat.schema = STR STRING KEY, LONG BIGINT` 
- CodeGen is used only if the SELECT is not `*`. For CodeGen, an intermediate schema is created as follows: Check if projection contains system or key columns. If not, the intermediate schema is the input schema. If there are any of these columns, the input schema is extended with `inputSchema	
.withPseudoAndKeyColsInValue(windowed);` that basically copies the key and system columns (rowtime, windwostart and windowend) into the value of the schema.
- The output schema (the schema of the row produced by the project operator), takes as input the materialization schema and: if the projection is  `SELECT *`, add windowstart and windowend to key columns and keep value columns the same as input. If projection is not `SELECT *`, then process each select and if it is a key or windowstart and windowend add them to the key part else add them to the value part. In our example, output schema is `STR STRING KEY, LONG BIGINT`.

#### Logical schema handling in logical plan builder. 
Each node of the logical plan has a logical schema associated with it and has access to its child logical schema. For pull queries the leaf is always a single `DataSourceNode`

- The output schema of the `DataSourceNode` (for the same example query as above) is `STR STRING KEY, LONG BIGINT, ROWTIME BIGINT, STR STRING`. This is the input to the Project operator, compare this to the `mat.schema`
- The output schema of the `FinalProjectNode` is `STR STRING KEY, STR STRING, LONG BIGINT` (it has the extra key column in the value)

#### Different queries that highlight how the schema handling of pull queries differs from persistent queries

1. `SELECT long from agg WHERE STR=foo`
-----> FinalProjectNode input schema: `STR STRING KEY, LONG BIGINT, ROWTIME BIGINT, STR STRING`
-----> mat.schema = `STR STRING KEY, LONG BIGINT`
-----> intermediate schema = `STR STRING KEY, LONG BIGINT, ROWTIME BIGINT, STR STRING`
-----> FinalProjectNode output schema=`STR STRING KEY, LONG BIGINT`
-----> pull query output schema= `LONG BIGINT`

2. `SELECT long, rowtime from agg WHERE STR=foo`
-----> FinalProjectNode input schema: `STR STRING KEY, LONG BIGINT, ROWTIME BIGINT, STR STRING`
-----> mat.schema = `STR STRING KEY, LONG BIGINT`
-----> intermediate schema= `STR STRING KEY, LONG BIGINT, ROWTIME BIGINT, STR STRING`
-----> Project output schema=`STR STRING KEY, LONG BIGINT, ROWTIME BIGINT`
-----> pull query output schema= `LONG BIGINT, ROWTIME BIGINT`

3. `SELECT long, windowstart, windowend from agg WHERE STR=foo` assuming table is windowed
-----> FinalProjectNode input schema: `STR STRING KEY, ROWTIME BIGINT, WINDOWSTART BIGINT, WINDOWEND BIGINT`
-----> mat.schema = `STR STRING KEY, LONG BIGINT, WINDOWSTART BIGINT, WINDOWEND BIGINT`
-----> intermediate schema= `STR STRING KEY, LONG BIGINT, ROWTIME BIGINT, STR STRING,  WINDOWSTART BIGINT, WINDOWEND BIGINT`
-----> Project output schema=`STR STRING KEY`
-----> pull query output schema= `LONG BIGINT,  WINDOWSTART BIGINT KEY, WINDOWEND BIGINT KEY`

#### Questions:

1. What is the correspondence between DataSource and Materialization? Why do their schemas differ? If the Materialization is a table, then shouldn't the schema of the DataSource match that of the table? I.e not have the rowtime and key in the value?

2. Why do the output schemas of the pull query and `FinalProjectNode` differ? Basically, it seems that the output schemas of persistent queries are different than that of pull queries? Moreover, the output schema of `FinalProjectNode` seems to have extra columns that are not in the projection. Why?

### Testing done 
Added new unit tests for building the pull query schema

### Reviewer checklist

### DIFF
--- a/ksqldb-engine/src/main/java/io/confluent/ksql/physical/pull/PullPhysicalPlanBuilder.java
+++ b/ksqldb-engine/src/main/java/io/confluent/ksql/physical/pull/PullPhysicalPlanBuilder.java
@@ -39,10 +39,10 @@ import io.confluent.ksql.physical.pull.operators.WhereInfo;
 import io.confluent.ksql.planner.LogicalPlanNode;
 import io.confluent.ksql.planner.plan.DataSourceNode;
 import io.confluent.ksql.planner.plan.FilterNode;
-import io.confluent.ksql.planner.plan.FinalProjectNode;
 import io.confluent.ksql.planner.plan.KsqlBareOutputNode;
 import io.confluent.ksql.planner.plan.OutputNode;
 import io.confluent.ksql.planner.plan.PlanNode;
+import io.confluent.ksql.planner.plan.PullProjectNode;
 import io.confluent.ksql.query.QueryId;
 import io.confluent.ksql.statement.ConfiguredStatement;
 import io.confluent.ksql.util.KsqlConfig;
@@ -124,8 +124,8 @@ public class PullPhysicalPlanBuilder {
     while (true) {
 
       AbstractPhysicalOperator currentPhysicalOp = null;
-      if (currentLogicalNode instanceof FinalProjectNode) {
-        currentPhysicalOp = translateProjectNode((FinalProjectNode)currentLogicalNode);
+      if (currentLogicalNode instanceof PullProjectNode) {
+        currentPhysicalOp = translateProjectNode((PullProjectNode)currentLogicalNode);
       } else if (currentLogicalNode instanceof FilterNode) {
         currentPhysicalOp = translateFilterNode((FilterNode)currentLogicalNode);
       } else if (currentLogicalNode instanceof DataSourceNode) {
@@ -159,15 +159,14 @@ public class PullPhysicalPlanBuilder {
     }
     return new PullPhysicalPlan(
         rootPhysicalOp,
-        ((FinalProjectNode)(rootPhysicalOp).getLogicalNode())
-            .getPullQueryOutputSchema().get(),
+        (rootPhysicalOp).getLogicalNode().getSchema(),
         queryId,
         keys,
         mat,
         dataSourceOperator);
   }
 
-  private ProjectOperator translateProjectNode(final FinalProjectNode logicalNode) {
+  private ProjectOperator translateProjectNode(final PullProjectNode logicalNode) {
     final ProcessingLogger logger = processingLogContext
         .getLoggerFactory()
         .getLogger(
--- a/ksqldb-engine/src/main/java/io/confluent/ksql/physical/pull/PullPhysicalPlanBuilder.java
+++ b/ksqldb-engine/src/main/java/io/confluent/ksql/physical/pull/PullPhysicalPlanBuilder.java
@@ -16,7 +16,6 @@
 package io.confluent.ksql.physical.pull;
 
 import com.google.common.collect.ImmutableList;
-import com.google.common.collect.Iterables;
 import io.confluent.ksql.GenericKey;
 import io.confluent.ksql.analyzer.ImmutableAnalysis;
 import io.confluent.ksql.analyzer.PullQueryValidator;
@@ -45,17 +44,12 @@ import io.confluent.ksql.planner.plan.KsqlBareOutputNode;
 import io.confluent.ksql.planner.plan.OutputNode;
 import io.confluent.ksql.planner.plan.PlanNode;
 import io.confluent.ksql.query.QueryId;
-import io.confluent.ksql.schema.ksql.PhysicalSchema;
-import io.confluent.ksql.serde.connect.ConnectSchemas;
 import io.confluent.ksql.statement.ConfiguredStatement;
 import io.confluent.ksql.util.KsqlConfig;
 import io.confluent.ksql.util.KsqlException;
 import io.confluent.ksql.util.PersistentQueryMetadata;
 import java.util.List;
 import java.util.Objects;
-import org.apache.kafka.connect.data.ConnectSchema;
-import org.apache.kafka.connect.data.Field;
-import org.apache.kafka.connect.data.Struct;
 
 /**
  * Traverses the logical plan top-down and creates a physical plan for pull queries.
@@ -215,17 +209,6 @@ public class PullPhysicalPlanBuilder {
       return new KeyedWindowedTableLookupOperator(
           mat, logicalNode, whereInfo.getWindowBounds().get());
     }
-  }
-
-  private Struct asKeyStruct(final Object keyValue, final PhysicalSchema physicalSchema) {
-    final ConnectSchema keySchema = ConnectSchemas
-        .columnsToConnectSchema(physicalSchema.keySchema().columns());
-
-    final Field keyField = Iterables.getOnlyElement(keySchema.fields());
-
-    final Struct key = new Struct(keySchema);
-    key.put(keyField, keyValue);
-    return key;
   }
 
   private QueryId uniqueQueryId() {
--- a/ksqldb-engine/src/main/java/io/confluent/ksql/physical/pull/PullPhysicalPlanBuilder.java
+++ b/ksqldb-engine/src/main/java/io/confluent/ksql/physical/pull/PullPhysicalPlanBuilder.java
@@ -16,25 +16,20 @@
 package io.confluent.ksql.physical.pull;
 
 import com.google.common.collect.ImmutableList;
+import com.google.common.collect.Iterables;
 import io.confluent.ksql.GenericKey;
 import io.confluent.ksql.analyzer.ImmutableAnalysis;
 import io.confluent.ksql.analyzer.PullQueryValidator;
 import io.confluent.ksql.execution.context.QueryContext.Stacker;
 import io.confluent.ksql.execution.context.QueryLoggerUtil;
 import io.confluent.ksql.execution.context.QueryLoggerUtil.QueryType;
-import io.confluent.ksql.execution.plan.SelectExpression;
 import io.confluent.ksql.execution.streams.materialization.Materialization;
-import io.confluent.ksql.execution.util.ExpressionTypeManager;
 import io.confluent.ksql.logging.processing.ProcessingLogContext;
 import io.confluent.ksql.logging.processing.ProcessingLogger;
 import io.confluent.ksql.metastore.MetaStore;
 import io.confluent.ksql.metastore.model.DataSource;
-import io.confluent.ksql.model.WindowType;
 import io.confluent.ksql.name.SourceName;
-import io.confluent.ksql.parser.tree.AllColumns;
 import io.confluent.ksql.parser.tree.Query;
-import io.confluent.ksql.parser.tree.Select;
-import io.confluent.ksql.parser.tree.SingleColumn;
 import io.confluent.ksql.physical.pull.operators.AbstractPhysicalOperator;
 import io.confluent.ksql.physical.pull.operators.DataSourceOperator;
 import io.confluent.ksql.physical.pull.operators.KeyedTableLookupOperator;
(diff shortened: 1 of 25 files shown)


ksqldb-engine/src/main/java/io/confluent/ksql/planner/plan/FinalProjectNode.java
```diff
@@ -56,23 +67,54 @@
...
+              metaStore
+          ))
+          .collect(ImmutableList.toImmutableList()));
+    } else {
+      this.isPullQuerySelectStar = false;
+      this.addAdditionalColumnsToIntermediateSchema = false;
+      this.pullQueryOutputSchema = Optional.empty();
+      this.pullQueryIntermediateSchema = Optional.empty();
+      this.compiledSelectExpressions = Optional.empty();
+      throwOnEmptyValueOrUnknownColumns();
```
agavra, line 115: >  What is the correspondence between DataSource and Materialization? Why do their schemas differ? If the Materialization is a table, then shouldn't the schema of the DataSource match that of the table? I.e not have the rowtime and key in the value?
>
>    Why do the output schemas of the pull query and FinalProjectNode differ? Basically, it seems that the output schemas of persistent queries are different than that of pull queries? Moreover, the output schema of FinalProjectNode seems to have extra columns that are not in the projection. Why?

Off the top of my head (see bottom), I don't know the answers to these questions (and I'm not sure that without digging anyone on the team knows) but I think these are exactly the ones we need to answer (and document for the future! Ideally in the code...) before merging this pr. Then, we can see if we can change it to be a single code path for both persistent/pull queries.

Otherwise (if we can't merge the code paths), I think we need to have two different logical nodes since they (logically) represent different operations and construct different schemas (they can extend a common base so that it's clear which code they share and which is different). We shouldn't make the pull and persistent code more intertwined but with a nuanced difference as it will make it harder to develop them independently without context on the other.

FWIW, you can change the behavior of any logical node of the persistent queries so long as the QTT tests pass. Any old query will continue to work because the physical plans are persisted, not the logical ones, in the command topic.

---


> Moreover, the output schema of FinalProjectNode seems to have extra columns that are not in the projection.

My best guess on this one is that the rowkey-in-value/rowtime-in-value/windowstart/windowend columns are needed to handle the processing. At the moment, the way we've coded it up, most operations accept only the value and not the key (or other metadata information) - so we need to copy information from the key into the value. I've looked into changing this and I got pretty close, but the problem is that the Joiner code in streams doesn't yet have the signature that I need to fully get rid of it so I abandoned the change. I'll try to revisit it (and hopefully change that in streams if necessary) at some point.

ksqldb-engine/src/main/java/io/confluent/ksql/physical/pull/operators/ProjectOperator.java
```diff
@@ -163,17 +86,17 @@ public Object next() {
     if (row == null) {
       return null;
     }
-    if (isSelectStar) {
+    if (logicalNode.getIsSelectStar()) {
       return createRow(row);
     }
-    final GenericRow intermediate = preSelectTransform.apply(row);
+    final GenericRow intermediate = getPreSelectTransform().apply(row);
```
AlanConfluent, line 92: If you just apply the transform immediately, can you just call it directly rather than returning a function?

ksqldb-engine/src/main/java/io/confluent/ksql/physical/pull/operators/ProjectOperator.java
```diff
@@ -207,8 +135,34 @@ public AbstractPhysicalOperator getChild(final int index) {
     throw new UnsupportedOperationException();
   }
 
-  public LogicalSchema getOutputSchema() {
-    return outputSchema;
+  private Function<TableRow, GenericRow> getPreSelectTransform() {
```
AlanConfluent, line 138: Maybe this should be called getIntermediateRow to match the terminology of the logical node?

ksqldb-engine/src/main/java/io/confluent/ksql/physical/pull/PullPhysicalPlanBuilder.java
```diff
@@ -137,17 +124,18 @@ public PullPhysicalPlan buildPullPhysicalPlan(final LogicalPlanNode logicalPlanN
     while (true) {
 
       AbstractPhysicalOperator currentPhysicalOp = null;
-      if (currentLogicalNode instanceof ProjectNode) {
-        currentPhysicalOp = translateProjectNode((ProjectNode)currentLogicalNode);
+      if (currentLogicalNode instanceof FinalProjectNode) {
```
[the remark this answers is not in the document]
AlanConfluent: I think this is outdated

ksqldb-engine/src/main/java/io/confluent/ksql/planner/plan/PullProjectNode.java
```diff
@@ -0,0 +1,260 @@
...
+   * @return the intermediate schema
+   */
+  @VisibleForTesting
+  protected LogicalSchema buildIntermediateSchema() {
+    final LogicalSchema parentSchema = getSource().getSchema();
+
+    if (!addAdditionalColumnsToIntermediateSchema) {
+      return parentSchema;
+    } else {
+      // SelectValueMapper requires the rowTime & key fields in the value schema :(
```
AlanConfluent, line 141: But that's fine because this intermediate schema is never exposed, right?  The output schema puts these fields correctly as keys.

ksqldb-engine/src/main/java/io/confluent/ksql/planner/plan/PullProjectNode.java
```diff
@@ -0,0 +1,260 @@
...
+ * <p>There are 3 schemas the node is handling, the input schema, the intermediate schema used
+ * for codegen and the output schema.
+ * <ul>
+ * <li>The input is the schema of the child node
+ *
+ * <li>CodeGen is used only if the projection is not SELECT *. For CodeGen, an intermediate schema
+ * is created as follows:
+ * Check if projection contains system or key columns. If not, the intermediate schema
+ * is the input schema. If there are any of these columns, the input schema is extended by copying
+ * the key and system columns (rowtime, windwostart and windowend) into the value of the schema.
```
AlanConfluent, line 54: This is just to use the SelectValueMapper, right?

ksqldb-engine/src/main/java/io/confluent/ksql/planner/plan/PullProjectNode.java
```diff
@@ -0,0 +1,260 @@
...
+
+  private LogicalSchema selectOutputSchema(
+      final MetaStore metaStore,
+      final List<SelectExpression> selectExpressions,
+      final boolean isWindowed
+  ) {
+    final Builder schemaBuilder = LogicalSchema.builder();
+    final LogicalSchema parentSchema = getSource().getSchema();
+
+    // Copy meta & key columns into the value schema as SelectValueMapper expects it:
```
AlanConfluent, line 239: Does this mean ExpressionTypeManager expects it?  That's the main place where this schema is used is just to evaluate the type of the expression.

ksqldb-engine/src/main/java/io/confluent/ksql/planner/LogicalPlanner.java
```diff
@@ -170,6 +171,25 @@ public OutputNode buildPlan() {
     return buildOutputNode(currentNode);
   }
 
+  public OutputNode buildPullLogicalPlan() {
```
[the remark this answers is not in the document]
vpapavas: Some of the checks about about for instance the group by are performed in the Analyzer and specifically, by the class `PullQueryValidator`. The rest of the checks about the WHERE clause will be performed when building the filter node and will be in the next PR

ksqldb-engine/src/main/java/io/confluent/ksql/planner/plan/PullProjectNode.java
```diff
@@ -0,0 +1,260 @@
...
+    this.isSelectStar = isSelectStar();
+    this.addAdditionalColumnsToIntermediateSchema = shouldAddAdditionalColumnsInSchema();
+    this.outputSchema = buildOutputSchema(metaStore);
+    this.intermediateSchema = buildIntermediateSchema();
+    this.compiledSelectExpressions = selectExpressions
+        .stream()
+        .map(selectExpression -> CodeGenRunner.compileExpression(
+            selectExpression.getExpression(),
+            "Select",
+            intermediateSchema,
```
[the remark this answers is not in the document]
vpapavas: Yeah, for the next PR that does the WHERE clause I would have to share this schema. I will build it once in the WHERE and then the PROJECT since it sits above, will be able to get if from it. If a WHERE doesn't exist, then the PROJECT will build it

ksqldb-engine/src/main/java/io/confluent/ksql/planner/plan/PullProjectNode.java
vpapavas: Yes, they are not exposed

ksqldb-engine/src/main/java/io/confluent/ksql/planner/plan/PullProjectNode.java
vpapavas: Correct

ksqldb-engine/src/main/java/io/confluent/ksql/planner/plan/PullProjectNode.java
vpapavas: Also, the `CodeGenRunner` uses the schema

ksqldb-engine/src/main/java/io/confluent/ksql/planner/plan/PullProjectNode.java
```diff
@@ -0,0 +1,257 @@
...
+import io.confluent.ksql.schema.ksql.types.SqlTypes;
+import io.confluent.ksql.util.KsqlConfig;
+import io.confluent.ksql.util.KsqlException;
+import java.util.List;
+import java.util.Objects;
+import java.util.Optional;
+import java.util.stream.Collectors;
+
+/**
+ * The projection of a Pull query.
```
agavra, line 42: thank you for the javadoc!

ksqldb-engine/src/main/java/io/confluent/ksql/planner/plan/PullProjectNode.java
```diff
@@ -0,0 +1,257 @@
...
+ *
+ * <p>There are 3 schemas the node is handling, the input schema, the intermediate schema used
+ * for codegen and the output schema.
+ * <ul>
+ * <li>The input is the schema of the child node
```
agavra, line 47: excuse me while I philosophize for a bit...

just wondering, is 'child' and 'parent' standard terminology in databases when discussing logical plans? I know we've standardized on it, but it always confuses me because the "child" is the "source node" which makes me feel like it's a "parent" 🤔 

What do you think about upstream/downstream nodes? That way it's clear that the direction of data is flowing from the source nodes (upstream) to the sink nodes (downstream)

ksqldb-execution/src/main/java/io/confluent/ksql/execution/transform/select/SelectValueMapperFactory.java
```diff
@@ -32,26 +33,34 @@
  * Factor class for {@link SelectValueMapper}.
  */
 public final class SelectValueMapperFactory {
-
   private static final String EXP_TYPE = "Select";
 
-  private final CodeGenRunner codeGenerator;
+  private CodeGenRunner codeGenerator;
```
agavra, line 38: any reason this is no longer `final`?

agavra (commented):
the structure lgtm! I didn't spend much time dissecting the functionality of the code though

ksqldb-engine/src/main/java/io/confluent/ksql/planner/plan/PullProjectNode.java
AlanConfluent: I guess I just meant that `schema` is only used with the `ExpressionTypeManager` and the comment is a bit misleading.

AlanConfluent (approved):

ksqldb-engine/src/main/java/io/confluent/ksql/planner/plan/PullProjectNode.java
vpapavas: The plans (logical, physical) are trees hence the terminology root, child, parent. The children are the leaves and as we move up the tree we visit from child to parent. Information and rows is flowing from the leaves towards the root.

Honestly, sorry but I find the upstream/downstream more confusing :(

agavra: 😂 haha i'll keep working at making it feel intuitive!

PR #6692, opened 2020-11-30
feat: support PROTOBUF keys

agavra:
### Description 

Supports protobuf keys (still under the config flag) 

### Testing done 

updated `protobuf.json` to have tests for protobuf keys

### Reviewer checklist

### DIFF
--- a/ksqldb-functional-tests/src/test/resources/query-validation-tests/protobuf.json
+++ b/ksqldb-functional-tests/src/test/resources/query-validation-tests/protobuf.json
@@ -294,6 +294,311 @@
       }
     },
     {
+      "name": "STRUCT BOOLEAN - key - no inference",
+      "comments": [
+        "note that null defaults to false in protobuf"
+      ],
+      "properties": {
+        "ksql.key.format.enabled": true
+      },
+      "statements": [
+        "CREATE STREAM INPUT (F1 BOOLEAN KEY, foo INT) WITH (kafka_topic='input_topic', format='PROTOBUF');",
+        "CREATE STREAM OUTPUT AS SELECT * FROM INPUT;"
+      ],
+      "inputs": [
+        {"topic": "input_topic", "key": {"F1": true}, "value": {"FOO": 10}},
+        {"topic": "input_topic", "key": {"F1": null}, "value": {"FOO": 10}},
+        {"topic": "input_topic", "key": null, "value": null}
+      ],
+      "outputs": [
+        {"topic": "OUTPUT", "key": {"F1": true}, "value": {"FOO": 10}},
+        {"topic": "OUTPUT", "key": {"F1": false}, "value": {"FOO": 10}},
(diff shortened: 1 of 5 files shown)


ksqldb-functional-tests/src/test/resources/query-validation-tests/protobuf.json
```diff
@@ -293,6 +293,311 @@
...
+    {
+      "name": "STRUCT BOOLEAN - key - no inference",
+      "comments": [
+        "note that null defaults to false in protobuf"
+      ],
+      "properties": {
+        "ksql.key.format.enabled": true
+      },
+      "statements": [
+        "CREATE STREAM INPUT (F1 BOOLEAN KEY, foo INT) WITH (kafka_topic='input_topic', format='PROTOBUF');",
```
vcrfxia, line 305: How come this works with declaring the key simply as `BOOLEAN` rather than `STRUCT<BOOLEAN>`? I thought keys were always forced to be unwrapped?

agavra: They are forced unwrapped if the format supports unwrapping, otherwise it uses what the format supports:

https://github.com/confluentinc/ksql/blob/master/ksqldb-engine/src/main/java/io/confluent/ksql/serde/SerdeFeaturesFactory.java#L90-L104

```java
  private static Optional<SerdeFeature> getKeyWrapping(
      final boolean singleKey,
      final Format keyFormat
  ) {
    // Until ksqlDB supports WRAP_SINGLE_KEYS in the WITH clause, we explicitly set
    // UNWRAP_SINGLE_KEYS for any key format that supports both wrapping and unwrapping to avoid
    // ambiguity later:
    if (singleKey
        && keyFormat.supportsFeature(SerdeFeature.UNWRAP_SINGLES)
        && keyFormat.supportsFeature(SerdeFeature.WRAP_SINGLES)
    ) {
      return Optional.of(SerdeFeature.UNWRAP_SINGLES);
    }
    return Optional.empty();
  }
```

vcrfxia: Ah, interesting. We should add tests for converting between protobuf and other key formats, and also joins between protobuf and other key formats. The rest of this PR LGTM, though! (Feel free to add in a separate PR if that's preferable.)

vcrfxia (approved):
LGTM, though see https://github.com/confluentinc/ksql/pull/6692#discussion_r532969819 for comment on additional test coverage. Thanks @agavra !

ksqldb-functional-tests/src/test/resources/query-validation-tests/protobuf.json
agavra: +1 I'll do this in a separate PR. I think the likelihood of an org using two formats is pretty low so it's OK if we release this with some rough edges around that (but definitely good to just have it working).

PR #6693, opened 2020-11-30
fix: Refactors rate limiter within LoggingHandler to be singleton

AlanConfluent:
### Description 
Changes the rate limiting logic within LoggingHandler to be a shared singleton.  Otherwise the limit is multiplied by the number of server verticles.

### Testing done 
Ran unit tests and ran high qps requests manually to verify correct logging.

### Reviewer checklist

### DIFF
--- a/ksqldb-rest-app/src/main/java/io/confluent/ksql/api/server/LoggingRateLimiter.java
+++ b/ksqldb-rest-app/src/main/java/io/confluent/ksql/api/server/LoggingRateLimiter.java
@@ -1,0 +1,67 @@
+/*
+ * Copyright 2020 Confluent Inc.
+ *
+ * Licensed under the Confluent Community License (the "License"); you may not use
+ * this file except in compliance with the License.  You may obtain a copy of the
+ * License at
+ *
+ * http://www.confluent.io/confluent-community-license
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OF ANY KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations under the License.
+ */
+
+package io.confluent.ksql.api.server;
+
+import static io.confluent.ksql.rest.server.KsqlRestConfig.KSQL_LOGGING_SERVER_RATE_LIMITED_REQUEST_PATHS_CONFIG;
+import static java.util.Objects.requireNonNull;
+
+import com.google.common.collect.ImmutableMap;
+import com.google.common.util.concurrent.RateLimiter;
+import io.confluent.ksql.rest.server.KsqlRestConfig;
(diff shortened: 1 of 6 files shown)


ksqldb-rest-app/src/main/java/io/confluent/ksql/api/server/LoggingRateLimiter.java
```diff
@@ -0,0 +1,67 @@
...
+  private final Map<String, Double> rateLimitedPaths;
+  private final Function<Double, RateLimiter> rateLimiterFactory;
+
+  private final Map<String, RateLimiter> rateLimiters = new ConcurrentHashMap<>();
+
+  public LoggingRateLimiter(final KsqlRestConfig ksqlRestConfig) {
+    this(ksqlRestConfig, RateLimiter::create);
+  }
+
+  public LoggingRateLimiter(
```
vcrfxia, line 41: This looks like it should be package-private and marked as VisibleForTesting?

ksqldb-rest-app/src/main/java/io/confluent/ksql/api/server/LoggingRateLimiter.java
```diff
@@ -0,0 +1,67 @@
...
+      final KsqlRestConfig ksqlRestConfig,
+      final Function<Double, RateLimiter> rateLimiterFactory) {
+    requireNonNull(ksqlRestConfig);
+    this.rateLimiterFactory = requireNonNull(rateLimiterFactory);
+    this.rateLimitedPaths = getRateLimitedRequestPaths(ksqlRestConfig);
+  }
+
+  public boolean shouldLog(final RoutingContext routingContext) {
+    if (rateLimitedPaths.containsKey(routingContext.request().path())) {
+      final String path = routingContext.request().path();
```
vcrfxia, line 51: Since only the path of the routing context is used here, would it make more sense to pass just the path into this method rather than the entire routing context?

ksqldb-rest-app/src/test/java/io/confluent/ksql/api/server/LoggingRateLimiterTest.java
```diff
@@ -0,0 +1,102 @@
...
+    assertThat(loggingRateLimiter.shouldLog(routingContext), is(true));
+    assertThat(loggingRateLimiter.shouldLog(routingContext), is(false));
+    assertThat(loggingRateLimiter.shouldLog(routingContext), is(false));
+
+    // Then:
+    verify(rateLimiter, times(4)).tryAcquire();
+  }
+
+  @Test
+  public void shouldLog_notIncluded() {
```
vcrfxia, line 94: nit (was confused by the name for a while):
```suggestion
  public void shouldLog_notRateLimited) {
```

vcrfxia (approved):
LGTM -- nice find and thanks for the fix!

PR #6694, opened 2020-12-01
feat: expose AVRO and JSON_SR as key formats

vcrfxia:
### Description 

This PR un-hides AVRO and JSON_SR from behind the feature flag, and cleans up a bunch of feature flag appearances in tests accordingly. Review the first commit to avoid scrolling through the new QTT historic plans. 

### Testing done 

QTT

### Reviewer checklist
- [ ] Ensure docs are updated if necessary. (eg. if a user visible feature is being added or changed).
- [ ] Ensure relevant issues are linked (description should include text like "Fixes #<issue number>")

PR #6695, opened 2020-12-01
chore: split key format feature flag

vcrfxia:
### Description 

This PR is stacked on https://github.com/confluentinc/ksql/pull/6694. Only the last commit needs to be reviewed.

This PR separates the current key format feature flag (`ksql.key.format.enabled`) into three:
- `ksql.key.format.enabled` : controls the addition of new key formats. I've kept the old name in order to avoid clashes with https://github.com/confluentinc/ksql/pull/6692
- `ksql.complex.key.format.enabled` : controls support for array and struct keys
- `ksql.multicol.key.format.enabled` : controls support for multi-column keys

### Testing done 

QTT

### Reviewer checklist

### DIFF
--- a/ksqldb-common/src/main/java/io/confluent/ksql/util/KsqlConfig.java
+++ b/ksqldb-common/src/main/java/io/confluent/ksql/util/KsqlConfig.java
@@ -138,7 +138,17 @@ public class KsqlConfig extends AbstractConfig {
   public static final String KSQL_KEY_FORMAT_ENABLED = "ksql.key.format.enabled";
   public static final Boolean KSQL_KEY_FORMAT_ENABLED_DEFAULT = false;
   public static final String KSQL_KEY_FORMAT_ENABLED_DOC =
-      "Feature flag for non-Kafka key formats";
+      "Feature flag for key formats under development";
+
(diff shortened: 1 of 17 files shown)


ksqldb-common/src/main/java/io/confluent/ksql/util/KsqlConfig.java
```diff
@@ -138,7 +138,17 @@
   public static final String KSQL_KEY_FORMAT_ENABLED = "ksql.key.format.enabled";
```
agavra, line 138: i've merged the other PR, so if you want to we can rename this (or just leave it as is, i'm OK with that too)

agavra (approved):
LGTM! Thanks @vcrfxia

ksqldb-common/src/main/java/io/confluent/ksql/util/KsqlConfig.java
vcrfxia: I'll leave it for now since the old name is already embedded in some historic plans (with value false), no need to introduce yet another config to clutter the history. Plus this flag is going away soon ;D

Issue #6557, opened 2020-11-02
latest_by_offset does not properly handle nulls with session windows


mikebin:
**Describe the bug**
`latest_by_offset` does not properly ignore `null` values when session windows are used in the aggregation query. A `null` value can be returned in this case. Other types of windows and non-windowed aggregations work fine.

**To Reproduce**
ksqlDB versions: 6.0.0, 0.12.0

```sql
create stream s (f1 string, f2 string) with (kafka_topic='s', value_format='json', partitions=1);

--Insert record with non-null field to be aggregated with latest_by_offset
insert into s (f1, f2) values ('a', 'v1');

--Works fine
select f1, latest_by_offset(f2) from s window session(10 minutes) group by f1 emit changes;
+-------------------------------------------------------------+-------------------------------------------------------------+
|F1                                                           |KSQL_COL_0                                                   |
+-------------------------------------------------------------+-------------------------------------------------------------+
|a                                                            |v1                                                           |
Query terminated

--Insert new record with null value for field f2
insert into s (f1) values ('a');

--null is returned for the aggregate value, instead of v1 as expected - this is the bug
select f1, latest_by_offset(f2) from s window session(10 minutes) group by f1 emit changes;
+-------------------------------------------------------------+-------------------------------------------------------------+
|F1                                                           |KSQL_COL_0                                                   |
+-------------------------------------------------------------+-------------------------------------------------------------+
|a                                                            |null                                                         |
Query terminated

--Works fine for other types of windows
select f1, latest_by_offset(f2) from s window tumbling(size 10 minutes) group by f1 emit changes;
+-------------------------------------------------------------+-------------------------------------------------------------+
|F1                                                           |KSQL_COL_0                                                   |
+-------------------------------------------------------------+-------------------------------------------------------------+
|a                                                            |v1                                                           |
Query terminated
```

**Expected behavior**
In the example above, value `v1` should be returned from the session-window aggregate query, instead of `null`

**Actual behaviour**
`null` is returned from the session-window aggregate, instead of being ignored.

**Additional context**
Problem appears to be a bug in the `merge` method.

vpapavas:
I am not sure if this should be classified as a bug of KSQL as it seems it depends on how streams handles session windows. For the tumbling window, there is only one window and both records belong to that window hence there is no merge called when aggregating. 
For the session window, since each key (even the same one) is a different session, during computation the two sessions are merged into one and the merge of the aggregator is called. 
[...]

PR #6699, opened 2020-12-01
fix: Fixes bug in latests-by-offset when using nulls and sessions windows

vpapavas:
### Description 
Fixes: https://github.com/confluentinc/ksql/issues/6557

### Testing done 
Added extra tests for ignoring nulls when merging

### Reviewer checklist

### DIFF
--- a/ksqldb-engine/src/main/java/io/confluent/ksql/function/udaf/offset/KudafByOffsetUtils.java
+++ b/ksqldb-engine/src/main/java/io/confluent/ksql/function/udaf/offset/KudafByOffsetUtils.java
@@ -72,19 +72,11 @@ final class KudafByOffsetUtils {
           return -1;
         } else if (struct2.get(VAL_FIELD) == null) {
           return 1;
+        } else if (struct1.get(VAL_FIELD) == null && struct2.get(VAL_FIELD) == null ){
+          return 0;
         }
 
-        // Deal with overflow - we assume if one is positive and the other negative then the
-        // sequence has overflowed - in which case the latest is the one with the smallest sequence
-        final long sequence1 = struct1.getInt64(SEQ_FIELD);
-        final long sequence2 = struct2.getInt64(SEQ_FIELD);
-        if (sequence1 < 0 && sequence2 >= 0) {
-          return 1;
-        } else if (sequence2 < 0 && sequence1 >= 0) {
-          return -1;
-        } else {
-          return Long.compare(sequence1, sequence2);
-        }
+        return INTERMEDIATE_STRUCT_COMPARATOR.compare(struct1, struct2);
       };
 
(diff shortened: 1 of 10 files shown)


ksqldb-engine/src/main/java/io/confluent/ksql/function/udaf/offset/KudafByOffsetUtils.java
```diff
@@ -64,6 +64,29 @@
     }
   };
 
+  static final Comparator<Struct> INTERMEDIATE_STRUCT_COMPARATOR_IGNORE_NULLS =
+      (struct1, struct2) -> {
+        // Ignore nulls: If one of the structs has a null value, then return the other irrespective
+        // of sequence.
```
agavra, line 70: nit: what if they're both null? then we should probably return 0 (I don't think it'll make a difference, but better to be correct)

agavra (approved):
LGTM! some minor comments inline

ksqldb-engine/src/main/java/io/confluent/ksql/function/udaf/offset/KudafByOffsetUtils.java
```diff
@@ -64,6 +64,21 @@
...
+        if (struct1.get(VAL_FIELD) == null) {
+          return -1;
+        } else if (struct2.get(VAL_FIELD) == null) {
+          return 1;
+        } else if (struct1.get(VAL_FIELD) == null && struct2.get(VAL_FIELD) == null) {
+          return 0;
+        }
```
agavra, line 77: this condition will never be hit - it needs to be the first in the if/else branches

ksqldb-engine/src/main/java/io/confluent/ksql/function/udaf/offset/LatestByOffset.java
```diff
@@ -50,6 +52,7 @@ private LatestByOffset() {
   }
 
   static AtomicLong sequence = new AtomicLong();
+  private static Comparator<Struct> comparator;
```
agavra, line 55: we should not be relying on static state - if people create two different latest by offests (one to ignore nulls and one not to) whichever is created second will override the comparator for the first. Instead, it should be created when `latestN` method is called as a local variable and referenced in the anonymous class 

In general, if you see code with static, non-final state it's probably a bug and best to avoid.

ksqldb-engine/src/main/java/io/confluent/ksql/function/udaf/offset/KudafByOffsetUtils.java
vpapavas: These are what happens when you work on the 3 PRs in parallel and the release

PR #6702, opened 2020-12-02
fix: use Java's Base64 instead of jersey's

niteshmor:
### Description 
Using Jersey's private utility function prevent Jersey upgrades in upstream rest-utils. This pull request replaces jersey's Base64 utility with Java's built-in Base64 in two tests.

### Testing done 
Tests run by CI

### Reviewer checklist


agavra (approved):
LGTM

PR #6707, opened 2020-12-02
Cleans up the tutorials section of the docs

MichaelDrogalis:
### Description 

We've had some content under the tutorials section for a while that doesn't match the use case mold and is duplicative. This patch redirects to the canonical sources of how to learn those things.


JimGalasyn (approved):
Nice!

PR #6709, opened 2020-12-02
fix: change locate() error message for a more user-friendly message

spena:
### Description 
Fixes https://github.com/confluentinc/ksql/issues/6249

Add a better error message when the issue in https://github.com/confluentinc/ksql/issues/6249 is thrown. 
```
ksql> CREATE TABLE MSG_COUNT WITH (KAFKA_TOPIC='MSG_COUNT', PARTITIONS=1, REPLICAS=1) AS SELECT
  'X' X,
  COUNT(*) MSG_CT
FROM PAGEVIEWS PAGEVIEWS
GROUP BY 'X'
EMIT CHANGES;

ksql> SELECT * FROM MSG_COUNT WHERE X='X';
Materialized data for key ['X'] is not available yet. Please try again later.
```

There was another fix that retry the failing request here (https://github.com/confluentinc/ksql/pull/6633), but seems pull queries must fail fast. Instead, the user should be warned about the issue and let them know they need to retry.

### Testing done 
Uni tests
Manually

### Reviewer checklist

### DIFF
--- a/ksqldb-streams/src/main/java/io/confluent/ksql/execution/streams/materialization/ks/KsLocator.java
+++ b/ksqldb-streams/src/main/java/io/confluent/ksql/execution/streams/materialization/ks/KsLocator.java
@@ -89,10 +89,11 @@ public final class KsLocator implements Locator {
 
       // Fail fast if Streams not ready. Let client handle it
       if (metadata == KeyQueryMetadata.NOT_AVAILABLE) {
-        LOG.debug("KeyQueryMetadata not available for state store {} and key {}",
+        LOG.debug("KeyQueryMetadata not available for state store '{}' and key {}",
             stateStoreName, key);
         throw new MaterializationException(String.format(
-            "KeyQueryMetadata not available for state store %s and key %s", stateStoreName, key));
+            "Materialized data for key %s is not available yet. "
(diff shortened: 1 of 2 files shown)


ksqldb-streams/src/main/java/io/confluent/ksql/execution/streams/materialization/ks/KsLocator.java
```diff
@@ -89,10 +89,11 @@
 
       // Fail fast if Streams not ready. Let client handle it
       if (metadata == KeyQueryMetadata.NOT_AVAILABLE) {
-        LOG.debug("KeyQueryMetadata not available for state store {} and key {}",
+        LOG.debug("KeyQueryMetadata not available for state store '{}' and key {}",
             stateStoreName, key);
         throw new MaterializationException(String.format(
-            "KeyQueryMetadata not available for state store %s and key %s", stateStoreName, key));
+            "Materialized data for key %s is not available yet. "
```
spena, line 95: I removed the `state store %s` because seems not useful for the user. The LOG will print it in the ksql.log debugging purposes.

AlanConfluent (approved):
LGTM

PR #6712, opened 2020-12-02
test: attempt to make CliTest more robust

agavra:
### Description 

We were getting failures like:
```
java.lang.UnsupportedOperationException: 
error msg: Could not write the statement 'TERMINATE CTAS_ID_15_9;' into the command topic.
Caused by: Timeout while waiting for command topic consumer to process command
	topic
	at io.confluent.ksql.cli.CliTest.setUp(CliTest.java:205)
```

I have no idea why those are happening, but I hope that this change (which just removes the parts which terminate queries) will make this better. Note that all queries are terminated after the class is done because `KSQL_REST_APP` is an `ExternalResource` which gets cleaned up by JUnit.

### Testing done 

test only change

### Reviewer checklist


spena (approved):
LGTM.

PR #6714, opened 2020-12-02
fix: Removes orphaned topics from transient queries

### LINKED ISSUE #4009: Restarting KSQL with active transient queries can leave orphaned internal topics
**Describe the bug**
It may be the case that restarting KSQL, while there are running push/transient query that uses internal changelog or repartition topics, results in the internal topics not being deleted

This has been removed on the community slack: 

https://confluentcommunity.slack.com/archives/C6UJNMY67/p1574980967041200

If this is the case, it's probably because we're not `close()`ing transient queries on shutdown, which would clean up these internal topics.  This would fix half the problem.

Of course, we may still have transient topics left around if a node _crashes_ or is killed while there are transient queries.  At the moment, a specific node would not be able to clean these up on restart as they could belong to another node.  We'd need to encode a node identify into the topic name to allow a restarted node to clean up orphaned topics.

AlanConfluent:
### Description 
Fixes #4009 

When a server is uncleanly shutdown or killed, it may leave orphaned topics from transient queries.  This PR attempts to clean up these topic when the server starts up.

This is accomplished by writing transient query application ids to disk of the local node.  This ensures that any topics are only attempted to be deleted by one node, the one that created it.  The data isn't durable in that a node can shutdown and not restart, potentially leaving state uncleaned up.

### Testing done 

### Reviewer checklist

### DIFF
--- a/ksqldb-engine/src/main/java/io/confluent/ksql/engine/OrphanedTransientQueryCleaner.java
+++ b/ksqldb-engine/src/main/java/io/confluent/ksql/engine/OrphanedTransientQueryCleaner.java
@@ -20,6 +20,7 @@ import static java.util.Objects.requireNonNull;
 import com.google.common.annotations.VisibleForTesting;
 import com.google.common.base.Strings;
 import io.confluent.ksql.config.SessionConfig;
+import io.confluent.ksql.exception.KafkaResponseGetFailedException;
 import io.confluent.ksql.services.KafkaTopicClient;
 import io.confluent.ksql.services.ServiceContext;
 import io.confluent.ksql.util.KsqlConfig;
@@ -56,7 +57,13 @@ public class OrphanedTransientQueryCleaner {
       return;
     }
     final KafkaTopicClient topicClient = serviceContext.getTopicClient();
-    final Set<String> topicNames = topicClient.listTopicNames();
+    final Set<String> topicNames;
+    try {
+      topicNames = topicClient.listTopicNames();
+    } catch (KafkaResponseGetFailedException e) {
+      LOG.error("Couldn't fetch topic names", e);
+      return;
+    }
     // Find any transient query topics
     final String queryApplicationIdPrefix = QueryApplicationId.buildPrefix(ksqlConfig, false);
     final List<String> orphanedTopics = topicNames.stream()
--- a/ksqldb-engine/src/main/java/io/confluent/ksql/engine/OrphanedTransientQueryCleaner.java
+++ b/ksqldb-engine/src/main/java/io/confluent/ksql/engine/OrphanedTransientQueryCleaner.java
@@ -17,19 +17,11 @@ package io.confluent.ksql.engine;
 
 import static java.util.Objects.requireNonNull;
 
-import com.google.common.annotations.VisibleForTesting;
-import com.google.common.base.Strings;
-import io.confluent.ksql.config.SessionConfig;
 import io.confluent.ksql.exception.KafkaResponseGetFailedException;
 import io.confluent.ksql.services.KafkaTopicClient;
 import io.confluent.ksql.services.ServiceContext;
-import io.confluent.ksql.util.KsqlConfig;
-import io.confluent.ksql.util.QueryApplicationId;
-import java.util.List;
 import java.util.Optional;
 import java.util.Set;
-import java.util.regex.Matcher;
-import java.util.regex.Pattern;
 import java.util.stream.Collectors;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
@@ -37,8 +29,6 @@ import org.slf4j.LoggerFactory;
 public class OrphanedTransientQueryCleaner {
 
   private static final Logger LOG = LoggerFactory.getLogger(OrphanedTransientQueryCleaner.class);
-
-  private static Pattern TRANSIENT_QUERY_APP_ID = Pattern.compile("(.*\\d+_\\d+)(-[a-zA-Z]+)+");
 
   private final QueryCleanupService cleanupService;
 
@@ -48,14 +38,8 @@ public class OrphanedTransientQueryCleaner {
 
   public void cleanupOrphanedInternalTopics(
       final ServiceContext serviceContext,
-      final SessionConfig config
+      final Set<String> queryApplicationIds
   ) {
-    final KsqlConfig ksqlConfig = config.getConfig(false);
-    final String nodeId = ksqlConfig.getString(KsqlConfig.KSQL_NODE_ID_CONFIG);
-    if (Strings.isNullOrEmpty(nodeId)) {
-      // If no node id is set, there are no internal topics to clean up
-      return;
-    }
     final KafkaTopicClient topicClient = serviceContext.getTopicClient();
     final Set<String> topicNames;
     try {
@@ -65,23 +49,12 @@ public class OrphanedTransientQueryCleaner {
       return;
     }
     // Find any transient query topics
-    final String queryApplicationIdPrefix = QueryApplicationId.buildPrefix(ksqlConfig, false);
-    final List<String> orphanedTopics = topicNames.stream()
-        .filter(topicName -> topicName.startsWith(queryApplicationIdPrefix))
-        .collect(Collectors.toList());
-    final Set<String> queryApplicationIds = orphanedTopics.stream()
-        .map(topicName -> {
-          final Optional<String> queryApplicationId
-              = parseTransientQueryApplicationIdFromTopicName(topicName);
-          if (!queryApplicationId.isPresent()) {
-            LOG.error("Couldn't parse application id from topic " + topicName);
-          }
-          return queryApplicationId;
-        })
+    final Set<String> orphanedQueryApplicationIds = topicNames.stream()
+        .map(topicName -> queryApplicationIds.stream().filter(topicName::startsWith).findFirst())
         .filter(Optional::isPresent)
         .map(Optional::get)
         .collect(Collectors.toSet());
-    for (final String queryApplicationId : queryApplicationIds) {
+    for (final String queryApplicationId : orphanedQueryApplicationIds) {
       cleanupService.addCleanupTask(
           new QueryCleanupService.QueryCleanupTask(
               serviceContext,
@@ -89,14 +62,5 @@ public class OrphanedTransientQueryCleaner {
               true
           ));
     }
-  }
-
-  @VisibleForTesting
-  static Optional<String> parseTransientQueryApplicationIdFromTopicName(final String topicName) {
-    final Matcher matcher = TRANSIENT_QUERY_APP_ID.matcher(topicName);
-    if (matcher.matches()) {
-      return Optional.of(matcher.group(1));
-    }
-    return Optional.empty();
   }
 }
--- a/ksqldb-engine/src/main/java/io/confluent/ksql/engine/OrphanedTransientQueryCleaner.java
+++ b/ksqldb-engine/src/main/java/io/confluent/ksql/engine/OrphanedTransientQueryCleaner.java
@@ -36,6 +36,12 @@ public class OrphanedTransientQueryCleaner {
     this.cleanupService = requireNonNull(cleanupService);
   }
 
+  /**
+   * Cleans up any internal topics that may exist for the given set of query application
+   * ids, since it's assumed that they are completed.
+   * @param serviceContext The service context
+   * @param queryApplicationIds The set of completed query application ids
+   */
   public void cleanupOrphanedInternalTopics(
       final ServiceContext serviceContext,
       final Set<String> queryApplicationIds
--- a/ksqldb-engine/src/main/java/io/confluent/ksql/engine/OrphanedTransientQueryCleaner.java
+++ b/ksqldb-engine/src/main/java/io/confluent/ksql/engine/OrphanedTransientQueryCleaner.java
@@ -1,0 +1,95 @@
+/*
+ * Copyright 2020 Confluent Inc.
+ *
+ * Licensed under the Confluent Community License (the "License"); you may not use
+ * this file except in compliance with the License.  You may obtain a copy of the
+ * License at
+ *
+ * http://www.confluent.io/confluent-community-license
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OF ANY KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations under the License.
+ */
+
+package io.confluent.ksql.engine;
+
+import static java.util.Objects.requireNonNull;
+
+import com.google.common.annotations.VisibleForTesting;
+import com.google.common.base.Strings;
+import io.confluent.ksql.config.SessionConfig;
+import io.confluent.ksql.services.KafkaTopicClient;
+import io.confluent.ksql.services.ServiceContext;
+import io.confluent.ksql.util.KsqlConfig;
+import io.confluent.ksql.util.QueryApplicationId;
+import java.util.List;
+import java.util.Optional;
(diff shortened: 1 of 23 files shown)


agavra:
There are some drawbacks with this approach:

- users have to carefully set the node.id for each of their nodes. this isn't always trivial in cloud environments (perhaps we could use k8s pod name - e.g. ksql-0?) when auto-scaling is enabled
- this approach makes nodes no longer interchangeable. if a customer scales down their cluster, a node.id would permanently disappear (leaving us open to having orphaned topics for that orphaned node.id)
- this opens the door to other features leveraging node.id, which is a double-edged sword (I've been arguing that we should have leader election and only have the leader write to the command topic, for example) because asymmetrical systems tend to be more difficult to operate, deploy and debug

I don't have any other alternatives off the top of my head, I'll give it some thought, but I think this merits further discussion. If we do decide to go down this path, might make sense to document this in a quick KLIP (I promise I'd review it promptly!).

AlanConfluent:
> * users have to carefully set the node.id for each of their nodes. this isn't always trivial in cloud environments (perhaps we could use k8s pod name - e.g. ksql-0?) when auto-scaling is enabled

That's pretty much what I had in mind.  You effectively are delegating sticky id generation (and enforcing id uniqueness) to whatever configures ksql -- in a cloud environment, it could be Kubernetes.  In a more static environment, it would just be a one-time generated id.

> * this approach makes nodes no longer interchangeable. if a customer scales down their cluster, a node.id would permanently disappear (leaving us open to having orphaned topics for that orphaned node.id)

That's true.  Given that this was being used for GCing, imperfect cleanup was considered acceptable, but I agree this isn't ideal.  I was trying to keep this somewhat simple since it's about cleanup rather than an essential part of functionality.

> * this opens the door to other features leveraging node.id, which is a double-edged sword (I've been arguing that we should have leader election and only have the leader write to the command topic, for example) because asymmetrical systems tend to be more difficult to operate, deploy and debug

That's fair.  People might not fully understand the limitations of how I had intended for this to be used and could misuse it.
 
> I don't have any other alternatives off the top of my head, I'll give it some thought, but I think this merits further discussion. If we do decide to go down this path, might make sense to document this in a quick KLIP (I promise I'd review it promptly!). (EDIT: or at least get multiple eyes on this)

I agree, I think that's fair.  Most of the alternatives I thought of for issuing/registering ids are a fair bit more complex.  For example:
 - You could use a consensus mechanism to try to take claim to an id on startup, knowing that an id is only issued to a single node.
 - Similarly, you could elect a coordinator and have it issue ids
 - We could do something similar to the way the command topic currently works (if my understanding is correct) and start a transaction and then do a read followed by a write (ensuring exclusive access) to the topic, allowing a test and set mechanism to be created.  Then you can effectively find the next id and claim it by writing to the topic.

Of course, in order to clean things up, you have to decide that the owner of some id is dead or been restarted and a similar protocol must then exist for a coordinator marking that id as dead, waiting a bit, and then GCing the topic.  We already do heartbeating for pull queries and could utilize that.

I actually thought of using the advertised listener instead of this since it's effectively the same idea (a unique address for each node) and already something we support.  The only issue is that not every configuration people seem to set up allows for inter-node communication.  If we assumed that were the case, then the problem is solved.

agavra:
> I agree, I think that's fair. Most of the alternatives I thought of for issuing/registering ids are a fair bit more complex.

Ideally, we'd have an alternative that doesn't involve ID generation at all. I think if we are going down that path, what you have here is the best for what we need. We can always make the ID generation more flexible (i.e. the suggestions you have) but we'll never be able to make it less - so it makes sense to start with static ID generation.

What do you think of a stupid-simple approach of just committing a list of transient queries to disk? That way we can track them and on start-up we can just delete all the ones that we were handling. It still has the potential to leak if a node goes down and never comes up, but that was a problem in either approach.

rodesai:
I agree that going the route of adding a per-node ID should have a KLIP. Adding this would be useful for more than just this type of garbage collection, but at the same time requires some care from an operator and has potential compatibility concerns. We should define:
    - how can the id be used. I think it's important to think ahead about what having an ID could buy us, because it will help us better think through any constraints it needs to have, and assess whether it's worth the compatibility issues raised below. Currently the proposal is to use it for resource tracking and garbage collection. But we could also use the id for:
        - static group membership in streams 
        - does it help our HA story to have predictable ids?
        - it could also be useful to store some metadata about each node, for example what it's last executed command is, so we can perform first-time-only initialization tasks for a query.
    - what are the requirements when setting these IDs
    - compatibility - what happens if the id is not specified? It would be simplest to just require it, but I think that's actually a pretty big requirement to add. But if we support not specifying the ID we need 2 behaviours every time the id is used.

That said, if we do have to set an ID it's trivial to set a unique predictable ID in CCloud, so that's not a concern, at least for us.

I do like the idea of keeping a registry of running queries on the local disk. It's expected that ksqlDB should have access to stable storage to function optimally, so we're not adding a new requirement here.

AlanConfluent:
> I agree that going the route of adding a per-node ID should have a KLIP. Adding this would be useful for more than just this type of garbage collection, but at the same time requires some care from an operator and has potential compatibility concerns. 

I agree that defining such a thing would require answering a lot of the questions you stated above.  I think that it's probably out of the scope of this GC case to try and tackle, since GC is much simpler potentially.


> I do like the idea of keeping a registry of running queries on the local disk as an alternative to the above. It's expected that ksqlDB should have access to stable storage to function optimally, so we're not adding a new requirement here.

I considered this and I think it's a reasonable approach. There's one drawback to only using local disk and that's if the same logical node gets restarted on a new machine, then cleanup isn't possible since there will be nothing saved to local disk.  Not sure how often that happens in various cloud environments, but it's certainly possible.  That said, if we're aiming to take care of the common case and this covers it, then that's fine by me.

agavra:
> That said, if we're aiming to take care of the common case and this covers it, then that's fine by me.

+1 I think that's what we should be aiming for - like you said earlier, it's mostly to GC things so if we miss it a few times in exceptional cases that's alright, the user can always cleanup these topics on their own (and there were exceptional cases in both proposals). It also makes life easier by avoiding thinking about the hard(er) problems 😅

AlanConfluent:
I've attempted to address your comments @rodesai @agavra and now write the query ids to disk.  Pleas take another look.

ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/LocalCommands.java
```diff
@@ -0,0 +1,182 @@
...
+    return currentLocalCommands.getFile();
+  }
+
+  public void processLocalCommandFiles(
+      final ServiceContext serviceContext
+  ) {
+    final FilenameFilter filter = (dir, fileName) -> fileName.endsWith(LOCAL_COMMANDS_FILE_SUFFIX);
+    final File[] files = directory.listFiles(filter);
+    if (files == null) {
+      throw new KsqlServerException("Bad directory " + directory.getAbsolutePath());
```
agavra, line 76:
```suggestion
      throw new KsqlServerException("Bad local commands directory " + directory.getAbsolutePath() + ". Please check your configuration for " + KsqlConfig.KSQL_LOCAL_COMMANDS_LOCATION_CONFIG);
```

ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/LocalCommands.java
```diff
@@ -0,0 +1,182 @@
...
+  }
+
+  public void write(final TransientQueryMetadata queryMetadata) {
+    try {
+      currentLocalCommands.write(
+          new LocalCommand(Type.TRANSIENT_QUERY, queryMetadata.getQueryApplicationId()));
+    } catch (IOException e) {
+      // Just log an error since not catching it would likely cause more cleanup work than this
+      // aims to fix.
+      LOG.error("Failed to write local command", e);
```
agavra, line 101:
```suggestion
      LOG.error("Failed to write local command for transient query:" + queryMetadata.getQueryApplicationId(), e);
```

ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/LocalCommands.java
```diff
@@ -0,0 +1,182 @@
...
+          + "' config in the properties file."
+      );
+    }
+    final File file = new File(directory, String.format("local_commands_%d_%s%s",
+        System.currentTimeMillis(), Integer.toHexString(RANDOM.nextInt()),
+        LOCAL_COMMANDS_FILE_SUFFIX));
+    return new LocalCommands(directory, ksqlEngine, LocalCommandsFile.createWriteable(file));
+  }
+
+  private void markFileAsProcessed(final File file) {
```
agavra, line 158: why not just delete the file? Do we every use these?

ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/LocalCommandsFile.java
```diff
@@ -0,0 +1,101 @@
...
+  public static LocalCommandsFile createWriteable(final File file) {
+    return new LocalCommandsFile(file, true);
+  }
+
+  public void write(final LocalCommand localCommand) throws IOException {
+    if (writer == null) {
+      throw new IOException("Write permission denied.");
+    }
+
+    final byte[] bytes = MAPPER.writeValueAsBytes(localCommand);
```
agavra, line 67: just making sure, this writes it in UTF8 so it's human readable json, right? I think it would be nice to make these files human readble

ksqldb-engine/src/main/java/io/confluent/ksql/engine/OrphanedTransientQueryCleaner.java
```diff
@@ -0,0 +1,66 @@
...
+      return;
+    }
+    // Find any transient query topics
+    final Set<String> orphanedQueryApplicationIds = topicNames.stream()
+        .map(topicName -> queryApplicationIds.stream().filter(topicName::startsWith).findFirst())
+        .filter(Optional::isPresent)
+        .map(Optional::get)
+        .collect(Collectors.toSet());
+    for (final String queryApplicationId : orphanedQueryApplicationIds) {
+      cleanupService.addCleanupTask(
```
agavra, line 58: since the file is only relevant on a new startup, does it need to be added to the cleanup service or can we just have it do this cleanup on startup?

agavra (commented):
overall lgtm, some inline comments

ksqldb-engine/src/main/java/io/confluent/ksql/engine/OrphanedTransientQueryCleaner.java
```diff
@@ -0,0 +1,66 @@
...
+    final Set<String> topicNames;
+    try {
+      topicNames = topicClient.listTopicNames();
+    } catch (KafkaResponseGetFailedException e) {
+      LOG.error("Couldn't fetch topic names", e);
+      return;
+    }
+    // Find any transient query topics
+    final Set<String> orphanedQueryApplicationIds = topicNames.stream()
+        .map(topicName -> queryApplicationIds.stream().filter(topicName::startsWith).findFirst())
```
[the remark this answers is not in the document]
AlanConfluent: So far as I can tell, nothing is very different in the cloud beyond setting some configs differently for making the query application id.  After all, so far as I can tell it's streams that is creating these topics and they just append to the query application id. 

I verified this manually with the transient topic names, so I'm pretty confident startsWith should work.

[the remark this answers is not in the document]
agavra: ack

ksqldb-engine/src/main/java/io/confluent/ksql/engine/OrphanedTransientQueryCleaner.java
```diff
@@ -0,0 +1,66 @@
...
+  }
+
+  public void cleanupOrphanedInternalTopics(
+      final ServiceContext serviceContext,
+      final Set<String> queryApplicationIds
+  ) {
+    final KafkaTopicClient topicClient = serviceContext.getTopicClient();
+    final Set<String> topicNames;
+    try {
+      topicNames = topicClient.listTopicNames();
```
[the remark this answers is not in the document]
AlanConfluent: In theory, it should process just one set of commands at a time during startup (it shouldn't be possible for there to be more than one unprocessed file).  When it does that, it should call this just once.  So it does this list call on node startup and uses that list of topics to compare with all of the application query ids in the commands file.  It would seem to be more expensive to say for every queryApplicationId, check the existence of a given topic.  Not only is that more rpcs, it also might check plenty of completed transient queries the cleaned up with no issue.  I considered adding a "done cleaning" entry to the commands so I could find the set of only orphaned queries, but that seems potentially error prone as well and requires writing another entry.  This seemed a little simpler.

Tell me if you'd prefer the "done cleaning" approach.

ksqldb-engine/src/main/java/io/confluent/ksql/engine/OrphanedTransientQueryCleaner.java
```diff
@@ -0,0 +1,66 @@
...
+    } catch (KafkaResponseGetFailedException e) {
+      LOG.error("Couldn't fetch topic names", e);
+      return;
+    }
+    // Find any transient query topics
+    final Set<String> orphanedQueryApplicationIds = topicNames.stream()
```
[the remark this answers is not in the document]
AlanConfluent: Yeah, I think we're on the same page now.  This is only ever run on startup and so by definition, the query application ids, if found as topic prefixes, would be orphaned.  Of course, that list is all transient queries run, not technically just the orphaned ones.

I'll improve the documentation.

ksqldb-engine/src/test/java/io/confluent/ksql/engine/OrphanedTransientQueryCleanerTest.java
```diff
@@ -0,0 +1,112 @@
...
+
+  @Before
+  public void setUp() {
+    when(serviceContext.getTopicClient()).thenReturn(topicClient);
+    cleaner = new OrphanedTransientQueryCleaner(queryCleanupService);
+  }
+
+  @Test
+  public void shouldCleanup_allApplicationIds() {
+    when(topicClient.listTopicNames()).thenReturn(ImmutableSet.of(TOPIC1, TOPIC2, TOPIC3));
```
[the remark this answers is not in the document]
AlanConfluent: Done

ksqldb-engine/src/main/java/io/confluent/ksql/engine/OrphanedTransientQueryCleaner.java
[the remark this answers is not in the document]
agavra: Yeah I agree the existing PR approach is better. At first, I didn't realize it happens just once on startup but since that's the case this is easier and less error-prone.

agavra (approved):
hitting the big green button

ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/LocalCommandsFile.java
AlanConfluent: Yep, it's UTF8: https://fasterxml.github.io/jackson-databind/javadoc/2.7/com/fasterxml/jackson/databind/ObjectMapper.html#writeValueAsBytes(java.lang.Object)

ksqldb-engine/src/main/java/io/confluent/ksql/engine/OrphanedTransientQueryCleaner.java
AlanConfluent: We chatted about this offline.  This takes advantage of the other cleanup routines done by `QueryCleanupService`, so needs access to it.  It does this cleanup once on startup

Issue #6650, opened 2020-11-20
Repartitioning tables for joins should result in a single changelog, not two


vcrfxia:
https://github.com/confluentinc/ksql/pull/6635 added support for repartitioning tables (for internal use cases only) in order to ensure records with logically equivalent keys join successfully when using Schema-Registry-enabled key formats. However, such repartitions currently result in two state stores (and changelog topics) per table, rather than one. This is because ksqlDB currently always materializes source tables so there's one materialization prior to the repartition, and also a second one after the repartition since a state store is needed for the join.

The main reason ksqlDB currently materializes all source tables is in order to prevent the source topic from being used as a changelog topic, should the query need a state store (see https://github.com/confluentinc/ksql/issues/5673 for details). However, in the case of repartitioning tables to enable joins, the source topic will not be used as a changelog topic as no state store is needed until after the repartition. As a result, the first state store (per table) is unnecessary, and should be removed to reduce duplication of data. Doing so requires one of two changes:
- finer-grained control over Kafka Streams topology optimizations. Specifically, ksqlDB would like to run with all topology optimizations enabled except the one for using source topics as changelogs. If this were possible, we could remove the logic in ksqlDB for forcing materializations on table sources, benefiting not just this join use case but also reducing data duplication for other table queries that do not need materialized stores.
- update the ksqlDB logical planner to traverse the execution step plan prior to passing the execution step plan to the plan builder for translation into a Kafka Streams topology. With this approach, the SourceBuilder can have information about the downstream topology in order to decide whether or not a materialization is needed at the table source.

PR #6715, opened 2020-12-03
chore: avoid extra state store when repartitioning tables for joins

vcrfxia:
### Description 

Fixes https://github.com/confluentinc/ksql/issues/6650 by taking the second of the two proposed solutions.

This PR introduces a `PlanInfoExtractor` which traverses the execution step plan to collect needed information to be passed to the `PlanBuilder`, for use when translating the execution step plan into an actual Kafka Streams topology. The `PlanInfoExtractor` collects information about whether sources are repartitioned downstream (prior to any joins). The `SourceBuilder` uses this information to decide whether or not to force materialization on source tables -- if the table is repartitioned downstream, it is not necessary to force materialization at the table source.

### Testing done 

Unit + QTT.

### Reviewer checklist


ksqldb-execution/src/main/java/io/confluent/ksql/execution/plan/ExecutionStep.java
```diff
@@ -60,7 +60,13 @@
   @JsonIgnore
   List<ExecutionStep<?>> getSources();
 
-  S build(PlanBuilder planBuilder);
+  default S build(PlanBuilder planBuilder) {
```
vcrfxia, line 63: The current expectation is that this variant is only called by QueryExecutor, on output nodes, and not from any intermediate nodes. Is it worth enforcing this assumption by throwing an exception if the method is called a non-output-node execution step?

ksqldb-streams/src/main/java/io/confluent/ksql/execution/streams/KSPlanInfoExtractor.java
```diff
@@ -0,0 +1,191 @@
...
+  @Override
+  public <K> KSPlanInfo visitTableTableJoin(final TableTableJoin<K> tableTableJoin) {
+    return visitJoinStep(tableTableJoin);
+  }
+
+  private KSPlanInfo visitSourceStep(final ExecutionStep<?> step) {
+    return new KSPlanInfo(step);
+  }
+
+  private KSPlanInfo visitRepartitionStep(final ExecutionStep<?> step) {
```
vcrfxia, line 177: Today, we only ever repartition tables to enable joins, and it is not possible to repartition a table after a table operation that requires materialization (e.g., joining on a table that is the result of grouping by another table, in the same query) so it is correct to only track repartitions when deciding whether to force materialization at the table source or not. However, if in the future we allow table repartitions after table operations that require materialization, then this logic will have to be updated to account for that. Does it make sense to future-proof this and add in the more complicated logic now, or would that be overcomplicating things?

ksqldb-execution/src/main/java/io/confluent/ksql/execution/plan/PlanInfo.java
```diff
@@ -0,0 +1,23 @@
...
+ * specific language governing permissions and limitations under the License.
+ */
+
+package io.confluent.ksql.execution.plan;
+
+/**
+ * Information about an execution step plan to be passed to a {@link PlanBuilder}
+ * for use when converting an execution step plan into a physical plan.
+ */
+public interface PlanInfo {
```
vcrfxia, line 22: Is it weird that this interface is empty? It's unclear to me which (if any) parts of the KSPlanInfo implementation are generic enough to warrant being moved into the interface.

ksqldb-functional-tests/src/test/resources/historical_plans/joins_-_stream-table_key-to-key_-_SR-enabled_key_format/6.2.0_1606950106050/topology
```diff
@@ -3,20 +3,17 @@ Topologies:
     Source: KSTREAM-SOURCE-0000000001 (topics: [t])
       --> KTABLE-SOURCE-0000000002
     Processor: KTABLE-SOURCE-0000000002 (stores: [])
-      --> KTABLE-MAPVALUES-0000000003
```
vcrfxia, line 6: Deleting and regenerating historic plans in this way is normally a major no-no but it was a conscious decision in this case as we have not released SR-enabled joins yet. This PR will be merged before 0.15 is cut, so these topology changes will not appear as a breaking change between releases.

ksqldb-engine/src/test/java/io/confluent/ksql/structured/SchemaKTableTest.java
```diff
@@ -15,24 +15,22 @@
 
 package io.confluent.ksql.structured;
 
-import static org.easymock.EasyMock.anyObject;
```
vcrfxia, line 18: Not sure why we had both easymock and mockito in this test file but easymock is gone now 💀

ksqldb-streams/src/main/java/io/confluent/ksql/execution/streams/SourceBuilder.java
```diff
@@ -330,12 +335,23 @@ private static PhysicalSchema getPhysicalSchema(final SourceStep<?> streamSource
...
+        // producer and KSQL (see https://issues.apache.org/jira/browse/KAFKA-10179
+        // and https://github.com/confluentinc/ksql/issues/5673 for more details)
+        table = source.mapValues(row -> row, materialized);
+      } else {
+        // if we know this table source is repartitioned later in the topology,
+        // we do not need to force a materialization at this source step since the
+        // re-partitioned topic will be used for any subsequent state stores, in lieu
+        // of the original source topic, thus avoiding the issues above.
+        // See https://github.com/confluentinc/ksql/issues/6650
+        table = source;
```
vcrfxia, line 353: I assumed if we're removing the Materialized we might as well remove the dummy mapValues step too. This change is backwards compatible (between releases) since this pathway is only hit in the case of forcing table repartitions to enable joins, which are new as of master (not currently released). However, if we think it's confusing that sometimes there is a dummy mapValues step on table sources and sometimes there isn't, I can put the dummy mapValues step back in with no materialization.

ksqldb-execution/src/main/java/io/confluent/ksql/execution/plan/PlanInfo.java
agavra: I think the abstract builder pattern makes sense for `PlanBuilder`, but less so for `PlanInfoExtractor` because we may want different `PlanBuilder`s if we decide to implement a non-KS physical plan (ignoring for a sec that `KStreamHolder` isn't actually generic enough to support non-KS implementations yet). The `PlanInfo`, however, should make sense to any corresponding physical plan we decide to build (it's an analysis of the steps and has nothing to do with Kafka Streams specifically). All we have to do is rename "repartition" to "keyChange" and this change becomes entirely generic. 

the tl;dr is that I don't think we need to have this be an interface (and I don't think we need an interface for `PlanInfoExtractor` either (though see my general comment about a different approach to this)

agavra (commented):
Nice! I think this definitely proves the approach that you had in mind, my comment is more from a code structure perspective/mental-modeling. It might make sense to just merge this now because we want to make sure it's in before 0.15 (and I want to leverge this change!) but I figured I'd toss down an alternative suggestion.

---

This PR takes the approach of flattening the physical plan tree into a single instance of `KSPlanInfo` and uses that as a dumping ground for information we might need along the path to building a physical plan (this is similar to the pattern we're using during our `Analysis` step of the AST before building the logical plan). This has some drawbacks; you already identified one of them: https://github.com/confluentinc/ksql/pull/6715/files?file-filters%5B%5D=.java#r534666939

Instead, I think we should take a page from what we have implemented for the `ExpressionVisitor` and implement something analogous for `ExecutionStep`. The top-level interface would be:
```java
interface ExecutionStepVisitor<R, C> {

   R visitStreamSource(StreamSource step, C context);
   R visitStreamFilter(StreamFilter step, C context);
   ...

}
```

Then, we could (in a future refactor) even make `KSPlanBuilder` just an implementation of `ExecutionStepVisitor<KStreamHolder, Void>` and we can also include the default traversal visitors to make for less boilerplate code going forward.

The first benefit is that we only have one mental model for working with `ExecutionStep`, making the code easier to pick up for new engineers working on this path. It also makes it easier to move from the expression handling code to the physical plan building code (you just need to learn the mental model once).

The second benefit, is that we don't have to have a single `PlanInfo` that represents everything we could ever want to know about the physical plan. We could take the path of just building mini-visitors each time we need it. This is OK because the physical plan tree is very small (it'll never get past tens of steps at the worst case) and the cost of traversing it multiple times is negligible (not to mention that anything done once-per-query is not worthy of optimization). This would mean that we could just implement:

```java
HasDownstreamRepartition implements ExecutionStepVisitor<Boolean, Void> {
  boolean visitTableSelectKey(TableSelectKey step, Void context) {
    return true;
  }

  boolean visitTableSink(TableSink step, Void context) {
    return false;
  }
}
```

And then we could get extra fancy, and even follow the path of having an `accept` method in the general `ExecutionStep` similar to the following in `Expression`:
```java
  protected abstract <R, C> R accept(ExpressionVisitor<R, C> visitor, C context);
```

vcrfxia:
> Instead, I think we should take a page from what we have implemented for the ExpressionVisitor and implement something analogous for ExecutionStep.

I actually did consider this while designing this PR -- I agree the visitor pattern is cleaner and I'd like for the `PlanBuilder` and `PlanInfoExtractor` to ideally share the same code. A sticking point for me is that the return types in PlanBuilder are strongly typed; each method is very particular about what its return type is, so the example `ExecutionStepVisitor` interface proposed above wouldn't work for `PlanBuilder`. I thought it was more beneficial for `PlanInfoExtractor` to follow the pattern set by `PlanBuilder` in terms of not requiring engineers to learn to work with multiple patterns for doing essentially the same thing (traversing the execution step plan), which is why I went with the approach in this PR. However, as you've pointed out, if we expect we'll want to add multiple execution step plan traversals in the future, refactoring `PlanInfoExtractor` into an `ExecutionStepVisitor` pattern would be better for modularity.

ksqldb-execution/src/main/java/io/confluent/ksql/execution/plan/PlanInfo.java
vcrfxia: I agree with your reasoning regarding the visitor pattern for `PlanBuilder`. My thinking in abstracting `PlanInfoExtractor` was that different plan builder implementations will likely want different information from the plan. For example, a non-Kafka-Streams implementation might not care about repartitions.

This is all very up in the clouds, though, since having a non-Kafka-Streams implementation is very far off. If you prefer nixing the interface I'm happy to do so.

agavra:
Talked offline - as a summary, we're going to stick with this approach and make it more generic if/when we need to add things to this that the model doesn't support.

ksqldb-execution/src/main/java/io/confluent/ksql/execution/plan/ExecutionStep.java
guozhangwang: +1. If we do not expect other callers, make it more strict is preferred.

guozhangwang:
> Talked offline - as a summary, we're going to stick with this approach and make it more generic if/when we need to add things to this that the model doesn't support.

I'm also in favor of decomposing the `PlanInfo` into finer modules that is only required at local visitors.

QQ: When we want to do any optimizations like filter pushdowns or other things that require the knowledge of the schema in the future, does that require this pattern to be modified again?

ksqldb-execution/src/main/java/io/confluent/ksql/execution/plan/PlanInfo.java
agavra: I'd rather nix the interface 😂 fewer lines of code to refactor when we inevitably do

agavra (approved):
LGTM! Thanks for the quick turnaround @vcrfxia

ksqldb-execution/src/main/java/io/confluent/ksql/execution/plan/ExecutionStep.java
vcrfxia: Turns out I was mistaken here. In the case of transient queries, there is no designated output node, so the QueryExecutor may call this variation of `build()` on essentially any node (whatever the last step in the execution step plan of a particular transient query happens to be).

ksqldb-streams/src/main/java/io/confluent/ksql/execution/streams/KSPlanInfoExtractor.java
vcrfxia: I'm going to merge this PR without future-proofing as there are other PRs blocked behind this.

@agavra do you think it makes sense to try to future-proof this logic in a follow-up PR? It should be a small change. The reason I'm hesitant is because it requires making strong assumptions about which steps require materialization at the Kafka Streams level, and it feels weird to bake in these assumptions as they could in theory evolve in the future.

vcrfxia:
> I'm also in favor of decomposing the PlanInfo into finer modules that is only required at local visitors.
> 
> QQ: When we want to do any optimizations like filter pushdowns or other things that require the knowledge of the schema in the future, does that require this pattern to be modified again?

Depends which schema you're referring to :) In the case of the filter pushdown optimization, only the expression analysis is needed. Because ksqlDB collects the full expression analysis before building even the "logical plan" (according to the terminology in the codebase, which is admittedly confusing / not 100% accurate), ksqlDB is able to, and in fact does even today, make this optimization without traversal of the execution step plan (as the optimization happens prior to building the execution step plan): https://github.com/confluentinc/ksql/blob/e6ba436536a10678eb9cbafa150c3fd0501d31ab/ksqldb-engine/src/main/java/io/confluent/ksql/planner/LogicalPlanner.java#L125-L129

However, if we wish to introduce other optimizations that require earlier execution steps to have information about later steps in the plan, then we will need to either enhance the PlanInfoExtractor introduced in this PR for extracting information about repartitions, or add another execution step visitor, at which point we would want to refactor the PlanInfoExtractor to become more modular in order for the new execution step visitor to share code with it.

guozhangwang:
> Depends which schema you're referring to :) In the case of the filter pushdown optimization, only the expression analysis is needed. Because ksqlDB collects the full expression analysis before building even the "logical plan" (according to the terminology in the codebase, which is admittedly confusing / not 100% accurate), ksqlDB is able to, and in fact does even today, make this optimization without traversal of the execution step plan (as the optimization happens prior to building the execution step plan):

That's crystal, thanks for the explanation!

ksqldb-streams/src/main/java/io/confluent/ksql/execution/streams/KSPlanInfoExtractor.java
agavra: The benefit of baking in the assumptions is that if they evolve we'll be conscious about it. I'm ok either way since we have pretty good test coverage here; we will catch in QTT if things change unexpectedly.

vcrfxia: Discussed (thoroughly 😂 ) offline: my concern is that if we support a new type of query that allows table aggregations prior to joins (in the same query), we may forget to update the logic in PlanInfoExtractor at that time, exposing us to the Avro schema name bug that the forced materialization at table sources tries to avoid. This would not be caught by existing QTTs as it would be for a new type of query. Summary of our discussion including how to proceed is here: https://github.com/confluentinc/ksql/issues/6750

PR #6718, opened 2020-12-04
chore: update processing log exists msg from info to warn (MINOR)

vcrfxia:
### Description 

The number of partitions for the processing log topic is configurable via the `ksql.logging.processing.topic.partitions` config. However, if the processing log topic already exists with a different number of partitions, the existing topic is left untouched and the ksqlDB server continues to start up after logging a message indicating that topic exists and the number of partitions was not as expected. (The same behavior is true as well for number of replicas.) This log message is currently logged at level INFO. This PR changes the log message to level WARN instead.

### Testing done 

Manual.

### Reviewer checklist

### DIFF
--- a/ksqldb-execution/src/main/java/io/confluent/ksql/exception/KafkaTopicExistsException.java
+++ b/ksqldb-execution/src/main/java/io/confluent/ksql/exception/KafkaTopicExistsException.java
@@ -17,6 +17,10 @@ package io.confluent.ksql.exception;
 
 public class KafkaTopicExistsException extends KafkaTopicClientException {
 
+  /**
+   * Whether the Kafka topic exists with an unexpected number of either
+   * partitions or replicas.
+   */
   private final boolean partitionOrReplicaMismatch;
 
   public KafkaTopicExistsException(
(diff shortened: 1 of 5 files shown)


ksqldb-execution/src/main/java/io/confluent/ksql/exception/KafkaTopicExistsException.java
```diff
@@ -17,8 +17,19 @@
 
 public class KafkaTopicExistsException extends KafkaTopicClientException {
 
-  public KafkaTopicExistsException(final String message) {
+  private final boolean partitionOrReplicaMismatch;
+
+  public KafkaTopicExistsException(
+      final String message,
+      final boolean partitionOrReplicaMismatch
```
vcrfxia, line 24: The value of this field is currently true in all cases, but given the generic name `KafkaTopicExistsException` I thought it'd be less confusing to add the extra field than assume that all instances of `KafkaTopicExistsException`s represent partition or replication factor mismatches.

agavra (approved):

PR #6721, opened 2020-12-04
docs: klip-42 - Schema Migrations Tool

spena:
### Description 
KLIP for the new schema migrations tool. 

### Testing done 

### Reviewer checklist

### DIFF
--- a/design-proposals/klip-42-schema-migrations-tool.md
+++ b/design-proposals/klip-42-schema-migrations-tool.md
@@ -389,6 +389,8 @@ Commands
   
 Options
   -c, --config-file  Specifies a configuration file
+  
+  -d, --define       Define variables for the client session (equivalent to the DEFINE statement).
  
   -h, --help         Shows this help  
     
--- a/design-proposals/klip-42-schema-migrations-tool.md
+++ b/design-proposals/klip-42-schema-migrations-tool.md
@@ -1,0 +1,442 @@
+# KLIP 42 - Schema Migrations Tool
+
+**Author**: Sergio Peña (@spena) |
+**Release Target**: TBD |
+**Status**: _In Discussion_ |
+**Discussion**: TBD
+
+**tl;dr:** _New tool to provide ksqlDB users for easy and automated schema migrations for their
+           ksqlDB environments. This allows users to version control their ksqlDB schema; recreate
+           their schema from scratch; and migrate their current schema to newer versions._
+
+## Motivation and background
+
+Schema migrations (also database migrations) refers to the process of performing updates or rollbacks on a database schema to a
+newer or older schema version. This practice is pretty common in any application lifecycle. Users use a version control for their
+applications which allow them to know if an application requires an upgrade or rolling back a buggy change. The same is necessary
+with a database schema. Users look for ways to version control schema changes along the application lifecycle.
+
+Existing tools exist to perform schema migrations on any database (MySQL, Postgres, Oracle, etc). These tools allow users to
+version the database schema, then automate the process to easily upgrade the schema to a newer version. If a schema is buggy,
+then users can rollback to the previous schema version. One more reason for using these tools is the integration with any
+CI/CD system, which allows users to test and verify a schema upgrade will work as expected.
+
+You can learn more about these existing tools:
+- [Flyway](https://flywaydb.org/)
+- [Liquibase](https://www.liquibase.org/)
+- [DBGeni](http://dbgeni.appsintheopen.com/index.html)
+
+ksqlDB requires the same kind of integration with CI/CD systems and automation to deploy ksqlDB schema changes in any environment.
+Some tools, like `Flyway`, support plugins to integrate schema migrations with a different non-supported database. However, our
+syntax is exclusively to ksqlDB; tools require JDBC support; and tools have several features that we cannot support
+(i.e. transactions & databases); so this integration becomes complex or unsupported. For that reason, ksqlDB must have its own
+schema migrations tool to provide the same migrations benefits to users who want to easily deploy and automate ksqlDB schema
+changes in all their ksqlDB environments.
+
+These benefits include:
+- Version the ksqlDB schema environments
+- Integrate ksqlDB schema changes with CI/CD environments
+- Simplify schema evolution changes across different environments
+
+This KLIP proposes a new tool for schema migrations. You will learn about the design aspects of the tool, and the features
+to support for a basic schema migration process.
+
+## What is in scope
+
+* Discuss design details for a new tool that provides schema migrations support for ksqlDB
+
+    Basic features to support:
+    - New CLI and API that can easily integrate with CI/CD environments
+    - Apply migrations on any ksqlDB environments
+    - Rollback migrated schemas to previous versions
+    - Version control ksqlDB schema
+
+* Provide a test plan that validates the new tool will meet the product requirements
+
+
+## What is not in scope
+
+* Some features found in existing migrations tools won't be supported
+
+  - `Execute entire migrations in a single transaction`
+
+    ksqlDB has support for a transactional metastore. However, this is limited to DDL statements that are persisted in the
+    Command topic. But DML statements, such as INSERT, write directly to the topic and do not work with transactions. This disallows
+    our tool to provide of a transaction support for the whole migration process. Also, DDL statements may create or delete topics,
+    which falls in the non-transactional process.
+
+  - `Support for simulated or dry-runs executions`
+
+    A dry-run requires the tool to know the current state of the ksqlDB schema before attempting to verify the new migrations scripts. This
+    requires a ksqlDB metastore exporting tool to work. Also, to make this simulation 100% safe, the tool requires a dummy Kafka and SR
+    environment that can validate issues with topics and SR subjects names, as well as security restrictions.
+
+  - `Other features, such as repeatable migrations and callbacks`
+
+    Not required for a basic migration.
+
+* Do performance analysis on migrations
+
+  In other DBs, there are operations that take too much time to complete. Such is the case of ALTER statements, which can add/remove columns
+  that would take time to complete on large tables. ksqlDB operations are quick unless an issue with the ksqlDB environment affects
+  these executions.
+
+## Value/Return
+
+Users will be able to integrate ksqlDB upgrades testing with any CI/CD environments of their choice. This is a huge benefit for users who want
+to automate ksqlDB upgrades with their application lifecycle.
+
+Also, a new tool will let users to easily automate schema evolution and migrations changes. They will be able to deploy new schema changes in
+all their environments (Prod, QA, Devel, etc).
+
+## Public APIS
+
+- No changes on current public APIs
+- A new Java API for Java developers
+
+  This seems important. However, it is in consideration if supporting a Java API is necessary.
+
+## Design
+
+I'm going to adopt `Flyway` and `DBGeni` tool syntax and behavior to design the ksqlDB migrations tool. Users will define a new migration in an SQL
+script. This new SQL script describes the changes to do to migrate the cluster from state A to state B. Then run the migration from the
+command line to apply the new state in the ksqlDB cluster. Users can also run this migration automatically as part of the build process and/or
+testing in a CI/CD environment.
+
+For instance, the following example creates a new file that setups the initial state of the cluster.
+
+`V1__Initial_setup.sql`
+```sql
+CREATE STREAM pageviews (
+    user_id INTEGER KEY, 
+    url STRING, 
+    status INTEGER
+) WITH (
+    KAFKA_TOPIC='pageviews', 
+    VALUE_FORMAT='JSON'    
+);
+
+CREATE TABLE pageviews_metrics AS
+ SELECT url, COUNT(*) AS num_views
+  FROM pageviews
+  GROUP BY url
+  EMIT CHANGES;
+```
+
+The user runs the migration tool on a specific ksqlDB cluster. The tool updates the cluster by running the SQL statements from the above file.
+Then sets the state of the cluster to version 1.
+
+```shell script
+$ ksql-migrations apply
+Current version of schema: << Empty Schema >>
+Migrating schema to version 1 - Initial setup
+```
+
+Later, the user needs new changes on the cluster. All previous migrations files are immutable. So, any changes on the cluster require a new migration
+file (or SQL script). Let's create one to create the users table.
+
+`V2__Add_users.sql`
+```sql
+CREATE TABLE users (
+   ID BIGINT PRIMARY KEY, 
+   STRING NAME, 
+   ADDRESS ADDRESS_TYPE
+ ) WITH (
+   KAFKA_TOPIC='users', 
+   VALUE_FORMAT='JSON' 
+ );
+```
+
+The user runs the migration tool on the same ksqlDB cluster. The tool detects the cluster has already version 1, so it executes only the newer version 2 migration
+file. It then sets the state of the cluster to version 2.
+
+```shell script
+$ ksql-migrations apply
+Current version of schema: 1
+Migrating schema to version 2 - Add users
+```
+
+The benefit of this tool is that it detects the required updates to execute in the ksqlDB cluster. So, users don't need to know which SQL statements need to perform
+to update the cluster. It also makes it easy to work with multiple clusters. Say that you have devel, stag and prod clusters. The tool will manage and track the version
+of these clusters and apply the right SQL operations.
+
+To be able to do that, the tool will use a metadata stream and table that contains the current schema version and all executed updates.
+
+### Schema metadata
+
+Each ksqlDB cluster requires metadata objects where to track the current schema state. This is not only useful for the tool to know the migrations files to apply, but also for
+users who can quickly verify if the cluster requires changes to fix a schema bug or add a major/minor improvement for their applications.
+
+The tool creates two metadata objects, a stream and table. automatically during the first migration; or when the user runs the tool with a parameter to initialize it.
+
+i.e.
+```
+$ ksql-migrations initialize
+Schema metadata initialized successfully
+```
+
+Due to some query limitations (lack of ORDER BY clause and pull queries working only on materialized views) in ksqlDB, the metadata will be stored in two places;  
+One stream (MIGRATION_EVENTS) and one table (SCHEMA_VERSION).
+
+The stream and table require topics unique for the cluster. In this case, these topics names will use the same convention as the processing log. It's not going
+to be an internal topic because it will be a user topic. The topic name is: `{clusterID}ksql_{StreamOrTableName}`.
+
+The user can specify a different name for the SCHEMA_VERSION table. This can be done through the tool configuration file. See `Configurations` for more details.
+
+The `MIGRATION_EVENTS` stream will keep track of every migration change (including undo changes). This will contain the history of changes the user has done. Also, this stream
+will contain a key to the current version of the schema (specified as `CURRENT` version). Every time a new migration or undo happens, the tool will insert a new event with the
+`CURRENT` key pointing to the current version.
+
+This is the `CREATE` statement for the `MIGRATION_EVENTS` stream:
+```sql
+CREATE STREAM migration_events (
+  version_key  STRING KEY,
+  version      STRING,
+  name         STRING,
+  state        STRING,
+  undoable     BOOLEAN,
+  checksum     STRING,
+  started_on   STRING,
+  completed_on STRING,
+  previous     STRING
+) WITH (  
+  KAFKA_TOPIC='default_ksql_migration_events',
+  VALUE_FORMAT='JSON',
+  PARTITIONS=1,
+  REPLICAS=1
+);
+```
+
+The `version_key` column has the version of the migration applied or undone. Other keys `CURRENT` and `LATEST` will be reserved for internal purposes.
+The `version` column has the version of the migration applied or undone.
+The `name` column has the name of the migration.
+The `state` column has the state of the migration process. It can be any of `Pending`, `Running`, `Migrated`, `Error`, `Undone`.
+The `undoable` column specifies if the migration can be undone or not. This requires an undoable migration file (See `Undo migrations`).
+The `checksum` column has the MD5 checksum of the migration file. It is used to validate the schema migrations with the local files.
+The `started_on` column has the date and time when the migration started.
+The `completed_on` column has the date and time when the migration finished.
+The `previous` column has the previous version applied.
+
+The `SCHEMA_VERSION` table will also keep track of every migration change (including undo changes), but with the difference that being a table the tool will see quickly if a schema
+version has been migrated or undone. It will also give us a quick view of the `CURRENT` state of the schema. The major advantage is that the tool will use pull queries in this materialized
+view to get the `CURRENT` state of the cluster.
+
+There is another reserved key `LATEST` that will point to the latest change applied (migrated or undone). This will be used by the tool to stream all changes up to the row that `LATEST` points, and stop.
+The `ksql-migrations info` command will use this to display information about the migrations that have been applied or undone. I cannot stream the `MIGRATION_EVENTS` or `SCHEMA_VERSION` directly because
+the tool does not know when to stop.
+
+
+This is the `CREATE` statement for the `SCHEMA_VERSION` table:
+```sql
+CREATE TABLE schema_version
+  WITH (
+    KAFKA_TOPIC='default_ksql_schema_version',
+    VALUE_FORMAT='JSON',
+    PARTITIONS=1
+  )
+  AS SELECT 
+    version_key, 
+    latest_by_offset(version) as version, 
+    latest_by_offset(name) AS name, 
+    latest_by_offset(state) AS state, 
+    latest_by_offset(undoable) AS undoable, 
+    latest_by_offset(checksum) AS checksum, 
+    latest_by_offset(started_on) AS started_on, 
+    latest_by_offset(completed_on) AS completed_on, 
+    latest_by_offset(previous) AS previous
+  FROM migration_events 
+  GROUP BY version_key;
+```
+
+The following are the outputs that we'll see on each stream and table, and how the tool will figure out the current schema version using pull queries.
+
+The `MIGRATION_EVENTS` stream output:
+```shell script
++-------------+---------+---------------+----------+----------+------------+---------------------+---------------------+----------+
+| version_key | version | name          | state    | undoable | checksum   | started_on          | completed_on        | previous |
++-------------+---------+---------------+----------+----------+------------+---------------------+---------------------+----------+
+| 1           | 1       | Initial setup | Migrated | No       | <MD5-sum>  | 12-01-2020 03:48:00 | 12-01-2020 03:48:05 | null     |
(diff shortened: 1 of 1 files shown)


design-proposals/klip-42-schema-migrations-tool.md
```diff
@@ -0,0 +1,442 @@
...
+    Basic features to support:
+    - New CLI and API that can easily integrate with CI/CD environments
+    - Apply migrations on any ksqlDB environments
+    - Rollback migrated schemas to previous versions
+    - Version control ksqlDB schema
+
+* Provide a test plan that validates the new tool will meet the product requirements
+
+
+## What is not in scope
```
MichaelDrogalis, line 57: Is squashing (periodically combining all previous migrations into one) out of scope? I don't think it needs to be a v1 thing, but this is historically important because over time, the value of having all migrations around diminishes.

design-proposals/klip-42-schema-migrations-tool.md
```diff
@@ -0,0 +1,442 @@
...
+
+## What is in scope
+
+* Discuss design details for a new tool that provides schema migrations support for ksqlDB
+
+    Basic features to support:
+    - New CLI and API that can easily integrate with CI/CD environments
+    - Apply migrations on any ksqlDB environments
+    - Rollback migrated schemas to previous versions
+    - Version control ksqlDB schema
```
MichaelDrogalis, line 52: One other thing that I think would be really useful would be producing a consolidated view of all the statements in one file. Rails, and I think Django, do this so you can see what is going on in one place. This consolidated view is typically checked into version control, and is always treated as read-only. @derekjn had some ideas about how this relates to or could be subsumed by supporting schema dumping.

design-proposals/klip-42-schema-migrations-tool.md
```diff
@@ -0,0 +1,442 @@
...
+
+Hope you have noticed the naming rules I followed in the previous examples for naming the migration files. Having a naming convention for files is necessary
+for the tool to detect what migration to apply or revert. Naming files are easier than creating configurations or command parameters to specify the same info.
+
+The migration files follow the same `Flyway` naming convention.
+`(Prefix)(Version)__(Description).sql`
+
+The `Prefix` specifies whether the file is a new migration (`V`) or undo (`U`) file.
+The `Version` is the version number used for the schema. For minor versions, such as `1.1`, an underscore is required (i.e. `1_1`)
+The `Description` is just a name or description of the new migration.
```
MichaelDrogalis, line 329: Is there any ceremony around the description, or can it be whatever? It looks like that is how you make those nice printouts when you migrate?

design-proposals/klip-42-schema-migrations-tool.md
```diff
@@ -0,0 +1,442 @@
...
+```
+
+### Configurations
+
+The tool will support a configuration file where all details for the server and tool migrations will be set. The following
+are the basic configurations that the tool should support.
+
+Example of `ksql-migrations.properties`
+```
+# Server URL and authentication
```
MichaelDrogalis, line 398: Is it worthwhile to support these as Docker environment variables, as we do with our general server configuration?

design-proposals/klip-42-schema-migrations-tool.md
```diff
@@ -0,0 +1,442 @@
...
+ksql.password='pass1'
+
+# Migrations details
+ksql.migrations.stream.name='migration_events'
+ksql.migrations.table.name='schema_version'
+```
+
+Note: The command line options and other configurations will also be defined during the implementation.
+
+### Questions
```
MichaelDrogalis, line 410: What would happen if multiple CI servers tried to migrate against the same cluster at the same time? Put differently, what sort of race conditions can happen?

MichaelDrogalis (approved):
This is awesome @spena! Incredible work. I really think this is one of the most important things we can add, and this it well thought out. Well done!

design-proposals/klip-42-schema-migrations-tool.md
```diff
@@ -0,0 +1,442 @@
...
+
+Undo a previous migration is necessary in the application lifecycle. A user sometimes want to revert the changes of an application because of a bug found in it.
+This also requires the database schema to be reverted or rollback to the previous version.
+
+The migrations tool allows users to revert changes. Note that undo is considered a forward migration. A new migration file is required which contains
+the SQL operations to revert the changes of a previous migration. The file in this case should contain the `U` prefix with the version number of the migration to rollback.
+
+For instance, let's undo the migration version 2 applied before. The migration file used before was named `V2__Add_users.sql`. For the undo file, we should add the `U` prefix.
+
+`U2_Add_users.sql`
```
colinhicks, line 302:
```suggestion
`U2__Add_users.sql`
```

design-proposals/klip-42-schema-migrations-tool.md
colinhicks: This will happen automatically, given the behavior of our docker utility belt.

design-proposals/klip-42-schema-migrations-tool.md
```diff
@@ -0,0 +1,442 @@
...
+Commands
+  initialize   Initializes the schema version table
+
+  apply ( all | next | until <target> )
+  
+              Migrates a schema to new available schema versions (default: next)
+              
+              If 'all' is specified, then it migrates all newer versions available
+              If 'next' is specified, then it migrates only the next available version
+              If 'until <target>' is specified, then it migrates all available versions before <target>
```
colinhicks, line 360: I would expect that it would apply all migrations to the target and including the target. I noticed this is also the behavior of dbgeni.

design-proposals/klip-42-schema-migrations-tool.md
```diff
@@ -0,0 +1,442 @@
...
+  
+  undo ( all | last | until <target> )
+  
+              Rollbacks a schema to the previous schema version (default: last)
+              
+              If 'all' is specified, then it rollbacks all previous versions
+              If 'last' is specified, then it rollbacks the previous version
```
colinhicks, line 367: Can we consider `previous` instead? I think this mirrors apply's `next` option better, personally.

design-proposals/klip-42-schema-migrations-tool.md
```diff
@@ -0,0 +1,442 @@
...
+  validate    Validate applied migrations against local files
+  
+              Compares local files checksum against the current metadata checksums to check for migrations files that have changed.
+              This tells the user that their schema might be not valid against their local files.
+  
+Options
+  -c, --config-file  Specifies a configuration file
+ 
+  -h, --help         Shows this help  
+
```
colinhicks, line 388: How feasible would it be to support variables / `--define` as a ksql-migrate CLI option?

Let's say I want to apply the same migration in dev and prod, but the topic names I use are different between environments. If I could specify `kafka_topic='${topicName}'` in my migration, I could reuse the migration with `ksql-migrate apply -d topicName=dev-mytopic`.

design-proposals/klip-42-schema-migrations-tool.md
```diff
@@ -0,0 +1,442 @@
...
+  previous     STRING
+) WITH (  
+  KAFKA_TOPIC='default_ksql_migration_events',
+  VALUE_FORMAT='JSON',
+  PARTITIONS=1,
+  REPLICAS=1
+);
+```
+
+The `version_key` column has the version of the migration applied or undone. Other keys `CURRENT` and `LATEST` will be reserved for internal purposes.
```
colinhicks, line 210:
```suggestion
The `version_key` column has the version of the migration applied or undone. Special values of the `version_key`, `CURRENT` and `LATEST`, will be reserved for internal purposes.
```

colinhicks (approved):
Great stuff, @spena!

design-proposals/klip-42-schema-migrations-tool.md
colinhicks: This will happen automatically, given the behavior of our docker utility belt.

design-proposals/klip-42-schema-migrations-tool.md
spena: It is out of scope for this tool. The proposal for squashing is to use a different tool that will export the ksqlDB metastore to a SQL file. Squashing will be a combination of exporting the metastore and replace the migrations files with this new file.

If we want to integrate squashing into the migrations tool, then we'll have to support the testing framework (YATT) in order to run every migration in a testing environment, then export the current env as SQL file. We still need the metastore export tool.

MichaelDrogalis: Makes sense, SGTM.

design-proposals/klip-42-schema-migrations-tool.md
spena: Just using underscores or spaces to separate words. If underscores are used, then they will automatically converted to spaces at runtime. I'll add that comment.

design-proposals/klip-42-schema-migrations-tool.md
spena: Locking the migrations stream and tables may help to avoid these race conditions. I need to think how to do that to also avoid locking an object during a failure in the migrations. Say the network connection fails, thus cannot unlock the stream and table.

I am thinking of adding another command to the tool to manually unlock or fix a failed migration. I'll add more info to the KLIP later.

design-proposals/klip-42-schema-migrations-tool.md
spena: Done

design-proposals/klip-42-schema-migrations-tool.md
spena: Yeah, I did that too initially, but I thought of keeping the `apply` and `undo` in sync 'cause they have `next` and `last`. Anyway, your expectation would be the same for many other users. I'll make that change.

design-proposals/klip-42-schema-migrations-tool.md
spena: Done

design-proposals/klip-42-schema-migrations-tool.md
spena: Done

design-proposals/klip-42-schema-migrations-tool.md
spena: Is this similar to Flyway's dry-runs feature? https://flywaydb.org/documentation/concepts/dryruns

design-proposals/klip-42-schema-migrations-tool.md
spena: Yes, those are automatically supported.

design-proposals/klip-42-schema-migrations-tool.md
spena: I considered using a second stream and table to store lock information, but it is getting complicated. I tried different ideas (which we can discuss offline), but I can't make it work. For instance, say we have a stream with schema `(id, locked, locked_by)` and the following rows:
```
id,  locked,  locked_by
* 1, true       u1
  1, true       u2
  1, true       u3
  1, false     u1
```
How can I create a table that shows me who has the lock? It should be `u1`. I was playing with `earliest_by_offset`, `latest_by_offset`. Also `latest_by_offset(id, 2)` to get the latest two to check in the application which one has the lock, but I forgot I could have more than 2 users trying to migrate. Without having transactions or a locking mechanism in ksqlDB, this is not reliable.

I think the only solution would be to support `LOCK` in ksqlDB, which requires session connections with the clients. This would let clients lock a table for the client session, thus avoiding other sessions read or write into a stream or table. @colinhicks We talked about sessions before, and I think we need this time. What do you think?

design-proposals/klip-42-schema-migrations-tool.md
spena: That's something we could add too. Seems useful. I'll add it to the klip.

spena: Done. I added the parameter to the tool help.

design-proposals/klip-42-schema-migrations-tool.md
MichaelDrogalis: Locking might get really complicated to add. We don't have to solve this right out of the gate, but we should know how it'll behave.

derekjn:
@spena fantastic KLIP, this looks great! I really like how you've designed the tool to leverage ksqlDB's core abstractions (streams, table, and pull queries).

Just a couple of points of feedback:

Firstly, since `undo` operations are forward migrations anyways, is there any benefit to managing a separate type of migration file for `undo` operations? In other words, can I do something with an `undo` operation that I couldn't just do with a forward migration? I understand that `undo` decreases the schema version, but are there any reasons why someone would do that instead of just using forward migrations? If users can achieve the same effect in two different ways, then I think we should consider keeping things simple by providing just one opinionated way to apply migrations, such as forward only.

I think that having an `undo` command would make more sense if each migration file contained any optional queries that would invert the migration (e.g. `DROP TABLE...`). That way, `undo` would tell the tool to run the undo component of the migration. I'm not suggesting that we add support for invertible migrations, just using it as an example. But if all migration files are technically forward migrations anyways, it seems to me that it isn't necessary to suggest otherwise in the tooling (unless of course there is some benefit of decreasing the schema version).

Secondly, I think it would be worth adding a command to the tool that creates an empty migration file for the next migration. This would eliminate a lot of the tedium and potential for mistakes around manually creating a properly named file. For example, this command would create an empty migration file based on the next schema version:

```
$ ksql-migrations add my_migration
Added V000099__my_migration.sql
```
_(Note that the naming convention should be sort friendly, which is why the schema version is 0-padded here)_

Users would then populate the migration file as necessary, and apply it like so:

```
$ ksql-migrations apply
Migrating schema to version 99 - my_migration
```

design-proposals/klip-42-schema-migrations-tool.md
derekjn: +1, I think this is a great idea @colinhicks.

derekjn:
@spena re: the above comment about `undo`, I think it would be most helpful if you could provide some examples/use cases for which having an `undo` operation would be valuable. cc @MichaelDrogalis

MichaelDrogalis:
I had a couple of ideas scratched down about that when I was spending some time on it...

1. Change the name of an object (stream, table, or column)
2. Drop an object
3. Seed a stream or table with rows once and only once

spena:
@derekjn There's no much benefits having an `undo` operation in migrations. Users can workaround it with a new migration file, and apply a forward migration. The only benefits I learned were:
 1) The user feeling that I can revert something I should haven't done.
 2) Remind users to always create migrations in pairs, with both 'V' and 'U' files, to have a revert operation in case of accidents.
     This recommendation helps to quickly revert something, and avoid users to pass through their application lifecycle to create,
     test, and deploy a new migration file. This can be even automated with `ksql-migrations add my_migration`

However, ksqlDB does not have many statements to revert. I walked through all of them, and I found these:

Supported
- CREATE has a DROP. However, If the user creates the topic ONLY IF it does not exist, would they want to undo with DELETE TOPIC?
- CREATE AS has DROP (terminates query automatically)
- DROP has CREATE. But, if I DROP DELETE TOPIC, then I will cause data loss that cannot be undone.

Non-supported
- CREATE OR REPLACE cannot remove columns. Removing columns is not supported due to schema evolution rules.
- ALTER ADD COLUMN does not have ALTER DROP COLUMN. Dropping columns is not supported due to schema evolution rules.
- TERMINATE does not have an undo operation (restart a query). I could drop the stream and create it again, but it will start with a new query ID, which may restart the offsets.
- DROP DELETE TOPIC cannot recover data. Unless I was dropping a CREATE AS stream, which can be recovered by creating  the stream & query again, and wait until the data is replayed from the source to the sink.
- INSERT does not have a DELETE (unsupported).

Based on that, you may be right that we don't need such a complex `undo` operation if users will hardly use it. We could add it in the future if users start asking for it and/or we support more syntax that can be reverted. Thoughts @MichaelDrogalis?

---

In regards to the `ksql-migrations add` proposal, that is very useful. I was debating with myself between using decimal versions or just integers (like DBGeni), but I felt inclined to the decimal versions based on this Hotfix use case (https://www.novatec-gmbh.de/en/blog/database-migration-flyway-large-project/). If you think that hotfix is not something we would see, then I can switch to your proposal. It is easier to sort.

MichaelDrogalis:
Thanks for the well thought out discussion @spena & @derekjn.

I think the way you laid it out is right—but I think (2) is much more important than we're estimating. Let's say I have a change that I deploy to production and something goes wrong. I immediately want to undo this change as fast as possible. If we don't support an undo operation, I'd have to go back and make a new migration, test it, check it in, and then apply it to my cluster. That is a lot of cognitive work to do when you're stressed out because production is broken.

The other use case that seems valuable is local development. Say I have 4 migrations, and I add another few queries in a fifth. I apply it my cluster, and it's _really_ close to correct, but not quite right. What do I do next? I could make a 6th migration to undo that, but because I haven't already checked in my 5th migration, it's kind of a waste. The other option is that I could nuke my development environment, but that seems extreme, too.

Anyway I don't feel super strongly that we need to support this in v1, but I do think it'll be a less than optimal experience for both of those situations soon down the road. Thoughts?

derekjn:
@spena @MichaelDrogalis if we're going to go with 0-padded migration numbers in the filenames, how many unique schema versions do you guys think we should support? I'm thinking that `6` digits would be plenty here.

MichaelDrogalis:
Seems sufficient to me, too.

design-proposals/klip-42-schema-migrations-tool.md
spena: I talked with @derekjn about this. This is the functionality we expect to have when working on the dump/export command for ksqlDB, which will export all metadata as a SQL file.

spena:
@derekjn @MichaelDrogalis I updated the KLIP with the latest feedback. If all looks good, can I have a quick dumps up so I can merge it?

The last 2 commits contain the following changes:

* Naming convention should use integers numbers (no decimals). It is easier to sort. There
  too few cases that require decimal versioning, and they might not even apply to ksqlDB
  migrations. Use 6 digit positions.
  
* Add undo operations as future work. ksqlDB has a few statements that support undo, and
  some of them are limited. Better to deal with these undo support in the future.
  
* Support dry-run operations. Dry-run should only verify what migrations will be applied
  in a specific cluster. There is no need to execute each migration statement.
  
* Consolidated views is what we expect dump/export command will do in ksqlDB. Nothing to
  add here.

MichaelDrogalis:
Perfect. Let's do this!

design-proposals/klip-42-schema-migrations-tool.md
```diff
@@ -0,0 +1,442 @@
...
+  - `Execute entire migrations in a single transaction`
+
+    ksqlDB has support for a transactional metastore. However, this is limited to DDL statements that are persisted in the
+    Command topic. But DML statements, such as INSERT, write directly to the topic and do not work with transactions. This disallows
+    our tool to provide of a transaction support for the whole migration process. Also, DDL statements may create or delete topics,
+    which falls in the non-transactional process.
+
+  - `Support for simulated executions`
+
+    A sumlation requires the tool to know the current state of the ksqlDB schema before attempting to verify the new migrations scripts. This
```
derekjn, line 68: Typo

derekjn (approved):

derekjn:
Excellent, looks good @spena!

design-proposals/klip-42-schema-migrations-tool.md
```diff
@@ -0,0 +1,442 @@
...
+Finally, let's look at the rest of the command parameters that will exist to facilitate schema migrations.
+
+```shell script
+Usage:
+  ksql-migrations [options] commands
+  
+Commands
+  new  <project-path>  Creates a new migrations project, directory structure and config file.
+
+  initialize   Initializes the schema version table
```
agavra, line 348: what happens if the user tries to initialize twice? I assume we should detect and fail?

design-proposals/klip-42-schema-migrations-tool.md
```diff
@@ -0,0 +1,442 @@
...
+ksql.migrations.tableName='schema_version'
+ksql.migrations.createSchemas='true'
+```
+
+Note: The command line options and other configurations will also be defined during the implementation.
+
+## Test plan
+
+- Verify positive and negative forward migrations
+- Verify integration with Github and Jenkins
```
agavra, line 419: do you have a description of how these integrations would look like? is that in-scope here or is that something that each developer would integrate on their own?

agavra (commented):
This is really neat (LGTM, post-merge)! Thanks for writing up such a detailed spec @spena. 

I have some comments in line, but what do you think about where this code should live? Does it make sense to have it in a separate repository? I think that would give a few benefits: (1) we could see the development experience for people using the java client and dogfood it ourselves without access to "internal" classes, (2) we wouldn't be tied to the build stability of ksql and (3) it would encourage developers to fork and contribute without overwhelming them with the burden of developing ksqlDB. Thoughts?

design-proposals/klip-42-schema-migrations-tool.md
```diff
@@ -0,0 +1,442 @@
...
+The `MIGRATION_EVENTS` stream output:
+```shell script
++-------------+---------+---------------+----------+------------+---------------------+---------------------+----------+
+| version_key | version | name          | state    | checksum   | started_on          | completed_on        | previous |
++-------------+---------+---------------+----------+------------+---------------------+---------------------+----------+
+| 1           | 1       | Initial setup | Migrated | <MD5-sum>  | 12-01-2020 03:48:00 | 12-01-2020 03:48:05 | null     |
+| 2           | 2       | Add users     | Migrated | <MD5-sum>  | 12-03-2020 10:34:30 | 12-03-2020 10:34:34 | 1        |
+| 2           | 2       | Add users     | Undone   | <MD5-sum>  | 12-03-2020 10:34:30 | 12-03-2020 10:34:34 | 1        |
+| CURRENT     | 1       | Initial setup | Migrated | <MD5-sum>  | 12-01-2020 03:48:00 | 12-01-2020 03:48:05 | null     |
+| LATEST      | 2       | Add users     | Undone   | <MD5-sum>  | 12-03-2020 10:34:30 | 12-03-2020 10:34:34 | 1        |
```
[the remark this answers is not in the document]
spena: Yes. CURRENT and LATEST will always point to the same row. The reason is the support of Undo operations, where CURRENT may point to a previous version. We're not going to support Undo for the moment. I left them just for future reference, but now I'm thinking about it, I think only one is necessary. We can add the LATEST pointer once Undo is supported.

design-proposals/klip-42-schema-migrations-tool.md
```diff
@@ -0,0 +1,442 @@
...
+The `Version` is the version number used for the schema. Versions will not support decimal versions. Integers with 6 digits are used.
+The `Description` is a name or description of the new migration. Description uses underscores (automatically replaced by spaces at runtime) or spaces separated the words.
+
+i.e.
+```
+- V000001__Initial_setup.sql    # a new version to migrate (v1) with name 'Initial setup'
+- V000002__Add_users.sql        # a new version to migrate (v2) with name 'Add users'
+```
+
+It is recommended the user creates these two files on any new migration. The tool will not enforce that. We need specify this recommendation in the ksqlDB documents.
```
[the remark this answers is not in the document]
spena: I need to remove that. We were going to support Undo operations, so these two files referenced to the `V` and `U` files. There is one file only now.

design-proposals/klip-42-schema-migrations-tool.md
spena: Yes. Although, you know. I think this is duplicated. The intention of `initialize` is to create an empty stream and table schema. And the intention of `baseline` is to create the stream and table schema too but with a version specified in the baseline parameter. Both are similar. 

I will remove the `baseline` and add the parameters to `initialize` to initialize to a specified version if desired.

design-proposals/klip-42-schema-migrations-tool.md
spena: Naa. I got excited in this plan. There is no integration. This plan is just run the main use case for having a migrations tool. So just a few tests that write a few migrations script, commit on github and automate Jenkins to pull and test. If everything works, then we're going to write a blog about it.

design-proposals/klip-42-schema-migrations-tool.md
```diff
@@ -0,0 +1,442 @@
...
+The `version` column has the version of the migration applied.
+The `name` column has the name of the migration.
+The `state` column has the state of the migration process. It can be any of `Pending`, `Running`, `Migrated`, `Error`, `Undone`.
+The `checksum` column has the MD5 checksum of the migration file. It is used to validate the schema migrations with the local files.
+The `started_on` column has the date and time when the migration started.
+The `completed_on` column has the date and time when the migration finished.
+The `previous` column has the previous version applied.
+
+The `SCHEMA_VERSION` table will also keep track of every migration change (including future work for undo changes), but with the difference that being a table the tool will see quickly if a schema
+version has been migrated or undone. It will also give us a quick view of the `CURRENT` state of the schema. The major advantage is that the tool will use pull queries in this materialized
```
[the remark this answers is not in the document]
spena: I'm not sure about this.

I understand that inserting a row in the stream may eventually be available in the MV for a pull query. So my idea was this:
- Insert a row in the stream for a Pending migration (i.e. v1)
- Run the pull query on the MV and wait until CURRENT is set to v1 and Pending (fail otherwise)
- Execute the v1 migration script
- Insert a row in the stream to change Pending -> Migrated
- Run the pull query on the MV and wait until CURRENT is set to v1 and Migrated (fail otherwise)
- Continue with the next migration script

But you mentioned queries at different times and different replicas. I think even a complete state store clean up would cause this consistency issues because when the ksql node is restarted, it will take time to sync the state store again. Do you have some ideas how to do it?

spena:
@agavra Thanks for the review. About using a different repo. I was thinking of a module only `ksqldb-tools` where we keep tools like this, export the metastore, etc. Using a different repo sound good too, but I'm not sure about it. A different repo will have a different version. We could match the version with ksqldb initially, but if the tool never changes, then it might look like an old version. @MichaelDrogalis What do you think about this? Would be good to have our own repo for tools?

MichaelDrogalis:
It's up to you all. But one heuristic that's always worked well for me in the past is to start with a monorepo and split it up over time as needed. There's more to maintain when you split it up, and you'll pretty quickly realize whether that is worth doing or not if you start with just one thing.

design-proposals/klip-42-schema-migrations-tool.md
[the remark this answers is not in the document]
agavra: I don't really have a good idea about how to work around this until we have vector clock support (cc @vvcephei) so that we can ensure we're at least up to date as of a certain point

dttouchdata:
First of all, sorry If I am posting this comment in wrong place. But I am running into an issue: I am trying to use ksql-migrations utility to deploy ksql scripts. Part of this, i am using environment variables (for endpoint, key, secrect etc.,) in ksql-migrations.properties file and setting values for these variables in the invoking unix shell (eg., export CCLOUD_KSQL_ENDPOINT=https://abcde.com/). But ksql-migrations utility isn’t taking value from environment variable and erroring out with url  CCLOUD_KSQL_ENDPOINT not found. If anyone used this utility, how did you handle this? Of course, it works fine if i specify actual values in properties file (but we don’t want to hardcode secrets there). Thanks

Issue #824, opened 2018-02-28
Add support for generic message keys (e.g. message keys in Avro or JSON format, STRUCT)


apurvam:
Right now, KSQL only uses a string Serde for keys in kafka messages. This works well for doing joins and aggregates on the entire key. However, in some cases a key (in json or avro or any other format) may have nested fields, and we may want to do joins and aggregates on a per field basis. 

Further, by supporting structured keys, we can do filters on the fields of the keys as well. 

This is related to #638 and #804.

We should also support non-string keys, such as Integer.

mquraishi:
Additionally, 1) If the value (not key) is an avro containtered object with the schema visible in clear text but the payload itself is a bytearray, KSQL cannot deserialize correctly.  In code it is easy.  This is once the message is retrieved from the topic.  2) If the value is json and one of the elements is another nested json object, creating a stream turns the nested son into some dictionary key=value object, resulting in deserialization errors.

miguno:
For the record:  Another +1 from a user (offline conversation).

big-andy-coates:
Hi @mquraishi,

> Additionally, 1) If the value (not key) is an avro containtered object with the schema visible in clear text but the payload itself is a bytearray, KSQL cannot deserialize correctly. In code it is easy. This is once the message is retrieved from the topic. 2) If the value is json and one of the elements is another nested json object, creating a stream turns the nested son into some dictionary key=value object, resulting in deserialization errors.

I think you're talking about an Avro serialized value where the schema is serialized as part of the message, right?  At the moment I don't believe KSQL supports such a serialization format.  KSQL does not expect the schema to be provided in the record.  It expects the schema from the Schema Registry or provided by the users when importing the data into KSQL.

Supporting embedded schemas would be a different issue to this one. If you're still interested in seeing this worked on would you mind detailing it in a new Github issue please?

big-andy-coates:
Please find attached a document comparing two potentially approaches we're looking into for structured key support in KSQL.  We are looking for feedback from the community on the two approaches.  We're wanting to move on this pretty quick: a decision on direction is likely to be taken by the end of this week, i.e. the Friday 8th Feb.  Please leave any comments below.  In addition, you can use the two additional comments below to vote for one of the approaches.

[Structured Keys Comparison (1).pdf](https://github.com/confluentinc/ksql/files/2830151/Structured.Keys.Comparison.1.pdf)

We look forward to hearing from you!

big-andy-coates:
[...]

PR #6722, opened 2020-12-04
feat: expose support for array and struct keys

vcrfxia:
### Description 

This PR adds more test coverage for array and struct keys, and exposes the feature to the world (by removing the relevant feature flag) 🎉 (partial progress towards https://github.com/confluentinc/ksql/issues/824)

Four commits for ease of reviewing:
- additional QTT test coverage
- adds integration test coverage, via the HTTP2 API and Java client tests
- removes the feature flag
- adds QTT historic plans

### Testing done 

See above -- lots of new tests!

### Reviewer checklist


ksqldb-serde/src/main/java/io/confluent/ksql/serde/GenericKeySerDe.java
```diff
@@ -148,21 +148,12 @@ private static void checkUnsupportedSchema(
...
-      }
-    } else {
-      for (final SimpleColumn column : columns) {
-        final SqlType sqlType = column.type();
-        if (!(sqlType instanceof SqlPrimitiveType || sqlType instanceof SqlDecimal)) {
-          throw new KsqlException("Unsupported key schema: " + columns);
-        }
+    for (final SimpleColumn column : columns) {
+      if (containsMapType(column.type())) {
+        throw new KsqlException("Map keys, including types that contain maps, are not supported "
```
agavra, line 153: (from a previous PR) we might want to link to https://github.com/confluentinc/ksql/issues/6621 in the exception of the message if anyone needs to "continue the conversation"

agavra (approved):
Thanks for all the added test coverage @vcrfxia!

ksqldb-serde/src/main/java/io/confluent/ksql/serde/GenericKeySerDe.java
vcrfxia: Sorry, must've missed the previous comment. Updated in this PR. I went with simply `See https://github.com/confluentinc/ksql/issues/6621 for more.` rather than language about "continuing the conversation" since the latter makes it sound like this is an ongoing discussion which it isn't (at least as of now).

agavra: ah sorry my comment wasn't clear! I meant (this is code from a previous PR) not (I left a comment on a previous PR). And definitely don't think we should include "continue the conversation" in the exception message! I meant it as that's what they'd be doing.

vcrfxia: Ah got'cha. Glad we're on the same page :)

PR #6726, opened 2020-12-05
feat: Adds table scans to pull queries

AlanConfluent:
### Description 
This PR adds the ability to table scan to pull queries.  Currently, projections work with them, though currently, nothing is supported in the where clause.  This should come later.  It's disabled by default, but can be enabled with the config `ksql.query.pull.table.scan.enabled`.

I didn't add any documentation yet because this should be considered an experimental change with limited functionality.  Benchmarking and where clause support are required before this should be considered a non experimental.

A table scan is issued like:
```
select * from RATINGS_TABLE;
```

### Testing done 
Unit tests + RQTT + manual testing

### Reviewer checklist


ksqldb-streams/src/main/java/io/confluent/ksql/execution/streams/materialization/ks/KsLocator.java
```diff
@@ -108,20 +146,42 @@
...
       keysByPartition.get(metadata.partition()).add(key);
+      metadataList.add(metadata);
+    }
+    return new Metadata(metadataList, Optional.of(keysByPartition));
+  }
 
-      if (locationsByPartition.containsKey(metadata.partition())) {
-        continue;
-      }
+  private Metadata getMetadataForAllPartitions(final Set<Integer> filterPartitions) {
```
vpapavas, line 154: We should add tests in `KsLocatorTest` to make sure that the data structures are populated correctly with all active and standby partitions the underlying store has

ksqldb-streams/src/main/java/io/confluent/ksql/execution/streams/materialization/ks/KsMaterializedSessionTable.java
```diff
@@ -65,6 +66,12 @@
     }
   }
 
+  @Override
+  public Iterator<WindowedRow> get(final int partition, final Range<Instant> windowStartBounds,
+      final Range<Instant> windowEndBounds) {
+    throw new MaterializationException("Table scan unsupported on session tables");
```
vpapavas, line 72: Why is this?

ksqldb-common/src/main/java/io/confluent/ksql/util/IteratorUtil.java
```diff
@@ -0,0 +1,75 @@
...
+  }
+
+  public static <T> Iterator<T> of(final T... elements) {
+    return ImmutableList.copyOf(elements).iterator();
+  }
+
+  private static class IteratorWithCallbacks<T> implements Iterator<T> {
+
+    private final Iterator<T> backingIterator;
+    private final Runnable completeRunnable;
```
vpapavas, line 37: What's the purpose of this iterator? Why do we need a special class? I see that this is used to close the iterator? Aren't iterators closed when done automatically?

vpapavas (approved):
LGTM! Thanks @AlanConfluent

PR #6732, opened 2020-12-08
fix: (minor) don't use deprecated jersey calls

niteshmor:
### Description 

This pull request updates the call signature for jersey's `ContainerRequest(...)` to a non-deprecated one in a test. Jersey 2.30 and above add a new `Configuration` parameter (see [here](https://javadoc.io/static/org.glassfish.jersey.bundles/jaxrs-ri/2.30/org/glassfish/jersey/server/ContainerRequest.html#constructor.summary)). 

The context for this pull request is that I am trying to update Jersey to 2.31 in upstream rest-utils (https://github.com/confluentinc/rest-utils/pull/210). Ksql treats warnings as errors, and the deprecation warning causes the upstream pull request from building correctly. Once that pull request actually upgrades the jetty version, ksql builds would fail without this current PR specifically because of treating deprecation warnings as errors. 

Merging this pull request will cause the immediate builds to break, but the current ksql build is already broken as of now. The reason for the current broken build is multiple jetty versions, which the upstream PR attempts to fix.

### Testing done 
None

### Reviewer checklist

### DIFF
--- a/ksql-rest-app/src/test/java/io/confluent/ksql/rest/server/filters/KsqlAuthorizationFilterTest.java
+++ b/ksql-rest-app/src/test/java/io/confluent/ksql/rest/server/filters/KsqlAuthorizationFilterTest.java
@@ -138,7 +138,8 @@ public class KsqlAuthorizationFilterTest {
         URI.create(path),
         method,
         securityContext,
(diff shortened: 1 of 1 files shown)


spena (approved):
lgtm

Issue #4219, opened 2020-01-04
Create Avro sink schema at time of stream/table creation


vcrfxia:
**Is your feature request related to a problem? Please describe.**

Currently, when a user creates an Avro sink stream/table, KSQL does not register the Avro schema for the new topic to Schema Registry until the first record is processed. (This is because KSQL relies on the `KafkaAvroSerializer` to automatically register the schema to Schema Registry, and the serializer is only invoked when a record is processed.) However, this leads to unintuitive behavior such as in https://github.com/confluentinc/ksql/issues/3883:
- user issues a query that creates a sink stream with Avro format
- user wishes to define a KSQL table on the sink topic (from the stream above), but the request fails due to `Avro schema for message values on topic <topic name> does not exist in the Schema Registry` 
- a record is sent to through the Streams app defined by the query
- user reissues the create table statement, and this time it succeeds

**Describe the solution you'd like**

KSQL should register the Avro schema with Schema Registry at the time the sink stream/table is defined, rather than waiting until the first record is processed.

**Describe alternatives you've considered**

N/A

**Additional context**

N/A

vcrfxia:
Copied over the 0.7.0 milestone from https://github.com/confluentinc/ksql/issues/3883

PR #6733, opened 2020-12-08
Add possibility to configure value.subject.name.strategy

Hubbitus:
### Description 
Also add strategy TopicNameStrategyPlain for search schemes <topic>-key for keys and <topic> for values.

Default [`SubjectNameStrategy`](https://github.com/confluentinc/schema-registry/blob/3037893a3a2fa9650999cbca11100f1b19095ba4/schema-serializer/src/main/java/io/confluent/kafka/serializers/subject/strategy/SubjectNameStrategy.java#L29) is [`TopicNameStrategy`](https://github.com/confluentinc/schema-registry/blob/3037893a3a2fa9650999cbca11100f1b19095ba4/schema-serializer/src/main/java/io/confluent/kafka/serializers/subject/TopicNameStrategy.java#L30): for any messages published to `topic` the schema of the message key is registered under the subject name `topic-key`, and the message value is registered under the subject name `topic-value.

In some cases, it is not desired to have `-key`, `-value` value suffixes. For example in case you are do **not** use schemas for keys (or rarely), and assume subject name in schema-registry equal to `topic` name is for values and `topic-key` for keys

That may be configured in properties file like:
```
value.subject.name.strategy=io.confluent.ksql.serde.avro.subject.TopicNameStrategyPlain
```
Please refer to the [documentation](https://docs.confluent.io/platform/current/schema-registry/serdes-develop/index.html#configuration-details) there explicitly stated limitation what such configuration is not supported currently in `ksqldb`.

**So that pull request brings it.**

Also, add strategy `TopicNameStrategyPlain` for search schemes `topic-key` for keys and `topic` for values.

### Testing done 
None autotests
Manual testing: configure ksql-server.properties to have line:
```
value.subject.name.strategy=io.confluent.ksql.serde.avro.subject.TopicNameStrategyPlain
```
and check it works for schemes subject in schema-registry match `<topic>` for values, and not `<topic>-value` name.

### Reviewer checklist


Hubbitus:
[clabot:check]

Hubbitus:
Jemkins fail is looks strange and not-related to my changes. Can (should) I deal with it?

agavra:
Thanks for the PR @Hubbitus! I'll take a look at this sometime soon and give you some feedback. In the meantime, the failure looks related to checkstyle on your changes:
```
[2020-12-08T15:06:30.823Z] [ERROR] /home/jenkins/workspace/confluentinc-pr_ksql_PR-6733/ksqldb-common/src/main/java/io/confluent/ksql/util/KsqlConfig.java:334: Line is longer than 100 characters (found 131). [LineLength]
[2020-12-08T15:06:30.823Z] [ERROR] /home/jenkins/workspace/confluentinc-pr_ksql_PR-6733/ksqldb-common/src/main/java/io/confluent/ksql/util/KsqlConfig.java:335: Line is longer than 100 characters (found 124). [LineLength]
[2020-12-08T15:06:30.824Z] [ERROR] /home/jenkins/workspace/confluentinc-pr_ksql_PR-6733/ksqldb-common/src/main/java/io/confluent/ksql/util/KsqlConfig.java:337: Line is longer than 100 characters (found 115). [LineLength]
[2020-12-08T15:06:30.824Z] [ERROR] /home/jenkins/workspace/confluentinc-pr_ksql_PR-6733/ksqldb-common/src/main/java/io/confluent/ksql/util/KsqlConfig.java:339: Line is longer than 100 characters (found 120). [LineLength]
[2020-12-08T15:06:30.824Z] [ERROR] /home/jenkins/workspace/confluentinc-pr_ksql_PR-6733/ksqldb-common/src/main/java/io/confluent/ksql/util/KsqlConfig.java:340: Line is longer than 100 characters (found 121). [LineLength]
```

As a hint, I recommend going to the pipeline steps page on Jenkins to look for the failing task:
image

Hubbitus:
Hello, @agavra.

Thanks for the feedback. I have truncated long lines.

Meantime is it possible for me to see such intentions from Jenkins?
If I follow to build [details](https://jenkins.confluent.io/job/Confluentinc%20Contributors/job/ksql/job/PR-6733/4/) I see build failed, and [console output](https://jenkins.confluent.io/job/Confluentinc%20Contributors/job/ksql/job/PR-6733/4/console) contains error:
```
java.io.IOException: cannot find current thread
	at org.jenkinsci.plugins.workflow.cps.CpsStepContext.doGet(CpsStepContext.java:300)
	at org.jenkinsci.plugins.workflow.cps.CpsBodySubContext.doGet(CpsBodySubContext.java:88)
	at org.jenkinsci.plugins.workflow.support.DefaultStepContext.get(DefaultStepContext.java:67)
	at org.jenkinsci.plugins.credentialsbinding.impl.BindingStep$Callback.finished(BindingStep.java:255)
	at org.jenkinsci.plugins.credentialsbinding.impl.BindingStep$Execution2$Callback2.finished(BindingStep.java:163)
	at org.jenkinsci.plugins.workflow.steps.GeneralNonBlockingStepExecution$TailCall.lambda$onSuccess$0(GeneralNonBlockingStepExecution.java:140)
	at org.jenkinsci.plugins.workflow.steps.GeneralNonBlockingStepExecution.lambda$run$0(GeneralNonBlockingStepExecution.java:77)
	at java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:511)
	at java.util.concurrent.FutureTask.run(FutureTask.java:266)
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1149)
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:624)
Finished: FAILURE
```

Meantime [pipeline steps](https://jenkins.confluent.io/job/Confluentinc%20Contributors/job/ksql/job/PR-6733/4/flowGraphTable/) are all green.

agavra:
That's really bizarre - I've never seen that before (the error you see on build 4). I think that one is a transient Jenkins issue. I took a look at the third build result: https://jenkins.confluent.io/job/Confluentinc%20Contributors/job/ksql/job/PR-6733/3/flowGraphTable/

Hubbitus:
Good point. Thanks. Now `mvn checkstyle:check` passes.

Did you consider add [`.editorconfig`](https://editorconfig.org/) file into the project?

Hubbitus:
Yeh. And noow Jenkins build link (https://jenkins.confluent.io/job/confluentinc-pr/job/ksql/job/PR-6733/7/display/redirect) lead to 404 error unfortunately.

agavra:
> Did you consider add .editorconfig file into the project?

We've standardized around checkstyle plugins across most of Confluent's products (see https://github.com/confluentinc/ksql/blob/master/CONTRIBUTING.md#code-style) though `.editorconfig` does seem like a neat lightweight solution. I'll definitely consider that if starting my own open source project!

ksqldb-serde/src/main/java/io/confluent/ksql/serde/avro/KsqlAvroSerdeFactory.java
```diff
@@ -140,6 +140,9 @@ private static AvroConverter getAvroConverter(
 
     avroConfig.put(AvroDataConfig.CONNECT_META_DATA_CONFIG, false);
 
+    avroConfig.put(KsqlConfig.KSQL_VALUE_SUBJECT_NAME_STRATEGY,
+            ksqlConfig.getString(KsqlConfig.KSQL_VALUE_SUBJECT_NAME_STRATEGY));
```
agavra, line 144: I finally had a chance to look into the contents of the PR. 

I'm worried that it's not as simple as this - we have quite a bit of code that relies on the `TopicNameStrategy`. For example, see usages of `KsqlConstants#getSRSubject`. 
- One such use case is that we want to subject to exist before any data exists in the topic (see https://github.com/confluentinc/ksql/issues/4219 and the crazy chain of issues that arose around that).
- Another is when we clean up a ksqlDB cluster, we look for all the subjects that we created and clean them up as well (you can see the code in `ClusterTerminator`) which also hardcodes the dependency on the TopicNameStrategy.

If we do manage to make all of the above generic (which may involve non-trivial amounts of work), I think we also need to explicitly allow-list the subject strategies that are supported. Some strategies (e.g. `RecordNameStrategy` don't play very well with ksqlDB's assumption that each topic only has one schema).

Let me know if this makes sense, but I think there's a non-trivial amount of work if we want this to work natively end-to-end (it probably even warrants a KLIP)

agavra (commented):
Thanks @Hubbitus - I appreciate the contribution. See my inline comment, I think there's a lot left to do for this patch.

ksqldb-serde/src/main/java/io/confluent/ksql/serde/avro/KsqlAvroSerdeFactory.java
Hubbitus: Yeeees. Basically, my patch is work (create tables and streams, read from topics), but yes, it is failed e.g. for statements create a table from external connector (Postgres in my case). I also debug a few hours of usages `KsqlConstants#getSRSubject`. And thank you for the explanation.

Meantime `KsqlConstants#getSRSubject` static. From the first point of view, we should not put there some logic like read configs (is it possible in such context?). On another side, if is public, it is API, so it is a good idea to deprecate it and support the strategy.

What is looks more reasonable to me is move such logic into `io.confluent.ksql.schema.ksql.inference.SchemaRegistryTopicSchemaSupplier`. Read configuration and in instantiation also create a configured strategy. What do you think?

Hubbitus: And then we may just refactor and delete `KsqlConstants#getSRSubject`

What you speak about possible `RecordNameStrategy` if it can't be used, we may just document it and give correct validation message.

Hubbitus: Could you please say or provide an example of what will be a correct way to get access to a parsed instance of `KsqlConfig` from within `SchemaRegistryTopicSchemaSupplier`?

Hubbitus: @agavra, could you please look at https://github.com/confluentinc/ksql/pull/6733/commits/1d4fe5f98c735bb13c1acc1c1f4063caa1f7c0fb commit?

I do not found a better way to inject configurated strategy into KsqlConstants and leave `static getSRSubject(...)` for backward compatibility.

Even if I introduce some DI mechanism and provide the bean like `subjectNameStrategy`, if I understand correctly we need to save public `getSRSubject`? Right?

agavra: Thanks @Hubbitus - I took another look. I think there's a good chunk of design work to be done before we can start thinking about the best approach (and it needs to be holistic, not just on the write-path). If we try to stick something in now, we'll be stuck with supporting it for the foreseeable future (and we have to make sure we don't break anything either). 

Is it blocking you to have ksql register things under the wrong subject name? You can always copy the schema across to be what you need it to be downstream.

Hubbitus: > I think there's a good chunk of design work to be done before we can start thinking about the best approach (and it needs to be holistic, not just on the write-path).
It is not write-path related! Furthermore, my first attempt which you comment, was the opposite, to support read data from topics as ksqldb streams and tables!

> If we try to stick something in now, we'll be stuck with supporting it for the foreseeable future (and we have to make sure we don't break anything either).
I agree I can miss something. It is my first contribution attempt. Maybe I need to write or extend some tests? Integration I think more interesting for that.

> Is it blocking you to have ksql register things under the wrong subject name? You can always copy the schema across to be what you need it to be downstream.
As I stated before ksqldb even can't read our topics! When I start to try to add source connectors, I was forced then configure schemes registration too. I had tried copy schemes under different names, on sandbox but it also near impossible when there several versions (or I do not know how).
So, really it is fully blocking us use ksqldb in our environment. Or use the patched version (that also is not a good alternative in the long-term perspective).

Hubbitus: @agavra, do you reject merging that pull request, or I can do something to move forward with it? I've ready to extend tests or adjust implementation.

agavra: Hello @Hubbitus - I don't think our team has enough bandwidth to help get this change across the line (as I mentioned, I think there's still a lot of work to be done). So I guess I'm "rejecting" it (though it's not because I don't think this feature is valuable, but I think it requires a lot more design from people familiar with all the implications).

Hubbitus: Maybe then you will be so kind as to plan it into some roadmap (milestone)?
And again, please say if I can doing something to help it moving forward.

agavra: cc @colinhicks for roadmapping purposes

PR #6182, opened 2020-09-10
fix: fail on non-string MAP keys

big-andy-coates:
### Description 

fixes: https://github.com/confluentinc/ksql/issues/6181

Fixes ksqlDB so that statements that create maps with non-string keys fail.

Crux of the issue was in `SqlTypeParser`, which ignored the key type.

Additionally, there were many places in the code that checked for non-STRING keys, but they always were STRING keys by this point.  These checks have generally been removed in prep for supporting non-STRING keys.  It is the format that should decide if it can handle non-STRING keys or not, hence the code now fails these statements because the key/value format rejects them as unsupported. This opens the way to PROTOBUF supporting integer keys in the future: only the Protobuf Serde could should need to change. (https://github.com/confluentinc/ksql/issues/6177).

Main changes:

- Fix to `SqlMap` and factory method `SqlTypes.map` to have the map type take a key type as well as a value type.
- Fix `SqlTypeParser` so that the parser does not ignore the key type.
- Fix `MapType` of the `ParamType` family of types in the same was as `SqlMap`.
- Fix `UdfUtil` to handle map keys
- Fix exception types in a few places.

Most of the rest is just test code change to take into account the new parameters.  Sorry, forgot to push as separate commits.

### Testing done 

Usual

### Reviewer checklist
- [ ] Ensure docs are updated if necessary. (eg. if a user visible feature is being added or changed).
- [ ] Ensure relevant issues are linked (description should include text like "Fixes #<issue number>")

PR #6735, opened 2020-12-08
docs: clarify that all data types except maps are supported as keys

vcrfxia:
### Description 

Docs for https://github.com/confluentinc/ksql/pull/6722. I don't see anywhere in the docs today where we say only primitive types are supported as keys. With 0.15, all non-map types will be supported, which this PR explicitly clarifies.

### Testing done 

Docs-only change.

### Reviewer checklist

### DIFF
--- a/docs/developer-guide/syntax-reference.md
+++ b/docs/developer-guide/syntax-reference.md
@@ -264,7 +264,8 @@ MAP<KeyType, ValueType>
 
 ksqlDB supports fields that are maps. A map has a key and value type. All
 of the keys must be of the same type, and all of the values must be also
-be of the same type.
+be of the same type. Currently only `STRING` keys are supported. The
+value type can be any valid SQL type.
 
 Access the values of a map by using the `[]` operator and passing in the
 key. For example, `SOME_MAP['cost']` retrieves the value for the entry
--- a/docs/developer-guide/syntax-reference.md
+++ b/docs/developer-guide/syntax-reference.md
@@ -255,12 +255,16 @@ MAP<KeyType, ValueType>
 ```
 
 !!! note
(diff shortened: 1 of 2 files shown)


docs/developer-guide/syntax-reference.md
```diff
@@ -254,13 +254,17 @@ SELECT ARRAY[s1.colA, s1.colB, s1.colC] FROM s1 EMIT CHANGES;
...
+    This restriction includes nested types containing maps as well.
+
 !!! note
 		The `DELIMITED` format doesn't support maps.
 
 ksqlDB supports fields that are maps. A map has a key and value type. All
 of the keys must be of the same type, and all of the values must be also
-be of the same type. Currently only `STRING` keys are supported. The
-value type can be any valid SQL type.
+be of the same type.
```
agavra, line 267: are you sure this is true? I thought there were some assumptions baked in at the ksql layer (not the serde layer) about string keys.

agavra (approved):
LGTM, but I think we need to verify the second claim.

docs/developer-guide/syntax-reference.md
vcrfxia: That's funny, the reason I made this change was because I remembered https://github.com/confluentinc/ksql/pull/6182 which removed all the assumptions that were previously baked in at non-serde layers. However. I forgot that the PR didn't actually add support for protobuf maps with non-string keys, even though we in theory could. So ultimately you're right ( for the wrong reason ;) ). I'll put the statement back in.

vcrfxia: Thanks for the catch!

JimGalasyn (approved):
Good tip!

vcrfxia:
Docs-only change. Merging despite build (flaky test failure).

PR #6737, opened 2020-12-08
chore: remove usage of SRClient#getAllSubjects

### LINKED ISSUE #6729: Only lookup one Schema Registry subject instead of all for CREATE STREAM|TABLE
With ksqlDB 0.8.0+, a schema is registered if needed during execution of `CREATE STREAM|TABLE`. Currently, the implementation looks up _all_ subjects in Schema Registry by sending a `GET /subjects` request. This can be problematic when using the `SimpleTopicAclAuthorizer` in Schema Registry, which requires superuser privileges for this operation.

Only lookup a single subject (based on the topic) when executing `CREATE STREAM|TABLE`, which does not require superuser privileges with `SimpleTopicAclAuthorizer`.

Use a different type of authorizer (such as RBAC) in Schema Registry.

agavra:
Fixes #6729 

### Description 
Don't use `getAllSubjects` when we can get look up just a single subject.

### Testing done 
- unit testing
- e2e testing to make sure the right exception is thrown with real SRClient

### Reviewer checklist

### DIFF
--- a/ksqldb-engine/src/main/java/io/confluent/ksql/schema/registry/SchemaRegistryUtil.java
+++ b/ksqldb-engine/src/main/java/io/confluent/ksql/schema/registry/SchemaRegistryUtil.java
@@ -22,7 +22,6 @@ import io.confluent.kafka.schemaregistry.client.rest.exceptions.RestClientExcept
 import io.confluent.ksql.util.ExecutorUtil;
 import io.confluent.ksql.util.KsqlConstants;
 import io.confluent.ksql.util.KsqlException;
-import java.io.IOException;
 import java.util.Optional;
 import java.util.stream.Stream;
 import org.slf4j.Logger;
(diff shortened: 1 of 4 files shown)


ksqldb-engine/src/main/java/io/confluent/ksql/schema/registry/SchemaRegistryUtil.java
```diff
@@ -89,6 +93,21 @@ public static void deleteSubjectWithRetries(
     }
   }
 
+  public static boolean subjectExists(
+      final SchemaRegistryClient srClient,
+      final String subject
+  ) {
+    try {
```
vcrfxia, line 100: Any reason we can't just call `getLatestSchema(...).isPresent()`?

vcrfxia (approved):
LGTM with a question inline. Thanks!

ksqldb-engine/src/main/java/io/confluent/ksql/schema/registry/SchemaRegistryUtil.java
agavra: that's a good idea, I have to do some refactoring to make it work but i'll just do that.

PR #6745, opened 2020-12-08
chore: support KEY_DELIMITER and tests for multi-col

agavra:
### Description 

Partial work toward #824 by supporting `KEY_DELIMITER` for delimited and adding some multi-col tests for delimited format.

### Testing done 

Added QTT tests

### Reviewer checklist

### DIFF
--- a/ksqldb-common/src/main/java/io/confluent/ksql/properties/with/CommonCreateConfigs.java
+++ b/ksqldb-common/src/main/java/io/confluent/ksql/properties/with/CommonCreateConfigs.java
@@ -143,7 +143,7 @@ public final class CommonCreateConfigs {
             "The delimiter to use when VALUE_FORMAT='DELIMITED'. Supports single "
               + "character to be a delimiter, defaults to ','. For space and tab delimited values "
               + "you must use the special values 'SPACE' or 'TAB', not an actual space or tab "
-              + "character.")
+              + "character. Also see " + KEY_DELIMITER_PROPERTY)
         .define(
             KEY_FORMAT_PROPERTY,
             ConfigDef.Type.STRING,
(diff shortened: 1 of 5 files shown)


ksqldb-common/src/main/java/io/confluent/ksql/properties/with/CommonCreateConfigs.java
```diff
@@ -123,6 +124,16 @@ public static void addToConfigDef(
...
+        .define(
+            KEY_DELIMITER_PROPERTY,
+            ConfigDef.Type.STRING,
+            null,
+            ConfigValidators.nullsAllowed(ConfigValidators.parses(Delimiter::parse)),
+            Importance.LOW,
+            "The delimiter to use when KEY_FORMAT='DELIMITED'. Supports single "
+              + "character to be a delimiter, defaults to ','. For space and tab delimited values "
+              + "you must use the special values 'SPACE' or 'TAB', not an actual space or tab "
+              + "character. Also see " + VALUE_DELIMITER_PROPERTY)
```
vcrfxia, line 136: Out of curiosity, is it standard (in our codebase, and in general) to cross-link between key and value properties like this? I haven't seen it done elsewhere, and this PR also doesn't add a link from `VALUE_DELIMITER_PROPERTY` to `KEY_DELIMITER_PROPERTY`. Curious what made you decide to link in this direction in this particular case.

vcrfxia (approved):
LGTM!

ksqldb-common/src/main/java/io/confluent/ksql/properties/with/CommonCreateConfigs.java
agavra: I just thought that with the format properties it might be a little confusing if you specify `FORMAT='DELIMITED'` and then have to specify individually the format properties (e.g. if you set just `KEY_DELIMITER` you might not get expected behavior). It could be overkill, but I'll just add the cross-link for the VALUE_FORMAT as well.

Issue #6734, opened 2020-12-08
Set and list against ksql.streams.topic.* prefixed properties is not working


spena:
**Describe the bug**
The properties prefixed with `ksql.streams.topic.*` are not listed with `LIST PROPERTIES` and cannot be set from CLI sessions either.

There is a confusion whether these properties should be listed or not. There is a fix to allow the streams topic prefix configs can be configured https://github.com/confluentinc/ksql/pull/3691 from the properties file. However, the `KsqlConfigResolver.java` has code with a difference on the topics to resolve and not.

These configs are resolved, which do not contain the topic prefix:
```
private static final List<PrefixedConfig> STREAM_CONFIG_DEFS = ImmutableList.of(
      new PrefixedConfig(StreamsConfig.CONSUMER_PREFIX, CONSUMER_CONFIG_DEF),
      new PrefixedConfig(StreamsConfig.PRODUCER_PREFIX, PRODUCER_CONFIG_DEF),
      new PrefixedConfig("", CONSUMER_CONFIG_DEF),
      new PrefixedConfig("", PRODUCER_CONFIG_DEF),
      new PrefixedConfig("", STREAMS_CONFIG_DEF)
  );
```
But in the `resolveStreamsConfig`, the topic prefix is added:
```
if (propertyName.startsWith(KSQL_STREAMS_PREFIX)
        && !propertyName.startsWith(KSQL_STREAMS_PREFIX + StreamsConfig.PRODUCER_PREFIX)
        && !propertyName.startsWith(KSQL_STREAMS_PREFIX + StreamsConfig.CONSUMER_PREFIX)
        && !propertyName.startsWith(KSQL_STREAMS_PREFIX + StreamsConfig.TOPIC_PREFIX)) {
      return Optional.empty();  // Unknown streams config
    }
```

Perhaps this is a bug, and we should add the topic prefix to the `STREAM_CONFIG_DEFS` as well.

**To Reproduce**
Use the latest master (6.2.0). Add the `ksql.streams.topic.min.insync.replicas` to the `ksql-server.properties`, and then run `list properties` from the CLI. The `ksql.streams.topic.min.insync.replicas` is not displayed.

**Expected behavior**
I'm not sure if these properties should be displayed or not, but looking at the code difference above, seems they should be displayed too.

agavra:
Looking at the `TopicConfig` class, I think it makes sense to list these properties. At least it seems valuable to me!

PR #6753, opened 2020-12-09
fix: include 'ksql.streams.topic.*' prefix properties on LIST PROPERTIES output

spena:
### Description 
Fixes https://github.com/confluentinc/ksql/issues/6734

Includes `ksql.streams.topic.*` prefixed properties into the `LIST PROPERTIES` output.
```
ksql> list properties;
...        
ksql.streams.topic.min.insync.replicas                 | KSQL  | SERVER           | 1        
...
```

It also allows users to set the `ksql.streams.topic.*` properties from the CLI.

### Testing done 
Added unit tests
Verified manually

### Reviewer checklist

### DIFF
--- a/ksqldb-common/src/main/java/io/confluent/ksql/util/KsqlConfig.java
+++ b/ksqldb-common/src/main/java/io/confluent/ksql/util/KsqlConfig.java
@@ -50,11 +50,15 @@ import org.apache.kafka.common.config.ConfigDef.Type;
 import org.apache.kafka.common.config.ConfigDef.ValidString;
 import org.apache.kafka.common.config.ConfigDef.Validator;
 import org.apache.kafka.common.config.SslConfigs;
+import org.apache.kafka.common.config.TopicConfig;
 import org.apache.kafka.common.utils.AppInfoParser;
 import org.apache.kafka.streams.StreamsConfig;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 @EffectivelyImmutable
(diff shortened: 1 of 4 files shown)


ksqldb-common/src/main/java/io/confluent/ksql/util/KsqlConfig.java
```diff
@@ -958,6 +963,25 @@ private KsqlConfig(final ConfigGeneration generation, final Map<?, ?> props) {
     this.ksqlStreamConfigProps = buildStreamingConfig(streamsConfigDefaults, originals());
   }
 
+  private static Set<String> streamTopicConfigNames() {
+    final ImmutableSet.Builder<String> configs = ImmutableSet.builder();
+
+    Arrays.stream(TopicConfig.class.getDeclaredFields())
```
agavra, line 969: heh.. this is a neat trick. I guess there's no other option to get these configs without reflection and just manually adding them?

agavra (approved):
LGTM!

ksqldb-common/src/main/java/io/confluent/ksql/util/KsqlConfig.java
spena: Right. I was thinking of manually creating the list, but TopicConfig can change in the future, and we'll have to adjust every time it changes. I'm not a fan of reflection, but I like the magic on it. This magic helps to forget about changes in TopicConfig as long as they stay public.

PR #6761, opened 2020-12-11
chore: only set unwrap_singles on internal topics if schema is single key

vcrfxia:
### Description 

### Testing done 

### Reviewer checklist

### DIFF
--- a/ksqldb-engine/src/main/java/io/confluent/ksql/structured/SchemaKTable.java
+++ b/ksqldb-engine/src/main/java/io/confluent/ksql/structured/SchemaKTable.java
@@ -49,6 +49,7 @@ import io.confluent.ksql.util.KsqlConfig;
 import io.confluent.ksql.util.KsqlException;
 import java.util.List;
 import java.util.Optional;
+import java.util.function.Function;
 
 // CHECKSTYLE_RULES.OFF: ClassDataAbstractionCoupling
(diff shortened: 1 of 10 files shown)


ksqldb-engine/src/main/java/io/confluent/ksql/structured/SchemaKTable.java
```diff
@@ -185,19 +186,28 @@ public SchemaKTable(
...
         .orElse(keyFormat);
 
-    final ExecutionStep<KTableHolder<K>> step = ExecutionStepFactory.tableSelectKey(
-        contextStacker,
-        sourceTableStep,
-        InternalFormats.of(newKeyFormat.getFormatInfo(), valueFormat),
-        keyExpression
-    );
+    final ExecutionStep<KTableHolder<K>> step = stepSupplier.apply(
+        InternalFormats.of(newSchema, newKeyFormat.getFormatInfo(), valueFormat));
```
vcrfxia, line 210: I'd like to unify this with the other methods by introducing an appropriate variant of `buildStepWithFormats()` but it'd be very custom since we also need access to `newKeyFormat` since it's needed in the SchemaKTable returned by the method. It's not clear to me that introducing such a custom variant is worth the unification.

vcrfxia:
Closing in favor of https://github.com/confluentinc/ksql/pull/6762.

PR #6762, opened 2020-12-11
chore: set internal key serde features when necessary

agavra:
### Description 

Pipe the key format that is determined from the repartition step into the select key, instead of creating it with empty serde features. This means that we will ensure that the serde features (wrapping/unwrapping) behave the way we want them to.

This is also a minor BREAKING CHANGE. Joins across different window types will no longer work. I think this is a good thing because the previous logic was somewhat insane and allowed you to join windows if they were semantically different but byte-equivalent.

### Testing done 

QTT testing/Unit Testing. I also hacked a QTT test to support wrapping single keys at the source level to see if it is properly resolved:
```json
    {
      "name": "stream-table key-to-key - key format properties mismatch",
      "statements": [
        "CREATE TABLE T (ID INT PRIMARY KEY, VAL STRING) WITH (kafka_topic='t', key_format='JSON', value_format='JSON', wrap_single_keys=true);",
        "CREATE STREAM S (ID INT KEY, FOO INT) WITH (kafka_topic='s', key_format='JSON', value_format='JSON');",
        "CREATE STREAM OUTPUT AS SELECT S.ID, VAL FROM S JOIN T ON S.ID = T.ID;"
      ],
      "inputs": [
        {"topic": "t", "key": {"ID": 10}, "value": {"VAL": "hello"}},
        {"topic": "s", "key": 10, "value": {"foo": 22}}
      ],
      "outputs": [
        {"topic": "OUTPUT", "key": 10, "value": {"VAL": "hello"}}
      ],
      "post": {
        "sources": [
          {"name": "OUTPUT", "type": "stream", "keyFormat": {"format": "JSON"}}
        ],
        "topics" : {
          "topics": [
            {
              "name": "t",
              "keyFormat": { "format": "JSON", "features": [ "WRAP_SINGLES" ] },
              "valueFormat": { "format": "JSON" },
              "partitions": 4
            },
            {
              "name": "_confluent-ksql-some.ksql.service.idquery_CSAS_OUTPUT_0-RightSourceKeyed-SelectKey-repartition",
              "keyFormat": { "format": "JSON", "features": [ "UNWRAP_SINGLES" ] },
              "valueFormat": { "format": "JSON" }
            },
            {
              "name": "OUTPUT",
              "keyFormat": { "format": "JSON", "features": [ "UNWRAP_SINGLES" ] },
              "valueFormat": { "format": "JSON" },
              "partitions": 4
            },
            {
              "name": "s",
              "keyFormat": { "format": "JSON", "features": [ "UNWRAP_SINGLES" ] },
              "valueFormat": { "format": "JSON" },
              "partitions": 4
            },
            {
              "name": "_confluent-ksql-some.ksql.service.idquery_CSAS_OUTPUT_0-RightSourceKeyed-Materialize-changelog",
              "keyFormat": { "format": "JSON", "features": [ "UNWRAP_SINGLES" ] },
              "valueFormat": { "format": "JSON" }
            }
          ]
        }
      }
    }
```

Here was the corresponding topology:
```
Topologies:
   Sub-topology: 0
      --> KTABLE-SOURCE-0000000002
      --> KTABLE-MAPVALUES-0000000003
      <-- KSTREAM-SOURCE-0000000001
    Processor: KTABLE-MAPVALUES-0000000003 (stores: [])
      --> KTABLE-TRANSFORMVALUES-0000000004
      <-- KTABLE-SOURCE-0000000002
    Processor: KTABLE-TRANSFORMVALUES-0000000004 (stores: [])
      --> KTABLE-TOSTREAM-0000000005
      <-- KTABLE-MAPVALUES-0000000003
    Processor: KTABLE-TOSTREAM-0000000005 (stores: [])
      --> RightSourceKeyed-SelectKey-Mapper
      <-- KTABLE-TRANSFORMVALUES-0000000004
    Processor: RightSourceKeyed-SelectKey-Mapper (stores: [])
      --> RightSourceKeyed-SelectKey-repartition-filter
      <-- KTABLE-TOSTREAM-0000000005
    Processor: RightSourceKeyed-SelectKey-repartition-filter (stores: [])
      --> RightSourceKeyed-SelectKey-repartition-sink
      <-- RightSourceKeyed-SelectKey-Mapper
    Sink: RightSourceKeyed-SelectKey-repartition-sink (topic: RightSourceKeyed-SelectKey-repartition)
      <-- RightSourceKeyed-SelectKey-repartition-filter

  Sub-topology: 1
    Source: KSTREAM-SOURCE-0000000012 (topics: [s])
      --> KSTREAM-TRANSFORMVALUES-0000000013
    Processor: KSTREAM-TRANSFORMVALUES-0000000013 (stores: [])
      --> PrependAliasLeft
      <-- KSTREAM-SOURCE-0000000012
    Processor: PrependAliasLeft (stores: [])
      --> Join
      <-- KSTREAM-TRANSFORMVALUES-0000000013
    Processor: Join (stores: [RightSourceKeyed-Materialize])
      --> Project
      <-- PrependAliasLeft
    Source: RightSourceKeyed-SelectKey-repartition-source (topics: [RightSourceKeyed-SelectKey-repartition])
      --> RightSourceKeyed-SelectKey
    Processor: Project (stores: [])
      --> KSTREAM-SINK-0000000017
      <-- Join
    Processor: RightSourceKeyed-SelectKey (stores: [RightSourceKeyed-Materialize])
      --> PrependAliasRight
      <-- RightSourceKeyed-SelectKey-repartition-source
    Sink: KSTREAM-SINK-0000000017 (topic: OUTPUT)
      <-- Project
    Processor: PrependAliasRight (stores: [])
      --> none
      <-- RightSourceKeyed-SelectKey
```

### Reviewer checklist


ksqldb-engine/src/main/java/io/confluent/ksql/serde/InternalFormats.java
```diff
@@ -50,16 +50,16 @@ private InternalFormats() {
...
+    if (keyFormat.getFormatInfo().getFormat().equals(NoneFormat.NAME)) {
       throw new IllegalArgumentException(NoneFormat.NAME + " can not be used for internal topics");
     }
 
     return Formats.of(
-        keyFormat,
+        keyFormat.getFormatInfo(),
         valueFormat,
-        SerdeFeaturesFactory.buildInternal(FormatFactory.fromName(keyFormat.getFormat())),
+        keyFormat.getFeatures(),
```
vcrfxia, line 62: It's odd to me that we're wiping value features on internal topics while passing along key features. I need to think more about the implications here. We might need to write key serde features into all the execution steps if we want this to be evolvable in the future.

ksqldb-engine/src/main/java/io/confluent/ksql/structured/SchemaKTable.java
```diff
@@ -224,12 +216,12 @@ public SchemaKGroupedTable groupBy(
       final Stacker contextStacker
   ) {
     final KeyFormat groupedKeyFormat = KeyFormat
-        .nonWindowed(keyFormat.getFormatInfo(), SerdeFeatures.of());
+        .nonWindowed(keyFormat.getFormatInfo(), keyFormat.getFeatures());
```
vcrfxia, line 219: This change is strange to me -- what was your motivation? This will have to be updated to support multi-column group by's, since the new schema (post group by) needs to be taken into account when deciding what key serde features to set. (I can do that as part of my group by PR, just curious why this change is being made here when it'll have to change again soon anyway.)

I also notice that the analogous change wasn't made in SchemaKStream.

ksqldb-functional-tests/src/test/resources/query-validation-tests/joins.json
```diff
@@ -1907,37 +1907,19 @@
       }
     },
     {
-      "name": "matching time-windowed",
+      "name": "matching time-windowed join with different windows fails",
```
vcrfxia, line 1910: Why does this test fail with this change? I don't see the connection to key serde features.

vcrfxia (commented):
Thanks for the quick PR! I need more time to think about this. Some initial questions inline.

ksqldb-functional-tests/src/test/resources/query-validation-tests/joins.json
agavra: because previously we were checking just that the `FormatInfo` matched, which didn't include window information. `KeyFormat` does

ksqldb-engine/src/main/java/io/confluent/ksql/serde/InternalFormats.java
vcrfxia: Took a deeper look -- looks like we already write serde features into the execution steps that need them, so that shouldn't be a concern. 

The only potential evolvability concern is that until now, our execution plans have always specified key wrapping in ambiguous situations (single key column with format that supports both wrapping and unwrapping). That changes with this PR, which means we'd be locking ourselves into the current default choices for each format -- namely, always wrap by default. Are we happy with this default, or should we consider defaulting key wrapping to unwrap even though value wrapping default wrapping? If most users unwrap single keys, we may prefer the former despite the inconsistency.

ksqldb-engine/src/main/java/io/confluent/ksql/structured/SchemaKTable.java
vcrfxia: I think I answered my own question: there is no analogous change in SchemaKStream since stream group bys don't need key format. Defaulting to propagating key serde features in table group bys means fewer historic execution plans change, since internal topics today always unwrap single keys (which is what gets propagated with this change, else we'd get new plans with empty serde features after table group bys).

I don't think this inconsistency between stream and table group bys makes sense in the long run, but the logic here will change with the next PR anyway so I think it's fine.

ksqldb-functional-tests/src/test/resources/query-validation-tests/joins.json
vcrfxia: Makes sense. We should call it out as a breaking change, though.

ksqldb-functional-tests/src/test/resources/historical_plans/group-by_-_AVRO_group_by_struct/6.2.0_1607703224334/plan.json
```diff
@@ -0,0 +1,204 @@
+{
```
vcrfxia, line 1: Adding a note for posterity: all the historic plan changes here appear to be the removal of UNWRAP_SINGLES from stream select key steps, which happens when grouping by a different key or when streams are repartitioned for joins. This makes sense in light of the changes in this PR.

vcrfxia (commented):
Had a deeper look and answered a bunch of my own questions. This approach makes sense to me, still need to weigh tradeoffs against other approaches. Good stuff!

ksqldb-engine/src/main/java/io/confluent/ksql/serde/InternalFormats.java
agavra: > What changes with this PR, which means we'd be locking ourselves into the current default choices for each format -- namely, always wrap by default.

Why does it mean that? If we write the serde features into the execution steps then we can always change the logical plan to produce different execution steps while keeping all of the old ones compatible. 

> If most users unwrap single keys, we may prefer the former despite the inconsistency.

Users shouldn't see what happens underneath the hood, we still "inherit" the desired serde features to the sink independent of what happens in the internal topics. You can see the example I've given in the "testing done" where the table is wrapped and the stream is unwrapped. The "final" chooses the left topic's information, which is to `UNWRAP_SINGLES`

agavra: > It's odd to me that we're wiping value features on internal topics while passing along key features. 

I agree, it felt a little sticky to me as well. I think it's possible to pass along the value format as well, but it's not as interesting (it would be mostly for hygiene) because we can always serialize it into the desired format at the edges (our semantics don't depend on the value format).

If we do this, we can do it as a follow-up.

ksqldb-engine/src/main/java/io/confluent/ksql/structured/SchemaKTable.java
agavra: Yeah, I assumed that with multi-col group bys you'd need to generate a new key format and probably change the key features. But what's nice with this approach is that whatever mechanism you decide should be used for generating the key format will automagically be propagated down the line.

ksqldb-functional-tests/src/test/resources/historical_plans/group-by_-_AVRO_group_by_struct/6.2.0_1607703224334/plan.json
agavra: Thanks for checking this! that was my next step actually to make sure it works 😂

ksqldb-serde/src/main/java/io/confluent/ksql/serde/KeyFormat.java
```diff
@@ -36,7 +37,7 @@
 public final class KeyFormat {
 
   private final FormatInfo format;
-  private final SerdeFeatures features;
+  private SerdeFeatures features;
```
vcrfxia, line 40: I thought this class was marked Immutable?

ksqldb-engine/src/main/java/io/confluent/ksql/structured/SchemaKStream.java
```diff
@@ -364,37 +358,33 @@ public SchemaKGroupedStream groupBy(
...
     }
 
-    final FormatInfo keyFmtInfo = keyFormat.getFormatInfo().getFormat().equals(NoneFormat.NAME)
-        ? FormatInfo.of(ksqlConfig.getString(KsqlConfig.KSQL_DEFAULT_KEY_FORMAT_CONFIG))
-        : keyFormat.getFormatInfo();
-
-    final KeyFormat rekeyedKeyFormat = KeyFormat
-        .nonWindowed(keyFmtInfo, SerdeFeatures.of());
-
+    final boolean isSingleKey = groupByExpressions.size() == 1;
```
vcrfxia, line 364: If this is removed and we instead hard-code `isSingleKey=true` for now, as we have been for purposes of setting internal key features, and is still the case in this PR for table group bys, do the changes in QTT historic plans go away? If so, I prefer this. I'll re-introduce this change in my multli-column group by PR where plans will change anyway.

ksqldb-engine/src/test/java/io/confluent/ksql/structured/SchemaKGroupedStreamTest.java
```diff
@@ -162,9 +163,10 @@ public void shouldBuildStepForAggregate() {
   }
 
   @Test
-  public void shouldBuildStepForAggregateWhereKeyFormatSupportsBothWrappingAndUnwrapping() {
+  public void shouldBuildStepForAggregateWithKeyFormatSerdeFeatures() {
```
vcrfxia, line 166: We're testing two things, right?
- If key wrapping is present, then the key serde feature is inherited
- If key wrapping is not present, then it is populated

Seems like it should be two separate unit tests? Same for the other test file below.

ksqldb-engine/src/main/java/io/confluent/ksql/planner/plan/JoinNode.java
```diff
@@ -254,16 +255,17 @@ private void ensureMatchingPartitionCounts(final KafkaTopicClient kafkaTopicClie
...
         .map(DataSource::getKsqlTopic)
         .map(KsqlTopic::getKeyFormat)
-        .map(KeyFormat::getFormatInfo)
-        .filter(format -> !format.getFormat().equals(NoneFormat.NAME))
+        .filter(format -> !format.getFormatInfo().getFormat().equals(NoneFormat.NAME))
         .findFirst()
-        .orElse(FormatInfo.of(defaultKeyFormat));
+        // if none exist, assume non-Windowed since that would mean that both sources
+        // were of NONE format, which doesn't support windowed operations
+        .orElse(KeyFormat.nonWindowed(FormatInfo.of(defaultKeyFormat), SerdeFeatures.of()));
```
vcrfxia, line 268: Upon first reading I was confused why this wasn't being sanitized with `SerdeFeaturesFactory#sanitizeKeyFeatures()`. Looks like the reason is all the sanitization happens in SchemaKStream and SchemaKTable instead. Might be worth adding that into the javadocs for the method?

ksqldb-engine/src/main/java/io/confluent/ksql/serde/SerdeFeaturesFactory.java
```diff
@@ -87,6 +88,57 @@ public static SerdeFeatures buildValueFeatures(
     return SerdeFeatures.from(builder.build());
   }
 
+  /**
+   * Not all {@code KeyFormat}s are valid internal topic formats. Specifically,
+   * we want to ensure that the key format is (1) not NONE and (2) explicitly sets
+   * the wrapping if it contains only a single column. This method ensures that
+   * both of these are eagerly set.
+   *
+   * @return the key format to use for the output of the aggregation
```
vcrfxia, line 97: This isn't just for aggregations, right? I'm confused by this javadoc.

ksqldb-engine/src/main/java/io/confluent/ksql/serde/SerdeFeaturesFactory.java
```diff
@@ -87,6 +88,57 @@ public static SerdeFeatures buildValueFeatures(
...
+   */
+  public static KeyFormat sanitizeKeyFormat(
+      final KsqlConfig ksqlConfig,
+      final KeyFormat keyFormat,
+      final boolean isSingleKey
+  ) {
+    if (keyFormat.getFormatInfo().getFormat().equals(NoneFormat.NAME)) {
+      // explicitly resolve the current default wrapping and write it into the
```
vcrfxia, line 105: I'm confused. Why are we populating a key format in the case where the key format is NONE? This seems like it changes behavior of things like `PARTITION BY NULL` where we in fact want the resulting key format to be NONE...?

ksqldb-engine/src/main/java/io/confluent/ksql/serde/SerdeFeaturesFactory.java
```diff
@@ -87,6 +88,57 @@ public static SerdeFeatures buildValueFeatures(
...
+      final SerdeFeatures defaultWrapping = getKeyWrapping(isSingleKey, defaultFormat.getFormat())
+          .map(SerdeFeatures::of)
+          .orElse(SerdeFeatures.of());
+
+      return KeyFormat.nonWindowed(defaultFormat, defaultWrapping);
+    }
+
+    // it is possible that the source format was either multi-key
+    // or no-key, in which case there would not have been specified
+    // a wrapping configuration - we should specify one here
```
vcrfxia, line 119: nit:
```suggestion
    // or no-key, in which case there would not have been
    // a wrapping configuration specified - we should specify one here
```

ksqldb-engine/src/main/java/io/confluent/ksql/serde/SerdeFeaturesFactory.java
```diff
@@ -87,6 +88,57 @@ public static SerdeFeatures buildValueFeatures(
...
+    } else {
+      return keyFormat;
+    }
+  }
+
+  private static Optional<SerdeFeature> getKeyWrapping(
+      final boolean singleKey,
+      final String format
+  ) {
+    return getKeyWrapping(singleKey, FormatFactory.fromName(format));
```
vcrfxia, line 139: nit: any particular reason we extract the format name from FormatInfo and then call `FormatFacory.fromName()`, rather than directly calling `FormatFactory.of()`?

ksqldb-functional-tests/src/test/resources/query-validation-tests/joins.json
vcrfxia: Thanks for adding the breaking change description to the PR! Friendly reminder to copy it into the commit message when this gets merged, in the format `BREAKING CHANGE: <text here>` so the auto-changelog-generator picks it up :)

vcrfxia (commented):
Awesome! LGTM overall -- some questions inline.

ksqldb-serde/src/main/java/io/confluent/ksql/serde/KeyFormat.java
agavra: um... yes it should be. I'm not sure why this changed (it might've been an accidental automated refactor)

ksqldb-engine/src/main/java/io/confluent/ksql/structured/SchemaKStream.java
agavra: sounds good, I can do that

ksqldb-engine/src/main/java/io/confluent/ksql/serde/SerdeFeaturesFactory.java
agavra: yes you are right. I wrote the javadoc and then I did some refactoring and it no longer applies! good catch

ksqldb-engine/src/main/java/io/confluent/ksql/serde/SerdeFeaturesFactory.java
agavra: hmm, let me look into this. I think it might be the same issue as above (I first used this just for aggregates and then I made it fit to everything). I'm not sure why this doesn't cause things to fail.

ksqldb-engine/src/main/java/io/confluent/ksql/serde/SerdeFeaturesFactory.java
agavra: changed, it was leftover from a refactor where I had a little bit more complicated logic

ksqldb-engine/src/main/java/io/confluent/ksql/serde/SerdeFeaturesFactory.java
agavra: I think it was working because we were always setting the right format in the output anyway and there wasn't any repartition step. I've changed the logic to only "fix" the format on group bys

vcrfxia (approved):
LGTM! Thanks @agavra , awesome stuff.

Issue #6360, opened 2020-10-05
ksqlDB leaks state stores for transient queries


agavra:
On a long-running cluster, I noted that there are many state stores on disk for transient queries:

```
root@ksql-1:/mnt/data/data/ksql-state# ls -al | grep transient | grep -oE 'Sep [0-9]+' | sort | uniq -c
     76 Sep 23
     91 Sep 24
    100 Sep 25
     96 Sep 26
    102 Sep 27
     88 Sep 28
     89 Sep 29
     11 Sep 30
root@ksql-1:/mnt/data/data/ksql-state# ls -al | grep transient | grep -E 'Sep [0-9]+' | head -5
4096 Sep 29 16:29 _confluent-ksql-pksqlc-gn6rmtransient_1012238981171586952_1601396723718
4096 Sep 28 03:38 _confluent-ksql-pksqlc-gn6rmtransient_1012260790516886546_1601264083968
4096 Sep 26 13:13 _confluent-ksql-pksqlc-gn6rmtransient_101964711127807535_1601125759174
4096 Sep 24 17:48 _confluent-ksql-pksqlc-gn6rmtransient_1020128588393177207_1600969460743
4096 Sep 30 00:33 _confluent-ksql-pksqlc-gn6rmtransient_1025443061838213674_1601425747128
```

These state stores all appear to have the same size (mostly), so I suspect they're for failed queries:
```
root@ksql-1:/mnt/data/data/ksql-state# du -sh * | grep transient | grep -oE '^[0-9.]+K' | sort | uniq -c
      4 164K
    640 4.0K
```

Furthermore, I noticed that the Kafka cluster was also overwhelmed with too many partitions for internal topics that were not cleaned up - all of them belonged to the transient queries above. The ksqlDB server was degraded when these queries were initially issued.

rodesai:
I had a look at our query-handling code and nothing obvious stood out to me that would explain this leak. We do have lots of logs around this codepath that would help, but they have been expired from the log store. So let's keep an eye out for a recurrence and debug when there are more logs.

rodesai:
Did a quick sanity-check here. It looks like we're leaking just the state store directory every time a transient query is terminated. The actual state store (directory contents) and internal topics are removed.

rodesai:
Looks like in some cases we _do_ leak internal topics and state stores. I think there's multiple issues going on here:

[...]

PR #6767, opened 2020-12-14
fix: closing a rebalancing transient query may leak the state store directory

spena:
### Description 
Fixes https://github.com/confluentinc/ksql/issues/6360

If a small timeout is set during the transient query close and the query is still rebalancing after the timeout expires, then the stream thread re-creates the internal topics and state stores causing the leak.

This PR derives the query timeout from the internal stream threads timeouts instead of using a configured timeout.

### Testing done 
Unit tests

### Reviewer checklist

### DIFF
--- a/ksqldb-engine/src/main/java/io/confluent/ksql/query/QueryExecutor.java
+++ b/ksqldb-engine/src/main/java/io/confluent/ksql/query/QueryExecutor.java
@@ -83,7 +83,7 @@ public final class QueryExecutor {
 
   // Each stream thread has two internal consumer threads running. When closing those consumer
   // threads, the close has a hard-coded 30 seconds timeout per thread.
-  private static long INTERNAL_CONSUMER_THREADS_TIMEOUT = 2 * 30000;
+  private static long INTERNAL_CONSUMER_THREADS_TIMEOUT_MS = 2 * 30_000;
 
   // CHECKSTYLE_RULES.ON: ClassDataAbstractionCoupling
   private final SessionConfig config;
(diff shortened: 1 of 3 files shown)


agavra (commented):
Where does this timeout get hit? Is it something that is user visible? I'm trying to understand the implications of having set this too high.

Otherwise, LGTM though we may want a streams expert like @guozhangwang take a look at this

spena:
@agavra The user does not have a way to set those internal timeouts. All internal closing methods have hard-coded values. The implication with a high value is the user may wait too much time until the query is closed. But setting the timeout too low also have the implication of leaking the state store if the stream was rebalancing. 

@rodesai What do you think about this value?

agavra:
> The implication with a high value is the user may wait too much time until the query is closed. 

Is this on the CLI path? e.g. if I ctrl-c a query does it hang for 5 min? Where does the user "see" this, or is it in the background?

rodesai:
> Is this on the CLI path? e.g. if I ctrl-c a query does it hang for 5 min? Where does the user "see" this, or is it in the background?

This should only come into play if the query is stuck for some reason (e.g. in a rebalance). On a normal close it should be quick. That said, the default timeout here is 5 minutes. So this patch would actually make us wait less. @spena I think based on your analysis in the ticket, we can't actually bound the wait time here, since the producer and admin clients wait indefinitely. Since this is the case, perhaps the right approach is to instead do a relatively short wait inline, and then if that times out, move the work to a background task that keeps trying to close the query. wdyt?

spena:
@rodesai Isn't the current close() called in a background thread already?

I looked at the `QueryStreamWriter`. If the client terminates the connection (stopped the transient query), then the writer will close the query. This writer, If I understand, runs in the background, isn't it?

guozhangwang:
just curious if this PR is still in progress?

spena:
@guozhangwang It needs more discussion I think. The PR does what it is expected, but it can take long to close if the timeout is large. Currently, ksql closes a transient query in the background if I understand, so I'm not sure about this:
```
... Since this is the case, perhaps the right approach is to instead do a relatively short wait inline, 
and then if that times out, move the work to a background task that keeps trying to close the query. wdyt?
```

I was wondering if by background approach means: all query metrics and metadata should be cleaned after a short period of time, then put the close() in the background to wait until it stream closes it, thus avoiding re-creating the state store directory? @rodesai is that a good understanding?

Issue #6371, opened 2020-10-07
Don't require AS_VALUE to access multi-column group by columns


agavra:
ksqlDB should allow access to the group-by columns without using `AS_VALUE` to manually copy them into the value.

Current Behavior:
```sql
ksql> CREATE STREAM a (id int key, col1 int) with (kafka_topic='a', value_format='json', partitions=1);
ksql> CREATE TABLE gb AS SELECT id, col1, COUNT(*) FROM a GROUP BY id, col1;

 Message
----------------------------------
 Created query with ID CTAS_GB_15
----------------------------------
ksql> SELECT id FROM gb;
Line: 1, Col: 8: SELECT column 'ID' cannot be resolved.
```

There are various ways to implement this:
1. we can actively copy any selected group by component into the value
1. we can leverage structured keys (#824) and decompose the key from the struct

I think the former is a good enough solution in the short term (which is how we implemented it before), while we should strive to implement the latter. The latter may require a refactor (TODO: link ticket here).

PR #6770, opened 2020-12-15
chore: support multi-column keys in GROUP BY

vcrfxia:
### Description 

Fixes https://github.com/confluentinc/ksql/issues/6371.

This PR adds support for multi-column GROUP BY. With this change, GROUP BY multiple columns will result in multiple key columns, rather than a single key column that is the string concatenation of the individual key fields (with delimiter `|+|`. A consequence of this change is that GROUP BY multiple columns is no longer allowed when using key formats that do not support it, such as the KAFKA format. 

This is a breaking change, and is currently guarded by a feature flag until we've added the corresponding pull query support and are ready to release. 

This PR adds test coverage in a new QTT file `group-by-multi-col.json` which is a copy-paste of the existing `group-by.json` with some additional tests added. Once the feature flag is removed, the current `group-by.json` will be deleted in favor of `group-by-multi-col.json` (which will be renamed to simply `group-by.json`). The easiest way to review the QTT changes, to see the breaking changes in this PR in addition to the next test coverage, is to diff the existing `group-by.json` with the new `group-by-multi-col.json`.

### Testing done 

Unit + QTT.

### Reviewer checklist

### DIFF
--- a/ksqldb-engine/src/main/java/io/confluent/ksql/engine/EngineExecutor.java
+++ b/ksqldb-engine/src/main/java/io/confluent/ksql/engine/EngineExecutor.java
@@ -478,7 +478,7 @@ final class EngineExecutor {
   }
 
   private String buildPlanSummary(final QueryId queryId, final ExecutionStep<?> plan) {
-    return new PlanSummary(queryId, config.getConfig(false), engineContext.getMetaStore())
+    return new PlanSummary(queryId, config.getConfig(true), engineContext.getMetaStore())
         .summarize(plan);
   }
 }
--- a/ksqldb-engine/src/main/java/io/confluent/ksql/structured/SchemaKGroupedStream.java
+++ b/ksqldb-engine/src/main/java/io/confluent/ksql/structured/SchemaKGroupedStream.java
@@ -85,7 +85,7 @@ public class SchemaKGroupedStream {
     } else {
       keyFormat = SerdeFeaturesFactory.sanitizeKeyFormat(
           this.keyFormat,
-          true
+          schema.key().size() == 1
       );
       step = ExecutionStepFactory.streamAggregate(
           contextStacker,
@@ -111,7 +111,7 @@ public class SchemaKGroupedStream {
             keyFormat.getFormatInfo(),
             keyFormat.getFeatures(),
             windowExpression.getKsqlWindowExpression().getWindowInfo()),
-        true
+        schema.key().size() == 1
     );
   }
 
--- a/ksqldb-streams/src/main/java/io/confluent/ksql/execution/streams/GroupByParamsFactory.java
+++ b/ksqldb-streams/src/main/java/io/confluent/ksql/execution/streams/GroupByParamsFactory.java
@@ -27,13 +27,16 @@ import io.confluent.ksql.logging.processing.NoopProcessingLogContext;
 import io.confluent.ksql.logging.processing.ProcessingLogger;
 import io.confluent.ksql.logging.processing.RecordProcessingError;
 import io.confluent.ksql.name.ColumnName;
+import io.confluent.ksql.schema.ksql.ColumnAliasGenerator;
 import io.confluent.ksql.schema.ksql.ColumnNames;
 import io.confluent.ksql.schema.ksql.LogicalSchema;
 import io.confluent.ksql.schema.ksql.types.SqlType;
 import io.confluent.ksql.schema.ksql.types.SqlTypes;
+import io.confluent.ksql.util.KsqlConfig;
 import java.util.List;
 import java.util.Objects;
 import java.util.function.Supplier;
+import java.util.stream.Stream;
 
 final class GroupByParamsFactory {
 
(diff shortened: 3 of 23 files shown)


ksqldb-engine/src/main/java/io/confluent/ksql/engine/EngineExecutor.java
```diff
@@ -478,7 +478,7 @@ private PersistentQueryMetadata executePersistentQuery(
   }
 
   private String buildPlanSummary(final QueryId queryId, final ExecutionStep<?> plan) {
-    return new PlanSummary(queryId, config.getConfig(false), engineContext.getMetaStore())
+    return new PlanSummary(queryId, config.getConfig(true), engineContext.getMetaStore())
```
vcrfxia, line 481: Is there a reason we weren't applying config overrides here? Seems like a bug. This fix is needed in order to pass the feature flag into the GroupByParamsBuilder.

ksqldb-engine/src/main/java/io/confluent/ksql/structured/SchemaKGroupedStream.java
```diff
@@ -85,7 +85,7 @@
     } else {
       keyFormat = SerdeFeaturesFactory.sanitizeKeyFormat(
           this.keyFormat,
-          true
+          schema.key().size() == 1
```
vcrfxia, line 88: This fix is needed, else `sanitizeKeyFormat()` adds in key wrapping after `groupBy()` removes it in the case of multi-column keys.

ksqldb-streams/src/test/java/io/confluent/ksql/execution/streams/GroupByParamsFactoryTest.java
```diff
@@ -238,37 +341,84 @@ public void shouldLogProcessingErrorIfAnyMultiExpressionResolvesToNull() {
   }
 
   @Test
-  public void shouldUseNullInGroupByIfOneExpressionFailsOrReturnsNullInMulti() {
```
vcrfxia, line 241: This test is identical to the existing `shouldReturnNullIfAnyMultiExpressionResolvesToNull()` and is therefore redundant.

ksqldb-engine/src/main/java/io/confluent/ksql/engine/EngineExecutor.java
agavra: off the top of my head, I don't think there's any good reason. We should make sure that it's not being used in the pull query path as this (calling `getConfig(true)`) is shockingly expensive at high throughput.

ksqldb-streams/src/main/java/io/confluent/ksql/execution/streams/GroupByParamsFactory.java
```diff
@@ -219,4 +229,69 @@ private static LogicalSchema multiExpressionSchema(
...
+        }
+
+        builder.append(result);
+      }
+
+      return builder.build();
+    }
+  }
+
+  private static LogicalSchema expressionSchema(
```
agavra, line 275: not your original code, but I'm wondering why we can't leverage the schema that we built in `LogicalPlanner`. It seems like these two schemas have to align perfectly or else this won't work (and the code is quite similar, this just handles `ExpressionMetadata` instead of `Expression` directly). Am I right? I haven't actually dug into it that much.

agavra (approved):
🎉  this is _supercalifragilisticexpialidocious_. Finally fixing one of the biggest historical pain points in ksqlDB (don't forget to add fixes #6371)

ksqldb-engine/src/main/java/io/confluent/ksql/engine/EngineExecutor.java
vcrfxia: This helper method is only called for persistent and push queries, not pull queries, so we should be in the clear. Good call on checking!

ksqldb-streams/src/main/java/io/confluent/ksql/execution/streams/GroupByParamsFactory.java
vcrfxia: Are you suggesting that the code for how we build the schema in the two places should be shared, or that the actual schema that gets built should be passed from `LogicalPlanner` to `GroupByParamsFactory`? I haven't looked too closely at the details of how the schema is built in `LogicalPlanner` but the latter doesn't make conceptual sense as the two are separated by the execution plan that gets serialized into the command topic.

agavra: I did mean the latter. Doesn't the serialized execution plan contain the schema (that was generated by the logical planner) too, though?

vcrfxia: We don't write the full schema in, no. Here's an example historic plan: https://github.com/confluentinc/ksql/blob/db2d01963930d2fa45f26670ecd6734fbf9f79e8/ksqldb-functional-tests/src/test/resources/historical_plans/group-by_-_AVRO_primitive_key/6.2.0_1607109704822/plan.json#L109 For the group by step we only write the group by columns, which is why we have to reconstruct the rest of the schema.

agavra: ack, out of scope then but seems like something that might be worth looking into some day

PR #6785, opened 2020-12-16
chore: fix multi-column GROUP BY support with new execution steps

vcrfxia:
### Description 

Follow-up to https://github.com/confluentinc/ksql/pull/6770 -- turns out the approach in that PR doesn't preserve backwards compatibility for existing multi-column GROUP BY queries since the logic for determining whether to concatenate grouping expressions into a single key column or not was contained purely in the engine, not in the execution plan itself. This PR fixes the issue by introducing new versions of the `StreamGroupBy` and `TableGroupBy` execution steps instead.

Note that nothing is currently broken for users since the new multi-column group by logic is currently behind a feature flag. The presence of this feature flag is also the reason no historic QTTs failed on the previous PR.

For reviewers: the raw number of lines changed in this PR is high but each of the changes are straightforward. Four commits for ease of reviewing:
- split `GroupByParamsFactory` into `GroupByParamsV1Factory` (old behavior, i.e., single, concatenated string key) and `GroupByParamsFactory` (new behavior, i.e., multi-column keys)
- introduce new execution steps
- introduce `StreamGroupByBuilder` and `StreamGroupByBuilderV1`, with the shared logic in `StreamGroupByBuilderBase`, and similarly for the table group by builders
- last few changes to finish wiring up the new classes

### Testing done 

Verified historic QTT pass once the feature flag is removed (tested locally).

### Reviewer checklist


ksqldb-execution/src/main/java/io/confluent/ksql/execution/plan/ExecutionStep.java
```diff
@@ -31,7 +31,8 @@
     @Type(value = StreamAggregate.class, name = "streamAggregateV1"),
     @Type(value = StreamFilter.class, name = "streamFilterV1"),
     @Type(value = StreamFlatMap.class, name = "streamFlatMapV1"),
-    @Type(value = StreamGroupBy.class, name = "streamGroupByV1"),
+    @Type(value = StreamGroupByV1.class, name = "streamGroupByV1"),
+    @Type(value = StreamGroupBy.class, name = "streamGroupByV2"),
```
agavra, line 35: to avoid confusion that I have with `StreamSelectKey` every time i look at the class, what do you think about (in a follow-up PR) changing the class name of `StreamGroupBy` to `StreamGroupByV2` and `StreamSelectKey` to `StreamSelectKeyV2`

agavra (approved):
LGTM! Great plumbing :)

ksqldb-execution/src/main/java/io/confluent/ksql/execution/plan/ExecutionStep.java
vcrfxia: I actually rather like the convention that the non-versioned class is always the latest. It makes it easier to find the latest code rather than needing to first check/reason about what the latest version is. 

Would you want to rename just the execution step classes or also the affiliated builders and execution plan traversal visitor methods? If the confusion is just around the execution step classes themselves, I could be on-board for that rename. I'm not sure renaming the others saves more confusion than is worth the toil it creates but I don't feel too strongly.

agavra: fair enough, I don't feel _too_ strongly and I guess I can see it either way

PR #6786, opened 2020-12-16
feat: ungate support for multi-column GROUP BY

vcrfxia:
### Description 

Exposes support for multi-column GROUP BY from behind the feature flag, and also adds docs. If we merge this PR now, we can catch 0.15. Ideally we'd wait until pull query support for multiple key columns has been added, though. 

BREAKING CHANGE: Queries with `GROUP BY` clauses that contain multiple grouping expressions now result in multiple key columns, one for each grouping expression, rather than a single key column that is the string-concatenation of the grouping expressions. Note that this new behavior (and breaking change) apply only to new queries; existing queries will continue to run uninterrupted with the previous behavior, even across ksqlDB server upgrades.

Three commits:
- docs updates
- remove feature flag
- QTT historic plans

### Testing done 

QTT.

### Reviewer checklist


agavra (approved):
So many back-to-back big wins 😂 excited to see this one go in. I didn't look into the historical plans in depth since I trust the change

JimGalasyn (approved):
Very cool!

vcrfxia:
Pull query support (for equality conditions) is added in https://github.com/confluentinc/ksql/pull/6796, and will also be merged in time for 0.15. Merging now 🎉

PR #6790, opened 2020-12-16
chore: improve error messages on invalid GROUP BYs

vcrfxia:
### Description 

tl;dr this PR improves error messages on invalid statements involving GROUP BYs.

The long version: Currently, in the logical planner when we build aggregate nodes, validation is split across the logic for building the aggregate schema (`LogicalPlanner#buildAggregateSchema()`) and for performing the aggregate analysis (`AggregateAnalyzer#analyze()`). The former throws a nice error message if the statement contains a function call that doesn’t exist. The latter throws a nice error message if the GROUP BY expressions contain window bounds (since doesn't support that today). However, the two pieces of logic throw unhelpful errors in the reverse scenarios: if the schema building logic encounters window bounds, it throws an error saying the column doesn’t exist. If the aggregate analyzer encounters an unknown function, it ignores it and returns an error about not finding any aggregate functions in the select expressions. In order to get helpful error messages in both scenarios, this PR updates `AggregateAnazlyer` to throw a helpful error message if a nonexistent function call is encountered, and moves the AggregateAnalyzer code above the schema building code in order to fail early (with helpful error messages).

### Testing done 

Unit + QTT.

### Reviewer checklist

### DIFF
--- a/ksqldb-common/src/main/java/io/confluent/ksql/function/FunctionRegistry.java
+++ b/ksqldb-common/src/main/java/io/confluent/ksql/function/FunctionRegistry.java
@@ -28,7 +28,7 @@ public interface FunctionRegistry {
   SqlType DEFAULT_FUNCTION_ARG_SCHEMA = SqlTypes.BIGINT;
 
   /**
-   * Test if the supplied {@code functionName} is an aggregate function.
+   * Test if there is an aggregate function with the supplied {@code functionName}.
    *
    * <p>Note: unknown functions result in {@code false} return value.
    *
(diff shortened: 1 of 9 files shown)


ksqldb-common/src/main/java/io/confluent/ksql/function/FunctionRegistry.java
```diff
@@ -47,6 +47,14 @@
    */
   boolean isTableFunction(FunctionName functionName);
 
+  /**
+   * Test whether this is a function with the supplied {@code functionName}.
```
agavra, line 51:
```suggestion
   * Test whether there is a function with the supplied {@code functionName}.
```

agavra (approved):
nice UX win! LGTM

PR #6797, opened 2020-12-17
feat: bound the size of the shared buffer cache used by queries

rodesai:
bound the total size of the shared buffer cache used by queries. The
bound is controlled by 2 configs, one for controlling the limit for
persistent queries, the other for transient queries. Today, this limit
is implemented by accounting the full limit for each query against the
total bound. In the future we may implement by using a shared cache.

### DIFF
--- a/ksqldb-common/src/main/java/io/confluent/ksql/util/KsqlConfig.java
+++ b/ksqldb-common/src/main/java/io/confluent/ksql/util/KsqlConfig.java
@@ -359,6 +359,17 @@ public class KsqlConfig extends AbstractConfig {
   private static final String KSQL_PROPERTIES_OVERRIDES_DENYLIST_DOC = "Comma-separated list of "
       + "properties that KSQL users cannot override.";
 
+  public static final String KSQL_TOTAL_CACHE_MAX_BYTES_BUFFERING =
+      "ksql.cache.max.bytes.buffering.total";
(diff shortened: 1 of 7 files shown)


ksqldb-common/src/main/java/io/confluent/ksql/util/KsqlConfig.java
```diff
@@ -359,6 +359,17 @@
   private static final String KSQL_PROPERTIES_OVERRIDES_DENYLIST_DOC = "Comma-separated list of "
       + "properties that KSQL users cannot override.";
 
+  public static final String KSQL_TOTAL_CACHE_MAX_BYTES_BUFFERING =
+      "ksql.cache.max.bytes.buffering.total";
```
stevenpyzhang, line 363: I think based on some of the other query configs we have already, these two configs name  would be more suitable
`ksql.query.transient.max.bytes.buffering.total`
`ksql.query.persistent.max.bytes.buffering.total`

ksqldb-engine/src/main/java/io/confluent/ksql/query/QueryExecutor.java
```diff
@@ -57,6 +57,7 @@
 import io.confluent.ksql.util.TransientQueryMetadata;
 import io.confluent.ksql.util.TransientQueryMetadata.ResultType;
 import java.util.Arrays;
+import java.util.Collection;
```
stevenpyzhang, line 60: nit: remove

stevenpyzhang (approved):
lgtm

Issue #5754, opened 2020-07-01
Support PARTITION BY on multiple expressions


big-andy-coates:
Resulting in a structured key.

PR #6803, opened 2020-12-18
feat: support PARTITION BY on multiple expressions

vcrfxia:
### Description 

Fixes https://github.com/confluentinc/ksql/issues/5754

This PR adds support for PARTITION BY with multiple expressions, resulting in multiple key columns. There are no backwards compatibility concerns as the ksqlDB syntax did not support this prior to this PR.

Docs will come in a separate PR.

### Testing done 

QTT.

### Reviewer checklist


ksqldb-engine/src/main/java/io/confluent/ksql/planner/LogicalPlanner.java
```diff
@@ -663,7 +665,7 @@ private LogicalSchema buildAggregateSchema(
 
   private LogicalSchema buildRepartitionedSchema(
       final PlanNode sourceNode,
-      final Expression partitionBy
+      final List<Expression> partitionBy // TODO: rename?
```
agavra: i think this is OK

ksqldb-engine/src/main/java/io/confluent/ksql/planner/LogicalPlanner.java
```diff
@@ -405,7 +407,7 @@ private PreJoinRepartitionNode buildInternalRepartitionNode(
         ExpressionTreeRewriter.rewriteWith(plugin, joinExpression);
 
     final LogicalSchema schema =
-        buildRepartitionedSchema(source, rewrittenPartitionBy);
+        buildRepartitionedSchema(source, Collections.singletonList(rewrittenPartitionBy));
```
agavra, line 410: nit: since we already import `ImmutableList` let's just use `ImmutableList.of` here

ksqldb-engine/src/main/java/io/confluent/ksql/planner/plan/PreJoinRepartitionNode.java
```diff
@@ -111,7 +112,7 @@ public void setKeyFormat(final KeyFormat format) {
     return getSource().buildStream(builder)
         .selectKey(
             valueFormat.getFormatInfo(),
-            partitionBy,
+            Collections.singletonList(partitionBy),
```
agavra, line 115: nit: same thing about `ImmutableList`

ksqldb-engine/src/main/java/io/confluent/ksql/planner/plan/UserRepartitionNode.java
```diff
@@ -103,9 +107,19 @@ public Expression resolveSelect(final int idx, final Expression expression) {
 
   @Override
   void validateKeyPresent(final SourceName sinkName, final Projection projection) {
-    if (!(partitionBy instanceof NullLiteral) && !projection.containsExpression(partitionBy)) {
-      final ImmutableList<Expression> keys = ImmutableList.of(originalPartitionBy);
+    if (!PartitionByParamsFactory.isPartitionByNull(partitionBy)
+        && !containsExpressions(projection, partitionBy)) {
+      final ImmutableList<Expression> keys = ImmutableList.copyOf(originalPartitionBy);
```
agavra: do we still need to make a copy of the original partition by now that it's a list?

ksqldb-engine/src/main/java/io/confluent/ksql/structured/SchemaKStream.java
```diff
@@ -314,15 +314,15 @@ static Expression rewriteTimeComparisonForFilter(final Expression expression) {
...
+      final List<Expression> keyExpression,
       final Optional<KeyFormat> forceInternalKeyFormat,
       final Stacker contextStacker,
       final boolean forceRepartition
   ) {
     final boolean keyFormatChange = forceInternalKeyFormat.isPresent()
         && !forceInternalKeyFormat.get().equals(keyFormat);
 
-    final boolean repartitionNeeded = repartitionNeeded(ImmutableList.of(keyExpression));
+    final boolean repartitionNeeded = repartitionNeeded(ImmutableList.copyOf(keyExpression));
```
agavra, line 325: ditto: no need for the copy

ksqldb-functional-tests/src/test/resources/query-validation-tests/partition-by.json
```diff
@@ -1025,6 +1025,140 @@
...
+      "post": {
+        "sources": [
+          {"name": "OUTPUT", "type": "stream", "schema": "ID INT KEY, NAME STRING KEY, AGE INT"}
+        ]
+      }
+    },
+    {
+      "name": "multiple key columns",
+      "statements": [
+        "CREATE STREAM INPUT (NAME STRING KEY, ID INT KEY, AGE INT) with (kafka_topic='input', format='JSON');",
```
agavra, line 1106: might be interesting to do this with `DELIMITED` so that we can assert that the key order actually changes

ksqldb-functional-tests/src/test/resources/query-validation-tests/partition-by.json
```diff
@@ -1025,6 +1025,140 @@
...
+      "inputs": [{"topic": "input", "key": {"NAME": "bob", "ID": 10}, "value": {"AGE": 30}}],
+      "outputs": [{"topic": "OUTPUT", "key": {"ID": 10, "NAME": "bob"}, "value": {"AGE": 30}}],
+      "post": {
+        "sources": [
+          {"name": "OUTPUT", "type": "stream", "schema": "ID INT KEY, NAME STRING KEY, AGE INT"}
+        ]
+      }
+    },
+    {
+      "name": "multiple columns with expressions",
```
agavra, line 1128: can we name one of the expressions in the projection to make sure that the key name is properly propagated to the schema?

ksqldb-parser/src/main/java/io/confluent/ksql/parser/tree/PartitionBy.java
```diff
@@ -17,27 +17,43 @@
...
+        .copyOf(requireNonNull(partitionByExpressions, "partitionByExpressions"));
+
+    if (partitionByExpressions.isEmpty()) {
+      throw new KsqlException("PARTITION BY requires at least one expression");
+    }
+
+    final HashSet<Object> partitionBys = new HashSet<>(partitionByExpressions.size());
+    partitionByExpressions.forEach(exp -> {
+      if (!partitionBys.add(exp)) {
+        throw new KsqlException("Duplicate PARTITION BY expression: " + exp);
```
agavra, line 50: what's wrong with a duplicate partition by expression? I don't see any reason why a user might want it, but I don't see why not either (e.g. maybe their output data expects `userId, userSpecialId` in the key and this stream always has the same value for both)

I get that the key name conflicts might be a little weird, so we can do this in a follow-up PR, but I don't think we should prohibit it

ksqldb-parser/src/main/java/io/confluent/ksql/parser/tree/PartitionBy.java
```diff
@@ -54,18 +70,18 @@ public boolean equals(final Object o) {
       return false;
     }
     final PartitionBy groupBy = (PartitionBy) o;
```
agavra: not your code, but can we rename this `partitionBy`?

ksqldb-streams/src/main/java/io/confluent/ksql/execution/streams/PartitionByParamsFactory.java
```diff
@@ -76,118 +79,168 @@ private PartitionByParamsFactory() {
...
+      final List<Object> newKeyComponents = evaluators.stream()
+          .map(evaluator -> evaluator.evaluate(oldK, row))
+          .collect(Collectors.toList());
+
+      final K key =
+          executionKeyFactory.constructNewKey(oldK, GenericKey.fromList(newKeyComponents));
+
+      if (row != null) {
+        for (int i = 0; i < partitionByCol.size(); i++) {
+          if (partitionByCol.get(i).shouldAppend) {
```
agavra: instead of relying on the implicit ordering of the partition by, it might make sense to lookup the `partitionByCol.name` in the `resultSchema` at the cost of a bit of performance. thoughts?

ksqldb-functional-tests/src/test/resources/query-validation-tests/partition-by.json
```diff
@@ -1098,12 +1150,27 @@
     },
     {
       "name": "multiple columns including ROWTIME",
+      "comments": [
+        "Ideally the third row of output would contain ROWTIME populated as 12 in the output key.",
+        "This doesn't happen today because ROWTIME is evaluated from the value, which is null in this",
+        "example. To fix this, we'd have to add special handling to detect when a key expression",
+        "depends only on ROWTIME, similar to how today we have special handling to detect when a key",
+        "expression depends only on key columns."
```
agavra, line 1158: or... we can finally swap over to using PAPI :)

agavra (approved):
LGTM! I went commit-by-commit so some of my comments may already be out-of-date, in which case you can just ignore them :)

ksqldb-engine/src/main/java/io/confluent/ksql/planner/LogicalPlanner.java
vcrfxia: You mean we import ImmutableList in this file or in the module? I don't see it in the file but I can add it.

ksqldb-engine/src/main/java/io/confluent/ksql/planner/plan/PreJoinRepartitionNode.java
vcrfxia: Sure. For my edification, why is ImmutableList preferred?

ksqldb-parser/src/main/java/io/confluent/ksql/parser/tree/PartitionBy.java
vcrfxia: This is consistent with what we do for multi-column GROUP BY (which has been around for a long time): https://github.com/confluentinc/ksql/blob/e7f1f4786e07463128d1ea73a010757406fc180e/ksqldb-parser/src/main/java/io/confluent/ksql/parser/tree/GroupBy.java#L49-L53

I believe the reason indeed has to do with naming.

ksqldb-engine/src/main/java/io/confluent/ksql/planner/plan/UserRepartitionNode.java
vcrfxia: Nope, removed.

ksqldb-engine/src/main/java/io/confluent/ksql/planner/LogicalPlanner.java
vcrfxia: Already renamed. Sorry for not squashing my pre-cleanup commits as I usually do. Not worth your time looking at my intermediate thought process while drafting, haha.

ksqldb-streams/src/main/java/io/confluent/ksql/execution/streams/PartitionByParamsFactory.java
vcrfxia: Do you mean replacing `shouldAppend` with a check to see whether the `partitionByCol.name` is present in the `resultSchema` as a value column, or do you mean iterating through `resultSchema` rather than `partitionByCols`? The former doesn't seem like an improvement to me, if we're still relying on the ordering of `partitionByCols` in the iteration. To remove reliance on ordering, we can do the latter and replace the lists (of columns and evaluators) with maps keyed on column name instead, but it's not clear to me that's better. It feels slightly harder to reason about code-wise but I wouldn't mind making the change. 

What's your concern regarding relying the ordering? Are you worried it's brittle, or something else?

ksqldb-functional-tests/src/test/resources/query-validation-tests/partition-by.json
vcrfxia: Done -- added a new test since delimited doesn't allow testing of null handling.

ksqldb-engine/src/main/java/io/confluent/ksql/planner/LogicalPlanner.java
agavra: oh huh, I must've been looking at a separate file. - in `PreJoinRepartitionNode` for example it was already there before this PR. It's nice to just standardize on one (and most of hte code uses `ImmutableList#of`, though there isn't really any good reason to use that over `Collections.singletonList` except for mabye that `ImmutableList` has more general usages)

ksqldb-parser/src/main/java/io/confluent/ksql/parser/tree/PartitionBy.java
agavra: sounds good, we can keep it like that for now

ksqldb-streams/src/main/java/io/confluent/ksql/execution/streams/PartitionByParamsFactory.java
agavra: When I was reviewing the code I thought we could just set it at the index (e.g. `row.set(resultSchema.get(partitionCol.name).index())`). So we still keep the iteration on `partitionByCol` but we set it in the row based on it's index in the schema.

> What's your concern regarding relying the ordering? Are you worried it's brittle, or something else?

yeah - I'm worried that it's brittle.

vcrfxia: Ah, interesting. (We'd have to first re-size the row before calling `.set()` but that's an implementation detail.) The advantage of your proposal is that we can change the details of the result schema without needing to update the logic here. If we extended that further, really we should also be setting the existing value fields in the new row based on the result schema, rather than leaving those intact and appending key columns. 

I guess I'm not convinced this change is necessary since our test coverage in this area is quite good -- lots of tests would break if someone modified the result schema without corresponding updates here. OTOH, it's very possible I'm biased towards thinking this code is understandable as is since I've been working on it for a while. If your assessment differs as someone who hasn't worked with this code as much, I'm inclined to go with your judgment.

As for performance, LogicalSchema stores a list of columns so finding a particular column might be slow. If we implemented this we'd want to build the index mapping outside the creation of the actual mapper, and have the mapper use the index mapping directly. I'm not opposed to this. If you think it's preferable I can open a follow-up PR.

PR #6809, opened 2020-12-21
test: additional test coverage for multi-column key null-handling (MINOR)

vcrfxia:
### Description 

Adds some additional  QTT coverage for null-handling in the context of multi-column keys:
- GROUP BY drops records if any key column evaluates to null
- Table sources do not drop records with null primary key columns, unless the entire key is null

### Testing done 

Test-only change.

### Reviewer checklist

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -31,4 +31,4 @@ logs/
 .settings
 .tern-port
 ui/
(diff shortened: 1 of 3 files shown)


vcrfxia:
Minor, test-only change. Merging without review.

Issue #6802, opened 2020-12-18
Improve error message for invalid pull queries


mikebin:
**Is your feature request related to a problem? Please describe.**
Error message for invalid pull queries mentions that only a single key can be specified in 0.14.0. Since multiple keys can now be specified in an `IN` clause, perhaps this error message should be updated.

```
Pull queries require a WHERE clause that:
 - limits the query to a single key, e.g. `SELECT * FROM X WHERE <key-column>=Y;`.
```

**Describe the solution you'd like**
Update error message to include mention of `IN` clause.

PR #6811, opened 2020-12-21
chore: add mention of IN clause to pull query error msg (MINOR)

vcrfxia:
### Description 

Fixes https://github.com/confluentinc/ksql/issues/6802

Previously:
```
ksql> select * from foo where v in ('val1', 'val2');

IN expression on unsupported column: V.  See https://cnfl.io/queries for more info.
Pull queries require a WHERE clause that:
 - limits the query to keys only, e.g. `SELECT * FROM X WHERE <key-column>=Y;`.
 - specifies an equality condition that is a conjunction of equality expressions that cover all keys.
```
Now:
```
ksql> select * from foo where v in ('val1', 'val2');

IN expression on unsupported column: V.  See https://cnfl.io/queries for more info.
Pull queries require a WHERE clause that:
 - limits the query to keys only, e.g. `SELECT * FROM X WHERE <key-column>=Y;`.
 - either:
   	 - specifies an equality condition that is a conjunction of equality expressions that cover all keys.
   	 - specifies a list of literal values for the key column of interest, e.g. `SELECT * FROM X WHERE <key-column> IN ('value_1', 'value_2');`
```

### Testing done 

The only change in this PR is an update to an error message.

### Reviewer checklist

### DIFF
--- a/ksqldb-engine/src/main/java/io/confluent/ksql/physical/pull/operators/WhereInfo.java
+++ b/ksqldb-engine/src/main/java/io/confluent/ksql/physical/pull/operators/WhereInfo.java
@@ -179,8 +179,8 @@ public final class WhereInfo {
             + System.lineSeparator()
             + " - either:"
             + System.lineSeparator()
-            + "   \t - specifies an equality condition that is a conjunction of equality expressions "
-            + "that cover all keys."
+            + "   \t - specifies an equality condition that is a conjunction of equality "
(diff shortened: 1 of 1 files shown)


mikebin (approved):
LGTM

AlanConfluent (approved):
Nice.  I should have updated this message initially, but must have missed it.  Thanks!

PR #6813, opened 2020-12-22
feat: Make pull queries streamed asynchronously

AlanConfluent:
### Description 
Before this PR, pull queries were gathered in memory as a `TableRows` object containing a `List<List<?>> rows` across all of the endpoints where pull queries are exposed.

This change adds the class `PullQueryQueue`, which is a queue of result rows meant to decouple the producers and consumers of row data, allowing for producer calls that are partially complete to be enqueued across both local and remote sources of data.  This largely follows the methodology used by push queries.  This allows:
 - Fetches for large batches of data to begin immediately returning results rather than waiting until each batch from each partition is complete before returning the first row.
 - Fetches for large amounts of data being that may be prohibitively expensive or not possible to hold it all in memory using the old methodology, whereas this change keeps relatively few rows in memory at a time and can apply back-pressure to the producers if the queue is filled.
 - Queries like table scans and range queries, which may return many more rows than past pull queries allowed for.

Specifically on each end of the queue:
 - New rows are produced and enqueued by `PullPhysicalPlan` if the request is being handled locally or `HARouting` if the request must be forwarded to another node.
 - Rows are consumed by the request thread of the endpoint.
 
 For each of the endpoints:
  - `StreamedQueryResource`: This uses chunked encoding responses using the class `PullQueryStreamWriter` which is a `StreamingOutput` response type.  This class periodically reads from the queue and writes a chunk to the response.
  - `QueryEndpoint`: This uses a `KsqlPullQueryHandle` with the existing `BlockingQueryPublisher` to connect a publisher to the queued data.
  - `PullQueryPublisher`: This uses `PullQuerySubscription`, a new `PollingSubscription` rather than the former block of logic that fed raw rows to the subscriber.

### Testing done 
Ran unit and integration tests.  Also manually experimented with batches of rows being streamed back to the user by introducing artificial delays.

### Reviewer checklist


ksqldb-engine/src/main/java/io/confluent/ksql/query/PullQueryQueue.java
```diff
@@ -0,0 +1,197 @@
...
+
+  public PullQueryQueue() {
+    this(BLOCKING_QUEUE_CAPACITY, 100);
+  }
+
+  public PullQueryQueue(
+      final int queueSizeLimit,
+      final int offerTimeoutMs) {
+    this.callback = new QueueToCompletionCallback(this::isClosed);
+    this.rowQueue = new LinkedBlockingQueue<>(queueSizeLimit);
```
agavra, line 52: `LinkedBlockingQueue` is pretty bad for GC characteristics since a new node is created on each enqueue and the amount of time it lives is likely to have it promoted into survivor space. Any reason not to use `ArrayBlockingQueue` since we know the size ahead of time?

ksqldb-engine/src/main/java/io/confluent/ksql/query/PullQueryQueue.java
```diff
@@ -0,0 +1,197 @@
...
+  public static final int BLOCKING_QUEUE_CAPACITY = 50;
+  private static final Function<PullQueryRow, KeyValue<List<?>, GenericRow>>
+      PQ_ROW_TO_KEY_VALUE = row -> KeyValue.keyValue(null, row.getGenericRow());
+
+  private final BlockingQueue<PullQueryRow> rowQueue;
+  private final int offerTimeoutMs;
+  private volatile boolean closed = false;
+
+  private LimitHandler limitHandler;
+  private LimitQueueCallback callback;
```
agavra, line 42:
```suggestion
  private LimitQueueCallback onLimitExceededCallback;
```

ksqldb-engine/src/main/java/io/confluent/ksql/query/PullQueryQueue.java
```diff
@@ -0,0 +1,197 @@
...
+        parent.setLimitHandler(limitHandler);
+      }
+    };
+  }
+
+  @Override
+  public KeyValue<List<?>, GenericRow> poll(final long timeout, final TimeUnit unit)
+      throws InterruptedException {
+    return Optional.ofNullable(rowQueue.poll(timeout, unit))
+        .map(PQ_ROW_TO_KEY_VALUE)
```
agavra, line 90: nit: if we implemented `PQ_ROW_TO_KEY_VALUE` to handle `nulls` (return `null` if the input is `null`) then we can avoid the Optional wrapping here and just return `PQ_ROW_TO_KEY_VALUE.apply(rowQueue.poll(timeout, unit))`

ksqldb-engine/src/main/java/io/confluent/ksql/query/PullQueryQueue.java
```diff
@@ -0,0 +1,197 @@
...
+  @Override
+  public KeyValue<List<?>, GenericRow> poll() {
+    return Optional.ofNullable(rowQueue.poll())
+        .map(PQ_ROW_TO_KEY_VALUE)
+        .orElse(null);
+  }
+
+  @Override
+  public void drainTo(final Collection<? super KeyValue<List<?>, GenericRow>> collection) {
+    final List<PullQueryRow> list = new ArrayList<>();
```
agavra, line 103: can we avoid iterating all the elements in the queue twice? seems unnecessary

ksqldb-engine/src/main/java/io/confluent/ksql/query/PullQueryQueue.java
```diff
@@ -0,0 +1,197 @@
...
+  }
+
+  // Unlike push queries that run forever until someone deliberately kills it, pull queries have an
+  // ending.  When they've reached their end, this is expected to be called.  Also, if the system
+  // wants to end pull queries prematurely, this should also be called.
+  public void close() {
+    closed = true;
+    // Unlike limits based on a number of rows which can be checked and possibly triggered after
+    // every queuing of a row, pull queries just declare they've reached their limit when close is
+    // called.
```
agavra, line 135: what happens when we implement a `LIMIT` clause on pull queries (such as table scans)? Would we want to call `limitReached` twice or would that just be a physical plan node that just calls close on the underlying queue?

ksqldb-engine/src/main/java/io/confluent/ksql/query/PullQueryQueue.java
```diff
@@ -0,0 +1,197 @@
...
+  public void acceptTableRows(final List<PullQueryRow> tableRows) {
+    if (tableRows == null) {
+      return;
+    }
+    for (PullQueryRow row : tableRows) {
+      acceptRow(row);
+    }
+  }
+
+  public void acceptRow(final PullQueryRow row) {
```
agavra, line 152: I think this should return a boolean `accepted` so that we can also short-circuit in `accetpTableRows` the first time `!callback.shouldQueue()`

ksqldb-functional-tests/src/test/resources/rest-query-validation-tests/pull-queries-against-materialized-aggregates.json
```diff
@@ -1797,22 +1797,23 @@
       "statements": [
         "CREATE STREAM INPUT (ID DOUBLE KEY, IGNORED INT) WITH (kafka_topic='test_topic', value_format='JSON');",
         "CREATE TABLE AGGREGATE AS SELECT ID, COUNT(1) AS COUNT FROM INPUT WINDOW TUMBLING(SIZE 1 SECOND) GROUP BY ID;",
-        "SELECT * FROM AGGREGATE WHERE ID IN (10.1, 8.1);",
+        "SELECT * FROM AGGREGATE WHERE ID IN (10.5, 8.5);",
```
agavra, line 1800: out of interest, why did you make this change?

ksqldb-rest-app/src/main/java/io/confluent/ksql/api/impl/BlockingQueryPublisher.java
```diff
@@ -59,13 +60,23 @@ public BlockingQueryPublisher(final Context ctx,
...
     this.queue.setQueuedCallback(this::maybeSend);
-    this.queue.setLimitHandler(() -> complete = true);
+    this.queue.setLimitHandler(() -> {
+      complete = true;
+      // This allows us to hit the limit without having to queue one last row
+      if (queue.isEmpty()) {
+        ctx.runOnContext(v -> sendComplete());
+      } else {
+        ctx.runOnContext(v -> doSend());
+      }
```
agavra, line 76: I'm confused. Why is this necessary? Why would we otherwise have to enqueue another row?

ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/resources/streaming/PullQueryPublisher.java
```diff
@@ -79,98 +78,68 @@
...
-        query
+    final RoutingOptions routingOptions = new PullQueryConfigRoutingOptions(
+        query.getSessionConfig().getConfig(false),
+        query.getSessionConfig().getOverrides(),
+        ImmutableMap.of()
+    );
+
+    PullQueryExecutionUtil.checkRateLimit(rateLimiter);
+
+    final PullQueryResult result = ksqlEngine.executePullQuery(
```
agavra, line 89: previously we were calling this inside the `Subscription#request` method, now we're doing it before calling the subscriber. I guess that this is intentional because of the new asynchronous nature of pull queries? Just wanted to confirm that I understand correctly (and to confirm that it is OK to do this here in the `subscribe` method instead of on the first `poll` call)

ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/resources/streaming/PullQueryStreamWriter.java
```diff
@@ -0,0 +1,215 @@
...
+import java.time.Clock;
+import java.util.List;
+import java.util.Objects;
+import java.util.Optional;
+import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.TimeUnit;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+@SuppressWarnings("checkstyle:CyclomaticComplexity")
```
agavra, line 42: nit: most of the code uses:
```
  // CHECKSTYLE_RULES.OFF: CyclomaticComplexity
```
just around the specific method that breaks the rule

agavra (commented):
Starting to lose steam, so I'll revisit in a bit. Note to myself, I've reviewed up until `PullQueryStreamWriter` and will continue later today!

ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/resources/streaming/PullQueryStreamWriter.java
```diff
@@ -0,0 +1,215 @@
...
+          // there a next.  This is to write proper json
+          final PullQueryRow toProcess = head;
+          head = row;
+          if (toProcess == null) {
+            continue;
+          }
+          writeRow(toProcess, head, sb);
+          if (sb.length() >= FLUSH_SIZE_BYTES || (clock.millis() - lastFlush) >= MAX_FLUSH_MS) {
+            output.write(sb.toString().getBytes(StandardCharsets.UTF_8));
+            output.flush();
```
agavra, line 106: I'm wondering how this works from a client perspective. It looks like we'd be flushing incomplete JSON, right? Is that what was happening previously for pull queries? Would clients know how to handle that properly?

ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/resources/streaming/PullQueryStreamWriter.java
```diff
@@ -0,0 +1,215 @@
...
+    this.disconnectCheckInterval = disconnectCheckInterval;
+    this.pullQueryQueue = Objects.requireNonNull(pullQueryQueue, "pullQueryQueue");
+    this.clock = Objects.requireNonNull(clock, "clock");
+    connectionClosedFuture.thenAccept(v -> connectionClosed = true);
+    result.onException(t -> pullQueryException = t);
+    result.onCompletion(v -> limitReached = true);
+  }
+
+  @Override
+  public void write(final OutputStream output) {
```
agavra, line 79: there's a lot of different levels going on and wrapping my head around passing around different heads/string buffers and flushing in the middle (etc...) is a struggle. At a minimum, I think we can do with inline comments and javadocing (including the private methods), though I challenge you to take another stab and see if you can refactor this to make it cleaner and reduce the cyclomatic complexity (this is one of those times where I think checkstyle got it right, it needs improvement).

There's a lot of complexity in creating proper JSON (commas in the right places, the header row being first, etc...). I wonder if there's a library that we can use to make that simpler.

ksqldb-rest-app/src/test/java/io/confluent/ksql/api/ApiTest.java
```diff
@@ -110,6 +115,53 @@ public void shouldExecutePullQuery() throws Exception {
     assertThat(queryId, is(nullValue()));
   }
 
+  public static void main(String[] args) throws ExecutionException, InterruptedException {
```
agavra, line 118: did you mean to check this in?

agavra (commented):
A few more comments, all that's left now are the tests :) I'll get to them tomorrow morning

ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/resources/streaming/PullQueryStreamWriter.java
```diff
@@ -0,0 +1,215 @@
...
+  }
+
+  private void writeRow(
+      final PullQueryRow row,
+      final PullQueryRow head,
+      final StringBuilder sb
+  ) {
+    // Send for a comma after the header
+    if (!sentAtLeastOneRow) {
+      sb.append(",").append(System.lineSeparator());
```
agavra, line 136: just wondering, how do we handle content that contains the line separator e.g. a column that has `foo\nbar` as its contents? will that throw anything off? might be worth testing that

ksqldb-rest-app/src/test/java/io/confluent/ksql/rest/server/resources/streaming/PullQueryStreamWriterTest.java
```diff
@@ -0,0 +1,268 @@
...
+    final List<String> lines = getOutput(out);
+    assertThat(lines, contains(
+        containsString("header"),
+        containsString("Row1"),
+        containsString("Row2"),
+        containsString("Row3")));
+  }
+
+  @Test
+  public void shouldExitAndDrainIfLimitReached() {
```
agavra, line 148: should we have negative tests for this and the above? i.e. does not drain if limit not reached or pull query not closed?

ksqldb-rest-app/src/test/java/io/confluent/ksql/rest/server/resources/streaming/PullQueryStreamWriterTest.java
```diff
@@ -0,0 +1,268 @@
...
+
+      return null;
+    };
+  }
+
+  private static PullQueryRow streamRow(final Object row) {
+    return new PullQueryRow(ImmutableList.of(row), SCHEMA, Optional.empty());
+  }
+
+  private static List<String> getOutput(final ByteArrayOutputStream out) {
```
agavra, line 262: can we add a test for valid JSON? (,s in the right place, pairs of [] etc...)

agavra (commented):
Well... morning quickly turned to afternoon and then evening but I finally got around to the tests for this PR.

agavra:
Overall, the PR seems like a huge improvement. I have some questions inline (amongst a less-important small army of nits) that I'd like addressed before giving the green light. I'll mark those with the 🚀 emoji so that we can focus on the important ones

ksqldb-engine/src/main/java/io/confluent/ksql/query/PullQueryQueue.java
AlanConfluent: I changed this to `Runnable queuedCallback` since that's all it does now.

ksqldb-engine/src/main/java/io/confluent/ksql/query/PullQueryQueue.java
AlanConfluent: I agree it's simpler to understand.  I just love Optional so much.

ksqldb-engine/src/main/java/io/confluent/ksql/query/PullQueryQueue.java
AlanConfluent: I could repeatedly call poll to drain it, though this is meant to be more efficient with all of the synchronization.  In the end, the array is only 50 elements long, max.

I think it's probably better to take the small allocation rather than the synchronization hit, but maybe you disagree. I can change it, if you prefer.

ksqldb-engine/src/main/java/io/confluent/ksql/query/PullQueryQueue.java
AlanConfluent: Nope, no reason.  Good suggestion.

ksqldb-engine/src/main/java/io/confluent/ksql/query/PullQueryQueue.java
AlanConfluent: Sure, that makes sense.  If it doesn't succeed, the thread has either been interrupted or the queue is closed, so no need to continue.

ksqldb-engine/src/test/java/io/confluent/ksql/query/PullQueryQueueTest.java
```diff
@@ -0,0 +1,148 @@
...
+    assertThat(queue.size(), is(QUEUE_SIZE));
+  }
+
+  private void givenWillCloseQueueAsync() {
+    executorService = Executors.newSingleThreadScheduledExecutor();
+    executorService.schedule(queue::close, 200, TimeUnit.MILLISECONDS);
+  }
+
+  private void givenQueue() {
+    queue = new PullQueryQueue(QUEUE_SIZE, 1);
```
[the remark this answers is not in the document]
AlanConfluent: It should effectively busy loop checking close.  If for some reason, there was a race in the test between closing and enqueuing with a full queue (because that requires a dequeue to happen), then it could be flaky, but there isn't.

ksqldb-engine/src/test/java/io/confluent/ksql/query/PullQueryQueueTest.java
```diff
@@ -0,0 +1,148 @@
...
+    // When:
+    queue.acceptRow(VAL_ONE);
+    queue.acceptRow(VAL_TWO);
+
+    // Then:
+    assertThat(drainValues(), contains(VAL_ONE, VAL_TWO));
+  }
+
+  @Test
+  public void shouldQueueNullKey() {
```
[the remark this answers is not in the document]
AlanConfluent: Me neither.  Removed.

ksqldb-functional-tests/src/test/resources/rest-query-validation-tests/pull-queries-against-materialized-aggregates.json
AlanConfluent: Because in floating point representation in the CPU (i.e. binary), `.1` cannot be represented exactly: https://www.exploringbinary.com/why-0-point-1-does-not-exist-in-floating-point

Doing a key lookup using equality with such a value creates weird flakey results sometimes: https://floating-point-gui.de/errors/comparison/

I was banging my head against the wall with these tests not returning the row I thought it should until I tried this.  Look at the result and you'll see that the existing answer is wrong!  `0.5` can be represented exactly in binary.  With all this, it probably doesn't make sense to do `double` key lookups, but I guess we allow it.  

User beware!

AlanConfluent (commented):
@agavra Followed up on some of your comments.  Still working through them.

AlanConfluent (commented):
@agavra I believe I've addressed your comments.  PTAL

ksqldb-engine/src/main/java/io/confluent/ksql/physical/pull/PullQueryResult.java
```diff
@@ -15,30 +15,39 @@
...
 
-  private final List<List<?>> tableRows;
-  private final Optional<List<KsqlNode>> sourceNodes;
   private final LogicalSchema schema;
+  private final Supplier<CompletableFuture<Void>> runner;
   private final QueryId queryId;
+  private final PullQueryQueue pullQueryQueue;
+  private final Optional<PullQueryExecutorMetrics> pullQueryMetrics;
+
+  private CompletableFuture<Void> future = new CompletableFuture<>();
```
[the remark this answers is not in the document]
agavra: ack - makes sense

ksqldb-engine/src/main/java/io/confluent/ksql/query/PullQueryQueue.java
agavra: ack - now that I understand the motivation we can leave it as is. in the future we can profile and if this ends up being a hotspot we can look into it then

ksqldb-functional-tests/src/test/resources/rest-query-validation-tests/pull-queries-against-materialized-aggregates.json
agavra: 🙀

ksqldb-rest-app/src/test/java/io/confluent/ksql/rest/server/resources/streaming/PullQueryPublisherTest.java
```diff
@@ -103,26 +118,40 @@ public void setUp() {
...
+      Collection<? super KeyValue<List<?>, GenericRow>> c = inv.getArgument(0);
+      if (times[0] == 0) {
+        c.add(KV1);
+      } else if (times[0] == 1) {
+        c.add(KV2);
+        completeCaptor.getValue().accept(null);
+      }
+      times[0]++;
+      return null;
+    }).when(pullQueryQueue).drainTo(any());
```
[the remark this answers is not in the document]
agavra: I trust that if you played around with it and it wasn't clear then it would probably require more work to retrofit it than it's worth. I'm happy leaving it as is

ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/resources/streaming/PullQueryStreamWriter.java
```diff
@@ -0,0 +1,369 @@
...
+      final PullQueryRow row = pullQueryQueue.pollRow(
+          disconnectCheckInterval,
+          TimeUnit.MILLISECONDS
+      );
+
+      if (row != null) {
+        // The head becomes the next thing to process and the newly polled row becomes the head.
+        final PullQueryRow toProcess = head;
+        head = row;
+        return toProcess;
```
agavra, line 354: i guess your algorithm above handles this gracefully, but the first time you call `pollNextRow` won't it always be `null`? might just want to add a little comment there as well so that the next person doesn't scratch their head 😄

ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/resources/streaming/PullQueryStreamWriter.java
agavra: 💯 this is _way_ cleaner and easier for me to reason about. Thanks for the heavy refactoring!

agavra (approved):
Thanks @AlanConfluent! Big step forward. I'm still having a little trouble wrapping my head around the limit handler/reached end code, but I don't want to block this PR on that as you've clearly thought that out and the longer it sits the more there's a chance for nasty merge conflicts. I'll dig into that offline 🙂

ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/resources/streaming/PullQueryStreamWriter.java
```diff
@@ -0,0 +1,369 @@
...
+   * @param hasAnotherRow if there's another row after this one.  This is used for determining how
+   *                      to write proper JSON, e.g. whether to add a comma.
+   */
+  private void writeRow(
+      final PullQueryRow row,
+      final WriterState writerState,
+      final boolean hasAnotherRow
+  ) {
+    // Send for a comma after the header
+    if (!sentAtLeastOneRow) {
```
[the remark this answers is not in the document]
AlanConfluent: You're actually right and that would honestly simplify my code a bit.  The only issue with this is that the code parsing this in both `KsqlTarget` and the CLI looks for "\n" and also assumes if there's a "," it's on the end.  I can potentially fix this, but I don't want to make a big change at this point.

ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/resources/streaming/PullQueryStreamWriter.java
AlanConfluent: Yes, that's right.  I'll add a comment.

ksqldb-engine/src/main/java/io/confluent/ksql/physical/pull/PullPhysicalPlan.java
```diff
@@ -57,22 +58,25 @@ public PullPhysicalPlan(
...
-    final List<List<?>> localResult = new ArrayList<>();
-    List<?> row = null;
+    List<?> row;
     while ((row = (List<?>)next()) != null) {
-      localResult.add(row);
+      if (pullQueryQueue.isClosed()) {
+        // If the queue has been closed, we stop adding rows and cleanup.
+        break;
+      }
+      pullQueryQueue.acceptRow(rowFactory.apply(row, schema));
```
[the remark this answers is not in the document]
guozhangwang: Just throwing out some more ideas here: if a query touches on multiple partitions, we can let the sorting to first happen on each partition itself, and then at the operator we can just do merge-sort based on the returned streaming of sorted rows, which would take const space only.

Even more general, if we allow aggregations in the future, such aggregation push-downs can be applied as well.

ksqldb-engine/src/main/java/io/confluent/ksql/query/PullQueryQueue.java
guozhangwang: Hmm... in SQL the LIMIT clause is deterministic usually, even if there's no ORDER BY (in which case, it would follow the row ordering).

In push / persistent queries this is not hard to maintain, as we naturally follow the offset ordering. But in pull query does this mean we may then return different results if the same query with LIMIT and no ORDER BY is issued multiple times?

ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/resources/streaming/PullQueryStreamWriter.java
guozhangwang: For my own clarification: were you referring to the JSON format of the HTTP response, or the JSON format of a single record?

PR #6820, opened 2020-12-29
chore: Update VERSION_PATTERN to accomodate Prereleased versioned artifacts

andrewegel:
- Add support for Prereleased values. Eg: -alpha1, -beta1234.
- Add unit tests for supported versions.
- Updated exception message to include used pattern for matching.

### Description 
The `6.2.x-alpha1` branch was failing to build in Jenkins due to a failure in some tests. They surfaced as `NoClassDefFound` exceptions, but if you watched the logs, there was actually an earlier `IllegalArgumentException` being thrown about the version not being in a format, so I tracked that down to here, made the regex a little more tolerant of this valid semver format, made the Exception message slightly more helpfull and added some Unit tests to support.

### Testing done 
This is a meta change. Check the `6.2.x-alpha1` branch job in Jenkins and you can see after these changes it passes.

### Reviewer checklist


rodesai (approved):
LGTM

PR #6821, opened 2021-01-01
fix: use the right name for fallback subject for transient queries

### LINKED ISSUE #6817: Wrong prefix for transient query changelog subject names in Schema Registry
**Describe the bug**
In some cases, ksqlDB may generate an invalid subject name for transient query schemas in Schema Registry. For example:

```
_confluent-ksql-default_query_1681835265420415213-KsqlTopic-Reduce-changelog-value
```

Note the missing the `transient` keyword in the subject name prefix.

In these cases, when the query terminates, the schemas are not automatically cleaned up and deleted as expected with https://github.com/confluentinc/ksql/issues/5714.

**To Reproduce**

**Expected behavior**
All schemas created as part of the transient query are hard-deleted when the query terminates.

**Actual behavior**
Misnamed schemas are not cleaned up, resulting in potential unbounded growth of registered schemas in Schema Registry

rodesai:
### Description 
Fixes #6817 plus a small refactor (most of the diff is the refactor):
- Split up KsqlQueryBuilder into PlanBuildContext and RuntimeBuildContext. PlanBuildContext
  contains all the context required to build an execution plan from a logical plan.
  RuntimeBuildContext contains all the context required to build the final runtime topology
  from an execution plan. Ultimately we should make this more abstract to support different
  runtimes, but for now the assumed runtime is Kafka Streams.
- Add the application id to RuntimeBuildContext
- Use the application id to construct the fallback subject name.

### Testing done 
updated unit tests according to the above

### DIFF
--- a/ksqldb-engine/src/main/java/io/confluent/ksql/planner/plan/DataSourceNode.java
+++ b/ksqldb-engine/src/main/java/io/confluent/ksql/planner/plan/DataSourceNode.java
@@ -111,10 +111,10 @@ public class DataSourceNode extends PlanNode {
   }
 
   @Override
-  public SchemaKStream<?> buildStream(final PlanBuildContext builder) {
-    final Stacker contextStacker = builder.buildNodeContext(getId().toString());
+  public SchemaKStream<?> buildStream(final PlanBuildContext buildContext) {
+    final Stacker contextStacker = buildContext.buildNodeContext(getId().toString());
     return schemaKStreamFactory.create(
-        builder,
+        buildContext,
         dataSource,
         contextStacker.push(SOURCE_OP_NAME)
     );
@@ -200,7 +200,7 @@ public class DataSourceNode extends PlanNode {
   interface SchemaKStreamFactory {
 
     SchemaKStream<?> create(
-        PlanBuildContext builder,
+        PlanBuildContext buildContext,
         DataSource dataSource,
         QueryContext.Stacker contextStacker
     );
--- a/ksqldb-engine/src/main/java/io/confluent/ksql/planner/plan/DataSourceNode.java
+++ b/ksqldb-engine/src/main/java/io/confluent/ksql/planner/plan/DataSourceNode.java
@@ -19,7 +19,6 @@ import static java.util.Objects.requireNonNull;
 
 import com.google.common.collect.ImmutableList;
 import com.google.errorprone.annotations.Immutable;
-import io.confluent.ksql.execution.builder.KsqlQueryBuilder;
 import io.confluent.ksql.execution.context.QueryContext;
 import io.confluent.ksql.execution.context.QueryContext.Stacker;
 import io.confluent.ksql.execution.expression.tree.ColumnReferenceExp;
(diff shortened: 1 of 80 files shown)


ksqldb-execution/src/main/java/io/confluent/ksql/execution/runtime/RuntimeBuildContext.java
```diff
@@ -44,7 +44,7 @@
 import org.apache.kafka.streams.StreamsBuilder;
 import org.apache.kafka.streams.kstream.Windowed;
 
-public final class KsqlQueryBuilder {
+public final class RuntimeBuildContext {
```
agavra, line 47: nit: can we add javaodc here that echos what you wrote in the commit description (specifically with the relationship between this and `PlanBuildContext`)

ksqldb-engine/src/main/java/io/confluent/ksql/planner/plan/DataSourceNode.java
```diff
@@ -104,7 +103,7 @@ public int getPartitions(final KafkaTopicClient kafkaTopicClient) {
   }
 
   @Override
-  public SchemaKStream<?> buildStream(final KsqlQueryBuilder builder) {
+  public SchemaKStream<?> buildStream(final PlanBuildContext builder) {
```
agavra, line 106: nit: please don't hate me 😇 but there's quite a few places in this pr where we don't rename `builder` to something like `buildContext`. Probably makes sense to do it now while the diff is in front of us

ksqldb-streams/src/main/java/io/confluent/ksql/execution/streams/SourceBuilder.java
```diff
@@ -387,13 +384,10 @@ private static PhysicalSchema getPhysicalSchema(final SourceStep<?> streamSource
    * </pre>
    */
   private static String changelogTopic(
-      final KsqlQueryBuilder queryBuilder,
+      final RuntimeBuildContext queryBuilder,
       final String stateStoreName
   ) {
-    return ReservedInternalTopics.KSQL_INTERNAL_TOPIC_PREFIX
```
agavra, line 393: to make sure that I understand, this line is the "meat" of the PR, right?

agavra (approved):
lgtm!

ksqldb-streams/src/main/java/io/confluent/ksql/execution/streams/SourceBuilder.java
rodesai: yep, that's right

ksqldb-engine/src/main/java/io/confluent/ksql/planner/plan/DataSourceNode.java
rodesai: fair enough. I got lazy :)

ksqldb-execution/src/main/java/io/confluent/ksql/execution/runtime/RuntimeBuildContext.java
rodesai: ack

PR #6846, opened 2021-01-08
fix: Make Select * avoid code gen for projections

AlanConfluent:
### Description 
Prior to https://github.com/confluentinc/ksql/pull/6684, select * didn't compile expressions for each of the columns, which makes them much more performant.  This PR remakes that change to avoid a regression.

### Testing done 
Unit tests and local benchmarking

### Reviewer checklist


agavra (commented):
LGTM - thanks for the quick fix on this one. For anyone following along, we'll need to figure out something long-term to make all pull queries (not just select * queries) efficient (or set expectations around what the expected performance of queries are).

agavra (approved):

PR #6851, opened 2021-01-11
feat: make UDAFs configurable and remove limit on COLLECT_LIST/SET

### LINKED ISSUE #5738: UDAFs do not respect the Configurable interface
**Describe the bug**

Version: 0.10.0, related to #5737 

[The docs](https://docs.ksqldb.io/en/latest/concepts/functions/#configurable-udf) describe that a UDF can be made configurable with external parameters by implementing the `Configurable` interface. This works as described for scalar functions. But it doesn't work for UDAFs. The `configure()` method is silently ignored. This might be a bit worse than #5737 because UDAF factories expect to return a `Udaf` or `TableUdaf`. Because Java doesn't support type unions, we might need to jump through some hoops to get ksqlDB to realize which UDAFs are configurable and which are not.

Either way, this behavior should also function for UDAFs.

### LINKED ISSUE #6711: Remove collection size limit for COLLECT_SET/LIST
Currently `COLLECT_LIST` and `COLLECT_SET` are limited to maintaining `1000` items. Any additions to these collections beyond this limit will be silently ignored, which can be confusing for users. We should remove this hard limit entirely, and introduce function-level configuration parameters for environments that need to enforce a hard limit on collection sizes.

agavra:
### Description 

fixes #5738
fixes #6711 

This commit adds the ability to configure UDAFs the same way that UDFs can be configured, by implementing the `Configurable` interface. By doing this, we can lift the restriction on the number of elements that `COLLECT_LIST` and `COLLECT_SET` have without opening the flood gates in the confluent cloud offering (we can tune that number later).

The "meat" of the PR is the on-liner in `UdafFactoryInvoker.java` and the rest is plumbing to make sure that we can get the `KsqlConfig` down to that level.

### Testing done 

- Some unit testing
- QTT testing

### Reviewer checklist

### DIFF
--- a/ksqldb-common/src/main/java/io/confluent/ksql/function/AggregateFunctionInitArguments.java
+++ b/ksqldb-common/src/main/java/io/confluent/ksql/function/AggregateFunctionInitArguments.java
@@ -15,9 +15,11 @@
 
 package io.confluent.ksql.function;
 
+import com.google.common.collect.ImmutableMap;
 import java.util.Arrays;
 import java.util.Collections;
 import java.util.List;
+import java.util.Map;
 import java.util.Objects;
 
 /**
@@ -28,18 +30,42 @@ import java.util.Objects;
  */
 public class AggregateFunctionInitArguments {
 
-  private final int udafIndex;
-  private final List<Object> initArgs;
-
   public static final AggregateFunctionInitArguments EMPTY_ARGS =
       new AggregateFunctionInitArguments();
 
-  public AggregateFunctionInitArguments(final int index, final Object... initArgs) {
-    this(index, Arrays.asList(initArgs));
+  private final int udafIndex;
+  private final List<Object> initArgs;
+  private final Map<String, ?> config;
+
+  /**
+   * This method should only be used for legacy "built-in" UDAF
+   * implementations that implement AggregateFunctionFactory directly
+   * such as TopKAggregateFuncitonFactory.
+   *
+   * Otherwise, the config will not be properly passed through
+   */
+  public AggregateFunctionInitArguments(
+      final int index,
+      final Object... initArgs
+  ) {
+    this(index, ImmutableMap.of(/* not a configurable function */), Arrays.asList(initArgs));
   }
 
-  public AggregateFunctionInitArguments(final int index, final List<Object> initArgs) {
+  public AggregateFunctionInitArguments(
+      final int index,
+      final Map<String, ?> config,
+      final Object... initArgs
+  ) {
+    this(index, config, Arrays.asList(initArgs));
+  }
+
+  public AggregateFunctionInitArguments(
+      final int index,
+      final Map<String, ?> config,
+      final List<Object> initArgs
+  ) {
     this.udafIndex = index;
+    this.config = Objects.requireNonNull(config, "config");
     this.initArgs = Objects.requireNonNull(initArgs);
 
     if (index < 0) {
@@ -49,6 +75,7 @@ public class AggregateFunctionInitArguments {
 
   private AggregateFunctionInitArguments() {
     this.udafIndex = 0;
+    this.config = ImmutableMap.of();
     this.initArgs = Collections.emptyList();
   }
 
@@ -62,5 +89,9 @@ public class AggregateFunctionInitArguments {
 
   public List<Object> args() {
     return initArgs;
+  }
+
+  public Map<String, ?> config() {
+    return config;
   }
 }
--- a/ksqldb-common/src/main/java/io/confluent/ksql/function/AggregateFunctionInitArguments.java
+++ b/ksqldb-common/src/main/java/io/confluent/ksql/function/AggregateFunctionInitArguments.java
@@ -40,9 +40,8 @@ public class AggregateFunctionInitArguments {
   /**
    * This method should only be used for legacy "built-in" UDAF
    * implementations that implement AggregateFunctionFactory directly
-   * such as TopKAggregateFuncitonFactory.
-   *
-   * Otherwise, the config will not be properly passed through
+   * such as TopKAggregateFuncitonFactory. Otherwise, the config will
+   * not be properly passed through to the aggregate function.
    */
   public AggregateFunctionInitArguments(
       final int index,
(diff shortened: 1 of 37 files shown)


ksqldb-common/src/main/java/io/confluent/ksql/function/AggregateFunctionInitArguments.java
```diff
@@ -28,18 +30,41 @@
...
   }
 
-  public AggregateFunctionInitArguments(final int index, final List<Object> initArgs) {
+  public AggregateFunctionInitArguments(
+      final int index,
+      final Map<String, ?> config,
+      final List<Object> initArgs
+  ) {
     this.udafIndex = index;
+    this.config = Objects.requireNonNull(config, "config");
```
spena, line 67: Should we call `ImmutableMap.copyOf(config)` so we keep the config completely immutable?

ksqldb-engine/src/main/java/io/confluent/ksql/function/udaf/array/CollectListUdaf.java
```diff
@@ -15,98 +15,105 @@
...
+  }
+
+  private static final class Collect<T> implements TableUdaf<T, List<T>, List<T>>, Configurable {
+
+    private int limit = Integer.MAX_VALUE;
+
+    @Override
+    public void configure(final Map<String, ?> map) {
+      final Object limit = map.get(LIMIT_CONFIG);
+      this.limit = (limit == null) ? this.limit : ((Number) limit).intValue();
```
spena, line 77: How are negative values handled?

ksqldb-engine/src/main/java/io/confluent/ksql/function/udaf/array/CollectSetUdaf.java
```diff
@@ -21,80 +21,90 @@
...
+  }
+
+  private static final class Collect<T> implements Udaf<T, List<T>, List<T>>, Configurable {
+
+    private int limit = Integer.MAX_VALUE;
+
+    @Override
+    public void configure(final Map<String, ?> map) {
+      final Object limit = map.get(LIMIT_CONFIG);
+      this.limit = (limit == null) ? this.limit : ((Number) limit).intValue();
```
spena, line 75: How are negative values handled?

ksqldb-engine/src/main/java/io/confluent/ksql/structured/SchemaKGroupedStream.java
```diff
@@ -67,7 +67,8 @@
       final List<FunctionCall> aggregations,
       final Optional<WindowExpression> windowExpression,
       final FormatInfo valueFormat,
-      final QueryContext.Stacker contextStacker
+      final Stacker contextStacker,
+      final KsqlConfig config
```
spena, line 71: This `config` is not used here, right?  I see `SchemaKTable` is created with `ksqlConfig`.

ksqldb-engine/src/test/java/io/confluent/ksql/function/udaf/array/CollectListUdafTest.java
```diff
@@ -65,14 +66,16 @@ public void shouldMergeIntLists() {
...
+    for (int i = 1; i < 25; i++) {
       runningList = udaf.aggregate(i, runningList);
     }
-    assertThat(runningList, hasSize(1000));
+    assertThat(runningList, hasSize(10));
     assertThat(runningList, hasItem(1));
-    assertThat(runningList, hasItem(1000));
-    assertThat(runningList, not(hasItem(1001)));
+    assertThat(runningList, hasItem(10));
+    assertThat(runningList, not(hasItem(11)));
```
spena, line 78: If limit is 10, why does it have the item 11?

ksqldb-engine/src/main/java/io/confluent/ksql/function/udaf/array/CollectListUdaf.java
agavra: good catch, I will treat negative as "no limit". I'll add that to the documentation

ksqldb-engine/src/test/java/io/confluent/ksql/function/udaf/array/CollectListUdafTest.java
agavra: 😉 the code is `not(hasItem(11))`

ksqldb-engine/src/main/java/io/confluent/ksql/structured/SchemaKGroupedStream.java
agavra: good catch, I used it in `SchemaKGroupedTable` because I didn't realize it had access to the field `ksqlConfig`. I removed it

ksqldb-common/src/main/java/io/confluent/ksql/util/KsqlConfig.java
```diff
@@ -509,6 +509,10 @@ String getName() {
   public static final Set<String> SSL_CONFIG_NAMES = sslConfigNames();
   public static final Set<String> STREAM_TOPIC_CONFIG_NAMES = streamTopicConfigNames();
 
+  public static KsqlConfig empty() {
+    return new KsqlConfig(ImmutableMap.of());
```
spena, line 513: What happened with the static `EMPTY`? That could be reused it, isn't it? Perhaps have the variable back and just return it from this method.

ksqldb-engine/src/test/java/io/confluent/ksql/function/udaf/array/CollectListUdafTest.java
spena: Pfff. I see it. sorry

spena (approved):
LGTM, I just left one single comment this time.

ksqldb-common/src/main/java/io/confluent/ksql/util/KsqlConfig.java
agavra: It was causing issues because of circular static dependencies. Some things depended on it in a static block and then things wouldn't load. It was very weird...

PR #6868, opened 2021-01-14
feat: add lambda syntax to grammar

stevenpyzhang:
### Description 
Adds syntax for lambda functions to the KSQL grammar

LambdaFunctionExpression for representing a lambda function node
LambdaLiteral for representing a lambda variable

During the AstBuilder phase, the lambda variables are treated as UnqualifiedColumnReferences since I couldn't pass in a proper context, so I decided to convert them with the AstSanitizer.

Currently, lambdas won't work properly because the corresponding SqlToJavaVisitor methods haven't been implemented yet.

### Testing done 
Unit test
Manual test

### Reviewer checklist

### DIFF
--- a/ksqldb-engine/src/main/java/io/confluent/ksql/engine/rewrite/AstSanitizer.java
+++ b/ksqldb-engine/src/main/java/io/confluent/ksql/engine/rewrite/AstSanitizer.java
@@ -17,11 +17,12 @@ package io.confluent.ksql.engine.rewrite;
 
 import static java.util.Objects.requireNonNull;
 
+import com.google.common.collect.ImmutableSet;
 import io.confluent.ksql.analyzer.Analysis.AliasedDataSource;
 import io.confluent.ksql.engine.rewrite.ExpressionTreeRewriter.Context;
 import io.confluent.ksql.execution.expression.tree.ColumnReferenceExp;
 import io.confluent.ksql.execution.expression.tree.Expression;
-import io.confluent.ksql.execution.expression.tree.LambdaFunctionExpression;
+import io.confluent.ksql.execution.expression.tree.LambdaFunctionCall;
 import io.confluent.ksql.execution.expression.tree.LambdaLiteral;
 import io.confluent.ksql.execution.expression.tree.QualifiedColumnReferenceExp;
 import io.confluent.ksql.execution.expression.tree.UnqualifiedColumnReferenceExp;
@@ -43,8 +44,10 @@ import io.confluent.ksql.schema.utils.FormatOptions;
 import io.confluent.ksql.util.AmbiguousColumnException;
 import io.confluent.ksql.util.KsqlException;
 import io.confluent.ksql.util.UnknownSourceException;
+import java.util.HashSet;
 import java.util.List;
 import java.util.Optional;
+import java.util.Set;
 import java.util.function.BiFunction;
 import java.util.stream.Collectors;
 
@@ -75,15 +78,15 @@ public final class AstSanitizer {
     final ExpressionRewriterPlugin expressionRewriterPlugin =
         new ExpressionRewriterPlugin(dataSourceExtractor);
 
-    final BiFunction<Expression, Object, Expression> expressionRewriter =
-        (e, v) -> ExpressionTreeRewriter.rewriteWith(expressionRewriterPlugin::process, e, v);
+    final BiFunction<Expression, SanitizerContext, Expression> expressionRewriter =
+        (e,v) -> ExpressionTreeRewriter.rewriteWith(expressionRewriterPlugin::process, e, v);
 
     return (Statement) new StatementRewriter<>(expressionRewriter, rewriterPlugin::process)
-        .rewrite(node, null);
+        .rewrite(node, new SanitizerContext());
   }
 
   private static final class RewriterPlugin extends
-      AstVisitor<Optional<AstNode>, StatementRewriter.Context<Object>> {
+      AstVisitor<Optional<AstNode>, StatementRewriter.Context<SanitizerContext>> {
 
     private final MetaStore metaStore;
     private final DataSourceExtractor dataSourceExtractor;
@@ -106,7 +109,7 @@ public final class AstSanitizer {
     @Override
     protected Optional<AstNode> visitInsertInto(
         final InsertInto node,
-        final StatementRewriter.Context<Object> ctx
+        final StatementRewriter.Context<SanitizerContext> ctx
     ) {
       final DataSource target = metaStore.getSource(node.getTarget());
       if (target == null) {
@@ -133,7 +136,7 @@ public final class AstSanitizer {
     @Override
     protected Optional<AstNode> visitSingleColumn(
         final SingleColumn singleColumn,
-        final StatementRewriter.Context<Object> ctx
+        final StatementRewriter.Context<SanitizerContext> ctx
     ) {
       if (singleColumn.getAlias().isPresent()) {
         return Optional.empty();
@@ -161,7 +164,8 @@ public final class AstSanitizer {
   }
 
   private static final class ExpressionRewriterPlugin extends
-      VisitParentExpressionVisitor<Optional<Expression>, Object> {
+      VisitParentExpressionVisitor<Optional<Expression>,
+      ExpressionTreeRewriter.Context<SanitizerContext>> {
 
     private final DataSourceExtractor dataSourceExtractor;
 
@@ -173,20 +177,14 @@ public final class AstSanitizer {
     @Override
     public Optional<Expression> visitUnqualifiedColumnReference(
         final UnqualifiedColumnReferenceExp expression,
-        final Object ctx
+        final ExpressionTreeRewriter.Context<SanitizerContext> ctx
     ) {
       final ColumnName columnName = expression.getColumnName();
-      if (ctx != null && ctx instanceof Context) {
-        final Context currentContext = (Context) ctx;
-        if (currentContext.getContext() instanceof LambdaContext) {
-          final LambdaContext lambdaContext =
-              (LambdaContext) currentContext.getContext();
-          if (lambdaContext.getLambdaArguments().size() > 0
-              && lambdaContext.getLambdaArguments().contains(columnName.text())) {
-            return Optional.of(new LambdaLiteral(columnName.text()));
-          }
-        }
+      if (ctx.getContext().getLambdaArgs().size() > 0
+          && ctx.getContext().getLambdaArgs().contains(columnName.text())) {
+        return Optional.of(new LambdaLiteral(columnName.text()));
       }
+
       final List<SourceName> sourceNames = dataSourceExtractor.getSourcesFor(columnName);
 
       if (sourceNames.size() > 1) {
@@ -209,19 +207,41 @@ public final class AstSanitizer {
 
     @Override
     public Optional<Expression> visitLambdaExpression(
-        final LambdaFunctionExpression expression,
-        final Object ctx
+        final LambdaFunctionCall expression,
+        final ExpressionTreeRewriter.Context<SanitizerContext> ctx
     ) {
       dataSourceExtractor.getAllSources().forEach(aliasedDataSource -> {
         for (String argument : expression.getArguments()) {
           if (aliasedDataSource.getDataSource().getSchema().columns().stream()
               .map(column -> column.name().text()).collect(Collectors.toList())
               .contains(argument)) {
-            throw new KsqlException("Lambda argument can't be a column name.");
+            throw new KsqlException(
+                String.format(
+                    "Lambda function argument can't be a column name: %s", argument));
           }
         }
       });
+
+      ctx.getContext().addLambdaArg(expression.getArguments());
       return visitExpression(expression, ctx);
+    }
+  }
+
+  private static class SanitizerContext {
+    final Set<String> lambdaArgs = new HashSet<>();
+
+    private void addLambdaArg(final List<String> newArguments) {
+      final int previousLambdaArgumentsLength = lambdaArgs.size();
+      lambdaArgs.addAll(newArguments);
+      if (new HashSet<>(lambdaArgs).size()
+          < previousLambdaArgumentsLength + 1) {
+        throw new KsqlException(
+            "Reusing lambda arguments in nested lambda is not allowed");
+      }
+    }
+
+    private Set<String> getLambdaArgs() {
+      return ImmutableSet.copyOf(lambdaArgs);
     }
   }
 }
--- a/ksqldb-engine/src/main/java/io/confluent/ksql/engine/rewrite/AstSanitizer.java
+++ b/ksqldb-engine/src/main/java/io/confluent/ksql/engine/rewrite/AstSanitizer.java
@@ -21,6 +21,8 @@ import io.confluent.ksql.analyzer.Analysis.AliasedDataSource;
 import io.confluent.ksql.engine.rewrite.ExpressionTreeRewriter.Context;
 import io.confluent.ksql.execution.expression.tree.ColumnReferenceExp;
 import io.confluent.ksql.execution.expression.tree.Expression;
+import io.confluent.ksql.execution.expression.tree.LambdaFunctionExpression;
+import io.confluent.ksql.execution.expression.tree.LambdaLiteral;
 import io.confluent.ksql.execution.expression.tree.QualifiedColumnReferenceExp;
 import io.confluent.ksql.execution.expression.tree.UnqualifiedColumnReferenceExp;
 import io.confluent.ksql.execution.expression.tree.VisitParentExpressionVisitor;
@@ -44,6 +46,7 @@ import io.confluent.ksql.util.UnknownSourceException;
 import java.util.List;
 import java.util.Optional;
 import java.util.function.BiFunction;
+import java.util.stream.Collectors;
 
 /**
  * Validate and clean ASTs generated from externally supplied statements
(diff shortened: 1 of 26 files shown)


ksqldb-engine/src/main/java/io/confluent/ksql/engine/rewrite/AstSanitizer.java
```diff
@@ -192,5 +206,22 @@ public static Statement sanitize(final Statement node, final MetaStore metaStore
...
+    public Optional<Expression> visitLambdaExpression(
+        final LambdaFunctionExpression expression,
+        final Object ctx
+    ) {
+      dataSourceExtractor.getAllSources().forEach(aliasedDataSource -> {
+        for (String argument : expression.getArguments()) {
+          if (aliasedDataSource.getDataSource().getSchema().columns().stream()
+              .map(column -> column.name().text()).collect(Collectors.toList())
+              .contains(argument)) {
+            throw new KsqlException("Lambda argument can't be a column name.");
```
guozhangwang, line 220: For my own education: could you elaborate the rationale of this verification? Maybe give me an example KSQL statement with lambda func?

ksqldb-engine/src/test/java/io/confluent/ksql/engine/rewrite/AstSanitizerTest.java
```diff
@@ -196,6 +203,35 @@ public void shouldAddQualifierForJoinColumnReferenceFromRight() {
     ))));
   }
 
+  @Test
+  public void shouldSanitizeLambdaArguments() {
+    // Given:
+    final Statement stmt = givenQuery(
+        "SELECT TRANSFORM_ARRAY(X => X + 5) FROM TEST2;");
```
guozhangwang, line 210: Is this a valid statement in practice? I.e. without a column reference of the entity `TEST`, how should we know which column to apply this lambda expression?

ksqldb-execution/src/main/java/io/confluent/ksql/execution/expression/tree/LambdaFunctionExpression.java
```diff
@@ -0,0 +1,91 @@
...
+    super(location);
+    this.arguments = ImmutableList.copyOf(requireNonNull(arguments, "arguments"));
+    if (arguments.size() == 0) {
+      throw new IllegalArgumentException(
+          String.format("Lambda expression must have at least 1 argument. => %s", body.toString()));
+    }
+    final Set<String> set = new HashSet<>(arguments);
+    if (set.size() < arguments.size()) {
+      throw new IllegalArgumentException(
+          String.format("Lambda arguments are duplicates: %s", arguments.toString()));
```
guozhangwang, line 56: nit: `have duplicates`.

ksqldb-parser/src/test/java/io/confluent/ksql/parser/AstBuilderTest.java
```diff
@@ -278,6 +284,34 @@ public void shouldIncludeSelectAliasIfPresent() {
     ))));
   }
 
+  @Test
+  public void shouldBuildLambdaFunction() {
+    // Given:
+    final SingleStatementContext stmt = givenQuery("SELECT TRANSFORM_ARRAY(Col4, X => X + 5) FROM TEST1;");
```
guozhangwang, line 290: For my own education: it seems that if Col is not specified, the sanitization phase would add a default column `COL_0` to it, is that right? If yes, when is the sanitization phase triggered? And how would `COL_0` be leverage in execution?

guozhangwang (commented):
Honestly I'm quite under-qualified yet to review this PR, since I'm still learning the AST code.. so I just left a few questions mainly for my own learning.

ksqldb-engine/src/test/java/io/confluent/ksql/engine/rewrite/AstSanitizerTest.java
stevenpyzhang: This particular one  wouldn't be valid.

We don't have to have a column reference though to apply the invocation function, we could have `Select Transform_Array(ARRAY[1, 2], x => x + 5) FROM TEST2 emit changes;`

guozhangwang: What does `ARRAY[1, 2]` mean here?

stevenpyzhang: https://docs.ksqldb.io/en/latest/developer-guide/syntax-reference/#array

It's how we create an inline array in the language. In this case we're not using any of the rows from the source and instead just using a constant array for use in the function

guozhangwang: I see. So in this case it is irrelevant to `TEST2`'s records, and would always return a single record as {key: null, value: [6,7]} right?

ksqldb-execution/src/main/java/io/confluent/ksql/execution/expression/tree/LambdaLiteral.java
```diff
@@ -0,0 +1,65 @@
...
+package io.confluent.ksql.execution.expression.tree;
+
+import com.google.errorprone.annotations.Immutable;
+import io.confluent.ksql.parser.NodeLocation;
+
+import java.util.Objects;
+import java.util.Optional;
+
+@Immutable
+public class LambdaLiteral extends Literal {
```
rodesai, line 25: question rather than suggestion - why do we need to use another type for these? An alternative would be to just use `UnqualifiedColumnReferenceExp` (and maybe rename it to `UnqualifiedReferenceExp`). Do you think it makes it easier to implement lambdas if they have different types?

ksqldb-engine/src/main/java/io/confluent/ksql/engine/rewrite/ExpressionTreeRewriter.java
```diff
@@ -465,6 +469,30 @@ public Expression visitCast(final Cast node, final C context) {
...
 
+    @Override
+    public Expression visitLambdaExpression(final LambdaFunctionExpression node, final C context) {
+      final Optional<Expression> result
+          = plugin.apply(node, new Context<>(context, this));
+      if (result.isPresent()) {
+        return result.get();
+      }
+
+      final LambdaContext lambdaContext =
```
rodesai, line 480: `LambdaContext` doesn't belong here - the rewriter shouldn't make any interpretation of the context. From this class's POV context is an opaque cookie to pass around to the rewriter so it can keep some internal state.

What's the intent of this LambdaContext stuff? is it just to screen for conflicts in variable names w/ enclosing variable names or column names? The better way to structure this would be to keep track of these in `AstSanitizer`. So from there you could do:

```
class AstSanitizer {
...
    private static final class ExpressionRewriterPlugin extends
      VisitParentExpressionVisitor<Optional<Expression>, Context<SanitizerContext>> {

          @Override
          public Optional<Expression> visitUnqualifiedColumnReference(
              final UnqualifiedColumnReferenceExp expression,
              final SanitizerContext ctx
          ) {
              final ColumnName columnName = expression.getColumnName();
              if (ctx.lambdaArgs().contains(columnName.text())) {
                  return Optional.of(new LambdaLiteral(columnName.text()));
              }
              ...
          }

          @Override
          public Optional<Expression> visitLambdaExpression(
              final LambdaFunctionExpression expression,
              final SanitizerContext ctx
           ) {
              dataSourceExtractor.getAllSources().forEach(aliasedDataSource -> {
              for (String argument : expression.getArguments()) {
                  if (aliasedDataSource.getDataSource().getSchema().columns().stream()
                      .map(column -> column.name().text()).collect(Collectors.toList())
                      .contains(argument)) {
                       throw new KsqlException("Lambda argument can't be a column name.");
                  }
                  if (ctx.lambdaArgs().contains(argument)) {
                       throw new KsqlException...
                  }
                  ctx.addLambdaArg(argument);
              }
              });
              return Optional.empty();  // just let the rewriter reconstruct the lambda - this function just validates and tracks args
          }
    }

    private static class SanitizerContext {
        Set<String> lambdaArgs = new HashSet<>();

        private void addLambdaArg(final String name) {
            reservedNames.add(name);
        }

        private Set<String> lambdaArgs() {
            return ImmutableSet.copyOf(lambdaArgs);
        }
    }
}
```

Then, this method should just become a dumb rewrite like everything else, and we can get rid of `LambdaContext`

rodesai (commented):
Thanks, @stevenpyzhang! Left a few comments/questions inline.

ksqldb-engine/src/test/java/io/confluent/ksql/engine/rewrite/AstSanitizerTest.java
stevenpyzhang: yup

ksqldb-parser/src/test/java/io/confluent/ksql/parser/AstBuilderTest.java
guozhangwang: Thanks! But I'm wondering, if `Col4` is not specified, i.e.

```
SELECT TRANSFORM_ARRAY(X => X + 5) FROM TEST1
```

Then the sanitizer would add a `COL_0` to the result AST, does that mean we default to the first column (rowkey?) if none is specified? I thought if it is not specified then the statement should fail.

ksqldb-execution/src/main/java/io/confluent/ksql/execution/expression/tree/LambdaFunctionExpression.java
```diff
@@ -0,0 +1,91 @@
...
+import io.confluent.ksql.parser.NodeLocation;
+
+import java.util.HashSet;
+import java.util.List;
+import java.util.Objects;
+import java.util.Optional;
+import java.util.Set;
+
+
+public class LambdaFunctionExpression extends Expression {
```
guozhangwang, line 30: Not a comment just a question regarding the class names: I saw the closest extension of `Expression` to `LambdaFunctionExpression` is called `FunctionCall`, which seems inconsistent in their naming. Is there any rationale behind this? Or should we either have:

* RegularFunctionCall, LambdaFunctionCall

Or:

* RegularFunctionExpression, LambdaFunctionException

ksqldb-engine/src/main/java/io/confluent/ksql/engine/rewrite/ExpressionTreeRewriter.java
guozhangwang: Sounds good to me. Then in https://github.com/confluentinc/ksql/pull/6868/files#diff-e5f9abbf3a2878a1dce09039136805764ecc31404a7eb27f2d1b49b9e038030dR81 we would just pass in a `new SanitizerContext()` instead of a `null`.

ksqldb-parser/src/main/java/io/confluent/ksql/parser/AstBuilder.java
```diff
@@ -1173,10 +1185,12 @@ public Node visitWhenClause(final SqlBaseParser.WhenClauseContext context) {
 
     @Override
     public Node visitFunctionCall(final SqlBaseParser.FunctionCallContext context) {
+      final List<Expression> expressionList = visit(context.expression(), Expression.class);
+      expressionList.addAll(visit(context.lambdaFunction(), Expression.class));
       return new FunctionCall(
```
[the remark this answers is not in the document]
stevenpyzhang: I'm going to update the g4 so that all the lambda functions have to be after all the other function arguments
```
    | identifier'(' (expression (',' expression)* (',' lambdaFunction)*)? ')'             #functionCall
```
So then this assumption can be made soundly

ksqldb-execution/src/main/java/io/confluent/ksql/execution/expression/tree/LambdaLiteral.java
stevenpyzhang: I think when implementing the type reference, it makes the code cleaner. I'd imagine when doing type checking in `ExpressionTypeManager`, instead of having to pass down in the context (similar to what's happening in AstSanitizer) that we're in a lambda function expression, we can just have a specific visitor for a LambdaLiteral.

ksqldb-parser/src/test/java/io/confluent/ksql/parser/AstBuilderTest.java
stevenpyzhang: the `KSQL_COL_0` refers to the name of the column in the output of the query. We need to have a column name for each column and if the user didn't specify what the column name should be, we should generate a unique name for that output column

```
ksql> select 5,3 as TEST,6,time from KSQL_PROCESSING_LOG emit changes;

+----------------------------------+----------------------------------+----------------------------------+----------------------------------+
|KSQL_COL_0                        |TEST                              |KSQL_COL_1                        |TIME                              |
+----------------------------------+----------------------------------+----------------------------------+----------------------------------+

```

ksqldb-engine/src/main/java/io/confluent/ksql/engine/rewrite/ExpressionTreeRewriter.java
rodesai: I suppose we could, though tracking the lambda args seems specific to `AstSanitizer` at the moment.

ksqldb-execution/src/main/java/io/confluent/ksql/execution/expression/tree/LambdaFunctionExpression.java
stevenpyzhang: I changed LambdaFunctionExpression to LambdaFunctionCall

ksqldb-execution/src/main/java/io/confluent/ksql/execution/expression/tree/LambdaLiteral.java
lct45: After working through some of the type stuff, I do think it's somewhat necessary to have the additional type. Right now, the array/map input is passed in as an `UnqualifiedReferenceExp` and it's nice to be able to distinguish that from the lambda inputs

ksqldb-parser/src/main/antlr4/io/confluent/ksql/parser/SqlBase.g4
```diff
@@ -292,13 +292,11 @@ primaryExpression
...
     | CAST '(' expression AS type ')'                                                     #cast
-    | identifier '=>' expression                                                          #lambda
-    | '(' identifier (',' identifier)*  ')' '=>' expression                               #lambda
     | ARRAY '[' (expression (',' expression)*)? ']'                                       #arrayConstructor
     | MAP '(' (expression ASSIGN expression (',' expression ASSIGN expression)*)? ')'     #mapConstructor
     | STRUCT '(' (identifier ASSIGN expression (',' identifier ASSIGN expression)*)? ')'  #structConstructor
-    | identifier '(' ASTERISK ')'                              		                      #functionCall
-    | identifier'(' (expression (',' expression)*)? ')' 						          #functionCall
+    | identifier '(' ASTERISK ')'                              		                        #functionCall
+    | identifier'(' (expression (',' expression)* (',' lambdaFunction)*)? ')'             #functionCall
```
guozhangwang, line 299: nit: space after `identifier`.

ksqldb-parser/src/main/java/io/confluent/ksql/parser/AstBuilder.java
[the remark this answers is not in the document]
guozhangwang: Sounds good to me.

ksqldb-parser/src/test/java/io/confluent/ksql/parser/AstBuilderTest.java
guozhangwang: Got it.

ksqldb-parser/src/test/java/io/confluent/ksql/parser/AstBuilderTest.java
```diff
@@ -285,25 +288,43 @@ public void shouldIncludeSelectAliasIfPresent() {
...
+                            column("X"),
+                            new IntegerLiteral(5))
+                  )
+                )
             ),
             Optional.empty())
     ))));
   }
 
+  @Test
```
guozhangwang, line 316: Since in g4 we allow multiple expressions followed by multiple lambda functions, could we also add a test for the error case where number of expressions is not equal to the number of lambda functions?

guozhangwang (approved):
Thanks for the new commit. I just have a few minor comments left, otherwise LGTM.

rodesai (approved):
LGTM!

ksqldb-parser/src/test/java/io/confluent/ksql/parser/AstBuilderTest.java
stevenpyzhang: I don't think we can make this restriction, not all invocation functions would fit this requirement. From KLIP-30
```
reduce_map(map, s, (k, v, s) => s) - Reduces the input Map down to a single value. s is the initial state and is passed into the scope of the lambda function. Each invocation returns a new value for s, which the next invocation will receive. reduce_map will return the final value of s.
```
^ this one wouldn't fit this restriction

guozhangwang: Got it, thanks.

PR #6872, opened 2021-01-15
feat: display precision and scale when describing DECIMAL columns

### LINKED ISSUE #6483: Left-join does not return matching rows the first time the query is issued
**Describe the bug**
A left-join between a stream and a table returns a non-matching row (for the side of the table, i.e nulls) although the natural-join finds a match. The second time I issue the LJ query, it finds a match.

**To Reproduce**
1. The version of KSQL.
Master-ish
2. Sample source data.
3. Any SQL statements you ran
```
create table account_detail (ref int primary key, mdate int) with (kafka_topic='account', value_format='json', partitions=1);
insert into orders (id, ref) values (123, 34);
```

```
create table account_detail (ref int primary key, mdate int) with (kafka_topic='account', value_format='json', partitions=1);
insert into account_detail(ref, mdate) values (34, 1002);
insert into account_detail(ref, mdate) values (36, 1007);
```

Natural join query:
```
select a.id, a.ref, b.ref, ifnull(cast(b.mdate as string), 'No Date') from orders a join account_detail b on a.ref=b.ref emit changes;
+-------------------------------+-------------------------------+-------------------------------+-------------------------------+
|ID                             |A_REF                          |B_REF                          |KSQL_COL_0                     |
+-------------------------------+-------------------------------+-------------------------------+-------------------------------+
|123                            |34                             |34                             |1002                           |
^CQuery terminated
```

First LJ query:
```
select a.id, a.ref, b.ref, ifnull(cast(b.mdate as string), 'No Date') from orders a left join account_detail b on a.ref=b.ref emit changes;
+-------------------------------+-------------------------------+-------------------------------+-------------------------------+
|ID                             |A_REF                          |B_REF                          |KSQL_COL_0                     |
+-------------------------------+-------------------------------+-------------------------------+-------------------------------+
|123                            |34                             |null                           |No Date                        |
^CQuery terminated
```

Second LJ query:
```
select a.id, a.ref, b.ref, ifnull(cast(b.mdate as string), 'No Date') from orders a left join account_detail b on a.ref=b.ref emit changes;
+-------------------------------+-------------------------------+-------------------------------+-------------------------------+
|ID                             |A_REF                          |B_REF                          |KSQL_COL_0                     |
+-------------------------------+-------------------------------+-------------------------------+-------------------------------+
|123                            |34                             |34                             |1002                           |
^CQuery terminated
```

**Expected behavior**
The LJ query should return the matching row the first time the query is issued.

### LINKED ISSUE #6843: Show decimal field precision/scale in DESCRIBE output
When describing a stream in ksqlDB, any decimal fields just display as type `DECIMAL`, with no precision or scale attributes. For example:

```
ksql> describe s;

Name                 : S
 Field | Type
-----------------
 VAL   | DECIMAL
-----------------
```

Display the defined precision and scale for the decimal type. For example:

```
 Field | Type
----------------------
 VAL   | DECIMAL(5,2)
----------------------
```

`DESCRIBE EXTENDED` can sometimes help by showing the original CS/CT statement (which includes precision and scale for decimal fields), but not if the stream was created from a query.

swist:
### Description 
Fixes #6843 
We want DESCRIBE queries to show the precision and scale for DECIMAL columns

It seems to me that maybe it would make sense to carry instances of
SqlBaseType on SchemaInfo instances somehow so that we could delegate
these formatting duties to them?

### Testing done 
- new unit tests added

### Reviewer checklist

### DIFF
--- a/ksqldb-cli/src/test/java/io/confluent/ksql/cli/console/ConsoleTest.java
+++ b/ksqldb-cli/src/test/java/io/confluent/ksql/cli/console/ConsoleTest.java
@@ -1001,7 +1001,13 @@ public class ConsoleTest {
                 SqlBaseType.STRUCT,
                 ImmutableList.of(
                     new FieldInfo("f1", new SchemaInfo(SqlBaseType.STRING, null, null), Optional.empty())),
-                null)
+                null),
+            "typeC", new SchemaInfo(
+                    SqlBaseType.DECIMAL,
+                    null,
+                    null,
+                    ImmutableMap.of("precision", 10, "scale", 9)
+                )
         ))
     ));
 
@@ -1035,6 +1041,15 @@ public class ConsoleTest {
           + "        }" + NEWLINE
           + "      } ]," + NEWLINE
           + "      \"memberSchema\" : null" + NEWLINE
+          + "    }," + NEWLINE
+          + "    \"typeC\" : {" + NEWLINE
+          + "      \"type\" : \"DECIMAL\"," + NEWLINE
+          + "      \"fields\" : null," + NEWLINE
+          + "      \"memberSchema\" : null," + NEWLINE
+          + "      \"parameters\" : {" + NEWLINE
+          + "        \"precision\" : 10," + NEWLINE
+          + "        \"scale\" : 9" + NEWLINE
+          + "      }" + NEWLINE
           + "    }" + NEWLINE
           + "  }," + NEWLINE
           + "  \"warnings\" : [ ]" + NEWLINE
@@ -1045,6 +1060,7 @@ public class ConsoleTest {
           + "----------------------------------------" + NEWLINE
           + " typeA     | STRUCT<f1 VARCHAR(STRING)> " + NEWLINE
           + " typeB     | ARRAY<VARCHAR(STRING)>     " + NEWLINE
+          + " typeC     | DECIMAL(10, 9)             " + NEWLINE
           + "----------------------------------------" + NEWLINE));
     }
   }
--- a/ksqldb-rest-app/src/test/java/io/confluent/ksql/rest/util/EntityUtilTest.java
+++ b/ksqldb-rest-app/src/test/java/io/confluent/ksql/rest/util/EntityUtilTest.java
@@ -23,6 +23,7 @@ import io.confluent.ksql.name.ColumnName;
 import io.confluent.ksql.rest.entity.FieldInfo;
 import io.confluent.ksql.rest.entity.FieldInfo.FieldType;
 import io.confluent.ksql.schema.ksql.LogicalSchema;
+import io.confluent.ksql.schema.ksql.types.SqlDecimal;
 import io.confluent.ksql.schema.ksql.types.SqlType;
 import io.confluent.ksql.schema.ksql.types.SqlTypes;
 import java.util.List;
@@ -75,6 +76,27 @@ public class EntityUtilTest {
     assertThat(fields.get(0).getSchema().getMemberSchema().get().getTypeName(),
         equalTo("INTEGER"));
   }
+
+  @Test
+  public void shouldBuildCorrectDecimalField() {
+    // Given:
+    final SqlDecimal decimal  = SqlTypes.decimal(10, 9);
+    final LogicalSchema schema = LogicalSchema.builder()
+            .valueColumn(ColumnName.of("field"), decimal)
+            .build();
+
+    // When:
+    final List<FieldInfo> fields = EntityUtil.buildSourceSchemaEntity(schema);
+
+    // Then:
+    assertThat(fields, hasSize(1));
+    assertThat(fields.get(0).getName(), equalTo("field"));
+    assertThat(fields.get(0).getSchema().getTypeName(), equalTo("DECIMAL"));
+    assertThat(fields.get(0).getSchema().getFields(), equalTo(Optional.empty()));
+    assertThat(fields.get(0).getSchema().getParameters().get(SqlDecimal.SCALE), equalTo(decimal.getScale()));
+    assertThat(fields.get(0).getSchema().getParameters().get(SqlDecimal.PRECISION), equalTo(decimal.getPrecision()));
+  }
+
 
   @Test
   public void shouldBuildCorrectArrayField() {
--- a/ksqldb-rest-model/pom.xml
+++ b/ksqldb-rest-model/pom.xml
@@ -59,6 +59,10 @@
             <version>${io.confluent.ksql.version}</version>
             <scope>test</scope>
         </dependency>
+        <dependency>
+            <groupId>com.fasterxml.jackson.datatype</groupId>
+            <artifactId>jackson-datatype-guava</artifactId>
+        </dependency>
     </dependencies>
 
     <build>
--- a/ksqldb-rest-model/src/main/java/io/confluent/ksql/rest/entity/SchemaInfo.java
+++ b/ksqldb-rest-model/src/main/java/io/confluent/ksql/rest/entity/SchemaInfo.java
@@ -17,12 +17,16 @@ package io.confluent.ksql.rest.entity;
 
 import com.fasterxml.jackson.annotation.JsonCreator;
 import com.fasterxml.jackson.annotation.JsonIgnoreProperties;
+import com.fasterxml.jackson.annotation.JsonInclude;
 import com.fasterxml.jackson.annotation.JsonProperty;
 import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableMap;
 import com.google.errorprone.annotations.Immutable;
 import io.confluent.ksql.schema.ksql.types.SqlBaseType;
+import io.confluent.ksql.schema.ksql.types.SqlDecimal;
 import io.confluent.ksql.testing.EffectivelyImmutable;
+import io.confluent.ksql.util.KsqlPreconditions;
+
 import java.util.List;
 import java.util.Objects;
 import java.util.Optional;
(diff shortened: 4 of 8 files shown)


ksqldb-rest-model/src/main/java/io/confluent/ksql/rest/entity/SchemaInfo.java
```diff
@@ -105,6 +128,21 @@ public int hashCode() {
                   .stream()
                   .map(f -> f.getName() + " " + f.getSchema().toTypeString())
                   .collect(Collectors.joining(", ", SqlBaseType.STRUCT + "<", ">")))
+          .put(
+              SqlBaseType.DECIMAL,
+              si -> {
+                final Object precision = si.getFieldProperties().get("precision");
+                final Object scale = si.getFieldProperties().get("scale");
+                String parameterString = "";
```
agavra, line 136: consider throwing an exception if exactly one of precision/scale is present (that should be a bug) - and add a test for the behavior as well

ksqldb-rest-model/src/main/java/io/confluent/ksql/rest/entity/SchemaInfo.java
```diff
@@ -36,6 +37,23 @@
   private final SqlBaseType type;
   private final ImmutableList<FieldInfo> fields;
   private final SchemaInfo memberSchema;
+  private final Map<String, Object> fieldProperties;
+
+  @JsonCreator
+  public SchemaInfo(
```
agavra, line 43: Can you also update `ConsoleTest#shouldPrintTypesList` to include a field that is a decimal? that will give us more confidence in what the user ends up seeing.

ksqldb-rest-model/src/main/java/io/confluent/ksql/rest/entity/SchemaInfo.java
```diff
@@ -105,6 +128,21 @@ public int hashCode() {
...
+                final Object scale = si.getFieldProperties().get("scale");
+                String parameterString = "";
+                if (precision != null) {
+                  parameterString += precision.toString();
+                  parameterString += ", ";
+                }
+                if (scale != null) {
+                  parameterString += scale.toString();
+                }
+                return SqlBaseType.DECIMAL + "(" + parameterString  + ")";
```
agavra, line 144: You asked for java-isms and you shall receive: 
```java
KsqlPreconditions.checkArgument(
  !(precision == null ^ scale == null), 
  "Unexpected Decimal SchemaInfo with only one of precision and scale: " + new ObjectMapper().writeValueAsString(si)
);

final String precisionString = (scale != null) 
  ? String.format("(%s, %s)", precision, scale) 
  : "";
```
Note that `%s` will format anything as a string.

ksqldb-rest-app/src/test/java/io/confluent/ksql/rest/util/EntityUtilTest.java
```diff
@@ -76,6 +77,27 @@ public void shouldBuildCorrectMapField() {
         equalTo("INTEGER"));
   }
 
+  @Test
+  public void shouldBuildCorrectDecimalField() {
+    // Given:
```
agavra, line 82: nice! you picked up on our Given/When/Then testing pattern quickly 😂

ksqldb-rest-model/src/main/java/io/confluent/ksql/rest/entity/SchemaInfo.java
```diff
@@ -48,6 +66,7 @@ public SchemaInfo(
         ? null
         : ImmutableList.copyOf(fields);
     this.memberSchema = memberSchema;
+    this.fieldProperties = ImmutableMap.of();
```
agavra, line 69: nit: if we have two constructors, it's better to have one call the other:
```java
this(type, fields, memberSchema, ImmutableMap.of();
```
That way the two won't ever get out of sync

ksqldb-rest-model/src/main/java/io/confluent/ksql/rest/entity/SchemaInfo.java
```diff
@@ -36,6 +37,23 @@
   private final SqlBaseType type;
   private final ImmutableList<FieldInfo> fields;
   private final SchemaInfo memberSchema;
+  private final Map<String, Object> fieldProperties;
+
+  @JsonCreator
```
agavra, line 42: according to https://stackoverflow.com/a/15931483/2258040 you can only have one constructor with `@JsonCreator` annotated. I think it should be this one - but we should test to see what happens if we try to deserialize a JSON object without `fieldProperties`. I'm pretty sure it'll work (it'll create an empty map) but if not we need to make `fieldProperties` optional.

agavra (commented):
Thanks @swist! I think this is a great first pass at the PR - no major comments inline, some recommendations about testing and java-isms.

In general, any PR that changes something that is visible to the user in the CLI should have a copy-paste terminal session in the "testing done" so that we can see what it would look like from the user perspective. Something like:
```
ksql> describe s;

 Name                 : S
 Field | Type
-----------------
 VAL   | DECIMAL(5, 2)
-----------------
```

ksqldb-rest-model/src/test/java/io/confluent/ksql/rest/entity/SchemaInfoTest.java
```diff
@@ -0,0 +1,24 @@
...
+import io.confluent.ksql.schema.ksql.types.SqlBaseType;
+import org.junit.Test;
+
+import static org.hamcrest.MatcherAssert.assertThat;
+import static org.hamcrest.Matchers.equalTo;
+
+public class SchemaInfoTest {
+
+  @Test
+  public void shouldCorrectlyFormatDecimals() {
```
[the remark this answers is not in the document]
swist: I wasn't able to cast a Decimal column without either, was going to ask you about this? It might be because the factory I've found doesn't implement these as optional? I'll see about passing the nulls

ksqldb-rest-model/src/main/java/io/confluent/ksql/rest/entity/SchemaInfo.java
swist: Ah couldn't find the right test on the other end, nice

ksqldb-rest-model/src/main/java/io/confluent/ksql/rest/entity/SchemaInfo.java
swist: Yeah ConsoleTest proves it's good

ksqldb-rest-model/src/main/java/io/confluent/ksql/rest/entity/SchemaInfo.java
swist: Apparently all constructors require both!

ksqldb-rest-app/src/test/java/io/confluent/ksql/rest/util/EntityUtilTest.java
swist: Big fan of bdd

ksqldb-cli/src/test/java/io/confluent/ksql/cli/console/ConsoleTest.java
```diff
@@ -1021,8 +1027,10 @@ public void shouldPrintTypesList() {
           + "      \"memberSchema\" : {" + NEWLINE
           + "        \"type\" : \"STRING\"," + NEWLINE
           + "        \"fields\" : null," + NEWLINE
-          + "        \"memberSchema\" : null" + NEWLINE
-          + "      }" + NEWLINE
+          + "        \"memberSchema\" : null," + NEWLINE
+          + "        \"parameters\" : { }" + NEWLINE
```
swist, line 1031: Is there a less crappy way of validating json returns? Load from file? (would also be good if it was order invariant, but i'm guessing in Java it will be fixed wrt to order in which the AST is walked?)

ksqldb-rest-model/src/test/java/io/confluent/ksql/rest/model/ImmutabilityTest.java
```diff
@@ -75,6 +75,7 @@ public void shouldBeImmutable() {
         .withKnownImmutableType(KGroupedStream.class)
         .withKnownImmutableType(KGroupedTable.class)
         .withKnownImmutableType(Serde.class)
+        .withKnownImmutableType(Number.class)
```
swist, line 78: I had to tell the test that we believe `Numbers` to be immutable because otherwise it wouldn't like the fact that we have an`ImmutableMap<String,Number>` to contain the params. Had to change it from `ImmutableMap<String, Object>` because that seemed too generic to make known immutable here. @agavra what do you think?

ksqldb-rest-model/pom.xml
```diff
@@ -59,6 +59,11 @@
             <version>${io.confluent.ksql.version}</version>
             <scope>test</scope>
         </dependency>
+        <dependency>
+            <groupId>com.fasterxml.jackson.datatype</groupId>
+            <artifactId>jackson-datatype-guava</artifactId>
+            <version>2.10.5</version>
```
swist, line 65: This adds support for guava serdes in jackson

ksqldb-rest-model/src/test/java/io/confluent/ksql/rest/model/ImmutabilityTest.java
agavra: I think we should keep it as `<String, Object>` and instead mark it as `@EffectivelyImmutable` so that it remains generic (in the future, we may have parameters like "structSchemaName" that may not be numbers). You can see that being used (for example) here: https://github.com/confluentinc/ksql/blob/3e48540fbe1a02ed02faa667e424b225f9d7ba20/ksqldb-udf/src/main/java/io/confluent/ksql/types/KsqlStruct.java#L42-L43

ksqldb-cli/src/test/java/io/confluent/ksql/cli/console/ConsoleTest.java
agavra: a less crappy way would also be when we eventually upgrade to java 13 to get multiline strings https://openjdk.java.net/jeps/355

agavra (approved):
LGTM! Some minor replies left on the previous comments, and I'm not sure I understand why we need the GuavaModule

ksqldb-rest-model/pom.xml
swist: We need the guava series because jackson otherwise doesn't understand how to serde an `ImmutableMap`

agavra: Can you get around that by declaring it as a `Map` (not `ImmutableMap` and just in the body of the method wrap it with `ImmutableMap.copyOf()`? In general, it's better practice to use interfaces in the fields/public accessors than the implementation class

ksqldb-rest-model/src/test/java/io/confluent/ksql/rest/model/ImmutabilityTest.java
swist: Done

swist: Done!

ksqldb-rest-model/pom.xml
agavra: Unresolved this one as I feel that it may still be relevant.

swist: Ow shoot, only saw it after I pressed the big green button. I don't remember exactly, but I think it would work now that you've showed me how to mark it as @EffectivelyImmutable. Do you want me to do a follow up PR?

agavra: no worries - I think a quick follow-up PR would be ideal. fewer dependencies is a good thing for build speed and package size :)

PR #6876, opened 2021-01-20
fix: fix how the buffer limit check evaluates streams config

rodesai:
The buffer limit check first creates a dummy streams config to get the current
max bytes buffering value. To do this it creates a map of the streams props and
creates the config (with dummy values for all the required fields). However
sometimes those fields are already provided and ImmutableMap doesn't like
overwrites in the builder. So, use a plain old HashMap.

### DIFF
--- a/config/ksql-server.properties
+++ b/config/ksql-server.properties
@@ -20,7 +20,8 @@
 # The default is any IPv4 interface on the machine.
 # NOTE: If set to wildcard or loopback set 'advertised.listener' to enable pull queries across machines
 listeners=http://0.0.0.0:8088
-
+ksql.service.id=test
+ksql.query.persistent.max.bytes.buffering.total=15000000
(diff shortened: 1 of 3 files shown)


config/ksql-server.properties
```diff
@@ -20,7 +20,8 @@
 # The default is any IPv4 interface on the machine.
 # NOTE: If set to wildcard or loopback set 'advertised.listener' to enable pull queries across machines
 listeners=http://0.0.0.0:8088
-
+ksql.service.id=test
```
vcrfxia, line 23: This does not look like it's meant to be checked in.

vcrfxia (approved):
LGTM besides the comment inline. Thanks @rodesai !

config/ksql-server.properties
rodesai: 😞 thanks

PR #6894, opened 2021-01-25
docs: update contribtuing.md to ensure that we don't squash commits (minor)

### LINKED ISSUE #6882: changelog generation is not able to handle squashed merged
Hit this when trying to generate the changelog for 0.15. When we released 0.14 we merged a bunch of changes from master into the 6.1.x and the 0.14 release branch. These changes all got squashed into one commit. Our changelog generation was subsequently not able to generate the correct changelog for 0.15 - it included all those 0.14 changes in 0.15's changelog.

I think I understand why. IIUC, the changelog generation works like this:
- use `conventional-changelog` to generate a changelog for the previous release
- use `conventional-changelog` to generate a changelog for the current release
- run `process_commit_files.py`. This script finds all the commits in the current release that aren't in the previous release. However since in this case those commits were squashed, they don't show up in the previous release's changelog and get added to the current release's

I'm not sure exactly how to fix this. It might be enough to just be careful not to squash the commits away. Another alternative might be to check the current changelog (generated by `conventional-changelog`) against the CHANGELOG.md on the release branch and remove any commits that are already in that changelog (assuming the RM for the release with squashed commits did the right thing and added those commits to the changelog manually)

agavra:
fixes #6882

### Description 

### Testing done 

### Reviewer checklist

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -252,3 +252,9 @@ Once enabled, commitlint will reject commits with improperly formatted commit me
    git push origin --force feature-xxx
    ```
 
+### Backporting Commits
+
+There might be times when a certain commit needs to be backported to a previous ksqlDB release (either community
+edition or confluent edition). In these situations, cherry-pick individual commits to the previous
+branches (do _not_ squash multiple commits and move them as one because our changelog generation tool
+does not handle squashed commits properly).


stevenpyzhang (approved):
LGTM

PR #6895, opened 2021-01-25
fix: Fixes auth for forwarded requests for pull queries

AlanConfluent:
### Description 
The PR https://github.com/confluentinc/ksql/pull/6665 made `HARouting` a singleton.  Unfortunately, it also kept a reference to the service context rather than using the user-aware context.  What this meant is that all forward requests did not provide authentication and failed.  This didn't fail in testing because it requires more than two nodes to expose the forwarding failure, since HA makes the request fall back to the fowarder itself.  For that reason, I made the routing functional test use auth, which does test the three node case and was able to expose the bug.

### Testing done 
`PullQueryRoutingFunctionalTest` with auth.

### Reviewer checklist

### DIFF
--- a/ksqldb-rest-app/src/test/java/io/confluent/ksql/rest/integration/HighAvailabilityTestUtil.java
+++ b/ksqldb-rest-app/src/test/java/io/confluent/ksql/rest/integration/HighAvailabilityTestUtil.java
@@ -140,8 +140,16 @@ class HighAvailabilityTestUtil {
   static void waitForStreamsMetadataToInitialize(
       final TestKsqlRestApp restApp, List<KsqlHostInfoEntity> hosts, String queryId
   ) {
+    waitForStreamsMetadataToInitialize(restApp, hosts, queryId, Optional.empty());
+  }
+
+  static void waitForStreamsMetadataToInitialize(
+      final TestKsqlRestApp restApp, List<KsqlHostInfoEntity> hosts, String queryId,
+      final Optional<BasicCredentials> credentials
+  ) {
     while (true) {
-      ClusterStatusResponse clusterStatusResponse = HighAvailabilityTestUtil.sendClusterStatusRequest(restApp);
+      ClusterStatusResponse clusterStatusResponse
+          = HighAvailabilityTestUtil.sendClusterStatusRequest(restApp, credentials);
       List<KsqlHostInfoEntity> initialized = hosts.stream()
           .filter(hostInfo -> Optional.ofNullable(
               clusterStatusResponse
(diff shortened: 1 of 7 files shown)


ksqldb-rest-app/src/test/java/io/confluent/ksql/rest/integration/HighAvailabilityTestUtil.java
```diff
@@ -392,11 +400,24 @@ public static void makeAdminRequest(TestKsqlRestApp restApp, final String sql) {
     RestIntegrationTestUtil.makeKsqlRequest(restApp, sql, Optional.empty());
   }
 
+  public static void makeAdminRequest(
```
guozhangwang, line 403: I'm curious why we have a `makeAdminRequest` and `makeAdminRequestWithResponse` here? Both of them call the same callee. Could we just have one `makeAdminRequest` that always returns (a potential empty list), and its callers can just ignore the returned value or not?

guozhangwang (commented):
Just a very nit question, otherwise LGTM!

ksqldb-rest-app/src/test/java/io/confluent/ksql/rest/integration/PullQueryRoutingFunctionalTest.java
```diff
@@ -122,6 +129,19 @@
       SerdeFeatures.of()
   );
 
+  private static final String PROPS_JAAS_REALM = "KsqlServer-Props";
+  private static final String KSQL_RESOURCE = "ksql-user";
+  private static final String USER_WITH_ACCESS = "harry";
```
agavra, line 134: do we have a test trying out a user with no access? that will make sure that we haven't accidentally just changed it to allow all users 😆

agavra (approved):
Great detective work figuring this out!

ksqldb-rest-app/src/test/java/io/confluent/ksql/rest/integration/PullQueryRoutingFunctionalTest.java
AlanConfluent: We do have tests elsewhere that test auth for pull queries and check the negative case.

ksqldb-rest-app/src/test/java/io/confluent/ksql/rest/integration/HighAvailabilityTestUtil.java
AlanConfluent: Sorry, just noticed your question.  I don't think there's a reason to have two separate methods.  I agree with you.  I'll do a quick followup PR to fix this.

PR #6905, opened 2021-01-26
Ksql 5716/close transient queries when network splits

swist:
### Description 
We've observed a behaviour where if there are two concurrent PushQueries running against a KSQL cluster, upon terminating the websocket in the browser, one of the queries lingers.

### Testing done 
This is a right pain to test because everything useful gets mocked, so I'm more than happy to take pointers about testing this unit-test wise. I have tested that the behaviour is now correct locally by doing the following:

1. deploy ksqldb, c3 and confluent platform
2. set up a socat relay between the frontend and ksqldb 
3. modify frontend to point at the relay instead of ksqldb
4. open two queries in separate tabs
5. kill socat to simulate a network split

Prior to the change one of the queries would always stay when running `show queries` from another client. After the change it does not. Body of commit 12fb072 has a decent explanation why this might be the case

### Reviewer checklist

### DIFF
--- a/ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/resources/streaming/WSQueryEndpoint.java
+++ b/ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/resources/streaming/WSQueryEndpoint.java
@@ -81,8 +81,6 @@ public class WSQueryEndpoint {
   private final HARouting routing;
   private final Optional<LocalCommands> localCommands;
 
-  private WebSocketSubscriber<?> subscriber;
-
   // CHECKSTYLE_RULES.OFF: ParameterNumberCheck
   public WSQueryEndpoint(
       // CHECKSTYLE_RULES.ON: ParameterNumberCheck
@@ -220,11 +218,6 @@ public class WSQueryEndpoint {
         throw new IllegalArgumentException("Unexpected statement type " + statement);
       }
 
-      webSocket.closeHandler(v -> {
-        if (subscriber != null) {
-          subscriber.close();
-        }
-      });
     } catch (final TopicAuthorizationException e) {
       log.debug("Error processing request", e);
       SessionUtil.closeSilently(
@@ -279,16 +272,29 @@ public class WSQueryEndpoint {
     }
   }
 
+  private void attachCloseHandler(final ServerWebSocket websocket,
+                                  final WebSocketSubscriber<?> subscriber) {
+    websocket.closeHandler(v -> {
+      if (subscriber != null) {
+        subscriber.close();
+        log.debug("Websocket {} closed, reason: {},  code: {}",
+                websocket.textHandlerID(),
+                websocket.closeReason(),
+                websocket.closeStatusCode());
+      }
+    });
+  }
+
   private void handleQuery(final RequestContext info, final Query query,
       final long startTimeNanos) {
     final Map<String, Object> clientLocalProperties = info.request.getConfigOverrides();
 
     final WebSocketSubscriber<StreamedRow> streamSubscriber =
         new WebSocketSubscriber<>(info.websocket);
-    this.subscriber = streamSubscriber;
 
-    final PreparedStatement<Query> statement =
-        PreparedStatement.of(info.request.getKsql(), query);
+    attachCloseHandler(info.websocket, streamSubscriber);
+
+    final PreparedStatement<Query> statement = PreparedStatement.of(info.request.getKsql(), query);
 
     final ConfiguredStatement<Query> configured = ConfiguredStatement
         .of(statement, SessionConfig.of(ksqlConfig, clientLocalProperties));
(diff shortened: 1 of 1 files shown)


ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/resources/streaming/WebSocketSubscriber.java
```diff
@@ -75,12 +75,13 @@ public void onError(final Throwable e) {
     final String msg = e.getMessage() == null || e.getMessage().trim().isEmpty()
         ? "KSQL exception: " + e.getClass().getSimpleName()
         : e.getMessage();
-
+    close();
```
colinhicks, line 78: Since `SessionUtil.closeSilently()` invokes `close()`, this seems unnecessary. It also would send the client the default close frame code, rather than the specific code currently passed to `closeSilently`. Same below in `onComplete`.

ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/resources/streaming/WSQueryEndpoint.java
```diff
@@ -220,11 +218,6 @@ public void executeStreamQuery(final ServerWebSocket webSocket, final MultiMap r
         throw new IllegalArgumentException("Unexpected statement type " + statement);
       }
 
-      webSocket.closeHandler(v -> {
-        if (subscriber != null) {
-          subscriber.close();
-        }
-      });
```
colinhicks, line 227: Don't we also need `attachCloseHandler(webSocket, subscriber);` here, too?

colinhicks: never mind, I misread the logic.

colinhicks:
@swist is it feasible to add test coverage in `io.confluent.ksql.rest.server.resources.WSQueryEndpointTest`? Something like calling `executeStreamQuery` twice, attaching a test close handler to each. Then, close the second websocket before the first, and assert that both close handlers were called?

ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/resources/streaming/WebSocketSubscriber.java
swist: Done!

ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/resources/streaming/WSQueryEndpoint.java
```diff
@@ -279,16 +272,28 @@ private KsqlRequest parseRequest(final MultiMap requestParams) {
...
 
+  private void attachCloseHandler(final ServerWebSocket websocket,
+                                  final WebSocketSubscriber<?> subscriber) {
+    websocket.closeHandler(v -> {
+      subscriber.close();
+      final String msg = String.format("Websocket %s closed, reason: %s,  code: %s)",
+              websocket.textHandlerID(),
+              websocket.closeReason(),
+              websocket.closeStatusCode());
+      log.debug(msg);
```
spena, line 283: I recommend using the `log.debug` parameters to pass variables to the string. Like:
```
log.debug("Websocket {} closed, reason: {},  code: {})", 
        websocket.textHandlerID(), 
        websocket.closeReason(), 
        websocket.closeStatusCode());
```

Passing the variables to the LOG instead of using the String.format() has the benefit that if DEBUG is disabled (in most cases), then the string will not be formatted. However, if using String.format(), then the string is always formatted even if the log doesn't print it because debug is disabled.

ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/resources/streaming/WSQueryEndpoint.java
```diff
@@ -279,16 +272,28 @@ private KsqlRequest parseRequest(final MultiMap requestParams) {
     }
   }
 
+  private void attachCloseHandler(final ServerWebSocket websocket,
+                                  final WebSocketSubscriber<?> subscriber) {
+    websocket.closeHandler(v -> {
+      subscriber.close();
```
spena, line 278: can subscriber be null? I see that validation in the code you removed. Should it be checked before creating the closeHandler?

vcrfxia (commented):
Thanks for the fix @swist ! LGTM pending the comments Colin and Sergio have left.

ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/resources/streaming/WSQueryEndpoint.java
swist: There isn't really a path that makes it null, but it makes sense to be more defensive there, I'll bring back the check!

ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/resources/streaming/WSQueryEndpoint.java
swist: Ah yes, was looking for something like this!

spena (approved):
lgtm

swist:
I'm going to merge this without the test for the private method behaviour, but will have a look at refactoring the websocket code a bit later

PR #6908, opened 2021-01-28
fix: Select star table scans need keys in the intermediate row

AlanConfluent:
### Description 
The intermediate row in a projection currently only contains a key if the `PullFilterNode` requires it, which means that `select * ` only works on key lookups.  With this change, we make sure to add keys if it's a `select * ` so that it's now found for table scan.

### Testing done 
RQTT tests and unit tests.

### Reviewer checklist


agavra (approved):

guozhangwang (commented):
LGTM!

PR #6909, opened 2021-01-28
fix: dont use the streamthread to construct the log-metric-and-continue logger

rodesai:
Don't use the streamthread to construct the log-metric-and-continue logger. Also set the level to
debug - this indicates an error with the data. In production usage of ksql, we've frequently seen
these deserialization errors spamming the ksql log. It was hard to tune the log level because we were
logging with the StreamThread category (which has lots of useful logs for streams internals). It also
didn't make sense to log these as warnings/errors since they didn't indicate a ksql bug or error.


agavra (approved):
Great! Thanks for digging into where this exception is thrown. I think it would be great to add a bit more background to the PR so that we have some tracking about why we did this

PR #6918, opened 2021-01-29
chore: setup the skeleton for a migrations tool

agavra:
### Description 

Set up all the boilerplate for the new migrations tool (see #6721)

### Testing done 

Ran the `help` commands for various use cases to make sure that it works as expected:

#### Help Command

This is the default command that runs with no arguments

```
usage: ksql-migrations [ {-c | --config-file} <config-file> ]
        [ {-n | --dry-run} ] <command> [ <args> ]

Commands are:
    apply      Migrates a schema to new available schema versions
    create     Create a pair of migration files with <desc> as description. This will created a pair of empty migration files based on the next schema version.
    info       Displays information about the current and available migrations
    new        Cleans all resources related to migrations. WARNING: this is not reversible!
    validate   Validate applied migrations against local files.

See 'ksql-migrations help <command>' for more information on a specific
command.
```

---

#### Apply Command

```
NAME
        ksql-migrations apply - Migrates a schema to new available schema
        versions

SYNOPSIS
        ksql-migrations [ {-c | --config-file} <config-file> ]
                [ {-n | --dry-run} ] apply [ {-a | --all} ] [ {-n | --next} ]
                [ {-u | --until} <version> ]

OPTIONS
        -a, --all
            run all available migrations

            This option is part of the group 'target' from which only one
            option may be specified


        -c <config-file>, --config-file <config-file>
            Specifies a configuration file

        -n, --dry-run
            dry run the clean, don't actually delete any ksqlDB resources

        -n, --next
            migrate the next available version

            This option is part of the group 'target' from which only one
            option may be specified


        -u <version>, --until <version>
            migrate until the specified version

            This option is part of the group 'target' from which only one
            option may be specified
```

### Reviewer checklist

### DIFF
--- a/ksqldb-tools/src/main/java/io/confluent/ksql/tools/migrations/commands/BaseCommand.java
+++ b/ksqldb-tools/src/main/java/io/confluent/ksql/tools/migrations/commands/BaseCommand.java
@@ -40,7 +40,7 @@ public abstract class BaseCommand implements Runnable {
   @Option(
       name = {"-n", "--dry-run"},
       title = "dry-run",
-      description = "dry run the clean, don't actually delete any ksqlDB resources",
+      description = "dry run the current command, no ksqlDB statements will be executed",
       type = OptionType.GLOBAL
   )
   protected boolean dryRun = false;
--- a/ksqldb-tools/src/main/java/io/confluent/ksql/tools/migrations/commands/BaseCommand.java
+++ b/ksqldb-tools/src/main/java/io/confluent/ksql/tools/migrations/commands/BaseCommand.java
@@ -1,0 +1,48 @@
+/*
+ * Copyright 2020 Confluent Inc.
+ *
+ * Licensed under the Confluent Community License (the "License"; you may not use
+ * this file except in compliance with the License. You may obtain a copy of the
+ * License at
+ *
+ * http://www.confluent.io/confluent-community-license
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OF ANY KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations under the License.
+ */
+
+package io.confluent.ksql.tools.migrations.commands;
+
+import com.github.rvesse.airline.annotations.Option;
+import com.github.rvesse.airline.annotations.OptionType;
+import edu.umd.cs.findbugs.annotations.SuppressFBWarnings;
+
+/**
+ * Defines common options across all of the migration
+ * tool commands.
+ */
+@SuppressFBWarnings(
+    value = "URF_UNREAD_PUBLIC_OR_PROTECTED_FIELD",
+    justification = "code is skeleton only at the moment, used to generate HELP message"
+)
+public abstract class BaseCommand implements Runnable {
+
+  @Option(
+      name = {"-c", "--config-file"},
+      title = "config-file",
+      description = "Specifies a configuration file",
+      type = OptionType.GLOBAL
+  )
+  protected String configFile;
+
+  @Option(
+      name = {"-n", "--dry-run"},
+      title = "dry-run",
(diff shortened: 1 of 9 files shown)


ksqldb-tools/src/main/java/io/confluent/ksql/tools/migrations/commands/CreateMigrationCommand.java
```diff
@@ -0,0 +1,55 @@
...
+import com.github.rvesse.airline.annotations.Arguments;
+import com.github.rvesse.airline.annotations.Command;
+import com.github.rvesse.airline.annotations.Option;
+import com.github.rvesse.airline.annotations.help.Examples;
+import com.github.rvesse.airline.annotations.restrictions.Required;
+
+@Command(
+    name = "create",
+    description = "Create a pair of migration files with <desc> as description. "
+        + "This will created a pair of empty migration files based on the "
```
guozhangwang, line 27: Why it creates "a pair of files", instead of a single file?

guozhangwang (commented):
I got a bit confused by the term "schema migration" from https://github.com/confluentinc/ksql/pull/6721/files. But after reading about it, it seems more like "persistent query group migration", is that right?

My understanding is that:

* A "version" here basically represents a group of persistent queries, and a newer version would always be a superset of an older version (assuming we do not take "drop"s into the version). In other words, we cannot modify an existing CSAS/CSAT schemas (like, adding a new column to the resulted entity) actually, but only to add new queries.
* We store the schema metadata both locally as a file and on a special internal Kafka topic, and hence we can do validation by fetching the metadata file from the Kafka topic, and compare that with the local file.

Are these understanding correct?

ksqldb-tools/src/main/java/io/confluent/ksql/tools/migrations/commands/CreateMigrationCommand.java
agavra: Oops! Copy paste issue. Originally it was going to be one for creating the migration and one for rolling it back. Rolling it back will no longer be something we support in V1. I'll fix this.

agavra:
>I got a bit confused by the term "schema migration" from https://github.com/confluentinc/ksql/pull/6721/files. But after reading about it, it seems more like "persistent query group migration", is that right?

This is mostly correct. I think it's somewhere in between, ideally you would be able to manage your entire database schemas using this tool - so it doesn't really "group" persistent queries together it just manages them. 

> A "version" here basically represents a group of persistent queries, and a newer version would always be a superset of an older version (assuming we do not take "drop"s into the version). In other words, we cannot modify an existing CSAS/CSAT schemas (like, adding a new column to the resulted entity) actually, but only to add new queries.

This is not necessarily correct, we can issue a `CREATE OR REPLACE` as the next statement in a sequence. You would be able to issue any query that you wanted via the schema migrations tool.

> We store the schema metadata both locally as a file and on a special internal Kafka topic, and hence we can do validation by fetching the metadata file from the Kafka topic, and compare that with the local file.

That is correct. The set of files on disk would ideally be checked in to version control and you would use that when spinning up alternative environments.

spena (approved):
Awesome! I like the simple structure of this tool. 

The code lgtm. I just left a comment about the clean command.

ksqldb-tools/src/main/java/io/confluent/ksql/tools/migrations/commands/BaseCommand.java
```diff
@@ -0,0 +1,48 @@
...
+      title = "config-file",
+      description = "Specifies a configuration file",
+      type = OptionType.GLOBAL
+  )
+  protected String configFile;
+
+  @Option(
+      name = {"-n", "--dry-run"},
+      title = "dry-run",
+      description = "dry run the clean, don't actually delete any ksqlDB resources",
```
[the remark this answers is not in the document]
agavra: oops! I had originally implemented this only for the `clean` command, and then moved it into the base class and forgot to update the docs. I'll update that

Issue #6596, opened 2020-11-10
Spotbugs raising issue on java11, not on java8


andrewegel:
**Describe the bug**
I was experimenting with build some of Confluent's projects with java11, and I discovered that on `master`, building with java11, a spot-bugs pops up:

```
% mvn --version
Apache Maven 3.6.3 (cecedd343002696d0abb50b32b541b8a6ba2883f)
Maven home: /usr/local/Cellar/maven/3.6.3/libexec
Java version: 11.0.9, vendor: AdoptOpenJDK, runtime: /Library/Java/JavaVirtualMachines/adoptopenjdk-11.jdk/Contents/Home
Default locale: en_US, platform encoding: UTF-8
OS name: "mac os x", version: "10.15.7", arch: "x86_64", family: "mac"
...
% mvn -D skipTests install
...
[ERROR] Redundant nullcheck of resourceAsStream, which is known to be non-null in io.confluent.ksql.util.AppInfo.<static initializer for AppInfo>() [io.confluent.ksql.util.AppInfo] Redundant null check at AppInfo.java:[line 43] RCN_REDUNDANT_NULLCHECK_OF_NONNULL_VALUE
```

This doesn't happen on java8 - its a pretty simple reproduction case.

**To Reproduce**
Steps to reproduce the behavior, include:
1. The version of KSQL. 

Check out `master` branch. Try building project:

```
mvn -D skipTests install
```

Platform: MacOS 10.15.7, adoptopenjdk-11

2. Sample source data.

N/A

3. Any SQL statements you ran

N/A

**Expected behavior**
The project builds & doesn't raise a spot-bugs issue.

**Actual behaviour**
See above

**Additional context**
I acknowledge this will probably be a low priority. You can find me on slack if you need additional context/details.

brianstrauch:
Just ran into this issue as well, after building the project for the first time. It looks like this was caused by a combination of https://github.com/confluentinc/ksql/pull/6509 and https://sourceforge.net/p/findbugs/bugs/1169/.

This could be fixed by:
1. Updating SpotBugs
2. Excluding the rule `RCN_REDUNDANT_NULLCHECK_OF_NONNULL_VALUE` in https://github.com/confluentinc/ksql/blob/master/findbugs/findbugs-exclude.xml
3. Reverting the changes from https://github.com/confluentinc/ksql/pull/6509

Not sure what the best route to take is, but hope this helps!

mjsax:
We should actually also ensure that we support Java17.

PR #6922, opened 2021-01-30
fix: Spotbugs raising RCN_REDUNDANT_NULLCHECK_OF_NONNULL_VALUE during build

chien-andalou:
### Description 
**Problem**
I was not able to build the master branch of ksql project on my Ubuntu 20.10 System with openjdk version "14.0.2" using "mvn clean package -DskipTests" command. I was getting the following error: 
`[ERROR] Medium: Redundant nullcheck of resourceAsStream, which is known to be non-null in io.confluent.ksql.util.AppInfo.<static initializer for AppInfo>() [io.confluent.ksql.util.AppInfo] Redundant null check at AppInfo.java:[line 43] RCN_REDUNDANT_NULLCHECK_OF_NONNULL_VALUE`

There is an issue that is very similar to the one I am trying to fix: https://github.com/confluentinc/ksql/issues/6596

My environment details:
```
mvn --version
Apache Maven 3.6.3
Maven home: /usr/share/maven
Java version: 14.0.2, vendor: Private Build, runtime: /usr/lib/jvm/java-14-openjdk-amd64
Default locale: en_US, platform encoding: UTF-8
OS name: "linux", version: "5.8.0-41-generic", arch: "amd64", family: "unix"
```
**Solution**
To get rid of the SpotBugs error, I have removed the null check and introduced a check for resource existence before getting it in a stream. After that, the build of ksql has succeeded.

### Testing done 
Because the problem was caused by SpotBugs, I only did a manual testing by running the build with and without the change. The behavior was not changed, therefore no unit or integration tests included. 

### Reviewer checklist

### DIFF
--- a/ksqldb-common/src/main/java/io/confluent/ksql/util/AppInfo.java
+++ b/ksqldb-common/src/main/java/io/confluent/ksql/util/AppInfo.java
@@ -40,13 +40,13 @@ public final class AppInfo {
 
     try {
       final Properties props = new Properties();
-      try (InputStream resourceAsStream = AppInfo.class.getResourceAsStream(
-          "/git.properties")) {
-        if (resourceAsStream != null) {
+      if (AppInfo.class.getResource("/git.properties") != null) {
+        try (InputStream resourceAsStream = AppInfo.class.getResourceAsStream(
+                "/git.properties")) {
           props.load(resourceAsStream);
         }
+        commitId = props.getProperty("git.commit.id", commitId).trim();
       }
-      commitId = props.getProperty("git.commit.id", commitId).trim();
     } catch (final Exception e) {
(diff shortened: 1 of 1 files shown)


agavra:
Thanks for reporting this @chien-andalou! This is a known limitation when compiling with JDK14 (I'm not sure if it's of findbugs or in our code), the entire dev team here uses JDK8 for development at this point. I can keep this ticket open until we migrate to JDK14 but that's not trivial as we need to update a lot of or build pipelines behind the scenes to get that to work.

jkao97:
Due to fork no longer existing, Jenkins is unable to do a repo scan to cleanup old PRs. Please move contents of this PR to a new branch/fork.

PR #6925, opened 2021-02-01
feat: Add lambda type base and context to support lambda type checking

lct45:
### Description 
Adds the basic sql types for lambdas as well as new contexts for `CodeGenRunner` and `SqlToJavaVisitor`. This is the setup for the type checking logic

### Testing done 
Add unit tests, modify existing tests to ignore lambda sql type

### Reviewer checklist

### DIFF
--- a/ksqldb-common/src/main/java/io/confluent/ksql/function/types/KsqlLambdaType.java
+++ b/ksqldb-common/src/main/java/io/confluent/ksql/function/types/KsqlLambdaType.java
@@ -1,0 +1,70 @@
+/*
+ * Copyright 2021 Confluent Inc.
+ *
+ * Licensed under the Confluent Community License (the "License"; you may not use
+ * this file except in compliance with the License. You may obtain a copy of the
+ * License at
+ *
+ * http://www.confluent.io/confluent-community-license
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OF ANY KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations under the License.
+ */
+
+package io.confluent.ksql.function.types;
+
+import java.util.Objects;
+
+public final class KsqlLambdaType extends ObjectType {
+
+  private final ParamType inputType;
+  private final ParamType returnType;
+
+  private KsqlLambdaType(
+      final ParamType inputType,
+      final ParamType returnType
+  ) {
+    this.inputType = inputType;
+    this.returnType = returnType;
+  }
+
+  public static KsqlLambdaType of(
+      final ParamType inputType,
+      final ParamType returnType
+  ) {
+    return new KsqlLambdaType(inputType, returnType);
+  }
+
+  public ParamType inputType() {
+    return inputType;
+  }
+
+  public ParamType returnType() {
+    return returnType;
+  }
+
+  @Override
+  public boolean equals(final Object o) {
+    if (this == o) {
+      return true;
+    }
+    if (o == null || getClass() != o.getClass()) {
+      return false;
+    }
+    final KsqlLambdaType arrayType = (KsqlLambdaType) o;
+    return Objects.equals(inputType, arrayType.inputType)
+        && Objects.equals(returnType, arrayType.returnType);
+  }
+
+  @Override
+  public int hashCode() {
+    return Objects.hash(inputType);
+  }
+
+  @Override
+  public String toString() {
+    return "KSQL_LAMBDA<" + inputType + ", " + returnType + ">";
+  }
+}
--- a/ksqldb-common/src/main/java/io/confluent/ksql/function/types/ParamTypes.java
+++ b/ksqldb-common/src/main/java/io/confluent/ksql/function/types/ParamTypes.java
@@ -23,6 +23,8 @@ import io.confluent.ksql.schema.ksql.types.SqlMap;
 import io.confluent.ksql.schema.ksql.types.SqlStruct;
 import io.confluent.ksql.schema.ksql.types.SqlStruct.Field;
 import io.confluent.ksql.schema.ksql.types.SqlType;
+
+import java.sql.Struct;
 import java.util.Map.Entry;
 import java.util.Optional;
 
@@ -38,6 +40,8 @@ public final class ParamTypes {
   public static final LongType LONG = LongType.INSTANCE;
   public static final ParamType DECIMAL = DecimalType.INSTANCE;
   public static final TimestampType TIMESTAMP = TimestampType.INSTANCE;
+  public static final KsqlLambdaType LAMBDALITERAL = KsqlLambdaType.of(null, null);
+
 
   public static boolean areCompatible(final SqlType actual, final ParamType declared) {
     return areCompatible(actual, declared, false);
--- a/ksqldb-common/src/main/java/io/confluent/ksql/function/types/ParamTypes.java
+++ b/ksqldb-common/src/main/java/io/confluent/ksql/function/types/ParamTypes.java
@@ -23,8 +23,6 @@ import io.confluent.ksql.schema.ksql.types.SqlMap;
 import io.confluent.ksql.schema.ksql.types.SqlStruct;
 import io.confluent.ksql.schema.ksql.types.SqlStruct.Field;
 import io.confluent.ksql.schema.ksql.types.SqlType;
-
-import java.sql.Struct;
 import java.util.Map.Entry;
 import java.util.Optional;
 
@@ -41,7 +39,6 @@ public final class ParamTypes {
   public static final ParamType DECIMAL = DecimalType.INSTANCE;
   public static final TimestampType TIMESTAMP = TimestampType.INSTANCE;
   public static final KsqlLambdaType LAMBDALITERAL = KsqlLambdaType.of(null, null);
-
 
   public static boolean areCompatible(final SqlType actual, final ParamType declared) {
     return areCompatible(actual, declared, false);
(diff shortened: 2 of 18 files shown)


ksqldb-common/src/test/java/io/confluent/ksql/schema/OperatorTest.java
```diff
@@ -95,11 +96,11 @@ public void shouldResolveModulusReturnType() {
   @Test
   public void shouldWorkUsingSameRulesAsBaseTypeUpCastRules() {
     allOperations().forEach(op -> {
-
-      for (final SqlBaseType leftBaseType : SqlBaseType.values()) {
+      SqlBaseType[] types = ArrayUtils.removeElement(SqlBaseType.values(), SqlBaseType.LAMBDA);
```
lct45, line 99: This logic is definitely messy - the goal is to keep the test the same for when we add adiditonal sql types but ignore lambdas as we don't need full functionality for them. The lambda sql type is meant to be an internal type so is not so relevant to many of these tests

ksqldb-common/src/main/java/io/confluent/ksql/function/types/KsqlLambdaType.java
```diff
@@ -0,0 +1,70 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OF ANY KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations under the License.
+ */
+
+package io.confluent.ksql.function.types;
+
+import java.util.Objects;
+
+public final class KsqlLambdaType extends ObjectType {
```
agavra, line 20: nit: do we need to prefix everything with `Ksql`? Personally, I prefer just `LambdaType` 😂

ksqldb-common/src/main/java/io/confluent/ksql/function/types/KsqlLambdaType.java
```diff
@@ -0,0 +1,70 @@
...
+      return false;
+    }
+    final KsqlLambdaType arrayType = (KsqlLambdaType) o;
+    return Objects.equals(inputType, arrayType.inputType)
+        && Objects.equals(returnType, arrayType.returnType);
+  }
+
+  @Override
+  public int hashCode() {
+    return Objects.hash(inputType);
```
agavra, line 63: do we not need to include the return type here?

ksqldb-common/src/main/java/io/confluent/ksql/function/types/KsqlLambdaType.java
```diff
@@ -0,0 +1,70 @@
...
+
+  @Override
+  public boolean equals(final Object o) {
+    if (this == o) {
+      return true;
+    }
+    if (o == null || getClass() != o.getClass()) {
+      return false;
+    }
+    final KsqlLambdaType arrayType = (KsqlLambdaType) o;
```
agavra, line 56: did you mean `lambdaType`?

ksqldb-common/src/main/java/io/confluent/ksql/function/types/ParamTypes.java
```diff
@@ -38,6 +38,7 @@ private ParamTypes() {
   public static final LongType LONG = LongType.INSTANCE;
   public static final ParamType DECIMAL = DecimalType.INSTANCE;
   public static final TimestampType TIMESTAMP = TimestampType.INSTANCE;
+  public static final KsqlLambdaType LAMBDALITERAL = KsqlLambdaType.of(null, null);
```
agavra, line 41: Looking at this out of context doesn't really make sense to me. Creating a lambda type with no input/output type seems like a bug (and using `null` to proxy this makes me feel even more nervous, in fact I'd probably recommend that we `requireNonNull` for these in the constructor). I'll try to look throughout at where this is used and comment with some more details later in the PR, but I suspect we should follow the pattern of `MapType` and `ArrayType` which are complex types (they don't have corresponding literals in this list)

ksqldb-common/src/main/java/io/confluent/ksql/schema/ksql/SchemaConverters.java
```diff
@@ -381,6 +383,7 @@ public SqlBaseType toSqlType(final Class<?> javaType) {
             .put(ParamTypes.LONG, SqlTypes.BIGINT)
             .put(ParamTypes.DOUBLE, SqlTypes.DOUBLE)
             .put(ParamTypes.TIMESTAMP, SqlTypes.TIMESTAMP)
+            .put(ParamTypes.LAMBDALITERAL, SqlTypes.LAMBDALITERAL)
```
agavra, line 386: as mentioned above, I think what we should do here is not put it into the map and instead use:
```java
if (paramType instanceOf LambdaType) {
  return SqlTypes.lambda(toSqlType(paramType.input), toSqlType(paramType.output));
}
```

ksqldb-common/src/main/java/io/confluent/ksql/schema/ksql/SchemaConverters.java
```diff
@@ -423,6 +426,7 @@ public SqlType toSqlType(final ParamType paramType) {
             .put(ParamTypes.DOUBLE, SqlBaseType.DOUBLE)
             .put(ParamTypes.DECIMAL, SqlBaseType.DECIMAL)
             .put(ParamTypes.TIMESTAMP, SqlBaseType.TIMESTAMP)
+            .put(ParamTypes.LAMBDALITERAL, SqlBaseType.LAMBDA)
```
agavra, line 429: this is error prone because any `LambdaType` that checks with this map that isn't _exactly_ `Lambda.of(null, null)` will not match and will throw an exception.

ksqldb-common/src/test/java/io/confluent/ksql/schema/OperatorTest.java
agavra: Why do we need this? Looking at the logic of the test, `shouldBeSupported` should return `false` here and we should get the exceptions we expect - right?

ksqldb-common/src/test/java/io/confluent/ksql/schema/ksql/SchemaConvertersTest.java
```diff
@@ -162,8 +166,7 @@ public void shouldHaveConnectTestsForAllSqlTypes() {
         .map(SqlType::baseType)
         .collect(Collectors.toSet());
 
-    final ImmutableSet<SqlBaseType> allTypes = ImmutableSet.copyOf(SqlBaseType.values());
-
+    final ImmutableSet<SqlBaseType> allTypes = ImmutableSet.copyOf(ArrayUtils.removeElement(SqlBaseType.values(), SqlBaseType.LAMBDA));
```
agavra, line 169: Ditto as above - I think if we implement this properly I don't see why lambdas should be any different than array types at this point. (I'll stop commenting this, but there are various places where we omit lambdas where I don't think we should treat them differently in our type system)

agavra (commented):
I think we're missing out on something here. There's already a generic type (see `GenericType`) to support the use case you have here (lambdas that you don't know the type for).

Let's consider the "map" lambda: `transform(array, x -> x + 1)`. This UDF would have the following java signature:
```
public List<T> transform(List<T> list, Function<T, T> mapper);
```

Since we already have the mechanism to resolve `List<T>` we can just apply that to `Function<T, T>` based on what we "learned" from `List<T>` (namely, we've already assigned `T` a concrete type by the time we look at the lambda).

You can see this logic in `GenericsUtil#resolveGenerics`, which resolves generics to their corresponding types.

ksqldb-common/src/main/java/io/confluent/ksql/function/types/KsqlLambdaType.java
stevenpyzhang: Sorry, that was my choice in the POC

ksqldb-common/src/main/java/io/confluent/ksql/function/types/KsqlLambdaType.java
```diff
@@ -0,0 +1,70 @@
...
+
+import java.util.Objects;
+
+public final class KsqlLambdaType extends ObjectType {
+
+  private final ParamType inputType;
+  private final ParamType returnType;
+
+  private KsqlLambdaType(
+      final ParamType inputType,
```
guozhangwang, line 26: Could a lambda type includes multiple input types?

ksqldb-udf/src/main/java/io/confluent/ksql/schema/ksql/types/SqlLambda.java
```diff
@@ -0,0 +1,81 @@
...
+import static java.util.Objects.requireNonNull;
+
+import com.google.errorprone.annotations.Immutable;
+import io.confluent.ksql.schema.utils.FormatOptions;
+import java.util.Objects;
+
+@Immutable
+public final class SqlLambda extends SqlType {
+
+  private final SqlType inputType;
```
guozhangwang, line 27: Just for my own understanding: how to represent `(X, Y) -> X + Y` with a single input type?

ksqldb-common/src/main/java/io/confluent/ksql/function/types/KsqlLambdaType.java
stevenpyzhang: It can, this PR was built off of my initial proof of concept which only supported 1 input type. The most recent version can support up to 3 input types https://github.com/confluentinc/ksql/pull/6919 (still a Draft for now though)

ksqldb-udf/src/main/java/io/confluent/ksql/schema/ksql/types/SqlLambda.java
stevenpyzhang: See comment above.

PR #6927, opened 2021-02-01
feat: Support timestamp protobuf serde

jzaralim:
### Description 
One of the reasons the decimal and timestamp types were not enabled in protobuf was because those two types have dependencies that need to be registered first in order for schema registry to be able to read schemas involving these types. This PR adds an extra step before registering protobuf schemas that resolves dependencies. 

### Testing done 
manually, unit and QTT tests

### Reviewer checklist

### DIFF
--- a/ksqldb-engine/src/main/java/io/confluent/ksql/schema/ksql/inference/SchemaRegisterInjector.java
+++ b/ksqldb-engine/src/main/java/io/confluent/ksql/schema/ksql/inference/SchemaRegisterInjector.java
@@ -215,7 +215,7 @@ public class SchemaRegisterInjector implements Injector {
           final ProtobufSchema resolved = AbstractKafkaProtobufSerializer.resolveDependencies(
               srClient,
               true,
-              true,
+              false,
               true,
               null,
               new DefaultReferenceSubjectNameStrategy(),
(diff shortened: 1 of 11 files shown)


jzaralim:
cc @rayokota

ksqldb-engine/src/main/java/io/confluent/ksql/schema/ksql/inference/SchemaRegisterInjector.java
```diff
@@ -205,8 +211,22 @@ private void registerSchema(
         final ParsedSchema parsedSchema = translator.toParsedSchema(
             PersistenceSchema.from(schema, serdeFeatures)
         );
-
-        srClient.register(subject, parsedSchema);
+        if (parsedSchema instanceof ProtobufSchema) {
+          final ProtobufSchema resolved = AbstractKafkaProtobufSerializer.resolveDependencies(
+              srClient,
+              true,
+              true,
```
rayokota, line 218: nit: the arg for `useLatestVersion` is not used because `autoRegisterSchema` is true, but it would probably be best to pass `false` anyway (since `false` is the default for `useLatestVersion` in general).  So the args should read `srClient, true, false, true`.

rayokota (approved):
@jzaralim LGTM!  Left a minor nit

agavra (approved):
Nice! Thanks @jzaralim - I have one comment for @rayokota inline

ksqldb-engine/src/main/java/io/confluent/ksql/schema/ksql/inference/SchemaRegisterInjector.java
rayokota: @agavra , sure I've filed https://confluentinc.atlassian.net/browse/DGS-1285 to track.  The method recursively registers referenced schemas.

spena (approved):
lgtm! Thanks Zara

PR #6930, opened 2021-02-02
fix: do not log data on JSON deserialization errors (MINOR)

vcrfxia:
### Description 

JsonParseExceptions are automatically populated with location information to indicate where the parse exception occurred. However, we don't want to log this from the ksqlDB JSON deserializer as it may result in topic data being logged. This PR clears the location data to avoid this.

Here's an example of what a stacktrace may look like prior to this PR:
```
Exception caught during Deserialization, taskId: 1_0, topic: my_topic, partition: 0, offset: 1901512
org.apache.kafka.common.errors.SerializationException: Failed to deserialize value from topic: my_topic. Illegal character ((CTRL-CHAR, code 0)): only regular white space (\r, \n, \t) is allowed between tokens
at [Source: (byte[])"REDACTED -- we don't want to log this"; line: 1, column: 2]
Caused by: com.fasterxml.jackson.core.JsonParseException: Illegal character ((CTRL-CHAR, code 0)): only regular white space (\r, \n, \t) is allowed between tokens
at [Source: (byte[])"REDACTED -- we don't want to log this"; line: 1, column: 2]
```
And here's what it looks like after this PR:
```
Exception caught during Deserialization, taskId: 1_0, topic: my_topic, partition: 0, offset: 1901512
org.apache.kafka.common.errors.SerializationException: Failed to deserialize value from topic: my_topic. Illegal character ((CTRL-CHAR, code 0)): only regular white space (\r, \n, \t) is allowed between tokens
Caused by: com.fasterxml.jackson.core.JsonParseException: Illegal character ((CTRL-CHAR, code 0)): only regular white space (\r, \n, \t) is allowed between tokens
```

### Testing done 

Added unit test.

### Reviewer checklist


rodesai (approved):
LGTM

guozhangwang (approved):
LGTM.

Issue #6932, opened 2021-02-02
GROUP BY with multiple fields fails if the source has `KEY_FORMAT='KAFKA'`


rodesai:
GROUP BY with multiple fields fails if the data source has `KEY_FORMAT='KAFKA'`, even if the user explicitly asks for the sink to have KEY_FORMAT JSON. It complains that the KAFKA serialization format does not support multiple fields:

```
$ cat statements.sql
CREATE STREAM CLICKS (IP_ADDRESS STRING, URL STRING, TIMESTAMP STRING)
    WITH (KAFKA_TOPIC = 'CLICKS',
          KEY_FORMAT = 'KAFKA',
          VALUE_FORMAT = 'JSON',
          TIMESTAMP = 'TIMESTAMP',
          TIMESTAMP_FORMAT = 'yyyy-MM-dd''T''HH:mm:ssXXX',
          PARTITIONS = 1);

CREATE TABLE DETECTED_CLICKS WITH (KEY_FORMAT='JSON') AS
    SELECT
        IP_ADDRESS AS KEY1,
        URL AS KEY2,
        TIMESTAMP AS KEY3,
        AS_VALUE(IP_ADDRESS) AS IP_ADDRESS,
        AS_VALUE(URL) AS URL,
        AS_VALUE(TIMESTAMP) AS TIMESTAMP
    FROM CLICKS WINDOW TUMBLING (SIZE 2 MINUTES, RETENTION 3650 DAYS)
    GROUP BY IP_ADDRESS, URL, TIMESTAMP
    HAVING COUNT(IP_ADDRESS) = 1;

CREATE STREAM RAW_DISTINCT_CLICKS (IP_ADDRESS STRING, URL STRING, TIMESTAMP STRING)
    WITH (KAFKA_TOPIC = 'DETECTED_CLICKS',
    KEY_FORMAT='JSON',
    VALUE_FORMAT = 'JSON');

CREATE STREAM DISTINCT_CLICKS AS
    SELECT
        IP_ADDRESS,
        URL,
        TIMESTAMP
    FROM RAW_DISTINCT_CLICKS
    WHERE IP_ADDRESS IS NOT NULL
    PARTITION BY IP_ADDRESS;

$ cat inputs.json
{
  "inputs": [
    {
      "topic": "CLICKS",
      "key": "\"10.0.0.1\"",
      "value": {
        "IP_ADDRESS": "10.0.0.1",
[...]

PR #6933, opened 2021-02-02
fix: use internal format with multi-col support when needed

vcrfxia:
### Description 

Fixes https://github.com/confluentinc/ksql/issues/6932 by having internal serdes use the JSON format when needed. 

### Testing done 

QTT.

### Reviewer checklist


rodesai (approved):
LGTM

agavra (approved):

ksqldb-engine/src/main/java/io/confluent/ksql/structured/SchemaKGroupedStream.java
```diff
@@ -85,7 +85,8 @@
     } else {
       keyFormat = SerdeFeaturesFactory.sanitizeKeyFormat(
           this.keyFormat,
-          schema.key().size() == 1
+          schema.key().size(),
```
guozhangwang, line 88: Just for clarification: why do we allow change key format change in groupBy / selectKey but not in aggregate? Is the internal formats of groupBy / selectKey used for the repartition topic, not the aggregate?

guozhangwang (commented):
Thanks @vcrfxia for the quick fix! I just had a qq about the physical plan behind it for my own education.

ksqldb-engine/src/main/java/io/confluent/ksql/structured/SchemaKGroupedStream.java
vcrfxia: The aggregate is always preceded by a groupBy, and the key schema does not change between the groupBy and the aggregate. As long as we properly sanitize the key format in groupBy, then we do not need to re-sanitize during aggregate (the aggregate should already inherit a sanitized key format from its parent step, the groupBy).

guozhangwang: Yup, thanks for the clarification. I was just wondering if the key format after the "groupBy" is used in the repartition topic, or the key format after the "aggregate" is used in the repartition topic, or none were used in the repartition topic.

So it seems that still, the key format after the "aggregate" is used, but that is always the same as the one after the "groupBy", is my understanding correct?

vcrfxia: You're asking about the changelog topic, right? If so, you're correct that it's the key format of the "aggregate" that's used ([link](https://github.com/confluentinc/ksql/blob/b69d74cfeb3033748c11e3a4ba7924d1e4123d58/ksqldb-streams/src/main/java/io/confluent/ksql/execution/streams/StreamAggregateBuilder.java#L96)) rather than the key format of the "groupBy", and that the two key formats should always be the same as the "aggregate" inherits its formats from the "groupBy".

guozhangwang: Yes, I meant changelog topic :) And thanks for the pointer!

PR #6940, opened 2021-02-03
minor: add parameters for specifying more stuff about releases

rodesai:
Adds build parameters for specifying:
- beta tag / build number. this should avoid some rebuild/cancel/replay/edit loops
- downstream ccloud git revision and image (for fix builds). This way you can kick
  off the whole pipeline for fix builds with 1 step (or 2 if you need new beta
  pkgs).

This should help us simplify some of our release instructions

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -20,7 +20,7 @@ def defaultParams = [
         description: 'Is this a release build. If so, GIT_REVISION and KSQLDB_VERSION must be specified, and the downstream CCloud job will not be triggered.',
         defaultValue: false),
     string(name: 'GIT_REVISION',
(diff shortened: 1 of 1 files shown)


vcrfxia (approved):
LGTM. The number of build params is getting rather intimidating, but I suppose that's not new 😅

rodesai:
> LGTM. The number of build params is getting rather intimidating, but I suppose that's not new 😅

Yeah. We will cleanup the base image one for cc-ksql-hotfix very soon (by basing the hotfix on nano-versions for cloud plugins). And if we switch up the build process so you don't have to worry about the cc-docker-ksql ones and just look at this build we should actually wind up with fewer params.

PR #6944, opened 2021-02-03
fix: Make pull query metrics apply only to pull and not also push

AlanConfluent:
### Description 
This changes pull query metrics to be reported only for pull queries, and not push, since both use the `/query` endpoint.

### Testing done 
Ran tests and manually verified.

### Reviewer checklist

### DIFF
--- a/ksqldb-rest-app/src/main/java/io/confluent/ksql/api/server/OldApiUtils.java
+++ b/ksqldb-rest-app/src/main/java/io/confluent/ksql/api/server/OldApiUtils.java
@@ -21,7 +21,6 @@ import static org.apache.hc.core5.http.HttpHeaders.TRANSFER_ENCODING;
 
 import io.confluent.ksql.api.auth.ApiSecurityContext;
 import io.confluent.ksql.api.auth.DefaultApiSecurityContext;
-import io.confluent.ksql.internal.PullQueryExecutorMetrics;
 import io.confluent.ksql.rest.EndpointResponse;
 import io.confluent.ksql.rest.Errors;
 import io.confluent.ksql.rest.entity.KsqlErrorMessage;
@@ -39,6 +38,7 @@ import java.io.OutputStream;
 import java.util.Optional;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.CompletionException;
+import java.util.concurrent.atomic.AtomicReference;
 import java.util.function.BiFunction;
 import org.apache.kafka.common.utils.Time;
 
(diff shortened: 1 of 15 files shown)


ksqldb-rest-app/src/main/java/io/confluent/ksql/api/server/OldApiUtils.java
```diff
@@ -176,4 +186,33 @@ public static EndpointResponse mapException(final Throwable exception) {
         .build();
   }
 
+  /**
+   * Interface for reporting metrics to a resource. A resource may choose to break things down
+   * arbitrarily, e.g. /query is used for both push and pull queries so we let the resource
+   * determine how to report the metrics.
+   */
+  public interface MetricsCallback {
```
agavra, line 194: nit: since this is used in the old and the new endpoints can we pull it out into its own class? (same for the impl below)

agavra (approved):
LGTM, though is it possible to add automated tests to make sure this doesnt' regress in a future refactor?

ksqldb-rest-app/src/main/java/io/confluent/ksql/api/server/OldApiUtils.java
AlanConfluent: It wasn't used in the new endpoint when I wrote this.  I moved them out and now call it from the new endpoint `stream-query` as well.  I still don't call it from `/ws/query` because it's a bit harder to do it in these scenarios, though we report latency and basic count from there.

ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/resources/streaming/PullQueryPublisher.java
```diff
@@ -108,15 +108,9 @@ public synchronized void subscribe(final Subscriber<Collection<StreamedRow>> sub
...
-      result.onCompletionOrException((v, t) -> decrementer.decrementAtMostOnce());
-      result.onCompletion(v -> {
-        pullQueryMetrics.ifPresent(p -> p.recordLatency(startTimeNanos));
-      });
       result.onCompletionOrException((v, throwable) -> {
         decrementer.decrementAtMostOnce();
-        if (throwable == null) {
-          pullQueryMetrics.ifPresent(p -> p.recordLatency(startTimeNanos));
-        }
+        pullQueryMetrics.ifPresent(p -> p.recordLatency(startTimeNanos));
```
guozhangwang, line 113: This seems changing the behavior that even if the result contains an error, we would still record latency. Is it intentional?

guozhangwang (commented):
@AlanConfluent for my understanding: which class/line is the actual fix -- it seems to be hidden somewhere with this refactoring change that I'm missing..

PR #6945, opened 2021-02-03
chore: update how AstBuilder and AstSanitizer handles lambda functions

stevenpyzhang:
### Description 
Instead of creating LambdaVariables in AstSanitizer, we do it in AstBuilder.
We need to do this because we need to be able to construct the LambdaVariable node from only AstBuilder since that's how KSQL plan's are deserialized (we can't do sanitization on expressions in the plan)

This is done by keeping track of what the LambdaArgs are in AstBuilder. Before visiting the body of the lambda expression, we add the args to the LambdaArgs set while keeping track of what the initial state of the set was before visiting the body. After returning from the body, we set the LambdaArgs set back to the initial state.

Renamed LambdaLiteral -> LambdaVariable 

### Testing done 
Updated unit tests

### Reviewer checklist

### DIFF
--- a/ksqldb-engine/src/main/java/io/confluent/ksql/engine/rewrite/AstSanitizer.java
+++ b/ksqldb-engine/src/main/java/io/confluent/ksql/engine/rewrite/AstSanitizer.java
@@ -216,9 +216,12 @@ public final class AstSanitizer {
         }
       });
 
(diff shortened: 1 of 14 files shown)


guozhangwang (approved):
LGTM!

PR #6948, opened 2021-02-04
test: add connect integration test for timestamps

jzaralim:
### Description 
Adds a test that reads an integer from a source connector and writes it out to a sink connector

### Reviewer checklist

### DIFF
--- a/ksqldb-rest-app/src/test/java/io/confluent/ksql/rest/server/ConnectIntegrationTest.java
+++ b/ksqldb-rest-app/src/test/java/io/confluent/ksql/rest/server/ConnectIntegrationTest.java
@@ -40,6 +40,8 @@ import io.confluent.ksql.rest.entity.KsqlEntityList;
 import io.confluent.ksql.rest.entity.StreamedRow;
 import io.confluent.ksql.rest.entity.WarningEntity;
 import java.io.ByteArrayOutputStream;
+import java.io.FileDescriptor;
+import java.io.FileOutputStream;
 import java.io.PrintStream;
 import java.util.Arrays;
 import java.util.HashSet;
@@ -129,6 +131,8 @@ public class ConnectIntegrationTest {
             .get(0)).getConnectors(),
         Matchers.empty()
     );
+
+    System.setOut(new PrintStream(new FileOutputStream(FileDescriptor.out)));
   }
 
   @Test
--- a/ksqldb-rest-app/src/test/java/io/confluent/ksql/rest/server/ConnectIntegrationTest.java
+++ b/ksqldb-rest-app/src/test/java/io/confluent/ksql/rest/server/ConnectIntegrationTest.java
@@ -22,6 +22,7 @@ import static org.hamcrest.Matchers.equalToIgnoringCase;
 import static org.hamcrest.Matchers.hasSize;
 import static org.hamcrest.Matchers.instanceOf;
 import static org.hamcrest.Matchers.is;
+import static org.hamcrest.Matchers.stringContainsInOrder;
 
 import com.google.common.base.Joiner;
 import com.google.common.collect.ImmutableMap;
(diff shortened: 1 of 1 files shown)


ksqldb-rest-app/src/test/java/io/confluent/ksql/rest/server/ConnectIntegrationTest.java
```diff
@@ -239,6 +252,68 @@ public void shouldReturnError() {
...
+        .put("topic", "foo")
+        .put("throughput", "5")
+        .put("id", "123")
+        .put("topic.creation.default.replication.factor", "1")
+        .put("topic.creation.default.partitions", "1")
+        .build());
+    assertThatEventually(
+        () -> ((ConnectorList) ksqlRestClient.makeKsqlRequest("SHOW CONNECTORS;").getResponse()
+            .get(0)).getConnectors().get(0).getState(),
+        is("RUNNING (1/1 tasks RUNNING)"));
```
spena, line 269: This is the `given` section. I don't think we should have any checks here. But you can move this inside the `create` method after the request is called. Other tests do this often, so it's good if you want to make sure that `create` will mock the right thing.

Btw, how does the assert know the running connector is the one previously created?

ksqldb-rest-app/src/test/java/io/confluent/ksql/rest/server/ConnectIntegrationTest.java
```diff
@@ -239,6 +252,68 @@ public void shouldReturnError() {
...
+
+    // When:
+    assertThatEventually(
```
spena, line 272: This assertion must be in the `then` section. It is when we do our checks that everything works. Btw, do you really need to  check the stream was created successfully? I see you have a `SELECT` clause in the Then section, so that assumes the streams is already created, it will fail if not, so you have your check there.

ksqldb-rest-app/src/test/java/io/confluent/ksql/rest/server/ConnectIntegrationTest.java
```diff
@@ -239,6 +252,68 @@ public void shouldReturnError() {
...
+            .isSuccessful(), is(true));
+
+    // Then:
+    final RestResponse<List<StreamedRow>> queryFoo = ksqlRestClient.makeQueryRequest("SELECT * FROM FOO EMIT CHANGES LIMIT 1;", 1L);
+    assertThat("successfully queried FOO", queryFoo.isSuccessful());
+    assertThat(queryFoo.getResponse().get(1).getRow().get().getColumns().get(0), is("1970-01-01T00:00:00.000"));
+
+    // When:
+    final RestResponse<KsqlEntityList> barQuery = ksqlRestClient
+        .makeKsqlRequest("CREATE STREAM BAR WITH (value_format='JSON') AS SELECT PAYLOAD FROM FOO;");
```
spena, line 283: Seems this test is different. Should have have another method for this? Ideally, we should have one test method per test case.

ksqldb-rest-app/src/test/java/io/confluent/ksql/rest/server/ConnectIntegrationTest.java
jzaralim: This was added because it takes some time for the topic to actually be created, so the command actually fails the first few times. I rewrote this to just keep trying until timeout instead of using an asset.

ksqldb-rest-app/src/test/java/io/confluent/ksql/rest/server/ConnectIntegrationTest.java
```diff
@@ -239,7 +254,70 @@ public void shouldReturnError() {
...
+    // When:
+    final RestResponse<List<StreamedRow>> queryFoo = ksqlRestClient.makeQueryRequest("SELECT * FROM FOO EMIT CHANGES LIMIT 1;", 1L);
+
+    // Then:
+    assertThat("successfully queried FOO", queryFoo.isSuccessful());
+    assertThat(queryFoo.getResponse().get(1).getRow().get().getColumns().get(0), is("1970-01-01T00:00:00.000"));
+  }
+
+  @Test
+  @Ignore
```
jzaralim, line 284: Setting this to ignore, because reading from stdout from a separate library seems to not work with Jenkins (and is probably not a good idea either). Swapping out `org.apache.kafka.connect.tools.VerifiableSinkConnector` with a connector to a real database and checking the values written to the database would work, but it would be a lot more effort than it is worth.

spena (approved):
Thanks, LGTM

PR #6965, opened 2021-02-08
docs: add clarifications and more details to klip 42

vcrfxia:
### Description 

As the title says. Notable updates and clarifications include:
- how we want to handle potential race conditions
- more details around configs
- clarifying that version numbers need not be consecutive
- a few fixes, such as the fact that `apply until <target>` is inclusive not exclusive

See inline comments for more.

### Testing done 

Docs-only change (plus a minor update to the migrations tool skeleton, which has yet to be released).

### Reviewer checklist


design-proposals/klip-42-schema-migrations-tool.md
```diff
@@ -196,35 +201,35 @@ One stream (MIGRATION_EVENTS) and one table (SCHEMA_VERSION).
...
-  REPLICAS=1
+  REPLICAS=1 /* value will be configurable */
 );
 ```
 
-The `version_key` column has the version of the migration applied or undone. Special values of the `version_key`, `CURRENT` and `LATEST` will be reserved for internal purposes.
+The `version_key` column has the version of the migration applied or undone. A special value of the `version_key`, `CURRENT`, will be reserved for internal purposes.
 The `version` column has the version of the migration applied.
 The `name` column has the name of the migration.
-The `state` column has the state of the migration process. It can be any of `Pending`, `Running`, `Migrated`, `Error`, `Undone`.
```
vcrfxia, line 227: What was the intended difference between `Pending` and `Running`? I've removed one for now to simplify the status flows (see "Migration state transitions and potential race conditions" below).

design-proposals/klip-42-schema-migrations-tool.md
```diff
@@ -239,11 +244,9 @@ The below schema is designed so we support undo operations in the future. When a
 
 This is the `CREATE` statement for the `SCHEMA_VERSION` table:
 ```sql
-CREATE TABLE schema_version
+CREATE TABLE schema_version /* name will be configurable */
   WITH (
-    KAFKA_TOPIC='default_ksql_schema_version',
-    VALUE_FORMAT='JSON',
```
vcrfxia, line 245: These extra configs are not necessary since they're inherited from the source by default. Also, replicas should be specified / inherited too.

design-proposals/klip-42-schema-migrations-tool.md
```diff
@@ -363,30 +370,33 @@ Commands
               
               If 'all' is specified, then it migrates all newer versions available
               If 'next' is specified, then it migrates only the next available version
-              If 'until <target>' is specified, then it migrates all available versions before <target>
+              If 'until <target>' is specified, then it migrates all available versions up to and including <target>
```
vcrfxia, line 373: This clarification / update is based on the conversation in https://github.com/confluentinc/ksql/pull/6721#discussion_r537631166

design-proposals/klip-42-schema-migrations-tool.md
```diff
@@ -363,30 +370,33 @@ Commands
...
   
   info        Displays information about the current and available migrations
   
-  baseline    Sets the current schema to the specified version
```
vcrfxia, line 370: This removal is based on the conversation in https://github.com/confluentinc/ksql/pull/6721#discussion_r545181281

design-proposals/klip-42-schema-migrations-tool.md
```diff
@@ -363,30 +370,33 @@ Commands
...
   
-  -d, --define       Define variables for the client session (equivalent to the DEFINE statement).
+  -d, --define       Define variables for the client session (equivalent to the DEFINE statement). Only applicable for `apply`.
   
   --dry-run          Simulates what migrations will be applied in the cluster. Dry-runs do not execute or simulate each migration
                      statement. Only displays the schema updates that may take place in the cluster without modifying the cluster.
+                     Applicable for `apply` as well as `initialize` and `clean` (to confirm the statements that will be executed to
+                     manipulate the metadata stream and table).               
  
+  --out-of-order     Allow out of order migrations. Only applicable for `apply`. This does not affec the behavior of `next` but will affect `all` and `until`.``
```
vcrfxia, line 398: Can someone sanity check that it makes sense for this flag to affect `all` and `until` but not `next`?

design-proposals/klip-42-schema-migrations-tool.md
```diff
@@ -399,18 +409,101 @@ are the basic configurations that the tool should support.
...
 
+The default stream and table metadata topic names will be `{clusterID}ksql_{StreamOrTableName}`, where the migrations
+tool with determine the ksqlDB cluster ID by accessing the `/info` endpoint of the ksqlDB server. If desired,
+these topic names may be overridden by the user.
+
 Note: The command line options and other configurations will also be defined during the implementation.
 
+### Migration state transitions and potential race conditions
+
+Below we have two potential options for how to manage migration state.
```
vcrfxia, line 439: Do people have thoughts between these two options? My preference is to start with the latter since it's simpler.

ksqldb-tools/src/main/java/io/confluent/ksql/tools/migrations/commands/BaseCommand.java
```diff
@@ -38,7 +38,7 @@
   protected String configFile;
 
   @Option(
-      name = {"-n", "--dry-run"},
```
vcrfxia, line 41: Allowing `-n` for this flag is extremely confusing since `-n` is also used for `--next` as an option in the `apply` command.

vcrfxia (commented):
@spena @colinhicks @agavra can you please comment on these additional clarifications/details to KLIP 42 when you get a chance? I want to check we're on the same page regarding design before I start the implementation. Thanks! 

cc @MichaelDrogalis @derekjn who also commented on the original KLIP.

design-proposals/klip-42-schema-migrations-tool.md
colinhicks: My understanding was that `Pending` represents migrations which have been defined but not yet applied, where `Running` represents a single in-progress migration.

design-proposals/klip-42-schema-migrations-tool.md
colinhicks: I think this makes sense because the single, next migration would always be applied by order.

design-proposals/klip-42-schema-migrations-tool.md
spena: I see the confusion. As @colinhicks mentioned, the `Pending` represents migrations not applied yet. However, this `Pending` status will not form part of the migration schema stream/table because those migrations hasn't run yet. The `Pending` is just a status that will be displayed by `ksql-migrations info` command, which looks at current migrations (in the schema table) and available migrations (in the local directory).

vcrfxia: OK, I will update the example flows below to use `Running` instead of `Pending`, and still leave `Pending` as a possible status, the only use for which is to show in the output of `info` for migrations which have yet to be applied. Thanks for clarifying!

design-proposals/klip-42-schema-migrations-tool.md
```diff
@@ -168,9 +170,12 @@ Current version of schema: 1
 Migrating schema to version 2 - Add users
 ```
 
-All migration files will support only integer versions (no decimal versions). Integer versions are easier to sort when are found in the file name. Also, there are
+All migration files will support only integer versions (no decimal versions). Integer versions are easier to sort when found in the file name. Also, there are
 too few cases that require decimal versioning (i.e. `1.1`) in schema changes. We don't expect users to use decimal versions on ksqlDB migrations.
 
+Version numbers need not be consecutive. In the event that earlier version numbers are added after a later version has already been migrated
+to, the tool will ignore the newly added versions by default, though we will support a flag to change this behavior and enable out-of-order migrations.
```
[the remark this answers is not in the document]
vcrfxia: Flyway allows this: https://flywaydb.org/documentation/configuration/parameters/outOfOrder

If we think it's confusing to allow out of order migrations, we can require that version numbers be consecutive. If we do this, then we should remove the `-v` option on `initialize` since the initial state must always have version 1 (or 0? Is 0 is valid version number?)

design-proposals/klip-42-schema-migrations-tool.md
spena: I'd like to keep this simple too. I raised this race condition issue with @MichaelDrogalis during the KLIP review, and we thought we could keep it simple for the first version of the tool, and work on improving the race conditions in the future. Upgrades are usually run by one user or automated tool during an software upgrade process, so I don't expect many issues with this.

design-proposals/klip-42-schema-migrations-tool.md
```diff
@@ -399,18 +409,101 @@ are the basic configurations that the tool should support.
...
+of race conditions if a user attempts to perform multiple migrations at once and opt for documentation
+warning about this instead.
+
+The flow for applying a migration would be:
+1. Check `CURRENT` state is either `Migrated` or `Error` (i.e., not `Running`), by issuing a pull query. Return error and abort if `Running`.
+1. Validate checksums for applied migrations, by issuing pull queries for `status='Migrated'`.
+1. Check `CURRENT` state has not changed from above, with another pull query. Return error and abort if a change is detected.
+1. Produce `Running` state to migrations stream.
+1. Perform migration. 
+1. If successful, produce `Migrated`, else `Error`.
```
spena, line 482: One extra step to consider is to check (with a pull query) that a schema version has been migrated before continuing applying the next schema version. You could retry this for sometime in case there is a pull query lag. This to ensure the node where the migration process is running will be in a good state at the end.

design-proposals/klip-42-schema-migrations-tool.md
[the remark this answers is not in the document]
spena: Oh, I didn't read that part. Then let's do it :)

design-proposals/klip-42-schema-migrations-tool.md
vcrfxia: Yeah, the migration status flows I wrote were only for single migrations. If we're applying multiple, we can just repeat the last four steps without re-doing the initial validation.

design-proposals/klip-42-schema-migrations-tool.md
vcrfxia: Sounds good, I just wanted to confirm that we're OK with these possible race conditions. Sounds like we are so that's great.

spena (approved):
LGTM then

design-proposals/klip-42-schema-migrations-tool.md
agavra: chiming in here as well, I'm OK with the potential race condition b/c of the justification that it's either run by a human or an automated tool. I don't expect a high throughput of users executing concurrent migrations (and it's not worth the ROI to implement something more complicated that would also make the minimum use requirements more difficult - e.g. I think you need to ensure your broker is set up with the correct configs to use transactions and you must have multiple brokers for it to "work" reliably)

agavra:
LGTM too

guozhangwang (commented):
@vcrfxia sorry I'm late on pushing my comments (I left them a couple of days ago and apparently I forgot to click `submit review`..) my main suggestion is to decouple the "states" from the migration_event log format, from the "states" displayed in the `info` cmd. LMK WDYT?

design-proposals/klip-42-schema-migrations-tool.md
```diff
@@ -399,18 +412,109 @@ are the basic configurations that the tool should support.
...
+running migrations simultaneously.
+
+#### Use Kafka transactions to avoid race conditions
+
+To avoid the possibility of multiple migrations being performed at the same time, we can emulate how the
+ksqlDB command topic uses Kafka transactions to ensure that only one migration takes place at a time. 
+We'd set the producer's transactional ID based on the ksqlDB cluster's service ID (which the migrations
+tool can obtain via the `/info` endpoint of the ksqlDB server), and the flow for applying a migration would be:
+1. Begin transaction.
+1. Check `CURRENT` state is either `Migrated` or `Error` (i.e., not `Running`), by creating a Kafka consumer. Return error and abort if `Running`.
```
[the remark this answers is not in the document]
vcrfxia: > nit: all steps are numbered 1.?

That's markdown syntax for creating a numbered list. You'll see it looks fine when rendered :) https://github.com/confluentinc/ksql/blob/master/design-proposals/klip-42-schema-migrations-tool.md#use-kafka-transactions-to-avoid-race-conditions

> Also I'm wondering what's the default "transaction.timeout" config in ksqlDB? Reading data within a producer txn could take some time and may cause the txn to expire; how is that resolved e.g. in command topic today?

I don't see anywhere ksqlDB sets the config so I assume that means we're using the default of 1 minute. If we hit the timeout when submitting commands to the command topic, the request is aborted and it's up to the user to try again. This might happen if the server that received the request is not caught up on processing the command topic: https://github.com/confluentinc/ksql/blob/e2b26c59667f605f9742b0dbf9780dee17c1f58e/ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/computation/DistributingExecutor.java#L141-L142

design-proposals/klip-42-schema-migrations-tool.md
vcrfxia: That's an interesting idea. It sounds like we're in agreement that the three states we need internally (in the metadata stream/table) are `running`, `complete`, and `error` (actual names subject to change) and the open discussion is about what states the user should see. The user states must include `running`, `complete`, and `pending` (not yet started), where again the actual names are subject to change. Is it useful for the user to see `error` (meaning a migration was attempted but failed)? I think the answer is yes because there should have a way to see the error message associated with a failure other than the output of the `apply` command itself. I'm imagining another column in the output of `info` called `failure reason` or similar, so we can save failures and show them to the user (rather than requiring them to re-run the `apply` command to remember why the command failed). This column will have to be added to the metadata stream/table as well. WDYT?

guozhangwang: That sounds good to me.

One thing I'd like to emphasize still, is that the output `info` should only contain one line per version, whereas the migration_events captures a whole history of migration actions, hence each version could have multiple records. I'd want to point out that `info` does not necessarily need to show the whole history, but only the latest snapshot of the migration status per version.

vcrfxia: We are in agreement here. `info` will only print one line per version (as discovered in the `/migrations` directory). The line printed will be the latest version, as pulled from the metadata table. `/info` will not consume from the metadata stream at all.

PR #6966, opened 2021-02-08
feat: add lambda type checking without adding lambda sql type

lct45:
### Description 
Resolves unknown types in lambda's mainly through the `ExpressionTypeManager` and passes additional information in the SqlArgument to the UDF to resolve function type issues there, instead of adding a new `SqlType.LAMBDA`

The sequence of PR's to review is:

1. https://github.com/confluentinc/ksql/pull/6989
2. https://github.com/confluentinc/ksql/pull/7011
3. This PR

### Testing done 

### Reviewer checklist


ksqldb-common/src/main/java/io/confluent/ksql/function/UdfFactory.java
```diff
@@ -79,6 +80,10 @@ public String toString() {
         + '}';
   }
 
+  public synchronized KsqlScalarFunction getUdfFunction(final List<SqlArgument> argTypes) {
```
lct45, line 83: Added another `getFunction` here instead of modifying the original because table functions still use the other `getFunction` method

ksqldb-common/src/main/java/io/confluent/ksql/function/UdfIndex.java
```diff
@@ -168,6 +197,37 @@ T getFunction(final List<SqlType> arguments) {
         .orElseThrow(() -> createNoMatchingFunctionException(arguments));
   }
 
+  private void getUdfCandidates(
+      final List<SqlArgument> arguments,
```
lct45, line 201: Similarly to above, kept a `SqlArgument` track and a `SqlType` track because of other function types that still use `SqlType`. It's an option to transition all functions to passing in `SqlArgument` but would be a large refactor

ksqldb-common/src/main/java/io/confluent/ksql/function/UdfIndex.java
```diff
@@ -168,6 +197,37 @@ T getFunction(final List<SqlType> arguments) {
...
+    if (argIndex == arguments.size()) {
+      if (current.value != null) {
+        candidates.add(current);
+      }
+      return;
+    }
+    final SqlType arg = arguments.get(argIndex).getSqlType();
+    for (final Entry<Parameter, Node> candidate : current.children.entrySet()) {
+      final Map<GenericType, SqlType> reservedCopy = new HashMap<>(reservedGenerics);
+      if (candidate.getKey().type instanceof LambdaType) {
```
lct45, line 217: break out here into a `lambdaType` track and a `SqlType` track

ksqldb-common/src/main/java/io/confluent/ksql/function/UdfFactory.java
```diff
@@ -79,6 +80,10 @@ public String toString() {
...
+    return udfIndex.getUdfFunction(argTypes);
+  }
+
   public synchronized KsqlScalarFunction getFunction(final List<SqlType> argTypes) {
```
lct45, line 87: Currently still used for testing purposes, _should_ be removable after tests are updated

ksqldb-common/src/main/java/io/confluent/ksql/function/UdfIndex.java
```diff
@@ -142,6 +145,32 @@ void addFunction(final T function) {
     curr.update(function, order);
   }
 
+  T getUdfFunction(final List<SqlArgument> arguments) {
```
lct45, line 148: Created a separate `SqlArgument` track because `TableFunctionFactory` still uses `getFunction` with a `List<SqlType>` passed in.

ksqldb-engine/src/main/java/io/confluent/ksql/function/FunctionLoaderUtils.java
```diff
@@ -177,14 +181,31 @@ static SchemaProvider handleUdfReturnSchema(
       final Map<GenericType, SqlType> genericMapping = new HashMap<>();
       for (int i = 0; i < Math.min(parameters.size(), arguments.size()); i++) {
         final ParamType schema = parameters.get(i);
```
lct45, line 183: I looked into changing this to also accept a `List<SqlArgument>` but it seemed borderline impossible as it involved changing the schema provider which caused somewhat of a chain reaction of other necessary changes. This way of checking inputs seemingly works although isn't very pretty. Open to ideas for how to improve this

guozhangwang:
Both PRs with 1200+ LOC :P

ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/resources/streaming/StreamedQueryResource.java
```diff
@@ -181,16 +181,23 @@ public EndpointResponse streamQuery(
...
+    try {
+      throwIfNotConfigured();
+      activenessRegistrar.updateLastRequestTime();
 
-    final PreparedStatement<?> statement = parseStatement(request);
+      final PreparedStatement<?> statement = parseStatement(request);
 
-    CommandStoreUtil.httpWaitForCommandSequenceNumber(
-        commandQueue, request, commandQueueCatchupTimeout);
+      CommandStoreUtil.httpWaitForCommandSequenceNumber(
```
stevenpyzhang, line 191: nit: we should remove this file from the PR (sorry leftover from my proof of concept)

ksqldb-udf/src/main/java/io/confluent/ksql/schema/ksql/SqlArgument.java
```diff
@@ -0,0 +1,39 @@
...
+ * WARRANTIES OF ANY KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations under the License.
+ */
+
+package io.confluent.ksql.schema.ksql;
+
+import io.confluent.ksql.schema.ksql.types.SqlLambda;
+import io.confluent.ksql.schema.ksql.types.SqlType;
+
+public class SqlArgument {
```
stevenpyzhang, line 21: Can we add javadocs on why we're using this (a wrapper for FunctionCall argument type)?

ksqldb-udf/src/main/java/io/confluent/ksql/schema/ksql/types/SqlLambda.java
```diff
@@ -0,0 +1,77 @@
...
+
+import static java.util.Objects.requireNonNull;
+
+import com.google.errorprone.annotations.Immutable;
+import io.confluent.ksql.schema.utils.FormatOptions;
+import java.util.List;
+import java.util.Objects;
+
+@Immutable
+public final class SqlLambda {
```
stevenpyzhang, line 26: javadocs

ksqldb-execution/src/main/java/io/confluent/ksql/execution/util/ExpressionTypeManager.java
```diff
@@ -96,12 +102,26 @@ public ExpressionTypeManager(
   public SqlType getExpressionSqlType(final Expression expression) {
```
stevenpyzhang, line 102: We can just reuse the method below in this function body

`getExpressionSqlType(expression, Collections.emptyMap(), Collections.emptyList())`

ksqldb-execution/src/main/java/io/confluent/ksql/execution/codegen/CodeGenRunner.java
```diff
@@ -274,5 +331,11 @@ private void addRequiredColumn(final ColumnName columnName) {
           column.index()
       );
     }
+
+    private Boolean shouldSetInputType(final FunctionCall node, final CodeGenTypeContext context) {
```
stevenpyzhang, line 335: I don't think we should be hard coding function names in here to know to put arguments into the context. Because of the syntax rule, we're guaranteed that only Lambdas will be at the end of a function call and no more arguments will be after them. We should just be able to put any argument we encounter into the context.

ksqldb-execution/src/main/java/io/confluent/ksql/execution/codegen/SqlToJavaVisitor.java
```diff
@@ -224,12 +230,44 @@ public String process(final Expression expression) {
...
   private String formatExpression(final Expression expression) {
+    final SqlToJavaTypeContext context = new SqlToJavaTypeContext();
     final Pair<String, SqlType> expressionFormatterResult =
-        new Formatter(functionRegistry).process(expression, null);
+        new Formatter(functionRegistry).process(expression, context);
     return expressionFormatterResult.getLeft();
   }
 
-  private class Formatter implements ExpressionVisitor<Pair<String, SqlType>, Void> {
+  private static class SqlToJavaTypeContext {
```
stevenpyzhang, line 239: The context currently between ExpressionTypeManager/SqlToJavaVisitor/CodeGenRunner does the exact same thing in each respective class, can we just have a singular context for now that can be used across all 3? There's nothing really differentiating the contexts currently so I don't see why we'd need to have a specific context class for each

ksqldb-execution/src/main/java/io/confluent/ksql/execution/util/ExpressionTypeManager.java
```diff
@@ -143,9 +192,8 @@ public Void visitArithmeticUnary(
     public Void visitLambdaExpression(
         final LambdaFunctionCall node, final ExpressionTypeContext context
     ) {
+      context.mapInputTypes(node.getArguments());
```
stevenpyzhang, line 195: I think we should throw an exception if there aren't an equal number of arguments from the lambda function call and the number of unmapped input types already in the context.

For example:
What if the user passes in MAP_REDUCE(map, state, (x,y) => ...)

We might be handling this already in `visitFunctionCall` when we get the Function from `UdfIndex` but we can be more specific about that scenario

ksqldb-common/src/main/java/io/confluent/ksql/function/UdfIndex.java
stevenpyzhang: Can we just change UdfIndex so that
```
  T getFunction(final List<SqlType> arguments) {
  // code to transform the list of SqlType to list of SqlArgument
  getUdfCandidates(listOfSqlArgument);
}

**Review by stevenpyzhang — COMMENTED** [review]
Left some initial thoughts

File: ksqldb-execution/src/main/java/io/confluent/ksql/execution/codegen/CodeGenRunner.java (hunk 11 above)
Comment (lct45) [reply ksqldb-execution/src/main/java/io/confluent/ksql/execution/codegen/CodeGenRunner.java:335]: I agree hard coding is unideal, but how does knowing lambdas are at the end help us determine what arguments to put into the context? The issue for `reduce_map(map, s, (k,v,s) ...)` is that we want to store the input type of `s` but don't want to double store x in something like `transform_array(arr, x => ucase(x))`, how will we distinguish between the cases?

**Comment by lct45** [comment]
@guozhangwang both PR are overly large, the hope was to pick the type checking approach we're going with and then I can break it out into manageable PRs

File: ksqldb-common/src/main/java/io/confluent/ksql/function/UdfIndex.java (hunk 2 above)
Comment (lct45) [reply ksqldb-common/src/main/java/io/confluent/ksql/function/UdfIndex.java:201]: We could for the current `getFunction(final List<SqlType> arguments)` but we'd still need one that accepts `SqlArguments` since the information we need for lambdas is only in `ExpressionTypeManager`, but we can do this approach for `public synchronized KsqlScalarFunction getFunction(final List<SqlType> argTypes)`

File: ksqldb-execution/src/main/java/io/confluent/ksql/execution/codegen/SqlToJavaVisitor.java (hunk 12 above)
Comment (lct45) [reply ksqldb-execution/src/main/java/io/confluent/ksql/execution/codegen/SqlToJavaVisitor.java:239]: I think it can definitely be shared between `SqlToJavaVisitor` and `CodeGenRunner`. The context in `ExpressionTypeManager` was heavily relied on before adding the others though and has some extra methods to resolve sql types so I lean towards leaving that one separate, WDYT?

File: ksqldb-execution/src/main/java/io/confluent/ksql/execution/util/ExpressionTypeManager.java (hunk 13 above)
Comment (lct45) [reply ksqldb-execution/src/main/java/io/confluent/ksql/execution/util/ExpressionTypeManager.java:195]: Yeah, I think that makes sense. This definitely relies on getting the number in inputs right from the `ExpressionTypeManager` which I think is somewhat of an open question if we're avoiding hard coding function names

File: ksqldb-execution/src/main/java/io/confluent/ksql/execution/codegen/CodeGenRunner.java (hunk 11 above)
Comment (stevenpyzhang) [reply ksqldb-execution/src/main/java/io/confluent/ksql/execution/codegen/CodeGenRunner.java]: Once lambda function start appearing, we should enforce that the number of args for each of the lambdas matches what we've stored so far.

So if we had a map and then another argument before lambdas, we store two input types from the map (assume order is key then value) and one input type for the other argument, we should make sure that all lambdas that occur afterward have 3 arguments so that we can map it exactly to what we've stored.

In the transform_array case, we encounter only the array before the lambda so we have one input type stored and wouldn't expect the lambda to be something like `(x,y) =>...`

Comment (stevenpyzhang) [reply ksqldb-execution/src/main/java/io/confluent/ksql/execution/codegen/CodeGenRunner.java]: It shouldn't matter what the function name is, we should just be keeping track of the input types we've encountered while processing the function's arguments

File: ksqldb-execution/src/main/java/io/confluent/ksql/execution/util/ExpressionTypeManager.java [hunk 14]
```diff
@@ -94,14 +101,25 @@ public ExpressionTypeManager(
...
 
   public SqlType getExpressionSqlType(final Expression expression) {
+    return getExpressionSqlType(expression, Collections.emptyMap(), Collections.emptyList());
+  }
+
+  public SqlType getExpressionSqlType(
+      final Expression expression, final Map<String, SqlType> inputMapping, final List<SqlType> inputTypes
+  ) {
     final ExpressionTypeContext expressionTypeContext = new ExpressionTypeContext();
+    expressionTypeContext.setLambdaTypes(inputMapping);
```
guozhangwang, line 111: For my own understanding: when we call setLambdaTypes here, is the mapping already populated or not? If yes, then we would not need to `mapInputTypes` anymore; if not, why?

guozhangwang (commented):
I think this is a cleaner approach compared with #6967 to not introduce a `LambdaType` as the returned column type.

I just have one meta question for my own understanding: I cannot fully get the difference between the `TypeContext` and the `ExpressionTypeContext` within `ExpressionTypeManager`. More specifically, when does the generic -> sqlType population happens (inside TypeContext, before visitor.process(), or inside ExpressionTypeContext, after visitor.process())? It seems that it differs between UDFs and lambdas, but I'm not sure why it's the case. @lct45 could you elaborate a bit more to me?

ksqldb-execution/src/main/java/io/confluent/ksql/execution/codegen/CodeGenRunner.java
lct45: Mmm I see what you mean, that should work

ksqldb-execution/src/main/java/io/confluent/ksql/execution/util/ExpressionTypeManager.java
lct45: The mapping is already set here, but it's set in either `CodeGenRunner` or `SqlToJavaVisitor`. this is a way to pass the mapping we've already done in a `TypeContext` into a new `ExpressionTypeContext` that's used to resolve ultimately resolve the types within `ExpressionTypeManager`. `ExpressionTypeManager` is used by both `SqlToJavaVisitor` and `CodeGenRunner` to do type resolution.

I think this ties into what @stevenpyzhang commented earlier, that it may make sense to combine `ExpressionTypeContext` and `TypeContext`, as it would make passing this information infinitely easier. I think I'll go ahead and do that unless @agavra has any objections.

Currently, the difference between the contexts is one, `ExpressionTypeContext`, is just used in `ExpressionTypeManager`. This existed before lambdas for passing around sql return types. `TypeContext`, as the code stands now, is used by `SqlToJavaVisitor` and `CodeGenRunner` to store information relevant to lambdas and pass it to the relevant methods. It doesn't contain the basic `sqlType` storage that `ExpressionTypeManager` does. `ExpressionTypeManager` doesn't currently use `TypeContext` at all. 

For lambdas, the generic -> sqlType population for the variables (x in x => ucase(x)) happens when the context (either `TypeContext` or `ExpressionTypeContext`) is passed information in the `visitLambdaExpression` function in `ExpressionTypeManager`, `CodeGenRunner`, and `SqlToJavaVisitor`. This happens after `visitor.process()` for each of the individual flows through one of the three relevant classes.

Let me know if this was more confusing than enlightening, I'm happy to have another go at explaining the relationship between all these pieces

agavra: Haven't looked at the code yet, but from your description combining the two makes sense to me!

ksqldb-engine/src/main/java/io/confluent/ksql/engine/rewrite/AstSanitizer.java
```diff
@@ -221,12 +221,11 @@ public static Statement sanitize(final Statement node, final MetaStore metaStore
       ctx.process(expression.getBody());
       ctx.getContext().removeLambdaArgs();
       ctx.getContext().addLambdaArgs(previousLambdaArgs);
-
       return visitExpression(expression, ctx);
     }
   }
 
-  private static class SanitizerContext {
+  public static class SanitizerContext {
```
stevenpyzhang, line 228: nit: sorry, another random tidbit leftover from my initial design, we should remove this

ksqldb-execution/src/main/java/io/confluent/ksql/execution/function/UdfUtil.java
```diff
@@ -39,6 +39,7 @@
 import java.util.function.BiFunction;
 import java.util.function.Function;
 import org.apache.kafka.connect.data.Struct;
+import sun.reflect.generics.reflectiveObjects.ParameterizedTypeImpl;
```
stevenpyzhang: nit: remove this line

ksqldb-execution/src/main/java/io/confluent/ksql/execution/codegen/helpers/TriFunction.java
```diff
@@ -28,4 +28,4 @@
     Objects.requireNonNull(after);
     return (A a, B b, C c) -> after.apply(apply(a, b, c));
   }
-}
+}
```
stevenpyzhang, line 31: nit: remove

ksqldb-execution/src/main/java/io/confluent/ksql/execution/codegen/helpers/LambdaUtil.java
```diff
@@ -29,7 +29,7 @@
 
   private LambdaUtil() {
   }
-
+
```
stevenpyzhang, line 32: nit: extra spacing

ksqldb-execution/src/test/java/io/confluent/ksql/execution/codegen/helpers/LambdaUtilTest.java
```diff
@@ -44,6 +45,7 @@ public void shouldGenerateFunctionCode() {
         .function(argName, argType, argName + " + 1");
 
     // Then:
+    System.out.println(javaCode);
```
stevenpyzhang, line 48: nit: remove

ksqldb-udf/src/main/java/io/confluent/ksql/function/TriFunction.java
```diff
@@ -0,0 +1,31 @@
...
+ * specific language governing permissions and limitations under the License.
+ */
+
+package io.confluent.ksql.function;
+
+import java.util.Objects;
+import java.util.function.Function;
+
+@FunctionalInterface
+public interface TriFunction<T,U,V,R> {
```
stevenpyzhang, line 22: I'm actually a bit confused as to why this is a new change, I thought I merged in Trifunction previously...
Can you check if we have duplicates and if so, remove this one?

ksqldb-engine/src/main/java/io/confluent/ksql/engine/EngineContext.java
```diff
@@ -211,6 +211,7 @@ private ParsedStatement substituteVariables(
     } catch (final KsqlStatementException e) {
       throw e;
     } catch (final Exception e) {
+      e.printStackTrace();
```
stevenpyzhang, line 214: nit: remove before merging

stevenpyzhang (commented):
some more nits

ksqldb-execution/src/main/java/io/confluent/ksql/execution/util/ExpressionTypeManager.java
guozhangwang: Thanks @lct45 , I think my confusion was that:

If in line 111 here, which is before we start visiting the node, the genericName -> sqlType `inputMapping` we are set via `expressionTypeContex.setLambdaTypes()` in is already *fully* populated, then inside `visitLambdaExpression` (line 200) which is part of the visiting traversal, we should not need to call `mapInputTypes` which is trying to build the mapping again.

Now I think I get your point, which is that the `inputMapping` passed in via `getExpressionSqlType` is not fully populated, i.e. it only contains the mapping entries for variables outside the lambda function, but not contain entries for variables that are only declared inside the lambda function. Is that right?

guozhangwang: In that case, if we consolidate these two into a single `TypeContext`, would a single `TypeContext` object be used in multiple lambda expressions? I'm asking this because if that's the case, then its `inputTypes` and `lambdaTypeMapping` would be a global structure containing mappings for multiple expressions, and we cannot enforce `inputTypes.size() == argumentList.size()` in `mapInputTypes`.

ksqldb-engine/src/main/java/io/confluent/ksql/engine/rewrite/AstSanitizer.java
lct45: IDE is complaining when I try to change it back to public

ksqldb-execution/src/main/java/io/confluent/ksql/execution/codegen/TypeContext.java
```diff
@@ -0,0 +1,58 @@
...
+    for (int i = 0; i < argumentList.size(); i++) {
+      this.lambdaTypeMapping.putIfAbsent(argumentList.get(i), inputTypes.get(i));
+    }
+  }
+
+  public SqlType getLambdaType(final String name) {
+    return lambdaTypeMapping.get(name);
+  }
+
+  Map<String, SqlType> getLambdaTypeMapping() {
```
guozhangwang, line 51: Seems this function is not used by anyone yet? And today we area only checking `notAllInputsSeen` below.

ksqldb-execution/src/main/java/io/confluent/ksql/execution/codegen/TypeContext.java
```diff
@@ -0,0 +1,58 @@
...
+
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+
+public class TypeContext {
+  private SqlType sqlType;
+  private final List<SqlType> inputTypes = new ArrayList<>();
+  private final Map<String, SqlType> lambdaTypeMapping = new HashMap<>();
```
guozhangwang, line 13: nit: `lambdaInputTypeMapping`.

ksqldb-execution/src/main/java/io/confluent/ksql/execution/util/ExpressionTypeManager.java
lct45: @stevenpyzhang may be able to weigh in more here, but I think all the lambdas inside of a UDF should have the same number of inputs so I don't think there would be a concern if the TypeContext is shared between them. The example I'm thinking of is from the KLIP, `transform_map(map, (k, v) => new_k, (k, v) => new_v)`, so having a global map of `k,v` should be fine. I did just run through it though, and the context is shared. This is necessary for the input though, since in a lambda it'll be one input mapped to the lambda variables (i.e. we only see the map once even if we have 2 lamda expressions). This does raise the issue of variable names. Should users be able to do a lambda like `transform_map(map, (x,y) => x + y, (k,v) => k - v)`? Because I think you're right @guozhangwang , right now I don't  it would accept that

stevenpyzhang: I think `transform_map(map, (x,y) => x + y, (k,v) => k - v)` should be supported. We could just make a copy of the TypeContext and pass a fresh instance to each argument that we visit so that we can have different mappings for each Lambda node.

ksqldb-common/src/main/java/io/confluent/ksql/function/types/ParamTypes.java
```diff
@@ -61,7 +83,11 @@ public static boolean areCompatible(
       final SqlMap sqlType = (SqlMap) argumentSqlType;
       final MapType mapType = (MapType) declared;
       return areCompatible(SqlArgument.of(sqlType.getKeyType()), mapType.key(), allowCast)
-          && areCompatible(SqlArgument.of(sqlType.getValueType()), mapType.value(), allowCast);
+          && areCompatible(
+          SqlArgument.of(sqlType.getValueType()),
```
lct45, line 87: With lambdas above we use `areCompatible(SqlType, ParamType)` which wraps to a `SqlArgument` and then passes down to `areCompatible(SqlArgument, ParamType, AllowCasts)`. Seems like we should either use the wrapper `areCompatible` for arrays and maps too or get rid of the extra class and set everything to `SqlArgument` here. Thoughts? @stevenpyzhang

JimGalasyn (approved):

ksqldb-common/src/main/java/io/confluent/ksql/function/types/ParamTypes.java
stevenpyzhang: Let's just change the function to take SqlArgument also.

ksqldb-udf/src/main/java/io/confluent/ksql/schema/ksql/SqlArgument.java
```diff
@@ -15,31 +15,48 @@
...
+ * lambdas as an argument. This class allows us to properly find the matching UDF and
+ * resolve the return type for lambda UDFs based on the given sqlLambda.
  */
 public class SqlArgument {
 
   private final SqlType sqlType;
+  private final SqlLambda sqlLambda;
 
-  public SqlArgument(final SqlType type) {
+  public SqlArgument(final SqlType type, final SqlLambda lambda) {
```
stevenpyzhang, line 32: We should change the sqlType and sqlLambda variables to be Optionals so we don't need to do null checks where SqlArgument is used

ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/resources/streaming/StreamedQueryResource.java
```diff
@@ -393,8 +393,8 @@ private EndpointResponse handlePrintTopic(
       final String reverseSuggestion = possibleAlternatives.isEmpty()
           ? ""
           : possibleAlternatives.stream()
-              .map(name -> "\tprint " + name + ";")
-              .collect(Collectors.joining(
+            .map(name -> "\tprint " + name + ";")
```
stevenpyzhang, line 396: nit: spaces

ksqldb-engine/src/test/java/io/confluent/ksql/function/UdfLoaderTest.java
```diff
@@ -187,6 +190,45 @@ public void shouldLoadDecimalUdfs() {
...
+  @Test
+  public void shouldLoadLambdaReduceUdfs() {
+    // Given:
+    final SqlLambda schema =
+        SqlLambda.of(
+            ImmutableList.of(SqlTypes.INTEGER, SqlTypes.INTEGER, SqlTypes.INTEGER),
+            SqlTypes.INTEGER);
+
+    // When:
+    final KsqlScalarFunction fun = FUNC_REG.getUdfFactory(FunctionName.of("reduce_map"))
```
stevenpyzhang, line 202: nit: Let's just use some random generic function names for these tests

ksqldb-execution/src/main/java/io/confluent/ksql/execution/codegen/SqlToJavaVisitor.java
```diff
@@ -458,7 +499,9 @@ private String formatExpression(final Expression expression) {
           paramType = function.parameters().get(i);
         }
 
-        joiner.add(process(convertArgument(arg, sqlType, paramType), context).getLeft());
+        final Pair<String, SqlType> pair =
```
stevenpyzhang, line 502: nit: this change should be unecessary

ksqldb-udf/src/main/java/io/confluent/ksql/schema/ksql/SqlArgument.java
stevenpyzhang: Now that I'm thinking about this more, I think we should actually have SqlArgument be a base class and two classes extend it SqlArgumentType (which has a SqlType) and SqlArgumentLambda(which has a SqlLambda). There are certain portions of the code such as the UDAF's where we now need to do some unnecessary null checks (they should only be supporting SqlArgumentType). The GenericsUtil can still take in SqlArgument and we can use `instance of` checks instead of null checks/optional checks when deciding what behavior to use.

ksqldb-common/src/test/java/io/confluent/ksql/function/UdfIndexTest.java
```diff
@@ -231,6 +238,127 @@ public void shouldChooseCorrectMap() {
...
+                        ImmutableList.of(SqlTypes.BOOLEAN, INTEGER),
+                        INTEGER))))
+    );
+
+    // Then:
+    assertThat(fun1.name(), equalTo(EXPECTED));
+    assertThat(fun2.name(), equalTo(EXPECTED));
+    assertThat(e.getMessage(), containsString("Valid alternatives are:"
+        + lineSeparator()
+        + "expected(MAP<VARCHAR, VARCHAR>, LAMBDA<[VARCHAR, VARCHAR], A>)"));
```
agavra, line 321: if this is shown to users, we might want to consider `toString` on lambdas to be something like `(VARCHAR, VARCHAR) -> A` instead of `LAMBDA<[VARCHAR, VARCHAR], A>` which is somewhat difficult to read

ksqldb-common/src/test/java/io/confluent/ksql/function/UdfIndexTest.java
```diff
@@ -231,6 +238,127 @@ public void shouldChooseCorrectMap() {
     assertThat(fun.name(), equalTo(EXPECTED));
   }
 
+  @Test
+  public void shouldChooseCorrectLambdaFunction() {
```
agavra, line 242: this is kind-of a weird test - we're testing the following signature:
```java
public String foo(List<T> list, Function<A, B> func);
```
by calling
```java
foo(list_of_ints, a_string -> an_int)
```

While there's nothing preventing that from working, I feel like we should be testing the case where input generic of the function should match the generic of the list. To make sure that it's working, we probably want to test that it chooses the right one if we do something more complicated:

```java
public String foo(Map<K, V> map, Function<K, K2> keyMapper);
public String foo(Map<K, V> map, Function<V, V2> valueMapper);
```

If I pass in `MAP<STRING, INTEGER>` and `LAMBDA<[STRING], STRING>` it should give me the first, but if I pass in `LAMBDA<[INTEGER], INTEGER>` it should give me the second.

ksqldb-common/src/test/java/io/confluent/ksql/function/UdfIndexTest.java
```diff
@@ -231,6 +238,127 @@ public void shouldChooseCorrectMap() {
...
+                        ImmutableList.of(SqlTypes.STRING,SqlTypes.STRING, SqlTypes.STRING),
+                        INTEGER)
+                )))
+    );
+
+    // Then:
+    assertThat(e1.getMessage(), containsString("Valid alternatives are:"
+        + lineSeparator()
+        + "other(MAP<A, B>, LAMBDA<[A, B], C>)"));
+    assertThat(e2.getMessage(), containsString("Number of lambda arguments doesn't match between schema and sql type"));
```
agavra, line 359: can we give an error message here that also shows what was passed in (e.g. `LAMBDA<[STRING, STRING, STRING], INTEGER>` in this case)

ksqldb-common/src/test/java/io/confluent/ksql/util/GenericsUtilTest.java
```diff
@@ -178,6 +204,93 @@ public void shouldIdentifyMapGeneric() {
...
+
+    // When:
+    final Exception e = assertThrows(
+        KsqlException.class,
+        () -> GenericsUtil.resolveGenerics(a, instance)
+    );
+
+    // Then:
+    assertThat(e.getMessage(), containsString(
+    "Found invalid instance of generic schema when mapping LAMBDA<[A], A> to Lambda<[DOUBLE], BOOLEAN>. "
```
agavra, line 259: nice! This is an excellent error message (super clear what's going on)

ksqldb-engine/src/main/java/io/confluent/ksql/function/FunctionLoaderUtils.java
```diff
@@ -178,15 +179,18 @@ static SchemaProvider handleUdfReturnSchema(
...
-        final SqlType instance = isVariadic && i == parameters.size() - 1
-            ? SqlTypes.array(arguments.get(i).getSqlType())
-            : arguments.get(i).getSqlType();
-
-        genericMapping.putAll(
-            GenericsUtil.resolveGenerics(schema, SqlArgument.of(instance)));
+        if (schema instanceof LambdaType) {
+          genericMapping.putAll(GenericsUtil.resolveGenerics(schema, arguments.get(i)));
+        } else {
+          // we resolve any variadic as if it were an array so that the type
```
agavra, line 185: we should fail variadic lambdas. I doubt anyone is crazy enough to do this, but let's make sure we're failing it gracefully

ksqldb-execution/src/main/java/io/confluent/ksql/execution/codegen/CodeGenRunner.java
```diff
@@ -185,10 +188,29 @@ public Void visitFunctionCall(final FunctionCall node, final TypeContext context
       final List<SqlArgument> argumentTypes = new ArrayList<>();
       final FunctionName functionName = node.getName();
       for (final Expression argExpr : node.getArguments()) {
-        process(argExpr, null);
-        argumentTypes.add(SqlArgument.of(
-            expressionTypeManager.getExpressionSqlType(argExpr)
-        ));
+        process(argExpr, context);
+        final SqlType newSqlType = expressionTypeManager.getExpressionSqlType(argExpr, context);
```
agavra, line 192:
```suggestion
        final SqlType resolvedArgType = expressionTypeManager.getExpressionSqlType(argExpr, context);
```
nit: "new" is a little confusing for me

ksqldb-execution/src/main/java/io/confluent/ksql/execution/codegen/CodeGenRunner.java
```diff
@@ -185,10 +188,29 @@ public Void visitFunctionCall(final FunctionCall node, final TypeContext context
...
+        // for lambdas - if we see this it's the  array/map being passed in we save the type
```
agavra, line 193: I don't understand this comment, can you flush it out with examples?

ksqldb-execution/src/main/java/io/confluent/ksql/execution/codegen/CodeGenRunner.java
```diff
@@ -185,10 +188,29 @@ public Void visitFunctionCall(final FunctionCall node, final TypeContext context
...
+        if (context.notAllInputsSeen()) {
+          if (newSqlType instanceof SqlArray) {
+            final SqlArray inputArray = (SqlArray) newSqlType;
+            context.addLambdaInputType(inputArray.getItemType());
+          } else if (newSqlType instanceof SqlMap) {
+            final SqlMap inputMap = (SqlMap) newSqlType;
+            context.addLambdaInputType(inputMap.getKeyType());
+            context.addLambdaInputType(inputMap.getValueType());
+          } else {
+            context.addLambdaInputType(newSqlType);
```
agavra, line 203: I feel like we can put all of this into a new method in `TypeContext` called `context#visitType`:
```java
void visitType(SqlType type) {
  if (notAllInputsSeen()) {
      if (newSqlType instanceof SqlArray) {
          final SqlArray inputArray = (SqlArray) type;
          addLambdaInputType(inputArray.getItemType());
      } else if (newSqlType instanceof SqlMap) {
         final SqlMap inputMap = (SqlMap) type;
         addLambdaInputType(inputMap.getKeyType());
         addLambdaInputType(inputMap.getValueType())
      } else {
        addLambdaInputType(type);
    }
  }
}
```

ksqldb-udf/src/main/java/io/confluent/ksql/schema/ksql/SqlArgument.java
stevenpyzhang: One idea for the optional is to have the getter functions like this
```
  public SqlType getSqlTypeOrThrow() {
    if (sqlType.isPresent()) {
      return sqlType.get();
    }
    if (sqlLambda.isPresent()) {
      throw new RuntimeException("Was expecting sql type as an argument");
    }
    return null;
  }

  public Optional<SqlLambda> getSqlLambda() {
    return sqlLambda;
  }

  public SqlLambda getSqlLambdaOrThrow() {
    if (sqlLambda.isPresent()) {
      return sqlLambda.get();
    }
    throw new RuntimeException("Was expecting sql lambda as an argument");
  }
```

We return null from `getSqlTypeOrThrow` if both the type and lambda are missing because this is when the function argument is `NULL`

ksqldb-execution/src/main/java/io/confluent/ksql/execution/codegen/CodeGenRunner.java
```diff
@@ -185,10 +188,29 @@ public Void visitFunctionCall(final FunctionCall node, final TypeContext context
...
       final FunctionName functionName = node.getName();
       for (final Expression argExpr : node.getArguments()) {
-        process(argExpr, null);
-        argumentTypes.add(SqlArgument.of(
-            expressionTypeManager.getExpressionSqlType(argExpr)
-        ));
+        process(argExpr, context);
+        final SqlType newSqlType = expressionTypeManager.getExpressionSqlType(argExpr, context);
+        // for lambdas - if we see this it's the  array/map being passed in we save the type
+        if (context.notAllInputsSeen()) {
```
agavra, line 194: not part of this PR, but I think I might be missing something 
```java
  public boolean notAllInputsSeen() {
    return lambdaInputTypeMapping.size() != lambdaInputTypes.size() || lambdaInputTypes.size() == 0;
  }
```

Can this handle lambdas with 0 arguments properly? If not that can be OK but we should make sure we fail that if anyone tries to register one.

ksqldb-execution/src/main/java/io/confluent/ksql/execution/codegen/SqlToJavaVisitor.java
```diff
@@ -360,15 +364,32 @@ private String formatExpression(final Expression expression) {
     }
 
     @Override
+    // CHECKSTYLE_RULES.OFF: TodoComment
```
agavra, line 367:
```suggestion
```

ksqldb-execution/src/main/java/io/confluent/ksql/execution/codegen/SqlToJavaVisitor.java
```diff
@@ -360,15 +364,32 @@ private String formatExpression(final Expression expression) {
...
+      final Pair<String, SqlType> lambdaBody = process(lambdaFunctionCall.getBody(), context);
+
+      final List<Pair<String, Class<?>>> argPairs = new ArrayList<>();
+      for (final String lambdaArg: lambdaFunctionCall.getArguments()) {
+        argPairs.add(new Pair<>(
+            lambdaArg,
+            SchemaConverters.sqlToJavaConverter().toJavaType(context.getLambdaType(lambdaArg))
+            ));
+      }
+      return new Pair<>(LambdaUtil.function(argPairs, lambdaBody.getLeft()),
```
agavra, line 381:
```suggestion
      return new Pair<>(LambdaUtil.toJavaCode(argPairs, lambdaBody.getLeft()),
```
let's name this method something more meaningful

ksqldb-udf/src/main/java/io/confluent/ksql/schema/ksql/SqlArgument.java
agavra: +1 this makes the contract way clearer

agavra (approved):
Contents LGTM so giving the green check (though wait for Steven's +1 before merging), I left a spattering of nits here and there - nothing major. With regards to these two questions Steven's asked:

>  the usage of SqlArgument as it seems a little hacky with how it is currently.

I like your inline suggestion of making different classes for them and using `instanceof` checks (or using polymorphism to naturally handle it)

> There's a lot of repeated code between SqlToJavaVisitor, CodeGenRunner, ExpressionTypeManager with adding types to the context for use in the child nodes to map lambda inputs to types, but we're not sure if there's any way around it.

Added a suggestion inline, maybe I'm missing something?

ksqldb-engine/src/test/java/io/confluent/ksql/function/UdfLoaderTest.java
lct45: I think you have to use an existing function name so it can find it in the function library, unless there's a way to mock that? When I changed it to `lambda_func` I got `io.confluent.ksql.util.KsqlException: Can't find any functions with the name 'lambda_func'`

ksqldb-common/src/test/java/io/confluent/ksql/function/UdfIndexTest.java
lct45: Do you think we should leave the `LAMBDA` identifier for the new `toString`? Just wondering if users will be able to easily identify if it's `((VARCHAR, VARCHAR) -> A`

ksqldb-common/src/test/java/io/confluent/ksql/function/UdfIndexTest.java
lct45: Just updated (cc @stevenpyzhang )

ksqldb-engine/src/test/java/io/confluent/ksql/function/UdfLoaderTest.java
stevenpyzhang: Oh really? That seems really odd since we'd break these tests if we changed the function name. It feels like we should have these functions mocked. If that's how it is then leaving this is fine then.

ksqldb-execution/src/main/java/io/confluent/ksql/execution/codegen/CodeGenRunner.java
lct45: Hmm... I think it would register more input types than we want it to, but I think lambdas with 0 arguments would still fail. Would that be something like `array_transform(testval1, 5 => 5)`? or `map_transform(MAP(5:=3, 1:=1), () => 2, () => 5)`?

lct45: Just looked back into it, in `LambdaFunctionCall` we check the argument size
`    if (arguments.size() == 0) {
      throw new IllegalArgumentException(
          String.format("Lambda expression must have at least 1 argument. => %s", body.toString()));
    }`

ksqldb-execution/src/main/java/io/confluent/ksql/execution/codegen/CodeGenRunner.java
```diff
@@ -175,20 +176,33 @@ private Visitor() {
...
-        ));
+        final TypeContext childContext = context.getCopy();
+        final SqlType resolvedArgType =
+            expressionTypeManager.getExpressionSqlType(argExpr, childContext);
+        process(argExpr, context.getCopy());
+
+        if (argExpr instanceof LambdaFunctionCall) {
+          argumentTypes.add(
+              SqlArgument.of(
+                  SqlLambda.of(context.getLambdaInputTypes(), childContext.getSqlType())));
```
guozhangwang, line 198: nit: I think this is cleaner to just use `resolvedArgType` as the second parameter, which will be the same as `childContext.getSqlType()`. Ditto on other two classes.

guozhangwang: Also maybe we could consolidate the shared logic in `FunctionCall`:

```
List<SqlArgument> argumentTypes resolveArgumentTypes(TypeContext context)
```

shared among `CodeGenRunner`, `SqlToJavaVisitor` and `ExpressionTypeManager`?

And I'm even wondering, why we need to try to resolve the argument types multiple times, rather than just cache the resolved types inside the `FunctionCall` node itself after done for the first time, and then re-use it in later iterations of traversals?

guozhangwang (commented):
Sorry I'm late on getting a final pass! I took another look, and had a meta question regarding the code gen traversal itself (not related to this PR specifically). 

Filed https://github.com/confluentinc/ksql/issues/7066 to summarize my thoughts.

ksqldb-execution/src/main/java/io/confluent/ksql/execution/codegen/CodeGenRunner.java
stevenpyzhang: Agreed @guozhangwang I think we can definitely optimize these 3 classes to avoid duplicate traversals. Thanks for filing the issue. I think for now since this was an existing issue even before lambdas, it's out of scope for the lambdas implementation and we can follow up on issue you filed in the future.

PR #6972, opened 2021-02-08
chore: close KafkaStreams applications created by sandbox execution in tests

stevenpyzhang:
### Description 
Looks like tests broke after https://github.com/apache/kafka/pull/9978

### Testing done 

### Reviewer checklist

### DIFF
--- a/ksqldb-common/src/main/java/io/confluent/ksql/util/AppInfo.java
+++ b/ksqldb-common/src/main/java/io/confluent/ksql/util/AppInfo.java
@@ -42,9 +42,7 @@ public final class AppInfo {
       final Properties props = new Properties();
       try (InputStream resourceAsStream = AppInfo.class.getResourceAsStream(
           "/git.properties")) {
-        if (resourceAsStream != null) {
-          props.load(resourceAsStream);
-        }
+        props.load(resourceAsStream);
       }
       commitId = props.getProperty("git.commit.id", commitId).trim();
     } catch (final Exception e) {
--- a/ksqldb-engine/src/test/java/io/confluent/ksql/engine/KsqlEngineTest.java
+++ b/ksqldb-engine/src/test/java/io/confluent/ksql/engine/KsqlEngineTest.java
@@ -719,7 +719,7 @@ public class KsqlEngineTest {
         () -> sandbox.execute(
             sandboxServiceContext,
             ConfiguredStatement.of(prepared, SessionConfig.of(KSQL_CONFIG, Collections.emptyMap()))
-        )
+        ).getQuery().ifPresent(QueryMetadata::close)
     );
 
     // Then:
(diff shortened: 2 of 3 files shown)


agavra (approved):
thanks for digging into this @stevenpyzhang

ksqldb-engine/src/test/java/io/confluent/ksql/materialization/ks/KsMaterializationFunctionalTest.java
```diff
@@ -167,6 +167,8 @@ public static void classSetUp() {
   public void setUp() {
     output = KsqlIdentifierTestUtil.uniqueIdentifierName();
 
+    final Random rand = new Random();
+    queryId = new QueryId("static" + rand.nextInt(1000));
```
ableegoldman, line 171: Might be a god idea to do something like what we do in Streams, to ensure a unique test while providing an id with some meaning for debugging purposes. Not necessarily right away, it's better to merge this ASAP to unblock the build -- but you can look at `IntegrationTestUtils#safeUniqueTestName` in Streams and see how it's used to get some idea of what we do

ksqldb-engine/src/test/java/io/confluent/ksql/engine/KsqlEngineTest.java
```diff
@@ -137,7 +137,7 @@ public void setUp() {
 
   @After
   public void closeEngine() {
-    ksqlEngine.close();
+    ksqlEngine.close(true);
```
ableegoldman, line 140: It looks like it's still failing. Is it possible this still isn't ultimately calling `close` on the `KafkaStreams`? (just fyi, it's `KafkaStreams#close` we need to call, since that unlocks the dir, not necessarily `KafkaStreams#cleanUp`. But it's probably good to also call `#cleanup` in tests as well, since it just cleans up whatever state is left behind.

ksqldb-common/src/main/java/io/confluent/ksql/util/AppInfo.java
```diff
@@ -42,9 +42,7 @@
       final Properties props = new Properties();
       try (InputStream resourceAsStream = AppInfo.class.getResourceAsStream(
           "/git.properties")) {
-        if (resourceAsStream != null) {
-          props.load(resourceAsStream);
-        }
+        props.load(resourceAsStream);
```
stevenpyzhang, line 45: this started throwing findbugs error so I changed

ksqldb-engine/src/test/java/io/confluent/ksql/materialization/ks/KsMaterializationFunctionalTest.java
stevenpyzhang: This actually doesn't help with the failures in this test class, because we create the KafkaStreams application twice as part of execution (one is done in a "sandboxed" context first and this first execution still creates the Streams app, but doesn't run it) so even with unique names, all the tests would fail.

mjsax: Would it be possible to use a different query-id in the sandbox?

stevenpyzhang:
closing in favor of https://github.com/confluentinc/ksql/pull/6979

PR #6979, opened 2021-02-09
build: fix build after AK bump

agavra:
### Description 

Fixes the build to make sure that state dirs are properly closed.

### Testing done 

ran `mvn clean install` locally

### Reviewer checklist

### DIFF
--- a/ksqldb-engine/src/main/java/io/confluent/ksql/engine/SandboxedExecutionContext.java
+++ b/ksqldb-engine/src/main/java/io/confluent/ksql/engine/SandboxedExecutionContext.java
@@ -127,11 +127,13 @@ final class SandboxedExecutionContext implements KsqlExecutionContext {
       final ServiceContext serviceContext,
       final ConfiguredKsqlPlan ksqlPlan
   ) {
-    return EngineExecutor.create(
+    final ExecuteResult result = EngineExecutor.create(
         engineContext,
         serviceContext,
         ksqlPlan.getConfig()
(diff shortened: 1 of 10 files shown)


ksqldb-engine/src/main/java/io/confluent/ksql/engine/SandboxedExecutionContext.java
```diff
@@ -127,11 +127,13 @@ public ExecuteResult execute(
       final ServiceContext serviceContext,
       final ConfiguredKsqlPlan ksqlPlan
   ) {
-    return EngineExecutor.create(
+    final ExecuteResult result = EngineExecutor.create(
         engineContext,
         serviceContext,
         ksqlPlan.getConfig()
     ).execute(ksqlPlan.getPlan());
+    result.getQuery().ifPresent(query -> query.getKafkaStreams().close());
```
spena, line 135: Why does it close it during the execution?

spena (approved):
Code looks good. I just have the question about why the query is closed when calling execute().

ksqldb-engine/src/main/java/io/confluent/ksql/engine/SandboxedExecutionContext.java
stevenpyzhang: It's so we don't have to close the KafkaStreams app in all the various locations where we use the sandbox execution.

stevenpyzhang (approved):

PR #6981, opened 2021-02-10
build: fix build after AK bump (#6979)

AlanConfluent:
### Description 
Cherry pick-ish of #6979 

### Testing done 
Unit tests.

### Reviewer checklist

### DIFF
--- a/ksqldb-engine/src/main/java/io/confluent/ksql/engine/SandboxedExecutionContext.java
+++ b/ksqldb-engine/src/main/java/io/confluent/ksql/engine/SandboxedExecutionContext.java
@@ -113,12 +113,14 @@ final class SandboxedExecutionContext implements KsqlExecutionContext {
       final ServiceContext serviceContext,
       final ConfiguredKsqlPlan ksqlPlan
   ) {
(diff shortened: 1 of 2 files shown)


stevenpyzhang (approved):
lgtm if test pass

I'm wondering though, I'm not sure if the rest of the changes outside of the Sandbox context are absolutely necessary since the reason we had to initialize the QueryMetadata in master was because of how CREATE or REPLACE logic works. I don't see any harm in including the refactor 6.0.0 though just to be safe.

AlanConfluent:
> lgtm if test pass
> 
> I'm wondering though, I'm not sure if the rest of the changes outside of the Sandbox context are absolutely necessary since the reason we had to initialize the QueryMetadata in master was because of how CREATE or REPLACE logic works. I don't see any harm in including the refactor 6.0.0 though just to be safe.

@stevenpyzhang You're right.  I started going down that path of porting the whole change and I was having to port lots of depended upon behavior that wasn't even part of the PR. It was getting nasty and would have required more of a review...

So I switched to just the sandbox change and all of the tests pass.  Going to merge this.

PR #6985, opened 2021-02-10
fix: get all available restore commands even on poll timeout

rodesai:
### Description 
Our restore logic differentiates on restore behaviour vs "live" behaviour
to avoid making some newly added referential integrity checks and therefore
ensure we execute anything that was already logged. While this isn't the
best way to guarantee compatibility, it's what we're doing now and we have
a problem: sometimes the restore poll times out and returns no records,
so we don't execute any "restore" phase. This patch fixes this by first
sampling the end offset of the log and ensuring that all rows between the
begin and end are returned.

### Testing done 
Unit tests for command store. Restart of a running server.

### Reviewer checklist


agavra (approved):
thanks for the quick fix to the incident!

PR #6989, opened 2021-02-10
feat: adding TypeContext for Lambdas

lct45:
### Description 
Updates the context to contain all the information we need for the upcoming lambda type resolution
### Testing done 

### Reviewer checklist

### DIFF
--- a/ksqldb-execution/src/main/java/io/confluent/ksql/execution/codegen/TypeContext.java
+++ b/ksqldb-execution/src/main/java/io/confluent/ksql/execution/codegen/TypeContext.java
@@ -1,0 +1,58 @@
+package io.confluent.ksql.execution.codegen;
+
+import io.confluent.ksql.schema.ksql.types.SqlType;
+
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+
+public class TypeContext {
+  private SqlType sqlType;
+  private final List<SqlType> inputTypes = new ArrayList<>();
+  private final Map<String, SqlType> lambdaTypeMapping = new HashMap<>();
+
+  public SqlType getSqlType() {
+    return sqlType;
+  }
+
+  public void setSqlType(final SqlType sqlType) {
+    this.sqlType = sqlType;
+  }
+
+  public List<SqlType> getInputTypes() {
+    if (inputTypes.size() == 0) {
+      return null;
+    }
+    return inputTypes;
+  }
+
+  public void addInputType(final SqlType inputType) {
+    this.inputTypes.add(inputType);
+  }
+
+  public void mapInputTypes(final List<String> argumentList) {
+    if (inputTypes.size() != argumentList.size()) {
+      throw new IllegalArgumentException("Was expecting " +
+          inputTypes.size() +
+          " arguments but found " +
+          argumentList.size() + "," + argumentList +
+          ". Check your lambda statement.");
+    }
+    for (int i = 0; i < argumentList.size(); i++) {
+      this.lambdaTypeMapping.putIfAbsent(argumentList.get(i), inputTypes.get(i));
+    }
+  }
+
+  public SqlType getLambdaType(final String name) {
+    return lambdaTypeMapping.get(name);
+  }
+
+  Map<String, SqlType> getLambdaTypeMapping() {
+    return this.lambdaTypeMapping;
+  }
+
+  public Boolean notAllInputsSeen() {
+    return lambdaTypeMapping.size() != inputTypes.size() || inputTypes.size() == 0;
+  }
+}
--- a/ksqldb-execution/src/main/java/io/confluent/ksql/execution/codegen/TypeContext.java
+++ b/ksqldb-execution/src/main/java/io/confluent/ksql/execution/codegen/TypeContext.java
@@ -1,3 +1,18 @@
+/*
+ * Copyright 2021 Confluent Inc.
+ *
+ * Licensed under the Confluent Community License (the "License"); you may not use
+ * this file except in compliance with the License.  You may obtain a copy of the
+ * License at
+ *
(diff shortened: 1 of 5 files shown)


stevenpyzhang (approved):
lgtm once build passes and a unit test

ksqldb-execution/src/main/java/io/confluent/ksql/execution/codegen/TypeContext.java
```diff
@@ -0,0 +1,58 @@
+package io.confluent.ksql.execution.codegen;
+
+import io.confluent.ksql.schema.ksql.types.SqlType;
+
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+
+public class TypeContext {
```
stevenpyzhang, line 10: Can we add unit tests for this class? (exception throwing)

ksqldb-execution/src/main/java/io/confluent/ksql/execution/codegen/TypeContext.java
```diff
@@ -0,0 +1,58 @@
...
+    return sqlType;
+  }
+
+  public void setSqlType(final SqlType sqlType) {
+    this.sqlType = sqlType;
+  }
+
+  public List<SqlType> getInputTypes() {
+    if (inputTypes.size() == 0) {
+      return null;
```
agavra, line 25: nit: when public APIs can return `null`, it's often preferable to wrap it in an `Optional` so that any method callee doesn't forget to handle it

alternatively, is there any reason not to just return the empty list?

ksqldb-execution/src/main/java/io/confluent/ksql/execution/codegen/TypeContext.java
```diff
@@ -0,0 +1,58 @@
...
+    }
+    return inputTypes;
+  }
+
+  public void addInputType(final SqlType inputType) {
+    this.inputTypes.add(inputType);
+  }
+
+  public void mapInputTypes(final List<String> argumentList) {
```
agavra, line 34: how will this be used? it's generally better to avoid side-effects (changing state) if we can just return the type mapping at this point. if that's unavoidable, then that's OK

ksqldb-execution/src/main/java/io/confluent/ksql/execution/codegen/TypeContext.java
```diff
@@ -0,0 +1,58 @@
...
+
+  public SqlType getLambdaType(final String name) {
+    return lambdaTypeMapping.get(name);
+  }
+
+  Map<String, SqlType> getLambdaTypeMapping() {
+    return this.lambdaTypeMapping;
+  }
+
+  public Boolean notAllInputsSeen() {
```
agavra, line 55:
```suggestion
  public boolean notAllInputsSeen() {
```

agavra (approved):
lgtm pending steven's comments!

ksqldb-execution/src/main/java/io/confluent/ksql/execution/codegen/TypeContext.java
lct45: There used to be a reason for the null, but there isn't anymore. Switched it to just return the list

ksqldb-execution/src/main/java/io/confluent/ksql/execution/codegen/TypeContext.java
lct45: It's used inside of the `VisitLambdaExpression` in `ExpressionTypeManager`, `CodeGenRunner` and `SqlToJavaVisitor` to associate the given lambda input variables (x, y, k, v, etc) with their SqlType. The mapping is referenced in `VisitLambdaLiteral`, so I think we do want it stored in the context for future use, rather than just returning the input type at this point

guozhangwang: On a side note: I gave some thoughts on that while reading the other PR, trying to see if we can just maintain one collection instead two (a list, and a map). But cannot come up with a solution that is as efficient as now. The main "nit" concern is that if should try to enforce "when the list of input types should not be appended any more" and "the mapping (this function call) should only be triggered once the list becomes immutable" -- today they are not enforced and hence our code may be vulnerable to bugs. If you have a better idea to keep it more restricted, that'd be appreciated :)

lct45: Hmm... Right now the `notallInputsSeen` is the gatekeeper for adding inputs. The issue is that we only know that we're done seeing inputs once we get `false` from that function, which would happen after we hit `visitLambdaExpression` and do the mapping... If all lambda functions accepted the same number of inputs it would be fairly straightforward to make the list immutable, but without knowing how many inputs to expect I can't think of a good way to make it more restricted. We can make the map immutable after we add everything, but with multiple lambdas of different names but the same input type (like we talked about on the other pr), the map may need to be wiped and remapped when we see a new lambda. I don't have any other ideas right now but I can keep thinking

ksqldb-execution/src/test/java/io/confluent/ksql/execution/codegen/TypeContextTest.java
```diff
@@ -0,0 +1,44 @@
+package io.confluent.ksql.execution.codegen;
```
stevenpyzhang, line 1: nit: need the copyright at top

ksqldb-execution/src/main/java/io/confluent/ksql/execution/codegen/TypeContext.java
guozhangwang: Sounds good; fair enough :)

PR #6992, opened 2021-02-10
chore: fix CliTest.testDisableVariableSubstitution()

spena:
### Description 

### Testing done 

### Reviewer checklist

### DIFF
--- a/ksqldb-cli/src/test/java/io/confluent/ksql/cli/CliTest.java
+++ b/ksqldb-cli/src/test/java/io/confluent/ksql/cli/CliTest.java
@@ -448,7 +448,7 @@ public class CliTest {
     assertThatEventually(() -> terminal.getOutputString(),
         containsString("Failed to Describe Kafka Topic(s): [${topicName}]"));
     assertThatEventually(() -> terminal.getOutputString(),
-        containsString("Caused by: This server does not host this topic-partition."));
+        containsString("Caused by: The request attempted to perform an operation on an invalid topic."));
   }
 
   @Test


agavra (commented):
huh, was this a change in AK that we recently picked up? I'm wondering when it happened

agavra (approved):

agavra:
merging to unblock other PRs - but still want the question answered! 😉

spena:
It changed, on both AK and CCS.

In CP 6.1, the error returned was an unknown partition:
https://github.com/confluentinc/kafka/blob/6.1.0-post/core/src/main/scala/kafka/server/KafkaApis.scala#L1182
```
if (isInternal(topic)) {
          val topicMetadata = createInternalTopic(topic)
          if (topicMetadata.errorCode == Errors.COORDINATOR_NOT_AVAILABLE.code)
            metadataResponseTopic(Errors.INVALID_REPLICATION_FACTOR, topic, true, util.Collections.emptyList())
          else
            topicMetadata
        } else if (allowAutoTopicCreation && config.autoCreateTopicsEnable) {
          createTopic(topic, config.numPartitions, config.defaultReplicationFactor)
        } else {
          metadataResponseTopic(Errors.UNKNOWN_TOPIC_OR_PARTITION, topic, false, util.Collections.emptyList())
        }
```

In CP master, the error returned depends on the topic name. In the case of CLITest, the topic name is invalid, so the `validate` throws an exception and returns a different error:
https://github.com/confluentinc/kafka/blob/master/core/src/main/scala/kafka/server/KafkaApis.scala#L1100
```
nonExistingTopics.map { topic =>
          val error = try {
            Topic.validate(topic)
            Errors.UNKNOWN_TOPIC_OR_PARTITION
          } catch {
            case _: InvalidTopicException =>
              Errors.INVALID_TOPIC_EXCEPTION
          }
```

```
INVALID_TOPIC_EXCEPTION(17, "The request attempted to perform an operation on an invalid topic.",
```

agavra:
Thanks @spena for digging into that!

PR #6994, opened 2021-02-11
feat: introduce transform and reduce invocation function for lambdas

stevenpyzhang:
### Description 
Instead of needing to specify the collection type in the function name, we should just have a `transform` and a `reduce` function.

We're changing reduce so that the initial state is first and then the collection item.

There's some weirdish test results due to https://github.com/confluentinc/ksql/issues/7065

### Testing done 
Unit test and QTT's

### Reviewer checklist


guozhangwang (approved):
Made a quick pass over the non-testing part. The refactoring makes sense to me and it's good to have reduce-array too :)

PR #7003, opened 2021-02-11
fix: Append state.dir directive to ksql-server.properties

andrewegel:
- Moved RPM scriptlet that appends state.dir to %post
- Appended newline to config/ksql-server.properties so the append above
doesn't wind up on the same line as another directive

### Description 
The debian packages' `/etc/ksqldb/ksql-server.properties` had this line incorrectly appended

```
# ksql.schema.registry.url=http://localhost:8081state.dir=/var/lib/kafka-streams
```

TO remedy that, I added an extra newline to `config/ksql-server.properties` where I believe this file is sourced from.

And the RPMs `/etc/ksqldb/ksql-server.properties` didn't have this at all. RPM didn't because the script (that I moved) Was running before the package was even installed, so the file `-f` check failed. I moved this check to %post so that the check/append occurs when the package & file are actually installed.

### Testing done 
Non-code change. Nightly builds will pick this change up where we can verify the change is present in the OS packages.

### Reviewer checklist

### DIFF
--- a/config/ksql-server.properties
+++ b/config/ksql-server.properties
@@ -79,3 +79,4 @@ compression.type=snappy
 
 # Uncomment and complete the following to enable KSQL's integration to the Confluent Schema Registry:
 # ksql.schema.registry.url=http://localhost:8081
+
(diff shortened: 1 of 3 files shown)


andrewegel:
@confluentinc/ksql this is still pending review from one of you - minor change.

spena (approved):
lgtm

PR #7010, opened 2021-02-16
chore: Use the new Streams Uncaught Exception Handler

wcarlson5:
### Description 
Move the restart logic to be on the thread level

### Testing done 

### Reviewer checklist


ksqldb-engine/src/main/java/io/confluent/ksql/engine/QueryMonitor.java
```diff
@@ -139,9 +139,7 @@ private void maybeRestartQueries() {
         final PersistentQueryMetadata queryMetadata = query.get();
         switch (queryMetadata.getState()) {
           case ERROR:
-            if (now >= retryEvent.nextRestartTimeMs()) {
```
wcarlson5, line 142: We can probably just remove this whole class, it does mark queries as healthy but I think that is just for use in this class

ksqldb-engine/src/main/java/io/confluent/ksql/util/QueryMetadata.java
```diff
@@ -66,18 +66,24 @@
...
 
+  private static final Ticker CURRENT_TIME_MILLIS_TICKER = new Ticker() {
+    @Override
+    public long read() {
+      return System.currentTimeMillis();
+    }
+  };
+
   // CHECKSTYLE_RULES.OFF: ParameterNumberCheck
-  @VisibleForTesting
```
wcarlson5, line 79: Is not just used in testing. Maybe vestigial or I am misunderstanding the tag?

ksqldb-engine/src/main/java/io/confluent/ksql/util/QueryMetadata.java
```diff
@@ -160,13 +170,12 @@ public void onStop(final Consumer<Boolean> onStop) {
...
   }
 
-  protected void uncaughtHandler(final Thread t, final Throwable e) {
+  protected StreamsUncaughtExceptionHandler.StreamThreadExceptionResponse uncaughtHandler(final Throwable e) {
     QueryError.Type errorType = Type.UNKNOWN;
     try {
       errorType = errorClassifier.classify(e);
     } catch (final Exception classificationException) {
       LOG.error("Error classifying unhandled exception", classificationException);
-      throw classificationException;
```
wcarlson5, line 169: We don't want to throw errors from the handler this will cause streams the silently kill threads. Either we restart like normal or we kill the client/application. This is set to retry here.

ksqldb-engine/src/main/java/io/confluent/ksql/util/QueryMetadata.java
```diff
@@ -181,8 +190,10 @@ protected void uncaughtHandler(final Thread t, final Throwable e) {
           );
       queryStateListener.ifPresent(lis -> lis.onError(queryError));
       queryErrors.add(queryError);
-      LOG.error("Unhandled exception caught in streams thread {}. ({})", t.getName(), errorType, e);
+      LOG.error("Unhandled exception caught in streams thread {}. ({})", Thread.currentThread().getName(), errorType, e);
     }
+    retryEvent.backOff();
+    return StreamsUncaughtExceptionHandler.StreamThreadExceptionResponse.REPLACE_THREAD;
```
wcarlson5, line 196: for now we will always retry after a back off period

ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/resources/streaming/PushQueryPublisher.java
```diff
@@ -93,7 +95,10 @@ public synchronized void subscribe(final Flow.Subscriber<Collection<StreamedRow>
 
       queryMetadata.setLimitHandler(this::setDone);
       queryMetadata.setUncaughtExceptionHandler(
-          (thread, e) -> setError(e)
+          e -> {
+            setError(e);
+            return StreamsUncaughtExceptionHandler.StreamThreadExceptionResponse.SHUTDOWN_CLIENT;
```
wcarlson5, line 100: For Push queries should we retry or just shutdown?

ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/resources/streaming/QueryStreamWriter.java
```diff
@@ -183,12 +185,13 @@ private void drain(final OutputStream out) throws IOException {
...
     @Override
-    public void uncaughtException(final Thread thread, final Throwable exception) {
-      streamsException = exception instanceof Exception
-          ? (Exception) exception
-          : new RuntimeException(exception);
+    public StreamThreadExceptionResponse handle(final Throwable throwable) {
+      streamsException = throwable instanceof Exception
+              ? (Exception) throwable
+              : new RuntimeException(throwable);
+      return StreamThreadExceptionResponse.SHUTDOWN_CLIENT;
```
wcarlson5, line 194: I couldn't find where this was retried so maybe just shutdown?

wcarlson5 (commented):
This should swap out the uncaughtExceptionHanlder with minimal changes. There are several other things that we can do with it in the future but this should remove the deprecation labels and not change the behavior much

ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/resources/streaming/QueryStreamWriter.java
guozhangwang: Just for my own education: we have two classes `QueryStreamWriter` and `PushQueryPublisher` for push queries, one called by the old API `StreamedQueryResource` and the other called by the new `WSQueryEndpoint`. Why we need to maintain two callees?

ksqldb-engine/src/main/java/io/confluent/ksql/util/QueryMetadata.java
guozhangwang: Back off within the handler means we would tie-up that calling thread. Would this be a problem?

wcarlson5: No, because once the handler returns the thread will spawn a new one and then die. We can do this because the calling thread will be the one with the error

ksqldb-engine/src/main/java/io/confluent/ksql/engine/QueryMonitor.java
```diff
@@ -232,7 +230,7 @@ public void restart(final PersistentQueryMetadata query) {
 
       LOG.info("Restarting query {} (attempt #{})", queryId, numRetries);
       try {
-        query.restart();
+//        query.restart();
```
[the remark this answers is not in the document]
wcarlson5: I believe we will remove this whole file but I wanted to get some more input

ksqldb-engine/src/main/java/io/confluent/ksql/util/QueryMetadata.java
guozhangwang: Yes, I know this function is triggered as part of `StreamThread.runLoop` where we call `this.streamsUncaughtExceptionHandler.accept(e);`, and in which the caller thread is the dying stream thread. I'm just wondering if we should let the dying thread fail fast and use a scheduler with backoff to create a new thread instead of letting it to wait for a backoff time and then die --- of course, this is more complex than the current simple design.

If the motivation is to not triggering consecutive rebalances moving tasks around assuming the decision is to REPLACE_THREAD, I think that makes sense. But then it should not backoff for longer than poll interval since otherwise the thread would be kicked out of the group and rebalance triggered anyways.

wcarlson5: Ah. We can't let the thread die individually with the current streams functionality. I believe the max back off we set is 1 second by default so that shouldn't be a problem. Although it would bet best to have the max backoff time <  poll interval.

ksqldb-engine/src/main/java/io/confluent/ksql/engine/QueryMonitor.java
wcarlson5: removed

ksqldb-engine/src/main/java/io/confluent/ksql/util/QueryMetadata.java
guozhangwang: Sounds good.

ksqldb-engine/src/main/java/io/confluent/ksql/engine/QueryMonitor.java
guozhangwang: With this class removed, it seems QueryMetadata.clearErrors would never be called (since we do not check if it has been resumed for long enough time anyways). Is the `queryErrors` queue inside QueryMetadata still needed?

ksqldb-engine/src/main/java/io/confluent/ksql/util/QueryMetadata.java
```diff
@@ -181,8 +197,15 @@ protected void uncaughtHandler(final Thread t, final Throwable e) {
           );
       queryStateListener.ifPresent(lis -> lis.onError(queryError));
       queryErrors.add(queryError);
-      LOG.error("Unhandled exception caught in streams thread {}. ({})", t.getName(), errorType, e);
+      LOG.error(
```
guozhangwang, line 200: Please see the other comment above: now we are only adding queryError to the queue of QueryMetadata, but does not clean it once it resumes back to normal (since it does not go through the fail and resume pattern any more). The errors would just be an LRU cache. Is that by design?

ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/resources/streaming/PushQueryPublisher.java
guozhangwang: Do you have any motivations to let push queries just shutdown upon errors here?

guozhangwang (commented):
Just a meta question, which is probably out of the scope of this PR: for persistent queries we would like to retry forever upon errors, while for push queries we shutdown the client immediately. Is this by design, and if yes what's the rationale behind it?

Also by the end of the day, if the classified error type is user or unknown, should we also keep retrying forever?

ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/resources/streaming/QueryStreamWriter.java
wcarlson5: This is a good question, I would also like to know the answer

ksqldb-engine/src/main/java/io/confluent/ksql/engine/QueryMonitor.java
wcarlson5: I plan on using that in the new uptime metric so even though it is not used right now it should be useful later

ksqldb-engine/src/main/java/io/confluent/ksql/util/QueryMetadata.java
ableegoldman: I think the annotation generally means that something is public/package-private instead of private so we can use it outside the class. So if it's used in testing at all, this still applies

ksqldb-engine/src/main/java/io/confluent/ksql/util/QueryMetadata.java
rodesai: The backoff starts at 15 seconds and goes up to 15 minutes. 

> and in which the caller thread is the dying stream thread. I'm just wondering if we should let the dying thread fail fast and use a scheduler with backoff to create a new thread instead of letting it to wait for a backoff time and then die --- of course, this is more complex than the current simple design.

I don't think we need another scheduler. Why not just have this stream thread clean itself up (close all clients), _then_ call the handler?

ksqldb-engine/src/main/java/io/confluent/ksql/util/QueryMetadata.java
rodesai: +1 I think this is an issue. The way I was thinking we could solve this would be to make the queryErrors queue a time-bounded queue. So we can have the entries automatically expired once they've been sitting in the queue longer than some time threshold (lets say 1 hour to start)

ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/resources/streaming/PushQueryPublisher.java
rodesai: Today for transient queries we don't retry, and instead just return the error as the final message in the response. I think we should keep the status quo for this patch (which it looks like you're doing)

ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/resources/streaming/QueryStreamWriter.java
rodesai: We originally had an http1 chunked-encoding implementation for transient queries that used `QueryStreamWriter`. Then, to support C3 we added the websockets stuff, which used `PushQueryPublisher`. The new http2 stuff seems to have it's own way of doing this that isn't even handling errors (see `QueryEndpoint` and `BlockingQueryPublisher`). So tl;dr: tech-debt.

I think we should consider 2 follow-ups:
1. clean up the legacy `/query` endpoint - I don't think anything is even using this anymore
2. add error handling to the new http2 endpoints

rodesai (commented):
Thanks, @wcarlson5 this mostly looks good to me. I think the only outstanding item that needs addressing is expiring the error queue.

ksqldb-engine/src/main/java/io/confluent/ksql/util/QueryMetadata.java
guozhangwang: If the decision was to shutdown the app, then sure we should just do that asap; but here the decision is to replace thread after backoff, then I think we do not need to set the max backoff to 15min since it is too long and would trigger a rebalance.

ksqldb-engine/src/main/java/io/confluent/ksql/util/QueryMetadata.java
wcarlson5: alright

ksqldb-engine/src/main/java/io/confluent/ksql/util/QueryMetadata.java
```diff
@@ -66,18 +66,24 @@
...
   private Optional<QueryStateListener> queryStateListener = Optional.empty();
   private boolean everStarted = false;
   protected boolean closed = false;
-  private UncaughtExceptionHandler uncaughtExceptionHandler = this::uncaughtHandler;
+  private StreamsUncaughtExceptionHandler uncaughtExceptionHandler = this::uncaughtHandler;
   private KafkaStreams kafkaStreams;
   private Consumer<Boolean> onStop = (ignored) -> { };
   private boolean initialized = false;
 
+  private static final Ticker CURRENT_TIME_MILLIS_TICKER = new Ticker() {
```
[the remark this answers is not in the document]
wcarlson5: The ticker lets the time change throughout test as multiple calls are made. We don't use the default ticker because it is in nano seconds

ksqldb-engine/src/main/java/io/confluent/ksql/util/QueryMetadata.java
wcarlson5: The max back off would only happen after there has been several failures. in the first five minutes there will be four tires at 15s, 30s, 1m, 2m. If these all fail I think there should probably be a rebalance and while this thread gets up an runnning

ksqldb-engine/src/main/java/io/confluent/ksql/util/QueryMetadata.java
wcarlson5: I made a time based queue wrapping around the eviction queue as I could not find a time based eviction queue. if you had some implementation in mind let me know and I can replace it.

wcarlson5:
@rodesai @guozhangwang I bounded the error queue by time as well as capacity, I think I got everything else as well.

ksqldb-engine/src/main/java/io/confluent/ksql/util/QueryMetadata.java
```diff
@@ -339,6 +362,38 @@ private void doClose(final boolean cleanUp) {
...
+  public static class TimeBoundedQueue {
+    private final Duration duration;
+    private final Queue<QueryError> queue;
+
+    TimeBoundedQueue(final Duration duration, final int capacity) {
+      queue = EvictingQueue.create(capacity);
+      this.duration = duration;
+    }
+
+    public boolean add(final QueryError e) {
```
rodesai, line 374: why does this return anything?

ksqldb-engine/src/main/java/io/confluent/ksql/util/QueryMetadata.java
```diff
@@ -339,6 +362,38 @@ private void doClose(final boolean cleanUp) {
...
+      final boolean result = queue.add(e);
+      while (queue.peek() != null) {
```
rodesai, line 376: nit: move this into an expire method

rodesai (approved):
LGTM - couple nits inline

ksqldb-engine/src/main/java/io/confluent/ksql/util/QueryMetadata.java
wcarlson5: it doesn't need to, just because the queue interface does. I'll make void

PR #7020, opened 2021-02-17
chore: refactor schema migrations tool

jzaralim:
### Description 
Refactors some of the code commited here: #6988 

Specifically,
* move java client creation to a new utils file
* handle timing in BaseCommand class
* add new MigrationException class

### Reviewer checklist

### DIFF
--- a/ksqldb-tools/src/main/java/io/confluent/ksql/tools/migrations/commands/BaseCommand.java
+++ b/ksqldb-tools/src/main/java/io/confluent/ksql/tools/migrations/commands/BaseCommand.java
@@ -18,9 +18,7 @@ package io.confluent.ksql.tools.migrations.commands;
 import com.github.rvesse.airline.annotations.Option;
 import com.github.rvesse.airline.annotations.OptionType;
 import edu.umd.cs.findbugs.annotations.SuppressFBWarnings;
-import io.confluent.ksql.tools.migrations.Migrations;
 import org.slf4j.Logger;
-import org.slf4j.LoggerFactory;
 
 /**
  * Defines common options across all of the migration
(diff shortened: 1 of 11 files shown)


ksqldb-tools/src/main/java/io/confluent/ksql/tools/migrations/commands/BaseCommand.java
```diff
@@ -45,4 +48,14 @@
   )
   protected boolean dryRun = false;
 
+  protected static final Logger LOGGER = LoggerFactory.getLogger(Migrations.class);
+
+  @Override
+  public void run() {
+    final long startTime = System.nanoTime();
+    command();
+    LOGGER.info("Execution time: " + (System.nanoTime() - startTime) / 1000000000);
```
vcrfxia, line 57: I think it's better to have the different classes each have their own loggers, rather than having them all share the `Migrations.class` logger. By having different loggers, users can more easily filter logs and can also configure different log levels for the different loggers.

I assume your motivation for switching to a shared logger was so that the execution time would be logged by the same logger as the other messages from the command execution itself? I don't have a great solution but one option could be to have each command implement a simple `getLogger()` method, which we can then call here.

vcrfxia (approved):
Thanks @jzaralim! One suggestion inline, LGTM otherwise.

ksqldb-tools/src/main/java/io/confluent/ksql/tools/migrations/commands/BaseCommand.java
```diff
@@ -45,4 +48,14 @@
...
+  }
+
+  protected abstract void command();
```
vcrfxia, line 60: Good call making this protected. Let's update the implementations in the subclasses to not be public?

PR #7030, opened 2021-02-18
feat(client): add getInfo method to java client

jzaralim:
### Description 
Add getInfo method to java client, which queries the `/info` endpoint

### Testing done 
unit test

### Reviewer checklist

### DIFF
--- a/ksqldb-api-client/src/main/java/io/confluent/ksql/api/client/ServerInfo.java
+++ b/ksqldb-api-client/src/main/java/io/confluent/ksql/api/client/ServerInfo.java
@@ -15,13 +15,23 @@
 
 package io.confluent.ksql.api.client;
 
+/**
+ * Metadata for a ksqlDB server.
+ */
 public interface ServerInfo {
 
-  String getVersion();
+  /**
+   * @return the ksqlDB version the server is running
+   */
+  String getServerVersion();
 
+  /**
+   * @return the Kafka cluster id
+   */
   String getKafkaClusterId();
 
+  /**
+   * @return the ksqlDB service id
+   */
   String getKsqlServiceId();
-
-  String getServerStatus();
 }
--- a/ksqldb-api-client/src/main/java/io/confluent/ksql/api/client/ServerInfo.java
+++ b/ksqldb-api-client/src/main/java/io/confluent/ksql/api/client/ServerInfo.java
@@ -1,0 +1,27 @@
+/*
+ * Copyright 2021 Confluent Inc.
+ *
+ * Licensed under the Confluent Community License (the "License"); you may not use
+ * this file except in compliance with the License.  You may obtain a copy of the
+ * License at
+ *
+ * http://www.confluent.io/confluent-community-license
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OF ANY KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations under the License.
+ */
+
+package io.confluent.ksql.api.client;
+
+public interface ServerInfo {
+
+  String getVersion();
+
+  String getKafkaClusterId();
+
+  String getKsqlServiceId();
+
+  String getServerStatus();
+}
--- a/ksqldb-api-client/src/main/java/io/confluent/ksql/api/client/impl/AdminResponseHandlers.java
+++ b/ksqldb-api-client/src/main/java/io/confluent/ksql/api/client/impl/AdminResponseHandlers.java
@@ -106,11 +106,18 @@ final class AdminResponseHandlers {
       final CompletableFuture<ServerInfo> cf
   ) {
     final JsonObject source = serverInfoEntity.getJsonObject("KsqlServerInfo");
-    cf.complete(new ServerInfoImpl(
-        source.getString("version"),
-        source.getString("kafkaClusterId"),
-        source.getString("ksqlServiceId")
-    ));
+
+    try {
+      final ServerInfoImpl serverInfo = new ServerInfoImpl(
+          source.getString("version"),
+          source.getString("kafkaClusterId"),
+          source.getString("ksqlServiceId")
+      );
+      cf.complete(serverInfo);
+    } catch (Exception e) {
+      cf.completeExceptionally(new IllegalStateException(
+          "Unexpected server response format. Response: " + serverInfoEntity));
+    }
   }
 
   static boolean isListStreamsResponse(final JsonObject ksqlEntity) {
--- a/ksqldb-api-client/src/main/java/io/confluent/ksql/api/client/impl/AdminResponseHandlers.java
+++ b/ksqldb-api-client/src/main/java/io/confluent/ksql/api/client/impl/AdminResponseHandlers.java
@@ -17,6 +17,7 @@ package io.confluent.ksql.api.client.impl;
 
 import io.confluent.ksql.api.client.QueryInfo;
 import io.confluent.ksql.api.client.QueryInfo.QueryType;
+import io.confluent.ksql.api.client.ServerInfo;
 import io.confluent.ksql.api.client.SourceDescription;
 import io.confluent.ksql.api.client.StreamInfo;
 import io.confluent.ksql.api.client.TableInfo;
@@ -28,7 +29,9 @@ import java.util.Optional;
 import java.util.concurrent.CompletableFuture;
 import java.util.stream.Collectors;
 
+// CHECKSTYLE_RULES.OFF: ClassDataAbstractionCoupling
 final class AdminResponseHandlers {
+  // CHECKSTYLE_RULES.ON: ClassDataAbstractionCoupling
 
   private AdminResponseHandlers() {
   }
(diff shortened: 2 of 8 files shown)


ksqldb-api-client/src/main/java/io/confluent/ksql/api/client/ServerInfo.java
```diff
@@ -0,0 +1,27 @@
...
+
+public interface ServerInfo {
+
+  String getVersion();
+
+  String getKafkaClusterId();
+
+  String getKsqlServiceId();
+
+  String getServerStatus();
```
vcrfxia, line 26: This is just the command runner status, right? I don't think it's useful to Java client users, as in, I can't think of what someone would do in response to getting this information. I say let's leave it out for now in order to keep the interface simpler. We can always add it in later if there's demand for it.

ksqldb-api-client/src/main/java/io/confluent/ksql/api/client/ServerInfo.java
```diff
@@ -0,0 +1,27 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OF ANY KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations under the License.
+ */
+
+package io.confluent.ksql.api.client;
+
+public interface ServerInfo {
```
vcrfxia, line 18: All public interfaces should have javadocs.

ksqldb-api-client/src/main/java/io/confluent/ksql/api/client/impl/AdminResponseHandlers.java
```diff
@@ -326,4 +342,26 @@ static boolean isConnectErrorResponse(final JsonObject ksqlEntity) {
...
     }
   }
+
+  /**
+   * Attempts to parse the provided response entity as a {@code ServerInfo}.
+   *
+   * @param serverInfoEntity response entity
+   * @return optional containing parsed result if successful, else empty
+   */
+  private static Optional<ServerInfo> getServerInfoResponse(
```
vcrfxia, line 352: We can inline this method. The reason the others aren't inlined is because they may be returned from the generic `/ksql` endpoint, in which case it's useful to have a separate helper method to check the type.

Or, if you think it's cleaner to have a separate method rather than inlining this, we should at least update the return type to not be optional. There is no situation in which we expect this method to be called and not result in a `ServerInfo` object.

ksqldb-api-client/src/main/java/io/confluent/ksql/api/client/impl/ClientImpl.java
```diff
@@ -355,7 +382,17 @@ public void close() {
       final T cf,
```
vcrfxia, line 382: Let's rename these existing `makeRequest()` methods `makePostRequest()` to clarify. Up to you whether you want the new one you added to be `makeRequest()` or `makeGetRequest()` (and hard-code `HttpMethod.GET`).

ksqldb-api-client/src/main/java/io/confluent/ksql/api/client/impl/ServerInfoImpl.java
```diff
@@ -0,0 +1,88 @@
...
+  private final String ksqlServiceId;
+  private final String serverStatus;
+
+  public ServerInfoImpl(
+      final String version,
+      final String kafkaClusterId,
+      final String ksqlServiceId,
+      final String serverStatus
+  ) {
+    this.version = version;
```
vcrfxia, line 33: nit: validate that these fields are non-null.

ksqldb-api-client/src/main/java/io/confluent/ksql/api/client/impl/ServerInfoImpl.java
```diff
@@ -0,0 +1,88 @@
...
+  @Override
+  public boolean equals(final Object o) {
+    if (this == o) {
+      return true;
+    }
+    if (o == null || getClass() != o.getClass()) {
+      return false;
+    }
+    final ServerInfoImpl that = (ServerInfoImpl) o;
+    return version == that.version
```
vcrfxia, line 68: Why `==` for string comparison here?

ksqldb-api-client/src/main/java/io/confluent/ksql/api/client/ServerInfo.java
```diff
@@ -0,0 +1,27 @@
...
+
+  String getVersion();
```
vcrfxia, line 20: nit: rename to either `getKsqlVersion` or `getServerVersion` to disambiguate. As written, people may accidentally think we're versioning the `ServerInfo`, rather than the server itself.

ksqldb-api-client/src/main/java/io/confluent/ksql/api/client/impl/ServerInfoImpl.java
vcrfxia: Relatedly, doesn't hurt to add a quick unit test to check that this `equals` implementation is correct. See, e.g., TableInfoImplTest.java for an example.

ksqldb-api-client/src/test/java/io/confluent/ksql/api/client/ClientTest.java
```diff
@@ -1390,6 +1391,15 @@ public void shouldDescribeSource() throws Exception {
...
 
+  @Test
+  public void shouldGetServerInfo() throws Exception {
+    final ServerInfo serverInfo = javaClient.serverInfo().get();
+    assertThat(serverInfo.getVersion(), is(AppInfo.getVersion()));
+    assertThat(serverInfo.getKsqlServiceId(), is("ksql-service-id"));
+    assertThat(serverInfo.getKafkaClusterId(), is("kafka-cluster-id"));
+    assertThat(serverInfo.getServerStatus(), is("server-status"));
+  }
+
```
vcrfxia, line 1402: Let's also add a quick integration test to `ClientIntegrationTest.java`, to protect against someone inadvertently updating the API format without updating the client.

vcrfxia (commented):
Thanks @jzaralim -- looking great! Some minor comments inline.

ksqldb-api-client/src/main/java/io/confluent/ksql/api/client/impl/AdminResponseHandlers.java
```diff
@@ -98,6 +101,18 @@ static void handleDescribeSourceResponse(
     }
   }
 
+  static void handleServerInfoResponse(
+      final JsonObject serverInfoEntity,
+      final CompletableFuture<ServerInfo> cf
+  ) {
+    final JsonObject source = serverInfoEntity.getJsonObject("KsqlServerInfo");
+    cf.complete(new ServerInfoImpl(
```
vcrfxia, line 109: Let's wrap the creation of the `ServerInfoImpl` in a try-catch block, so we can fail the future with a more informative error message if something goes wrong (nothing should go wrong today, but it'd be good to future-proof in the event that the response format is inadvertently changed in an incompatible way).

ksqldb-api-client/src/main/java/io/confluent/ksql/api/client/impl/ServerInfoImpl.java
```diff
@@ -0,0 +1,78 @@
...
+  private final String serverVersion;
+  private final String kafkaClusterId;
+  private final String ksqlServiceId;
+
+  public ServerInfoImpl(
+      final String serverVersion,
+      final String kafkaClusterId,
+      final String ksqlServiceId
+  ) {
+    this.serverVersion = Objects.requireNonNull(serverVersion);
```
vcrfxia, line 31: nit: use second argument in `Objects.requireNonNull()` to produce a more informative error message if something goes wrong here.
```suggestion
    this.serverVersion = Objects.requireNonNull(serverVersion. "serverVersion");
```
and similar for the others.

ksqldb-api-client/src/main/java/io/confluent/ksql/api/client/impl/ServerInfoImpl.java
```diff
@@ -0,0 +1,78 @@
...
+
+  @Override
+  public int hashCode() {
+    return Objects.hash(serverVersion, kafkaClusterId, ksqlServiceId);
+  }
+
+  @Override
+  public String toString() {
+    return "ServerInfo{"
+        + "version='" + serverVersion + '\''
```
vcrfxia, line 73: nit: update name
```suggestion
        + "serverVersion='" + serverVersion + '\''
```

ksqldb-api-client/src/test/java/io/confluent/ksql/api/client/integration/ClientIntegrationTest.java
```diff
@@ -956,6 +958,21 @@ public void shouldHandleErrorResponseFromDescribeSource() {
...
   }
 
+  @Test
+  public void shouldGetServerInfo() throws Exception {
+    // Given:
+    final String expectedClusterId = REST_APP.getServiceContext().getAdminClient().describeCluster().clusterId().get();
+
+    // When:
+    final ServerInfo serverInfo = client.serverInfo().get();
+    Object o = REST_APP.getServiceContext().getAdminClient().describeCluster().clusterId().get();
```
vcrfxia, line 968: Is this used?

vcrfxia (approved):
Thanks @jzaralim ! This is awesome :) Minor comments inline. Also I updated the PR title to reflect that is a new feature, available for general consumption 😁

PR #7046, opened 2021-02-19
chore: refactor migrations commands to not call exit from within command

vcrfxia:
### Description 

BaseCommand should handle returning the appropriate exit code after printing the execution time, rather than exiting from within the commands themselves.

### Testing done 

Minor refactor, more thorough testing (in general) coming in a later PR.

### Reviewer checklist

### DIFF
--- a/ksqldb-tools/src/main/java/io/confluent/ksql/tools/migrations/Migrations.java
+++ b/ksqldb-tools/src/main/java/io/confluent/ksql/tools/migrations/Migrations.java
@@ -18,6 +18,7 @@ package io.confluent.ksql.tools.migrations;
 import com.github.rvesse.airline.Cli;
 import com.github.rvesse.airline.help.Help;
 import io.confluent.ksql.tools.migrations.commands.ApplyMigrationCommand;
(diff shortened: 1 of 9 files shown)


agavra (approved):

ksqldb-tools/src/main/java/io/confluent/ksql/tools/migrations/Migrations.java
```diff
@@ -52,10 +53,10 @@
   private Migrations() {}
 
   public static void main(final String[] args) {
-    // all Migrations commands must implement Runnable, so we
+    // all Migrations commands must implement BaseCommand, so we
     // are safe to assume that the type returned is Cli<Runnable>
```
jzaralim, line 57: nit:
```suggestion
    // are safe to assume that the type returned is Cli<BaseCommand>
```

jzaralim (approved):

ksqldb-tools/src/main/java/io/confluent/ksql/tools/migrations/Migrations.java
vcrfxia: Nice eye!

PR #7047, opened 2021-02-19
chore: update some pull query error messages and configs

### LINKED ISSUE #4025: ksqlDB 0.6 - improve pull/push error messages
I have a stream that is CTAS plus INSERT INTO: 

```
Queries that write from this STREAM
CSAS_ALL_TWEETS_7 : CREATE STREAM ALL_TWEETS WITH (KAFKA_TOPIC='ALL_TWEETS', PARTITIONS=1, REPLICAS=1) AS SELECT *
FROM TWEETS_01 TWEETS_01
EMIT CHANGES;
InsertQuery_8 : INSERT INTO ALL_TWEETS SELECT * FROM TWEETS_02;

```

😒When I try to select from it I don't get the helpful message guiding me to `EMIT CHANGES`: 

```
ksql> select * from all_tweets;
Multiple queries currently materialize '`ALL_TWEETS`'. KSQL currently only supports pull queries when the table has only been materialized once.
ksql>
```

😄 much better is this message; we should do this for the one above too: 

```
ksql> select * from tweets_03;
Table 'TWEETS_03' is not materialized. Refer to https://cnfl.io/queries for info on query types. If you intended to issue a push query, resubmit with the EMIT CHANGES clause
 KSQL currently only supports pull queries on materialized aggregate tables. i.e. those created by a 'CREATE TABLE AS SELECT <fields>, <aggregate_functions> FROM <sources> GROUP BY <key>' style statement.
Query syntax in KSQL has changed. There are now two broad categories of queries:
- Pull queries: query the current state of the system, return a result, and terminate.
- Push queries: query the state of the system in motion and continue to output results until they meet a LIMIT condition or are terminated by the user.

'EMIT CHANGES' is used to to indicate a query is a push query. To convert a pull query into a push query, which was the default behavior in older versions of KSQL, add `EMIT CHANGES` to the end of the statement before any LIMIT clause.

For example, the following are pull queries:
        'SELECT * FROM X WHERE ROWKEY=Y;' (non-windowed table)
        'SELECT * FROM X WHERE ROWKEY=Y AND WINDOWSTART>=Z;' (windowed table)

The following is a push query:
        'SELECT * FROM X EMIT CHANGES;'

Note: Persistent queries, e.g. `CREATE TABLE AS ...`, have an implicit `EMIT CHANGES`, but we recommend adding `EMIT CHANGES` to these statements.
ksql>
```

### LINKED ISSUE #5821: Enable pull query metrics by default
In the run up to launching pull queries, we kept all aspects of it behind a feature flag. One vestigial part of that is the `ksql.query.pull.metrics.enabled` server configuration, which is false by default. Now that pull queries are available, this configuration leads to general confusion. It is unintuitive that metrics would be disabled out of the box.

Let's either ditch this flag and enable them always, or at least change the default to true.

agavra:
### Description 

fixes #4025 
fixes #5821 

### Testing done 

Added test to cover new error message

### Reviewer checklist

### DIFF
--- a/ksqldb-common/src/main/java/io/confluent/ksql/util/KsqlConfig.java
+++ b/ksqldb-common/src/main/java/io/confluent/ksql/util/KsqlConfig.java
@@ -762,7 +762,7 @@ public class KsqlConfig extends AbstractConfig {
         .define(
             KSQL_QUERY_PULL_METRICS_ENABLED,
             Type.BOOLEAN,
-            false,
+            true,
             Importance.LOW,
             KSQL_QUERY_PULL_METRICS_ENABLED_DOC
         )
--- a/ksqldb-engine/src/main/java/io/confluent/ksql/engine/PullQueryExecutionUtil.java
+++ b/ksqldb-engine/src/main/java/io/confluent/ksql/engine/PullQueryExecutionUtil.java
@@ -68,34 +68,32 @@ public final class PullQueryExecutionUtil {
   static PersistentQueryMetadata findMaterializingQuery(
       final EngineContext engineContext, final ImmutableAnalysis analysis) {
 
-    final SourceName sourceName = getSourceName(analysis);
+    final DataSource source = analysis.getFrom().getDataSource();
+    final SourceName sourceName = source.getName();
 
     final Set<QueryId> queries = engineContext.getQueriesWithSink(sourceName);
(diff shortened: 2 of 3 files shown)


guozhangwang (approved):
LGTM!

PR #7050, opened 2021-02-19
chore(migrations): update MigrationConfig to query /info with Java client

jzaralim:
### Description 
Also refactors the class to throw exceptions instead of calling `System.exit`
### Reviewer checklist

### DIFF
--- a/ksqldb-tools/src/main/java/io/confluent/ksql/tools/migrations/MigrationConfig.java
+++ b/ksqldb-tools/src/main/java/io/confluent/ksql/tools/migrations/MigrationConfig.java
@@ -103,7 +103,7 @@ public final class MigrationConfig extends AbstractConfig {
     }
 
     final Client client = MigrationsUtil.getKsqlClient(ksqlServerUrl);
-    CompletableFuture<ServerInfo> response = client.serverInfo();
+    final CompletableFuture<ServerInfo> response = client.serverInfo();
 
     try {
       final String serviceId = response.get().getKsqlServiceId();
--- a/ksqldb-tools/src/main/java/io/confluent/ksql/tools/migrations/MigrationConfig.java
+++ b/ksqldb-tools/src/main/java/io/confluent/ksql/tools/migrations/MigrationConfig.java
@@ -15,23 +15,17 @@
 
 package io.confluent.ksql.tools.migrations;
 
-import edu.umd.cs.findbugs.annotations.SuppressFBWarnings;
+import io.confluent.ksql.api.client.Client;
+import io.confluent.ksql.api.client.ServerInfo;
 import io.confluent.ksql.properties.PropertiesUtil;
-import io.confluent.ksql.rest.client.KsqlRestClient;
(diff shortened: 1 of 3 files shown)


ksqldb-tools/src/main/java/io/confluent/ksql/tools/migrations/MigrationConfig.java
```diff
@@ -111,36 +96,29 @@ private MigrationConfig(final Map<String, String> configs, final String id) {
...
-      LOGGER.error("Failed to query " + MigrationConfig.KSQL_SERVER_URL + "/info");
-      System.exit(1);
+    try {
+      final String serviceId = response.get().getKsqlServiceId();
+      return serviceId;
+    } catch (InterruptedException e) {
+      throw new MigrationException("Interrupted while attempting to connect to "
+          + ksqlServerUrl + "/info");
+    } catch (ExecutionException e) {
+      if (e.getCause() instanceof IllegalStateException) {
```
vcrfxia, line 115: What results in an IllegalStateException? The error message suggests that's what we get if the server URL is invalid? I'm surprised that'd result in an IllegalStateException.

vcrfxia (approved):
LGTM with a question inline. Thanks @jzaralim ! 

Looks like there's slight overlap between this PR and https://github.com/confluentinc/ksql/pull/7053. You should go ahead and merge first, I'll update my PR afterwards.

ksqldb-tools/src/main/java/io/confluent/ksql/tools/migrations/MigrationConfig.java
jzaralim: It occurs when the response is in an unexpected format (missing a field etc). That would happen if the url provided also happened to have an /info endpoint, but I'm also starting to realize that it could also happen if the java client and the server have incompatible versions.

Will update the error message

vcrfxia: We should check up front whether the version is compatible or not, and throw a more specific error message if so. If we do that, hopefully we won't need to check for this specific exception type here anymore (it's weird to check for IllegalStateException since those shouldn't happen under normal operation).

ksqldb-tools/src/main/java/io/confluent/ksql/tools/migrations/MigrationConfig.java
```diff
@@ -15,23 +15,17 @@
 
 package io.confluent.ksql.tools.migrations;
 
-import edu.umd.cs.findbugs.annotations.SuppressFBWarnings;
+import io.confluent.ksql.api.client.Client;
+import io.confluent.ksql.api.client.ServerInfo;
 import io.confluent.ksql.properties.PropertiesUtil;
-import io.confluent.ksql.rest.client.KsqlRestClient;
```
vcrfxia, line 20: Are we able to remove the ksqldb-rest-client dependency in the pom for this module now that this is gone?

PR #7055, opened 2021-02-20
build: fix build after AK bump (#6979)

ableegoldman:
Full cherrypick of https://github.com/confluentinc/ksql/pull/6979

### DIFF
--- a/ksqldb-engine/src/main/java/io/confluent/ksql/engine/EngineContext.java
+++ b/ksqldb-engine/src/main/java/io/confluent/ksql/engine/EngineContext.java
@@ -228,7 +228,7 @@ final class EngineContext {
     return result.getMessage();
   }
 
-  void registerQuery(final QueryMetadata query) {
+  void registerQuery(final QueryMetadata query, final boolean createAsQuery) {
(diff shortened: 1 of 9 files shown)


guozhangwang:
The fix of 6979 is primarily in `SandboxedExecutionContext.java`, but it is not included in this PR?

stevenpyzhang (approved):
lgtm if build passes

stevenpyzhang:
@guozhangwang that fix is already in the 6.1.x branch

The original fix was in master.
For older branches, the only branch we cherry-picked something to was 6.0.x and for that branch, we only need the sandbox fix for that branch. When that got merged into 6.0.x and we pint merged, 6.1.x picked up the sandbox fix.

https://github.com/confluentinc/ksql/blob/6.1.x/ksqldb-engine/src/main/java/io/confluent/ksql/engine/SandboxedExecutionContext.java#L124

So now we just need the rest of the fix to 6.1.x

guozhangwang:
> @guozhangwang that sandbox fix is already in the 6.1.x branch

Ack, thanks!

PR #7056, opened 2021-02-20
feat: implement correct logic for nested lambdas and more complex lambda expressions

stevenpyzhang:
### Description 
When processing a function call, we only want to use the context with the input type list populated for the LambdaFunctionCalls since those are where we need the input type list for mapping arguments. For other FunctionCall argument expressions, we need to pass in the original context that we arrived at the node with.

We also need to be passing copies of the contexts in order to prevent corruption when mapping the input type to the lambda arguments from other subtrees.

### Testing done 
Unit tests so far, QTT test coming soon.

### Reviewer checklist


ksqldb-execution/src/main/java/io/confluent/ksql/execution/codegen/CodeGenRunner.java
```diff
@@ -183,24 +183,31 @@ public Void visitLikePredicate(final LikePredicate node, final TypeContext conte
...
     public Void visitFunctionCall(final FunctionCall node, final TypeContext context) {
-      final List<SqlArgument> argumentTypes = new ArrayList<>();
       final FunctionName functionName = node.getName();
+
+      final TypeContext contextCopy = context.getCopy();
+      final List<SqlArgument> argumentTypes = new ArrayList<>();
       final boolean hasLambda = node.hasLambdaFunctionCallArguments();
       for (final Expression argExpr : node.getArguments()) {
-        final TypeContext childContext = context.getCopy();
+        final TypeContext childContext;
```
guozhangwang, line 192: Better add some comments for other readers why we are doing this: since the context need to be shared across function args that involves lambdas, blah blah.

ksqldb-execution/src/main/java/io/confluent/ksql/execution/codegen/CodeGenRunner.java
```diff
@@ -183,24 +183,31 @@ public Void visitLikePredicate(final LikePredicate node, final TypeContext conte
 
     @Override
     public Void visitFunctionCall(final FunctionCall node, final TypeContext context) {
-      final List<SqlArgument> argumentTypes = new ArrayList<>();
       final FunctionName functionName = node.getName();
+
+      final TypeContext contextCopy = context.getCopy();
```
guozhangwang, line 188: nit: maybe rename to `lambdaSharedContext`?

ksqldb-execution/src/main/java/io/confluent/ksql/execution/codegen/CodeGenRunner.java
```diff
@@ -183,24 +183,31 @@ public Void visitLikePredicate(final LikePredicate node, final TypeContext conte
...
+        final TypeContext childContext;
+        if (argExpr instanceof LambdaFunctionCall) {
+          childContext = contextCopy.getCopy();
+        } else {
+          childContext = context.getCopy();
+        }
         final SqlType resolvedArgType =
-            expressionTypeManager.getExpressionSqlType(argExpr, childContext);
-        process(argExpr, context.getCopy());
+            expressionTypeManager.getExpressionSqlType(argExpr, childContext.getCopy());
```
guozhangwang, line 199: We are a bit duplicate copying context here. Maybe we can wrap the logic of

```
        final TypeContext childContext;
        if (argExpr instanceof LambdaFunctionCall) {
          childContext = contextCopy.getCopy();
        } else {
          childContext = context.getCopy();
        }
```

as a `argTypeContext(argExpr, parentContext, lambdaSharedContext)`, which would be called for both `expressionTypeManager#getExpressionSqlType` and `process`? Ditto on other classes.

stevenpyzhang:
PR is blocked on https://github.com/confluentinc/ksql/pull/7093 right now

ksqldb-execution/src/main/java/io/confluent/ksql/execution/codegen/CodeGenRunner.java
agavra: took me a while as well to understand what was going on. I'd name them like this:
```
// this is the one that's passed in, it represents the context at the time the parent called this method
TypeContext parentTypeContext; 

// this is a copy of the parent type context, additionally populated with the result of visiting non-lambda
// expressions - it starts as a copy of the parent typeContext
TypeContext currentTypeContext; 

// a copy of either parent or current type context to be passed to the child - in the case of lambdas
// we pass in the parent type context without the result of resolving the current context because 
// there may be valid overlapping lambda parameter names
TypeContext childContext;
```

agavra (approved):
The approach makes sense to me (if I'm understanding the problem correctly, see comment below with suggested naming)

PR #7059, opened 2021-02-21
chore(migrations): require server url as argument to `new` command

vcrfxia:
### Description 

This PR is stacked on https://github.com/confluentinc/ksql/pull/7053. Only the commits starting from `chore(migrations): require server url as argument to command` need to be reviewed.

The `ksql-migrations new` command now requires two arguments, the migrations directory and the ksqlDB server URL, which is used to initialize the migrations config file (the server URL is a required config). This means `initialize` may now be called immediately after `new`. 

In a future PR we can consider initializing the config file with other configs (set to their default values) and helpful comments for how to use them.

### Testing done 

Updated unit tests.

### Reviewer checklist

### DIFF
--- a/ksqldb-tools/src/main/java/io/confluent/ksql/tools/migrations/commands/NewMigrationCommand.java
+++ b/ksqldb-tools/src/main/java/io/confluent/ksql/tools/migrations/commands/NewMigrationCommand.java
@@ -27,6 +27,8 @@ import java.io.File;
 import java.io.FileNotFoundException;
 import java.io.IOException;
 import java.io.PrintWriter;
+import java.io.UnsupportedEncodingException;
+import java.nio.charset.Charset;
 import java.nio.file.FileSystemException;
 import java.nio.file.Files;
 import java.nio.file.Paths;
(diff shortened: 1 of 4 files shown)


ksqldb-tools/src/main/java/io/confluent/ksql/tools/migrations/commands/NewMigrationCommand.java
```diff
@@ -92,6 +116,16 @@ private boolean tryCreatePropertiesFile(final String path) {
       LOGGER.error(String.format("Failed to create file %s: %s", path, e.getMessage()));
       return false;
     }
+
+    final String initialConfig = MigrationConfig.KSQL_SERVER_URL + "=" + ksqlServerUrl;
```
jzaralim, line 122: How will we add more default lines to the config file in the future?

jzaralim (approved):
LGTM, with a question

ksqldb-tools/src/main/java/io/confluent/ksql/tools/migrations/commands/NewMigrationCommand.java
vcrfxia: Not sure I understand your question -- any extra lines we add to the default config should also be added here (in the same place). We can pull this out into a helper method and switch to using StringBuilder when the time comes.

jzaralim: Was asking since this section could get long if we include comments and TLS configs. What you suggested works though.

PR #7061, opened 2021-02-21
feat: classify authorization exception as user error

mjsax:
### Description 
If we receive a `AuthorizationException` it indicates as configuration issue, thus, we should classify it as a user error.

### Testing done 
Added corresponding unit tests.

### Reviewer checklist


ksqldb-engine/src/main/java/io/confluent/ksql/query/TopicAuthorizationClassifier.java
```diff
@@ -0,0 +1,52 @@
...
+    this.queryId = Objects.requireNonNull(queryId, "queryId");
+  }
+
+  @Override
+  public Type classify(final Throwable e) {
+    final Type type = e instanceof TopicAuthorizationException ? Type.USER : Type.UNKNOWN;
+
+    if (type == Type.USER) {
+      LOG.info(
+          "Classified error as USER error based on missing topic. Query ID: {} Exception: {}",
```
agavra, line 44: I think you meant to update this error message

agavra (approved):
LGTM! It might make sense to test this locally to make sure that this is exactly the error type that is raised (and not wrapped in something else)

ksqldb-engine/src/test/java/io/confluent/ksql/query/TopicAuthorizationClassifierTest.java
```diff
@@ -0,0 +1,55 @@
...
+
+    // When:
+    final Type type = new TopicAuthorizationClassifier("").classify(e);
+
+    // Then:
+    assertThat(type, is(Type.USER));
+  }
+
+  @Test
+  public void shouldClassifyNoMissingTopicAsUnknownError() {
```
vcrfxia, line 44:
```suggestion
  public void shouldClassifyNoMissingAclAsUnknownError() {
```

ksqldb-engine/src/main/java/io/confluent/ksql/query/TopicAuthorizationClassifier.java
```diff
@@ -0,0 +1,52 @@
+/*
+ * Copyright 2020 Confluent Inc.
```
vcrfxia, line 2: nit: year (and also for the new test file)

vcrfxia (approved):
+1 to @agavra 's comment on checking that we're catching the right exception -- LGTM otherwise. Thanks @mjsax !

rodesai:
@mjsax can we also knock out https://confluentinc.atlassian.net/browse/KCI-237 in this patch?

guozhangwang:
Sorry to keep you adding more into this PR, but besides KCI-237, here are a few more exceptions we can knock out :)

* ClusterAuthorizationException
* AuthenticationException
* TransactionalIdAuthorizationException

mjsax:
@guozhangwang It seem an `AuthenticationException` would never be thrown, but only sub-types. Thus I think we could skip this part? Also, for KafkaStreams, `ClusterAuthorizationException` should never be thrown (only kslqDB itself needs this), but it would be a different code path from my understanding to handle this case.

ksqldb-engine/src/main/java/io/confluent/ksql/query/TransactionAuthorizationClassifier.java
```diff
@@ -0,0 +1,57 @@
...
+  private final String queryId;
+
+  public TransactionAuthorizationClassifier(final String queryId) {
+    this.queryId = Objects.requireNonNull(queryId, "queryId");
+  }
+
+  @Override
+  public Type classify(final Throwable e) {
+    final Type type =
+        e instanceof StreamsException
```
mjsax, line 42: @guozhangwang It's unclear to me, why this exception is wrapped with `StreamsException` while `TopicAuthorizationException` and `GroupAuthorizationException` are not?

For `MissingSourceTopicException` it's clear because it extends `StreamsException` and is thrown by KS directly.

ksqldb-engine/src/main/java/io/confluent/ksql/query/MissingTopicClassifier.java
```diff
@@ -22,9 +22,7 @@
 import org.slf4j.LoggerFactory;
 
 /**
- * {@code MissingTopicClassifier} classifies errors by querying the broker
- * to check that all topics that the query relies on being accessible exist
- * and are accessible
+ * {@code MissingTopicClassifier} classifies missing source topic exceptions as user error
```
vcrfxia, line 25: Unrelated to this PR but I'm curious: what happens if an internal (changelog or repartition) topic is missing, or if a sink topic is missing? Does Streams throw a different type of exception in these cases?

ksqldb-engine/src/test/java/io/confluent/ksql/query/GroupAuthorizationClassifierTest.java
```diff
@@ -0,0 +1,55 @@
...
+import org.apache.kafka.streams.errors.StreamsException;
+import org.junit.Test;
+import org.junit.runner.RunWith;
+import org.mockito.junit.MockitoJUnitRunner;
+
+@RunWith(MockitoJUnitRunner.class)
+public class GroupAuthorizationClassifierTest {
+
+  @Test
+  public void shouldClassifyTopicAuthorizationExceptionAsUserError() {
```
vcrfxia, line 32:
```suggestion
  public void shouldClassifyGroupAuthorizationExceptionAsUserError() {
```

ksqldb-engine/src/test/java/io/confluent/ksql/integration/SecureIntegrationTest.java
```diff
@@ -321,6 +527,29 @@ private void assertCanRunKsqlQuery(
...
+  private void assertQueryFailsWithUserError(
+      final String query,
+      final String errorMsg
+  ) {
+    final QueryMetadata queryMetadata = KsqlEngineTestUtil
+        .execute(serviceContext, ksqlEngine, query, ksqlConfig, Collections.emptyMap()).get(0);
+
+    queryMetadata.start();
+    assertThatEventually(
+        "",
```
vcrfxia, line 539: nit: empty error message.

ksqldb-engine/src/test/java/io/confluent/ksql/integration/SecureIntegrationTest.java
```diff
@@ -243,6 +254,196 @@ public void shouldWorkWithMinimalPrefixedAcls() {
     assertCanAccessClusterConfig(prefix);
   }
 
+  @Test
+  public void shouldClassifyMissingSourceTopicExceptionAsUserError() {
+    // Given:
+    final String serviceId = "my-service-id_";  // Defaults to "default_"
```
vcrfxia, line 260: These new tests are hard to read since it's not clear to me which parts of the "givens" are relevant for the test and which are just boilerplate. Can we refactor the common parts/boilerplate into a helper method (e.g., `givenTestSetupWithAclsForQuery()`) so only the relevant parts are left in each test?

ksqldb-engine/src/test/java/io/confluent/ksql/integration/SecureIntegrationTest.java
```diff
@@ -243,6 +254,196 @@ public void shouldWorkWithMinimalPrefixedAcls() {
...
+
+  @Test
+  public void shouldClassifyTransactionIdAuthorizationExceptionAsUserError() {
+    // Given:
+    final String serviceId = "my-service-id_";  // Defaults to "default_"
+    final String prefix = "_confluent-ksql-" + serviceId;
+
+    final Map<String, Object> ksqlConfig = getKsqlConfig(NORMAL_USER);
+    ksqlConfig.put(KSQL_SERVICE_ID_CONFIG, serviceId);
+    ksqlConfig.put(StreamsConfig.PROCESSING_GUARANTEE_CONFIG, StreamsConfig.EXACTLY_ONCE_BETA);
```
vcrfxia, line 409: Why does this cause a transaction ID authorization exception? (Can we add a quick comment so readers know that's the purpose of setting this config in this test?)

vcrfxia (approved):
Thanks @mjsax -- LGTM with minor suggestions inline. I'm not sure about the answer to your question about why some exceptions are wrapped and others aren't. ksqlDB directly passes the exception (from the call to `kafkaStreams.setUncaughtExceptionHandler(...)`) to the `QueryErrorClassifier` so there's no unwrapping done in ksqlDB itself. Would have to dive into the Streams code to understand why some are wrapped and others aren't.

ksqldb-engine/src/main/java/io/confluent/ksql/query/TransactionAuthorizationClassifier.java
guozhangwang: I think it should not, and it should be fixed via https://cwiki.apache.org/confluence/display/KAFKA/KIP-691%3A+Enhance+Transactional+Producer+Exception+Handling

cc @abbccdda

ksqldb-engine/src/main/java/io/confluent/ksql/query/MissingTopicClassifier.java
mjsax: Atm, for changelogs and repartition topic they would be created. We would fail if the expected config does not match. (this will "change" with https://cwiki.apache.org/confluence/display/KAFKA/KIP-698%3A+Add+Explicit+User+Initialization+of+Broker-side+State+to+Kafka+Streams) -- Or course, it's only verified in a rebalance, but checking the source topic is also only done during a rebalance.

For output topics, Kafka Streams won't do anything, and thus the producer would fill up its write buffer and eventually block. To eventually Streams would crash.

But as ksqlDB checks for output topic, it should be ok?

guozhangwang:
> @guozhangwang It seem an `AuthenticationException` would never be thrown, but only sub-types. Thus I think we could skip this part? Also, for KafkaStreams, `ClusterAuthorizationException` should never be thrown (only kslqDB itself needs this), but it would be a different code path from my understanding to handle this case.

Why `CLUSTER_AUTHORIZATION_FAILED` should never be thrown? From javadocs I see it is possible to return in produce response?

ksqldb-engine/src/main/java/io/confluent/ksql/query/TransactionAuthorizationClassifier.java
mjsax: But the producer should not wrap anything with `StreamsException` -- so why does KS wrap the one but not the other producer exception? And how would the KIP change it (it seem the KIP is for the producer but not for KS)?

guozhangwang: `TopicAuthorizationException`: this can be thrown from both consumer and producer, and for consumer it could throw from `partitionsFor` which is used in KS via global store, in that place we just do not wrap it and let it throw all the way up. For producer it may be thrown from the callback, in which we do wrap --- this is an inconsistency.

`GroupAuthorizationException`: only thrown from producer in latest version with KIP-447, for which we should wrap in KS; in older version it could be thrown from consumer and again we do not wrap in KS.

`TransactionalIdAuthorizationException`: thrown from producer, in the callback where we always wrap.

rodesai: In our other classifiers (e.g. the regex classifier we walk the cause stack and try and find any exception that matches). That would be an approach to make this check more robust.

ksqldb-engine/src/main/java/io/confluent/ksql/query/MissingTopicClassifier.java
rodesai: > For output topics, Kafka Streams won't do anything, and thus the producer would fill up its write buffer and eventually block. To eventually Streams would crash.

this is true for internal topics too right? It seems like right now for internal topics that are deleted we'd block for max.block.ms and then throw an error classified as SYSTEM. Then, retry and recreate. After KIP-698 we'd block and then throw an error we will classify as USER on every retry.

> But as ksqlDB checks for output topic, it should be ok?

Where do we check?

rodesai (commented):
LGTM! nice integration tests :)

ksqldb-engine/src/main/java/io/confluent/ksql/query/MissingTopicClassifier.java
mjsax: > this is true for internal topics too right?

Yes. Not sure how the error would be classified atm or after KIP-698. But it might be out-of-scope for this PR. Would like to focus on authorization errors for now.

> Where do we check?

Doesn't kslqDB create output topics explicitly if they don't exist?

PR #7063, opened 2021-02-22
fix: need to name schema field after tightening upstream validation

ableegoldman:
Similar to https://github.com/confluentinc/ksql/pull/7052

Issue surfaced after tightening up the compatibility check in confluentinc/schema-registry#1775

### DIFF
--- a/ksqldb-functional-tests/src/test/resources/query-validation-tests/serdes.json
+++ b/ksqldb-functional-tests/src/test/resources/query-validation-tests/serdes.json
@@ -81,7 +81,7 @@
       "topics": [
         {
           "name": "input_topic",
-          "schema": {"name": "ignored", "type": "record", "fields": [{"name": "FOO", "type": "int"}]}
+          "schema": {"name": "KsqlDataSourceSchema", "type": "record", "fields": [{"name": "FOO", "type": "int"}]}
         }
       ],
       "inputs": [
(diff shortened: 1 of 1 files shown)


agavra:
Aren't this and #7052 are indicative of backwards incompatible changes? Unless I'm not understanding something, we should allow users to specify whatever schema they want, they shouldn't need to use our internal name when declaring a schema.

ableegoldman:
@agavra this doesn't have anything to do with users, it's just causing our own internal tests to fail (due to incompatible schema). But maybe I'm missing your point?

agavra:
The idea behind the QTT test here is that the `topics` node specifies the input topics and their schemas as they would be registered in SR (exactly as the user has them). If this test is failing, then that means that previously a user was allowed to specify an input schema of 
```
{"name": "ignored", "type": "record", "fields": [{"name": "FOO", "type": "int"}]}
```

And still declare a stream on top of it with the following:
```
CREATE STREAM INPUT (foo INT) WITH (WRAP_SINGLE_VALUE=true, kafka_topic='input_topic', value_format='AVRO');
```

After https://github.com/confluentinc/schema-registry/pull/1775, it looks like we would no longer be allowing that. I think that's a bug, we should never require our users to specify something like `KsqlDataSourceSchema` in their schemas in order to use them in ksql.

agavra:
This might not still be the case in more recent versions (we changed it so that we don't register schemas into the user's subject for the source topic) but we should backport that change here if we need it.

It would be good to post the error message and stack trace as part of this PR.

rayokota:
@agavra , the tests were relying on buggy behavior in SR.  The first version was registered with record name `ignored`, but the second version was registered with record name `KsqlDataSourceSchema`.  According to the Avro spec, records with different names are incompatible.  This was recently fixed in SR.  It appears these tests are broken on 5.5.x through 6.1.x but fine on `master`.

agavra:
@rayokota - yup, I think this is a bug in ksql not schema registry. I'm just saying that we can't release this SR fix without also fixing ksqlDB.

ableegoldman:
Closing in favor of https://github.com/confluentinc/ksql/pull/7096

PR #7064, opened 2021-02-22
chore(migrations): utility functions for retrieving migration files and finding the current version

jzaralim:
### Description 
* Moves `MigrationUtil` into a util folder that also includes `MetadataUtil` for handling the migration stream/table, and `MigrationsDirectoryUtil` for handling the migrations directory.
* Adds functions in title

### Reviewer checklist

### DIFF
--- a/ksqldb-tools/src/main/java/io/confluent/ksql/tools/migrations/util/MetadataUtil.java
+++ b/ksqldb-tools/src/main/java/io/confluent/ksql/tools/migrations/util/MetadataUtil.java
@@ -23,19 +23,23 @@ import io.confluent.ksql.tools.migrations.MigrationException;
 import java.util.List;
 import java.util.concurrent.ExecutionException;
 
-public class MetadataUtil {
+public final class MetadataUtil {
+
+  public static final String NONE_VERSION = "<none>";
+  public static final String CURRENT_VERSION = "CURRENT";
+
   private MetadataUtil() {
   }
 
-  public static String getCurrentVersion(final MigrationConfig config) {
-    final Client client = MigrationsUtil.getKsqlClient(config);
+  public static String getCurrentVersion(final MigrationConfig config, final Client client) {
     final String migrationTableName = config.getString(MigrationConfig.KSQL_MIGRATIONS_TABLE_NAME);
-    BatchedQueryResult result = client.executeQuery(
-        "SELECT VERSION FROM " + migrationTableName + " WHERE version_key = 'CURRENT';");
+    final BatchedQueryResult result = client.executeQuery(
+        "SELECT VERSION FROM " + migrationTableName + " WHERE version_key = '"
+            + CURRENT_VERSION + "';");
     try {
-      List<Row> resultRows = result.get();
+      final List<Row> resultRows = result.get();
       if (resultRows.size() == 0) {
-        return "0";
+        return NONE_VERSION;
       }
       return resultRows.get(0).getString("VERSION");
     } catch (InterruptedException | ExecutionException e) {
--- a/ksqldb-tools/src/main/java/io/confluent/ksql/tools/migrations/util/MetadataUtil.java
+++ b/ksqldb-tools/src/main/java/io/confluent/ksql/tools/migrations/util/MetadataUtil.java
@@ -26,7 +26,7 @@ import java.util.concurrent.ExecutionException;
 public final class MetadataUtil {
 
   public static final String NONE_VERSION = "<none>";
-  public static final String CURRENT_VERSION = "CURRENT";
+  public static final String CURRENT_VERSION_KEY = "CURRENT";
 
   private MetadataUtil() {
   }
@@ -35,7 +35,7 @@ public final class MetadataUtil {
     final String migrationTableName = config.getString(MigrationConfig.KSQL_MIGRATIONS_TABLE_NAME);
     final BatchedQueryResult result = client.executeQuery(
         "SELECT VERSION FROM " + migrationTableName + " WHERE version_key = '"
-            + CURRENT_VERSION + "';");
+            + CURRENT_VERSION_KEY + "';");
     try {
       final List<Row> resultRows = result.get();
       if (resultRows.size() == 0) {
@@ -45,8 +45,6 @@ public final class MetadataUtil {
     } catch (InterruptedException | ExecutionException e) {
       throw new MigrationException(
           String.format("Could not query %s: %s", migrationTableName, e.getMessage()));
-    } finally {
-      client.close();
     }
   }
 }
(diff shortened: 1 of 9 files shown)


ksqldb-tools/src/main/java/io/confluent/ksql/tools/migrations/util/MetadataUtil.java
```diff
@@ -0,0 +1,48 @@
...
+import io.confluent.ksql.tools.migrations.MigrationConfig;
+import io.confluent.ksql.tools.migrations.MigrationException;
+import java.util.List;
+import java.util.concurrent.ExecutionException;
+
+public class MetadataUtil {
+  private MetadataUtil() {
+  }
+
+  public static String getCurrentVersion(final MigrationConfig config) {
```
vcrfxia, line 30: Since both `apply` and `validate` will be making multiple client calls, I think it'd be better to have this method accept the client as a param (and leave it up to the calling commands to handle properly closing the client). 

If you can think of use cases where the only client call is to get the current version, we can have a convenience overload that creates and closes the client as this implementation currently does, but primarily we should be using the other version (that supplies the client).

ksqldb-tools/src/main/java/io/confluent/ksql/tools/migrations/util/MetadataUtil.java
```diff
@@ -0,0 +1,48 @@
...
+    final Client client = MigrationsUtil.getKsqlClient(config);
+    final String migrationTableName = config.getString(MigrationConfig.KSQL_MIGRATIONS_TABLE_NAME);
+    BatchedQueryResult result = client.executeQuery(
+        "SELECT VERSION FROM " + migrationTableName + " WHERE version_key = 'CURRENT';");
+    try {
+      List<Row> resultRows = result.get();
+      if (resultRows.size() == 0) {
+        return "0";
```
vcrfxia, line 38: Let's pull this out into a static constant.

ksqldb-tools/src/main/java/io/confluent/ksql/tools/migrations/util/MetadataUtil.java
```diff
@@ -0,0 +1,48 @@
...
+
+public class MetadataUtil {
+  private MetadataUtil() {
+  }
+
+  public static String getCurrentVersion(final MigrationConfig config) {
+    final Client client = MigrationsUtil.getKsqlClient(config);
+    final String migrationTableName = config.getString(MigrationConfig.KSQL_MIGRATIONS_TABLE_NAME);
+    BatchedQueryResult result = client.executeQuery(
+        "SELECT VERSION FROM " + migrationTableName + " WHERE version_key = 'CURRENT';");
```
vcrfxia, line 34: Let's also pull `CURRENT` out into a static constant.

ksqldb-tools/src/main/java/io/confluent/ksql/tools/migrations/util/MigrationsDirectoryUtil.java
```diff
@@ -0,0 +1,55 @@
...
+import java.nio.charset.StandardCharsets;
+import java.nio.file.Files;
+import java.nio.file.Paths;
+import org.apache.commons.lang3.StringUtils;
+
+public class MigrationsDirectoryUtil {
+  private MigrationsDirectoryUtil() {
+  }
+
+  public static String getFileNameForVersion(final String version, final String path) {
```
vcrfxia, line 30:
```suggestion
  public static String getFileNameForVersion(final String version, final String migrationsDir) {
```

ksqldb-tools/src/main/java/io/confluent/ksql/tools/migrations/util/MigrationsDirectoryUtil.java
```diff
@@ -0,0 +1,55 @@
...
+    final File directory = new File(path);
+    for (String name : directory.list()) {
+      if (name.startsWith(prefix)) {
+        return name;
+      }
+    }
+    return null;
+  }
+
+  public static String getFileForVersion(final String version, final String path) {
```
vcrfxia, line 41:
```suggestion
  public static String getFileContentsForVersion(final String version, final String migrationsDir) {
```

ksqldb-tools/src/main/java/io/confluent/ksql/tools/migrations/util/MigrationsDirectoryUtil.java
```diff
@@ -0,0 +1,55 @@
...
+
+  public static String getFileNameForVersion(final String version, final String path) {
+    final String prefix = "V" + StringUtils.leftPad(version, 6, "0");
+    final File directory = new File(path);
+    for (String name : directory.list()) {
+      if (name.startsWith(prefix)) {
+        return name;
+      }
+    }
+    return null;
```
vcrfxia, line 38: Make the return type `Optional<String>`, rather than returning null?

vcrfxia (approved):
Thanks @jzaralim ! LGTM with minor comments inline.

ksqldb-tools/src/main/java/io/confluent/ksql/tools/migrations/util/MetadataUtil.java
vcrfxia: We can have `public static final String NONE_VERSION = "<none>";` and use it for two purposes: (1) no current version when this method is called, (2) the "previous version" for the first migration applied. By having a more user-readable name, it won't be confusing when displayed as part of the output of `info`.

ksqldb-tools/src/main/java/io/confluent/ksql/tools/migrations/util/MetadataUtil.java
```diff
@@ -0,0 +1,52 @@
...
+      final List<Row> resultRows = result.get();
+      if (resultRows.size() == 0) {
+        return NONE_VERSION;
+      }
+      return resultRows.get(0).getString("VERSION");
+    } catch (InterruptedException | ExecutionException e) {
+      throw new MigrationException(
+          String.format("Could not query %s: %s", migrationTableName, e.getMessage()));
+    } finally {
+      client.close();
```
vcrfxia, line 49: This method should not close the client; it should be up to the individual commands to create and close the client. That way we can reuse the client for multiple requests.

ksqldb-tools/src/main/java/io/confluent/ksql/tools/migrations/util/MetadataUtil.java
```diff
@@ -0,0 +1,52 @@
...
+import io.confluent.ksql.api.client.Row;
+import io.confluent.ksql.tools.migrations.MigrationConfig;
+import io.confluent.ksql.tools.migrations.MigrationException;
+import java.util.List;
+import java.util.concurrent.ExecutionException;
+
+public final class MetadataUtil {
+
+  public static final String NONE_VERSION = "<none>";
+  public static final String CURRENT_VERSION = "CURRENT";
```
vcrfxia, line 29: nit:
```suggestion
  public static final String CURRENT_VERSION_KEY = "CURRENT";
```

PR #7067, opened 2021-02-23
docs: fix synthetic key example

agavra:
### Description 

Original example had a table repartition, which is not supported.

### Testing done 

Ran the example locally

### Reviewer checklist


JimGalasyn (approved):
Thanks for the fix! I'll cherry-pick to live and dev branches.

PR #7078, opened 2021-02-23
fix: quick fix for appender

wcarlson5:
update the default for the kafka log appender to the producer send buffer does not fill up on a bad configuration

### Reviewer checklist


guozhangwang:
Should we fix `log4j-files` and `log4j` properties file as well?

wcarlson5:
@guozhangwang they don't use the same appender, I don't think they have the same problem

rodesai (approved):
LGTM

PR #7085, opened 2021-02-24
feat: Introduce materialization support for tables using CTAS to make tables queryable

### LINKED ISSUE #4614: Improved error message when querying non-materialized tables

Today, when I try to run a pull query against a table that isn't materialized, I get a pretty large error message:

```
'MOVIE_TICKET_SALES' is not materialized. Refer to https://cnfl.io/queries for info on query types. If you intended to issue a push query, resubmit with the EMIT CHANGES clause
 KSQL currently only supports pull queries on materialized aggregate tables. i.e. those created by a 'CREATE TABLE AS SELECT <fields>, <aggregate_functions> FROM <sources> GROUP BY <key>' style statement.
Query syntax in KSQL has changed. There are now two broad categories of queries:
- Pull queries: query the current state of the system, return a result, and terminate. 
- Push queries: query the state of the system in motion and continue to output results until they meet a LIMIT condition or are terminated by the user.

'EMIT CHANGES' is used to to indicate a query is a push query. To convert a pull query into a push query, which was the default behavior in older versions of KSQL, add `EMIT CHANGES` to the end of the statement before any LIMIT clause.

For example, the following are pull queries:
	'SELECT * FROM X WHERE ROWKEY=Y;' (non-windowed table)
	'SELECT * FROM X WHERE ROWKEY=Y AND WINDOWSTART>=Z;' (windowed table)

The following is a push query:
	'SELECT * FROM X EMIT CHANGES;'

Note: Persistent queries, e.g. `CREATE TABLE AS ...`, have an implicit `EMIT CHANGES`, but we recommend adding `EMIT CHANGES` to these statements for clarify.
```

There are lots of good things about this message. It basically says three things:

1. pull queries only work on materialized tables
2. there are new queries (push and pull), and that's different than previous versions
3. emit changes is how you control the query type

This is good information, but we've heard a few times that people read this and still don't quite understand what they did wrong. The root of this is understanding that the table isn't materialized, and really understanding what that is. We describe that it's a chained table from another statement, but that's still murky for newcomers.

We should explain what a materialized table actually is. I don't think this needs to be big -- it can even link out the docs. But without that, the real source of the error doesn't make sense.

cprasad1:
### Description 
Introduces materialization support for tables using CTAS to make tables queryable. 
We can make tables queryable by using:
`CREATE TABLE QUERYABLE_TABLE AS SELECT...`

This would help fix #4614 by adding a more helpful error message as well.

### Testing done 
- Unit tests
- Query translation tests

### Reviewer checklist

### DIFF
--- a/ksqldb-engine/src/main/java/io/confluent/ksql/engine/PullQueryExecutionUtil.java
+++ b/ksqldb-engine/src/main/java/io/confluent/ksql/engine/PullQueryExecutionUtil.java
@@ -110,8 +110,4 @@ public final class PullQueryExecutionUtil {
                     + PullQueryValidator.PULL_QUERY_SYNTAX_HELP
     );
   }
-
-
-
-
 }
--- a/ksqldb-engine/src/main/java/io/confluent/ksql/engine/PullQueryExecutionUtil.java
+++ b/ksqldb-engine/src/main/java/io/confluent/ksql/engine/PullQueryExecutionUtil.java
@@ -101,12 +101,12 @@ public final class PullQueryExecutionUtil {
   private static KsqlException notMaterializedException(final SourceName sourceTable) {
     final String tableName = sourceTable.toString().replaceAll("`", "");
     return new KsqlException(
-            "The " + sourceTable + " table isn't queriable. To derive a queriable table, "
-                    + "you can do 'CREATE TABLE QUERIABLE_"
+            "The " + sourceTable + " table isn't queryable. To derive a queryable table, "
+                    + "you can do 'CREATE TABLE QUERYABLE_"
                     + tableName
                     + " AS SELECT * FROM "
                     + tableName
-                    + "."
+                    + "'."
                     + PullQueryValidator.PULL_QUERY_SYNTAX_HELP
     );
   }
--- a/ksqldb-engine/src/main/java/io/confluent/ksql/engine/PullQueryExecutionUtil.java
+++ b/ksqldb-engine/src/main/java/io/confluent/ksql/engine/PullQueryExecutionUtil.java
@@ -99,11 +99,18 @@ public final class PullQueryExecutionUtil {
   }
 
   private static KsqlException notMaterializedException(final SourceName sourceTable) {
+    final String tableName = sourceTable.toString().replaceAll("`", "");
     return new KsqlException(
-        "Can't pull from " + sourceTable + " as it's not a materialized table."
-            + PullQueryValidator.PULL_QUERY_SYNTAX_HELP
+            "The " + sourceTable + " table isn't queriable. To derive a queriable table, "
+                    + "you can do 'CREATE TABLE QUERIABLE_"
+                    + tableName
+                    + " AS SELECT * FROM "
+                    + tableName
+                    + "."
+                    + PullQueryValidator.PULL_QUERY_SYNTAX_HELP
     );
   }
+
 
 
 
--- a/ksqldb-engine/src/main/java/io/confluent/ksql/physical/pull/PullPhysicalPlanBuilder.java
+++ b/ksqldb-engine/src/main/java/io/confluent/ksql/physical/pull/PullPhysicalPlanBuilder.java
@@ -203,14 +203,14 @@ public class PullPhysicalPlanBuilder {
   }
 
   private KsqlException notMaterializedException(final SourceName sourceTable) {
-    final String tableName = sourceTable.toString().replaceAll("`", "");
+    final String tableName = sourceTable.text();
     return new KsqlException(
-        "The " + sourceTable + " table isn't queriable. To derive a queriable table, "
-                + "you can do 'CREATE TABLE QUERIABLE_"
+        "The " + sourceTable + " table isn't queryable. To derive a queryable table, "
+                + "you can do 'CREATE TABLE QUERYABLE_"
                 + tableName
                 + " AS SELECT * FROM "
                 + tableName
-                + "."
+                + "'."
                 + PullQueryValidator.PULL_QUERY_SYNTAX_HELP
     );
   }
--- a/ksqldb-engine/src/main/java/io/confluent/ksql/physical/pull/PullPhysicalPlanBuilder.java
+++ b/ksqldb-engine/src/main/java/io/confluent/ksql/physical/pull/PullPhysicalPlanBuilder.java
@@ -203,9 +203,15 @@ public class PullPhysicalPlanBuilder {
   }
 
   private KsqlException notMaterializedException(final SourceName sourceTable) {
+    final String tableName = sourceTable.toString().replaceAll("`", "");
     return new KsqlException(
-        "Can't pull from " + sourceTable + " as it's not a materialized table."
-            + PullQueryValidator.PULL_QUERY_SYNTAX_HELP
+        "The " + sourceTable + " table isn't queriable. To derive a queriable table, "
+                + "you can do 'CREATE TABLE QUERIABLE_"
+                + tableName
+                + " AS SELECT * FROM "
+                + tableName
+                + "."
+                + PullQueryValidator.PULL_QUERY_SYNTAX_HELP
     );
   }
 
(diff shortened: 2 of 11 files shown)


ksqldb-engine/src/main/java/io/confluent/ksql/engine/PullQueryExecutionUtil.java
```diff
@@ -99,12 +99,19 @@ private static SourceName getSourceName(final ImmutableAnalysis analysis) {
...
+    final String tableName = sourceTable.toString().replaceAll("`", "");
     return new KsqlException(
-        "Can't pull from " + sourceTable + " as it's not a materialized table."
-            + PullQueryValidator.PULL_QUERY_SYNTAX_HELP
+            "The " + sourceTable + " table isn't queriable. To derive a queriable table, "
+                    + "you can do 'CREATE TABLE QUERIABLE_"
+                    + tableName
+                    + " AS SELECT * FROM "
+                    + tableName
+                    + "."
```
vvcephei, line 109:
```suggestion
                    + "'."
```

(my entry for most cryptic feedback of the year)

ksqldb-functional-tests/src/test/resources/rest-query-validation-tests/pull-queries-against-ctas-materialized.json
```diff
@@ -0,0 +1,235 @@
...
+        {"admin": {"@type": "currentStatus"}},
+        {"admin": {"@type": "currentStatus"}},
+        {"query": [
+          {"header":{"schema":"`ID` BIGINT KEY, `USERTIMESTAMP` BIGINT, `GENDER` STRING, `REGION_ID` STRING"}},
+          {"row":{"columns":[2,2000,"f","e"]}}
+        ]}
+      ]
+    },
+    {
+      "name": "select * against mat table windowed table scan",
```
vvcephei, line 51: These queries look like key lookups. Did I miss something?

vvcephei (commented):
Thanks, @cprasad1 !

I can't speak to many of the little changes in internal method calls, etc., but at a high level, the error message and the tests look good: exactly what we wanted to achieve.

Minor note: we prefer "queryable" to "queriable".

Also, specifically in the tests, I'm wondering if we can re-structure them to be a little more compact and have better coverage, something like this:
* Have each case set up a CT or CTAS statement for query in a slightly different way, covering all the possible permutations (like a basic select *, one that includes a WHERE clause, one that includes a more complicated projection, one that does both, one that does a groupBy agg from a stream, one that does a join, etc. etc.).
* Then, inside each case, we just run multiple pull queries, for each slight variant of pull queries we want to test

In other words, we vary the queriable object between tests and vary the pill queries within each test. That way, we strike a good balance between code coverage and proliferation of test cases.

ksqldb-streams/src/main/java/io/confluent/ksql/execution/streams/TableTableJoinBuilder.java
```diff
@@ -52,6 +53,6 @@ private TableTableJoinBuilder() {
         throw new IllegalStateException("invalid join type");
     }
 
-    return left.withTable(result, joinParams.getSchema());
+    return left.withTable(result, joinParams.getSchema(), Optional.empty());
```
AlanConfluent, line 56: Why do you need to explicitly avoid passing the materialization info builder here?  Just curious.  Is it because joins aren't materialized and the old code would have implied that they were?

You could probably do `KTableHolder.unmaterialized(...)` and pass in the ExecutionKeyFactory from left to make it a bit more explicit.

ksqldb-functional-tests/src/test/resources/rest-query-validation-tests/pull-queries-against-ctas-materialized.json
```diff
@@ -0,0 +1,235 @@
...
+      "statements": [
+        "CREATE TABLE users (ID BIGINT PRIMARY KEY, USERTIMESTAMP BIGINT,gender VARCHAR,region_id VARCHAR) WITH (KAFKA_TOPIC = 'users_topic',VALUE_FORMAT = 'JSON',PARTITIONS=1);",
+        "CREATE TABLE queriable_users AS SELECT ID, COUNT(*) AS COUNT from users GROUP BY ID;",
```
AlanConfluent, line 54: This would have been materialized previously.  I think we have a lot of tests for this already.

AlanConfluent (commented):
I just had a few small comments, but looks good.

ksqldb-streams/src/main/java/io/confluent/ksql/execution/streams/TableTableJoinBuilder.java
cprasad1: >  Is it because joins aren't materialized and the old code would have implied that they were?

exactly this reason

ksqldb-engine/src/main/java/io/confluent/ksql/physical/pull/PullPhysicalPlanBuilder.java
```diff
@@ -203,9 +203,15 @@ private QueryId uniqueQueryId() {
   }
 
   private KsqlException notMaterializedException(final SourceName sourceTable) {
+    final String tableName = sourceTable.toString().replaceAll("`", "");
```
agavra, line 206: just noting that you can use `sourceTable.text()` instead of using a replace here

vvcephei: Oh, man, I overlooked this. Yeah, we shouldn't ever rely on `toString()` for program logic.

ksqldb-execution/src/main/java/io/confluent/ksql/execution/plan/KTableHolder.java
```diff
@@ -88,6 +88,13 @@ public LogicalSchema getSchema() {
     return new KTableHolder<>(table, schema, executionKeyFactory, materializationBuilder);
   }
 
+  public KTableHolder<K> withTable(
+          final KTable<K, GenericRow> table,
+          final LogicalSchema schema,
+          final Optional<MaterializationInfo.Builder> builder) {
+    return new KTableHolder<>(table, schema, executionKeyFactory, builder);
```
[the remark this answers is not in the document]
guozhangwang: I think the reason is to inherit the `executionKeyFactory` from the parent while all others are overridden.

I'm thinking that since we are replacing all others except the `executionKeyFactory`, maybe we should just remove this function and on the caller just call

```
materialized(table, schema, parent.getExecutionKeyFactory(), materializedBuilder)
```

instead?

ksqldb-streams/src/main/java/io/confluent/ksql/execution/streams/TableTableJoinBuilder.java
guozhangwang: +1 on using `unmaterialized()` instead.

guozhangwang (commented):
The PR LGTM, but I had a meta question regarding the description:

```
CREATE TABLE QUERYABLE_TABLE AS SELECT...
```

Are we really going to augment the syntax? In this PR it seems we just make CTAT always materialized anyways.

cprasad1:
> The PR LGTM, but I had a meta question regarding the description:
> 
> ```
> CREATE TABLE QUERYABLE_TABLE AS SELECT...
> ```
> 
> Are we really going to augment the syntax? In this PR it seems we just make CTAT always materialized anyways.

@guozhangwang what are you referring to when you say "augment"? If you were referring to the `QUERYABLE_` part of `QUERYABLE_TABLE`, then that is just the dummy name of the table that is just used as an example here

guozhangwang:
> @guozhangwang what are you referring to when you say "augment"? If you were referring to the `QUERYABLE_` part of `QUERYABLE_TABLE`, then that is just the dummy name of the table that is just used as an example here

My bad, I thought it was some new keywords :P

AlanConfluent (approved):

vvcephei (approved):
Thanks, @cprasad1 !

PR #7086, opened 2021-02-24
fix: Table Scan: Filters out StreamsMetadata for wrong subtopology

AlanConfluent:
### Description 
The call to `kafkaStreams.allMetadataForStore(stateStoreName)` returns metadata for all subtopologies associated with the store, including ones which don't actually contain the store.  The calling code assumed that it was for just one subtopology, so this resulted in a bug where TopicPartitions from the wrong subtopology could overwrite the intended one.

### Testing done 
Unit tests and manual testing with table scans in a multi node environment.

### Reviewer checklist

### DIFF
--- a/ksqldb-streams/src/main/java/io/confluent/ksql/execution/streams/materialization/ks/KsLocator.java
+++ b/ksqldb-streams/src/main/java/io/confluent/ksql/execution/streams/materialization/ks/KsLocator.java
@@ -189,8 +189,6 @@ public final class KsLocator implements Locator {
       streamsMetadata.topicPartitions().forEach(
           tp -> {
             if (sourceTopics.contains(tp.topic())) {
-              LOG.info("Active: store: " + stateStoreName + " topic partition: " + tp + " host: "
-                  + streamsMetadata.hostInfo());
               activeHostByPartition.put(tp.partition(), streamsMetadata.hostInfo());
             }
           });
@@ -198,8 +196,6 @@ public final class KsLocator implements Locator {
       streamsMetadata.standbyTopicPartitions().forEach(
           tp -> {
             if (sourceTopics.contains(tp.topic())) {
-              LOG.info("Standby: store: " + stateStoreName + " topic partition: " + tp + " host: "
-                  + streamsMetadata.hostInfo());
               standbyHostsByPartition.computeIfAbsent(tp.partition(), p -> new HashSet<>());
               standbyHostsByPartition.get(tp.partition()).add(streamsMetadata.hostInfo());
             }
@@ -222,10 +218,6 @@ public final class KsLocator implements Locator {
           Collections.emptySet());
       metadataList.add(
           new PartitionMetadata(activeHost, standbyHosts, partition, Optional.empty()));
-    }
-
-    for (PartitionMetadata pm : metadataList) {
-      LOG.info("Got metadata partition: " + pm.getPartition() + " active: " + pm.getActiveHost() + " standbys: " + pm.getStandbyHosts());
     }
     return metadataList;
   }
--- a/ksqldb-streams/src/main/java/io/confluent/ksql/execution/streams/materialization/ks/KsLocator.java
+++ b/ksqldb-streams/src/main/java/io/confluent/ksql/execution/streams/materialization/ks/KsLocator.java
@@ -81,7 +81,7 @@ public final class KsLocator implements Locator {
       final String applicationId
   ) {
     this.kafkaStreams = requireNonNull(kafkaStreams, "kafkaStreams");
-    this.topology = topology;
+    this.topology = requireNonNull(topology, "topology");
     this.keySerializer = requireNonNull(keySerializer, "keySerializer");
     this.stateStoreName = requireNonNull(stateStoreName, "stateStoreName");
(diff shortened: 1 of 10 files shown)


ksqldb-streams/src/main/java/io/confluent/ksql/execution/streams/materialization/ks/KsLocator.java
```diff
@@ -207,6 +222,38 @@
     return metadataList;
   }
 
+  private Set<String> findSubtopologySourceTopics() {
```
agavra, line 225: yikes 😨 - probably makes sense to heavily document what this is doing

agavra (commented):
Thanks for both finding and fixing this gnarly bug! I'll wait for a KS guru to review, but was wondering if this bug is reproducible in a unit test?

guozhangwang (commented):
I only checked the `findSubtopologySourceTopics` logic and it lgtm.

ksqldb-streams/src/main/java/io/confluent/ksql/execution/streams/materialization/ks/KsLocator.java
```diff
@@ -174,16 +182,23 @@
    */
   private List<PartitionMetadata>  getMetadataForAllPartitions(
       final Set<Integer> filterPartitions) {
+    final Set<String> sourceTopics = findSubtopologySourceTopics();
```
vvcephei, line 185: If you choose to add some docs, it'd be worth explaining that the sources may have different numbers of partitions, but the store partitioning is determined by the _maximum_ number of partitions of _any_ source topic, which the algorithm mapping partition to host, below, achieves. It took me a minute to see that it's correct..

vvcephei (approved):
Thanks, @AlanConfluent !

This looks right to me; I just have some sanity-check questions about the interpretation of the TopologyDescription.

ksqldb-streams/src/main/java/io/confluent/ksql/execution/streams/materialization/ks/KsLocator.java
```diff
@@ -207,6 +222,38 @@
...
+        continue;
+      }
+
+      for (final TopologyDescription.Node node : subtopology.nodes()) {
+        if (node instanceof Source) {
+          final Source source = (Source) node;
+          if (source.topicSet() != null) {
+            return source.topicSet()
+                .stream()
+                .map(name -> applicationId + "-" + name)
```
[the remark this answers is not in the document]
guozhangwang: +1. Good catch.

AlanConfluent:
> Thanks for both finding and fixing this gnarly bug! I'll wait for a KS guru to review, but was wondering if this bug is reproducible in a unit test?

I did exercise this case in the table scan test in `KsLocatorTest`.  This is still a unit test though.  It would be a bit harder to do that in a functional test, though that's not a bad idea.  Let me look into that.

ksqldb-streams/src/main/java/io/confluent/ksql/execution/streams/materialization/ks/KsLocator.java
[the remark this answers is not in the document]
AlanConfluent: If the user chose a name that collided with an internal name, couldn't this incorrectly choose the wrong source topic?  I know it's a bit of a contrived case, but it seems possible.

ksqldb-streams/src/main/java/io/confluent/ksql/execution/streams/materialization/ks/KsLocator.java
AlanConfluent: I added a bunch of comments that aim to clarify this.

ksqldb-streams/src/main/java/io/confluent/ksql/execution/streams/materialization/ks/KsLocator.java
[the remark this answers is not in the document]
guozhangwang: Today we do not forbid that programmatically and if that happens, it would be of a bigger issue: during the first rebalance when we would like to create the intermediate topic, if collides with the existing source topics, then it would throw if the number of partitions do not match expected value, or would assume the topic is already created if they match. And we could potentially have a loop where a task read as source and write as sink the same topic.

AlanConfluent:
> Thanks for both finding and fixing this gnarly bug! I'll wait for a KS guru to review, but was wondering if this bug is reproducible in a unit test?

I added a functional test that should exercise this too.  It's a bit complex in that we have no guarantee on the order that the API returns TopicPartitions and that the assignments are made, but at least it should exercise it.

ksqldb-streams/src/main/java/io/confluent/ksql/execution/streams/materialization/ks/KsLocator.java
```diff
@@ -174,16 +183,34 @@
    */
   private List<PartitionMetadata>  getMetadataForAllPartitions(
       final Set<Integer> filterPartitions) {
+    // It's important that we consider only the source topics for the subtopology that contains the
```
agavra, line 186: thanks for the comments :)

guozhangwang (approved):

PR #7087, opened 2021-02-24
feat(migrations): implement `validate` command

vcrfxia:
### Description 

The `validate` command of the migrations tool:
1. Finds the latest migrated version by querying the metadata table for the current version and following "previous" pointers until one with status MIGRATED is found. This should never involve more than one hop following pointers since the `apply` command should always use the latest MIGRATED version as "previous". Currently an error is thrown if this is not the case, but we can revisit when we introduce the `abort` command to get out of bad states due to race conditions.
2. Continuing to follow "previous" pointers, validating each file's checksum (and that each version has status MIGRATED) along the way.
3. The "previous" pointers eventually lead to the special `NONE_VERSION`, indicating the first migration version has been reached.

Once table scans are implemented, we can simplify the algorithm by simply doing a table scan and filtering for status MIGRATED.

### Testing done 

Lots of unit tests. Integration tests will be added with the `apply` implementation.

### Reviewer checklist

### DIFF
--- a/ksqldb-tools/src/main/java/io/confluent/ksql/tools/migrations/commands/ValidateMigrationsCommand.java
+++ b/ksqldb-tools/src/main/java/io/confluent/ksql/tools/migrations/commands/ValidateMigrationsCommand.java
@@ -15,6 +15,9 @@
 
 package io.confluent.ksql.tools.migrations.commands;
 
+import static io.confluent.ksql.tools.migrations.util.MetadataUtil.getInfoForVersion;
+import static io.confluent.ksql.tools.migrations.util.MetadataUtil.getLatestMigratedVersion;
+import static io.confluent.ksql.tools.migrations.util.MetadataUtil.validateVersionIsMigrated;
 import static io.confluent.ksql.tools.migrations.util.MigrationsDirectoryUtil.computeHashForFile;
 import static io.confluent.ksql.tools.migrations.util.MigrationsDirectoryUtil.getFilePathForVersion;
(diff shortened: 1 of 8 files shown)


ksqldb-tools/src/main/java/io/confluent/ksql/tools/migrations/util/MetadataUtil.java
```diff
@@ -47,4 +55,105 @@ public static String getCurrentVersion(final MigrationConfig config, final Clien
           String.format("Could not query %s: %s", migrationTableName, e.getMessage()));
     }
   }
+
+  public static String getLatestMigratedVersion(
```
vcrfxia, line 59: @jzaralim this method will be useful for implementing `apply`, in order to get the latest version to start applying commands from.

ksqldb-tools/src/main/java/io/confluent/ksql/tools/migrations/commands/ValidateMigrationsCommand.java
```diff
@@ -37,12 +55,102 @@
...
 
   @Override
   protected Logger getLogger() {
     return null;
   }
 
+  /**
+   * @return true if validation passes, else false.
+   */
+  static boolean validate(
```
vcrfxia, line 118: This method is non-private since we'll want to call it from the `apply` command, in order to validate applied migrations prior to submitting any new ones ( cc @jzaralim )

jzaralim (approved):
this is super helpful!

PR #7093, opened 2021-02-24
fix: get unique function names in correct order

stevenpyzhang:
### Description 
The code generated for nested UDFs is done incorrectly currently. If I have the following expression
`REGEXP_EXTRACT('.*', REGEXP_EXTRACT('(.*) (.*)', input_string, 2))`, this fails to process records correctly because the wrong function identifier is in the generated code.

Currently in CodeGenRunner, we add the unique function name to the CodeGenSpec after we've processed all function call arguments (child nodes) of the parent node so a PostOrder traversal to add function names to spec

In SqlToJavaVisitor, we get the function name of the parent node before we process the child node (and get their function name from spec) which is a PreOrder traversal to get the function names from the spec

This results in the ordering of the function names to be wrong in the generated code.

We need to align the order we get the function name with the way we add it. I ended up going with making the CodeGenRunner a PreOrder traversal also

### Testing done 
Add a new QTT test

### Reviewer checklist


agavra (approved):
Great job finding this bug and very elegant fix :)

guozhangwang (approved):

PR #7096, opened 2021-02-25
chore: remove unnecessary output schema in serde QTT

agavra:
### Description 

https://github.com/confluentinc/schema-registry/pull/1775 exposed an issue in our test, we were "overwriting" schemas that we had registered under the output subject in our tests. Registering these are pointless, as up until now they haven't enforced any true compatibility and in future versions we have `PostTopicNode` to ensure that the generated schema is as intended 

see https://github.com/confluentinc/ksql/blob/d064342cc0224569fc400d02b113c6432ed63cb7/ksqldb-functional-tests/src/main/java/io/confluent/ksql/test/model/PostConditionsNode.java#L169

### Testing done 

Ran the test

### Reviewer checklist

### DIFF
--- a/ksqldb-functional-tests/src/test/resources/query-validation-tests/serdes.json
+++ b/ksqldb-functional-tests/src/test/resources/query-validation-tests/serdes.json
@@ -229,10 +229,6 @@
         {
           "name": "input_topic",
           "schema": {"type": "map", "values": ["null", "int"]}
-        },
-        {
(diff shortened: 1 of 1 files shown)


ableegoldman (approved):
LGTM, assuming tests are green. Thanks for helping to look into this!

guozhangwang (approved):

PR #7097, opened 2021-02-25
feat: CLI should fail for unsupported server version

### LINKED ISSUE #3860: Throw an error if CLI attempts to connect to an incompatible server version

As mentioned in https://github.com/confluentinc/ksql/issues/3813, 5.4+ KSQL CLI verisons are incompatible with pre-5.4 server versions (i.e., 5.3 and below). Currently, a user can use a 5.4+ CLI and connect to a pre-5.4 server successfully, but any push queries issued will fail with bizarre error messages (related to the introduction of pull queries in 5.4).

Rather than allowing the connection followed by poor UX, the 5.4+ CLI should simply throw an error when attempting to connect to a pre-5.4 server.

n/a

**Additional context**

mjsax:
### Description 

Fixes #3860

The ksqlDB CLI require server version 5.4.0 and we should fail early
with a clear error message if an older server version is detected.

### Testing done 
If an error is detected, the following output will be printed
```
Copyright 2017-2021 Confluent Inc.

CLI v7.0.0-0, Server v0.10.0-0 located at http://someserver:8008

WARNING: CLI and server version don't match. This may lead to unexpected errors.
         It is recommended to use a CLI that matches the server version.

Server Status: status

Having trouble? Type 'help' (case-insensitive) for a rundown of how things work!
```

### Reviewer checklist

### DIFF
--- a/ksqldb-cli/src/test/java/io/confluent/ksql/cli/CliTest.java
+++ b/ksqldb-cli/src/test/java/io/confluent/ksql/cli/CliTest.java
@@ -910,10 +910,10 @@ public class CliTest {
   }
 
   @Test
-  public void shouldFailOnUnsupportedServerVersion() throws Exception {
+  public void shouldFailOnUnsupportedCpServerVersion() throws Exception {
     givenRunInteractivelyWillExit();
 
-    final KsqlRestClient mockRestClient = givenMockRestClient("5.3.0-0");
+    final KsqlRestClient mockRestClient = givenMockRestClient("5.5.0-0");
 
     assertThrows(
         KsqlUnsupportedServerException.class,
@@ -923,10 +923,23 @@ public class CliTest {
   }
 
   @Test
-  public void shouldPrintWarningOnDifferentServerVersion() throws Exception {
+  public void shouldFailOnUnsupportedStandaloneServerVersion() throws Exception {
     givenRunInteractivelyWillExit();
 
-    final KsqlRestClient mockRestClient = givenMockRestClient("5.4.0-0");
+    final KsqlRestClient mockRestClient = givenMockRestClient("0.9.0-0");
+
+    assertThrows(
+        KsqlUnsupportedServerException.class,
+        () -> new Cli(1L, 1L, mockRestClient, console)
+            .runInteractively()
+    );
+  }
+
+  @Test
+  public void shouldPrintWarningOnDifferentCpServerVersion() throws Exception {
+    givenRunInteractivelyWillExit();
+
+    final KsqlRestClient mockRestClient = givenMockRestClient("6.0.0-0");
 
     new Cli(1L, 1L, mockRestClient, console)
         .runInteractively();
@@ -939,22 +952,33 @@ public class CliTest {
   }
 
   @Test
-  public void shouldNotPrintWarningOnDifferentBugFixServerVersion() throws Exception {
+  public void shouldPrintWarningOnDifferentStandaloneServerVersion() throws Exception {
     givenRunInteractivelyWillExit();
 
-    final String cliVersion = AppInfo.getVersion();
-    final String[] tokens = cliVersion.split("-")[0].split("\\.");
-    final KsqlRestClient mockRestClient = givenMockRestClient(
-        tokens[0] + "." + tokens[1] + ".1" + tokens[2] + "-0"
-    );
+    final KsqlRestClient mockRestClient = givenMockRestClient("1.0.0-0");
 
     new Cli(1L, 1L, mockRestClient, console)
         .runInteractively();
 
     assertThat(
         terminal.getOutputString(),
-        not(containsString("WARNING: CLI and server version don't match."
-            + " This may lead to unexpected errors."))
+        containsString("WARNING: CLI and server version don't match."
+            + " This may lead to unexpected errors.")
+    );
+  }
+
+  @Test
+  public void shouldPrintWarningOnUnknownServerVersion() throws Exception {
+    givenRunInteractivelyWillExit();
+
+    final KsqlRestClient mockRestClient = givenMockRestClient("bad-version");
+
+    new Cli(1L, 1L, mockRestClient, console)
+        .runInteractively();
+
+    assertThat(
+        terminal.getOutputString(),
+        containsString("WARNING: Could not identify server version.")
     );
   }
 
@@ -1379,7 +1403,7 @@ public class CliTest {
     verify(mockRestClient).makeKsqlRequest(statementText, seqNum);
   }
 
-  @SuppressWarnings({"unchecked", "rawtypes"})
+  @SuppressWarnings("unchecked")
   private static Matcher<String>[] prependWithKey(final String key, final List<?> values) {
 
     final Matcher<String>[] allMatchers = new Matcher[values.size() + 1];
--- a/ksqldb-cli/src/test/java/io/confluent/ksql/cli/CliTest.java
+++ b/ksqldb-cli/src/test/java/io/confluent/ksql/cli/CliTest.java
@@ -955,7 +955,7 @@ public class CliTest {
   public void shouldPrintWarningOnDifferentStandaloneServerVersion() throws Exception {
     givenRunInteractivelyWillExit();
 
-    final KsqlRestClient mockRestClient = givenMockRestClient("1.0.0-0");
+    final KsqlRestClient mockRestClient = givenMockRestClient("0.10.0-0");
 
     new Cli(1L, 1L, mockRestClient, console)
         .runInteractively();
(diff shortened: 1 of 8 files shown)


vcrfxia:
Hey @mjsax I think we want a more general solution here, rather than a patch that only addresses the incompatibility between 5.4+ CLI versions and pre-5.4 server versions. If you look at the additional comments on the ticket (https://github.com/confluentinc/ksql/issues/3860), there are also incompatibilities between the 0.12.0 CLI and the 0.14.0 server, as well as between the 0.14.0 CLI and the 6.0.1 server. 

I think it makes sense to:
- print a large warning banner if the minor version of the CLI does not match that of the server, to the effect of "WARNING: the CLI version you are using (0.14) does not match the version of the ksqlDB server (0.15). This may lead to unexpected behavior. Please download the 0.15 CLI version at <insert link>". Note that we should only check minor versions, since the details of the version string may not match for expected reasons (IIRC, ccloud versions strings look a bit different). We may or may not want to build in conversion between CP versions and ksqlDB standalone versions (e.g., CP 6.1 aligns with ksqlDB 0.14). I don't think we need to, but we should ensure that the download link we provide is meaningful for both types of versions.
- fail the connection for known incompatibilities. If the user report that the 0.14.0 CLI is incompatible with the 6.0.1 server, then we should fail (on master) for connections to a CP 6.0 / ksqlDB 0.10.0 server. We can have a variable that we update when newer incompatibilities are discovered. If we choose to backport this patch to older versions, we can update the variable on older branches as necessary. (I don't think we need to backport, just throwing this out there as an "in case we do.")

It'd be good to get opinions from @colinhicks on the desired behavior as well.

Also, if you add "Fixes https://github.com/confluentinc/ksql/issues/3860" into the PR description, then the corresponding issue will be automatically closed when your PR is merged. Linking PRs and issues together is also useful for tracking :)

MichaelDrogalis:
Tagging in for @colinhicks on this one:

@vcrfxia I think what you described is exactly what we want here. We need a general mechanism to both warn on potentially bad server/CLI combinations and to fail known incompatible ones. I agree that we don't need to backport this—just to adopt it going forward.

vcrfxia (commented):
Hey @mjsax I haven't read through the entire PR in detail yet but I don't think this does quite what we want. There are two possible sets of ksqlDB version numbers: standalone ksqlDB versions (0.6.0, 0.7.0, ... 0.15.0) and Confluent Platform ksqlDB version numbers (5.4.0, 5.5.0, ... 6.1.0). We need to account for the possibility of encountering both, and we also need to account for the fact that some version numbers may be "unusual": for example, the 0.14.0 ksqlDB release is actually called 0.14.0-rc732 (due to a quirk in the release process), and we also want developers to be able to test with different version types (snapshots, etc). In general, I think we should err on the side of not failing connections if we encounter unexpected formats and can't confirm that versions are compatible, rather than hard-failing the connection attempt.

I recently hacked some server version code together as part of working on the migrations tool: it's definitely not perfect but maybe it'll be helpful as inspiration? https://github.com/confluentinc/ksql/blob/master/ksqldb-tools/src/main/java/io/confluent/ksql/tools/migrations/util/ServerVersionUtil.java#L108 If manipulating server versions become common and we want to standardize, we can move the existing code into the ksqldb-common repo and pull it in from both ksqldb-tools and ksqldb-cli.

Also, could you paste an example or two of what the ksqlDB CLI output on incompatible versions looks like into the PR description? That'll be helpful for reviewers and anyone else to wants to get a quick sense of what to expect from this PR.

spena (commented):
@mjsax, I was wondering if we should have an endpoint where to pull versions that are compatible with the server. That way we could just add new incompatibles versions in the future and let the CLI check for compatibility. 

This PR does the check in the CLI and it is hard-coded. If we have an incompatible version in the future, then the CLI with the hard-coded version will not fail correctly.

spena (approved):
LGTM

mjsax:
> I was wondering if we should have an endpoint where to pull versions that are compatible with the server. That way we could just add new incompatibles versions in the future and let the CLI check for compatibility.

I agree that we should have something like this. But it's out of scope for this ticket. We need a more generic compatibility story to address what you suggested. The purpose of this PR is just to incrementally improve the current state.

ksqldb-common/src/main/java/io/confluent/ksql/util/KsqlVersion.java
```diff
@@ -0,0 +1,137 @@
...
+  }
+
+  public boolean isAtLeast(final KsqlVersion version) {
+    if (versionType == version.versionType) {
+      return isAtLeastVersion(this, version.majorVersion, version.minorVersion);
+    }
+
+    if (versionType == VersionType.KSQLDB_STANDALONE) {
+      final KsqlVersion otherStandalone = versionMapping.get(version);
+      return otherStandalone != null
```
vcrfxia, line 73: Looks like this method returns false if the two versions are not comparable (because one is a standalone version and one is a CP version). Can we add this as a javadoc so future readers aren't left wondering?

ksqldb-common/src/main/java/io/confluent/ksql/util/KsqlVersion.java
```diff
@@ -0,0 +1,137 @@
...
+              otherStandalone.minorVersion
+          );
+    }
+
+    final KsqlVersion standalone = versionMapping.get(this);
+    return standalone != null
+        && isAtLeastVersion(standalone, version.majorVersion, version.minorVersion);
+  }
+
+  private boolean isAtLeastVersion(final KsqlVersion version, final int major, final int minor) {
```
vcrfxia, line 86: nit: static

ksqldb-common/src/main/java/io/confluent/ksql/util/KsqlVersion.java
```diff
@@ -0,0 +1,137 @@
...
+      return true;
+    } else if (version.majorVersion < major) {
+      return false;
+    } else {
+      return version.minorVersion >= minor;
+    }
+  }
+
+  public boolean same(final KsqlVersion version) {
+    if (versionType == version.versionType) {
```
vcrfxia, line 97: Is it simpler to replace this with `return isAtLeastVersion(version) && version.istALeastVersion(this);`?

ksqldb-common/src/test/java/io/confluent/ksql/util/KsqlVersionTest.java
```diff
@@ -0,0 +1,100 @@
...
+
+  @Test
+  public void shouldCompareCpVersionToStandaloneVersionInSame() {
+    // known mappings
+    assertThat(new KsqlVersion("6.0.").same(new KsqlVersion("0.10.")), is(true));
+    assertThat(new KsqlVersion("0.10.").same(new KsqlVersion("6.0.")), is(true));
+
+    assertThat(new KsqlVersion("6.1.").same(new KsqlVersion("0.14.")), is(true));
+    assertThat(new KsqlVersion("0.14.").same(new KsqlVersion("6.1.")), is(true));
+
```
vcrfxia, line 45: Add tests for known mapping that compare to false?

Same for the `isAtLeast()` tests below.

ksqldb-rest-client/src/main/java/io/confluent/ksql/rest/client/KsqlUnsupportedServerException.java
```diff
@@ -0,0 +1,29 @@
...
+ */
+
+package io.confluent.ksql.rest.client;
+
+public class KsqlUnsupportedServerException extends RuntimeException {
+
+  public KsqlUnsupportedServerException(final String cliVersion, final String serverVersion) {
+    super(String.format(
+        "Unsupported server version %s detected."
+        + " CLI version %s requires a minimum server version 5.4.x.",
```
vcrfxia, line 23: We should not hard-code `5.4.x` here.

ksqldb-cli/src/test/java/io/confluent/ksql/cli/CliTest.java
```diff
@@ -906,6 +909,79 @@ public void shouldPrintErrorOnUnsupportedAPI() throws Exception {
...
+        containsString("WARNING: CLI and server version don't match."
+            + " This may lead to unexpected errors.")
+    );
+  }
+
+  @Test
+  public void shouldPrintWarningOnDifferentStandaloneServerVersion() throws Exception {
+    givenRunInteractivelyWillExit();
+
+    final KsqlRestClient mockRestClient = givenMockRestClient("1.0.0-0");
```
vcrfxia, line 958: nit: can we use an old version rather than a new one, so this test doesn't unexpectedly start failing once we reach the future version?

vcrfxia (approved):
Thanks @mjsax -- LGTM overall with a few minor things inline. Could you update the PR description with an example warning as well, just as a sanity check?

ksqldb-common/src/main/java/io/confluent/ksql/util/KsqlVersion.java
mjsax: Good catch!

ksqldb-common/src/main/java/io/confluent/ksql/util/KsqlVersion.java
mjsax: Works for me.

vcrfxia (approved):
LGTM -- thanks @mjsax !

PR #7099, opened 2021-02-25
feat(migrations): apply command

jzaralim:
### Description 
Migrations tool `apply` command. Follows the algorithm outlined [here](https://github.com/jzaralim/ksql/pull/new/metadata-util). Also includes helper functions for writing metadata, loading all migration files and extracting the name of the migration from the file name, and introduces a new class, `Migration`

### Testing done 
unit and integration tests

### Reviewer checklist

### DIFF
--- a/ksqldb-tools/src/main/java/io/confluent/ksql/tools/migrations/commands/ApplyMigrationCommand.java
+++ b/ksqldb-tools/src/main/java/io/confluent/ksql/tools/migrations/commands/ApplyMigrationCommand.java
@@ -261,6 +261,7 @@ public class ApplyMigrationCommand extends BaseCommand {
     final String executionEnd = (state == MigrationState.MIGRATED || state == MigrationState.ERROR)
         ? Long.toString(clock.millis())
         : "";
+    final String checksum = MigrationsDirectoryUtil.computeHashForFile(migration.getFilepath());
     try {
       MetadataUtil.writeRow(
           config,
@@ -270,7 +271,8 @@ public class ApplyMigrationCommand extends BaseCommand {
           executionStart,
           executionEnd,
           migration,
-          previous
+          previous,
+          checksum
       ).get();
       MetadataUtil.writeRow(
           config,
@@ -280,7 +282,8 @@ public class ApplyMigrationCommand extends BaseCommand {
           executionStart,
           executionEnd,
           migration,
-          previous
+          previous,
+          checksum
       ).get();
       return true;
     } catch (InterruptedException | ExecutionException e) {
--- a/ksqldb-tools/src/main/java/io/confluent/ksql/tools/migrations/commands/ApplyMigrationCommand.java
+++ b/ksqldb-tools/src/main/java/io/confluent/ksql/tools/migrations/commands/ApplyMigrationCommand.java
@@ -15,6 +15,7 @@
 
 package io.confluent.ksql.tools.migrations.commands;
 
+import static io.confluent.ksql.tools.migrations.util.MigrationsDirectoryUtil.getAllMigrations;
 import static io.confluent.ksql.tools.migrations.util.MigrationsDirectoryUtil.getMigrationsDirFromConfigFile;
 
 import com.github.rvesse.airline.annotations.Command;
@@ -28,15 +29,14 @@ import io.confluent.ksql.tools.migrations.MigrationException;
 import io.confluent.ksql.tools.migrations.util.MetadataUtil;
 import io.confluent.ksql.tools.migrations.util.MetadataUtil.MigrationState;
 import io.confluent.ksql.tools.migrations.util.MetadataUtil.VersionInfo;
-import io.confluent.ksql.tools.migrations.util.MigrationsDirectoryUtil;
 import io.confluent.ksql.tools.migrations.util.MigrationsUtil;
 import io.confluent.ksql.util.KsqlException;
-import java.util.ArrayList;
+import java.time.Clock;
 import java.util.Arrays;
 import java.util.List;
-import java.util.Optional;
 import java.util.concurrent.ExecutionException;
 import java.util.function.Function;
+import java.util.stream.Collectors;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
@@ -92,7 +92,8 @@ public class ApplyMigrationCommand extends BaseCommand {
     return command(
         config,
         MigrationsUtil::getKsqlClient,
-        getMigrationsDirFromConfigFile(configFile)
+        getMigrationsDirFromConfigFile(configFile),
+        Clock.systemDefaultZone()
     );
   }
 
@@ -100,7 +101,8 @@ public class ApplyMigrationCommand extends BaseCommand {
   int command(
       final MigrationConfig config,
       final Function<MigrationConfig, Client> clientSupplier,
-      final String migrationsDir
+      final String migrationsDir,
+      final Clock clock
   ) {
     final Client ksqlClient;
     try {
@@ -110,77 +112,50 @@ public class ApplyMigrationCommand extends BaseCommand {
       return 1;
     }
 
-    if (!ValidateMigrationsCommand.validate(config, migrationsDir, ksqlClient)) {
+    if (ValidateMigrationsCommand.validate(config, migrationsDir, ksqlClient)
+        && apply(config, ksqlClient, migrationsDir, clock)) {
+      ksqlClient.close();
+      return 0;
+    } else {
+      ksqlClient.close();
       return 1;
     }
+  }
 
-    final int startVersion = getStartVersion(config, ksqlClient);
-    final int endVersion;
-    try {
-      endVersion = getEndVersion(startVersion, migrationsDir);
-    } catch (MigrationException e) {
-      LOGGER.error(e.getMessage());
-      return 1;
-    }
-
+  private boolean apply(
+      final MigrationConfig config,
+      final Client ksqlClient,
+      final String migrationsDir,
+      final Clock clock
+  ) {
+    final int start = getStartVersion(config, ksqlClient);
+    final List<Migration> migrations;
     LOGGER.info("Loading migration files");
 
-    final List<Migration> migrations;
-    try {
-      migrations = loadMigrations(startVersion, endVersion, migrationsDir);
-    } catch (MigrationException e) {
-      LOGGER.error(e.getMessage());
-      return 1;
+    if (next) {
+      migrations = getAllMigrations(start, start, migrationsDir);
+    } else if (version >= start) {
+      migrations = getAllMigrations(start, version, migrationsDir);;
+    } else if (version > 0 && version < start) {
+      LOGGER.error("Specified version is lower than latest migrated version");
+      return false;
+    } else {
+      migrations = getAllMigrations(start, migrationsDir);
     }
 
     for (Migration migration : migrations) {
-      if (!applyMigration(config, ksqlClient, migration)) {
-        return 1;
+      if (!applyMigration(config, ksqlClient, migration, clock)) {
+        return false;
       }
     }
-    return 0;
-  }
-
-  private List<Migration> loadMigrations(
-      final int start,
-      final int end,
-      final String migrationsDir
-  ) {
-    final List<Migration> migrations = new ArrayList<>();
-    for (int version = start; version <= end; version++) {
-      migrations.add(loadMigration(version, migrationsDir));
-    }
-    return migrations;
-  }
-
-  private Migration loadMigration(
-      final int version,
-      final String migrationsDir
-  ) {
-    final String versionString = Integer.toString(version);
-
-    final Optional<String> migrationFilePath =
-        MigrationsDirectoryUtil.getFilePathForVersion(versionString, migrationsDir);
-    if (!migrationFilePath.isPresent()) {
-      throw new MigrationException("Failed to find file for version " + versionString);
-    }
-
-    final String migrationCommand =
-        MigrationsDirectoryUtil.getFileContentsForVersion(versionString, migrationsDir);
-
-    LOGGER.info(migrationFilePath.get() + " loaded");
-    return new Migration(
-        version,
-        MigrationsDirectoryUtil.getNameFromMigrationFilePath(migrationFilePath.get()),
-        MigrationsDirectoryUtil.computeHashForFile(migrationFilePath.get()),
-        migrationCommand
-    );
+    return true;
   }
 
   private boolean applyMigration(
       final MigrationConfig config,
       final Client ksqlClient,
-      final Migration migration
+      final Migration migration,
+      final Clock clock
   ) {
     LOGGER.info("Applying " + migration.getName() + " version " + migration.getVersion());
     LOGGER.info(migration.getCommand());
@@ -190,27 +165,31 @@ public class ApplyMigrationCommand extends BaseCommand {
     }
 
     if (!verifyMigrated(config, ksqlClient, migration.getPrevious(), MAX_RETRIES)) {
-      LOGGER.error("Failed to verify status of version" + migration.getPrevious());
+      LOGGER.error("Failed to verify status of version " + migration.getPrevious());
       return false;
     }
 
-    final String executionStart = Long.toString(System.nanoTime() / 1000000);
+    final String executionStart = Long.toString(clock.millis());
 
-    if (!updateState(config, ksqlClient, MigrationState.RUNNING, executionStart, migration)) {
+    if (
+        !updateState(config, ksqlClient, MigrationState.RUNNING, executionStart, migration, clock)
+    ) {
       return false;
     }
 
     try {
-      final List<String> commands = Arrays.asList(migration.getCommand().split(";"));
+      final List<String> commands = Arrays.stream(migration.getCommand().split(";"))
+          .filter(s -> s.length() > 1)
+          .collect(Collectors.toList());
       for (final String command : commands) {
         ksqlClient.executeStatement(command + ";").get();
       }
     } catch (InterruptedException | ExecutionException e) {
       LOGGER.error(e.getMessage());
-      updateState(config, ksqlClient, MigrationState.ERROR, executionStart, migration);
+      updateState(config, ksqlClient, MigrationState.ERROR, executionStart, migration, clock);
       return false;
     }
-    updateState(config, ksqlClient, MigrationState.MIGRATED, executionStart, migration);
+    updateState(config, ksqlClient, MigrationState.MIGRATED, executionStart, migration, clock);
     LOGGER.info("Successfully migrated");
     return true;
   }
@@ -221,26 +200,6 @@ public class ApplyMigrationCommand extends BaseCommand {
       return 1;
     } else {
       return Integer.parseInt(latestMigratedVersion) + 1;
-    }
-  }
-
-  private int getEndVersion(final int startVersion, final String migrationsDir) {
-    if (next) {
-      return startVersion;
-    } else if (version >= startVersion) {
-      return version;
-    } else if (version > 0 && version < startVersion) {
-      throw new MigrationException("Specified version is lower than latest migrated version");
-    } else {
-      int endVersion = startVersion;
-      while (
-          MigrationsDirectoryUtil
-              .getFilePathForVersion(Integer.toString(endVersion), migrationsDir)
-              .isPresent()
-      ) {
-        endVersion++;
-      }
-      return endVersion - 1;
     }
   }
 
@@ -259,7 +218,7 @@ public class ApplyMigrationCommand extends BaseCommand {
         return true;
       } else {
         LOGGER.info(String.format(
-            "Could not verify status of version %s. Retrying (%i/%i)", version, i, retries));
+            "Could not verify status of version %s. Retrying (%d/%d)", version, i, retries));
       }
     }
     return false;
@@ -270,10 +229,11 @@ public class ApplyMigrationCommand extends BaseCommand {
       final Client ksqlClient,
       final MigrationState state,
       final String executionStart,
-      final Migration migration
+      final Migration migration,
+      final Clock clock
   ) {
     final String executionEnd = (state == MigrationState.MIGRATED || state == MigrationState.ERROR)
-        ? Long.toString(System.nanoTime() / 1000000)
+        ? Long.toString(clock.millis())
         : "";
     try {
       MetadataUtil.writeRow(
--- a/ksqldb-tools/src/main/java/io/confluent/ksql/tools/migrations/commands/ApplyMigrationCommand.java
+++ b/ksqldb-tools/src/main/java/io/confluent/ksql/tools/migrations/commands/ApplyMigrationCommand.java
@@ -15,9 +15,28 @@
 
 package io.confluent.ksql.tools.migrations.commands;
 
+import static io.confluent.ksql.tools.migrations.util.MigrationsDirectoryUtil.getMigrationsDirFromConfigFile;
+
 import com.github.rvesse.airline.annotations.Command;
 import com.github.rvesse.airline.annotations.Option;
 import com.github.rvesse.airline.annotations.restrictions.MutuallyExclusiveWith;
+import com.google.common.annotations.VisibleForTesting;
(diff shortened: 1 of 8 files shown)


ksqldb-tools/src/main/java/io/confluent/ksql/tools/migrations/commands/ApplyMigrationCommand.java
```diff
@@ -56,7 +77,188 @@
...
+      final Client ksqlClient,
+      final String migrationsDir,
+      final Clock clock
+  ) {
+    final int start = getStartVersion(config, ksqlClient);
+    final List<Migration> migrations;
+    LOGGER.info("Loading migration files");
+
+    if (next) {
+      migrations = getAllMigrations(start, start, migrationsDir);
```
vcrfxia, line 136: As above, this logic doesn't quite work since the next version does not have to be consecutive. The "next" version can use the same logic as "all" to find future migration files (from the latest applied version) but should only apply the first version found. 

We might also need to check whether the version(s) to be applied have already been applied or not. It's possible that we went "backwards" in version due to an out of order migration. We should check with @colinhicks what the expected user experience is here. Specifically, suppose a user applies migrations with versions 1, 2, and 6, then applies version 3 out of order. If the user then tries to apply the "next" migration, do we apply the migration with version 4 or 7?

ksqldb-tools/src/main/java/io/confluent/ksql/tools/migrations/commands/ApplyMigrationCommand.java
```diff
@@ -56,7 +77,188 @@
...
+    ) {
+      return false;
+    }
+
+    try {
+      final List<String> commands = Arrays.stream(migration.getCommand().split(";"))
+          .filter(s -> s.length() > 1)
+          .collect(Collectors.toList());
+      for (final String command : commands) {
+        ksqlClient.executeStatement(command + ";").get();
```
vcrfxia, line 185: This doesn't work for INSERT VALUES statements. We can handle this case in a follow-up PR, but INSERT VALUES statements need to be submitted via the `insertInto()` method rather than `executeStatement()`.

ksqldb-tools/src/main/java/io/confluent/ksql/tools/migrations/commands/ApplyMigrationCommand.java
```diff
@@ -56,7 +77,188 @@
...
+      final MigrationConfig config,
+      final Client ksqlClient,
+      final String migrationsDir,
+      final Clock clock
+  ) {
+    final int start = getStartVersion(config, ksqlClient);
+    final List<Migration> migrations;
+    LOGGER.info("Loading migration files");
+
+    if (next) {
```
vcrfxia, line 135: The logic in this if-else stack is great for the non-out-of-order case. In a follow-up PR, we should add a flag to the "apply" command to allow for out-of-order migrations. If this turns out to be too much effort, we can defer this to v2 of the migrations tool instead.

ksqldb-tools/src/main/java/io/confluent/ksql/tools/migrations/commands/ApplyMigrationCommand.java
```diff
@@ -56,7 +77,188 @@
...
+  private boolean verifyMigrated(
+      final MigrationConfig config,
+      final Client ksqlClient,
+      final String version,
+      final int retries
+  ) {
+    if (version.equals(MetadataUtil.NONE_VERSION)) {
+      return true;
+    }
+    for (int i = 1; i <= retries; i++) {
```
vcrfxia, line 215: We need to have a delay between retries here. See the methods in RetryUtil.

ksqldb-tools/src/main/java/io/confluent/ksql/tools/migrations/util/MetadataUtil.java
```diff
@@ -122,9 +158,9 @@ public static VersionInfo getInfoForVersion(
             "Failed to query state for migration with version " + version
                 + ": no such migration is present in the migrations metadata table");
       }
-      expectedHash = resultRows.get(0).getString(0);
-      prevVersion = resultRows.get(0).getString(1);
-      state = resultRows.get(0).getString(2);
+      expectedHash = resultRows.get(0).getString(1);
```
vcrfxia, line 161: Thanks for the fix!

ksqldb-tools/src/main/java/io/confluent/ksql/tools/migrations/commands/ApplyMigrationCommand.java
```diff
@@ -56,7 +77,188 @@
...
+      final Clock clock
+  ) {
+    final int start = getStartVersion(config, ksqlClient);
+    final List<Migration> migrations;
+    LOGGER.info("Loading migration files");
+
+    if (next) {
+      migrations = getAllMigrations(start, start, migrationsDir);
+    } else if (version >= start) {
+      migrations = getAllMigrations(start, version, migrationsDir);;
```
vcrfxia, line 138: nit: extra semicolon.

ksqldb-tools/src/main/java/io/confluent/ksql/tools/migrations/commands/ApplyMigrationCommand.java
```diff
@@ -56,7 +77,188 @@
...
+    } else if (version > 0 && version < start) {
+      LOGGER.error("Specified version is lower than latest migrated version");
+      return false;
+    } else {
+      migrations = getAllMigrations(start, migrationsDir);
```
vcrfxia, line 143: @colinhicks would you expect the default "apply" mode (out of "next", "all", and "until <version>") to be "all", "next", or should we fail unless one of the three is explicitly specified?

ksqldb-tools/src/test/java/io/confluent/ksql/tools/migrations/MigrationsTest.java
```diff
@@ -88,8 +96,44 @@ public static void classTearDown() {
...
+    final int status = MIGRATIONS_CLI.parse("--config-file", configFilePath, "apply").run();
+    assertThat(status, is(0));
+
+    // verify FOO and BAR were registered
+    describeSource("FOO");
+    describeSource("BAR");
+
+    // verify current
+    final List<String> current = makeKsqlQuery("SELECT * FROM migration_schema_versions WHERE VERSION_KEY='CURRENT';");
+    assertThat(current.size(), is(2));
```
vcrfxia, line 122: Why do we get two result rows from this query? It's a pull query, right? Shouldn't that only return the latest row? I feel like there's something I don't understand about pull queries.

ksqldb-tools/src/test/java/io/confluent/ksql/tools/migrations/MigrationsTest.java
```diff
@@ -88,8 +96,44 @@ public static void classTearDown() {
...
+    assertThat(currentRow.get(7), is("1"));
+    // verify version 1
+    final List<String> version1 = makeKsqlQuery("SELECT * FROM migration_schema_versions WHERE VERSION_KEY='1';");
+    assertThat(current.size(), is(2));
+    final JSONArray version1Row = new JSONObject(version1.get(1)).getJSONObject("row").getJSONArray("columns");
+    assertThat(version1Row.get(1), is("1"));
+    assertThat(version1Row.get(2), is("foo"));
+    assertThat(version1Row.get(3), is("MIGRATED"));
+    assertThat(version1Row.get(7), is("<none>"));
+
```
vcrfxia, line 136: Let's also verify version 2 for completeness.

ksqldb-tools/src/test/java/io/confluent/ksql/tools/migrations/MigrationsTest.java
```diff
@@ -176,6 +220,10 @@ private static SourceDescription describeSource(final String name) {
     return RestIntegrationTestUtil.makeKsqlRequest(REST_APP, sql);
   }
 
+  private static List<String> makeKsqlQuery(final String sql) {
+    return RestIntegrationTestUtil.makeWsRequest(REST_APP.getWsListener(), sql, Optional.empty(), Optional.empty(), Optional.empty());
```
vcrfxia, line 224: How come we're using the websocket endpoint? The regular query endpoint (see `RestIntegrationTestUtil #makeQueryRequest()`) is preferable, though in an ideal world we'd refactor RestIntegrationTestUtil to use the new HTTP/2 endpoints instead.

ksqldb-tools/src/test/java/io/confluent/ksql/tools/migrations/MigrationsTest.java
```diff
@@ -88,8 +96,44 @@ public static void classTearDown() {
...
+    );
+    final int status = MIGRATIONS_CLI.parse("--config-file", configFilePath, "apply").run();
+    assertThat(status, is(0));
+
+    // verify FOO and BAR were registered
+    describeSource("FOO");
+    describeSource("BAR");
+
+    // verify current
+    final List<String> current = makeKsqlQuery("SELECT * FROM migration_schema_versions WHERE VERSION_KEY='CURRENT';");
```
vcrfxia, line 121: Feel free to defer to a follow-up PR, but we should probably switch these assertions (probably also the `describeSource()` assertions above and in the `initialize` tests) to `assertThatEventually()` to account for the async nature of these commands (especially for updating the metadata table's state store).

ksqldb-tools/src/main/java/io/confluent/ksql/tools/migrations/commands/ApplyMigrationCommand.java
vcrfxia: Yeah on second thought, let's defer applying out of order migrations to v2.

ksqldb-tools/src/main/java/io/confluent/ksql/tools/migrations/commands/ApplyMigrationCommand.java
```diff
@@ -56,7 +77,188 @@
...
+    }
```
vcrfxia, line 144: I think we should add another variant of the "apply" command that just applies a single, specific version. (We can use the "-v" flag.) Let's do this in a follow-up PR (I can take it up if you're tired of working on "apply" haha).

ksqldb-tools/src/test/java/io/confluent/ksql/tools/migrations/commands/ApplyMigrationCommandTest.java
```diff
@@ -0,0 +1,340 @@
...
+        .thenReturn(infoQueryResult);
+    when(insertResult.get()).thenReturn(null);
+    when(statementResult.get()).thenReturn(null);
+
+    migrationsDir = folder.getRoot().getPath();
+  }
+
+  @Test
+  public void shouldApplyFirstMigration()
+      throws IOException, ExecutionException, InterruptedException {
```
vcrfxia, line 110: nit: let's just use `throws Exception` so we don't have to keep these lists up to date.

ksqldb-tools/src/test/java/io/confluent/ksql/tools/migrations/commands/ApplyMigrationCommandTest.java
```diff
@@ -0,0 +1,340 @@
...
+    when(infoQueryResult.get()).thenReturn(ImmutableList.of(createInfoRow(1, NAME, MigrationState.MIGRATED)));
+
+    // When:
+    final int result = command.command(config, cfg -> ksqlClient, migrationsDir, Clock.fixed(
+        Instant.ofEpochMilli(1000), ZoneId.systemDefault()));
+
+    // Then:
+    assertThat(result, is(1));
+    Mockito.verify(ksqlClient, Mockito.times(3)).executeQuery(any());
+  }
```
vcrfxia, line 233: Verify that `ksqlClient.executeStatement()` and `ksqlClient.insertInto()` were never called.

ksqldb-tools/src/test/java/io/confluent/ksql/tools/migrations/commands/ApplyMigrationCommandTest.java
```diff
@@ -0,0 +1,340 @@
...
+    // Then:
+    assertThat(result, is(0));
+    final InOrder inOrder = inOrder(ksqlClient);
+    verifyMigratedVersion(inOrder, 2, MigrationState.MIGRATED);
+    inOrder.verify(ksqlClient).close();
+    inOrder.verifyNoMoreInteractions();
+  }
+
+  @Test
+  public void shouldApplyMultipleMigrations()
```
vcrfxia, line 151: Also add a unit test for "until" (positive test case)?

ksqldb-tools/src/test/java/io/confluent/ksql/tools/migrations/commands/ApplyMigrationCommandTest.java
```diff
@@ -0,0 +1,340 @@
...
+    inOrder.verify(ksqlClient).insertInto(
+        MIGRATIONS_STREAM,
+        createKsqlObject(MetadataUtil.CURRENT_VERSION_KEY, version, NAME, finalState, "1000", "1000")
+    );
+    inOrder.verify(ksqlClient).insertInto(
+        MIGRATIONS_STREAM,
+        createKsqlObject(Integer.toString(version), version, NAME, finalState, "1000", "1000")
+    );
+  }
+}
```
vcrfxia, line 340: Nice test coverage! This is awesome :)

vcrfxia (commented):
Thanks @jzaralim this is really great! A few concerns and the usual boatload of nits. A lot of my comments are around supporting non-consecutive migration versions and (in the future) out-of-order migrations. Feel free to defer any of the comments related specifically to out-of-order migrations, though I do think we should support non-consecutive version numbers out of the box.

ksqldb-tools/src/test/java/io/confluent/ksql/tools/migrations/MigrationsTest.java
jzaralim: This queries the websocket endpoint, which returns the column names as the first item in the list.

ksqldb-tools/src/main/java/io/confluent/ksql/tools/migrations/commands/ApplyMigrationCommand.java
vcrfxia: It just occurred to me that translating `INSERT VALUES` statements into `insertInto()` calls is actually really annoying. Let me think some more about the best way to handle this. We might want to make another minor update to the client.

ksqldb-tools/src/main/java/io/confluent/ksql/tools/migrations/commands/ApplyMigrationCommand.java
jzaralim: The design doc suggests this option:

```
  --out-of-order     Allow out of order migrations. Only applicable for `apply`. This does not affect the behavior of `next` but will affect `all` and `until`.``
```

vcrfxia: I think it's still ambiguous. For example, if you apply 1 then 2 then 6 then 3, does "CURRENT" now point to 6 or 3? I think it should point to 3, and that the "previous" state linked to from 3 should be 6. If that's the case then "next" from the "CURRENT" of 3 is still 4.

ksqldb-tools/src/test/java/io/confluent/ksql/tools/migrations/MigrationsTest.java
jzaralim: `makeQueryRequest ` was a private method. It seems safe to change to public though, so that's updated in the latest commit

ksqldb-tools/src/main/java/io/confluent/ksql/tools/migrations/commands/ApplyMigrationCommand.java
colinhicks: I agree with @vcrfxia above.

> Specifically, suppose a user applies migrations with versions 1, 2, and 6, then applies version 3 out of order. If the user then tries to apply the "next" migration, do we apply the migration with version 4 or 7?

As a user, I would expect version 4.

> I think it's still ambiguous. For example, if you apply 1 then 2 then 6 then 3, does "CURRENT" now point to 6 or 3? I think it should point to 3, and that the "previous" state linked to from 3 should be 6. If that's the case then "next" from the "CURRENT" of 3 is still 4.

This also sounds good to me.

spena: @colinhicks Won't that behavior cause confusions with users? If I migrated to 6, and for some reason I applied an out-of-order migration 3, why would we continue from 4 after that? Btw, I don't understand out-of-order migrations yet, but if users use it, then those should be in special cases, i.e. to apply a migration that they forgot (oops!). I would expect the tool continues applying migrations from the highest version migrated, in this case next one should be 7. 

Btw, do we need to support out-of-order migrations?

colinhicks: @spena can you clarify the line here?

>   --out-of-order     Allow out of order migrations. Only applicable for `apply`. This does not affect the behavior of `next` but will affect `all` and `until`.``

I think I might be misinterpreting it. Are we actually saying that `--out-of-order` handles the case where the current migration is 6 and a new migration, 4, was added late, after we had applied 1, 2, 3, 5, and 6?

vcrfxia: > Are we actually saying that --out-of-order handles the case where the current migration is 6 and a new migration, 4, was added late, after we had applied 1, 2, 3, 5, and 6?

I added that line to the KLIP. My intent in saying "This does not affect the behavior of next" was that if the current migration is 6, then `next` will not apply 4 even if `--out-of-order` is set. Upon thinking about it more, I think we should either disallow setting `--out-of-order` for the `next` mode (to avoid confusion) or allow it with the semantics of finding the first, unapplied migration, which may be before the "current" version.

vcrfxia: > Btw, do we need to support out-of-order migrations?

We've already decided to not support this for the first version of the migrations tool, so I assume you're asking about the long-term plan instead? I thought it was a useful feature to have in the long run but maybe not if it's generating so much confusion. I'll let @colinhicks make the call here.

A related question is whether we want to allow non-consecutive migration versions (e.g., as a user can I apply version 1 and then version 3 without applying version 2). Doing so without supporting out of order migrations means I'll never be able to apply version 2 (on this cluster). This may still be valuable to users who, for example, start using the migrations tool from a non-clean slate and create migration 1 to capture the current state but cannot apply it (because those streams and tables already exist), and want to start from migration 2. Does this make sense, or would supporting non-consecutive migrations introduce more confusion than it's worth, if we don't also support out of order migrations (at least in v1 of the migrations tool)?

colinhicks: Thanks for the explanation.

> I think we should either disallow setting --out-of-order for the next mode (to avoid confusion) or allow it with the semantics of finding the first, unapplied migration

Based on the conversation above, my vote is to disallow setting it for `next` in order to avoid confusion :). How do others feel about this?

colinhicks: I do think we want to keep the option open to support other out-of-order migration scenarios in future versions. Ideally we can be informed by user feedback when considering cases.

ksqldb-tools/src/main/java/io/confluent/ksql/tools/migrations/commands/ApplyMigrationCommand.java
```diff
@@ -56,7 +77,201 @@
...
+      final Clock clock
+  ) {
+    String previous = MetadataUtil.getLatestMigratedVersion(config, ksqlClient);
+    final int minimumVersion = previous.equals(MetadataUtil.NONE_VERSION)
+        ? 1
+        : Integer.parseInt(previous) + 1;
+    LOGGER.info("Loading migration files");
+    final List<Migration> migrations = getAllMigrations(migrationsDir).stream()
+        .filter(migration -> {
+          if (version > 0) {
```
vcrfxia, line 138: Let's validate (at the beginning of this method) that version is non-negative.

ksqldb-tools/src/main/java/io/confluent/ksql/tools/migrations/commands/ApplyMigrationCommand.java
```diff
@@ -56,7 +77,201 @@
...
+            return migration.getVersion() <= version && migration.getVersion() >= minimumVersion;
+          } else {
+            return migration.getVersion() >= minimumVersion;
+          }
+        })
+        .collect(Collectors.toList());
+
+    if (migrations.size() == 0) {
+      LOGGER.error("No eligible migrations found.");
```
vcrfxia, line 147: I don't think we should return an error if there are no migrations and the mode is `until` or `all`. We should check that we have log messages that make it clear which file(s) are being applied, and simply print a message saying there's nothing to apply in these modes, but I think the exit code should still be 0 rather than 1.

ksqldb-tools/src/main/java/io/confluent/ksql/tools/migrations/util/MigrationsDirectoryUtil.java
```diff
@@ -99,4 +94,34 @@ public static String computeHashForFile(final String filename) {
...
   }
+
+  public static String getNameFromMigrationFilePath(final String filename) {
+    return filename
+        .substring(filename.indexOf("__") + 2, filename.indexOf(".sql"))
+        .replace('_', ' ');
+  }
+
+  public static int getVersionFromMigrationFilePath(final String filename) {
+    return Integer.parseInt(filename.substring(1, 7));
```
vcrfxia, line 105: We should validate the filename format, rather than letting this throw if this particular substring is not an integer. If we detect files that don't fit the expected format (`V<six digit number>__<name>.sql`), we should log a warning saying that those files are being ignored.

ksqldb-tools/src/main/java/io/confluent/ksql/tools/migrations/util/MigrationsDirectoryUtil.java
```diff
@@ -99,4 +94,34 @@ public static String computeHashForFile(final String filename) {
...
+    final String[] names = directory.list();
+    if (names == null) {
+      throw new MigrationException("Failed to retrieve files from " + migrationsDir);
+    }
+
+    return Arrays.stream(names)
+        .map(name ->new Migration(
+            getVersionFromMigrationFilePath(name),
+            getNameFromMigrationFilePath(name),
+            computeHashForFile(migrationsDir + "/" + name),
```
vcrfxia, line 123: Let's not load all the contents of every file and compute the hash for every file each time -- we should only do this for the files we actually need to do it for.

ksqldb-tools/src/main/java/io/confluent/ksql/tools/migrations/commands/ApplyMigrationCommand.java
```diff
@@ -56,7 +77,201 @@
...
+    }
+    try {
+      RetryUtil.retryWithBackoff(
+          retries,
+          1000,
+          1000,
+          () -> {
+            if (!MetadataUtil.getInfoForVersion(version, config, ksqlClient)
+                .getState().equals(MigrationState.MIGRATED)) {
+              throw new MigrationException("Could not verify status of version " + version);
```
vcrfxia, line 226: nit: make the error message more specific, something like `Expected status MIGRATED for version <version>. Got <other status>`?

vcrfxia (approved):
Thanks @jzaralim ! A few minor things inline. LGTM otherwise!

PR #7101, opened 2021-02-26
chore: remove references to lambdas from functions and return error message when user tries to write expression with lambda

stevenpyzhang:
### Description 
lambdas are not released in 0.16.0

Removed the invocation functions added for lambdas and also throw an error message when user tries to write a lambda expression.

Before removing the invocation function, tried to transform an array and saw the error message

```
select transform(array[1,2,3], x => x) from test emit changes;
Lambda expressions are not supported
```

### Testing done 

### Reviewer checklist


lct45 (approved):
LGTM, thanks @stevenpyzhang

vvcephei (approved):
Thanks, @stevenpyzhang !

guozhangwang (approved):

PR #7103, opened 2021-02-26
refactor: pull lambda type to sql lambda check out of UdfIndex and GenericUtils

stevenpyzhang:
### Description 
The previous implementation of lambda type checking was too reliant on the ordering of the previous invocation function arguments to work properly. This new approach gets around this limitation by assigning types for the SqlLambda directly from the LambdaType from the function.parameters() list. We use the existing generic resolving code in order to determine the correct types for the lambda arguments and pass that in through the context as a list (which we were doing previously. The core change is how this list gets generated)

This is done by doing two passes over the FunctionCall arguments, the first pass, any lambda functions are treated as SqlLambda which only has the num input types as information. During this first pass, we fetch the appropriate function from Udffactory (this can be done because of how Java handles function erasure). During the second pass, since we now have the function and can access the function argument ParamType, when we encounter a lambda function, we can look up the corresponding LambdaType from the function signature and construct the input type list.

I've also now updated the reduce UDF so that the type signature is
`reduce(array, state, (s, x))`
`reduce(map, state, (s, k, v))`

### Testing done 
Some unit tests (need a bit more)
QTT tests were updated with the new reduce 

### Reviewer checklist


ksqldb-execution/src/main/java/io/confluent/ksql/execution/util/ExpressionTypeManager.java
```diff
@@ -474,36 +484,90 @@ public Void visitFunctionCall(
...
+            final List<SqlType> sqlTypesForLambdaArgMapping =
+                new ArrayList<>(childContext.getLambdaInputTypes());
+            process(expression, childContext);
+
+            final SqlType resolvedArgType = childContext.getSqlType();
+            final SqlArgument lambdaArgument = SqlArgument.of(
+                SqlLambda.of(sqlTypesForLambdaArgMapping, resolvedArgType));
+
+            argForReturnType.add(lambdaArgument);
+            GenericsUtil.reserveGenerics(
```
lct45, line 551: What's the scenario where there's a generic inside of a lambda that hasn't already been mapped?

ksqldb-execution/src/main/java/io/confluent/ksql/execution/codegen/SqlToJavaVisitor.java
```diff
@@ -448,49 +452,108 @@ private String formatExpression(final Expression expression) {
...
+              }
+            }
+            final List<SqlType> sqlTypesForLambdaArgMapping =
+                new ArrayList<>(childContext.getLambdaInputTypes());
+            final SqlType resolvedArgType = expressionTypeManager.getExpressionSqlType(
+                expression,
+                childContext.getCopy()
+            );
+            final SqlArgument lambdaArgument = SqlArgument.of(
+                SqlLambda.of(sqlTypesForLambdaArgMapping, resolvedArgType));
```
lct45, line 525: Why did you do `sqlTypesForLambdaArgMapping` instead of doing `childContext.getLambdaInputTypes()`? Since we passs in a copy to `ExpressionTypeManager` won't the input type list be unchanged when we create the sqlLambda?

ksqldb-udf/src/main/java/io/confluent/ksql/schema/ksql/types/SqlLambda.java
```diff
@@ -41,12 +42,27 @@ public static SqlLambda of(
...
+
+  public SqlLambda(
+      final Integer numInputs
+  ) {
+    this.inputTypes = null;
+    this.returnType = null;
+    this.numInputs = numInputs;
   }
 
   public List<SqlType> getInputType() {
```
lct45, line 68: I feel like it's unlikely that we'd ever hit this when we have `null` for `inputType` but do we want to consider doing optionals like we did for `SqlArgument`?

ksqldb-common/src/main/java/io/confluent/ksql/function/GenericsUtil.java
```diff
@@ -291,4 +313,27 @@ public static boolean instanceOf(final ParamType schema, final SqlArgument insta
 
     return true;
   }
+
+  public static void reserveGenerics(
```
lct45, line 317: Did you not want to use `reserveGenerics` inside of `UdfIndex` because of the return type + access level (aka the other one is private)? JW because I was confused about `reserveGenerics` being new since it's a function I've seen before. If we leave it here can we come up with another name? IMO all the stuff shared between UDF/Generics is confusing enough w/o duplicate names

ksqldb-execution/src/main/java/io/confluent/ksql/execution/codegen/CodeGenRunner.java
```diff
@@ -184,45 +190,103 @@ public Void visitLikePredicate(final LikePredicate node, final TypeContext conte
...
+                parameter,
+                argTypesForFirstPass.get(i),
+                reservedGenerics
+            );
+          }
+        }
+        argumentSchemas.addAll(argForReturnType);
+      } else {
+        argumentSchemas.addAll(argTypesForFirstPass);
+        argumentSchemas.forEach(argument -> typeContextsForChildren.add(context.getCopy()));
```
lct45, line 283: It doesn't look like this is used later, and if we're checking if the node has a lambda we won't hit both line 280 and this line, right? So do we need to add all the `argTypesForFirstPass` to `argumentSchemas` or can we just do the `forEach` on `argTypesForFirstPass` directly?

ksqldb-execution/src/main/java/io/confluent/ksql/execution/codegen/SqlToJavaVisitor.java
```diff
@@ -448,49 +452,108 @@ private String formatExpression(final Expression expression) {
...
+                parameter,
+                argTypesForFirstPass.get(i),
+                reservedGenerics
+            );
+          }
+        }
+        functionReturnSchema = function.getReturnType(argForReturnType);
+        argumentSchemas.addAll(argForReturnType);
+      } else {
+        functionReturnSchema = function.getReturnType(argTypesForFirstPass);
```
lct45, line 549: If i'm reading everything correctly, getting the return type is the one line different between here and `CodeGenRunner` ): IT looks like if we did pull this out into something like `genericsUtil` it would require a huge argument list... but it would feel nice to reuse some of the code at least between this and `CodeGenRunner`. At the very least WDYT about pulling out the stuff inside of `if(node.hasLambdaFunctionCall())` into a helper function? Just thinking that if other people have to work on this and don't want to get bogged down by lambdas that could help readability

ksqldb-execution/src/main/java/io/confluent/ksql/execution/util/ExpressionTypeManager.java
```diff
@@ -474,36 +484,90 @@ public Void visitFunctionCall(
...
+            }
+
+            final LambdaType lambdaParameter = (LambdaType) parameter;
+            for (ParamType inputParam : lambdaParameter.inputTypes()) {
+              if (inputParam instanceof GenericType) {
+                final GenericType genericParam = (GenericType) inputParam;
+                if (!reservedGenerics.containsKey(genericParam)) {
+                  throw new KsqlException(
+                      String.format(
+                          "Could not verify type for generic %s.",
```
agavra, line 533: also makes sense here to output the reserved generics map and what we expected (but didn't find)

ksqldb-execution/src/main/java/io/confluent/ksql/execution/util/ExpressionTypeManager.java
```diff
@@ -474,36 +484,90 @@ public Void visitFunctionCall(
...
+
+            argForReturnType.add(lambdaArgument);
+            GenericsUtil.reserveGenerics(
+                parameter,
+                lambdaArgument,
+                reservedGenerics
+            );
+          } else {
+            argForReturnType.add(argTypesForFirstPass.get(i));
+            GenericsUtil.reserveGenerics(
```
agavra, line 558: I don't know if we support lambdas in the middle of function declarations, but I think we'll want to process all non-lambda arguments before we process lambda arguments instead of doing them in the order they show up - otherwise we may not have the info we need to resolve the lambdas:
```
my_filter(x -> true, list)
```

if we don't support this, then we should fail hard if we see any non-lambda parameter passed in after a lambda paramter (which we might already be doing, I haven't looked at tests here)

ksqldb-functional-tests/src/test/resources/historical_plans/reduce-map_-_reduce_map_with_null_initial_state/7.0.0_1614328638924/spec.json
```diff
@@ -0,0 +1,113 @@
+{
```
agavra, line 1: can we delete the old historical plans? no need to regularly test things that never shipped (for follow up PR)

ksqldb-udf/src/main/java/io/confluent/ksql/schema/ksql/types/SqlLambda.java
agavra: I think we should take it a step further, this class is coupling two different things: resolved lambdas and unresolved lambdas. Do we even need the former? If so, let's break it out into two separate classes - if code needs one it's unlikely it needs the other (and it can get the size from using `inputTypes.size()`)

agavra (commented):
Thanks @stevenpyzhang! Great job getting this out on such a quick turnaround time 🥳 we get the best of both worlds.

I quickly skimmed it (just looked at ExpressionTypeManager, a few of the util classes and the tests) so I'll let @lct45 review the rest but the approach LGTM (even cleaner than what I had in mind).

I think the real test would be to see if we can implement a reduce that also takes in the index (i.e. a lambda parameter that has a strongly typed, non-generic input). If we did this correctly, it should "just work"

ksqldb-execution/src/main/java/io/confluent/ksql/execution/util/ExpressionTypeManager.java
stevenpyzhang: This logic should be fine. The parser only recognizes lambda functions at the end of a function https://github.com/confluentinc/ksql/blob/master/ksqldb-parser/src/main/antlr4/io/confluent/ksql/parser/SqlBase.g4#L300

We may want to change this in the future though if we want to support UDF with functions anywhere in the signature but I think this is low priority and should be a lower priority. 

I filed an issue to keep track of this https://github.com/confluentinc/ksql/issues/7107

agavra: +1 from me - thanks for filing the ticket

ksqldb-execution/src/main/java/io/confluent/ksql/execution/util/ExpressionTypeManager.java
stevenpyzhang: We wouldn't support something like this
```
  public <T,S, U> S randomFunction(
      @UdfParameter(description = "some array.") final List<T> list,
      @UdfParameter(description = "some state.") final S initialState,
      @UdfParameter(description = "some function.") final TriFunction<S, U, T, S> biFunction
  ) {
```
The LambdaType for the last argument would have Generics [S, U, T] in the input list and we wouldn't be able to handle U

stevenpyzhang: `select transform(array[1], (x,y) => 5) from hello emit changes;` would error out here

lct45: Hmm yeah I see what you mean. That transform example isn't supported right now anyways, right? I guess it doesn't hurt to leave it in

ksqldb-udf/src/main/java/io/confluent/ksql/schema/ksql/types/SqlLambda.java
lct45: I do think we need the first to properly resolve generics when we get the return type, but you're right that we don't need both

ksqldb-execution/src/main/java/io/confluent/ksql/execution/util/ExpressionTypeManager.java
stevenpyzhang: It's just a temporary example, and I don't think we need to support having unresolved generics in the lambda function in the near future.

ksqldb-execution/src/main/java/io/confluent/ksql/execution/util/ExpressionTypeManager.java
```diff
@@ -474,36 +482,98 @@ public Void visitFunctionCall(
...
-      final boolean hasLambda = node.hasLambdaFunctionCallArguments();
-      for (final Expression expression : node.getArguments()) {
-        final TypeContext childContext = TypeContextUtil.contextForExpression(
-            expression, expressionTypeContext, currentTypeContext
-        );
-        process(expression, childContext);
-        final SqlType resolvedArgType = childContext.getSqlType();
+      final List<Expression> arguments = node.getArguments();
 
+      for (final Expression expression : arguments) {
```
rodesai, line 489: this loop can probably be it's own method

ksqldb-execution/src/main/java/io/confluent/ksql/execution/util/ExpressionTypeManager.java
```diff
@@ -474,36 +482,98 @@ public Void visitFunctionCall(
...
+            // at this index in the function arguments if there's a
+            // lambda node at this index in the function node argument list
+            if (!(parameter instanceof LambdaType)) {
+              throw new KsqlException("Error while processing lambda function. "
+                  + "This is most likely an internal error and a "
+                  + "Github issue should be filed for debugging.");
+            }
+
+            final LambdaType lambdaParameter = (LambdaType) parameter;
+            for (ParamType inputParam : lambdaParameter.inputTypes()) {
```
rodesai, line 525: maybe move to another method like `reserveGenericsForLambdaReturnType`

rodesai (commented):
Thanks @stevenpyzhang! took a first pass through the type resolution code and left a few questions

ksqldb-execution/src/main/java/io/confluent/ksql/execution/codegen/SqlToJavaVisitor.java
stevenpyzhang: This was from an earlier iteration where the input type list was getting wiped while handling the child, so I added this in. Removed now.

ksqldb-udf/src/main/java/io/confluent/ksql/schema/ksql/types/SqlLambda.java
stevenpyzhang: I broke it out into SqlLambda (which only has information on the num input types) and a child class SqlLambdaResolved (which actually has the input types and return type values)

ksqldb-common/src/main/java/io/confluent/ksql/function/GenericsUtil.java
```diff
@@ -213,23 +221,29 @@ private static boolean resolveGenerics(
...
+          return resolveGenerics(
+              mapping,
+              lambdaType.returnType(),
+              SqlArgument.of(sqlLambda.getReturnType()),
+              resolveLambdaType
+          );
+        } else {
+          return true;
+        }
+      } else {
```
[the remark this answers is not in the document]
stevenpyzhang: You're right, I changed this back to returning false.

ksqldb-execution/src/main/java/io/confluent/ksql/execution/util/ExpressionTypeManager.java
```diff
@@ -474,36 +482,98 @@ public Void visitFunctionCall(
...
+              if (inputParam instanceof GenericType) {
+                final GenericType genericParam = (GenericType) inputParam;
+                if (!reservedGenerics.containsKey(genericParam)) {
```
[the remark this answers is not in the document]
stevenpyzhang: This is making the assumption that lambda functions are only found at the end of a function argument list. We'll have already processed all the generics from non-lambda function arguments and if the lambda function signature has a generic we haven't encountered, then we throw the exception. This does mean we're not able to have lambdas anywhere in the UDF argument list but that's a limitation of the parser also and we can address both later.

https://github.com/confluentinc/ksql/issues/7107

ksqldb-udf/src/main/java/io/confluent/ksql/schema/ksql/types/SqlLambda.java
```diff
@@ -15,47 +15,36 @@
...
+public class SqlLambda {
 
-  private final List<SqlType> inputTypes;
-  private final SqlType returnType;
+  private final int numInputs;
 
   public static SqlLambda of(
-      final List<SqlType> inputType,
-      final SqlType returnType
+      final Integer numInputs
```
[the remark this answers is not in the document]
stevenpyzhang: Java doesn't recognize that the parameters of the bifunction are different, if we were to write the UDF `foo` and try to have those two variants of the function, Java would complain since they have the same erasure. That's why we're able to just check the num inputs when getting the correct function from UDF factory

ksqldb-execution/src/main/java/io/confluent/ksql/execution/util/ExpressionTypeManager.java
```diff
@@ -474,36 +482,98 @@ public Void visitFunctionCall(
...
+          final Expression expression = arguments.get(i);
+          final ParamType parameter = paramTypes.get(i);
+          if (expression instanceof LambdaFunctionCall) {
+            final TypeContext childContext = expressionTypeContext.getCopy();
+
+            // the function returned from the UDF factory should have lambda
+            // at this index in the function arguments if there's a
+            // lambda node at this index in the function node argument list
+            if (!(parameter instanceof LambdaType)) {
+              throw new KsqlException("Error while processing lambda function. "
```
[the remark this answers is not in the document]
stevenpyzhang: changed it to run time exception

ksqldb-common/src/main/java/io/confluent/ksql/function/types/ParamTypes.java
```diff
@@ -57,26 +58,32 @@ public static boolean areCompatible(
...
-        return false;
-      }
-      int i = 0;
-      for (final ParamType paramType: declaredLambda.inputTypes()) {
-        if (!areCompatible(
-            SqlArgument.of(sqlLambda.getInputType().get(i)),
-            paramType,
-            allowCast)
-        ) {
+      final SqlLambda sqlLambda = sqlLambdaOptional.get();
```
guozhangwang, line 62: For my understanding: the caller path from `UdfIndex#getCandidates` should always see unresolved lambda, hence only comparing num.params, whereas the other caller `FunctionLoaderUtils#handleUdfReturnSchema` should always see resolved lambda and hence fall into the first condition, right?

ksqldb-common/src/main/java/io/confluent/ksql/function/GenericsUtil.java
```diff
@@ -213,23 +219,32 @@ private static boolean resolveGenerics(
...
-            "Number of lambda arguments doesn't match between schema and sql type");
-      }
-
-      int i = 0;
-      for (final ParamType paramType : lambdaType.inputTypes()) {
-        if (!resolveGenerics(
-            mapping, paramType, SqlArgument.of(sqlLambda.getInputType().get(i))
-        )) {
-          return false;
+      if (lambdaType.inputTypes().size() == sqlLambda.getNumInputs()) {
```
guozhangwang, line 222: This piece of logic seems to be the same as in `ParamType#areCompatible` for lambdas, could we extract them into a shared util?

ksqldb-common/src/main/java/io/confluent/ksql/function/GenericsUtil.java
guozhangwang: +1 !

ksqldb-common/src/main/java/io/confluent/ksql/function/GenericsUtil.java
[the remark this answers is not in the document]
guozhangwang: For my own education, it seems that if users indeed register two function as above, then 1) `UdfIndex#getCandidates` would return more than one candidate, and 2) we have to rely on `UdfIndex#Node#compare` to pick the right one, which would be relying on `order` since `int compare = compareFunctions.compare(value, other.value)` would return zero.

Does that mean, that users declaring these functions under the same class should always declare the first and then the second so that `transform(list, (x, i) -> 'foo')` has a higher order? I.e. do we enforce a declaration ordering for cases like above?

ksqldb-execution/src/main/java/io/confluent/ksql/execution/codegen/CodeGenRunner.java
```diff
@@ -184,45 +190,107 @@ public Void visitLikePredicate(final LikePredicate node, final TypeContext conte
...
-        // expression subtree doesn't interfere with type checking in another one
-        final SqlType resolvedArgType =
-            expressionTypeManager.getExpressionSqlType(argExpr, childContext.getCopy());
-
-        if (argExpr instanceof LambdaFunctionCall) {
-          argumentTypes.add(
+      final List<SqlArgument> argTypesForFirstPass = new ArrayList<>();
+      final List<Expression> arguments = node.getArguments();
+
+      for (final Expression expression : arguments) {
```
guozhangwang, line 198: nit: it would be better to add a high-level comment describing the two-pass algorithm here, e.g. why and how we do it.

ksqldb-execution/src/main/java/io/confluent/ksql/execution/codegen/CodeGenRunner.java
```diff
@@ -184,45 +190,107 @@ public Void visitLikePredicate(final LikePredicate node, final TypeContext conte
...
+          );
+          argTypesForFirstPass.add(SqlArgument.of(resolvedArgType));
         }
       }
 
       final UdfFactory holder = functionRegistry.getUdfFactory(functionName);
-      final KsqlScalarFunction function = holder.getFunction(argumentTypes);
+      final KsqlScalarFunction function = holder.getFunction(argTypesForFirstPass);
+
+      final List<ParamType> paramTypes = function.parameters();
```
guozhangwang, line 217: Should we check if `function.parameters().size() == arguments.size()` first?

ksqldb-execution/src/main/java/io/confluent/ksql/execution/codegen/SqlToJavaVisitor.java
```diff
@@ -448,49 +451,115 @@ private String formatExpression(final Expression expression) {
...
-
-        // pass a copy of the context to the type checker so that type checking in one
-        // expression subtree doesn't interfere with type checking in another one
-        final SqlType resolvedArgType =
-            expressionTypeManager.getExpressionSqlType(argExpr, childContext.getCopy());
-        if (argExpr instanceof LambdaFunctionCall) {
-          argumentSchemas.add(
+      final List<Expression> arguments = node.getArguments();
+
+      for (final Expression expression : arguments) {
```
guozhangwang, line 462: Not for this PR, but I'm feeling a bit harder for remembering the resolved types now since we are making two passes.. :)

ksqldb-execution/src/main/java/io/confluent/ksql/execution/util/ExpressionTypeManager.java
guozhangwang: In addition, though I understand we need to do similar two-pass loops on all three classes today, but could we consider at least extracting them into a single util function call?

guozhangwang (commented):
Made a pass, left mostly questions rather than comments :)

ksqldb-udf/src/main/java/io/confluent/ksql/schema/ksql/types/SqlLambda.java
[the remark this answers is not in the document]
rodesai: Ah right my bad I forgot it just gets turned into BiFunction at the end of the day.

ksqldb-common/src/main/java/io/confluent/ksql/function/GenericsUtil.java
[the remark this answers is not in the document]
stevenpyzhang: I don't think the ordering would matter, in Almog's example, they lambda has different input types so there shouldn't be confusion when verifying if the variation the user is trying to use matches one of the candidates, it should only be possible to return one of them

ksqldb-common/src/main/java/io/confluent/ksql/function/GenericsUtil.java
stevenpyzhang: I have UdfIndex using this one now, not sure if there's a better way other than having UdfIndex catch the exceptions and return false.

ksqldb-common/src/main/java/io/confluent/ksql/function/types/ParamTypes.java
stevenpyzhang: Yeah I think so, for completeness we have both cases here still.

ksqldb-functional-tests/src/test/resources/historical_plans/reduce-map_-_reduce_map_with_null_initial_state/7.0.0_1614328638924/spec.json
stevenpyzhang: deleted the old ones.

ksqldb-execution/src/main/java/io/confluent/ksql/execution/util/ExpressionTypeManager.java
```diff
@@ -474,36 +482,98 @@ public Void visitFunctionCall(
 
       final UdfFactory udfFactory = functionRegistry.getUdfFactory(node.getName());
 
-      // this context gets updated as we process non lambda arguments
-      final TypeContext currentTypeContext = expressionTypeContext.getCopy();
+      final List<SqlArgument> argTypesForFirstPass = new ArrayList<>();
```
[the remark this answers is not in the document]
stevenpyzhang: done

ksqldb-common/src/main/java/io/confluent/ksql/function/GenericsUtil.java
lct45: Hmm I guess if we really wanted to avoid checking errors in UDFIndex we could have expression type manager call a helper function inside of generics util and have _that_ function call `reserveGenerics` and if `reserveGenerics` returns false it could throw an error... Overall tho I think it'd be nice to not add more small functions that call other functions so I think I lean on the side of leaving it as is

lct45: Sorry, a few last things here. In this method we call the `resolveGenerics` with two arguments, which does pretty much the exact same thing as this method and throws the same errors. Can we skip the error handling here and leave it in `resolveGenerics`? It's hard for me to think this through logically without running through the code but is there a scenario where we don't catch a generics mismatch in `resolveGenerics` because we don't have the passed in map from `reserveGenerics` and therefore find the exception here? The original `reserveGenerics` did this second check which makes me _think_ it's necessary but I'm not sure.

ksqldb-common/src/main/java/io/confluent/ksql/function/GenericsUtil.java
```diff
@@ -291,4 +324,35 @@ public static boolean instanceOf(final ParamType schema, final SqlArgument insta
 
     return true;
   }
+
+  public static void reserveGenerics(
+      final ParamType schema,
+      final SqlArgument argument,
+      final Map<GenericType, SqlType> reservedGenerics
+  ) {
+    if (!GenericsUtil.instanceOf(schema, argument)) {
```
lct45, line 333: re: above: I think even if we want the later error handling in this method this scenario should be covered in the call to `resolveGenerics`

ksqldb-common/src/main/java/io/confluent/ksql/function/UdfIndex.java
```diff
@@ -366,33 +363,17 @@ boolean accepts(final SqlArgument argument, final Map<GenericType, SqlType> rese
...
-      final Map<GenericType, SqlType> genericMapping =
-          GenericsUtil.resolveGenerics(schema, argument);
-
-      for (final Entry<GenericType, SqlType> entry : genericMapping.entrySet()) {
-        final SqlType old = reservedGenerics.putIfAbsent(entry.getKey(), entry.getValue());
-        if (old != null && !old.equals(entry.getValue())) {
+        try {
+          GenericsUtil.reserveGenerics(type, argument, reservedGenerics);
+          return true;
+        } catch (final Exception e) {
```
lct45, line 369: I guess the other thing here is if we end up swallowing exceptions we don't want to, which actually seems pretty possible. In the old way, we had `udfUtil.resolveGenerics` -> `resolveGenerics(2)` -> `resolveGenerics(3)`. `resolveGenerics(2)` threw errors depending on what happened in `resolveGenerics(3)` and we didn't catch those anywhere, right? So in theory those should still be bubbling up through the new `genericsUtil.reserveGenerics` but in this case we'd catch those too. Sorry for going back and forth, but now it seems like we should change `reserveGenerics` to return a `bool` and call another helper function in `genericsUtil` to throw an error, or just keep both `reserveGenerics` with new names. It'd be helpful to see what some other people think though. cc: @agavra @guozhangwang

ksqldb-common/src/main/java/io/confluent/ksql/function/GenericsUtil.java
lct45: Hmm. It is the same logic, both checks make recursive calls to their respective functions so for that reason, I think it still makes sense to keep them in separate loops. It looks like the base case here is to add a generic to the list and the base case in `paramTypes` is to check to see if the `schema` and `instance` are a primitive match

ksqldb-common/src/test/java/io/confluent/ksql/function/types/ParamTypesTest.java
```diff
@@ -102,12 +102,27 @@ public void shouldPassCompatibleSchemas() {
...
     assertThat(ParamTypes.areCompatible(
-        SqlArgument.of(new SqlLambda(ImmutableList.of(SqlTypes.STRING), SqlTypes.STRING)),
+        SqlArgument.of(new SqlLambdaResolved(ImmutableList.of(SqlTypes.STRING), SqlTypes.STRING)),
         LambdaType.of(ImmutableList.of(ParamTypes.STRING), ParamTypes.STRING),
         false),
         is(true));
   }
 
+  @Test
+  public void shouldOnlyLookAtInputTypeSizeForCompatibleLambdaTypeAndSqlLambda() {
```
lct45, line 112: Thanks for adding these

ksqldb-execution/src/main/java/io/confluent/ksql/execution/util/FunctionArgumentsUtil.java
```diff
@@ -0,0 +1,228 @@
...
+import io.confluent.ksql.schema.ksql.SqlArgument;
+import io.confluent.ksql.schema.ksql.types.SqlLambda;
+import io.confluent.ksql.schema.ksql.types.SqlLambdaResolved;
+import io.confluent.ksql.schema.ksql.types.SqlType;
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+
+public final class FunctionArgumentsUtil {
```
lct45, line 38: Love this class, super helpful to share all the code and to keep the other methods clean. For my own information, why is `FunctionArgumentsAndContext` a sub-class instead of the main class?

ksqldb-common/src/main/java/io/confluent/ksql/function/GenericsUtil.java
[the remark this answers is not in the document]
guozhangwang: Got it, so in what scenarios the `UdfIndex#getCandidates` could return multiple results?

ksqldb-execution/src/main/java/io/confluent/ksql/execution/codegen/SqlToJavaVisitor.java
stevenpyzhang: hopefully the new refactor into the utility class makes it clearer

ksqldb-execution/src/main/java/io/confluent/ksql/execution/codegen/CodeGenRunner.java
stevenpyzhang: I don't think we need to since this would have been already checked in the UdfIndex as part of verifying the candidates with resolveGenerics.

ksqldb-execution/src/main/java/io/confluent/ksql/execution/codegen/CodeGenRunner.java
stevenpyzhang: added that as a java doc as part of the new util function used in all 3 classes

ksqldb-common/src/main/java/io/confluent/ksql/function/GenericsUtil.java
stevenpyzhang: It's very similar, but as Leah said, it's slightly different, I think having them separate is OK for now.

ksqldb-common/src/main/java/io/confluent/ksql/function/GenericsUtil.java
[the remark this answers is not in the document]
stevenpyzhang: I think the only scenario would be the second getCandidates call
```
    // if none were found (candidates is empty) try again with
    // implicit casting
    getCandidates(arguments, 0, root, candidates, new HashMap<>(), true);
```
when we allow casting of the function arguments when we do `return ParamTypes.areCompatible(argument, type, allowCasts);` in `accepts()`

ksqldb-common/src/main/java/io/confluent/ksql/function/GenericsUtil.java
stevenpyzhang: I was originally having issues combining the one in UdfIndex with the one I added since we need to be able to surface the error for invalid generic mappings in the util with lambda functions whereas it's OK to just return true/false in UdfIndex since it's checking candidate functions. I resolved this by having the new function I add return a Pair of boolean and exception to get around this. This still leaves us with a ResolveGeneric and ReserveGeneric in GenericUtils which is definitely confusing, but there's subtle differences to each functions behavior even though it looks like the same logic (`reserve generic` uses an instance of check before resolving, which `resolve generic` doesn't do)

In the end, GenericUtils already had a number of functions that appear to be doing the same logic (with different variations) even before lambdas so I would want to leave an overall refactor of the util out of this pr for now.

stevenpyzhang: Never mind, I think I'll be able to cut down on some of the generic util code..

stevenpyzhang: Done

stevenpyzhang:
@lct45 @agavra @guozhangwang @rodesai the PR is ready for a second look. It's changed quite drastically from the first iteration due to refactoring. If there was any comment I didn't address (I think I should've hit most of them...), could you please resolve the original and then make a new one? It would make keeping track of what needs to be done much easier.

@agavra it'd be good to especially sanity check GenericUtils since I reworked some existing functions

ksqldb-execution/src/main/java/io/confluent/ksql/execution/util/FunctionArgumentsUtil.java
```diff
@@ -0,0 +1,227 @@
...
+import java.util.Optional;
+
+public final class FunctionArgumentsUtil {
+
+  private FunctionArgumentsUtil() {
+
+  }
+
+  public static class FunctionArgumentsAndContext {
+    private final List<TypeContext> contexts;
```
rodesai, line 48: Couple things about `TypeContext`:

1. I don't think you need `visitType` in there anymore. Seems unused.
2. I think you could simplify `TypeContext` by just building the mapping from lambda variable to type in `getLambdaContextAndType` and creating a child `TypeContext` with this mapping, e.g.:

```
class TypeContext {
    // no list of types
    final Map<String, SqlType> lambdaInputTypeMapping;

    public TypeContext copyWithLambdaVariableTypeMapping(final Map<String, SqlType> variableToType) {
        final HashMap<String, SqlType> mapping = new HashMap<>(lambdaInputTypeMapping);
        for (final String v : variableToType.keySet()) {
          if (mapping.containsKey(v) && !mapping.get(v).equals(variableToType.get(v))) {
              throw new IllegalStateException();
          }
          mapping.put(v, variableToType.get(v));
        }
        return new TypeContext(ImmutableMap.copyOf(mapping));
    }
}

Then, in `getLambdaContextAndType` you can do:
```
if (expression instanceof LambdaFunctionCall) {
          final LambdaFunctionCall lambdaFunctionCall = (LambdaFunctionCall) expression;

          // the function returned from the UDF factory should have lambda
          // at this index in the function arguments if there's a
          // lambda node at this index in the function node argument list
          if (!(parameter instanceof LambdaType)) {
            throw new RuntimeException("Error while processing lambda function. "
                + "This is most likely an internal error and a "
                + "Github issue should be filed for debugging. "
                + "Include the function name, the parameters passed in, the expected "
                + "signature, and any other relevant information.");
          }

          final LambdaType lambdaParameter = (LambdaType) parameter;
          final Iterator<String> variablesI = lambdaFunctionCall.getArguments().listIterator();
          final HashMap<String, SqlType> variableTypeMapping = new HashMap<>();
          for (ParamType inputParam : lambdaParameter.inputTypes()) {
            if (inputParam instanceof GenericType) {
              final GenericType genericParam = (GenericType) inputParam;
              if (!reservedGenerics.containsKey(genericParam)) {
                throw new RuntimeException(
                    String.format(
                        "Could not resolve type for generic %s. "
                            + "The generic mapping so far: %s",
                        genericParam.toString(), reservedGenerics.toString()));
              }
              variableTypeMapping.put(variablesI.next(), reservedGenerics.get(genericParam));
            } else {
              variableTypeMapping.put(
                  variablesI.next(),
                  SchemaConverters.functionToSqlConverter().toSqlType(inputParam)
              );
            }
          }
          final TypeContext childContext = context.copyWithMapping(variableTypeMapping);
          final SqlType resolvedLambdaReturnType =
              expressionTypeManager.getExpressionSqlType(expression, childContext);
```

It's not a big change, but I think it helps readability because the relationship between the lambda parameters in the registered function and those in the expressions is tracked in one place (rather than this intermediate list and map in `TypeContext`.

It also makes `TypeContext` immutable which makes it easier to reason about.

File: ksqldb-execution/src/main/java/io/confluent/ksql/execution/util/ExpressionTypeManager.java [hunk 27]
```diff
@@ -78,7 +77,6 @@
 import io.confluent.ksql.util.KsqlConfig;
 import io.confluent.ksql.util.KsqlException;
 import io.confluent.ksql.util.VisitorUtil;
-import java.util.ArrayList;
```
rodesai, line 81: rather than exposing `TypeContext` in `getExpressionSqlType` I think it would be better to just add a parameter that's a map from String to SqlType that contains the types for all the lambda parameters. `TypeContext` really feels like it should be an internal detail of this class. I know we have it being passed around in `SqlToJavaVisitor` and `CodeGenRunner`, but I don't think we should use a common type unless there's really a common set of information that makes sense to pass around. In `TypeContext` case there's only 2 fields (well 3, but I think we should drop the list and make the map immutable as described below), and 1 (`sqlType`) is only used from `ExpressionTypeManager`.

ksqldb-execution/src/main/java/io/confluent/ksql/execution/util/FunctionArgumentsUtil.java
```diff
@@ -0,0 +1,227 @@
...
+   * @param udfFactory  udf factory for the function in the expression
+   * @param context a type context
+   *
+   * @return a wrapper that contains a list of function arguments 
+   *         (any lambdas are SqlLambdaResolved), the ksql function, 
+   *         type contexts for use in further processing the function 
+   *         argument child nodes, and the return type of the ksql function
+   */
+  // CHECKSTYLE_RULES.OFF: CyclomaticComplexity
+  public static FunctionArgumentsAndContext getLambdaContextAndType(
```
rodesai, line 105: nit: naming - `getFunctionTypeInfo`? - it's really returning you all the info about the function argument types and return types, not just for lambdas

ksqldb-execution/src/main/java/io/confluent/ksql/execution/util/FunctionArgumentsUtil.java
```diff
@@ -0,0 +1,227 @@
...
+      return returnType;
+    }
+
+    public KsqlScalarFunction getFunction() {
+      return function;
+    }
+  }
+
+  /**
+   * Given a function call node, we have to do a two pass processing of the
```
rodesai, line 83: nit: explain what the function is doing first:

something like
"Compute type information given a function call node. Specifically, computes the function return type, the types of all arguments, and the types of all lambda parameters for arguments that are lambda expressions."

ksqldb-execution/src/test/java/io/confluent/ksql/execution/util/FunctionArgumentsUtilTest.java
```diff
@@ -0,0 +1,176 @@
...
+    givenUdfWithNameAndReturnType("ComplexFunction", SqlTypes.DOUBLE);
+    when(function.parameters()).thenReturn(
+        ImmutableList.of(
+            ArrayType.of(GenericType.of("X")),
+            MapType.of(GenericType.of("K"), GenericType.of("V")),
+            GenericType.of("Q"),
+            LambdaType.of(ImmutableList.of(GenericType.of("K"), GenericType.of("V"), GenericType.of("Q")), GenericType.of("K")),
+            LambdaType.of(ImmutableList.of(GenericType.of("Q"), GenericType.of("V")), GenericType.of("X"))));
+
+    final FunctionCall expression =
```
rodesai, line 96: nit: move to it's own method like `givenFunctionCallWithMultipleLambdas`

ksqldb-execution/src/test/java/io/confluent/ksql/execution/util/FunctionArgumentsUtilTest.java
```diff
@@ -0,0 +1,176 @@
...
+  private ExpressionTypeManager expressionTypeManager;
+
+  @Rule
+  public final MockitoRule mockitoRule = MockitoJUnit.rule();
+
+  @Before
+  public void init() {
+    expressionTypeManager = new ExpressionTypeManager(SCHEMA, functionRegistry);
+  }
+
```
rodesai, line 83: add some test cases?
- no lambdas
- no generics in the lambdas

rodesai (commented):
Left some more comments, mostly some feedback on the structure of the code.

ksqldb-execution/src/main/java/io/confluent/ksql/execution/util/FunctionArgumentsUtil.java
```diff
@@ -0,0 +1,227 @@
...
+  }
+
+  public static class FunctionArgumentsAndContext {
+    private final List<TypeContext> contexts;
+    private final List<SqlArgument> arguments;
+    private final SqlType returnType;
+    private final KsqlScalarFunction function;
+
+    public FunctionArgumentsAndContext(
+        final List<TypeContext> contexts,
```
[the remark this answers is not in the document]
lct45: Overall I think this makes sense. We do need to track inputs somewhere though because we depend on input type ordering when creating a `SqlLambdaResolved`, so we can't just add the map keys as the input types since they don't have guaranteed ordering. I went ahead and added a list to `mapLambdaParametersToTypes` that just adds the inputs' sqlTypes, but do you think the list should be in the context instead? It doesn't need to be shared but it does feel kinda repetitive to build a map and a list of the values in the map at the same time

ksqldb-execution/src/main/java/io/confluent/ksql/execution/util/ExpressionTypeManager.java
lct45: I'm not sure I follow. Would we add the map parameter to all the methods in `ExpressionTypeManager`? Or change `TypeContext` back into `ExpressionTypeManagerContext` and include in that a mapping from string to SqlType for lambdas? We do still need the lambda variable mapping passed around, we can pass _in_ a list from `codeGenRunner` and `SqlToJavaVisitor` and then store it back in the `ExpressionTypeManagerContext` to be passed down where the lambda literal is resolved. I'm not sure I really prefer that method though since we'd also need to change `SqlToJavaVisitor` and potentially `CodeGenRunner` to pass around a map of lambda types since they also need the information shared between method calls. We could go back to an earlier iteration and have a separate context for each of the three classes, that would keep `SqlType` just in `ExpressionTypeManager`, but we'd be doing a lot of list copying from one context type to a different context type

ksqldb-execution/src/main/java/io/confluent/ksql/execution/util/FunctionArgumentsUtil.java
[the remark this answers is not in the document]
guozhangwang: I had similar thoughts when reviewing another PR by @lct45, and she convinced me that we indeed need both the list and the map because the ordering is required. Now with two passes, my understanding is that the second pass would still rely on the ordering so `SqlLambdaResolved` itself still need both list and map? maybe @lct45 can correct me here.

ksqldb-execution/src/main/java/io/confluent/ksql/execution/util/ExpressionTypeManager.java
stevenpyzhang: @rodesai what are your thoughts on creating a Context class that `SqlToJavaVisitor` and `CodeGenRunner` both use and then a `TypeContext` with `sqlType` field that ExpressionTypeManager uses?

ksqldb-execution/src/main/java/io/confluent/ksql/execution/util/FunctionArgumentsUtil.java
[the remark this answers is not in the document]
lct45: Yeah that exactly right, we do need some way of keeping the inputs in the correct order so having a map alone won't do. _where_ we put the list is somewhat flexible though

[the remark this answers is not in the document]
rodesai: We need the order at this level (the enclosing function call). But when we recurse down to the lambda expression itself, why do we need it any more? We should know the types for all the variables.

In today's implementation we basically do:
- build up a list of argument types
- recurse down to LambdaFunction
- correlate the list with the parameter list in LambdaFunction
- clear the list (at this point the list is gone so we don't care about the order anymore)

I'm proposing we just do all 4 in 1 place and store it as a map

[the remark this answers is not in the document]
rodesai: Never mind - I see what you mean now. I think we can just keep the list to within the implementation of `getLambdaContextAndType` and maybe `mapLambdaParametersToTypes` (you could just build it from the mapping and the params in the expression) rather than passing it around.

ksqldb-execution/src/main/java/io/confluent/ksql/execution/util/ExpressionTypeManager.java
rodesai: > change TypeContext back into ExpressionTypeManagerContext and include in that a mapping from string to SqlType for lambdas?
This would be better IMO.

> We could go back to an earlier iteration and have a separate context for each of the three classes, that would keep SqlType just in ExpressionTypeManager, but we'd be doing a lot of list copying from one context type to a different context type

Personally I prefer keeping a separate context type internal to each class that contains a visitor because it keeps things decoupled. Yes, you might end up with a few context types that look similar but I think this is generally more extensible and easier to understand. Over time, we may want to do different things from different visitors that require different state to be passed around. If we use a common context type, we'll have to have it support the union of the use cases from the different visitors (we already see this today with the `sqlType` method which only `ExpressionTypeManager` uses). This makes it harder to understand/maintain the codebase. If I want to change this context I have to think about all the different things that are going to use it. If it's private within some class I know it's only being used there.

> but we'd be doing a lot of list copying from one context type to a different context type
Unless we're on a performance path I'd optimize for readability/maintainability.

ksqldb-udf/src/main/java/io/confluent/ksql/schema/ksql/types/SqlLambda.java
```diff
@@ -15,47 +15,34 @@
...
   ) {
-    return new SqlLambda(inputType, returnType);
+    return new SqlLambda(numInputs);
   }
 
   public SqlLambda(
```
rodesai, line 36: make this private (or package-private if you need it for tests)

ksqldb-udf/src/main/java/io/confluent/ksql/schema/ksql/types/SqlLambdaResolved.java
```diff
@@ -0,0 +1,100 @@
...
+  }
+
+  public static SqlLambdaResolved of(
+      final List<SqlType> inputType,
+      final SqlType returnType
+  ) {
+    return new SqlLambdaResolved(inputType, returnType);
+  }
+
+  public SqlLambdaResolved(
```
rodesai, line 51: make this private

ksqldb-execution/src/main/java/io/confluent/ksql/execution/util/FunctionArgumentsUtil.java
```diff
@@ -0,0 +1,227 @@
...
+        final Expression expression = arguments.get(i);
+        final ParamType parameter = paramTypes.get(i);
+        if (expression instanceof LambdaFunctionCall) {
+          final TypeContext childContext = context.getCopy();
+
+          // the function returned from the UDF factory should have lambda
+          // at this index in the function arguments if there's a
+          // lambda node at this index in the function node argument list
+          if (!(parameter instanceof LambdaType)) {
+            throw new RuntimeException("Error while processing lambda function. "
```
guozhangwang, line 185: nit: maybe print the `ParamType` in the error message as well?

ksqldb-execution/src/test/java/io/confluent/ksql/execution/codegen/SqlToJavaVisitorTest.java
```diff
@@ -1046,8 +1046,8 @@ public void shouldGenerateCorrectCodeForNestedLambdas() {
             ArrayType.of(ParamTypes.DOUBLE),
             ParamTypes.DOUBLE,
             LambdaType.of(
-                ImmutableList.of(ParamTypes.DOUBLE, ParamTypes.DOUBLE),
-                ParamTypes.DOUBLE))
+                ImmutableList.of(ParamTypes.DOUBLE, ParamTypes.INTEGER),
```
guozhangwang, line 1049: Not clear why we need such changes in this test?

ksqldb-common/src/main/java/io/confluent/ksql/function/UdfIndex.java
```diff
@@ -203,9 +202,7 @@ private KsqlException createNoMatchingFunctionException(final List<SqlArgument>
           if (argument == null) {
             return "null";
           } else {
-            final Optional<SqlLambda> sqlLambda = argument.getSqlLambda();
-            return sqlLambda.map(lambda -> lambda.toString(FormatOptions.noEscape()))
-                .orElseGet(() -> argument.getSqlTypeOrThrow().toString(FormatOptions.noEscape()));
+            return argument.toString(FormatOptions.noEscape());
```
guozhangwang, line 205: Nice refactoring.

ksqldb-execution/src/main/java/io/confluent/ksql/execution/util/FunctionArgumentsUtil.java
```diff
@@ -0,0 +1,300 @@
...
+
+    final KsqlScalarFunction function = udfFactory.getFunction(functionArgumentTypes);
+    final SqlType returnSchema;
+    final List<TypeContext> typeContextsForChildren = new ArrayList<>();
+
+    if (!functionCall.hasLambdaFunctionCallArguments()) {
+      returnSchema = function.getReturnType(functionArgumentTypes);
+      return FunctionTypeInfo.of(
+          functionArgumentTypes.stream()
+              .map(argument -> ArgumentInfo.of(argument, context.getCopy()))
```
guozhangwang, line 166: Just wondering: could we reuse the copied context from `firstPassOverFunctionArguments` since that one seems not modified till here?

guozhangwang:
I think this is a great refactoring :) Made another pass.

ksqldb-execution/src/main/java/io/confluent/ksql/execution/util/FunctionArgumentsUtil.java
```diff
@@ -0,0 +1,300 @@
...
+    public KsqlScalarFunction getFunction() {
+      return function;
+    }
+  }
+
+  public static final class ArgumentInfo {
+    final SqlArgument sqlArgument;
+    final TypeContext context;
+
+    public static ArgumentInfo of(
```
rodesai, line 91: make private

ksqldb-execution/src/main/java/io/confluent/ksql/execution/util/FunctionArgumentsUtil.java
```diff
@@ -0,0 +1,300 @@
...
+  private FunctionArgumentsUtil() {
+
+  }
+
+  public static final class FunctionTypeInfo {
+    private final List<ArgumentInfo> argumentInfos;
+    private final SqlType returnType;
+    private final KsqlScalarFunction function;
+
+    public static FunctionTypeInfo of(
```
rodesai, line 56: make this private - you shouldn't be able to construct this outside of this class

ksqldb-execution/src/main/java/io/confluent/ksql/execution/util/FunctionArgumentsUtil.java
```diff
@@ -0,0 +1,300 @@
...
+    }
+
+    public KsqlScalarFunction getFunction() {
+      return function;
+    }
+  }
+
+  public static final class ArgumentInfo {
+    final SqlArgument sqlArgument;
+    final TypeContext context;
```
rodesai, line 89: does this need to be `TypeContext`? It should be a pretty small change to just put an optional map here and construct the context in the caller.

rodesai (approved):
LGTM based on offline discussion. Let's favor merging this PR and doing any more refactors in follow-ups, since we're coming down to the wire on 0.17 and there's a couple changes that possibly depend/conflict.

ksqldb-execution/src/main/java/io/confluent/ksql/execution/util/FunctionArgumentsUtil.java
stevenpyzhang: All the contexts need to be separate instances when processing child nodes of the function call arguments so this seems like the easiest way to do it.

ksqldb-execution/src/test/java/io/confluent/ksql/execution/codegen/SqlToJavaVisitorTest.java
stevenpyzhang: This test was written when we still had the restriction where we were getting the input type list from the ordering of the other function arguments. If you look at the FunctionCall arguments in the test, it's an Array<Double> and Integer so the previous InputTypeList was [Double, Integer]. To get that same input type list with the refactor, we need to update the LambdaType here.

ksqldb-execution/src/main/java/io/confluent/ksql/execution/util/FunctionArgumentsUtil.java
stevenpyzhang: done

ksqldb-execution/src/main/java/io/confluent/ksql/execution/util/FunctionArgumentsUtil.java
stevenpyzhang: done

ksqldb-execution/src/main/java/io/confluent/ksql/execution/util/FunctionArgumentsUtil.java
stevenpyzhang: I'll address this in the follow up context refactor, it'll naturally get addressed as part of that

ksqldb-execution/src/main/java/io/confluent/ksql/execution/util/FunctionArgumentsUtil.java
stevenpyzhang: done

ksqldb-common/src/main/java/io/confluent/ksql/function/GenericsUtil.java
```diff
@@ -270,25 +280,50 @@ private static boolean matches(final ParamType schema, final SqlArgument instanc
...
-    final Map<ParamType, SqlType> asMap = new HashMap<>();
-    for (final Entry<GenericType, SqlType> entry : mappings) {
-      final SqlType old = asMap.putIfAbsent(entry.getKey(), entry.getValue());
+    for (final Entry<GenericType, SqlType> entry : genericMapping) {
+      final SqlType old = reservedGenerics.putIfAbsent(entry.getKey(), entry.getValue());
       if (old != null && !old.equals(entry.getValue())) {
-        return false;
+        return new Pair<>(false, Optional.of(new KsqlException(String.format(
+            "Found invalid instance of generic schema when mapping %s to %s. "
+                + "Cannot map %s to both %s and %s",
```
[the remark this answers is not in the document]
stevenpyzhang: done

ksqldb-common/src/main/java/io/confluent/ksql/function/GenericsUtil.java
```diff
@@ -153,40 +156,35 @@ public static SqlType applyResolved(
...
+    if (!result.getLeft() && result.getRight().isPresent()) {
+      throw result.getRight().get();
     }
 
-    return ImmutableMap.copyOf(mapping);
+    return ImmutableMap.copyOf(genericMapping);
   }
 
+  /**
+   * Identifies a mapping from generic type to concrete type based on a {@code schema} and
```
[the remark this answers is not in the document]
stevenpyzhang: I changed both these instances to reserverGeneric now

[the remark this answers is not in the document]
stevenpyzhang: Ok to summarize what we have
`resolveGeneric` has the actual logic to return a list of generic mappings based on a single argument and schema
`reserveGeneric` turns that list into a map and verifies that there are no overlaps, one version of reserveGeneric takes an existing map and verifies the new mappings don't clash

ksqldb-execution/src/test/java/io/confluent/ksql/execution/codegen/SqlToJavaVisitorTest.java
```diff
@@ -1113,17 +1113,6 @@ public void shouldGenerateCorrectCodeForNestedLambdas() {
                 + "})) + 5)"));
   }
 
-  @Test
-  public void shouldThrowErrorOnEmptyLambdaInput() {
```
stevenpyzhang, line 1117: Leftover from when we had a list in the type context so this test got removed.

ksqldb-common/src/main/java/io/confluent/ksql/function/GenericsUtil.java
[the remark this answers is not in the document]
guozhangwang: Great, thanks for the clarification too!

stevenpyzhang:
I'm merging this PR in and will be working on an immediate follow up to allow for decoupling of the contexts from each of the visitors. I'll also keep an eye out on this PR for any straggler comments and look to address it in the follow up

PR #7109, opened 2021-02-26
feat: limit the number of active push queries everywhere using "ksql.max.push.queries" config

cprasad1:
### Description 
This PR limits the number of active push queries using the `ksql.max.push.queries` config.

### Testing done 
- Unit testing
- Manual Testing
- 
### Reviewer checklist

### DIFF
--- a/ksqldb-rest-app/src/main/java/io/confluent/ksql/api/impl/QueryEndpoint.java
+++ b/ksqldb-rest-app/src/main/java/io/confluent/ksql/api/impl/QueryEndpoint.java
@@ -36,6 +36,7 @@ import io.confluent.ksql.query.BlockingRowQueue;
 import io.confluent.ksql.rest.server.LocalCommands;
 import io.confluent.ksql.rest.server.resources.streaming.PullQueryConfigPlannerOptions;
 import io.confluent.ksql.rest.server.resources.streaming.PullQueryConfigRoutingOptions;
+import io.confluent.ksql.rest.util.QueryCapacityUtil;
 import io.confluent.ksql.schema.ksql.Column;
 import io.confluent.ksql.schema.utils.FormatOptions;
 import io.confluent.ksql.services.ServiceContext;
@@ -112,6 +113,14 @@ public class QueryEndpoint {
       final WorkerExecutor workerExecutor
   ) {
     final BlockingQueryPublisher publisher = new BlockingQueryPublisher(context, workerExecutor);
+
+    if (QueryCapacityUtil.exceedsPushQueryCapacity(ksqlEngine, ksqlConfig)) {
+      QueryCapacityUtil.throwTooManyActivePushQueriesException(
+              ksqlEngine,
+              ksqlConfig,
+              statement.getStatementText()
+      );
+    }
 
     final TransientQueryMetadata queryMetadata = ksqlEngine
         .executeQuery(serviceContext, statement, true);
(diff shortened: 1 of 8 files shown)


ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/util/QueryCapacityUtil.java
```diff
@@ -53,4 +54,36 @@ public static void throwTooManyActivePersistentQueriesException(
...
+
+  public static int getNumLivePushQueries(final KsqlExecutionContext ctx) {
+    return ctx.getAllLiveQueries().size() - ctx.getPersistentQueries().size();
+  }
+
+  public static int getPushQueryLimit(final KsqlConfig ksqlConfig) {
+    return Integer.parseInt(ksqlConfig
+            .originals()
+            .getOrDefault(KsqlRestConfig.MAX_PUSH_QUERIES, String.valueOf(Integer.MAX_VALUE))
+            .toString());
```
cprasad1, line 86: I would love to avoid this weird and ugly gymnastics that is required for getting the `MAX_PUSH_QUERIES` config. Any suggestions?

ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/resources/streaming/StreamedQueryResource.java
```diff
@@ -353,6 +354,14 @@ private EndpointResponse handlePushQuery(
     final ConfiguredStatement<Query> configured = ConfiguredStatement
         .of(statement, SessionConfig.of(ksqlConfig, streamsProperties));
 
+    final int numPushQueries = QueryCapacityUtil.getNumLivePushQueries(ksqlEngine);
```
AlanConfluent, line 357: This is just one of the endpoints `/query`.  We want both the others.  There's already a check for `/query-stream`, but it's inconsistent with this.  I'm referring to the code in `io.confluent.ksql.api.server.Server.registerQuery`.  It seems like it was going to be part of some greater organization of connections, but wasn't extended to other endpoints.  

@vcrfxia and @agavra  I think reviewed some of that original code and can maybe comment.  If we want to keep that, then we can declare `/query-stream` done. Otherwise removing it and being consistent with this code would be good.

We should at least be sure to add this to `/ws/query` as well.

agavra: I have a meta-question here, why do we track this at the endpoint instead of in the engine? Seems a little weird to me...

ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/resources/streaming/StreamedQueryResource.java
```diff
@@ -172,6 +175,7 @@ public void configure(final KsqlConfig config) {
     }
 
     ksqlConfig = config;
+    ksqlRestConfig = new KsqlRestConfig(ksqlConfig.originals());
```
AlanConfluent, line 178: Nit: You can just pass this in the constructor when created in `KsqlRestApplication` rather than creating it here.

ksqldb-rest-app/src/main/java/io/confluent/ksql/api/impl/QueryEndpoint.java
```diff
@@ -77,6 +80,7 @@ public QueryEndpoint(
   ) {
     this.ksqlEngine = ksqlEngine;
     this.ksqlConfig = ksqlConfig;
+    this.ksqlRestConfig = new KsqlRestConfig(ksqlConfig.originals());
```
AlanConfluent, line 83: Nit: You can just pass this in from `KsqlServerEndpoints` -> `KsqlRestApplication` if you wanted.  That's a little more consistent with how we usually create it.

AlanConfluent (approved):

PR #7120, opened 2021-03-01
build: add back maven settings file into release commit (MINOR)

vcrfxia:
### Description 

After the recent refactoring to the build script, the maven-settings.xml file is no longer getting checked into the git tag, which means the release branches are no longer buildable. This PR adds back the file.

### Testing done 

Wasn't able to test this with jenkins replay since the fix requires changes to build-packages.sh which we can't edit in replay. @andrewegel do you know of a workaround for testing this?

### Reviewer checklist

### DIFF
--- a/build-packages.sh
+++ b/build-packages.sh
@@ -122,6 +122,7 @@ echo "Versioning Diff:"
 git diff | cat
 
 # Commit changes
+git add .
 git commit -a -m "build: Setting project version ${FULL_VERSION} and parent version ${UPSTREAM_VERSION}."
 
 # We run things through fakeroot, which causes issues with finding an .m2/repository location to write. We set the homedir where it does
--- a/build-packages.sh
+++ b/build-packages.sh
@@ -122,7 +122,7 @@ echo "Versioning Diff:"
 git diff | cat
 
 # Commit changes
-git add .
+git add maven-settings.xml
 git commit -a -m "build: Setting project version ${FULL_VERSION} and parent version ${UPSTREAM_VERSION}."
 
 # We run things through fakeroot, which causes issues with finding an .m2/repository location to write. We set the homedir where it does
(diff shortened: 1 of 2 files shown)


build-packages.sh
```diff
@@ -122,6 +122,7 @@ echo "Versioning Diff:"
 git diff | cat
 
 # Commit changes
+git add .
```
andrewegel, line 125:
```suggestion
git add maven-settings.xml
```

I would make this more specific so we know we're only adding this file, and not any other "junk"

vcrfxia: Are we okay with hard-coding the name of the settings file like this? It seems unlikely to change so it might be OK. We'd have to update it in a ton of other places if we ever changed it.

Other options would be to pass in the name of the file as a param (seems like overkill) or to try to copy it out of the .m2, and control the name as a result (is this feasible)?

andrewegel:
> Wasn't able to test this with jenkins replay since the fix requires changes to build-packages.sh which we can't edit in replay. @andrewegel do you know of a workaround for testing this?

Theres not a way to work around that - Relay will only work for the code in the Jenkinsfile or code imported from jenkins-common.

build-packages.sh
andrewegel: > Are we okay with hard-coding the name of the settings file like this? It seems unlikely to change so it might be OK. We'd have to update it in a ton of other places if we ever changed it.
> 
> Other options would be to pass in the name of the file as a param (seems like overkill) or to try to copy it out of the .m2, and control the name as a result (is this feasible)?

I think you said it already, its unlikely to change so its probably OK - Copying it out of ~/.m2/settings.xml is also viable.

andrewegel (approved):

agavra (approved):

PR #7132, opened 2021-03-02
chore(migrations): fail early on incompatible server version

vcrfxia:
### Description 

Two main changes in this PR:
- check server version in `migrations initialize` and fail if the server version is incompatible
- check whether `migrations initialize` has been called (specifically, whether the metadata stream and table exist) from each of `validation`, `apply`, and `info`, and fail if not

Also adds a helper method that will be used in `migrations info`, and cleans up a couple minor bugs.

### Testing done 

New unit tests added. Existing integration tests continue to pass.

### Reviewer checklist

### DIFF
--- a/ksqldb-tools/src/main/java/io/confluent/ksql/tools/migrations/commands/MigrationInfoCommand.java
+++ b/ksqldb-tools/src/main/java/io/confluent/ksql/tools/migrations/commands/MigrationInfoCommand.java
@@ -16,8 +16,8 @@
 package io.confluent.ksql.tools.migrations.commands;
 
 import static io.confluent.ksql.tools.migrations.util.MigrationsDirectoryUtil.getMigrationsDirFromConfigFile;
-import static io.confluent.ksql.tools.migrations.util.MigrationsUtil.getServerInfo;
-import static io.confluent.ksql.tools.migrations.util.MigrationsUtil.versionSupportsMultiKeyPullQuery;
+import static io.confluent.ksql.tools.migrations.util.ServerVersionUtil.getServerInfo;
+import static io.confluent.ksql.tools.migrations.util.ServerVersionUtil.versionSupportsMultiKeyPullQuery;
 
 import com.github.rvesse.airline.annotations.Command;
 import com.google.common.annotations.VisibleForTesting;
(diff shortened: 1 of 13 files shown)


ksqldb-tools/src/main/java/io/confluent/ksql/tools/migrations/commands/MigrationInfoCommand.java
```diff
@@ -29,12 +42,59 @@
...
+      LOGGER.error(e.getMessage());
+      return 1;
+    }
+
+    if (!validateMetadataInitialized(ksqlClient, config)) {
+      ksqlClient.close();
+      return 1;
+    }
+
+    throw new NotImplementedException();
```
vcrfxia, line 83: This implementation will be finished in a future PR. (I wanted to open this one first in order to merge the refactors sooner to minimize merge conflicts as other work is ongoing.)

ksqldb-tools/src/main/java/io/confluent/ksql/tools/migrations/util/ServerVersionUtil.java
```diff
@@ -0,0 +1,134 @@
...
+      version = new KsqlServerVersion(ksqlServerVersion);
+    } catch (IllegalArgumentException e) {
+      throw new MigrationException("Could not parse ksqlDB server version to "
+          + "verify compatibility. Version: " + ksqlServerVersion);
+    }
+
+    return version.isAtLeast(6, 0, 0, 10);
+  }
+
+  public static boolean versionSupportsMultiKeyPullQuery(final String ksqlServerVersion) {
```
vcrfxia, line 74: This method will be used by `migrations info` in order to determine whether to issue a single, multi-key pull query to fetch version info, or whether we'll need to fall back to multiple, single-key lookups instead.

jzaralim (approved):
lgtm. will there ever be a scenario where the patch number determines compatibility (6.0.1 vs 6.0.2 for example)?

vcrfxia:
> will there ever be a scenario where the patch number determines compatibility (6.0.1 vs 6.0.2 for example)?

Hopefully not, but we can always add support for this in the future if we need it :)

PR #7133, opened 2021-03-02
feat(migrations): implement `create` command

vcrfxia:
### Description 

The `migrations create` command creates an empty migrations file, of the appropriate filename format.

### Testing done 

Unit + integration.

### Reviewer checklist

### DIFF
--- a/ksqldb-tools/src/main/java/io/confluent/ksql/tools/migrations/commands/CreateMigrationCommand.java
+++ b/ksqldb-tools/src/main/java/io/confluent/ksql/tools/migrations/commands/CreateMigrationCommand.java
@@ -15,21 +15,33 @@
 
 package io.confluent.ksql.tools.migrations.commands;
 
+import static io.confluent.ksql.tools.migrations.util.MigrationsDirectoryUtil.getAllVersions;
+import static io.confluent.ksql.tools.migrations.util.MigrationsDirectoryUtil.getFilePathForVersion;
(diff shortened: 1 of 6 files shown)


ksqldb-tools/src/main/java/io/confluent/ksql/tools/migrations/commands/CreateMigrationCommand.java
```diff
@@ -15,49 +15,154 @@
...
+   */
+  private boolean validateVersionDoesNotAlreadyExist(final String migrationsDir) {
+    // no explicit version was specified, nothing to verify
+    // (airline actually can't distinguish between explicit 0 and nothing specified,
+    // but we won't worry about this edge case for now)
+    if (version == 0) {
+      return true;
+    }
+
+    final Optional<String> existingFile;
```
jzaralim, line 106: We should also do this check in apply, since `validate` only checks for duplicate versions for versions that have already been migrated. Currently, apply will still run if there are two unmigrated files with the same version number. This can be done in a follow-up.

jzaralim (approved):

ksqldb-tools/src/main/java/io/confluent/ksql/tools/migrations/commands/CreateMigrationCommand.java
vcrfxia: Sure, I'm working on a follow-up with some changes to `apply` right now anyway, so I'll slip this into my next PR. Good call!

PR #7143, opened 2021-03-02
feat: Send taskMetadata to health check for uptime metric

wcarlson5:
### Description 
This includes the task metadata for each query in the healthcheck response

### Testing done 

### Reviewer checklist

### DIFF
--- a/ksqldb-engine/src/main/java/io/confluent/ksql/util/QueryMetadata.java
+++ b/ksqldb-engine/src/main/java/io/confluent/ksql/util/QueryMetadata.java
@@ -39,12 +39,14 @@ import java.util.Optional;
 import java.util.Queue;
 import java.util.Set;
 import java.util.function.Consumer;
+import java.util.stream.Collectors;
 import org.apache.kafka.streams.KafkaStreams;
 import org.apache.kafka.streams.KafkaStreams.State;
 import org.apache.kafka.streams.LagInfo;
 import org.apache.kafka.streams.Topology;
 import org.apache.kafka.streams.errors.StreamsException;
 import org.apache.kafka.streams.errors.StreamsUncaughtExceptionHandler;
+import org.apache.kafka.streams.processor.TaskMetadata;
 import org.apache.kafka.streams.state.StreamsMetadata;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
@@ -208,6 +210,13 @@ public abstract class QueryMetadata {
     }
     retryEvent.backOff();
     return StreamsUncaughtExceptionHandler.StreamThreadExceptionResponse.REPLACE_THREAD;
+  }
+
+  public Set<TaskMetadata> getTaskMetadata() {
+    return kafkaStreams.localThreadsMetadata()
+                       .stream()
+                       .flatMap(t -> t.activeTasks().stream())
+                       .collect(Collectors.toSet());
   }
 
   public Map<String, Object> getOverriddenProperties() {
--- a/ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/healthcheck/HealthCheckAgent.java
+++ b/ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/healthcheck/HealthCheckAgent.java
@@ -36,10 +36,12 @@ import java.util.Collections;
 import java.util.List;
 import java.util.Map;
 import java.util.Objects;
+import java.util.Set;
 import java.util.stream.Collectors;
 import org.apache.commons.lang3.exception.ExceptionUtils;
 import org.apache.kafka.clients.admin.DescribeTopicsOptions;
 import org.apache.kafka.common.errors.UnknownTopicOrPartitionException;
+import org.apache.kafka.streams.processor.TaskMetadata;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
@@ -49,6 +51,7 @@ public class HealthCheckAgent {
   public static final String METASTORE_CHECK_NAME = "metastore";
   public static final String KAFKA_CHECK_NAME = "kafka";
   public static final String COMMAND_RUNNER_CHECK_NAME = "commandRunner";
+  public static final String QUERY_TASKS_METADATA_NAME = "metadata";
 
   private static final List<Check> DEFAULT_CHECKS = ImmutableList.of(
       new ExecuteStatementCheck(METASTORE_CHECK_NAME, "list streams; list tables; list queries;"),
@@ -84,7 +87,15 @@ public class HealthCheckAgent {
         ));
     final boolean allHealthy = results.values().stream()
         .allMatch(HealthCheckResponseDetail::getIsHealthy);
-    return new HealthCheckResponse(allHealthy, results);
+    final Map<String, Set<TaskMetadata>> metadata = ksqlClient
+            .makeClusterStatusRequest(serverEndpoint)
+            .getResponse()
+            .getClusterStatus()
+            .values()
+            .stream()
+            .flatMap(t -> t.getTasksMetadataPerQuery().entrySet().stream())
+            .collect(Collectors.toMap(Map.Entry::getKey, Map.Entry::getValue));
+    return new HealthCheckResponse(allHealthy, results, metadata);
   }
 
   private interface Check {
--- a/ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/resources/ClusterStatusResource.java
+++ b/ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/resources/ClusterStatusResource.java
@@ -36,6 +36,8 @@ import java.util.Map;
 import java.util.Optional;
 import java.util.Set;
 import java.util.stream.Collectors;
+
+import org.apache.kafka.streams.processor.TaskMetadata;
 import org.apache.kafka.streams.state.HostInfo;
 import org.apache.kafka.streams.state.StreamsMetadata;
 
@@ -79,7 +81,8 @@ public class ClusterStatusResource {
                                           entry.getValue().getLastStatusUpdateMs(),
                                           getActiveStandbyInformation(entry.getKey()),
                                           getHostStoreLags(entry.getKey(),
-                                              entry.getValue().isHostAlive()))
+                                          entry.getValue().isHostAlive()),
+                                              getTaskMetadata(entry.getKey()))
         ));
 
     return new ClusterStatusResponse(response);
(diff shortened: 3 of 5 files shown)


ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/healthcheck/HealthCheckAgent.java
```diff
@@ -84,7 +87,15 @@ public HealthCheckResponse checkHealth() {
         ));
     final boolean allHealthy = results.values().stream()
         .allMatch(HealthCheckResponseDetail::getIsHealthy);
-    return new HealthCheckResponse(allHealthy, results);
+    final Map<String, Set<TaskMetadata>> metadata = ksqlClient
+            .makeClusterStatusRequest(serverEndpoint)
```
wcarlson5, line 91: I am not sure if the correct thing todo was this, I could maybe make a new endpoint but that seemed even more disruptive. Also does this query the one instance or the whole cluster. If it is the later we can have the health check just hit this instead of adding it to the health check response.

ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/resources/ClusterStatusResource.java
```diff
@@ -119,6 +122,19 @@ private HostStoreLags getHostStoreLags(final KsqlHostInfo ksqlHostInfo, final bo
     return perQueryMap;
   }
 
+  private Map<String, Set<TaskMetadata>> getTaskMetadata(
+          final KsqlHostInfo ksqlHostInfo
+  ) {
+    final Map<String, Set<TaskMetadata>> perQueryMap = new HashMap<>();
+    for (PersistentQueryMetadata persistentQueryMetadata: engine.getPersistentQueries()) {
```
wcarlson5, line 129: This was the best way I found to get to the `PersistentQueryMetadata` to call `getTaskMetadata()` please let me know if there is a better way

ksqldb-rest-model/src/main/java/io/confluent/ksql/rest/entity/HealthCheckResponse.java
```diff
@@ -22,23 +22,29 @@
...
   private final boolean isHealthy;
   private final ImmutableMap<String, HealthCheckResponseDetail> details;
+  private final ImmutableMap<String, Set<TaskMetadata>> metadata;
 
   @JsonCreator
   public HealthCheckResponse(
       @JsonProperty("isHealthy") final boolean isHealthy,
-      @JsonProperty("details") final Map<String, HealthCheckResponseDetail> details
+      @JsonProperty("details") final Map<String, HealthCheckResponseDetail> details,
+      @JsonProperty("taskMetadata") final Map<String, Set<TaskMetadata>> metadata
```
wcarlson5, line 43: I thought about adding it to details and making this its own `Check` but it didn't fit the pattern

wcarlson5 (commented):
I am really not sure that this is the best way to go about it but it should work for a POC

ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/healthcheck/HealthCheckAgent.java
```diff
@@ -49,6 +51,7 @@
   public static final String METASTORE_CHECK_NAME = "metastore";
   public static final String KAFKA_CHECK_NAME = "kafka";
   public static final String COMMAND_RUNNER_CHECK_NAME = "commandRunner";
+  public static final String QUERY_TASKS_METADATA_NAME = "metadata";
```
spena, line 54: Use a different name for this. Metadata in ksql will be confused with metastore.

ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/healthcheck/HealthCheckAgent.java
```diff
@@ -84,7 +87,15 @@ public HealthCheckResponse checkHealth() {
         ));
     final boolean allHealthy = results.values().stream()
         .allMatch(HealthCheckResponseDetail::getIsHealthy);
-    return new HealthCheckResponse(allHealthy, results);
+    final Map<String, Set<TaskMetadata>> metadata = ksqlClient
```
spena, line 90: Could you implement this check in another class, i.e. QueryUpstreamCheck? Like we do with the metastore, kafka, and commandrunner checks. It is more structured and won't require changes in HealthCheckResponse.

ksqldb-engine/src/main/java/io/confluent/ksql/util/QueryMetadata.java
```diff
@@ -210,6 +212,13 @@ public void onStop(final Consumer<Boolean> onStop) {
     return StreamsUncaughtExceptionHandler.StreamThreadExceptionResponse.REPLACE_THREAD;
   }
 
+  public Set<TaskMetadata> getTaskMetadata() {
+    return kafkaStreams.localThreadsMetadata()
+                       .stream()
+                       .flatMap(t -> t.activeTasks().stream())
+                       .collect(Collectors.toSet());
```
spena, line 219: Shouldn't it be better if an aggregation is returned here? Or are there plans to look at each task of the stream per query? 

Say a user has 100 queries running, how many tasks are expected per query (in average)? Would this make the healthcheck slow if the list is high?

ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/healthcheck/HealthCheckAgent.java
spena: I am thinking if we should return the overall status of the queries in the whole cluster. If we need a more detailed response, maybe a new endpoint should be created (say `/diagnostics` or something) that returns the issues per instance?

ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/resources/ClusterStatusResource.java
spena: `engine` is the right place to get the persistent queries. Btw, this is only getting the queries from the current ksql node. What about the rest of the nodes? If the query is idle in this node but active in another, this won't return the uptime of the active one.

ksqldb-rest-model/src/main/java/io/confluent/ksql/rest/entity/HealthCheckResponse.java
spena: Well, the healthcheck is checking the query uptime, so it fits, right? Btw, why returning all details of the streams instead of just the aggregated values of all tasks in a query?

ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/resources/ClusterStatusResource.java
wcarlson5: That is good. I think the idea was to only check the active queries and then we aggregate them in the health check and send them to dd and metabase

ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/healthcheck/HealthCheckAgent.java
wcarlson5: We actually can just remove it

ksqldb-engine/src/main/java/io/confluent/ksql/util/QueryMetadata.java
wcarlson5: There are plans to look at each task. It really depends on how complicated the query is but there should not be too many tasks and the taskMetadata is not too heavy of an object.

ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/healthcheck/HealthCheckAgent.java
wcarlson5: I did do that at first but that would require a change to the HealthCheckResponseDetail which then would complicate all the other checks. If you think that is better we can do that instead though

ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/healthcheck/HealthCheckAgent.java
wcarlson5: I'm not sure I follow. what do you mean by "the overall status of the queries in the whole cluster"?

The health check is hitting the `/healthcheck` endpoint. are you saying it should get the taskMetadata from a `/diagnostics` endpoint of the status request here should hit a more targeted endpoint than the entire `ClusterStatus`

ksqldb-rest-model/src/main/java/io/confluent/ksql/rest/entity/HealthCheckResponse.java
wcarlson5: The Checks return a boolean in the detail. But we have the information to make that boolean until we receive it all in the health check. That is also why we are sending all the details

wcarlson5 (commented):
I am a little confused. I think I need to look more at exactly what the health check receives and from where. @spena maybe you can clarify how the health check views a cluster vs one instance

ksqldb-rest-model/src/main/java/io/confluent/ksql/rest/entity/HealthCheckResponse.java
```diff
@@ -22,23 +22,29 @@
...
 import java.util.Objects;
+import java.util.Set;
 
 @JsonIgnoreProperties(ignoreUnknown = true)
 @Immutable
 public final class HealthCheckResponse {
 
   private final boolean isHealthy;
   private final ImmutableMap<String, HealthCheckResponseDetail> details;
+  private final ImmutableMap<String, Set<TaskMetadata>> metadata;
```
abbccdda, line 37: tasksMetadata?

ksqldb-rest-model/src/main/java/io/confluent/ksql/rest/entity/HealthCheckResponse.java
abbccdda: I feel doing a pre-aggregation here might help, but is this gonna be per query or per instance?

wcarlson5:
I talked with @rodesai a bit and it looks like we are going to add the `tasksMetadata` to the describe and add a separate thread in the health check to take care or it. So @abbccdda and @spena I will close this PR in favor of that once I get it up and running. Thanks for the feedback!

wcarlson5:
closed in favor of https://github.com/confluentinc/ksql/pull/7151

PR #7144, opened 2021-03-03
chore(migrations): enable auth for migrations tool

jzaralim:
### Description 
Reads auth configs from the config file and passes them to ClientOptions.

### Testing done 
unit test for creating ClientOptions.

### Reviewer checklist

### DIFF
--- a/ksqldb-tools/src/main/java/io/confluent/ksql/tools/migrations/MigrationConfig.java
+++ b/ksqldb-tools/src/main/java/io/confluent/ksql/tools/migrations/MigrationConfig.java
@@ -190,7 +190,7 @@ public final class MigrationConfig extends AbstractConfig {
       serviceId = ServerVersionUtil.getServerInfo(ksqlClient, ksqlServerUrl).getKsqlServiceId();
       ksqlClient.close();
       return serviceId;
-    } catch(MigrationException e) {
+    } catch (MigrationException e) {
       ksqlClient.close();
       throw e;
     }
--- a/ksqldb-tools/src/main/java/io/confluent/ksql/tools/migrations/MigrationConfig.java
+++ b/ksqldb-tools/src/main/java/io/confluent/ksql/tools/migrations/MigrationConfig.java
@@ -15,7 +15,9 @@
 
 package io.confluent.ksql.tools.migrations;
 
+import io.confluent.ksql.api.client.Client;
 import io.confluent.ksql.properties.PropertiesUtil;
+import io.confluent.ksql.tools.migrations.util.MigrationsUtil;
 import io.confluent.ksql.tools.migrations.util.ServerVersionUtil;
 import java.io.File;
 import java.util.Map;
@@ -27,6 +29,17 @@ import org.apache.kafka.common.config.ConfigDef.Type;
 public final class MigrationConfig extends AbstractConfig {
 
   public static final String KSQL_SERVER_URL = "ksql.server.url";
+
+  public static final String KSQL_USERNAME = "ksql.username";
+  public static final String KSQL_PASSWORD = "ksql.password";
+
+  public static final String SSL_TRUSTSTORE_LOCATION = "ssl.truststore.location";
+  public static final String SSL_TRUSTSTORE_PASSWORD = "ssl.truststore.password";
+  public static final String SSL_KEYSTORE_LOCATION = "ssl.keystore.location";
+  public static final String SSL_KEYSTORE_PASSWORD = "ssl.keystore.password";
+  public static final String SSL_KEY_PASSWORD = "ssl.key.password";
+  public static final String SSL_KEY_ALIAS = "ssl.key.alias";
+  public static final String SSL_ALPN = "ssl.alpn";
 
   public static final String KSQL_MIGRATIONS_STREAM_NAME = "ksql.migrations.stream.name";
   public static final String KSQL_MIGRATIONS_STREAM_NAME_DEFAULT = "migration_events";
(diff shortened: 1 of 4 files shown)


ksqldb-tools/src/main/java/io/confluent/ksql/tools/migrations/MigrationConfig.java
```diff
@@ -94,11 +160,24 @@ private MigrationConfig(final Map<String, String> configs, final String id) {
...
 
   private static String getServiceId(final Map<String, String> configs) throws MigrationException {
-    final String ksqlServerUrl = configs.get(MigrationConfig.KSQL_SERVER_URL);
+    final String ksqlServerUrl = configs.get(KSQL_SERVER_URL);
     if (ksqlServerUrl == null) {
       throw new MigrationException("Missing required property: " + MigrationConfig.KSQL_SERVER_URL);
     }
 
-    return ServerVersionUtil.getServerInfo(ksqlServerUrl).getKsqlServiceId();
+    final Client ksqlClient = MigrationsUtil.getKsqlClient(
```
vcrfxia, line 168: Where is this client closed? The reason we used to have two versions of `ServerVersionUtil.getServerInfo()` was so that one didn't require client management (creation and deletion) and the other did.

ksqldb-tools/src/main/java/io/confluent/ksql/tools/migrations/util/MigrationsUtil.java
```diff
@@ -47,6 +88,29 @@ public static Client getKsqlClient(final String ksqlServerUrl) throws MigrationE
         .setHost(url.getHost())
         .setPort(url.getPort());
 
-    return Client.create(options);
+    if (username != null || password != null) {
+      options.setBasicAuthCredentials(username, password);
+    }
+
+    // CHECKSTYLE_RULES.OFF: BooleanExpressionComplexity
+    final boolean useTls = !Strings.isNullOrEmpty(sslTrustStoreLocation)
```
vcrfxia, line 96: The logic here is not correct. It's possible that a user will want to use TLS even without any of these optional properties set. We should infer whether to use TLS or not based on the ksql server URL (whether it starts with HTTPS) instead.

ksqldb-tools/src/main/java/io/confluent/ksql/tools/migrations/MigrationConfig.java
```diff
@@ -52,8 +65,61 @@ private MigrationConfig(final Map<String, String> configs, final String id) {
...
+            "The key password"
+        ).define(
+            SSL_KEY_ALIAS,
+            Type.STRING,
+            null,
+            Importance.MEDIUM,
+            "The key alias"
+        ).define(
+            SSL_ALPN,
+            Type.STRING,
```
vcrfxia, line 118: Any reason this isn't a boolean type?

ksqldb-tools/src/main/java/io/confluent/ksql/tools/migrations/MigrationConfig.java
```diff
@@ -52,8 +65,61 @@ private MigrationConfig(final Map<String, String> configs, final String id) {
...
+            false,
+            Importance.MEDIUM,
+            "Whether ALPN should be used. It defaults to false."
+        ).define(
```
vcrfxia, line 122: Let's also add a config to control whether hostname verification is enabled or not.

ksqldb-tools/src/main/java/io/confluent/ksql/tools/migrations/util/MigrationsUtil.java
```diff
@@ -47,6 +88,29 @@ public static Client getKsqlClient(final String ksqlServerUrl) throws MigrationE
...
+        || !Strings.isNullOrEmpty(sslTrustStorePassword)
+        || !Strings.isNullOrEmpty(sslKeystoreLocation)
+        || !Strings.isNullOrEmpty(sslKeystorePassword)
+        || !Strings.isNullOrEmpty(sslKeyPassword)
+        || !Strings.isNullOrEmpty(sslKeyAlias)
+        || useAlpn;
+    // CHECKSTYLE_RULES.ON: BooleanExpressionComplexity
+
+    options.setUseTls(useTls);
+    options.setTrustStore(sslTrustStoreLocation);
```
vcrfxia, line 106: nit: these options only need to be set if `useTls` is true.

ksqldb-tools/src/main/java/io/confluent/ksql/tools/migrations/MigrationConfig.java
```diff
@@ -28,6 +30,17 @@
 
   public static final String KSQL_SERVER_URL = "ksql.server.url";
 
+  public static final String KSQL_USERNAME = "ksql.username";
```
vcrfxia, line 33: nit: rename to clarify that this is used for basic auth
```suggestion
  public static final String KSQL_BASIC_AUTH_USERNAME = "ksql.auth.basic.username";
```
and similarly for password below.

vcrfxia (commented):
Thanks @jzaralim ! Quick question: how did you test this? Ideally we'd add an integration test but we can defer on that for now (until after the release deadline) if we've tested manually.

ksqldb-tools/src/main/java/io/confluent/ksql/tools/migrations/MigrationConfig.java
jzaralim: Added `ksqlClient.close()` to this function.

jzaralim:
> Thanks @jzaralim ! Quick question: how did you test this? Ideally we'd add an integration test but we can defer on that for now (until after the release deadline) if we've tested manually.

I tested using a Confluent Cloud ksql server

ksqldb-tools/src/main/java/io/confluent/ksql/tools/migrations/MigrationConfig.java
```diff
@@ -167,17 +174,20 @@ private static String getServiceId(final Map<String, String> configs) throws Mig
...
+        configs.get(KSQL_BASIC_AUTH_USERNAME),
+        configs.get(KSQL_BASIC_AUTH_PASSWORD),
         configs.get(SSL_TRUSTSTORE_LOCATION),
         configs.get(SSL_TRUSTSTORE_PASSWORD),
         configs.get(SSL_KEYSTORE_LOCATION),
         configs.get(SSL_KEYSTORE_PASSWORD),
         configs.get(SSL_KEY_ALIAS),
         configs.get(SSL_KEY_PASSWORD),
-        configs.get(SSL_ALPN).equals("true")
+        configs.getOrDefault(SSL_ALPN, "false").equals("true"),
```
vcrfxia, line 185: Should this check be case-insensitive? I don't remember off the top of my head how AbstractConfig handles case-sensitivity for booleans. My guess is that it's case-insensitive but we should verify.

ksqldb-tools/src/main/java/io/confluent/ksql/tools/migrations/MigrationConfig.java
```diff
@@ -167,17 +174,20 @@ private static String getServiceId(final Map<String, String> configs) throws Mig
...
+        configs.getOrDefault(SSL_VERIFY_HOST, "true").equals("true")
     );
-
-    return ServerVersionUtil.getServerInfo(ksqlClient, ksqlServerUrl).getKsqlServiceId();
+    final String serviceId;
+    serviceId = ServerVersionUtil.getServerInfo(ksqlClient, ksqlServerUrl).getKsqlServiceId();
```
vcrfxia, line 189: This method could throw. We need to wrap this in a try-catch to ensure the client gets closed even if the server info call fails.

ksqldb-tools/src/main/java/io/confluent/ksql/tools/migrations/util/MigrationsUtil.java
```diff
@@ -92,25 +98,19 @@ static ClientOptions createClientOptions(
...
-    final boolean useTls = !Strings.isNullOrEmpty(sslTrustStoreLocation)
-        || !Strings.isNullOrEmpty(sslTrustStorePassword)
-        || !Strings.isNullOrEmpty(sslKeystoreLocation)
-        || !Strings.isNullOrEmpty(sslKeystorePassword)
-        || !Strings.isNullOrEmpty(sslKeyPassword)
-        || !Strings.isNullOrEmpty(sslKeyAlias)
-        || useAlpn;
-    // CHECKSTYLE_RULES.ON: BooleanExpressionComplexity
-
+    final boolean useTls = ksqlServerUrl.trim().startsWith("https");
```
vcrfxia, line 101: Let's lowercase ksqlServerUrl first to be safe.

vcrfxia (approved):
LGTM besides the comments inline. We should add an integration test for this, but we can do that as a follow-up. I'll file a ticket.

PR #7145, opened 2021-03-03
feat(migrations): implement `info` command

vcrfxia:
### Description 

This PR implements the `migrations info` command. The command does two things: prints the current migration version, and prints a table with information about the different available migration versions.

As part of implementing `info`, this PR also adds a new column (`error_reason`) to the migrations metadata stream and table, and contains some minor refactors such as renaming `Migration` to `MigrationFile`, renaming `VersionInfo` to `MigrationVersionInfo` and moving it into its own file, and cleaning up some test code. The refactors are in the first few commits and can be reviewed separately if that's easier.

### Testing done 

Unit + integration testing.

### Reviewer checklist


ksqldb-tools/src/test/java/io/confluent/ksql/tools/migrations/MigrationsTest.java
```diff
@@ -108,42 +134,74 @@ public void testApply() throws IOException {
...
+  }
 
+  private static void verifyMigrationsApplied() {
     // verify FOO and BAR were registered
     describeSource("FOO");
     describeSource("BAR");
 
-    // verify current
-    final List<StreamedRow> current = makeKsqlQuery("SELECT * FROM migration_schema_versions WHERE VERSION_KEY='CURRENT';");
-    assertThatEventually(() -> current.size(), is(2));
```
vcrfxia, line 126: These old `assertThatEventually` calls didn't actually do anything since `current` wasn't being refreshed (i.e., `makeKsqlQuery()` was only being called once, outside the `assertThatEventually`). I've gone through and cleaned up the usage of `assertThatEventually` in this test file.

ksqldb-tools/src/test/java/io/confluent/ksql/tools/migrations/MigrationsTest.java
```diff
@@ -108,42 +134,74 @@ public void testApply() throws IOException {
...
+
+      // Then:
+      assertThat(infoStatus, is(0));
+
+      verify(logAppender, atLeastOnce()).doAppend(logCaptor.capture());
+      final List<String> logMessages = logCaptor.getAllValues().stream()
+          .map(LoggingEvent::getRenderedMessage)
+          .collect(Collectors.toList());
+      assertThat(logMessages, hasItem(containsString("Current migration version: 2")));
+      assertThat(logMessages, hasItem(matchesRegex(
```
vcrfxia, line 162: This test uses a regex matcher since (1) the timestamp depends on when the test is run, and (2) the timezone used is local.

jzaralim (approved):

ksqldb-tools/src/main/java/io/confluent/ksql/tools/migrations/util/MigrationVersionInfo.java
```diff
@@ -0,0 +1,130 @@
...
+import io.confluent.ksql.api.client.Row;
+import io.confluent.ksql.tools.migrations.util.MetadataUtil.MigrationState;
+import java.text.SimpleDateFormat;
+import java.util.Date;
+import java.util.Objects;
+
+public final class MigrationVersionInfo {
+
+  private static final String DATE_FORMAT = "yyyy-MM-dd HH:mm:ss.SSS z";
+  private static final String EMPTY_MIGRATION_TIMESTAMP = "N/A";
```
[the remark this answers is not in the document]
vcrfxia: I don't feel strongly. I think we can leave it as is. In all likelihood, the only way users will view the contents of the metadata stream and table will be through `migrations info`, rather than inspecting the stream/table directly.

PR #7146, opened 2021-03-03
chore(migrations): add script to run migration tool

jzaralim:
### Description 
Script to run migration tool from command line as `ksql-migrations ...`

To test on docker, follow the instructions here:
https://github.com/confluentinc/ksql/tree/master/ksqldb-docker

And run the command `docker exec <container> ksql-migrations ...`

To build the tar, run this command:
`make VERSION=0.14.0 PACKAGE_TYPE=archive -f debian/Makefile archive`

If you're building on a Macbook, you might have to change `iPatch` to `IPatch` in debian/Makefile

### Reviewer checklist

### DIFF
--- a/bin/ksql-migrations
+++ b/bin/ksql-migrations
@@ -1,6 +1,6 @@
 #!/bin/bash
 
-# (Copyright) [2018 - 2018] Confluent, Inc.
+# (Copyright) [2021 - 2021] Confluent, Inc.
 
 : "${JAVA_HOME:=""}"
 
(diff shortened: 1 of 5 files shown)


bin/ksql-migrations
```diff
@@ -0,0 +1,43 @@
...
+done
+
+DIR="$base_dir/share/java/ksqldb"
+if [ -d "$DIR" ]; then
+  KSQL_CLASSPATH="$DIR/*:$KSQL_CLASSPATH"
+fi
+
+: "${KSQL_CONFIG_DIR:="$base_dir/config"}"
+
+exec "$JAVA" -cp "$KSQL_CLASSPATH" io.confluent.ksql.tools.migrations.Migrations "$@"
```
vcrfxia, line 43: The log4j config isn't customizable with this, right? We should make it customizable, in case users want to separate info and error logs, or for whatever reason can't use our default config.

bin/ksql-migrations
```diff
@@ -0,0 +1,43 @@
+#!/bin/bash
+
+# (Copyright) [2018 - 2018] Confluent, Inc.
```
vcrfxia, line 3: nit: year

vcrfxia (commented):
Thanks @jzaralim . How did you test this? Have you verified that the binary is available from the different methods of packaging (tarball, docker, etc)?

bin/ksql-migrations
jzaralim: Done. The user can customize the logger file by setting the environment variable `KSQL_LOG4J_OPTS`. If it's not present then the script will look for `log4j-console.properties`.

vcrfxia (approved):
LGTM. Thanks @jzaralim !

PR #7148, opened 2021-03-03
feat: add lambda feature flag

lct45:
### Description 
Adds a lambda feature flag so we can disable lambdas if anything awful happens after we ship them.

A bit of refactoring to pass in ksql config information all the way to AST sanitizer, could be used in the future if we want access to feature flag info in these classes
### Testing done 
unit test to make sure we throw the proper error

### Reviewer checklist


stevenpyzhang (approved):
lgtm, could you do a quick manual verification of behavior just to sanity check? 

Start up server without setting config to false
Issue a `Create Stream As` Query
Stop server
Set the config in the server properties
Start up server
Verify old query is still running
Verify new queries with lambdas can't be issued.

lct45:
Confirmed that the feature flag works

PR #7151, opened 2021-03-03
feat: expose TaskMetadata in QueryDescription

wcarlson5:
### Description 
Expose the TaskMetadata for each query in the description for the health check to use
### Testing done 

### Reviewer checklist


ksqldb-cli/src/test/java/io/confluent/ksql/cli/console/ConsoleTest.java
```diff
@@ -393,22 +394,28 @@ public void shouldPrintExplainQueryWithError() {
...
+                  Optional.empty())),
+          ImmutableSet.of("source"),
+          ImmutableSet.of("sink"),
+          "topology",
+          "executionPlan",
+          ImmutableMap.of("overridden.prop", 42),
+          ImmutableMap.of(new KsqlHostInfoEntity("foo", 123), KsqlQueryStatus.ERROR),
+          KsqlQueryType.PERSISTENT,
+          ImmutableList.of(new QueryError(timestamp, "error", Type.SYSTEM)),
+          ImmutableSet.of(
```
wcarlson5, line 413: The changes have not been merged into streams yet to use the new TaskMetadata. Also I have not updated the condition for this test so it will fail rn

ksqldb-rest-model/src/main/java/io/confluent/ksql/rest/entity/QueryDescription.java
```diff
@@ -67,7 +72,8 @@ public QueryDescription(
       @JsonProperty("ksqlHostQueryStatus") final Map<KsqlHostInfoEntity, KsqlQueryStatus>
           ksqlHostQueryStatus,
       @JsonProperty("queryType") final KsqlQueryType queryType,
-      @JsonProperty("queryErrors") final List<QueryError> queryErrors
+      @JsonProperty("queryErrors") final List<QueryError> queryErrors,
+      @JsonProperty("tasksMetadata") final Set<TaskMetadata> tasksMetadata
```
wcarlson5, line 76: It is having trouble Serializing `TaskMetadata` because it can't find the getters because of AK's naming conventions. I can fix this by adding `.setVisibility(PropertyAccessor.FIELD, JsonAutoDetect.Visibility.ANY)` either at the mapper or just in `Console.java` but either risks exposing private fields. WDYT @abbccdda @spena ?

wcarlson5:
The test failures are because the changes in https://github.com/apache/kafka/pull/10211 are not merged yet and the localThreadsMetadata can't be called in created

ksqldb-rest-model/src/main/java/io/confluent/ksql/rest/entity/QueryDescription.java
abbccdda: It looks like all other serialization methods are depending on Ksql internally defined classes. Maybe we could define a class like `StreamTaskMetadata` to translate the original `TaskMetadata` into a class having getters.

ksqldb-cli/src/test/java/io/confluent/ksql/cli/console/ConsoleTest.java
abbccdda: Could you rebase and update this PR?

wcarlson5: I don't think the sync job has finished. The changes are not showing up yet even after a rebase  and clean install

abbccdda: Cool, sounds good, so should we proceed with the PR or wait for the sync to be done?

ksqldb-rest-model/src/main/java/io/confluent/ksql/rest/entity/QueryDescription.java
rodesai: +1 for defining our own API model. Can you put this in `ksqldb-reset-model`?

ksqldb-common/src/main/java/io/confluent/ksql/util/StreamsTaskMetadata.java
```diff
@@ -0,0 +1,61 @@
...
+
+import com.google.common.collect.ImmutableMap;
+
+import java.util.Optional;
+import java.util.Set;
+
+import org.apache.kafka.common.TopicPartition;
+import org.apache.kafka.streams.processor.TaskMetadata;
+
+public class StreamsTaskMetadata {
```
rodesai, line 26: assuming this is the API model you'd need to put this in `ksqldb-rest-model`. You'd also need to add Jackson annotations to the constructor that specify which json properties map to which parameters. `QueryDescription` is a good model to follow.

ksqldb-common/src/main/java/io/confluent/ksql/util/StreamsTaskMetadata.java
```diff
@@ -0,0 +1,61 @@
...
+
+  private final String taskId;
+  private final Set<TopicPartition> topicPartitions;
+  private final ImmutableMap<TopicPartition, Long> endOffsets;
+  private final ImmutableMap<TopicPartition, Long> committedOffsets;
+  private final Optional<Long> timeCurrentIdlingStarted;
+
+  StreamsTaskMetadata(final TaskMetadata taskMetadata) {
```
rodesai, line 34: nit: we usually have the constructor take each of the properties as a parameter, and move things like this to a static factory method that creates an instance from a `TaskMetadata`, e.g.:

```
@JsonCreator
StreamsTaskMetadata(
    @JsonProperty("taskId") String taskId,
    ...
)

static StreamsTaskMetadata fromTaskMetadata(final TaskMetadata taskMetadata) {
    return new StreamsTaskMetadata(...);
}
```

ksqldb-common/src/main/java/io/confluent/ksql/util/StreamsTaskMetadata.java
```diff
@@ -0,0 +1,61 @@
...
+import java.util.Set;
+
+import org.apache.kafka.common.TopicPartition;
+import org.apache.kafka.streams.processor.TaskMetadata;
+
+public class StreamsTaskMetadata {
+
+  private final String taskId;
+  private final Set<TopicPartition> topicPartitions;
+  private final ImmutableMap<TopicPartition, Long> endOffsets;
```
rodesai, line 30: I don't think `TopicPartition` is the model we use in our API. To make sure things stay compatible and that we don't add unnecessary fields, convert these to `TopicPartitionEntity`

rodesai (commented):
Thanks @wcarlson5! Is there going to be a follow-on that gathers the current task status from all the nodes in the cluster?

ksqldb-cli/src/test/java/io/confluent/ksql/cli/console/ConsoleTest.java
wcarlson5: We can proceed, the only changes after the sync is that there will be a few fields added to `StreamsTaskMetadata` but they will only change the json out put. Everything else will be the same.

ksqldb-rest-app/src/test/java/io/confluent/ksql/rest/entity/QueryDescriptionFactoryTest.java
```diff
@@ -169,6 +170,7 @@ public void setUp() {
         0L,
         0L
     );
+    persistentQuery.initialize();
```
guozhangwang, line 173: For my understanding: why we need to call `initialize` now? It seems just building the topology / set exception handler?

guozhangwang (approved):
Just nit questions, otherwise lgtm.

ksqldb-rest-app/src/test/java/io/confluent/ksql/rest/entity/QueryDescriptionFactoryTest.java
wcarlson5: `queryStreams.localThreadsMetadata()` needed the KS object to be not null

ksqldb-common/src/main/java/io/confluent/ksql/util/StreamsTaskMetadata.java
wcarlson5: @rodesai I converted these, however there is a key deserializer. I am adding a  `TopicPartitionEntityDeserializer extends KeyDeserializer` to `TopicPartitionEntity`. Or is there some other way this is normally handled?

ksqldb-rest-model/src/main/java/io/confluent/ksql/rest/entity/TopicPartitionEntity.java
```diff
@@ -72,4 +76,19 @@ public String toString() {
...
   }
+
+  @JsonIgnoreProperties(ignoreUnknown = true)
+  public static class TopicPartitionEntityDeserializer extends KeyDeserializer {
+    @Override
+    public TopicPartitionEntity deserializeKey(
+        final String key,
+        final DeserializationContext ctxt
+    ) throws IOException {
+      final String[] split = key.split("=");
```
wcarlson5, line 87: If anyone has a fancy regex solution I would prefer that, this was just a quick workaround

wcarlson5 (commented):
I resolved the test failures.

@rodesai yes the gather will not be here so should be in a follow up

ksqldb-rest-model/src/main/java/io/confluent/ksql/rest/entity/StreamsTaskMetadata.java
```diff
@@ -0,0 +1,135 @@
...
+import java.util.Set;
+import java.util.stream.Collectors;
+
+import org.apache.kafka.streams.processor.TaskMetadata;
+
+@JsonIgnoreProperties(ignoreUnknown = true)
+public class StreamsTaskMetadata {
+
+  private final String taskId;
+  private final Set<TopicPartitionEntity> topicPartitions;
```
rodesai, line 34: what's the rationale for putting these in separate sets, as opposed to having the `TopicOffset` just have the end committed and end offsets and the TopicPartitionEntity? Then the client doesn't have to think about coherence between the two, and the content is smaller.

rodesai (commented):
Thanks @wcarlson5! This is looking good. One last comment/question inline.

ksqldb-rest-model/src/main/java/io/confluent/ksql/rest/entity/StreamsTaskMetadata.java
wcarlson5: The is a few cases where the TopicPartition won't have committed yet so I left them separate. We could have a zero default if that would be better

guozhangwang: I'm in favor of putting them into a single collection with defaults if not committed yet -- could be less vulnerable to bugs that result in inconsistent views.

ksqldb-rest-model/src/main/java/io/confluent/ksql/rest/entity/QueryDescription.java
```diff
@@ -83,6 +85,7 @@ public QueryDescription(
         new HashMap<>(Objects.requireNonNull(ksqlHostQueryStatus, "ksqlHostQueryStatus"));
     this.queryType = Objects.requireNonNull(queryType, "queryType");
     this.queryErrors = new ArrayList<>(Objects.requireNonNull(queryErrors, "queryErrors"));
+    this.tasksMetadata = ImmutableSet.copyOf(tasksMetadata);
```
spena, line 88:
```suggestion
    this.tasksMetadata = ImmutableSet.copyOf(Objects.requireNonNull(tasksMetadata, "tasksMetadata"));
```

spena (approved):
LGTM

ksqldb-rest-model/src/main/java/io/confluent/ksql/rest/entity/StreamsTaskMetadata.java
wcarlson5: I condensed them into one set that has the TopicPartition, endOffset and the committedOffset

wcarlson5:
Once the build pass I will merge. Thanks for the reviews @rodesai @spena @guozhangwang !

rodesai (approved):
LGTM

PR #7152, opened 2021-03-03
feat: Adds Lambda functionality to the interpreter

AlanConfluent:
### Description 
Adds lambda functionality to the interpreter used with pull queries.  Now you should be able to use expressions like `SELECT ID, REDUCE(ID, 0, (x,y) => x + y) as sum FROM TABLE;` or `SELECT ID, count, VAL_LIST FROM AGGREGATE WHERE REDUCE(VAL_LIST, 0, (x,y) => x + y) > 4` where lambdas exist in either the `SELECT` or `WHERE `clauses. 

### Testing done 
Wrote a bunch of unit tests and RQTT tests.

### Reviewer checklist

### DIFF
--- a/ksqldb-engine/src/main/java/io/confluent/ksql/planner/plan/PullFilterNode.java
+++ b/ksqldb-engine/src/main/java/io/confluent/ksql/planner/plan/PullFilterNode.java
@@ -395,13 +395,6 @@ public class PullFilterNode extends SingleSourcePlanNode {
             : (comp.getLeft() instanceof UnqualifiedColumnReferenceExp ? comp.getLeft() : null));
   }
 
-  private boolean isSingleColumnReference(final ComparisonExpression comp) {
-    return (comp.getRight() instanceof UnqualifiedColumnReferenceExp
-        && !(comp.getLeft() instanceof UnqualifiedColumnReferenceExp))
-            || (comp.getLeft() instanceof UnqualifiedColumnReferenceExp
-        && !(comp.getRight() instanceof UnqualifiedColumnReferenceExp));
-  }
-
   private Expression getNonColumnRefSide(final ComparisonExpression comparison) {
     return comparison.getRight() instanceof UnqualifiedColumnReferenceExp
         ? comparison.getLeft()
--- a/ksqldb-engine/src/main/java/io/confluent/ksql/planner/plan/PullFilterNode.java
+++ b/ksqldb-engine/src/main/java/io/confluent/ksql/planner/plan/PullFilterNode.java
@@ -395,6 +395,13 @@ public class PullFilterNode extends SingleSourcePlanNode {
             : (comp.getLeft() instanceof UnqualifiedColumnReferenceExp ? comp.getLeft() : null));
   }
 
+  private boolean isSingleColumnReference(final ComparisonExpression comp) {
+    return (comp.getRight() instanceof UnqualifiedColumnReferenceExp
+        && !(comp.getLeft() instanceof UnqualifiedColumnReferenceExp))
+            || (comp.getLeft() instanceof UnqualifiedColumnReferenceExp
+        && !(comp.getRight() instanceof UnqualifiedColumnReferenceExp));
+  }
+
   private Expression getNonColumnRefSide(final ComparisonExpression comparison) {
     return comparison.getRight() instanceof UnqualifiedColumnReferenceExp
         ? comparison.getLeft()
--- a/ksqldb-execution/src/main/java/io/confluent/ksql/execution/interpreter/TermEvaluationContext.java
+++ b/ksqldb-execution/src/main/java/io/confluent/ksql/execution/interpreter/TermEvaluationContext.java
@@ -59,7 +59,7 @@ public final class TermEvaluationContext {
   /**
    * Looks up a variable from the mappings made with {@code pushVariableMappings} that have not yet
    * been popped with {@code popVariableMappings}.
-   * @param name
+   * @param name The name of the variable to lookup
    * @return
    */
   public Object lookupVariable(final String name) {
--- a/ksqldb-execution/src/main/java/io/confluent/ksql/execution/interpreter/TermEvaluationContext.java
+++ b/ksqldb-execution/src/main/java/io/confluent/ksql/execution/interpreter/TermEvaluationContext.java
@@ -16,10 +16,14 @@
 package io.confluent.ksql.execution.interpreter;
 
 import io.confluent.ksql.GenericRow;
+import java.util.Deque;
+import java.util.LinkedList;
+import java.util.Map;
 
 public final class TermEvaluationContext {
 
   private final GenericRow row;
+  private final Deque<Map<String, Object>> deque = new LinkedList<>();
 
   public TermEvaluationContext(final GenericRow row) {
     this.row = row;
@@ -27,5 +31,22 @@ public final class TermEvaluationContext {
 
   public GenericRow getRow() {
     return row;
+  }
+
+  public void pushVariableMappings(final Map<String, Object> mappings) {
+    deque.push(mappings);
+  }
+
+  public void popVariableMappings() {
+    deque.pop();
+  }
+
+  public Object lookupVariable(final String name) {
+    for (Map<String, Object> mappings : deque) {
+      if (mappings.containsKey(name)) {
+        return mappings.get(name);
+      }
+    }
+    throw new IllegalStateException("Can't find lambda variable " + name);
   }
 }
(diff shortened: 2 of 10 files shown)


stevenpyzhang (commented):
Thanks @AlanConfluent , could you add some nested lambda tests/complex lambdas RQQT tests (also unit tests but not as necessary) just as a sanity check to make sure the integration is done properly? you can look at `complex-lambda.json` QTT for some inspiration (or just copy it over)

Also a test where we use different lambda variables when there's multiple lambdas in a single function

transform(map, (x,y) => ... , (k,v) => )

The complex nested lambdas and different variable names in same function call were some cases that we stumbled on early on in lambda implementation so we should have coverage for those.

ksqldb-execution/src/main/java/io/confluent/ksql/execution/interpreter/terms/LambdaFunctionTerms.java
```diff
@@ -0,0 +1,125 @@
...
+import java.util.function.Function;
+
+public class LambdaFunctionTerms {
+
+  public static class LambdaFunction1Term implements Term {
+
+    private final List<Pair<String, SqlType>> argNamesToTypes;
+    private final Term body;
+
+    public LambdaFunction1Term(final List<Pair<String, SqlType>> argNamesToTypes, final Term body) {
```
guozhangwang, line 35: Could we enforce the `body` to be `FunctionCallTerm`?

ksqldb-execution/src/main/java/io/confluent/ksql/execution/interpreter/TermEvaluationContext.java
```diff
@@ -16,10 +16,14 @@
...
 
 import io.confluent.ksql.GenericRow;
+import java.util.Deque;
+import java.util.LinkedList;
+import java.util.Map;
 
 public final class TermEvaluationContext {
 
   private final GenericRow row;
+  private final Deque<Map<String, Object>> deque = new LinkedList<>();
```
guozhangwang, line 26: nit: lambdaArgumentsMappingDeque?

ksqldb-execution/src/main/java/io/confluent/ksql/execution/interpreter/terms/LambdaFunctionTerms.java
```diff
@@ -0,0 +1,125 @@
...
+import io.confluent.ksql.execution.interpreter.TermEvaluationContext;
+import io.confluent.ksql.schema.ksql.types.SqlType;
+import io.confluent.ksql.util.Pair;
+import java.util.List;
+import java.util.function.BiFunction;
+import java.util.function.Function;
+
+public class LambdaFunctionTerms {
+
+  public static class LambdaFunction1Term implements Term {
```
guozhangwang, line 30: Could we use 

```
(Object[] args) -> { 
    ... // for loop to populate a hashmap, then
    context.pushVariableMappings(new ImmutableMap(map));
    ... 
}
``` 

to workaround the num.argument limitation? Also we would just have one term.

guozhangwang (commented):
Made a pass, mostly minor comments and just one meta question: I'm wondering if we can workaround a hard-coded num.argument limitation?

guozhangwang (commented):
For testing, could we add coverage for 1) lambdas in UDF and 2) lambdas in UDTF and 3) nested lambdas?

ksqldb-engine/src/main/java/io/confluent/ksql/planner/plan/PullFilterNode.java
```diff
@@ -351,6 +357,13 @@ private UnqualifiedColumnReferenceExp getColumnRefSide(final ComparisonExpressio
             ? comp.getRight() : comp.getLeft());
   }
 
+  private boolean isSingleColumnReference(final ComparisonExpression comp) {
+    return (comp.getRight() instanceof UnqualifiedColumnReferenceExp
+        && !(comp.getLeft() instanceof UnqualifiedColumnReferenceExp))
```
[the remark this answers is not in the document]
guozhangwang: NVM! Figured it out while looking at https://github.com/confluentinc/ksql/pull/7155 :)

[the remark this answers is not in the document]
AlanConfluent: After I merge that PR, this will go away :-)

ksqldb-execution/src/main/java/io/confluent/ksql/execution/interpreter/TermEvaluationContext.java
AlanConfluent: I ended up calling it `variableMappingsStack` since it's being used to mimic lambda call stacks.

ksqldb-execution/src/main/java/io/confluent/ksql/execution/interpreter/terms/LambdaFunctionTerms.java
AlanConfluent: I tried a few different things here, both this and `(Object... args)`.  Unfortunately, it requires both that a lambda implements some interface, and then that the parameters match it.  Since we're using three different interface on the consuming end (`Function`, `BiFunction`, `TriFunction`), we need to match their signatures.  Unfortunately, this is a limitation of that requirement and the code compiler does a similar thing.

ksqldb-execution/src/main/java/io/confluent/ksql/execution/interpreter/terms/LambdaFunctionTerms.java
AlanConfluent: The body is whatever is actually in the body of the lambda, which can be anything (e.g. `X+1` of the lambda definition `X => X+1`).  I think you're meaning the parent of this term, which is always a function call.  We don't actually keep parent references, and the children of a function call can be anything, so I don't think we can enforce types here.

ksqldb-execution/src/main/java/io/confluent/ksql/execution/interpreter/terms/LambdaFunctionTerms.java
```diff
@@ -0,0 +1,125 @@
...
+          return body.getValue(context);
+        } finally {
+          context.popVariableMappings();
+        }
+      };
+    }
+
+    @Override
+    public SqlType getSqlType() {
+      return null;
```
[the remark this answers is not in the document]
AlanConfluent: Done

ksqldb-execution/src/main/java/io/confluent/ksql/execution/interpreter/terms/LambdaFunctionTerms.java
AlanConfluent: I cleaned this class up with a common base class and some utility methods to try to make the child classes very minimal.

AlanConfluent:
> Thanks @AlanConfluent , could you add some nested lambda tests/complex lambdas RQQT tests (also unit tests but not as necessary) just as a sanity check to make sure the integration is done properly? you can look at `complex-lambda.json` QTT for some inspiration (or just copy over the same test case but for pull queries)
> 
> Also, could you add a test where we use different lambda variables when there's multiple lambdas in a single function
> transform(map, (x,y) => ... , (k,v) => )
> 
> The complex nested lambdas and different variable names in same function call were some cases that we stumbled on early on in lambda implementation so we should have coverage for those.

@stevenpyzhang , I just replicated your complex-lambda.json but for pull queries and they pass.  Also, I have some fairly complex unit tests which tests some of these same cases too.

stevenpyzhang:
@AlanConfluent thanks for adding the tests, I'll take another look today, that's really promising 👍

AlanConfluent:
@guozhangwang 

> Made a pass, mostly minor comments and just one meta question: I'm wondering if we can workaround a hard-coded num.argument limitation?

I don't think there is.  I explained in greater detail inline.  This same method is used for code generation.

> For testing, could we add coverage for 1) lambdas in UDF and 2) lambdas in UDTF and 3) nested lambdas?

I have many for 1) and 3).  For 2) I may have misspoken before about the support for UDTFs.  These aren't supported in evaluation since they aren't supported in pull queries.  Also, the current code generator `SqlToJavaVisitor` doesn't support them either.  From what I understand, both aggregation and table functions are invoked only by streams (at the moment!), so this is why.  If pull queries supports either in the future, we'll certainly implement support for this.

guozhangwang (approved):

stevenpyzhang (approved):
lgtm

PR #7155, opened 2021-03-04
feat: Allows lots of table scans cases where keys cannot easily be extracted

### LINKED ISSUE #4484: Support calculated predicates in pull queries
While fixed predicates work 

```
ksql> SELECT TIMESTAMPTOSTRING(WINDOWSTART, 'yyyy-MM-dd HH:mm:ss') AS WINDOW_START_TS, FULL_NAME,                                            >       RATINGS_COUNT
>  FROM RATINGS_PER_CUSTOMER_PER_MINUTE
>  WHERE ROWKEY='Rica Blaisdell'
>    AND WINDOWSTART > '2020-02-07T16:57:00.000';
+---------------------------------------------+---------------------------------------------+---------------------------------------------+
|WINDOW_START_TS                              |FULL_NAME                                    |RATINGS_COUNT                                |
+---------------------------------------------+---------------------------------------------+---------------------------------------------+
|2020-02-07 16:58:00                          |Rica Blaisdell                               |10                                           |
|2020-02-07 16:59:00                          |Rica Blaisdell                               |7                                            |
|2020-02-07 17:00:00                          |Rica Blaisdell                               |7                                            |
|2020-02-07 17:01:00                          |Rica Blaisdell                               |10                                           |
|2020-02-07 17:02:00                          |Rica Blaisdell                               |5                                            |
|2020-02-07 17:03:00                          |Rica Blaisdell                               |8                                            |
|2020-02-07 17:04:00                          |Rica Blaisdell                               |4                                            |
Query terminated
ksql> SELECT TIMESTAMPTOSTRING(WINDOWSTART, 'yyyy-MM-dd HH:mm:ss') AS WINDOW_START_TS, FULL_NAME, 
>                                                 RATINGS_COUNT
>  FROM RATINGS_PER_CUSTOMER_PER_MINUTE
>  WHERE ROWKEY='Rica Blaisdell'
>    AND WINDOWSTART > 1581008904345;
+---------------------------------------------+---------------------------------------------+---------------------------------------------+
|WINDOW_START_TS                              |FULL_NAME                                    |RATINGS_COUNT                                |
+---------------------------------------------+---------------------------------------------+---------------------------------------------+
|2020-02-07 16:35:00                          |Rica Blaisdell                               |4                                            |
|2020-02-07 16:36:00                          |Rica Blaisdell                               |7                                            |
|2020-02-07 16:37:00                          |Rica Blaisdell                               |8                                            |
|2020-02-07 16:38:00                          |Rica Blaisdell                               |6                                            |
|2020-02-07 16:39:00                          |Rica Blaisdell                               |9                                            |
```

Calculated predicate does n't 

```
ksql> SELECT TIMESTAMPTOSTRING(WINDOWSTART, 'yyyy-MM-dd HH:mm:ss') AS WINDOW_START_TS, FULL_NAME,
>       RATINGS_COUNT
>  FROM RATINGS_PER_CUSTOMER_PER_MINUTE
>  WHERE ROWKEY='Rica Blaisdell'
>    AND WINDOWSTART > (UNIX_TIMESTAMP() - 86400000);
WINDOWSTART bounds must be BIGINT.
ksql> SELECT TIMESTAMPTOSTRING(WINDOWSTART, 'yyyy-MM-dd HH:mm:ss') AS WINDOW_START_TS, FULL_NAME,
>       RATINGS_COUNT
>  FROM RATINGS_PER_CUSTOMER_PER_MINUTE
>  WHERE ROWKEY='Rica Blaisdell'
>    AND WINDOWSTART > CAST((UNIX_TIMESTAMP() - 86400000) AS BIGINT);
WINDOWSTART bounds must be BIGINT.
```
Thanks @rmoff for reporting this!

AlanConfluent:
fixes #4484 

### Description 
This allows many cases when table scans are enabled.  Prior to this, a comparison must have been of the form `KEY=12345` or possibly `KEY=1 + 2` where there was a column on one side and a resolvable expression on the other.  When table scans are enabled, now these can be done:
- Both sides are column references (e.g. `KEY = COL`)
- Neither side is a column reference (e.g. `5 > 3`, `KEY + 1 = COL + 1` )
- One side is a column reference and the other is an unresolveable expression (e.g. `KEY = COL + 10`)

Resolves #4484 

### Testing done 
RQTT tests and unit tests

### Reviewer checklist

### DIFF
--- a/ksqldb-engine/src/main/java/io/confluent/ksql/engine/generic/GenericRecordFactory.java
+++ b/ksqldb-engine/src/main/java/io/confluent/ksql/engine/generic/GenericRecordFactory.java
@@ -146,7 +146,8 @@ public class GenericRecordFactory {
           column,
           functionRegistry,
           config,
-          "insert value").resolve(valueExp);
+          "insert value",
+          false).resolve(valueExp);
 
       values.put(column, value);
     }
--- a/ksqldb-engine/src/main/java/io/confluent/ksql/planner/plan/PullFilterNode.java
+++ b/ksqldb-engine/src/main/java/io/confluent/ksql/planner/plan/PullFilterNode.java
@@ -31,6 +31,7 @@ import io.confluent.ksql.execution.expression.tree.Literal;
 import io.confluent.ksql.execution.expression.tree.LogicalBinaryExpression;
 import io.confluent.ksql.execution.expression.tree.LongLiteral;
 import io.confluent.ksql.execution.expression.tree.NullLiteral;
+import io.confluent.ksql.execution.expression.tree.QualifiedColumnReferenceExp;
 import io.confluent.ksql.execution.expression.tree.StringLiteral;
 import io.confluent.ksql.execution.expression.tree.TraversalExpressionVisitor;
 import io.confluent.ksql.execution.expression.tree.UnqualifiedColumnReferenceExp;
@@ -302,7 +303,17 @@ public class PullFilterNode extends SingleSourcePlanNode {
                 isWindowed));
         return null;
       }
+
       final UnqualifiedColumnReferenceExp column = getColumnRefSide(node);
+      final Expression other = getNonColumnRefSide(node);
+      final NonColumnRefValidator nonColumnRefValidator = new NonColumnRefValidator();
+      nonColumnRefValidator.process(other, null);
+
+      if (nonColumnRefValidator.isNonColumnRefUnresolvable()) {
+        setTableScanOrElseThrow(() ->
+            invalidWhereClauseException("Non column reference must be resolvable", isWindowed));
+        return null;
+      }
 
       final ColumnName columnName = column.getColumnName();
       if (columnName.equals(SystemColumns.WINDOWSTART_NAME)
@@ -349,6 +360,28 @@ public class PullFilterNode extends SingleSourcePlanNode {
       } else {
         throw exceptionSupplier.get();
       }
+    }
+  }
+
+  private final class NonColumnRefValidator extends TraversalExpressionVisitor<Object> {
+
+    private boolean nonColumnRefUnresolvable;
+
+    public NonColumnRefValidator() {
+      nonColumnRefUnresolvable = false;
+    }
+
+    @Override
+    public Void visitUnqualifiedColumnReference(
+        final UnqualifiedColumnReferenceExp node,
+        final Object context
+    ) {
+      nonColumnRefUnresolvable = true;
+      return null;
+    }
+
+    public boolean isNonColumnRefUnresolvable() {
+      return nonColumnRefUnresolvable;
     }
   }
 
@@ -438,7 +471,8 @@ public class PullFilterNode extends SingleSourcePlanNode {
             keyColumn.name(),
             metaStore,
             config,
-            "pull query"
+            "pull query",
+            ksqlConfig.getBoolean(KsqlConfig.KSQL_QUERY_PULL_INTERPRETER_ENABLED)
         ).resolve(exp);
       }
 
--- a/ksqldb-engine/src/main/java/io/confluent/ksql/planner/plan/PullFilterNode.java
+++ b/ksqldb-engine/src/main/java/io/confluent/ksql/planner/plan/PullFilterNode.java
@@ -31,7 +31,6 @@ import io.confluent.ksql.execution.expression.tree.Literal;
 import io.confluent.ksql.execution.expression.tree.LogicalBinaryExpression;
 import io.confluent.ksql.execution.expression.tree.LongLiteral;
 import io.confluent.ksql.execution.expression.tree.NullLiteral;
-import io.confluent.ksql.execution.expression.tree.QualifiedColumnReferenceExp;
 import io.confluent.ksql.execution.expression.tree.StringLiteral;
 import io.confluent.ksql.execution.expression.tree.TraversalExpressionVisitor;
 import io.confluent.ksql.execution.expression.tree.UnqualifiedColumnReferenceExp;
@@ -45,6 +44,7 @@ import io.confluent.ksql.schema.ksql.Column.Namespace;
 import io.confluent.ksql.schema.ksql.DefaultSqlValueCoercer;
 import io.confluent.ksql.schema.ksql.LogicalSchema;
 import io.confluent.ksql.schema.ksql.SystemColumns;
+import io.confluent.ksql.schema.ksql.types.SqlTypes;
 import io.confluent.ksql.schema.utils.FormatOptions;
 import io.confluent.ksql.structured.SchemaKStream;
 import io.confluent.ksql.util.KsqlConfig;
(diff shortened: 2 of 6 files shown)


ksqldb-engine/src/main/java/io/confluent/ksql/engine/generic/GenericRecordFactory.java
```diff
@@ -146,7 +146,8 @@ private static LogicalSchema withRowTime(final LogicalSchema schema) {
           column,
           functionRegistry,
           config,
-          "insert value").resolve(valueExp);
+          "insert value",
+          false).resolve(valueExp);
```
agavra, line 150: i believe this is used on the pull query path as well as the insert into path, in which case should we be using the interpreter? might be worth confirming the code flow

ksqldb-engine/src/main/java/io/confluent/ksql/planner/plan/PullFilterNode.java
```diff
@@ -345,6 +363,28 @@ private void setTableScanOrElseThrow(final Supplier<KsqlException> exceptionSupp
     }
   }
 
+  private final class NonColumnRefValidator extends TraversalExpressionVisitor<Object> {
```
agavra, line 366:
```suggestion
  private final class HasColumnRef extends TraversalExpressionVisitor<Object> {
```
It was really tough to grok `nonColumnRefUnresolvable` - it looks like what this does is answer "did we visit a column reference?"

ksqldb-functional-tests/src/test/resources/rest-query-validation-tests/pull-queries-against-materialized-aggregates.json
```diff
@@ -2060,6 +2060,171 @@
...
+          {"row":{"columns":[11, 12000, 13000, 1, 59874.0]}}
+        ]}
+      ]
+    },
+    {
+      "name": "non-windowed - Table Scan - non column ref is not literal",
+      "statements": [
+        "CREATE STREAM INPUT (ID INTEGER KEY, IGNORED INT) WITH (kafka_topic='test_topic', value_format='JSON');",
+        "CREATE TABLE AGGREGATE AS SELECT ID, COUNT(1) AS COUNT FROM INPUT GROUP BY ID;",
+        "SELECT * FROM AGGREGATE WHERE ID = 20 - 10;",
```
agavra, line 2146:
```sql
SELECT * FROM AGGREGATE WHERE ID = 20 - 10;
```

at some point, this should result in a Key lookup right? What's the status of #4484 after we merge this PR? We might want to consider renaming that one (or opening a new one). I think it's important to consider that alongside this PR because we'd probably want a single-pass strategy (naively, we could just use the GenericExpressionResolver to try to resolve the non-key side of the equation and if it throws we assume it's not resolvable).

agavra (approved):
LGTM! Nice quick win in usability

ksqldb-functional-tests/src/test/resources/rest-query-validation-tests/pull-queries-against-materialized-aggregates.json
AlanConfluent: Yes, this should result in a key lookup.  It should fall into the case of non column ref needs to be resolved.

I think #4484 should be resolved after this.

I was debating whether to do it that way, and I'm not opposed to doing it like that.  The current code is a common pattern where there's a validation and an extraction pass.  In this case, it makes dealing with the consequences a bit easier (we can just declare table scan required with the other similar checks and avoid extraction). Doing some of this in extraction would just slightly complicate that code a bit. Or did you mean merging Validation and Extraction more broadly to a single pass?

ksqldb-engine/src/main/java/io/confluent/ksql/engine/generic/GenericRecordFactory.java
agavra: thanks! I must've confused the two

ksqldb-functional-tests/src/test/resources/rest-query-validation-tests/pull-queries-against-materialized-aggregates.json
agavra: Ah ignore that original comment, I hadn't properly understood the code on my first pass. Originally I had thought that we would table scan for `ID = 20 - 10` but now I'm seeing that we don't. I think the current implementation is fine.

ksqldb-engine/src/main/java/io/confluent/ksql/planner/plan/PullFilterNode.java
```diff
@@ -564,8 +613,21 @@ private Instant asInstant(final Expression other) {
...
+      final Object obj = new GenericExpressionResolver(
+          SqlTypes.BIGINT,
+          name,
+          metaStore,
+          ksqlConfig,
+          "pull query window bounds extractor",
+          ksqlConfig.getBoolean(KsqlConfig.KSQL_QUERY_PULL_INTERPRETER_ENABLED)
+      ).resolve(other);
+
+      if (obj instanceof Integer || obj instanceof Long) {
```
guozhangwang, line 625: Should we allow float/double to be casted to INT as well?

ksqldb-engine/src/main/java/io/confluent/ksql/planner/plan/PullFilterNode.java
```diff
@@ -295,7 +297,23 @@ public Void visitComparisonExpression(
...
     ) {
+      if (!isSingleColumnReference(node)) {
+        setTableScanOrElseThrow(() ->
+            invalidWhereClauseException("A comparison must reference a key column and literal",
+                isWindowed));
+        return null;
+      }
+
       final UnqualifiedColumnReferenceExp column = getColumnRefSide(node);
+      final Expression other = getNonColumnRefSide(node);
```
guozhangwang, line 308: nit: I'm wondering if we can merge these two conditions into a single one, like:

```
first = getFirstColumnRefSide(); // if both are column ref, return left; if non are column ref, return null;
if (first != null) {
    second = getOtherSide();
    secondHasColumnRef = new ... ; secondHasColumnRef.process(second);
    if (second isColumnRef || secondHasColumnRef.hasColumnRef)
        setTableScan();
} else {
    setTableScan();    // i.e. for `WHERE 100 = 100` or `100 = 101`, we would always do table scan
}
```

ksqldb-engine/src/main/java/io/confluent/ksql/planner/plan/PullFilterNode.java
```diff
@@ -295,7 +297,23 @@ public Void visitComparisonExpression(
         final ComparisonExpression node,
         final Object context
     ) {
+      if (!isSingleColumnReference(node)) {
```
guozhangwang, line 300: Just occurred to me: if the `WHERE` clause is just ` 100 = 101` then we would still do a table scan instead of immediately return empty values?

guozhangwang (commented):
Made a pass, overall LGTM!

ksqldb-engine/src/main/java/io/confluent/ksql/planner/plan/PullFilterNode.java
AlanConfluent: Sure, I reoriented the logic similar to this and allowed getFirstColumnRefSide to return null so that it can be checked for, allowing me to get rid of the other method.

ksqldb-engine/src/main/java/io/confluent/ksql/planner/plan/PullFilterNode.java
AlanConfluent: The GenericExpressionResolver is in strict mode which I don't believe allows this.  We don't do a lot of conversions elsewhere automatically, such as with the key values (and would be a bit harder to generify), so I think this if fine for consistency.

guozhangwang (approved):

PR #7159, opened 2021-03-04
fix: return null on null input for timestamp math functions

jzaralim:
### Description 
Checks function inputs for null values

## Testing
QTT and unit tests

### Reviewer checklist

### DIFF
--- a/ksqldb-functional-tests/src/test/resources/query-validation-tests/timestampsub.json
+++ b/ksqldb-functional-tests/src/test/resources/query-validation-tests/timestampsub.json
@@ -41,6 +41,23 @@
         {"topic": "TEST2", "key": 0, "value": {"VALUE": 5}},
         {"topic": "TEST2", "key": 0, "value": {"VALUE": -95}}
       ]
+    },
+    {
+      "name": "handles null values",
+      "statements": [
+        "CREATE STREAM TEST (ID INT KEY, time TIMESTAMP, num INTEGER) WITH (kafka_topic='test', value_format='AVRO');",
+        "CREATE STREAM TEST2 AS SELECT id, timestampsub(MILLISECONDS, num, time) AS VALUE FROM TEST;"
+      ],
+      "inputs": [
+        {"topic": "test", "key": 0, "value": {"time": null, "num": 5}},
+        {"topic": "test", "key": 0, "value": {"time": 5, "num": null}},
+        {"topic": "test", "key": 0, "value": {"time": null, "num": null}}
+      ],
+      "outputs": [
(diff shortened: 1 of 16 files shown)


ksqldb-functional-tests/src/test/resources/query-validation-tests/transform-array.json
```diff
@@ -93,6 +93,19 @@
...
       ]
+    },
+    {
+      "name": "apply transform lambda function to array of timestamps",
+      "statements": [
+        "CREATE STREAM test (ID STRING KEY, times ARRAY<TIMESTAMP>) WITH (kafka_topic='test_topic', value_format='JSON');",
+        "CREATE STREAM OUTPUT AS SELECT ID, TRANSFORM(times, x => timestampadd(milliseconds, 5, x)) AS c FROM test;"
+      ],
+      "inputs": [
+        {"topic": "test_topic", "key": "one", "value": {"times": [3, 6]}}
```
stevenpyzhang, line 104: could we add one more input record where there's a null in the array? I'm curious what the behavior would be, I'd expect the output to just have a null at the same array index but I suspect the entire record value would be null

jzaralim: Added a fix to this PR.

ksqldb-functional-tests/src/test/resources/query-validation-tests/timestampsub.json
```diff
@@ -41,6 +41,23 @@
         {"topic": "TEST2", "key": 0, "value": {"VALUE": 5}},
         {"topic": "TEST2", "key": 0, "value": {"VALUE": -95}}
       ]
+    },
+    {
+      "name": "handles null values",
+      "statements": [
+        "CREATE STREAM TEST (ID INT KEY, time TIMESTAMP, num INTEGER) WITH (kafka_topic='test', value_format='AVRO');",
+        "CREATE STREAM TEST2 AS SELECT id, timestampsub(MILLISECONDS, num, time) AS VALUE FROM TEST;"
```
stevenpyzhang, line 49: Not too familiar with the new timestamp code, is milliseconds a reserved keyword in the syntax now for use cases like these?

stevenpyzhang (approved):
lgtm with build pass

ksqldb-functional-tests/src/test/resources/query-validation-tests/timestampsub.json
jzaralim: yup, its the same set of keywords that window expressions use

PR #7161, opened 2021-03-05
feat(migrations): enable insert values and create connector for migrations tool

jzaralim:
### Description
Adds logic to decide where to send a command, instead of sending them all to `Client.executeStatement`

After the migration file gets split up into separate commands, the command matches against different regex patterns.

If the command matches `INSERT INTO ... VALUES ...` then parse it with the KsqlParser to evaluate the column names and values.

If the command matches `CREATE SINK|SOURCE CONNECTOR ...` then it's send to the old rest client. In the future, this will also support `DROP CONNECTOR;`.

Otherwise, it's sent to `Client.executeStatement`.

### Testing done 
Added some unit and integration tests. Nothing too complicated though.

### Reviewer checklist

### DIFF
--- a/ksqldb-tools/src/main/java/io/confluent/ksql/tools/migrations/commands/ApplyMigrationCommand.java
+++ b/ksqldb-tools/src/main/java/io/confluent/ksql/tools/migrations/commands/ApplyMigrationCommand.java
@@ -24,8 +24,15 @@ import com.github.rvesse.airline.annotations.Option;
 import com.github.rvesse.airline.annotations.restrictions.RequireOnlyOne;
 import com.google.common.annotations.VisibleForTesting;
 import io.confluent.ksql.api.client.Client;
+import io.confluent.ksql.api.client.FieldInfo;
+import io.confluent.ksql.api.client.KsqlObject;
+import io.confluent.ksql.execution.expression.tree.Expression;
 import io.confluent.ksql.tools.migrations.MigrationConfig;
 import io.confluent.ksql.tools.migrations.MigrationException;
+import io.confluent.ksql.tools.migrations.util.CommandParser;
+import io.confluent.ksql.tools.migrations.util.CommandParser.SqlCommand;
+import io.confluent.ksql.tools.migrations.util.CommandParser.SqlInsertValues;
+import io.confluent.ksql.tools.migrations.util.CommandParser.SqlStatement;
 import io.confluent.ksql.tools.migrations.util.MetadataUtil;
 import io.confluent.ksql.tools.migrations.util.MetadataUtil.MigrationState;
 import io.confluent.ksql.tools.migrations.util.MigrationFile;
@@ -34,9 +41,10 @@ import io.confluent.ksql.tools.migrations.util.MigrationsUtil;
 import io.confluent.ksql.util.KsqlException;
 import io.confluent.ksql.util.RetryUtil;
 import java.time.Clock;
-import java.util.Arrays;
 import java.util.Collections;
+import java.util.HashMap;
 import java.util.List;
+import java.util.Map;
 import java.util.Optional;
 import java.util.concurrent.ExecutionException;
 import java.util.function.Function;
@@ -274,17 +282,31 @@ public class ApplyMigrationCommand extends BaseCommand {
       return false;
     }
 
-    final List<String> commands = Arrays.stream(migrationFileContent.split(";"))
-        .map(String::trim)
-        .filter(s -> !s.isEmpty())
-        .map(s -> s + ";")
-        .collect(Collectors.toList());
-    for (final String command : commands) {
+    final List<SqlCommand> commands;
+    try {
+      commands = CommandParser.parse(migrationFileContent);
+    } catch (MigrationException e) {
+      LOGGER.error(e.getMessage());
+      return false;
+    }
+
+    for (final SqlCommand command : commands) {
       try {
-        ksqlClient.executeStatement(command).get();
+        if (command instanceof SqlStatement) {
+          ksqlClient.executeStatement(command.getCommand()).get();
+        } else if (command instanceof SqlInsertValues) {
+          final List<FieldInfo> fields =
+              ksqlClient.describeSource(((SqlInsertValues) command).getSourceName()).get().fields();
+          ksqlClient.insertInto(
+              ((SqlInsertValues) command).getSourceName(),
+              getRow(
+                  fields,
+                  ((SqlInsertValues) command).getColumns(),
+                  ((SqlInsertValues) command).getValues())).get();
+        }
       } catch (InterruptedException | ExecutionException e) {
         final String errorMsg = String.format(
-            "Failed to execute sql: %s. Error: %s", command, e.getMessage());
+            "Failed to execute sql: %s. Error: %s", command.getCommand(), e.getMessage());
         LOGGER.error(errorMsg);
         updateState(config, ksqlClient, MigrationState.ERROR,
             executionStart, migration, clock, previous, Optional.of(errorMsg));
@@ -296,6 +318,44 @@ public class ApplyMigrationCommand extends BaseCommand {
         executionStart, migration, clock, previous, Optional.empty());
     LOGGER.info("Successfully migrated");
     return true;
+  }
+
+  private KsqlObject getRow(
+      final List<FieldInfo> sourceFields,
+      final List<String> insertColumns,
+      final List<Expression> insertValues
+  ) {
+    final Map<String, Object> row = new HashMap<>();
+    if (insertColumns.size() > 0) {
+      if (insertColumns.size() != insertValues.size()) {
+        throw new MigrationException("mismatch");
+      }
+      for (int i = 0 ; i < insertColumns.size(); i++) {
+        final String columnName = insertColumns.get(i);
+        final List<FieldInfo> matchingFields = sourceFields.stream()
+            .filter(f -> f.name().equals(columnName)).collect(Collectors.toList());
+        if (matchingFields.size() == 0) {
+          throw new MigrationException("Could not find column named " + columnName);
+        } else if (matchingFields.size() > 1) {
+          throw new MigrationException("Found multiple columns named " + columnName);
+        } else {
+          row.put(
+              columnName,
+              CommandParser.toFieldType(
+                  insertValues.get(i), matchingFields.get(0).type().getType())
+          );
+        }
+      }
+    } else {
+      for (int i = 0 ; i < sourceFields.size(); i++) {
+        row.put(
+            sourceFields.get(i).name(),
+            CommandParser.toFieldType(insertValues.get(i), sourceFields.get(i).type().getType())
+        );
+      }
+    }
+
+    return new KsqlObject(row);
   }
 
   private boolean verifyMigrated(
--- a/ksqldb-tools/src/main/java/io/confluent/ksql/tools/migrations/commands/ApplyMigrationCommand.java
+++ b/ksqldb-tools/src/main/java/io/confluent/ksql/tools/migrations/commands/ApplyMigrationCommand.java
@@ -27,10 +27,12 @@ import io.confluent.ksql.api.client.Client;
 import io.confluent.ksql.api.client.FieldInfo;
 import io.confluent.ksql.api.client.KsqlObject;
 import io.confluent.ksql.execution.expression.tree.Expression;
+import io.confluent.ksql.rest.client.KsqlRestClient;
 import io.confluent.ksql.tools.migrations.MigrationConfig;
 import io.confluent.ksql.tools.migrations.MigrationException;
 import io.confluent.ksql.tools.migrations.util.CommandParser;
 import io.confluent.ksql.tools.migrations.util.CommandParser.SqlCommand;
+import io.confluent.ksql.tools.migrations.util.CommandParser.SqlConnectorStatement;
 import io.confluent.ksql.tools.migrations.util.CommandParser.SqlInsertValues;
 import io.confluent.ksql.tools.migrations.util.CommandParser.SqlStatement;
 import io.confluent.ksql.tools.migrations.util.MetadataUtil;
@@ -124,6 +126,7 @@ public class ApplyMigrationCommand extends BaseCommand {
     return command(
         config,
         MigrationsUtil::getKsqlClient,
+        MigrationsUtil::createRestClient,
         getMigrationsDirFromConfigFile(configFile),
         Clock.systemDefaultZone()
     );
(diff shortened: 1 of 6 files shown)


ksqldb-tools/src/main/java/io/confluent/ksql/tools/migrations/util/CommandParser.java
```diff
@@ -0,0 +1,209 @@
...
+      case MAP:
+        final Map<Object, Object> resolvedMap = new HashMap<>();
+        ((CreateMapExpression) expressionValue).getMap()
+            .forEach((k, v) -> resolvedMap.put(getPrimitiveValue(k), getPrimitiveValue(v)));
+        return resolvedMap;
+      case STRUCT:
+        final Map<Object, Object> resolvedStruct = new HashMap<>();
+        ((CreateStructExpression) expressionValue)
+            .getFields().stream().forEach(
+                field -> resolvedStruct.put(field.getName(), getPrimitiveValue(field.getValue())));
```
jzaralim, line 115: Complex nested types are not supported

ksqldb-tools/src/main/java/io/confluent/ksql/tools/migrations/util/CommandParser.java
```diff
@@ -0,0 +1,209 @@
...
+import java.util.Map;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+import java.util.stream.Collectors;
+import javax.ws.rs.NotSupportedException;
+
+public class CommandParser {
+  private static final Pattern STRING_WS_SEMICOLON =
+      Pattern.compile("('([^']*|(''))*')|;", Pattern.DOTALL);
+  private static final Pattern INSERT_VALUES_PATTERN = Pattern.compile(
```
vcrfxia, line 47: Does this work if there are appearances of parens or the string `VALUES` within quotes within the statement? I think it should be fine but regexes are notoriously hard to reason about. Might be nice to add quick unit tests just for this pattern to cover some of these edge cases.

vcrfxia: This regex doesn't look like it works with quoted stream names (both backticks and double quotes are supported). I think the more reliable thing to do is to chop up the statement to remove quoted strings (similar to the logic above) and then check for the tokens (minus whitespace) `INSERT` and `INTO` at the front and `VALUES` appearing anywhere else in the statement, rather than relying on a single regex.

ksqldb-tools/src/test/java/io/confluent/ksql/tools/migrations/MigrationsTest.java
```diff
@@ -229,6 +234,16 @@ private static void verifyMigrationsApplied() {
     assertThat(current.get(1).getRow().get().getColumns().get(2), is("bar bar BAR"));
     assertThat(current.get(1).getRow().get().getColumns().get(3), is("MIGRATED"));
     assertThat(current.get(1).getRow().get().getColumns().get(7), is("1"));
+
+    // verify foo
+    final List<StreamedRow> foo = assertThatEventually(
+        () -> makeKsqlQuery("SELECT * FROM FOO EMIT CHANGES LIMIT 2;"),
+        hasSize(4));
```
vcrfxia, line 241: nit: add a comment explaining why size 4. Is it because there's a header and a footer?

ksqldb-tools/src/test/java/io/confluent/ksql/tools/migrations/util/CommandParserTest.java
```diff
@@ -0,0 +1,106 @@
...
+        + "    \"mode\"='bulk',\n"
+        + "    \"topic.prefix\"='jdbc-',\n"
+        + "    \"table.whitelist\"='users',\n"
+        + "    \"key\"='username');";
+    List<SqlCommand> commands = CommandParser.parse(command);
+    assertThat(commands.size(), is(1));
+    assertThat(commands.get(0), instanceOf(SqlConnectorStatement.class));
+    assertThat(commands.get(0).getCommand(), is(command));
+  }
+}
```
vcrfxia, line 106: We can do this as a follow-up PR but it'd be good to add more unit test coverage here for different types of statements: create sink connector, different types of insert values statements (with and without explicit field names, with quoted field names, with quoted stream name), insert into ... as select (to ensure it's not confused with insert values), unsupported statement types (to ensure a sane error message is returned), etc.

vcrfxia (commented):
Thanks @jzaralim -- this is really great!! Some thoughts inline. Most can be done as follow-ups.

ksqldb-tools/src/main/java/io/confluent/ksql/tools/migrations/util/MigrationsUtil.java
```diff
@@ -113,4 +116,13 @@ static ClientOptions createClientOptions(
     }
     return options;
   }
+
+  public static KsqlRestClient createRestClient(final MigrationConfig config) {
+    return KsqlRestClient.create(
```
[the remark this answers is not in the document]
jzaralim: Will defer this to a follow-up.

ksqldb-tools/src/main/java/io/confluent/ksql/tools/migrations/util/CommandParser.java
jzaralim: This is now supported

ksqldb-tools/src/main/java/io/confluent/ksql/tools/migrations/commands/ApplyMigrationCommand.java
```diff
@@ -298,6 +331,44 @@ private boolean applyMigration(
...
+  ) {
+    final Map<String, Object> row = new HashMap<>();
+    if (insertColumns.size() > 0) {
+      if (insertColumns.size() != insertValues.size()) {
+        throw new MigrationException("mismatch");
+      }
+      for (int i = 0 ; i < insertColumns.size(); i++) {
+        final String columnName = insertColumns.get(i);
+        final List<FieldInfo> matchingFields = sourceFields.stream()
+            .filter(f -> f.name().equals(columnName)).collect(Collectors.toList());
```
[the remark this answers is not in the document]
jzaralim: Originally the `sourceFields` were providing the type information, but that's unnecessary since the value expressions contain that information. I updated the PR so that we don't need to check anymore.  This also resolves the nested complex types problem.

ksqldb-tools/src/main/java/io/confluent/ksql/tools/migrations/util/CommandParser.java
```diff
@@ -0,0 +1,228 @@
...
+  private static final String INSERT = "INSERT";
+  private static final String INTO = "INTO";
+  private static final String VALUES = "VALUES";
+  private static final String SELECT = "SELECT";
+  private static final String CREATE = "CREATE";
+  private static final String SINK = "SINK";
+  private static final String SOURCE = "SOURCE";
+  private static final String CONNECTOR = "CONNECTOR";
+  private static final String SHORT_COMMENT_OPENER = "--";
+  private static final String SHORT_COMMENT_CLOSER = "\n";
```
vcrfxia, line 53: Do we have integration tests / have we run manual tests with files with newlines? It'd be good to sanity-check that carriage returns don't pose a problem.

ksqldb-tools/src/main/java/io/confluent/ksql/tools/migrations/util/CommandParser.java
```diff
@@ -0,0 +1,228 @@
...
+  *
+  * @return a list of strings
+  * */
+  @VisibleForTesting
+  static List<String> splitSql(final String sql) {
+    final List<String> commands = new ArrayList<>();
+    String current = "";
+    for (int i = 0; i < sql.length(); i++) {
+      if (sql.charAt(i) == SINGLE_QUOTE) {
+        final int closingToken = sql.indexOf(SINGLE_QUOTE, i + 1);
```
vcrfxia, line 79: What happens if the sql string is invalid and doesn't contain a closing quote to go with the opening quote? Same question for malformed comments.

vcrfxia: Also a note for anyone else looking at this: this doesn't explicitly handle escaped quotes but that's fine since quotes are pieced back together in the final string anyway. Could be good to add this as a comment in the code itself so future readers don't wonder the same thing.

ksqldb-tools/src/main/java/io/confluent/ksql/tools/migrations/util/CommandParser.java
```diff
@@ -0,0 +1,228 @@
...
+        current += sql.substring(i, closingToken + 1);
+        i = closingToken;
+      } else if (i < sql.length() - 1 && sql.startsWith(SHORT_COMMENT_OPENER, i)) {
+        i = sql.indexOf(SHORT_COMMENT_CLOSER, i + 1);
+      } else if (i < sql.length() - 1 && sql.startsWith(LONG_COMMENT_OPENER, i)) {
+        i = sql.indexOf(LONG_COMMENT_CLOSER, i + 1) + 1;
```
vcrfxia, line 85: Manipulating `i` inside the for-loop feels error-prone, and also requires reasoning about offsets like this. (This hard-coded `+ 1` should more accurately be `+ LONG_COMMENT_CLOSER.length - 1`, and similarly for the others.) Simplify the code by using a while loop instead?

ksqldb-tools/src/main/java/io/confluent/ksql/tools/migrations/util/CommandParser.java
```diff
@@ -0,0 +1,228 @@
...
+  * Determines the type of command a sql string is and returns a SqlCommand.
+  **/
+  private static SqlCommand transformToSqlCommand(final String sql) {
+    final List<String> tokens = Arrays
+        .stream(sql.toUpperCase().split(QUOTED_STRING_OR_WHITESPACE))
+        .filter(s -> !s.isEmpty())
+        .collect(Collectors.toList());
+    if (tokens.get(0).equals(INSERT)
+        && tokens.get(1).equals(INTO)
+        && !tokens.get(3).equals(SELECT)
```
vcrfxia, line 145: What is this line for? If we're trying to rule out `INSERT INTO ... AS SELECT`, we should be checking for the token with index 4 (not 3), right?

ksqldb-tools/src/main/java/io/confluent/ksql/tools/migrations/util/CommandParser.java
```diff
@@ -0,0 +1,228 @@
...
+      return new SqlInsertValues(
+          sql,
+          parsedStatement.getTarget().text(),
+          parsedStatement.getValues(),
+          parsedStatement.getColumns().stream()
+              .map(name -> name.text()).collect(Collectors.toList()));
+    } else if (tokens.get(0).equals(CREATE)
+        && (tokens.get(1).equals(SINK) || tokens.get(1).equals(SOURCE))
+        && tokens.get(2).equals(CONNECTOR)
+    ) {
```
vcrfxia, line 159: What about `DROP CONNECTOR` statements? We should support those too.

ksqldb-tools/src/main/java/io/confluent/ksql/tools/migrations/util/CommandParser.java
```diff
@@ -0,0 +1,228 @@
...
+      } else {
+        current += sql.charAt(i);
+      }
+    }
+
+    return commands;
+  }
+
+  /*
+  * Converts an expression into a Java object.
```
vcrfxia, line 99:
```suggestion
  * Converts a generic expression into the proper Java type.
```

ksqldb-tools/src/main/java/io/confluent/ksql/tools/migrations/util/CommandParser.java
```diff
@@ -0,0 +1,228 @@
...
+      this.command = command;
+    }
+
+    public String getCommand() {
+      return command;
+    }
+  }
+
+  /*
+   * Represents ksqlDb `INSERT INTO ... VALUES ...;` statements
```
vcrfxia, line 179: nit:
```suggestion
   * Represents ksqlDB `INSERT INTO ... VALUES ...;` statements
```

ksqldb-tools/src/main/java/io/confluent/ksql/tools/migrations/util/CommandParser.java
```diff
@@ -0,0 +1,228 @@
...
+      return values;
+    }
+
+    public List<String> getColumns() {
+      return columns;
+    }
+  }
+
+  /*
+  * Represents commands that can be sent directly to the the Java client's `executeStatement` method
```
vcrfxia, line 212: nit:
```suggestion
  * Represents commands that can be sent directly to the Java client's `executeStatement` method
```

ksqldb-tools/src/main/java/io/confluent/ksql/tools/migrations/commands/ApplyMigrationCommand.java
```diff
@@ -274,28 +291,95 @@ private boolean applyMigration(
...
+        row.put(sourceFields.get(i).name(), CommandParser.toFieldType(insertValues.get(i)));
+      }
+    }
+
+    return new KsqlObject(row);
+  }
+
+  private void verifyColumnValuesMatch(final List<String> columns, final List<Expression> values) {
+    if (columns.size() != values.size()) {
+      throw new MigrationException("Expected number columns and values to match: "
```
vcrfxia, line 380: nit: reword the error message to make it clear that this is thrown due to user error (rather than something wrong with the migrations tool):
```suggestion
      throw new MigrationException("Invalid `INSERT VALUES` statement. Number of columns and values must match. Got: Columns: %d. Values: %d."
```

vcrfxia (approved):
Hey @jzaralim -- this is awesome! Minor comments inline; feel free to defer to follow-ups as you see fit. I didn't look at the test coverage super thoroughly since we've got follow-ups planned for that anyway. Could you also quickly update the PR description? I think the mentions of regexes are out of date with the latest version. Great work!

ksqldb-tools/src/main/java/io/confluent/ksql/tools/migrations/util/CommandParser.java
jzaralim: `INSERT INTO .. SELECT` doesn't have `AS`

ksqldb-tools/src/main/java/io/confluent/ksql/tools/migrations/util/CommandParser.java
jzaralim: Will add this to follow-up PR.

ksqldb-tools/src/main/java/io/confluent/ksql/tools/migrations/util/CommandParser.java
jzaralim: Added a check to throw on malformed string/comments

jzaralim:
Leaving out connector tests from the integration test because several other tests are flaky and this PR is blocking other work. Will add the tests in another PR.

PR #7162, opened 2021-03-05
fix: prevent IOB when printing topics with a key/value with an empty string

mmolimar:
### Description 
When printing topics, the record formatter throws and Index Out of Bounds exception in case the key or value is an empty string because the deserializers might not have a format to use. This PR fix this checking if the possible formats are not empty.

### Testing done 
Included empty values in the tests which already exist to validate this behavior.

### Reviewer checklist

### DIFF
--- a/ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/resources/streaming/RecordFormatter.java
+++ b/ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/resources/streaming/RecordFormatter.java
@@ -109,15 +109,15 @@ public final class RecordFormatter {
 
     final List<String> formatted = formatRecords(records);
 
-    final boolean sameKeyFormatChanged = keyDeserializers.getPossibleFormats().stream()
-        .map(activeKeyFormat::equals)
-        .findFirst()
-        .orElse(false);
+    final boolean sameKeyFormatChanged = keyDeserializers
+        .getPossibleFormats()
+        .stream()
+        .anyMatch(activeKeyFormat::equals);
 
-    final boolean sameValueFormatChanged = valueDeserializers.getPossibleFormats().stream()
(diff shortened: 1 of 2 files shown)


ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/resources/streaming/RecordFormatter.java
```diff
@@ -109,11 +109,15 @@ public RecordFormatter(
...
 
-    final boolean sameKeyFormatChanged = activeKeyFormat
-        .equals(keyDeserializers.getPossibleFormats().get(0));
-
-    final boolean sameValueFormatChanged = activeValueFormat
-        .equals(valueDeserializers.getPossibleFormats().get(0));
+    final boolean sameKeyFormatChanged = keyDeserializers.getPossibleFormats().stream()
+        .map(activeKeyFormat::equals)
+        .findFirst()
+        .orElse(false);
```
agavra, line 115: nit, this and the one below can be replaced with this pattern:
```java
keyDeserializers
  .getPossibleFormats()
  .stream()
  .anyMatch(activeKeyFormat::equals);
```

agavra (approved):
Nice catch! Thanks @mmolimar

guozhangwang (approved):
LGTM!

ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/resources/streaming/RecordFormatter.java
mmolimar: Done!

agavra: Thanks again! I'll wait for the build to be green and merge.

PR #7178, opened 2021-03-08
fix: incorporate lambdas into CoercionUtil and change LambdaVariable from Literal to Expression

stevenpyzhang:
### Description 
We need to pass lambda type mappings to coercion util which uses ExpressionTypeManager.

### Testing done 
Unit test and QTT

### Reviewer checklist

### DIFF
--- a/ksqldb-execution/src/main/java/io/confluent/ksql/execution/expression/tree/LambdaVariable.java
+++ b/ksqldb-execution/src/main/java/io/confluent/ksql/execution/expression/tree/LambdaVariable.java
@@ -22,7 +22,7 @@ import java.util.Objects;
 import java.util.Optional;
 
 @Immutable
-public class LambdaVariable extends Literal {
+public class LambdaVariable extends Expression {
 
   private final String lambdaCharacter;
 
@@ -35,8 +35,7 @@ public class LambdaVariable extends Literal {
     this.lambdaCharacter = lambdaCharacter;
   }
 
-  @Override
-  public String getValue() {
+  public String getLambdaCharacter() {
     return lambdaCharacter;
   }
 
--- a/ksqldb-execution/src/main/java/io/confluent/ksql/execution/interpreter/TermCompiler.java
+++ b/ksqldb-execution/src/main/java/io/confluent/ksql/execution/interpreter/TermCompiler.java
@@ -109,6 +109,7 @@ import io.confluent.ksql.util.KsqlConfig;
 import io.confluent.ksql.util.KsqlException;
 import io.confluent.ksql.util.Pair;
 import java.util.ArrayList;
+import java.util.Collections;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
@@ -191,7 +192,7 @@ public class TermCompiler implements ExpressionVisitor<Term, Context> {
       final Context context
   ) {
     final InPredicate preprocessed = InListEvaluator
-        .preprocess(inPredicate, expressionTypeManager);
+        .preprocess(inPredicate, expressionTypeManager, Collections.emptyMap());
 
     final Term value = process(preprocessed.getValue(), context);
 
--- a/ksqldb-execution/src/main/java/io/confluent/ksql/execution/interpreter/TermCompiler.java
+++ b/ksqldb-execution/src/main/java/io/confluent/ksql/execution/interpreter/TermCompiler.java
@@ -109,7 +109,6 @@ import io.confluent.ksql.util.KsqlConfig;
 import io.confluent.ksql.util.KsqlException;
 import io.confluent.ksql.util.Pair;
 import java.util.ArrayList;
-import java.util.Collections;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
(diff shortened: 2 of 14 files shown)


ksqldb-execution/src/main/java/io/confluent/ksql/execution/expression/tree/LambdaVariable.java
```diff
@@ -22,7 +22,7 @@
 import java.util.Optional;
 
 @Immutable
-public class LambdaVariable extends Literal {
+public class LambdaVariable extends Expression {
```
lct45, line 25: To make sure I understand - coercion util looks at expressions to fix type issues, so we need this to be an expression to fit with that. Is the scenario if we have something like `transform(arr, x => x + 2.3)` where arr is an array of decimals so x will be coerced as a lambda variable?

ksqldb-execution/src/main/java/io/confluent/ksql/execution/interpreter/TermCompiler.java
```diff
@@ -191,7 +192,7 @@ public Term visitInPredicate(
       final Context context
   ) {
     final InPredicate preprocessed = InListEvaluator
-        .preprocess(inPredicate, expressionTypeManager);
+        .preprocess(inPredicate, expressionTypeManager, Collections.emptyMap());
```
lct45, line 195: Why don't you get the lambda mapping from the context here?

ksqldb-execution/src/main/java/io/confluent/ksql/execution/util/CoercionUtil.java
```diff
@@ -101,11 +103,19 @@ private CoercionUtil() {
    * @param typeManager the type manager used to resolve expression types.
    * @return the coercion result.
    */
-  public static Result coerceUserList(
+  static Result coerceUserList(
```
lct45, line 106: Did you keep this for testing purposes? IIUC all the places in code that this was used have been replaced by the other method. If it's there for testing, should we note that or leave it public?

ksqldb-execution/src/main/java/io/confluent/ksql/execution/expression/tree/LambdaVariable.java
guozhangwang: I'd like to ask the same, for my own understanding as well :)

ksqldb-execution/src/main/java/io/confluent/ksql/execution/interpreter/TermCompiler.java
guozhangwang: cc @AlanConfluent for a quick look.

ksqldb-execution/src/main/java/io/confluent/ksql/execution/util/ExpressionTypeManager.java
```diff
@@ -189,13 +189,13 @@ public Void visitLambdaVariable(
         final LambdaVariable node, final Context context
     ) {
       context.setSqlType(
-          context.getLambdaSqlTypeMapping().get(node.getValue())
+          context.getLambdaSqlTypeMapping().get(node.getLambdaCharacter())
```
guozhangwang, line 192: nit: this is for line 130 above. I'm wondering if we should not provide two overloaded `getExpressionSqlType` and just enforce all callers to explicitly create `emptyMap` if necessary, to avoid potential bugs like this?

ksqldb-execution/src/main/java/io/confluent/ksql/execution/expression/tree/LambdaVariable.java
stevenpyzhang: Lambda Variables don't represent any values, they're basically like "local column references". If you look in coercion util at
```
    private static Expression convertToCommonType(
        final TypedExpression exp,
        final SqlType commonType
    )
```
Before I changed this to an expression, it was trying to convert the "value" of the lambda variable to whatever type was needed, which isn't the correct behavior, we should just be casting the lambda variable to the correct type.

ksqldb-execution/src/main/java/io/confluent/ksql/execution/interpreter/TermCompiler.java
stevenpyzhang: that's a miss, I'll put in the lambda mapping

ksqldb-execution/src/main/java/io/confluent/ksql/execution/util/CoercionUtil.java
stevenpyzhang: Let me rework the javadocs, this is just for testing purposes yes

ksqldb-execution/src/main/java/io/confluent/ksql/execution/util/ExpressionTypeManager.java
stevenpyzhang: Yeah that sounds reasonable. I'll do it in a follow up at some point.

PR #7180, opened 2021-03-09
fix: check for null inputs for various timestamp functions

### LINKED ISSUE #7160: Most of the timestamp functions throw NPE on null input
**Describe the bug**
Some of them should return null instead of throwing NPE. For example,

```
timestampadd(SECONDS, 5, null);
```
should return null.

jzaralim:
### Description 
Fixes #7160

### Testing done 
Updated unit and QTT tests

### Reviewer checklist


spena (approved):
LGTM

PR #7186, opened 2021-03-09
chore: Put interpreter use of lambdas behind the flag ksql.lambdas.enabled

AlanConfluent:
### Description 
Now throws an error if the use of lambdas in the interpreter is attempted without setting `ksql.lambdas.enabled=true`

### Testing done 
Added RQTT

### Reviewer checklist


agavra (approved):

jnh5y:
@AlanConfluent do we still want this change?

If so, there may be some plans which need to be generated.

PR #7187, opened 2021-03-09
chore: add plugin to generate licences file

mjsax:
### Description 

### Testing done 

### Reviewer checklist


pom.xml
```diff
@@ -635,6 +635,52 @@
...
+                <executions>
+                    <execution>
+                        <id>create-licenses</id>
+                        <configuration>
+                            <mainClass>io.confluent.licenses.LicenseFinder</mainClass>
+                            <arguments>
+                                <!-- Note use of development instead of package so we pick up all dependencies.
+                                     This assumes both subprojects will be packaged together. -->
+                                <argument>-i ${project.build.directory}/${project.build.finalName}-package/share/java/acl</argument>
+                                <argument>-o ${project.parent.basedir}/licenses/overrides</argument>
```
ewencp, line 652: If this is in the root pom, `project.parent.basedir` won't work, you'd just want `project.basedir` since the licenses/override directory is at the root directory as well.

But I think from a quick glance this is in the wrong place anyway. You want it in any pom file where you are using maven-assembly plugin to generate packages:
* -i is the directory that the packaged version lands in, we run assembly plugin then use those jars to generate license info
* You can keep the `parent` part for -o if your overrides are in the root and reused by all components.
* -h -l -n -t should all be fine as is I think
* -x is an ignore flag, it is intended to list any jars your project produces since we don't need to include anything for them -- they are already covered just by the LICENSE file we ship

pom.xml
```diff
@@ -635,6 +635,52 @@
...
+                    </execution>
+                </executions>
+                <configuration>
+                    <includeProjectDependencies>true</includeProjectDependencies>
+                    <includePluginDependencies>true</includePluginDependencies>
+                    <executableDependency>
+                        <groupId>io.confluent</groupId>
+                        <artifactId>licenses</artifactId>
+                    </executableDependency>
+                    <mainClass>io.confluent.controlcenter.acl.StartupAcl</mainClass>
```
ewencp, line 674: I don't actually remember what this part of the config does, but you'll want to change this to the main class for each module you put this in.

pom.xml
```diff
@@ -635,6 +635,52 @@
...
+                                <argument>-o ${project.parent.basedir}/licenses/overrides</argument>
+                                <argument>-f</argument>
+                                <argument>-h ${project.build.directory}/${project.build.finalName}-package/share/doc/${project.name}/licenses.html</argument>
+                                <argument>-l ${project.build.directory}/${project.build.finalName}-package/share/doc/${project.name}/licenses</argument>
+                                <argument>-n ${project.build.directory}/${project.build.finalName}-package/share/doc/${project.name}/notices</argument>
+                                <argument>-t ${project.name}</argument>
+                                <argument>-x connect-api-${ce.kafka.version}.jar,connect-json-${ce.kafka.version}.jar,connect-runtime-${ce.kafka.version}.jar,control-center-${io.confluent.blueway.version}.jar,licenses-${io.confluent.license-file-generator.version}.jar,confluent-licensing-new-${ce.kafka.version}.jar,control-center-frontend-${io.confluent.blueway-fe.version}.jar,confluent-command-${io.confluent.blueway.version}.jar</argument>
```
mjsax, line 658: @ewencp Just cycling back to this PR. Seems the `-x` flag is used to exclude stuff?

I copied this snipped from `blueway` -- to I am wondering what we would want/need to exclude for ksqlDB ?
```
connect-api-${ce.kafka.version}.jar,
connect-json-${ce.kafka.version}.jar,
connect-runtime-${ce.kafka.version}.jar,
control-center-${io.confluent.blueway.version}.jar,
licenses-${io.confluent.license-file-generator.version}.jar,
confluent-licensing-new-${ce.kafka.version}.jar,
control-center-frontend-${io.confluent.blueway-fe.version}.jar,
confluent-command-${io.confluent.blueway.version}.jar
```

ewencp: It's a list of packages that are also just under the confluent (enterprise) license, which is actually located elsewhere. So basically just proprietary ones, which shouldn't exist for this repo.

Generally the easiest way to generate the list though is to look at the packaged jars and filter out anything that is proprietary or belongs to this project anyway (i.e. you don't need to extract licenses from ksql jars since they are covered by the top-level LICENSE file within the package).

PR #7190, opened 2021-03-10
feat(migrations): enable set/unset statements for migration tool

jzaralim:
### Description 
As the title states. Start reviewing from `feat(migrations): enable set/unset statements for migration tool`

### Testing done 
Added unit and integration test

### Reviewer checklist


ksqldb-tools/src/main/java/io/confluent/ksql/tools/migrations/util/CommandParser.java
```diff
@@ -176,6 +186,11 @@ private static SqlCommand transformToSqlCommand(final String sql) {
         return new SqlConnectorStatement(sql);
       case STATEMENT:
         return new SqlStatement(sql);
+      case SET_PROPERTY:
+        final List<String> strings = getStringsInCommand(sql);
+        return new SqlPropertyCommand(sql, true, strings.get(0), strings.get(1));
```
vcrfxia, line 191: I think we should be stricter about verifying what the statement looks like here. Currently, we allow malformed statements such as `SET 'string1' 'string2' 'string3';` which has no actual meaning. Because `SET` and `UNSET` statements are so simple, I think we can use regexes here. The only edge case I can think of that we'll want to check for is whether we require whitespace after `SET`/`UNSET` or not (I don't remember off the top of my head), i.e., is `SET'prop' = 'val';` a valid statement, according to the ksqlDB parser?

ksqldb-tools/src/main/java/io/confluent/ksql/tools/migrations/util/CommandParser.java
```diff
@@ -263,4 +293,37 @@ public String getSourceName() {
...
+    SqlPropertyCommand(
+        final String command,
+        final boolean set,
+        final String property,
+        final String value
+    ) {
+      super(command);
+      this.set = set;
+      this.property = property;
+      this.value = value;
```
vcrfxia, line 314: nit: convert to Optional to clarify that this may not be present. Then we can also add null checks on `property` and `value` for extra clarity.

ksqldb-tools/src/main/java/io/confluent/ksql/tools/migrations/util/CommandParser.java
```diff
@@ -263,4 +293,37 @@ public String getSourceName() {
...
+    }
+
+    public boolean isSet() {
```
vcrfxia, line 317: nit:
```suggestion
    public boolean isSetCommand() {
```

ksqldb-tools/src/main/java/io/confluent/ksql/tools/migrations/commands/ApplyMigrationCommand.java
```diff
@@ -319,11 +320,38 @@ private void executeCommands(
       final Clock clock,
       final String previous
   ) {
+    final Map<String, Object> properties = new HashMap<>();
     final List<SqlCommand> commands = CommandParser.parse(migrationFileContent);
     for (final SqlCommand command : commands) {
       try {
-        executeCommand(command, ksqlClient, restClient);
```
vcrfxia, line 325: I think the code was more readable with this helper method. We can pass in `properties` to the helper method and let it manipulate the properties from within the method. If your concern was about properties changing within the helper, I don't think it's a huge deal since it's a private helper but we can also add a javadoc on the helper to make that explicit so future readers won't be confused.

vcrfxia (approved):
LGTM with minor comments inline. Thanks @jzaralim !

Issue #6869, opened 2021-01-14
ksql.service.id should not be usable as a query parameter


MichaelDrogalis:
**Describe the bug**

The server parameter `ksql.service.id` is currently allowed to be used as a query-level parameter. This should only be settable once at the server level.

PR #7192, opened 2021-03-10
fix: ksql.service.id should not be usable as a query parameter

tolgadur:
Issue: [#6869](https://github.com/confluentinc/ksql/issues/6869)
ksql.service.id was settable as a query-level parameter but it should only be settable as a server-level parameter.

### Description 
The query-parameters that aren't allowed to be set are defined in the denylist that is imported from the ksql-server.properties. I hardcoded the 'ksql.servier.id' in the denylist as I wasn't sure if I should edit the config files. Hence, alternative to my approach we could edit the config files from where the denylist is imported.

### Testing done 
Just made sure that all tests run as there were already tests checking whether the properties set in the denylist actually are actually denied from query-level parameter setting.

### Reviewer checklist

### DIFF
--- a/ksqldb-common/src/main/java/io/confluent/ksql/properties/DenyListPropertyValidator.java
+++ b/ksqldb-common/src/main/java/io/confluent/ksql/properties/DenyListPropertyValidator.java
@@ -17,6 +17,7 @@ package io.confluent.ksql.properties;
 
 import com.google.common.collect.ImmutableSet;
 import com.google.common.collect.Sets;
+import io.confluent.ksql.util.KsqlConfig;
 import io.confluent.ksql.util.KsqlException;
 
 import java.util.Collection;
@@ -32,8 +33,8 @@ public class DenyListPropertyValidator {
   private final Set<String> immutableProps;
 
   public DenyListPropertyValidator(final Collection<String> immutableProps) {
-    this.immutableProps = ImmutableSet.copyOf(
-        Objects.requireNonNull(immutableProps, "immutableProps"));
+    this.immutableProps = ImmutableSet.<String>builder().addAll(
+        Objects.requireNonNull(immutableProps, "immutableProps")).add(KsqlConfig.KSQL_SERVICE_ID_CONFIG).build();
   }
 
   /**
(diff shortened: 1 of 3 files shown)


ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/KsqlRestApplication.java
```diff
@@ -618,6 +618,7 @@ public static KsqlRestApplication buildApplication(final KsqlRestConfig restConf
     final String ksqlServerId = ksqlConfig.getString(KsqlConfig.KSQL_SERVICE_ID_CONFIG);
     updatedRestProps.putAll(
         MetricCollectors.addConfluentMetricsContextConfigs(ksqlServerId, kafkaClusterId));
+    updatedRestProps.put(KsqlConfig.KSQL_PROPERTIES_OVERRIDES_DENYLIST, "ksql.service.id");
```
agavra, line 621: if this was already specified, wouldn't this overwrite the previous value?

tolgadur: Yes, you are right :sweat_smile:. I changed it.

agavra (approved):
lgtm :) thanks for patching a long-standing bug!

ksqldb-common/src/main/java/io/confluent/ksql/properties/DenyListPropertyValidator.java
```diff
@@ -32,8 +32,8 @@
   private final Set<String> immutableProps;
 
   public DenyListPropertyValidator(final Collection<String> immutableProps) {
-    this.immutableProps = ImmutableSet.copyOf(
-        Objects.requireNonNull(immutableProps, "immutableProps"));
+    this.immutableProps = ImmutableSet.<String>builder().addAll(
+        Objects.requireNonNull(immutableProps, "immutableProps")).add("ksql.service.id").build();
```
stevenpyzhang, line 36: I think you should be able to specify `KsqlConfig.KSQL_SERVICE_ID_CONFIG` here instead of hardcoding the property

stevenpyzhang (approved):
LGTM, could you also add a unit test in DenyListPropertyValidatorTest.java?

PR #7193, opened 2021-03-10
fix: Pull Queries: Avoids inefficient KsqlConfig copy with overrides

AlanConfluent:
### Description 
The call to `sessionConfig.getConfig(true);` copies fields of ksqlconfig in order to apply overrides.  This is very inefficient, slowing pull query performance locally on my Mac from 10k qps for key lookups to 6k, a 40% reduction.

I had made this change recently with #7006 so that the interpreter override would be available for pull queries.  It had this unfortunate effect.  The solution is to avoid the large copy and check overrides deliberately.  We already have a pattern for this in `PullQueryConfigPlannerOptions`.

### Testing done 
Ran unit tests

### Reviewer checklist

### DIFF
--- a/ksqldb-engine/src/main/java/io/confluent/ksql/engine/EngineExecutor.java
+++ b/ksqldb-engine/src/main/java/io/confluent/ksql/engine/EngineExecutor.java
@@ -163,7 +163,10 @@ final class EngineExecutor {
           queryAnalyzer.analyze(statement.getStatement(), Optional.empty()),
           new PullQueryExecutionUtil.ColumnReferenceRewriter()::process
       );
-      final KsqlConfig ksqlConfig = sessionConfig.getConfig(true);
+      // Do not set sessionConfig.getConfig to true! The copying is inefficient and slows down pull
+      // query performance significantly.  Instead use PullPlannerOptions which check overrides
+      // deliberately.
+      final KsqlConfig ksqlConfig = sessionConfig.getConfig(false);
       final LogicalPlanNode logicalPlan = buildAndValidateLogicalPlan(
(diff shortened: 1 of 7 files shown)


ksqldb-engine/src/main/java/io/confluent/ksql/engine/EngineExecutor.java
```diff
@@ -163,7 +163,10 @@ PullQueryResult executePullQuery(
           queryAnalyzer.analyze(statement.getStatement(), Optional.empty()),
           new PullQueryExecutionUtil.ColumnReferenceRewriter()::process
       );
-      final KsqlConfig ksqlConfig = sessionConfig.getConfig(true);
+      // Do not set sessionConfig.getConfig to true! The copying is inefficient and slows down pull
+      // query performance significantly.  Instead use PullPlannerOptions which check overrides
+      // deliberately.
+      final KsqlConfig ksqlConfig = sessionConfig.getConfig(false);
```
agavra, line 169: yikes. i wonder if there's a way for us to return a "view" when we call `getConfig(true)` instead of copying. This has bitten us twice 😢

agavra (approved):
Good catch! Thanks Alan

guozhangwang (approved):
LGTM!

ksqldb-engine/src/main/java/io/confluent/ksql/engine/EngineExecutor.java
AlanConfluent: Yeah, I think rather than copy, we should more or less do what we do in this PR.  Don't copy, but just check the overrides first.

PR #7212, opened 2021-03-12
fix: Ensures BaseSubscriber.makeRequest is called on context in PollableSubscriber

### LINKED ISSUE #7115: ksqldb-api-client: Cannot retrieve more than 100 rows in pull/push queries using StreamedQueryResult
I have a simple stream that contains 500+ records and I'm trying to query those records with some added filters in the push query. The result set should contain 200+ records easily but only 100 rows are being returned.

To reproduce:
- Create a test Ksql Table
- Insert 100+ records
- Try executing the pull/push query on this table using `this.ksqlDBJavaClient.streamQuery(sql, Collections.singletonMap("auto.offset.reset", "earliest"));`

@vcrfxia - Thanks for your contribution to the api client. I see you added the Request batch size limit [here](https://github.com/confluentinc/ksql/commit/280ef0ca8aa02693db1427ece08cd7863abed98d#diff-a111db1fb7491d6b0b2400e4f5a8d9c4e3c2956a5cf3568f9bb048b02f9c029bR30). Do you think this needs to be configurable?

Error stacktrace:
```
2021-02-28 23:27:07.736 ERROR 46923 --- [   scheduling-1] o.s.s.s.TaskUtils$LoggingErrorHandler    : Unexpected error occurred in scheduled task

java.lang.IllegalStateException: On wrong context or worker
	at io.confluent.ksql.util.VertxUtils.checkContext(VertxUtils.java:38) ~[ksqldb-common-6.1.0.jar:na]
	at io.confluent.ksql.reactive.BaseSubscriber.checkContext(BaseSubscriber.java:147) ~[ksqldb-common-6.1.0.jar:na]
	at io.confluent.ksql.reactive.BaseSubscriber.makeRequest(BaseSubscriber.java:122) ~[ksqldb-common-6.1.0.jar:na]
	at io.confluent.ksql.api.client.impl.PollableSubscriber.checkRequestTokens(PollableSubscriber.java:115) ~[ksqldb-api-client-6.1.0.jar:na]
	at io.confluent.ksql.api.client.impl.PollableSubscriber.poll(PollableSubscriber.java:90) ~[ksqldb-api-client-6.1.0.jar:na]
	at io.confluent.ksql.api.client.impl.StreamedQueryResultImpl.poll(StreamedQueryResultImpl.java:105) ~[ksqldb-api-client-6.1.0.jar:na]
	at io.confluent.ksql.api.client.impl.StreamedQueryResultImpl.poll(StreamedQueryResultImpl.java:86) ~[ksqldb-api-client-6.1.0.jar:na]
	at io.confluent.ksql.api.client.impl.StreamedQueryResultImpl.poll(StreamedQueryResultImpl.java:81) ~[ksqldb-api-client-6.1.0.jar:na]
```

AlanConfluent:
### Description 

Fixes a small bug where `BaseSubscriber.makeRequest` can be called from outside the Vertx context, which should be disallowed.

`makeRequest` is called from `checkRequestTokens` and there are currently a few paths into `checkRequestTokens`:
- One is `afterSubscribe` which is called from the context
- The other is `poll` which is inherently blocking and shouldn't be called from the context.

For this reason, this PR exposes `runOnRightContext` and invokes `makeRequest` using that.

Fixes #7115 

### Testing done 
Ran tests and manually exercised this path by fetching more than the batch size.

### Reviewer checklist

### DIFF
--- a/ksqldb-api-client/src/main/java/io/confluent/ksql/api/client/impl/PollableSubscriber.java
+++ b/ksqldb-api-client/src/main/java/io/confluent/ksql/api/client/impl/PollableSubscriber.java
@@ -23,6 +23,7 @@ import java.util.Objects;
 import java.util.concurrent.BlockingQueue;
 import java.util.concurrent.LinkedBlockingQueue;
 import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicInteger;
 import java.util.function.Consumer;
 import org.reactivestreams.Subscription;
 
@@ -33,7 +34,7 @@ public class PollableSubscriber extends BaseSubscriber<Row> {
 
   private final BlockingQueue<Row> queue = new LinkedBlockingQueue<>();
   private final Consumer<Throwable> errorHandler;
-  private int tokens;
+  private final AtomicInteger tokens = new AtomicInteger(0);
   private volatile boolean complete;
   private volatile boolean closed;
   private volatile boolean failed;
@@ -85,8 +86,8 @@ public class PollableSubscriber extends BaseSubscriber<Row> {
       try {
         final Row row = queue.poll(pollTime, TimeUnit.NANOSECONDS);
         if (row != null) {
-          tokens--;
-          checkRequestTokens();
+          tokens.decrementAndGet();
+          context.runOnContext(v -> checkRequestTokens());
           return row;
         } else if (complete) {
           // If complete, close once the queue has been emptied
@@ -109,9 +110,8 @@ public class PollableSubscriber extends BaseSubscriber<Row> {
   }
 
   private void checkRequestTokens() {
-    if (tokens == 0) {
-      tokens += REQUEST_BATCH_SIZE;
-      context.runOnContext(v -> makeRequest(REQUEST_BATCH_SIZE));
+    if (tokens.compareAndSet(0, REQUEST_BATCH_SIZE)) {
+      makeRequest(REQUEST_BATCH_SIZE);
     }
   }
 }
--- a/ksqldb-api-client/src/main/java/io/confluent/ksql/api/client/impl/PollableSubscriber.java
+++ b/ksqldb-api-client/src/main/java/io/confluent/ksql/api/client/impl/PollableSubscriber.java
@@ -27,10 +27,9 @@ import java.util.function.Consumer;
 import org.reactivestreams.Subscription;
 
 public class PollableSubscriber extends BaseSubscriber<Row> {
-
-  private static final int REQUEST_BATCH_SIZE = 100;
   // 100ms in ns
   private static final long MAX_POLL_NANOS = TimeUnit.MILLISECONDS.toNanos(100);
+  static final int REQUEST_BATCH_SIZE = 100;
 
   private final BlockingQueue<Row> queue = new LinkedBlockingQueue<>();
   private final Consumer<Throwable> errorHandler;
@@ -112,7 +111,7 @@ public class PollableSubscriber extends BaseSubscriber<Row> {
   private void checkRequestTokens() {
     if (tokens == 0) {
       tokens += REQUEST_BATCH_SIZE;
-      runOnRightContext(() -> makeRequest(REQUEST_BATCH_SIZE));
+      context.runOnContext(v -> makeRequest(REQUEST_BATCH_SIZE));
     }
   }
 }
--- a/ksqldb-api-client/src/main/java/io/confluent/ksql/api/client/impl/PollableSubscriber.java
+++ b/ksqldb-api-client/src/main/java/io/confluent/ksql/api/client/impl/PollableSubscriber.java
@@ -112,7 +112,7 @@ public class PollableSubscriber extends BaseSubscriber<Row> {
   private void checkRequestTokens() {
     if (tokens == 0) {
       tokens += REQUEST_BATCH_SIZE;
-      makeRequest(REQUEST_BATCH_SIZE);
+      runOnRightContext(() -> makeRequest(REQUEST_BATCH_SIZE));
     }
   }
 }
--- a/ksqldb-api-client/src/test/java/io/confluent/ksql/api/client/impl/PollableSubscriberTest.java
+++ b/ksqldb-api-client/src/test/java/io/confluent/ksql/api/client/impl/PollableSubscriberTest.java
@@ -90,11 +90,13 @@ public class PollableSubscriberTest {
     publisher.subscribe(pollableSubscriber);
 
     Row row = pollableSubscriber.poll(POLL_DURATION);
-    for (int i = 0; row != null; i++) {
+    int i = 0;
+    for (; row != null; i++) {
       Long col1 = row.getLong(COLUMN_NAME);
       assertThat(col1, is((long) i));
       row = pollableSubscriber.poll(POLL_DURATION);
     }
+    assertThat(i, is(numRows));
     assertThat(throwable, is(nullValue()));
   }
 
--- a/ksqldb-api-client/src/test/java/io/confluent/ksql/api/client/impl/PollableSubscriberTest.java
+++ b/ksqldb-api-client/src/test/java/io/confluent/ksql/api/client/impl/PollableSubscriberTest.java
@@ -1,0 +1,108 @@
+/*
+ * Copyright 2021 Confluent Inc.
+ *
+ * Licensed under the Confluent Community License (the "License"); you may not use
+ * this file except in compliance with the License.  You may obtain a copy of the
+ * License at
+ *
+ * http://www.confluent.io/confluent-community-license
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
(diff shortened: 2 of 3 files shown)


ksqldb-api-client/src/main/java/io/confluent/ksql/api/client/impl/PollableSubscriber.java
```diff
@@ -112,7 +112,7 @@ synchronized boolean isClosed() {
   private void checkRequestTokens() {
     if (tokens == 0) {
       tokens += REQUEST_BATCH_SIZE;
-      makeRequest(REQUEST_BATCH_SIZE);
+      runOnRightContext(() -> makeRequest(REQUEST_BATCH_SIZE));
```
agavra, line 115: fascinating. so without this we always hit the batch limit and then stopped? can we add some tests to make sure we can exceed the `BATCH_SIZE` - this bug is a little embarrassing! 😳

ksqldb-common/src/main/java/io/confluent/ksql/reactive/BaseSubscriber.java
```diff
@@ -147,7 +147,7 @@ protected final void checkContext() {
     VertxUtils.checkContext(context);
   }
 
-  private void runOnRightContext(final Runnable runnable) {
+  protected void runOnRightContext(final Runnable runnable) {
```
agavra, line 150: I'm a little confused, the first line in `makeRequest` checks the context. if that doesn't ensure our property, then what does it do? 

is there any reason not to just _always_ call `makeRequest` wrapped in a `runOnRightContext`? If that's the case could we keep this private and just make `makeRequest` do that?

agavra (commented):
Thanks @AlanConfluent - I'm always impressed with how fast you can get to the root of an issue! 

It LGTM (well, as much as I understand this code - which is not much at all, the Vert.X context and threading model always confuses me...), but I think we should add some test coverage to make sure #7115 doesn't come back.

ksqldb-common/src/main/java/io/confluent/ksql/reactive/BaseSubscriber.java
vcrfxia, line 150: I don't think we should expose this method. Let's just update the problematic line in `PollableSubscriber#poll()` to call `context.runOnContext(v -> checkRequestTokens());` instead of `checkRequestTokens()`. 

The purpose of having strict context checks is to ensure that implementers of BaseSubscriber are cognizant of which methods are being run where. When additional, user-facing methods outside the public methods of BaseSubscriber are exposed, it's preferable to have the implementer intentionally use the correct context, rather than throwing a blanket of "use this context if needed" at the method. If you're curious, there's precedent here: https://github.com/confluentinc/ksql/blob/ed2a1860e9c0b85388b41a2f8049d328cc2aba1d/ksqldb-rest-app/src/main/java/io/confluent/ksql/api/impl/InsertsSubscriber.java#L232

vcrfxia (commented):
Hey @AlanConfluent thanks for tracking down this bug! I'm +1 to Almog's comment on adding testing to prevent future regressions, and also left a comment inline regarding the fix.

ksqldb-common/src/main/java/io/confluent/ksql/reactive/BaseSubscriber.java
AlanConfluent: > I'm a little confused, the first line in makeRequest checks the context. if that doesn't ensure our property, then what does it do?

The call to `checkContext` just asserts that we're on the context and is the basis of the exceptions we've seen because we're not in fact on it in some call paths.

AlanConfluent: > I don't think we should expose this method. Let's just update the problematic line in PollableSubscriber#poll() to call context.runOnContext(v -> checkRequestTokens()); instead of checkRequestTokens().

Ok, sounds reasonable.  Rather than make token volatile, which you'd have to do if it was updated in the context, I just changed this within `checkRequestTokens` to:
```
context.runOnContext(v -> makeRequest(REQUEST_BATCH_SIZE));
```
 
This fits the invariant you describe about makeRequest being called on context as it is everywhere else.  Is that reasonable?

AlanConfluent:
> Hey @AlanConfluent thanks for tracking down this bug! I'm +1 to Almog's comment on adding testing to prevent future regressions, and also left a comment inline regarding the fix.

Added some tests that trigger the bug.

agavra (approved):
:) LGTM!

ksqldb-api-client/src/test/java/io/confluent/ksql/api/client/impl/PollableSubscriberTest.java
```diff
@@ -0,0 +1,108 @@
...
+    for (int i = 0; i < numRows; i++) {
+      rows.add(createRow(i));
+    }
+
+    publisher = new BufferedPublisher<>(context, rows);
+
+    publisher.subscribe(pollableSubscriber);
+
+    Row row = pollableSubscriber.poll(POLL_DURATION);
+    for (int i = 0; row != null; i++) {
```
vcrfxia, line 93: This does not actually verify we received the expected number of rows?

ksqldb-common/src/main/java/io/confluent/ksql/reactive/BaseSubscriber.java
vcrfxia: > Or do all non-public methods generally get called on the context?

Are you asking whether the entire `poll()` call should be run on the context? If so, the answer is no. Blocking operations are not meant to be performed on event loops.

> Rather than use synchonization with token and use an AtomicInteger, which you'd have to do if it was updated in the context, I just changed this within checkRequestTokens to: `context.runOnContext(v -> makeRequest(REQUEST_BATCH_SIZE));`

My only potential concern with this is that because `checkRequestTokens()` is called from `afterSubscribe()` which is already wrapped in a call to `context.runOnContext(...);`, we now have two layers of wrapping in this case. I think this is fine but it'd be good to double-check. (I hope we have automated testing that would fail if this was disallowed/blocking!)

ksqldb-api-client/src/test/java/io/confluent/ksql/api/client/impl/PollableSubscriberTest.java
AlanConfluent: You're right.  I forgot to check the index at the end.  Pulled it out so I can check it against numRows after the loop.

ksqldb-common/src/main/java/io/confluent/ksql/reactive/BaseSubscriber.java
AlanConfluent: > Are you asking whether the entire poll() call should be run on the context? If so, the answer is no. Blocking operations are not meant to be performed on event loops.

I definitely get that that's not allowed on the context.  Poll is a public method, so I wasn't meaning to include that.  I was more generally asking about the convention about when to ensure code is running on the context vs potentially some other thread.  "Do the onramps to the context only happen at a few well defined public API functions in these publishers and subscribers?" was what I meant.
 
> My only potential concern with this is that because checkRequestTokens() is called from afterSubscribe() which is already wrapped in a call to context.runOnContext(...);, we now have two layers of wrapping in this case. I think this is fine but it'd be good to double-check. (I hope we have automated testing that would fail if this was disallowed/blocking!)

Ah, you're right.  Didn't realize that about `afterSubscribe`.  Ok, tokens was already being accessed by multiple threads, so I just made it an AtomicInteger and took your original suggestion.

vcrfxia: > I was more generally asking about the convention about when to ensure code is running on the context vs potentially some other thread. "Do the onramps to the context only happen at a few well defined public API functions in these publishers and subscribers?" was what I meant.

In general or in how the ksql code in particular is setup? I don't know the general answer but the ksql setup makes sense to me: BaseSubscriber ensures all the subscriber methods are run on the correct context, since otherwise all subscriber implementations would have to do that separately. For all other methods (not included in BaseSubscriber), the subscriber implementation must ensure the relevant calls are run on the correct context, while minimizing the code that is run on the context to what must be run on the context, in order to not consume event loop resources unnecessarily. 

Not sure if I actually answered your question. LMK if I'm still misunderstanding.

vcrfxia (approved):
Thanks @AlanConfluent ! LGTM.

ksqldb-common/src/main/java/io/confluent/ksql/reactive/BaseSubscriber.java
AlanConfluent: I think that answers my question.  I was kind of wanting to know if in our codebase we had some kind of philosophy since reasoning at a glance about whether some method in one of these classes is run on or off the context is hard. In this case, it seems like `BaseSubscriber` handles most of that, which makes it easier.  Looking through the code, `PollableSubscriber` seems like a bit of an exception since it provides another public method that we know is run off the context, whereas all of these other uses are all async by nature.

PR #7218, opened 2021-03-12
chore: bump clients and connect logging level to INFO

ableegoldman:
The consumer and producer logs in particular are often critical for telling the whole story. They should be logged at the same level as Streams itself by default.


wcarlson5 (approved):

vcrfxia (approved):
LGTM!

PR #7222, opened 2021-03-12
feat(java client): Add connector functions to java client

jzaralim:
### Description 
Adds `createConnector`, `dropConnector`, `listConnector` and `describeConnector` to the Java client. I can split this into separate PRs if that makes it easier to review.

### Testing done 
Unit and integration tests

### Reviewer checklist


ksqldb-api-client/src/main/java/io/confluent/ksql/api/client/ConnectorDescription.java
```diff
@@ -0,0 +1,48 @@
...
+
+public interface ConnectorDescription {
+  /**
+   * @return class of this connector
+   */
+  String connectorClass();
+
+  /**
+   *
+   * @return a list of sources that this connector reads/writes to
```
vcrfxia, line 28: What sources are these? External sources (as opposed to ksqlDB streams/tables)?

ksqldb-api-client/src/main/java/io/confluent/ksql/api/client/ConnectorDescription.java
```diff
@@ -0,0 +1,48 @@
...
+   */
+  List<String> sources();
+
+  /**
+   *
+   * @return a list of topics consumed by this connector
```
vcrfxia, line 34: "consumed" meaning this is only present for sink connectors?

ksqldb-api-client/src/main/java/io/confluent/ksql/api/client/ConnectorDescription.java
```diff
@@ -0,0 +1,48 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OF ANY KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations under the License.
+ */
+
+package io.confluent.ksql.api.client;
+
+import java.util.List;
+
+public interface ConnectorDescription {
```
vcrfxia, line 20: This interface should contain the connector name as well. I think we should simply have it extend `ConnectorInfo` and add the two additional fields (sources and topics).

ksqldb-api-client/src/main/java/io/confluent/ksql/api/client/ConnectorType.java
```diff
@@ -0,0 +1,27 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OF ANY KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations under the License.
+ */
+
+package io.confluent.ksql.api.client;
+
+public interface ConnectorType {
+  enum Type { SOURCE, SINK, UNKNOWN }
```
vcrfxia, line 19: We should add a javadoc about what situations might cause the connector type to be `UNKNOWN`, since I imagine will be confused about this otherwise.

ksqldb-api-client/src/main/java/io/confluent/ksql/api/client/Client.java
```diff
@@ -234,6 +234,52 @@
...
+   *
+   * <p>If a non-200 response is received from the server, the {@code CompletableFuture} will be
+   * failed.
+   *
+   * @param connectorName name of the connector
+   * @param isSource true if the connector is a source connector, false if it is a sink connector
+   * @param properties connector properties
+   * @return result of connector creation
+   */
+  CompletableFuture<CreateConnectorResult> createConnector(
```
vcrfxia, line 248: @colinhicks do you want to sanity-check these new Java client interfaces for managing connectors, from the user/product perspective?

ksqldb-api-client/src/main/java/io/confluent/ksql/api/client/impl/ConnectorCommandResponseHandler.java
```diff
@@ -0,0 +1,120 @@
...
+  /**
+   * Attempts to parse the provided response entity as a {@code List<ConnectorInfo>}.
+   *
+   * @param connectorsEntity response entity
+   * @return optional containing parsed result if successful, else empty
+   */
+  private static Optional<List<ConnectorInfo>> getListConnectorsResponse(
+      final JsonObject connectorsEntity) {
+    try {
+      final JsonArray connectors = connectorsEntity.getJsonArray("connectors");
```
vcrfxia, line 107: A bit late for this PR but we've been wanting to change the pattern of how the Java client response interfaces relate to the server response objects for a while now, in order to avoid custom JSON parsing like this. Here's the ticket with context: https://github.com/confluentinc/ksql/issues/6042

We should really do this before the next time we add additional interfaces, but it's outside the scope for this PR. (If you're keen on helping clean up the codebase and want to tackle it as a follow-up PR that'd be super, but no pressure 😁 )

ksqldb-api-client/src/main/java/io/confluent/ksql/api/client/impl/ConnectorDescriptionImpl.java
```diff
@@ -0,0 +1,103 @@
...
+
+  public ConnectorDescriptionImpl(
+      final String connectorClass,
+      final List<String> sources,
+      final List<String> topics,
+      final ConnectorType type,
+      final String state
+
+  ) {
+    this.connectorClass = Objects.requireNonNull(connectorClass);
```
vcrfxia, line 39: nit: add second argument to name failure should this ever happen (and similarly for below and other classes).
```suggestion
    this.connectorClass = Objects.requireNonNull(connectorClass, "connectorClass");
```

ksqldb-api-client/src/test/java/io/confluent/ksql/api/client/integration/ClientIntegrationTest.java
```diff
@@ -972,6 +1028,66 @@ public void shouldGetServerInfo() throws Exception {
     assertThat(serverInfo.getKafkaClusterId(), is(expectedClusterId));
   }
 
+  @Test
+  public void shouldListConnectors() throws Exception {
+    // Given:
+    makeKsqlRequest("CREATE SOURCE CONNECTOR " + TEST_CONNECTOR + " WITH ('connector.class'='" + MOCK_SOURCE_CLASS + "');");
```
vcrfxia, line 1034: If all three of these tests create the same connector, why do whichever tests run second and third not fail (because the connector already exists)?

ksqldb-api-client/src/test/java/io/confluent/ksql/api/client/integration/ClientIntegrationTest.java
```diff
@@ -972,6 +1028,66 @@ public void shouldGetServerInfo() throws Exception {
...
+    assertThat(connector.topics().size(), is(0));
+    assertThat(connector.sources().size(), is(0));
+    assertThat(connector.connectorClass(), is(MOCK_SOURCE_CLASS));
+  }
+
+  @Test
+  public void shouldDropConnector() throws Exception {
+    // Given:
+    makeKsqlRequest("CREATE SOURCE CONNECTOR " + TEST_CONNECTOR + " WITH ('connector.class'='" + MOCK_SOURCE_CLASS + "');");
+
```
vcrfxia, line 1067: Add validation that the connector was successfully created into the "Given" here, else the test could be a no-op (as far as the reader is aware).

ksqldb-api-client/src/main/java/io/confluent/ksql/api/client/Client.java
```diff
@@ -234,6 +234,52 @@
...
+   *
+   * <p>If a non-200 response is received from the server, the {@code CompletableFuture} will be
+   * failed.
+   *
+   * @return a list of connectors
+   */
+  CompletableFuture<List<ConnectorInfo>> listConnectors();
+
+  /**
+   * Returns metadata about a connector .
```
vcrfxia, line 273: nit: extra space before period.

vcrfxia (commented):
Hey @jzaralim -- awesome test coverage! Are you planning to add docs in a separate PR? If you wouldn't mind, we should also add docs for the new `serverInfo()` method you added as part of the 0.16 release (could be a separate PR).

Left some questions inline, mostly about the interfaces, but the high-level LGTM!

ksqldb-api-client/src/main/java/io/confluent/ksql/api/client/CreateConnectorResult.java
```diff
@@ -0,0 +1,36 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OF ANY KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations under the License.
+ */
+
+package io.confluent.ksql.api.client;
+
+import java.util.Map;
+
+public interface CreateConnectorResult {
```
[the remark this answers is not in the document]
jzaralim: I'm not sure, but this is the response that Connect returns when creating a new connector. If this is not useful for the Java client, we can remove it.

ksqldb-api-client/src/main/java/io/confluent/ksql/api/client/ConnectorDescription.java
jzaralim: Updated the Javadoc to make this more clear.

ksqldb-api-client/src/main/java/io/confluent/ksql/api/client/ConnectorDescription.java
jzaralim: Updated the Javadoc to make this more clear.

ksqldb-api-client/src/test/java/io/confluent/ksql/api/client/integration/ClientIntegrationTest.java
jzaralim: The statement fails but doesn't throw an error. I ended up moving this to setUp and tearDown though.

ksqldb-api-client/src/main/java/io/confluent/ksql/api/client/CreateConnectorResult.java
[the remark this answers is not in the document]
jzaralim: Hm actually I guess this wouldn't be the only place the client shows less info than the old rest api. I'll remove it, having 3 very similar classes for connector responses is clunky anyway.

ksqldb-api-client/src/main/java/io/confluent/ksql/api/client/ConnectorType.java
```diff
@@ -0,0 +1,34 @@
...
+ */
+
+package io.confluent.ksql.api.client;
+
+public interface ConnectorType {
+  enum Type {
+    SOURCE,
+    SINK,
+    /**
+     * Denotes an unknown connector type. This is used when there were errors in connector creation.
```
vcrfxia, line 23: Is this the only reason the connector type might be `UNKNOWN` or are there others as well, for example, when listing existing connectors? For some reason I thought there were other possible reasons for an `UNKNOWN` type as well but I could be wrong.

ksqldb-api-client/src/test/java/io/confluent/ksql/api/client/integration/ClientIntegrationTest.java
```diff
@@ -226,6 +283,12 @@ public static void classTearDown() {
   public void setUp() {
     vertx = Vertx.vertx();
     client = createClient();
+    makeKsqlRequest("CREATE SOURCE CONNECTOR " + TEST_CONNECTOR + " WITH ('connector.class'='" + MOCK_SOURCE_CLASS + "');");
```
vcrfxia, line 286: I don't think this belongs here. Most tests don't need the connector so it feels wasteful. How about we wrap this in a `givenConnectorExists()` method and only call it from the `Given:` portion of the tests that actually need it instead?

ksqldb-api-client/src/test/java/io/confluent/ksql/api/client/integration/ClientIntegrationTest.java
```diff
@@ -237,6 +300,9 @@ public void tearDown() {
       vertx.close();
     }
     REST_APP.getServiceContext().close();
+
+    ((ConnectorList) makeKsqlRequest("SHOW CONNECTORS;").get(0)).getConnectors()
```
vcrfxia, line 304: How does this work if it's being called after `REST_APP.getServiceContext().close();`? 

Instead of trying to put this in `tearDown()`, we can wrap it into `givenConnectorExists()` (see comment above) and `classTearDown()` instead.

vcrfxia (approved):
Thanks @jzaralim -- LGTM once the inline comments are addressed and docs are added. It'd be nice to get another opinion on the new client interfaces as well cc @colinhicks .

ksqldb-api-client/src/main/java/io/confluent/ksql/api/client/ConnectorType.java
jzaralim: I couldn't find any other references to `UNKNOWN`, but there could be a reason on the Connect side of things that I'm missing. Updated docs to be more accurate

ksqldb-api-client/src/test/java/io/confluent/ksql/api/client/integration/ClientIntegrationTest.java
```diff
@@ -1073,8 +1072,13 @@ public void shouldDropConnector() throws Exception {
     client.dropConnector(TEST_CONNECTOR).get();
 
     // Then:
-    final List<ConnectorInfo> connectors = client.listConnectors().get();
-    assertThat(connectors.size(), is(0));
+    assertThatEventually(() -> {
+      try {
+        return client.listConnectors().get().size();
+      } catch (InterruptedException | ExecutionException e) {
```
vcrfxia, line 1078: Is this exception expected (under normal cases)? What would cause this?

ksqldb-api-client/src/test/java/io/confluent/ksql/api/client/integration/ClientIntegrationTest.java
```diff
@@ -1123,11 +1127,29 @@ private void verifyNumQueries(final int numQueries) {
...
+    // Make sure we are starting from a clean slate before creating a new connector.
+    cleanupConnectors();
+
     makeKsqlRequest("CREATE SOURCE CONNECTOR " + TEST_CONNECTOR + " WITH ('connector.class'='" + MOCK_SOURCE_CLASS + "');");
+
+    assertThatEventually(
+        () -> {
+          try {
+            return ((ConnectorList) makeKsqlRequest("SHOW CONNECTORS;").get(0)).getConnectors().size();
+          } catch (AssertionError e) {
```
vcrfxia, line 1139: Same questions as above. I notice there's no analogous try-catch around the `SHOW CONNECTORS;` request in the `cleanupConnectors()` helper method, curious why this situation is different.

vcrfxia (commented):
Two more minor questions inline -- LGTM overall! (Plus the comments from before on docs and getting product input on the new interfaces) Thanks @jzaralim .

ksqldb-api-client/src/test/java/io/confluent/ksql/api/client/integration/ClientIntegrationTest.java
jzaralim: It takes a while to provision a connector, even after `CREATE CONNECTOR` returns a successful response. When ksqlDB calls `SHOW CONNECTORS;`, it makes a call to '/info' for reach connector, so if the new connector is not done provisioning, it returns an error response which propagates.

ksqldb-api-client/src/test/java/io/confluent/ksql/api/client/integration/ClientIntegrationTest.java
jzaralim: It's not. It's just that since this is a part of a lambda, we need to do a try-catch here in order for the code to compile.

ksqldb-api-client/src/main/java/io/confluent/ksql/api/client/Client.java
```diff
@@ -234,6 +234,52 @@
...
+   * <p>If a non-200 response is received from the server, the {@code CompletableFuture} will be
+   * failed.
+   *
+   * @param connectorName name of the connector
+   * @param isSource true if the connector is a source connector, false if it is a sink connector
+   * @param properties connector properties
+   * @return result of connector creation
+   */
+  CompletableFuture<Void> createConnector(
+      String connectorName, boolean isSource, Map<String, String> properties);
```
derekjn, line 249: I believe we use `Map<String, Object>` to pass properties elsewhere in the client interface. Is there a reason for using `<String, String>` here, or should we keep this consistent with the other interface methods?

jzaralim: Good catch! Will update the PR.

PR #7238, opened 2021-03-16
fix: Bubble up errors from HARouting unless using StandbyFallbackException

### LINKED ISSUE #6799: Pull Queries: HARouting does not show correct exception
If an exception is thrown during the physical plan execution, HARouting eats that exception and always throws 
```
"Unable to execute pull query: %s. Exhausted standby hosts to try.",
```
which is wrong since this is not the reason the query failed.

For example, trying this query 
```
select * from tab2 where id=1 and windowstart >= 1608151320000 and windowend > 0 and windowend < 1;
```

returns 
```
Unable to execute pull query: select * from tab2 where id=1 and windowstart >= 1608151320000 and windowend > 0 and windowend < 1;. Exhausted standby hosts to try.
```

whereas the exception is 
```
io.confluent.ksql.execution.streams.materialization.MaterializationException: Failed to get value from materialized table
	at io.confluent.ksql.execution.streams.materialization.ks.KsMaterializedWindowTable.get(KsMaterializedWindowTable.java:104)
	at io.confluent.ksql.execution.streams.materialization.KsqlMaterialization$KsqlMaterializedWindowedTable.get(KsqlMaterialization.java:160)
	at io.confluent.ksql.physical.pull.operators.KeyedWindowedTableLookupOperator.open(KeyedWindowedTableLookupOperator.java:71)
	at io.confluent.ksql.physical.pull.operators.SelectOperator.open(SelectOperator.java:55)
	at io.confluent.ksql.physical.pull.operators.ProjectOperator.open(ProjectOperator.java:71)
	at io.confluent.ksql.physical.pull.PullPhysicalPlan.open(PullPhysicalPlan.java:81)
	at io.confluent.ksql.physical.pull.PullPhysicalPlan.execute(PullPhysicalPlan.java:67)
	at io.confluent.ksql.physical.pull.HARouting.executeOrRouteQuery(HARouting.java:253)
	at io.confluent.ksql.physical.pull.HARouting.lambda$handlePullQuery$1(HARouting.java:154)
	at java.util.concurrent.FutureTask.run$$$capture(FutureTask.java:266)
	at java.util.concurrent.FutureTask.run(FutureTask.java)
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1149)
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:624)
Caused by: java.lang.IllegalArgumentException: fromKey > toKey
	at java.util.TreeMap$NavigableSubMap.<init>(TreeMap.java:1368)
	at java.util.TreeMap$AscendingSubMap.<init>(TreeMap.java:1855)
	at java.util.TreeMap.subMap(TreeMap.java:913)
	at org.apache.kafka.streams.state.internals.AbstractSegments.segments(AbstractSegments.java:125)
	at org.apache.kafka.streams.state.internals.WindowKeySchema.segmentsToSearch(WindowKeySchema.java:98)
	at org.apache.kafka.streams.state.internals.AbstractRocksDBSegmentedBytesStore.fetch(AbstractRocksDBSegmentedBytesStore.java:83)
	at org.apache.kafka.streams.state.internals.AbstractRocksDBSegmentedBytesStore.fetch(AbstractRocksDBSegmentedBytesStore.java:69)
	at org.apache.kafka.streams.state.internals.RocksDBWindowStore.fetch(RocksDBWindowStore.java:83)
	at org.apache.kafka.streams.state.internals.RocksDBWindowStore.fetch(RocksDBWindowStore.java:29)
	at org.apache.kafka.streams.state.WindowStore.fetch(WindowStore.java:110)
	at io.confluent.ksql.execution.streams.materialization.ks.WindowStoreCacheBypass.fetchUncached(WindowStoreCacheBypass.java:148)
	at io.confluent.ksql.execution.streams.materialization.ks.WindowStoreCacheBypass.fetch(WindowStoreCacheBypass.java:101)
	at io.confluent.ksql.execution.streams.materialization.ks.KsMaterializedWindowTable.get(KsMaterializedWindowTable.java:71)
	... 13 more
```

The query is wrong but I am only able to see the exception because I am debugging.

Throw the correct exception

AlanConfluent:
### Description 
`HARouting` has previously treated all Exceptions the same and used them as reason to retry with standbys.  The issue is that certain errors are deterministic and often caused by some form of user error or code bug.  Not only is there no use in retrying something that will have the same effect on a standby, but if we treat it similar to a transient error and print the error message `Unable to execute pull query: "%s". Exhausted standby hosts to try.`, it's rather confusing since trying again won't fix the issue.

The goal of this PR is to surface any bug by requiring any standby retries to explicitly use `StandbyFallbackException` and by bubbling up other exceptions by default.  In particular, any network error gets wrapped in a `StandbyFallbackException` and everything else, including local lookup errors, is surfaced.

Fixes #6799 

### Testing done 
I rewrote much of `HARoutingTest` since it wasn't well tested before and stopped at the routing step.  It's now tested all the way down to either the network call or execution at the physical plan layer.

### Reviewer checklist

### DIFF
--- a/ksqldb-engine/src/main/java/io/confluent/ksql/physical/pull/HARouting.java
+++ b/ksqldb-engine/src/main/java/io/confluent/ksql/physical/pull/HARouting.java
@@ -167,16 +167,13 @@ public final class HARouting implements AutoCloseable {
 
       // Make requests to each host, specifying the partitions we're interested in from
       // this host.
-      final Map<KsqlNode, Future<Void>> futures = new LinkedHashMap<>();
+      final Map<KsqlNode, Future<RoutingResult>> futures = new LinkedHashMap<>();
       for (Map.Entry<KsqlNode, List<KsqlPartitionLocation>> entry : groupedByHost.entrySet()) {
         final KsqlNode node = entry.getKey();
         futures.put(node, executorService.submit(
-            () -> {
-              routeQuery.routeQuery(
-                  node, entry.getValue(), statement, serviceContext, routingOptions,
-                  pullQueryMetrics, pullPhysicalPlan, outputSchema, queryId, pullQueryQueue);
-              return null;
-            }
+            () -> routeQuery.routeQuery(
+                node, entry.getValue(), statement, serviceContext, routingOptions,
+                pullQueryMetrics, pullPhysicalPlan, outputSchema, queryId, pullQueryQueue)
         ));
       }
 
@@ -184,20 +181,23 @@ public final class HARouting implements AutoCloseable {
       // the locations to the nextRoundRemaining list.
       final ImmutableList.Builder<KsqlPartitionLocation> nextRoundRemaining
           = ImmutableList.builder();
-      for (Map.Entry<KsqlNode, Future<Void>> entry : futures.entrySet()) {
-        final Future<Void> future = entry.getValue();
+      for (Map.Entry<KsqlNode, Future<RoutingResult>> entry : futures.entrySet()) {
+        final Future<RoutingResult> future = entry.getValue();
         final KsqlNode node = entry.getKey();
+        RoutingResult routingResult = null;
         try {
-          future.get();
+          routingResult = future.get();
         } catch (ExecutionException e) {
           LOG.warn("Error routing query {} to host {} at timestamp {} with exception {}",
               statement.getStatementText(), node, System.currentTimeMillis(), e.getCause());
-          if (!(e.getCause() instanceof StandbyFallbackException)) {
-            throw new MaterializationException(String.format(
-                "Unable to execute pull query \"%s\". %s",
-                statement.getStatementText(), e.getCause().getMessage()));
-          }
+          throw new MaterializationException(String.format(
+              "Unable to execute pull query \"%s\". %s",
+              statement.getStatementText(), e.getCause().getMessage()));
+        }
+        if (routingResult == RoutingResult.STANDBY_FALLBACK) {
           nextRoundRemaining.addAll(groupedByHost.get(node));
+        } else {
+          Preconditions.checkState(routingResult == RoutingResult.SUCCESS);
         }
       }
       remainingLocations = nextRoundRemaining.build();
@@ -239,7 +239,7 @@ public final class HARouting implements AutoCloseable {
 
   @VisibleForTesting
   interface RouteQuery {
-    void routeQuery(
+    RoutingResult routeQuery(
         KsqlNode node,
         List<KsqlPartitionLocation> locations,
         ConfiguredStatement<Query> statement,
@@ -254,7 +254,7 @@ public final class HARouting implements AutoCloseable {
   }
 
   @VisibleForTesting
-  static void executeOrRouteQuery(
+  static RoutingResult executeOrRouteQuery(
       final KsqlNode node,
       final List<KsqlPartitionLocation> locations,
       final ConfiguredStatement<Query> statement,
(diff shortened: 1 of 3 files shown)


ksqldb-engine/src/main/java/io/confluent/ksql/physical/pull/HARouting.java
```diff
@@ -273,29 +276,37 @@ static void executeOrRouteQuery(
...
       try {
         LOG.debug("Query {} routed to host {} at timestamp {}.",
-                  statement.getStatementText(), node.location(), System.currentTimeMillis());
+            statement.getStatementText(), node.location(), System.currentTimeMillis());
         pullQueryMetrics
             .ifPresent(queryExecutorMetrics -> queryExecutorMetrics.recordRemoteRequests(1));
         forwardTo(node, locations, statement, serviceContext, pullQueryQueue, rowFactory,
             outputSchema);
+      } catch (StandbyFallbackException e) {
+        LOG.error("Standby Fallback: Error forwarding query {} to node {} with exception {}",
```
guozhangwang, line 301: Do we want to log it as error really, since we are about to retry on the standby right?

Also if we intend to expose this to users, they may not really understand what `standby fallback` means. What about just "Failed to execute query {} on node {}, will try to switch to execute on the standby query state which may result in stale results"?

ksqldb-engine/src/main/java/io/confluent/ksql/physical/pull/HARouting.java
```diff
@@ -273,29 +276,37 @@ static void executeOrRouteQuery(
         pullQueryMetrics
             .ifPresent(queryExecutorMetrics -> queryExecutorMetrics.recordLocalRequests(1));
         pullPhysicalPlan.execute(locations, pullQueryQueue,  rowFactory);
+      } catch (StandbyFallbackException e) {
+        LOG.error("Standby Fallback: Error executing query {} locally at node {} with exception",
+            statement.getStatementText(), node, e.getCause());
+        throw e;
```
guozhangwang, line 282: Ditto here.

ksqldb-engine/src/main/java/io/confluent/ksql/physical/pull/HARouting.java
```diff
@@ -273,29 +276,37 @@ static void executeOrRouteQuery(
...
-                  statement.getStatementText(), node.location(), System.currentTimeMillis());
+            statement.getStatementText(), node.location(), System.currentTimeMillis());
         pullQueryMetrics
             .ifPresent(queryExecutorMetrics -> queryExecutorMetrics.recordRemoteRequests(1));
         forwardTo(node, locations, statement, serviceContext, pullQueryQueue, rowFactory,
             outputSchema);
+      } catch (StandbyFallbackException e) {
+        LOG.error("Standby Fallback: Error forwarding query {} to node {} with exception {}",
+            statement.getStatementText(), node, e.getCause());
+        throw e;
```
guozhangwang, line 303: Since on the caller we are going to swallow the `StandbyFallbackException` and retry on standbys, how about letting `executeOrRouteQuery` to return a result (for now, just boolean is okay; in general, maybe a enum can do more things) where the caller can check on the `futures`, instead of capturing and opening the exception envelope to check?

guozhangwang (commented):
Made a first pass on non-testing code. I'm wondering if we are going to capture/swallow the new type of exceptions anyways, could we just let the `executeOrRouteQuery` to return something for the caller that captures `standby fallback` instead of the exception throwing path?

cprasad1 (approved):
LGTM

ksqldb-engine/src/main/java/io/confluent/ksql/physical/pull/HARouting.java
AlanConfluent: Sure, good suggestion.  I just created an enum to make it a little easier to read.

guozhangwang:
LGTM!

PR #7243, opened 2021-03-16
fix(java client): allow java client to accept statements with more than one semicolon

jzaralim:
### Description 
The `executeStatement` Java client method fails if the given statement contains a string with semicolons. This PR updates the validator logic to ignore semicolons in strings when counting semicolons.

### Testing done 
Added a unit test

### Reviewer checklist

### DIFF
--- a/ksqldb-api-client/src/main/java/io/confluent/ksql/api/client/impl/DdlDmlRequestValidators.java
+++ b/ksqldb-api-client/src/main/java/io/confluent/ksql/api/client/impl/DdlDmlRequestValidators.java
@@ -46,6 +46,16 @@ final class DdlDmlRequestValidators {
     return true;
   }
 
+  /**
+   * Counts the number of sql statements in a string by
+   *  1. Removing all of the sql strings
+   *  2. Splitting the remaining substrings by ';'. The -1 argument in the split
+   *     function call ensures that each ';' will always have two partitions surrounding it, so that
+   *     the number of partitions is the same whether or not the final ';' has whitespace after it.
+   *  3. Counting the partitions
+   * @param sql a string containing sql statements
+   * @return the number of sql statements in the string
+   */
   private static int countStatements(final String sql) {
     return Arrays.stream(sql.split(QUOTED_STRING))
         .mapToInt(part -> part.split(";", -1).length - 1)
--- a/ksqldb-api-client/src/main/java/io/confluent/ksql/api/client/impl/DdlDmlRequestValidators.java
+++ b/ksqldb-api-client/src/main/java/io/confluent/ksql/api/client/impl/DdlDmlRequestValidators.java
@@ -22,7 +22,7 @@ import java.util.concurrent.CompletableFuture;
 
 final class DdlDmlRequestValidators {
 
-  private static final String QUOTED_STRING_OR_SEMICOLON = "(`([^`]*|(``))*`)|('([^']*|(''))*')";
+  private static final String QUOTED_STRING = "(`([^`]*|(``))*`)|('([^']*|(''))*')";
 
   private DdlDmlRequestValidators() {
   }
@@ -37,15 +37,18 @@ final class DdlDmlRequestValidators {
       return false;
     }
 
-    if (Arrays.stream(sql.split(QUOTED_STRING_OR_SEMICOLON))
-        .mapToInt(part -> part.split(";", -1).length - 1)
-        .sum() > 1
-    ) {
+    if (countStatements(sql) > 1) {
       cf.completeExceptionally(new KsqlClientException(
           "executeStatement() may only be used to execute one statement at a time."));
       return false;
     }
 
     return true;
+  }
+
+  private static int countStatements(final String sql) {
+    return Arrays.stream(sql.split(QUOTED_STRING))
+        .mapToInt(part -> part.split(";", -1).length - 1)
+        .sum();
   }
 }
--- a/ksqldb-api-client/src/main/java/io/confluent/ksql/api/client/impl/DdlDmlRequestValidators.java
+++ b/ksqldb-api-client/src/main/java/io/confluent/ksql/api/client/impl/DdlDmlRequestValidators.java
@@ -17,9 +17,12 @@ package io.confluent.ksql.api.client.impl;
 
 import io.confluent.ksql.api.client.ExecuteStatementResult;
 import io.confluent.ksql.api.client.exception.KsqlClientException;
+import java.util.Arrays;
 import java.util.concurrent.CompletableFuture;
 
 final class DdlDmlRequestValidators {
+
+  private static final String QUOTED_STRING_OR_SEMICOLON = "(`([^`]*|(``))*`)|('([^']*|(''))*')";
 
   private DdlDmlRequestValidators() {
   }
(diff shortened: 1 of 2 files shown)


ksqldb-api-client/src/main/java/io/confluent/ksql/api/client/impl/DdlDmlRequestValidators.java
```diff
@@ -34,7 +37,10 @@ static boolean validateExecuteStatementRequest(
       return false;
     }
 
-    if (sql.indexOf(";") != sql.lastIndexOf(";")) {
+    if (Arrays.stream(sql.split(QUOTED_STRING_OR_SEMICOLON))
```
vcrfxia, line 40: nit: improve readability by pulling this logic out into a method called `containsMultipleStatements(sql)` (or similar).

ksqldb-api-client/src/main/java/io/confluent/ksql/api/client/impl/DdlDmlRequestValidators.java
```diff
@@ -34,7 +37,10 @@ static boolean validateExecuteStatementRequest(
...
+        .mapToInt(part -> part.split(";", -1).length - 1)
```
vcrfxia, line 41: Why is `-1` being passed as the second argument rather than using the default `0`? We shouldn't reject statements with extra white space after the semicolon. Might even be nice to trim the resulting pieces to see if there really are multiple pieces on either side of the semicolon.

ksqldb-api-client/src/test/java/io/confluent/ksql/api/client/ClientTest.java
```diff
@@ -782,6 +782,31 @@ public void shouldHandleErrorResponseFromExecuteStatement() {
     assertThat(e.getCause().getMessage(), containsString("something bad"));
   }
 
+  @Test
+  public void shouldHandleSingleWithStatementMultipleSemicolonsFromExecuteStatement() throws Exception {
```
vcrfxia, line 786:
```suggestion
  public void shouldExecuteSingleStatementWithMultipleSemicolons() throws Exception {
```

ksqldb-api-client/src/test/java/io/confluent/ksql/api/client/ClientTest.java
```diff
@@ -782,6 +782,31 @@ public void shouldHandleErrorResponseFromExecuteStatement() {
...
+    // Given
+    final CommandStatusEntity entity = new CommandStatusEntity(
+        "C';'SAS;",
```
vcrfxia, line 789: nit: make this more realistic so as to not confuse readers (else someone might see this test and wonder why we want to support this). As one example, we could have a CSAS where one of the columns being selected is a constant string containing a semicolon.

vcrfxia (commented):
Thanks @jzaralim -- great stuff! Comments inline.

ksqldb-api-client/src/main/java/io/confluent/ksql/api/client/impl/DdlDmlRequestValidators.java
jzaralim: This counts the number of semicolons in outside of strings. The `-1` means that empty strings before or after a semicolon are counted. The name of the regex is a misnomer, which has been fixed.

ksqldb-api-client/src/main/java/io/confluent/ksql/api/client/impl/DdlDmlRequestValidators.java
```diff
@@ -22,7 +22,7 @@
 
 final class DdlDmlRequestValidators {
 
-  private static final String QUOTED_STRING_OR_SEMICOLON = "(`([^`]*|(``))*`)|('([^']*|(''))*')";
+  private static final String QUOTED_STRING = "(`([^`]*|(``))*`)|('([^']*|(''))*')";
```
vcrfxia, line 25: I'm confused. Is this saying that strings are allowed to be enclosed in backticks? I thought only single quotes were allowed for string literals.

ksqldb-api-client/src/main/java/io/confluent/ksql/api/client/impl/DdlDmlRequestValidators.java
vcrfxia: Thanks. To check my understanding (again): the purpose of the -1 is to ensure that there are always 2 pieces on either side of the semicolon, which means the length of the (split) array minus 1 is the number of semicolons in the statement? Can we add a short comment into the code to explain this for future readers?

Would be great to also add a short javadoc explaining exactly what this method checks. For example, this method does not validate that there are no hanging statements (text at the end not ended with a semicolon). This is fine since the client passes the original text to the server and the server will throw if such a statement is received, but it's not immediately clear to a reader that this is intentional rather than an oversight.

ksqldb-api-client/src/main/java/io/confluent/ksql/api/client/impl/DdlDmlRequestValidators.java
jzaralim: Backquoted identifiers can have semicolons in them.

ksqldb-api-client/src/main/java/io/confluent/ksql/api/client/impl/DdlDmlRequestValidators.java
jzaralim: Yup, that is correct. Will add a javadoc.

ksqldb-api-client/src/main/java/io/confluent/ksql/api/client/impl/DdlDmlRequestValidators.java
vcrfxia: Got it. Can we rename this variable to `QUOTED_STRING_OR_IDENTIFIER` in that case?

ksqldb-api-client/src/test/java/io/confluent/ksql/api/client/ClientTest.java
```diff
@@ -782,6 +782,31 @@ public void shouldHandleErrorResponseFromExecuteStatement() {
     assertThat(e.getCause().getMessage(), containsString("something bad"));
   }
 
+  @Test
+  public void shouldExecuteSingleStatementWithMultipleSemicolons() throws Exception {
+    // Given
+    final CommandStatusEntity entity = new CommandStatusEntity(
+        "CREATE STREAM FOO AS CONCAT(A, `wow;`) FROM `BAR`;  ",
```
vcrfxia, line 789: Let's also update this example to reflect proper usage of quoted strings, to avoid confusion (contributed to my question above).
```suggestion
        "CREATE STREAM FOO AS CONCAT(A, 'wow;') FROM `BAR`;  ",
```

vcrfxia (approved):
LGTM once the outstanding comments are addressed. Thanks for the fix, @jzaralim !

ksqldb-api-client/src/main/java/io/confluent/ksql/api/client/impl/DdlDmlRequestValidators.java
```diff
@@ -46,6 +46,16 @@ static boolean validateExecuteStatementRequest(
     return true;
   }
 
+  /**
+   * Counts the number of sql statements in a string by
+   *  1. Removing all of the sql strings
```
vcrfxia, line 51: nit: and identifiers.

Issue #7084, opened 2021-02-24
Support Pull Queries on Table-Table Joins


agavra:
**Is your feature request related to a problem? Please describe.**

Currently, ksqlDB materializes all table queries (see #6750), making nearly every table query-able through pull queries without any additional changes. This is not the case for Table-Table joins, because the pull query engine currently does not know how to merge the two materialized stores.

**Describe the solution you'd like**

- we could proactively materialize the result, and query the result directly (adding an additional state store)
- we could have the physical planner intelligently merge the results from querying two state stores

PR #7246, opened 2021-03-17
feat: Materialize Table-Table join results

cprasad1:
### Description 
Materialize Table-Table join results to enable pull queries on join results. #7084 

### Testing done 
- RQTT
- Unit Testing
- Manual Testing

### Reviewer checklist


guozhangwang (commented):
A meta question: if we do not provide a name, then KS would auto-generate one that increments the suffix index. Would this be a concern for existing queries compatibility?

ksqldb-execution/src/main/java/io/confluent/ksql/execution/plan/TableSelect.java
```diff
@@ -68,9 +71,14 @@ private TableSelect(
       @JsonProperty(value = "source", required = true) final ExecutionStep<KTableHolder<K>> source,
       @JsonProperty(value = "keyColumnNames") final Optional<List<ColumnName>> keyColumnNames,
       @JsonProperty(value = "selectExpressions", required = true) final
-      List<SelectExpression> selectExpressions
+      List<SelectExpression> selectExpressions,
+      @JsonProperty(value = "internalFormats", required = true) final Formats internalFormats
```
cprasad1, line 75: This will change ~650 historical plans

ksqldb-streams/src/main/java/io/confluent/ksql/execution/streams/TableSelectBuilder.java
```diff
@@ -56,20 +69,71 @@ private TableSelectBuilder() {
 
     final Named selectName = Named.as(StreamsUtil.buildOpName(queryContext));
 
+    final Optional<MaterializationInfo.Builder> matBuilder =  table.getMaterializationBuilder();
+
+    final boolean forceMaterialize = !matBuilder.isPresent();
```
cprasad1, line 74: heuristic for determining if we should force a materializing at this node or not

ksqldb-execution/src/main/java/io/confluent/ksql/execution/plan/TableSelect.java
vvcephei: Will KSQL still be able to load historical plans that don't include this field?

ksqldb-streams/src/main/java/io/confluent/ksql/execution/streams/TableSelectBuilder.java
```diff
@@ -56,20 +69,71 @@ private TableSelectBuilder() {
...
+
+    //Feauture flag for turning T-T Join queryability on or off
+    final boolean turnItOn = true;
```
vvcephei, line 77: Seems like this could have a better name ;) 

I.e., the name should say what the comment says. Then, you don't need the comment.

ksqldb-streams/src/main/java/io/confluent/ksql/execution/streams/TableSelectBuilder.java
vvcephei: Not sure I follow. It looks like the heuristic is that if it's not materialized, materialize it.

It looks like we did ultimately go the route of simply materializing every SELECT node, with the assumption that there will always be exactly one of these, and it will be at the end of the topology. Is that right?

vvcephei (commented):
Thanks for this, @cprasad1 !

Just a few comments.

ksqldb-streams/src/main/java/io/confluent/ksql/execution/streams/TableSelectBuilder.java
cprasad1: The heuristic is that if we can't fetch a state store from upstream, then Materialize it

vvcephei: Aaah. That makes more sense. Thanks!

cprasad1: It is an indirect way of saying only the select nodes at the end of Joins would be materialized (therefore, "heuristic")

ksqldb-engine/src/main/java/io/confluent/ksql/planner/plan/ProjectNode.java
```diff
@@ -45,7 +46,12 @@ public ProjectNode(
...
         getSelectExpressions(),
         buildContext.buildNodeContext(getId().toString()),
-        buildContext
+        buildContext,
+        getFormatInfo()
     );
   }
+
+  public FormatInfo getFormatInfo() {
+    return getLeftmostSourceNode().getDataSource().getKsqlTopic().getValueFormat().getFormatInfo();
```
AlanConfluent, line 55: Does this format hold for anything that the leftmost source may join with?

ksqldb-execution/src/main/java/io/confluent/ksql/execution/plan/PlanInfo.java
```diff
@@ -70,6 +70,7 @@ public PlanInfo merge(final PlanInfo other) {
     return new PlanInfo(allSources.merge(other.allSources));
   }
 
+
```
AlanConfluent, line 73: Stray line?

ksqldb-execution/src/main/java/io/confluent/ksql/execution/plan/TableSelect.java
AlanConfluent: I would think you would have to remove required = true to handle it without.  I thought the idea was that this should exist only for plans made by newer ksql versions.

AlanConfluent: I'm assuming that this is the internal format of the materialization if it's to be materialized?  Maybe it makes more sense to wrap it in another object called Materialization with format as a field if we think there could be more metadata?  Or just call this materializationFormat?

ksqldb-streams/src/main/java/io/confluent/ksql/execution/streams/TableSelectBuilder.java
AlanConfluent: Just so I understand, I assume that to be materialized, there must be two conditions met:
- Formats must be non-null, indicating that the plan came from a newer version of ksql which wants to materialize such a join (and not an old ksql which wouldn't have done this)
- The topology is such that there isn't a state store upstream already and so we would want one here.

Is that correct?

ksqldb-streams/src/main/java/io/confluent/ksql/execution/streams/TableSelectBuilder.java
```diff
@@ -26,18 +29,28 @@
...
 
   @SuppressWarnings("unchecked")
   public static <K> KTableHolder<K> build(
-      final KTableHolder<K> table,
-      final TableSelect<K> step,
-      final RuntimeBuildContext buildContext
+          final KTableHolder<K> table,
+          final TableSelect<K> step,
+          final RuntimeBuildContext buildContext,
+          final Formats formats,
```
AlanConfluent, line 52: Should formats be Optional?  It will only exist if the plan comes from a new version of kql, right?

cprasad1: I have made it Optional for the constructor that is invoked by Jackson

ksqldb-streams/src/main/java/io/confluent/ksql/execution/streams/TableSelectBuilder.java
cprasad1: yes, @AlanConfluent that's correct

ksqldb-functional-tests/src/test/resources/rest-query-validation-tests/pull-queries-against-materialized-table-table-joins.json
```diff
@@ -0,0 +1,466 @@
...
+        {"topic": "right_topic", "key": 15, "value": {"F1": "c", "F2": 20}, "timestamp": 15500},
+        {"topic": "left_topic", "key": 0, "value": {"NAME": "bar", "VALUE": 99}, "timestamp": 16000}
+      ],
+      "responses": [
+        {"admin": {"@type": "currentStatus"}},
+        {"admin": {"@type": "currentStatus"}},
+        {"admin": {"@type": "currentStatus"}},
+        {"query": [
+          {"header":{"schema":"`ID` BIGINT KEY, `NAME` STRING, `VALUE` BIGINT, `F1` STRING, `F2` BIGINT"}},
+          {"row":{"columns":[15, null, null, "c", 20]}},
```
guozhangwang, line 228: Hmm.. why the keys are not sorted here? Does ordering actually not matter in the test format?

ksqldb-functional-tests/src/test/resources/rest-query-validation-tests/pull-queries-against-materialized-table-table-joins.json
```diff
@@ -0,0 +1,466 @@
...
+          {"header":{"schema":"`ID` BIGINT KEY, `T_ID` BIGINT, `TT_ID` BIGINT, `NAME` STRING, `VALUE` BIGINT, `F1` STRING, `F2` BIGINT"}},
+          {"row":{"columns":[1, 1, 1, "bar", 99, "a", 10]}},
+          {"row":{"columns":[15, 0, 15, "" , 0, "c", 20]}},
+          {"row":{"columns":[10, 10, 0, "100", 5, "", 0]}}
+        ]}
+      ]
+    },
+    {
+      "name": "table join pipeline",
+      "format": ["JSON"],
```
guozhangwang, line 265: Ditto, why not include AVRO as well?

ksqldb-functional-tests/src/test/resources/rest-query-validation-tests/pull-queries-against-materialized-table-table-joins.json
```diff
@@ -0,0 +1,466 @@
...
+
+        ]}
+      ]
+    },
+    {
+      "name": "table table join with where clause",
+      "statements": [
+        "CREATE TABLE TEST (ID BIGINT PRIMARY KEY, NAME varchar, VALUE bigint) WITH (kafka_topic='left_topic', value_format='JSON');",
+        "CREATE TABLE TEST_TABLE (ID BIGINT PRIMARY KEY, F1 varchar, F2 bigint) WITH (kafka_topic='right_topic', value_format='JSON');",
+        "CREATE TABLE OUTPUT as SELECT t.id, name, tt.f1, f2 FROM test t JOIN test_table tt ON t.id = tt.id WHERE t.value > 10 AND tt.f2 > 5;",
```
guozhangwang, line 306: Is `table table inner join - filter in topology` subsumed by this test?

ksqldb-functional-tests/src/test/resources/rest-query-validation-tests/pull-queries-against-materialized-table-table-joins.json
```diff
@@ -0,0 +1,466 @@
...
+        ]}
+      ]
+    },
+    {
+      "name": "table-table-table - inner-full",
+      "statements": [
+        "CREATE TABLE T1 (ID INT PRIMARY KEY, V0 bigint) WITH (kafka_topic='left', value_format='JSON');",
+        "CREATE TABLE T2 (ID INT PRIMARY KEY, V0 bigint) WITH (kafka_topic='right', value_format='JSON');",
+        "CREATE TABLE T3 (ID INT PRIMARY KEY, V0 bigint) WITH (kafka_topic='right2', value_format='JSON');",
+        "CREATE TABLE OUTPUT as SELECT ROWKEY AS ID, T1.V0, T2.V0, T3.V0 FROM T1 JOIN T2 ON T1.ID = T2.ID FULL JOIN T3 ON T1.ID = T3.ID;",
```
guozhangwang, line 439: What's the semantic difference between `FULL JOIN` and `OUTER JOIN`?

ksqldb-streams/src/main/java/io/confluent/ksql/execution/streams/TableSelectBuilder.java
guozhangwang: Sounds good to have the simple heuristic for now. We can discuss about if we want to be smarter with trade-offs between materialization and query serving in the near future.

guozhangwang (commented):
Made another pass. Thanks @cprasad1 !

Just wondering, if we are going to update `TableFilter` for filter-after-joins in another PR?

ksqldb-streams/src/main/java/io/confluent/ksql/execution/streams/TableSelectBuilder.java
cprasad1: I will try

ksqldb-functional-tests/src/test/resources/rest-query-validation-tests/pull-queries-against-materialized-table-table-joins.json
cprasad1: yes, good catch

ksqldb-functional-tests/src/test/resources/rest-query-validation-tests/pull-queries-against-materialized-table-table-joins.json
```diff
@@ -0,0 +1,466 @@
...
+        {"query": [
+          {"header":{"schema":"`T_ID` BIGINT KEY, `NAME` STRING, `VALUE` BIGINT, `F1` STRING, `F2` BIGINT"}},
+          {"row":{"columns":[0,"bar",99,"a",10]}},
+          {"row":{"columns":[10,"100",5,"",0]}},
+          {"row":{"columns":[90,"ninety",90,"",0]}}
+        ]}
+      ]
+    },
+    {
+      "name": "table table inner join",
```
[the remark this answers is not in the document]
cprasad1: good catch

ksqldb-functional-tests/src/test/resources/rest-query-validation-tests/pull-queries-against-materialized-table-table-joins.json
cprasad1: technically, we only support `INNER`, `LEFT OUTER`, and `FULL OUTER` joins between tables. So, depends on what you mean by `OUTER JOIN`. Source: https://docs.ksqldb.io/en/latest/developer-guide/joins/join-streams-and-tables/#table-table-joins

ksqldb-engine/src/main/java/io/confluent/ksql/planner/plan/ProjectNode.java
cprasad1: @guozhangwang yes, it is sufficient

ksqldb-functional-tests/src/test/resources/rest-query-validation-tests/pull-queries-against-materialized-table-table-joins.json
guozhangwang: I was just confused to see `FULL JOIN` here (because elsewhere we use the full name `FULL OUTER JOIN`) and the name of the test is `inner-full` while other test names say `outer join`, and wonder if these are different. 

So is `OUTER JOIN` just a syntax sugar for `FULL OUTER JOIN`, similar to `JOIN` being a syntax sugar of `INNER JOIN`?

ksqldb-functional-tests/src/test/resources/rest-query-validation-tests/pull-queries-against-materialized-table-table-joins.json
cprasad1: added at the top.

ksqldb-functional-tests/src/test/resources/rest-query-validation-tests/pull-queries-against-materialized-table-table-joins.json
```diff
@@ -0,0 +1,466 @@
...
+        {"admin": {"@type": "currentStatus"}},
+        {"query": [
+          {"header":{"schema":"`T_ID` BIGINT KEY, `NAME` STRING, `VALUE` BIGINT, `F1` STRING, `F2` BIGINT"}},
+          {"row":{"columns":[0,"bar",99,"a",10]}}
+        ]}
+      ]
+    },
+    {
+      "name": "table table outer join",
+      "format": ["JSON"],
```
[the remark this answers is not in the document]
cprasad1: added at the top.

ksqldb-streams/src/main/java/io/confluent/ksql/execution/streams/TableSelectBuilder.java
cprasad1: done

ksqldb-functional-tests/src/test/resources/rest-query-validation-tests/pull-queries-against-materialized-table-table-joins.json
cprasad1: yes, I believe that is correct

ksqldb-functional-tests/src/test/resources/rest-query-validation-tests/pull-queries-against-materialized-table-table-joins.json
```diff
@@ -0,0 +1,466 @@
...
+        {"topic": "left_topic", "key": 0, "value": {"NAME": "bar", "VALUE": 99}, "timestamp": 16000},
+        {"topic": "left_topic", "key": 90, "value": {"NAME": "ninety", "VALUE": 90}, "timestamp": 17000}
+      ],
+      "responses": [
+        {"admin": {"@type": "currentStatus"}},
+        {"admin": {"@type": "currentStatus"}},
+        {"admin": {"@type": "currentStatus"}},
+        {"query": [
+          {"header":{"schema":"`T_ID` BIGINT KEY, `NAME` STRING, `VALUE` BIGINT, `F1` STRING, `F2` BIGINT"}},
+          {"row":{"columns":[0,"bar",99,"a",10]}},
```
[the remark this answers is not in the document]
cprasad1: Protobuf handles null values differently than AVRO and JSON. Protobuf doesn't have the concept of a null value, so the conversion between PROTOBUF and Java (Kafka Connect) objects is undefined. Usually, Protobuf resolves a "missing field" to the default value of its type, and I believe the default values we have in ksqlDB are:
String: the default value is the empty string.
Byte: the default value is empty bytes.
Bool: the default value is false.
Numeric type: the default value is zero.
Enum: the default value is the first defined enum value, which must be zero.
Message field: the field is not set. Its exact value is language-dependent.

ksqldb-functional-tests/src/test/resources/rest-query-validation-tests/pull-queries-against-materialized-table-table-joins.json
cprasad1: The ordering does matter in the test format. For a single partition case the order is always certain (That's why the test cases pass every time). The definitive order is determined by the state store itself. I don't know the underlying data structure that stores all the records, so I can't speak to the logic behind the ordering. It is 'random' in a definite way. 
For multiple partitions, there is no determined ordering because of async processing

ksqldb-functional-tests/src/test/resources/rest-query-validation-tests/pull-queries-against-materialized-table-table-joins.json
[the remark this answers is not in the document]
guozhangwang: Thanks!

ksqldb-functional-tests/src/test/resources/rest-query-validation-tests/pull-queries-against-materialized-table-table-joins.json
guozhangwang: In that case, why it is ordered as 15 -> 0 -> 10? Shouldn't it be 0 -> 10 -> 15?

cprasad1: The ordering is definitive, not sorted

guozhangwang: From my understanding on the table-table join logic, the emitted records' keys should be 0, 10, 15; hence I'm curious to learn why it is 15,0,10.

cprasad1: yes, but it gets stored in a state store from which pull queries fetch the records. We don't offer any guarantees on the ordering in which the state stores serves the records to the pull query afaik. We can dig into it, but there is a plausible explanation for why the pull query ordering doesnt match the persistent query ordering.

cprasad1: I guess a higher order question/task would be: do we want to offer any ordering guarantees around pull query records? Should they match persistent query ordering in some way?

cprasad1: @AlanConfluent thoughts?

guozhangwang: Since this test is on a single partition, the ordering is dependent on that single state store's range query -- although we do not commit to any ordering today, our current implementation on the range query on persistent rocksDB does indeed return records by key order. That's why I'm curious why only this single test returns results not in key order, while all other tests seem to return results in key order.

ksqldb-functional-tests/src/test/resources/rest-query-validation-tests/pull-queries-against-materialized-table-table-joins.json
```diff
@@ -0,0 +1,468 @@
...
+        {"admin": {"@type": "currentStatus"}},
+        {"query": [
+          {"header":{"schema":"`T_ID` BIGINT KEY, `NAME` STRING, `VALUE` BIGINT, `F1` STRING, `F2` BIGINT"}},
+          {"row":{"columns":[100,"100",5,"c",20]}}
+        ]}
+      ]
+    },
+    {
+      "name": "table table left join",
+      "format": ["AVRO", "JSON"],
```
AlanConfluent, line 121: Presumably this tests first avro for both types and then json for both types.  I remember we were wondering if it was possible to join different types, one avro, one json because the code seemed to assume they were the same format.  Might be good to verify that it's disallowed.

ksqldb-functional-tests/src/test/resources/rest-query-validation-tests/pull-queries-against-materialized-table-table-joins.json
AlanConfluent: We generally don't have ordering guarantees since we do requests in parallel and they stream back results.  In this case, if it's a single partition, it should presumably be the ordering of the underlying store.  I'm honestly not sure why it wouldn't be sorted, though in tests, they're certainly consistent.  We could spend a few minutes looking into this if we're curious, but it doesn't seem like a big deal since we can't make guarantees anyway.

AlanConfluent (approved):

ksqldb-functional-tests/src/test/resources/rest-query-validation-tests/pull-queries-against-materialized-table-table-joins.json
guozhangwang: I guess I'm just curious since 1) this result is consistent, and 2) all other test results are ordered indeed, and I cannot think of a reason why it is different from other results by just looking at the statements. But I agree that at the moment we do not provide any guarantees anyways. So this question is mainly for our own understanding.

cprasad1: @guozhangwang there are other tests that have unsorted ordering as well

guozhangwang: I spent some more time thinking about it and still cannot come up with the potential root cause (I looked at the source code at KS too but did not find anything there). Filed https://github.com/confluentinc/ksql/issues/7391.

guozhangwang: Just to clarify, this should not be a blocker for merging this PR as is, we could just investigate it separately as we continue making progress on this one.

guozhangwang (approved):
LGTM!

PR #7262, opened 2021-03-17
chore: add rocksdbjni as explicit dependency to config setter module

vcrfxia:
### Description 

The master build is currently failing because the ksqldb-rocksdb-config-setter module can't find dependencies from `org.rocksdb`. This dependency was previously being pulled in transitively from the module's kafka-streams dependency but this no longer appears to be the case. (The transitive dependency is now pulled in with scope `runtime` which I believe explains the compilation error.) This PR fixes the build by adding the dependency explicitly. 

~~Weirdly, I can't find any recent changes to explain the change in behavior; if anyone else has more ideas for where to look that'd be great. The build started failing between 10:18am and 11:51am PST today.~~

Update: mystery has been solved (thanks, Sophie!). Here's the culprit (a change in Streams's transitive dependency handling): https://github.com/confluentinc/kafka/pull/536/commits/7a3ebbebbc6aed2049f22c650a52f6f685546207#diff-49a96e7eea8a94af862798a45174e6ac43eb4f8b4bd40759b5da63ba31ec3ef7R1428

### Testing done 

Build passes.

### Reviewer checklist

### DIFF
--- a/ksqldb-rocksdb-config-setter/pom.xml
+++ b/ksqldb-rocksdb-config-setter/pom.xml
@@ -36,6 +36,12 @@
         </dependency>
 
         <dependency>
+            <groupId>org.rocksdb</groupId>
+            <artifactId>rocksdbjni</artifactId>
+            <version>5.18.4</version>
+        </dependency>
+
+        <dependency>
             <groupId>junit</groupId>
             <artifactId>junit</artifactId>
             <scope>test</scope>


ksqldb-rocksdb-config-setter/pom.xml
```diff
@@ -35,6 +35,12 @@
             <artifactId>kafka-streams</artifactId>
         </dependency>
 
+        <dependency>
+            <groupId>org.rocksdb</groupId>
+            <artifactId>rocksdbjni</artifactId>
+            <version>5.18.4</version>
```
stevenpyzhang, line 41: nit:  we could put this in the properties section in the ksql parent pom

stevenpyzhang (approved):
🤷‍♂️  LGTM

ableegoldman:
I think it was probably [this ](https://github.com/apache/kafka/pull/10203) which was just synced over from AK earlier today.

guozhangwang (approved):

ksqldb-rocksdb-config-setter/pom.xml
vcrfxia: Normally I'd agree but I don't feel strongly in this case since ksqlDB shouldn't be in the business of interacting with RocksDB directly (we should be going through Streams APIs for that) which means this dependency is specific to this config setter module. 

OTOH, it's very little effort to move it and maybe we'll add more special modules later. I'll move it.

ableegoldman (approved):
I'm pretty concerned that we would need to pull in this dependency manually from now on, so we might need to revert this change in AK. But that will probably take a while to sync over so we should go ahead and merge this to unblock the build for now -- thanks for the PR

ableegoldman:
Update: we can revert this once https://github.com/apache/kafka/pull/10341 is merged and has been synced over to ccs-kafka

PR #7264, opened 2021-03-18
test: fix MigrationsTest to not fail while waiting for pull query materialization

vcrfxia:
### Description 

MigrationsTest is currently flaky in that if the migrations metadata table is not yet ready to serve pull queries, an error will be returned from `RestIntegrationTestUtil.makeQueryRequest()` when a pull query is issued, which exits the `assertThatEventually()` call rather than triggering a retry as desired. This PR fixes the flakiness by wrapping the pull query request in order to avoid this and allow the retry mechanism to work as intended.

### Testing done 

Test-only change.

### Reviewer checklist


jzaralim (approved):

agavra (approved):
thanks for the fix @vcrfxia!

abbccdda (approved):

Issue #7184, opened 2021-03-09
An NPE while backing the metadata causes KSQL fail to start


spena:
**Describe the bug**
If a null key or value is being written to the command topic, then the backup process will fail with an NPE causing the KSQL to stop working. 

This is the stack trace of the NPE:
```
java.lang.NullPointerException
	at java.base/java.io.FileOutputStream.write(FileOutputStream.java:341)
	at io.confluent.ksql.rest.server.BackupReplayFile.write(BackupReplayFile.java:87)
	at io.confluent.ksql.rest.server.CommandTopicBackupImpl.writeRecord(CommandTopicBackupImpl.java:150)
	at io.confluent.ksql.rest.server.CommandTopic.backupRecord(CommandTopic.java:159)
	at io.confluent.ksql.rest.server.CommandTopic.getRestoreCommands(CommandTopic.java:117)
	at io.confluent.ksql.rest.server.computation.CommandStore.getRestoreCommands(CommandStore.java:268)
	at io.confluent.ksql.rest.server.computation.CommandRunner.processPriorCommands(CommandRunner.java:254)
	at io.confluent.ksql.rest.server.KsqlRestApplication.initialize(KsqlRestApplication.java:459)
	at io.confluent.ksql.rest.server.KsqlRestApplication.startKsql(KsqlRestApplication.java:385)
	at io.confluent.ksql.rest.server.KsqlRestApplication.startAsync(KsqlRestApplication.java:367)
	at io.confluent.ksql.rest.server.KsqlServerMain.tryStartApp(KsqlServerMain.java:89)
	at io.confluent.ksql.rest.server.KsqlServerMain.main(KsqlServerMain.java:64)
```

**To Reproduce**
KSQL version is `0.15`
Steps to reproduce TBD

**Expected behavior**
The KSQL server should not stop working. The backups should be marked as degradation so a warning is displayed to the user.

vcrfxia:
root cause + add logging.

spena:
I couldn't find the root cause of this null key in the command topic. The logs do not contain any new commands sent by the user. The Kafka logs does not include records written either. This could be probably caused by a producer used by the user to write directly to the command topic, thus causing the server to crash.

PR #7268, opened 2021-03-22
fix: fix NPE when backing a record that has null key/values

spena:
### Description 

Fixes https://github.com/confluentinc/ksql/issues/7184

### Testing done 

### Reviewer checklist

### DIFF
--- a/ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/BackupReplayFile.java
+++ b/ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/BackupReplayFile.java
@@ -84,15 +84,6 @@ public final class BackupReplayFile implements Closeable {
       throw new IOException("Write permission denied.");
     }
 
-    if (record == null) {
-      throw new IllegalArgumentException("Record is null");
-    }
-
-    if (record.key() == null || record.value() == null) {
-      throw new IllegalArgumentException(String.format(
-          "Record contain a null key/value: key=%s, value=%s", record.key(), record.value()));
-    }
-
     writer.write(record.key());
     writer.write(KEY_VALUE_SEPARATOR_BYTES);
     writer.write(record.value());
--- a/ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/BackupReplayFile.java
+++ b/ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/BackupReplayFile.java
@@ -84,6 +84,15 @@ public final class BackupReplayFile implements Closeable {
       throw new IOException("Write permission denied.");
     }
 
+    if (record == null) {
+      throw new IllegalArgumentException("Record is null");
+    }
+
+    if (record.key() == null || record.value() == null) {
+      throw new IllegalArgumentException(String.format(
+          "Record contain a null key/value: key=%s, value=%s", record.key(), record.value()));
+    }
+
     writer.write(record.key());
     writer.write(KEY_VALUE_SEPARATOR_BYTES);
     writer.write(record.value());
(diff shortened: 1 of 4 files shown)


ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/BackupReplayFile.java
```diff
@@ -84,6 +84,15 @@ public void write(final ConsumerRecord<byte[], byte[]> record) throws IOExceptio
       throw new IOException("Write permission denied.");
     }
 
+    if (record == null) {
+      throw new IllegalArgumentException("Record is null");
+    }
+
+    if (record.key() == null || record.value() == null) {
+      throw new IllegalArgumentException(String.format(
+          "Record contain a null key/value: key=%s, value=%s", record.key(), record.value()));
```
agavra, line 93: it might also help to output the partition/offset (even though partition is always 0 here) - ditto for other error messages

ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/BackupReplayFile.java
```diff
@@ -84,6 +84,15 @@ public void write(final ConsumerRecord<byte[], byte[]> record) throws IOExceptio
       throw new IOException("Write permission denied.");
     }
 
+    if (record == null) {
+      throw new IllegalArgumentException("Record is null");
```
agavra, line 88: when can this ever happen? can we throw a more useful error message?

agavra (approved):

ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/BackupReplayFile.java
guozhangwang: Because any records read from the cmd topic should not have null records, I feel this can only happen if the local file was manipulated somehow?

ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/CommandTopicBackupImpl.java
```diff
@@ -132,6 +132,17 @@ public void writeRecord(final ConsumerRecord<byte[], byte[]> record) {
           "Failed to write record due to out of sync command topic and backup file: " + record);
     }
 
+    if (record == null || record.key() == null || record.value() == null) {
+      String recordString = null;
+      if (record != null) {
+        recordString = String.format("key=%s, value=%s", record.key(), record.value());
+      }
+
+      LOG.warn("Cannot write a null record or null key/value to the backup file. This could be "
```
guozhangwang, line 141: Ditto here: could this ever happen, if we already check for null record/key/value when writing to cmd topic?

ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/BackupReplayFile.java
spena: Done

ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/BackupReplayFile.java
spena: I don't think this never happen. I added just in case, but seems unnecessary. I removed it.

ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/CommandTopicBackupImpl.java
spena: I removed the redundant null check from the BackupReplayFile, which is called by the CommandTopicBackupImpl. I just left the one from this file.

PR #7271, opened 2021-03-22
feat: enable variable substitution for /query-stream and /ksql endpoints

jzaralim:
### Description 
Adds an extra parameter, `sessionVariables` to the /query-stream and /ksql endpoints so that users can pass in a map of variable substitutions and use them in statements. For `/ksql`, `sessionVariables` is a map of initial variable values that can be modified with `define` and `undefine` statements.

This is a prerequisite for enabling variable substitutions in the java client.

### Testing done 
integration and unit tests

### Reviewer checklist

### DIFF
--- a/ksqldb-rest-app/src/test/java/io/confluent/ksql/api/integration/ApiIntegrationTest.java
+++ b/ksqldb-rest-app/src/test/java/io/confluent/ksql/api/integration/ApiIntegrationTest.java
@@ -175,6 +175,22 @@ public class ApiIntegrationTest {
   }
 
   @Test
+  public void shouldExecutePushQueryWithVariableSubstitution() {
+
+    // Given:
+    String sql = "SELECT DEC AS ${name} from " + TEST_STREAM + " EMIT CHANGES LIMIT 2;";
+
+    // When:
+    QueryResponse response = executeQueryWithVariables(sql, new JsonObject().put("name", "COL"));
+
+    // Then:
+    assertThat(response.rows, hasSize(2));
+    assertThat(response.responseObject.getJsonArray("columnNames"), is(new JsonArray().add("COL")));
+    assertThat(response.responseObject.getJsonArray("columnTypes"), is(new JsonArray().add("DECIMAL(4, 2)")));
+    assertThat(response.responseObject.getString("queryId"), is(notNullValue()));
+  }
+
+  @Test
   public void shouldFailPushQueryWithInvalidSql() {
 
     // Given:
@@ -613,9 +629,13 @@ public class ApiIntegrationTest {
   }
 
   private QueryResponse executeQuery(final String sql) {
+    return executeQueryWithVariables(sql, new JsonObject());
+  }
+
+  private QueryResponse executeQueryWithVariables(final String sql, final JsonObject variables) {
     JsonObject properties = new JsonObject();
     JsonObject requestBody = new JsonObject()
-        .put("sql", sql).put("properties", properties);
+        .put("sql", sql).put("properties", properties).put("sessionVariables", variables);
     HttpResponse<Buffer> response = sendRequest("/query-stream", requestBody.toBuffer());
     return new QueryResponse(response.bodyAsString());
   }
--- a/ksqldb-rest-app/src/test/java/io/confluent/ksql/api/integration/ApiIntegrationTest.java
+++ b/ksqldb-rest-app/src/test/java/io/confluent/ksql/api/integration/ApiIntegrationTest.java
@@ -307,6 +307,35 @@ public class ApiIntegrationTest {
   }
 
   @Test
+  public void shouldExecutePullQueryWithVariableSubstitution() {
+
+    // Given:
+    String sql = "SELECT * from ${AGG_TABLE} WHERE K=" + AN_AGG_KEY + ";";
+    final JsonObject variables = new JsonObject().put("AGG_TABLE", AGG_TABLE);
+
+    // When:
+    // Maybe need to retry as populating agg table is async
+    AtomicReference<QueryResponse> atomicReference = new AtomicReference<>();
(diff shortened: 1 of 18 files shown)


vcrfxia:
Hey @jzaralim , thanks for taking this on. Do you have a writeup somewhere for what the end state will look like, in terms of what variable substitution will look like via HTTP and also the Java client? I ask because the proposal in this PR seems to differ from the earlier [KLIP](https://github.com/confluentinc/ksql/blob/master/design-proposals/klip-38-variable-substitution.md) which envisions variable substitution via HTTP requests as happening through the use of `DEFINE` and `UNDEFINE` statements, and proposes that the CLI and Java client perform substitutions prior to sending statements to the server. It'd be good to get alignment on what we're building towards before investing in code changes to ensure everyone's on the same page.

jzaralim:
@vcrfxia Sure, I added up update to klip-38 here: https://github.com/confluentinc/ksql/pull/7288

ksqldb-rest-app/src/test/java/io/confluent/ksql/api/integration/ApiIntegrationTest.java
```diff
@@ -174,6 +174,22 @@ public void shouldExecutePushQueryWithLimit() {
...
+    // When:
+    QueryResponse response = executeQueryWithVariables(sql, new JsonObject().put("name", "COL"));
+
+    // Then:
+    assertThat(response.rows, hasSize(2));
+    assertThat(response.responseObject.getJsonArray("columnNames"), is(new JsonArray().add("COL")));
+    assertThat(response.responseObject.getJsonArray("columnTypes"), is(new JsonArray().add("DECIMAL(4, 2)")));
+    assertThat(response.responseObject.getString("queryId"), is(notNullValue()));
+  }
+
```
vcrfxia, line 192: Let's also add a pull query integration test, for completeness.

ksqldb-rest-app/src/test/java/io/confluent/ksql/rest/integration/RestApiTest.java
```diff
@@ -674,6 +674,22 @@ public void shouldDeleteTopic() {
     assertThat("Expected topic X to be deleted", !topicExists("X"));
   }
 
+  @Test
+  public void shouldCreateStreamWithVariableSubstitution() {
+    // Given:
+    // When:
+    makeKsqlRequestWithVariables(
+        "CREATE STREAM X AS SELECT * FROM " + PAGE_VIEW_STREAM + " WHERE USERID='${id}';",
```
vcrfxia, line 682: This test does not actually validate that variable substitution works. The test would succeed just as well if variable substitution were not supported and `'${id}'` were simply treated as a string key.

ksqldb-rest-model/src/main/java/io/confluent/ksql/rest/SessionProperties.java
```diff
@@ -36,6 +37,27 @@
...
   private final Map<String, String> sessionVariables;
 
+  /**
+   * @param mutableScopedProperties   The streamsProperties of the incoming request
+   * @param ksqlHostInfo              The ksqlHostInfo of the server that handles the request
+   * @param localUrl                  The url of the server that handles the request
+   * @param internalRequest           Flag indicating if request is from within the KSQL cluster
+   * @param sessionVariables          Initial session variables
+   */
+  public SessionProperties(
```
vcrfxia, line 47: Is this used?

vcrfxia (commented):
Thanks @jzaralim -- LGTM overall! Some notes on test coverage inline. Can we also update the docs for this?

ksqldb-rest-model/src/main/java/io/confluent/ksql/rest/SessionProperties.java
jzaralim: Yup, https://github.com/confluentinc/ksql/blob/0c34efeb57d37ae51b4241f64cfad5e1cb2ea6ea/ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/resources/KsqlResource.java#L283

ksqldb-rest-model/src/main/java/io/confluent/ksql/rest/SessionProperties.java
```diff
@@ -36,6 +37,27 @@
...
+      final Map<String, Object> mutableScopedProperties,
+      final KsqlHostInfo ksqlHostInfo,
+      final URL localUrl,
+      final boolean internalRequest,
+      final Map<String, Object> sessionVariables
+  ) {
+    this(mutableScopedProperties, ksqlHostInfo, localUrl, internalRequest);
```
vcrfxia, line 54: nit: can we have the other constructor call this one instead? I think constructor chains are more intuitive / easier to read when the ones with fewer arguments call the ones with more arguments (with intuitive defaults) since then you can just read the code in the final constructor in the chain in order to understand what's happening.

ksqldb-rest-model/src/main/java/io/confluent/ksql/rest/SessionProperties.java
```diff
@@ -36,6 +37,27 @@
...
+    if (sessionVariables != null) {
```
vcrfxia, line 55: Is there ever a valid use case for this to be null? If not, let's use `Objects.requireNonNull()` to validate that. If so, we should consider passing an `Optional` instead to make it explicit.

vcrfxia (approved):
Thanks @jzaralim -- LGTM minus the two minor comments inline, plus the request for docs from earlier. (Approving now so you can merge without docs if you have other changes blocked behind this one, but ideally the docs changes would be part of the same PR since they aid in a reader's understanding of the PR.)

PR #7288, opened 2021-03-24
docs: add details for java client to klip-38

jzaralim:
### Description 
As the title says.

### Reviewer checklist

### DIFF
--- a/design-proposals/klip-38-variable-substitution.md
+++ b/design-proposals/klip-38-variable-substitution.md
@@ -264,7 +264,21 @@ Scope: SESSION
 In Java API client, the session is opened when `Client.create()` is called, and closed when `Client.close()` is called. Variables
 may be defined using a client method instead of a new syntax. This is to ensure users know the scope starts after the `create()` call.
 
-TBD: Define a Java API method for variable substitution.
+Users will define variables using the functions, `define(variable_name, value)` and `undefine(variable_name)`.
+The client will keep track of the calls made in a map. Alternatively, the `executeStatement` and `executeQuery`
+functions could accept a third parameter, `sessionVariables` that is a map of variables to variable definitions.
+
+There are two options for executing the variable execution. The first is to do it in the client by
+using the variable substitutor in `ksqldb-parser`. The disadvantages to this are that `ksqldb-parser`
+would be a dependancy of the Java client, and that the client and server versions would have
+to match in order for the parser to work.
+
+The other option is to update the endpoints, `/ksql` and `/query-streams` to accept `sessionVariables`
+as a parameter. The variable substitution would then be done on the server side. The disadvantage is
+that an API change would be required.
+
+Because the Java client is meant to be lightweight and flexible, the disadvantages of the first option
+are stronger than the disadvantage of the second, so we will go with the second option.
 
 Note: Users can use variables and replace variables in strings using Java language features too.
 
--- a/design-proposals/klip-38-variable-substitution.md
+++ b/design-proposals/klip-38-variable-substitution.md
@@ -269,12 +269,12 @@ The client will keep track of the calls made in a map. Alternatively, the `execu
 `insertInto`, `streamInserts` and `createConnector` functions could accept a third parameter, `sessionVariables`
 that is a map of variables to variable definitions.
 
-There are two options for executing the variable execution. The first is to do it in the client by
+There are two options for executing the variable substitution. The first is to do it in the client by
 using the variable substitutor in `ksqldb-parser`. The disadvantages to this are that `ksqldb-parser`
-would be a dependancy of the Java client, and that the client and server versions would have
+would be a dependency of the Java client, and that the client and server versions would have
 to match in order for the parser to work.
 
-The other option is to update the endpoints `/ksql`, `/inserts-stream` and `/query-streams` to accept `sessionVariables`
+The other option is to update the endpoints `/ksql`, `/inserts-stream` and `/query-stream` to accept `sessionVariables`
 as a parameter. The variable substitution would then be done on the server side. The disadvantage is
 that an API change would be required.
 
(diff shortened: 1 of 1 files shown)


design-proposals/klip-38-variable-substitution.md
```diff
@@ -264,7 +264,21 @@ Scope: SESSION
 In Java API client, the session is opened when `Client.create()` is called, and closed when `Client.close()` is called. Variables
 may be defined using a client method instead of a new syntax. This is to ensure users know the scope starts after the `create()` call.
 
-TBD: Define a Java API method for variable substitution.
+Users will define variables using the functions, `define(variable_name, value)` and `undefine(variable_name)`.
+The client will keep track of the calls made in a map. Alternatively, the `executeStatement` and `executeQuery`
```
vcrfxia, line 268: Don't we also want to support variable substitution in other methods as well, such as `insertInto()`, `streamInserts()`, and `createConnector()`, just to name a few? I don't think it makes sense to essentially double the number of methods on the Java client in order to add versions that accept a map of variables, so I'm a fan of `define()` and `undefine()` instead.

design-proposals/klip-38-variable-substitution.md
```diff
@@ -264,7 +264,21 @@ Scope: SESSION
...
+functions could accept a third parameter, `sessionVariables` that is a map of variables to variable definitions.
+
+There are two options for executing the variable execution. The first is to do it in the client by
+using the variable substitutor in `ksqldb-parser`. The disadvantages to this are that `ksqldb-parser`
```
vcrfxia, line 272: I haven't looked into the code for this at all but is it not possible to separate out parser functionality for variable substitution into a library in ksqldb-common? Or is the variable substitution somehow intertwined with the actual statement parsing itself.

design-proposals/klip-38-variable-substitution.md
```diff
@@ -264,7 +264,21 @@ Scope: SESSION
...
+would be a dependancy of the Java client, and that the client and server versions would have
+to match in order for the parser to work.
+
+The other option is to update the endpoints, `/ksql` and `/query-streams` to accept `sessionVariables`
```
vcrfxia, line 276: As mentioned above, there are other endpoints that we'd want to support variable substitution on as well. Definitely `/inserts-stream` for now, potentially others in the future.

vcrfxia (commented):
Thanks @jzaralim for starting this conversation in the KLIP. Some thoughts/questions inline.

design-proposals/klip-38-variable-substitution.md
jzaralim: Added `/inserts-stream` to the klip. The other two endpoints used by the client, `/close-query` and `/info` will be left as is (although technically the user could pass in a variable to `/close-query`, I'm not sure how useful supporting this actually is.)

design-proposals/klip-38-variable-substitution.md
jzaralim: The statements have to be parsed because the substitutor needs to be able to differentiate between literals, identifiers and keywords (substitution is not allowed for keywords).

vcrfxia: Ah bummer... in light of that, if we think this is the least of all evils and @colinhicks is OK with the added complexity to our HTTP endpoints, then I'm happy with this approach.

design-proposals/klip-38-variable-substitution.md
```diff
@@ -264,7 +264,22 @@ Scope: SESSION
...
 
-TBD: Define a Java API method for variable substitution.
+Users will define variables using the functions, `define(variable_name, value)` and `undefine(variable_name)`.
+The client will keep track of the calls made in a map. Alternatively, the `executeStatement`, `executeQuery`, `streamQuery`
+`insertInto`, `streamInserts` and `createConnector` functions could accept a third parameter, `sessionVariables`
+that is a map of variables to variable definitions.
+
+There are two options for executing the variable execution. The first is to do it in the client by
+using the variable substitutor in `ksqldb-parser`. The disadvantages to this are that `ksqldb-parser`
+would be a dependancy of the Java client, and that the client and server versions would have
```
vcrfxia, line 274: nit:
```suggestion
would be a dependency of the Java client, and that the client and server versions would have
```

design-proposals/klip-38-variable-substitution.md
```diff
@@ -264,7 +264,22 @@ Scope: SESSION
 In Java API client, the session is opened when `Client.create()` is called, and closed when `Client.close()` is called. Variables
 may be defined using a client method instead of a new syntax. This is to ensure users know the scope starts after the `create()` call.
 
-TBD: Define a Java API method for variable substitution.
+Users will define variables using the functions, `define(variable_name, value)` and `undefine(variable_name)`.
+The client will keep track of the calls made in a map. Alternatively, the `executeStatement`, `executeQuery`, `streamQuery`
+`insertInto`, `streamInserts` and `createConnector` functions could accept a third parameter, `sessionVariables`
+that is a map of variables to variable definitions.
+
+There are two options for executing the variable execution. The first is to do it in the client by
```
vcrfxia, line 272:
```suggestion
There are two options for executing the variable substitution. The first is to do it in the client by
```

design-proposals/klip-38-variable-substitution.md
```diff
@@ -264,7 +264,22 @@ Scope: SESSION
...
+using the variable substitutor in `ksqldb-parser`. The disadvantages to this are that `ksqldb-parser`
+would be a dependancy of the Java client, and that the client and server versions would have
+to match in order for the parser to work.
+
+The other option is to update the endpoints `/ksql`, `/inserts-stream` and `/query-streams` to accept `sessionVariables`
```
vcrfxia, line 277:
```suggestion
The other option is to update the endpoints `/ksql`, `/inserts-stream` and `/query-stream` to accept `sessionVariables`
```

vcrfxia (approved):

PR #7294, opened 2021-03-25
test: fix QTT tests after upstream change

mjsax:
Cf. https://issues.apache.org/jira/browse/KAFKA-12508

Reverts from changes from #7285 for older brnanches.

PR #7300, opened 2021-03-25
Revert "test: fix QTT tests after upstream change"

swist:
Reverts confluentinc/ksql#7294

Master is back to having 8 rows


swist:
dupe of #7299

PR #7304, opened 2021-03-26
fix: only list known connect worker properties when listing server props

vcrfxia:
### Description 

When ksqlDB runs with embedded Connect via the `ksql.connect.worker.config` config, if a user lists server properties with `list properties;`, the entire contents of the connect worker config file is displayed in the output, including any potentially sensitive configs. This is in contrast to sensitive ksqlDB or Streams configs which are properly sanitized (displayed as `[hidden]`) in the properties listing output, including any/all UDF-specific configs. This PR fixes the issue for the connect worker configs by only displaying known connect worker properties (as specified by `DistributedConfig`) as we know these are safe to display.

### Testing done 

Unit test added.

### Reviewer checklist

### DIFF
--- a/ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/execution/ListPropertiesExecutor.java
+++ b/ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/execution/ListPropertiesExecutor.java
@@ -35,6 +35,7 @@ import java.util.Optional;
 import java.util.Properties;
 import java.util.Set;
 import java.util.stream.Collectors;
+import org.apache.kafka.common.config.ConfigDef.Type;
 import org.apache.kafka.common.utils.Utils;
 import org.apache.kafka.connect.runtime.distributed.DistributedConfig;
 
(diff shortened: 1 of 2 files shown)


ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/execution/ListPropertiesExecutor.java
```diff
@@ -97,9 +99,17 @@ private ListPropertiesExecutor() {
...
-        ? Utils.propsToStringMap(getWorkerProps(configFile))
-        : Collections.emptyMap();
+    if (configFile.isEmpty()) {
+      return Collections.emptyMap();
+    }
+
+    final Map<String, String> workerProps = Utils.propsToStringMap(getWorkerProps(configFile));
+    // only list known connect worker properties to avoid showing potentially sensitive data
+    // in other configs
+    final Set<String> allowList = new DistributedConfig(workerProps).values().keySet();
```
vcrfxia, line 109: This change is very contained. The only potential risk is that `DistributedConfig` has a number of required properties and this line will fail if those properties weren't set. However, the ksqlDB server should not have been able to start if `KsqlConfig.CONNECT_WORKER_CONFIG_FILE_PROPERTY` is set and a `DistributedConfig` cannot be created from it, so this should be fine. @agavra can you confirm?

agavra (approved):
Thanks @vcrfxia!

ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/execution/ListPropertiesExecutor.java
agavra: That sounds right to me - it's unfortunate that the config def is private :/ otherwise we could've just taken a look at that directly.

agavra: Alternatively, what do you think about catching the corresponding exception and then log-and-continue? (and set the allowList to empty)

vcrfxia: Sure, I'm happy with that especially since we're backporting this change to older branches. I considered adding unit test coverage for this case but I don't think it's necessary; can add a quick one if you disagree.

PR #7306, opened 2021-03-26
test: fix QTT tests after upstream change

mjsax:
Cf. https://issues.apache.org/jira/browse/KAFKA-12508

### DIFF
--- a/ksqldb-functional-tests/src/test/resources/historical_plans/joins_-_table_to_table_wrapped_single_field_value_schema_on_inputs/5.5.0_1581572097364/spec.json
+++ b/ksqldb-functional-tests/src/test/resources/historical_plans/joins_-_table_to_table_wrapped_single_field_value_schema_on_inputs/5.5.0_1581572097364/spec.json
@@ -106,6 +106,14 @@
     }, {
       "topic" : "OUTPUT",
       "key" : "0",
+      "value" : {
+        "ID1" : null,
+        "ID2" : null
+      },
+      "timestamp" : 40
+    }, {
+      "topic" : "OUTPUT",
+      "key" : "0",
       "value" : null,
       "timestamp" : 50
     } ],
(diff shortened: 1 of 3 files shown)


ksqldb-functional-tests/src/test/resources/historical_plans/joins_-_table_to_table_wrapped_single_field_value_schema_on_inputs/5.5.0_1581572097364/spec.json
```diff
@@ -103,6 +103,14 @@
...
       },
       "timestamp" : 30
+    }, {
+      "topic" : "OUTPUT",
+      "key" : "0",
+      "value" : {
+        "ID1" : null,
+        "ID2" : null
+      },
+      "timestamp" : 40
```
mjsax, line 113: @vcrfxia @agavra -- for my own education: why is this output (and the previous one with ts=30) `<null, null>` and not just `null` as for ts=50 below.

We are testing an inner join here so I though the value should be tombstone here?

vcrfxia: The input record is not a tombstone: https://github.com/confluentinc/ksql/blob/c2ace3cb4ad05b03796b007350efc26b1846ef62/ksqldb-functional-tests/src/test/resources/query-validation-tests/joins.json#L1391
ksqlDB distinguishes between "null value" and "object containing null value".

vcrfxia (approved):
LGTM. Thanks for the fix!

ksqldb-functional-tests/src/test/resources/historical_plans/joins_-_table_to_table_wrapped_single_field_value_schema_on_inputs/5.5.0_1581572097364/spec.json
mjsax: Ah. that makes sense. Missed this "detail". Thanks for the pointer.

PR #7307, opened 2021-03-26
feat: add maven wrapper

xli1996:
### Description 
Disscusion around adopt maven wrapper to speed up building ksql. Ksql build can be slow using default maven due to https://issues.apache.org/jira/browse/MRESOLVER-164. A customized maven https://github.com/confluentinc/maven/tree/maven-3-6-3-patch with -DversionFilter parameter help reduce that issue.

### Testing done 
Replace `mvn` with `./mvnw`

### Reviewer checklist

I am also consider if versionFilter should be default true, but probably need more testing.

### DIFF
--- a/.mvn/wrapper/MavenWrapperDownloader.java
+++ b/.mvn/wrapper/MavenWrapperDownloader.java
@@ -1,0 +1,117 @@
+/*
+ * Copyright 2007-present the original author or authors.
+ *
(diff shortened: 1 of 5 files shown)


guozhangwang (approved):
Thanks @xli1996 . I think it makes sense to add this wrapper until MRESOLVER-164 is resolved. LGTM!

PR #7317, opened 2021-03-29
docs: add documentation for variable substitution in /ksql and /query-stream

jzaralim:
### Description 
docs for changes made in #7271 

### Reviewer checklist

### DIFF
--- a/docs/developer-guide/ksqldb-rest-api/ksql-endpoint.md
+++ b/docs/developer-guide/ksqldb-rest-api/ksql-endpoint.md
@@ -31,7 +31,7 @@ SELECT statements use the `/query` endpoint.
 - named by ``property-name``. Both the value and ``property-name`` should be
   strings.
 - **sessionVariables** (map): Optional. Starting from 0.18, the parameter `sessionVariables` accepts
-  a map of variable names and values as initial variable substitution values. See
+  a map of string variable names and values of any type as initial variable substitution values. See
   [ksqlDB Variable Substitution](../../../how-to-guides/substitute-variables) for more information
   on variable substitution.
 - **commandSequenceNumber** (long): Optional. If specified, the statements will
--- a/docs/developer-guide/ksqldb-rest-api/ksql-endpoint.md
+++ b/docs/developer-guide/ksqldb-rest-api/ksql-endpoint.md
@@ -30,6 +30,10 @@ SELECT statements use the `/query` endpoint.
 - **streamsProperties[``property-name``]** (string): The value of the property
 - named by ``property-name``. Both the value and ``property-name`` should be
   strings.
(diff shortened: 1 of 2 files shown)


JimGalasyn (approved):
LGTM!

docs/developer-guide/ksqldb-rest-api/streaming-endpoint.md
```diff
@@ -77,6 +77,25 @@ won't contain newlines.
 ]
 ```
 
+Starting from 0.18, variable substitution can be applied by passing a map of variables and
+definitions to the `sessionVariables` argument of the request body and referencing variables by
+enclosing them in `${}`. For example, the following requests are equivalent:
+
+```json
+{
+"sql": "select concat(A, '${fruit}') from foo",
```
vcrfxia, line 86: Can we make this example more realistic to avoid user confusion? A user seeing this might think this is a real query (that should be accepted by the ksqlDB server), which it's not.

docs/developer-guide/ksqldb-rest-api/ksql-endpoint.md
```diff
@@ -30,6 +30,10 @@ SELECT statements use the `/query` endpoint.
 - **streamsProperties[``property-name``]** (string): The value of the property
 - named by ``property-name``. Both the value and ``property-name`` should be
   strings.
+- **sessionVariables** (map): Optional. Starting from 0.18, the parameter `sessionVariables` accepts
+  a map of variable names and values as initial variable substitution values. See
+  [ksqlDB Variable Substitution](../../../how-to-guides/substitute-variables) for more information
```
vcrfxia, line 35: Does this link work without the `.md`  suffix?

docs/developer-guide/ksqldb-rest-api/ksql-endpoint.md
```diff
@@ -30,6 +30,10 @@ SELECT statements use the `/query` endpoint.
 - **streamsProperties[``property-name``]** (string): The value of the property
 - named by ``property-name``. Both the value and ``property-name`` should be
   strings.
+- **sessionVariables** (map): Optional. Starting from 0.18, the parameter `sessionVariables` accepts
+  a map of variable names and values as initial variable substitution values. See
```
vcrfxia, line 34: Does this map have to have string keys and string values or is it more flexible than that? (We should clarify if so.)

vcrfxia (approved):
Thanks @jzaralim -- a few questions inline. LGTM otherwise!

docs/developer-guide/ksqldb-rest-api/ksql-endpoint.md
jzaralim: Yup, this would point to the link on the website

PR #7320, opened 2021-03-30
refactor: reorganize query management and clean up listeners

rodesai:
This patch cleans up our code for managing the life-cycle of continuous queries. In particular:
- Break out query management from EngineContext into its own module (called QueryRegistry)
  with more explicit interfaces. EngineContext should really just be a place for pointers to
  context - it shouldn't actually be executing query management logic. This is moved into
  QueryRegistry and QueryRegistryImpl. This patch also adds detailed doc-comments about the
  QueryRegistry interface.
- Clean up query event listeners. We had various bits of code that listened on different events
  from a query. For example, we had a listener that tracked query status metrics, another that
  listened for close queries to clean up state, and another to manage the set of live queries.
  In a future PR I intend to add another listener for doing validation against the runtime (e.g.
  Streams). There were also multiple ways of listening on an event in some places. In this patch,
  I've created an explicit interface for listeners of query events that contains all the possible
  events to listen on. Listeners are provided to QueryRegistryImpl in its constructor.
- We were not explicitly sandboxing listeners, which is dangerous because we may accidentally
  take some action in response to an event on a sandbox. This patch adds an interface for
  listeners to handle the case where the engine is being sandboxed for validation.

There's also a couple of more minor changes:
- move `stop` to PersistentQueryMetadata. Transient queries can't be stopped - they can only
  be closed. We were previously jsut calling close from stop, but I think this is cleaner and
  more explicit.
- call initialize from one place for all queries (from registerQuery)
- drop uptime from QueryMetadata - we don't use this anymroe

### Testing done 
Added unit tests for query registry


ksqldb-engine/src/main/java/io/confluent/ksql/engine/QueryEventListener.java
```diff
@@ -0,0 +1,95 @@
...
+import io.confluent.ksql.services.ServiceContext;
+import io.confluent.ksql.util.QueryMetadata;
+import java.util.Optional;
+import org.apache.kafka.streams.KafkaStreams;
+
+/**
+ * Defines the events that clients of {@link io.confluent.ksql.query.QueryRegistry} can register
+ * to listen on.
+ */
+public interface QueryEventListener {
```
rodesai, line 29: I went back and forth between what's implemented here, and a structure with 1 interface (e.g. `QueryRegistrationListener`), which listens on new/deregister queries, and another interface like this one for events on a single query. In this model you'd add the listeners after the query is created (like we do before this patch). Something like:

```
interface QueryRegistrationListener {
  void onCreate(...) {}
  void onDeregister(...) {}  
  void onSandbox(QueryRegistry sanboxRegistry) {}
}

interface QueryRegistry {
    void addQueryRegistrationListener(QueryRegistrationListener l);
}

interface QueryEventListener {
    void onStateChange(...){}
    void onError(...){}
    void onClose(...){}
    void onSandbox(QueryMetadata sandboxQueryMetadata) {}
}

class QueryMetadata {
    void addQueryEventListener(QueryEventListener l);
}
```

Then you would write code like:
```
class SomethingThatCaresAboutQueries implements QueryRegistrationListener {
    void onCreate(...) {
        query.addListener(new PerQueryListener());
    }

    void onSandbox(QueryRegistry sandbox) {
        for (QueryMetadata sandboxQuery : sandbox.getAllLiveQueries()) {
            sandboxQuery.addListener(/* create a listener for the sandbox to react to sandbox events */);
        }
    }

    class PerQueryListener implements QueryEventListener {
         // per-query state and code that listens and reacts to events
    }
}
```

They're functionally the same, but keeping it all in one interface seemed a little cleaner and easier to follow, at the expense of needing to track a mapping from query id to per-query-state in the listener classes.

ksqldb-engine/src/main/java/io/confluent/ksql/util/TransientQueryMetadata.java
```diff
@@ -75,18 +74,28 @@ public TransientQueryMetadata(
...
         closeTimeout,
         new QueryId(queryApplicationId),
         QueryErrorClassifier.DEFAULT_CLASSIFIER,
         maxQueryErrorsQueueSize,
         retryBackoffInitialMs,
-        retryBackoffMaxMs
+        retryBackoffMaxMs,
+        listener
     );
-    this.initialize();
```
rodesai, line 86: I just had the registry call `initialize` for all queries - I think it's easier to reason about this code if we keep the behavior the methods common to PersistentQueryMetadata and TransientQueryMetadata (constructor, initialize, start, close) the same, and make any differences explicit in the API (e.g. by moving stop to PersistentQueryMetadata)

ksqldb-engine/src/main/java/io/confluent/ksql/internal/QueryStateListener.java
```diff
@@ -16,97 +16,132 @@
...
-  }
 
-  @Override
-  public void onError(final QueryError error) {
-    this.error = error.getType().name();
-  }
+    public void onChange(final State newState, final State oldState) {
+      state = newState.toString();
 
-  public long uptime() {
```
rodesai, line 104: we don't seem to be using this anymore

ksqldb-engine/src/main/java/io/confluent/ksql/util/TransientQueryMetadata.java
```diff
@@ -128,15 +137,9 @@ public void setLimitHandler(final LimitHandler limitHandler) {
...
-  @Override
-  public void stop() {
-    // for transient queries, anytime they are stopped
-    // we should also fully clean them up
-    close();
-  }
-
   @Override
   public void close() {
+    this.isRunning.set(false);
```
agavra, line 142: I think this used to happen at the very end of `super.close()` as part of the close callback and now it's happening at the start. I'm not sure that this is a bug, but might be worth moving it back to the end just in case (this area of the code is very brittle)

agavra (approved):
Thanks @rodesai! The change is definitely a positive one IMO, though this area of the code is quite brittle and there's lots of parallel changes going on here so it's a little hard to check for correctness. I spot-checked a few scenarios that I recall causing bugs and they seemed to be covered in the new code as well, so hopefully whatever test coverage we have here is good :)

ksqldb-engine/src/main/java/io/confluent/ksql/internal/QueryStateListener.java
```diff
@@ -16,97 +16,132 @@
...
 import java.util.Objects;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.ConcurrentMap;
 import org.apache.kafka.common.MetricName;
 import org.apache.kafka.common.metrics.Gauge;
 import org.apache.kafka.common.metrics.Metrics;
 import org.apache.kafka.streams.KafkaStreams.State;
 
-public class QueryStateListener implements ErrorStateListener {
+public class QueryStateListener implements QueryEventListener {
```
guozhangwang, line 34: nit: could we rename to `MetricsReportingListener`?

ksqldb-engine/src/main/java/io/confluent/ksql/query/QueryRegistryImpl.java
```diff
@@ -0,0 +1,366 @@
...
+  interface QueryExecutorFactory {
+    QueryExecutor create(
+        SessionConfig config,
+        ProcessingLogContext processingLogContext,
+        ServiceContext serviceContext,
+        FunctionRegistry functionRegistry
+    );
+  }
+
+  private class PerQueryListener implements QueryMetadata.Listener {
```
guozhangwang, line 348: We have two `PerQueryListener` now, one is here and another one is inside `QueryStateListener`, and the call trace is `QueryMetadata.PerQueryListener -> QueryMetadata#notifyXXX -> QueryStateListener.PerQueryListener`. 

Could we rename them differently, e.g. rename this one as just `ListenerImpl` since this is not supposed to be customized and only exposed as an interface for testing purposes.

ksqldb-engine/src/main/java/io/confluent/ksql/query/QueryExecutor.java
```diff
@@ -152,7 +143,8 @@ public TransientQueryMetadata buildTransientQuery(
       final LogicalSchema schema,
       final OptionalInt limit,
       final Optional<WindowInfo> windowInfo,
-      final boolean excludeTombstones
+      final boolean excludeTombstones,
+      final QueryMetadata.Listener listener
```
guozhangwang, line 147: Personally I found passing in this listener, which is not supposed to be changed in production, is a bit odd -- maybe for testing purposes we can expose a setter to pass in a mock after the queryMetadata is created? Anyways, if this is the pattern we've followed for testing for a while feel free to ignore it.

ksqldb-engine/src/main/java/io/confluent/ksql/query/QueryRegistry.java
```diff
@@ -0,0 +1,160 @@
...
+ * QueryRegistry also supports tracking the sink data source for each query. In particular, callers
+ * can ask for the query that was started by the C(S|T)AS statement that created a source, and can
+ * get a list of all queries writing into a source.
+ * </p>
+ * <p>
+ * Finally, the implementation should support subscribing to events about queries by implementing
+ * the {@link io.confluent.ksql.engine.QueryEventListener} interface.
+ * </p>
+ */
+public interface QueryRegistry {
```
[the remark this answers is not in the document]
guozhangwang: If we expect to have a mock for this interface for testing purposes in the future, that makes sense to me.

ksqldb-engine/src/main/java/io/confluent/ksql/util/TransientQueryMetadata.java
guozhangwang: SGTM.

guozhangwang (approved):
Thanks for the huge improvements! I just have a few minor questions. After replying / incorporating them please feel free to merge.

ksqldb-engine/src/main/java/io/confluent/ksql/engine/QueryEventListener.java
rodesai: In my view, `QueryMetadata` is meant to model a single query, so in that case we'd still have a different `QueryMetadata` for each query.  We'd need to split out some of this functionality into another class that models a Kafka Streams runtime, and tracks the queries assigned to each runtime. The state change and error notifications should go into that class, and QueryStateListener should listen on that instead of QueryMetadata as it does today. So, as a client of the engine you can register to be notified on per-query events, or per-runtime events depending on the use case. Some use cases for the former would be per-query cleanup on terminate, per-query limits checking (e.g. whats in https://github.com/confluentinc/ksql/pull/7323), and tracking errors for each query that can be surfaced in the API. A use case for the latter would be emitting metrics on the current status and error state of each runtime.

ksqldb-engine/src/main/java/io/confluent/ksql/query/QueryRegistry.java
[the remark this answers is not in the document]
rodesai: I wouldn't expect alternative implementations other than mocks/fakes. The main reason I split the interface out is so there's a clear boundary between the engine-context/engine-executor and query management.

ksqldb-engine/src/main/java/io/confluent/ksql/util/QueryMetadata.java
```diff
@@ -119,69 +114,63 @@ public long read() {
...
-    this.queryId = other.queryId;
+    this.queryId = other.getQueryId();
     this.errorClassifier = other.errorClassifier;
     this.uncaughtExceptionHandler = other.uncaughtExceptionHandler;
-    this.queryStateListener = other.queryStateListener;
     this.everStarted = other.everStarted;
-    this.queryErrors = other.queryErrors;
-    this.retryEvent = other.retryEvent;
+    this.queryErrors = new TimeBoundedQueue(Duration.ZERO, 0);
+    this.retryEvent = new RetryEvent(
```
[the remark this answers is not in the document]
rodesai: Not really. this should only get used by the sandbox, which should never actually start the query.

ksqldb-engine/src/main/java/io/confluent/ksql/util/QueryMetadata.java
```diff
@@ -475,4 +445,21 @@ private long getWaitingTimeMs() {
...
+     * throws an uncaught exception.
+     *
+     * @param error the error that occurred
+     */
+    void onError(QueryMetadata queryMetadata, QueryError error);
+
+    void onStateChange(
+        QueryMetadata queryMetadata,
+        KafkaStreams.State var1,
+        KafkaStreams.State var2);
```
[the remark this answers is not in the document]
rodesai: ah yeah this is auto-generated. I think the only case we're calling it explicitly is when restarting the query so that the callback gets poked with the new state. In that case we don't know the old value so we're just passing the current value for both.

ksqldb-engine/src/main/java/io/confluent/ksql/engine/QueryEventListener.java
```diff
@@ -0,0 +1,95 @@
...
+
+  /**
+   * Called when the query registry is creating a validation sandbox. If the event listener
+   * implementation wants to listen for events on the sandbox registry, it should return a
+   * QueryEventListener implementation when this method is called. Otherwise, return
+   * {@code Optional.empty()}
+   *
+   * @return the sandbox event listener
+   */
+  default Optional<QueryEventListener> createSandbox() {
```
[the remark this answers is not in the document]
rodesai: > you need to have a QueryEventListener object in order to create a new one, which seems a bit awkward.

This is intentional. The server uses a protocol like the following when accepting a new query:

1. start a new txn on the command topic
2. drain the command topic and execute all commands
3. create a snapshot of the ksql engine (essentially a read-only copy)
4. validate the new query by executing it using the snapshot engine
5. produce the query's physical plan to the command topic and commit

This ensures we generate a log where all commands can be executed, and so the command runner knows to never skip a command and we guarantee that we produce an identical meta-store on every restart (or in the worst case in the presence of a validation bug replay only a prefix of the command topic and alert ourselves).

This interface is meant to be invoked by step 4 so that your listeners can be notified that the engine is being snapshotted and return the appropriate listener for the snapshot. For most listeners (e.g. those that publish metrics, the cleanup topics listener), the snapshot listeners are Optional.empty(), meaning they don't run for the snapshot. https://github.com/confluentinc/ksql/pull/7323 implements a listener that does return a snapshot listener. In this case, it creates a copy of itself that enforces constraints on the number of stores.

ksqldb-engine/src/main/java/io/confluent/ksql/internal/QueryStateListener.java
```diff
@@ -16,97 +16,132 @@
...
+  public void onDeregister(final QueryMetadata query) {
+    perQuery.get(query.getQueryId()).onDeregister();
+    perQuery.remove(query.getQueryId());
+  }
 
-    if (newState != State.ERROR) {
-      this.error = NO_ERROR;
+  private static final String NO_ERROR = "NO_ERROR";
+
+  static class PerQueryListener {
```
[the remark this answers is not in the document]
rodesai: I considered this and opted not to, because they are really just implementation details of the classes they're nested in. I'll make them private.

ksqldb-engine/src/main/java/io/confluent/ksql/internal/QueryStateListener.java
rodesai: definitely - the diff was already too big :)

ksqldb-engine/src/main/java/io/confluent/ksql/query/QueryExecutor.java
rodesai: Yeah I agree it's a bit strange. I prefer not to add setters and have whatever's needed passed in when the query is created. I guess this class could really be put into QueryRegistryImpl - it's really only meant to be used from there. I opted not to to keep the registry focused. At minimum we can make this class package-private.

ksqldb-engine/src/main/java/io/confluent/ksql/internal/QueryStateListener.java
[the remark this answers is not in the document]
guozhangwang: Please ignore this comment --- it was my original question but after I read further on the PR I realized they should not be consolidated, and rather consider renaming, as mentioned in https://github.com/confluentinc/ksql/pull/7320#discussion_r607239530 --- I thought I deleted this one when creating the new one but obviously it did not go away, sorry..

PR #7321, opened 2021-03-30
fix: Make topic offset public

wcarlson5:
make a returned type public so I can use it in HC

### Reviewer checklist


stevenpyzhang (approved):
lgtm

PR #7323, opened 2021-03-31
feat: add a limit on the total number of stores

rodesai:
### Description 
Adds a config to limit the total number of state stores. To implement
this check, this patch converts KafkaStreamsQueryValidator into a
QueryEventListener, and uses the returned topology to compute an
estimate of the number of stores by multiplying the stores in the
topology by the number of partitions. It's convenient to implement
this as a QueryEventListener because we need to inspect the built
topology to determine the number of stores KS creates.

Note that only the last commit in the patch needs to be reviewed for
this PR. The previous commits are covered by #7320 

### Testing done 
Unit tests added

### Reviewer checklist


agavra:
Thanks @rodesai - I haven't had a chance to look through this yet, but I'm interested in understanding the motivation/background for the change. I'm on board, but just want to know the history.

vvcephei:

PR #7331, opened 2021-03-31
feat: scatter gather task metadata for query descriptions

stevenpyzhang:
### Description 
`show queries extended` now combines the task metadata information from remote hosts in the response.

If we ran a query where the source topic has two partitions and then got a QueryDescription of it, the description would only have 1 task metadata before this patch. Now it has both.

### Testing done 
Updated integration test

Sample response with two ksqlDB servers running

```
[
    {
        "@type": "query_descriptions",
        "statementText": "show queries extended;",
        "queryDescriptions": [
            {
                "id": "CSAS_TEST_3",
                "statementText": "CREATE STREAM TEST WITH (KAFKA_TOPIC='TEST', PARTITIONS=2, REPLICAS=1) AS SELECT *\nFROM PAGEVIEWS PAGEVIEWS\nEMIT CHANGES;",
                ...
                "ksqlHostQueryStatus": {
                    "192.168.86.31:8089": "RUNNING",
                    "192.168.86.31:8088": "RUNNING"
                },
                "queryType": "PERSISTENT",
                "queryErrors": [],
                "tasksMetadata": [
                    {
                        "taskId": "0_1",
                        "topicOffsets": [
                            {
                                "topicPartitionEntity": {
                                    "topic": "pageviews",
                                    "partition": 1
                                },
                                "endOffset": 0,
                                "committedOffset": 10949
                            }
                        ],
                        "timeCurrentIdlingStarted": null
                    },
                    {
                        "taskId": "0_0",
                        "topicOffsets": [
                            {
                                "topicPartitionEntity": {
                                    "topic": "pageviews",
                                    "partition": 0
                                },
                                "endOffset": 0,
                                "committedOffset": 11004
                            }
                        ],
                        "timeCurrentIdlingStarted": null
                    }
                ],
                "state": "RUNNING"
            }
        ],
        "warnings": []
    }
]
```

### Reviewer checklist

### DIFF
--- a/ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/execution/ListQueriesExecutor.java
+++ b/ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/execution/ListQueriesExecutor.java
@@ -204,6 +204,7 @@ public final class ListQueriesExecutor {
       final QueryId queryId = q.getId();
 
       // If the query has already been discovered, add to the ksqlQueryHostStatus mapping
+      // and the streams metadata task set
       if (allResults.containsKey(queryId)) {
         for (Map.Entry<KsqlHostInfoEntity, KsqlQueryStatus> entry :
             q.getKsqlHostQueryStatus().entrySet()) {
@@ -211,6 +212,8 @@ public final class ListQueriesExecutor {
               .get(queryId)
               .updateKsqlHostQueryStatus(entry.getKey(), entry.getValue());
         }
+        
+        allResults.get(queryId).updateTaskMetadata(q.getTasksMetadata());
       } else {
         allResults.put(queryId, q);
       }
--- a/ksqldb-rest-model/src/main/java/io/confluent/ksql/rest/entity/QueryDescription.java
+++ b/ksqldb-rest-model/src/main/java/io/confluent/ksql/rest/entity/QueryDescription.java
@@ -51,7 +51,7 @@ public class QueryDescription {
   private final Map<KsqlHostInfoEntity, KsqlQueryStatus> ksqlHostQueryStatus;
   private final KsqlQueryType queryType;
   private final List<QueryError> queryErrors;
-  private ImmutableSet<StreamsTaskMetadata> tasksMetadata;
+  private final Set<StreamsTaskMetadata> tasksMetadata;
 
   // CHECKSTYLE_RULES.OFF: ParameterNumberCheck
   @SuppressWarnings("WeakerAccess") // Invoked via reflection
@@ -86,7 +86,7 @@ public class QueryDescription {
         new HashMap<>(Objects.requireNonNull(ksqlHostQueryStatus, "ksqlHostQueryStatus"));
     this.queryType = Objects.requireNonNull(queryType, "queryType");
     this.queryErrors = new ArrayList<>(Objects.requireNonNull(queryErrors, "queryErrors"));
-    this.tasksMetadata = ImmutableSet.copyOf(Objects.requireNonNull(tasksMetadata));
+    this.tasksMetadata = new HashSet<>(Objects.requireNonNull(tasksMetadata));
   }
 
   public QueryId getId() {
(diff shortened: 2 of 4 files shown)


ksqldb-rest-model/src/main/java/io/confluent/ksql/rest/entity/QueryDescription.java
```diff
@@ -50,7 +51,7 @@
   private final Map<KsqlHostInfoEntity, KsqlQueryStatus> ksqlHostQueryStatus;
   private final KsqlQueryType queryType;
   private final List<QueryError> queryErrors;
-  private final Set<StreamsTaskMetadata> tasksMetadata;
```
wcarlson5, line 53: Can this still be final?

wcarlson5 (commented):
When you say "`show queries extended` now combines the task metadata information from remote hosts in the response" is it just this statement? Would an explain statement also get all the information as well?

stevenpyzhang:
@wcarlson5 `explain` is not set up to scatter gather right now, it only returns the local results for the query

ksqldb-rest-model/src/main/java/io/confluent/ksql/rest/entity/QueryDescription.java
stevenpyzhang: I ended up doing this way because I wanted to restrict how someone could interface with a QueryDescription object, they have to call the `updateTaskMetadata()` function to modify the set.

rodesai: would be a little more idiomatic to make the set final but not immutable and to have the getter make immutable copies.

ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/execution/ListQueriesExecutor.java
```diff
@@ -204,13 +204,16 @@ private ListQueriesExecutor() {
...
+      // and the streams metadata task set
       if (allResults.containsKey(queryId)) {
         for (Map.Entry<KsqlHostInfoEntity, KsqlQueryStatus> entry :
             q.getKsqlHostQueryStatus().entrySet()) {
           allResults
               .get(queryId)
               .updateKsqlHostQueryStatus(entry.getKey(), entry.getValue());
         }
+        
+        allResults.get(queryId).updateTaskMetadata(q.getTasksMetadata());
```
rodesai, line 216: can we add a unit test case to `ListQueriesExecutorTest`?

ksqldb-rest-model/src/main/java/io/confluent/ksql/rest/entity/QueryDescription.java
stevenpyzhang: ^ ok did that

ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/execution/ListQueriesExecutor.java
stevenpyzhang: I extended the `show queries extended` unit test

guozhangwang (approved):
Hey @stevenpyzhang just for my own understanding: it seems the `q.getTasksMetadata()` is from `kafkaStreams.localThreadsMetadata()` which only counts for local thread's metadata; is the `remoteResults` object created by calling `localThreadsMetadata` across all instances that may host some partitions of the query?

Otherwise, LGTM!

stevenpyzhang:
>it seems the q.getTasksMetadata() is from kafkaStreams.localThreadsMetadata() which only counts for local thread's metadata; is the remoteResults object created by calling localThreadsMetadata across all instances that may host some partitions of the query?

@guozhangwang q.getTasksMetadata() is referring to the QueryDescription class method in ListQueriesExecutor.java, not the QueryMetadata object that calls localThreadsMetadata() in getTaskMetadata(). We're issuing `show queries extended` to all instances across the cluster, so we'd get the QueryDescription for a Query from each instance, these QueryDescriptions contain whatever is returned from localThreadsMetadata on that instance.

We can then just take the set of StreamsTaskMetadata from a remote query description and add it to aggregated QueryDescription of a particular query.

guozhangwang:
> @guozhangwang q.getTasksMetadata() is referring to the QueryDescription class method in ListQueriesExecutor.java, not the QueryMetadata object that calls localThreadsMetadata() in getTaskMetadata()

Hmm.. but from the source code, the `QueryDescription#tasksMetadata` is initialized as `queryMetadata.getTaskMetadata()` at construction time, which triggers `localThreadsMetadata()`?

stevenpyzhang:
I think I misunderstood your original question. The simple answer is yes.

An example for what remote result would contain:

3 server ksqlDB cluster, server A1, B1, C1

Query X is running from a source with 2 partitions
Task 0_0 is running on server A1
Task 0_1 is running on server C1

Show queries extended is issued to server A1.

remoteResults would contain 2 pairs: (B1 hostname, B1 query description), (C1 hostname, C1 query description)
The query description from C1 would have the 0_1 task metadata information since C1 server has that information when calling `localThreadsMetadata`, the query description from B1 would have an empty set for task metadata.

guozhangwang:
> remoteResults would contain 2 pairs: (B1 hostname, B1 query description of Query X), (C1 hostname, C1 query description of Query X)

Got it, thank you!

PR #7332, opened 2021-03-31
test: Adds smoke tests to the packaging for debs

AlanConfluent:
### Description 
Adds a smoke tests for deb packages.  This test:
- Installs packages for zookeeper and kafka
- Installs the deb package
- starts all three services
- Runs some simple ksql queries for creating a stream, table
- Runs a push query
- Runs a pull query

### Testing done 
This is a test!

### Reviewer checklist

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -267,14 +267,6 @@ def job = {
                                 
                                 """
                             }
-                            // Run smoke tests for the packages produced above.  This utilizes
-                            // docker and must be wrapped in a withDockerServer
-                            sh """
-                                echo "Package Smoke tests"
-                                # The last argument is the public CP version to use for dependencies
-                                # such as Zookeeper and Kafka when testing ksqlDB. 
-                                ${env.WORKSPACE}/smoke/run_smoke.sh ${env.WORKSPACE} ./output/ "6.1"
-                            """
                             step([$class: 'hudson.plugins.findbugs.FindBugsPublisher', pattern: '**/*bugsXml.xml'])
 
                             if (!config.isPrJob) {
@@ -288,6 +280,19 @@ def job = {
                     }
                 }
             }
+        }
+    }
+
+    stage('KSQLDB package smoke tests') {
+        withDockerServer([uri: dockerHost()]) {
+            // Run smoke tests for the packages produced above.  This utilizes
+            // docker and must be wrapped in a withDockerServer
+            sh """
+                echo "Package Smoke tests"
+                # The last argument is the public CP version to use for dependencies
+                # such as Zookeeper and Kafka when testing ksqlDB. 
+                ${env.WORKSPACE}/smoke/run_smoke.sh ${env.WORKSPACE} ./output/ "6.1"
+            """
         }
     }
 
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -270,8 +270,8 @@ def job = {
                             // Run smoke tests for the packages produced above
                             sh """
                                 echo "Package Smoke tests"
-                                DEBS=( \\\$(find "${WORKSPACE}/output/" -name '*.deb' -print) )
-                                ${env.WORKSPACE}/smoke/run_smoke.sh "${env.WORKSPACE}" "${DEBS[0]}"    
+                                DEBS=( \$(find ./output/ -name '*.deb' -print) )
+                                ${env.WORKSPACE}/smoke/run_smoke.sh "${env.WORKSPACE}" "\${DEBS[0]}"    
                             """
                             step([$class: 'hudson.plugins.findbugs.FindBugsPublisher', pattern: '**/*bugsXml.xml'])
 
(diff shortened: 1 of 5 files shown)


build-packages.sh
```diff
@@ -161,3 +166,8 @@ if "${BUILD_JAR}"; then
         "-Dskip.docker.build=false"
 fi 
 
+if "${RUN_SMOKE_TESTS}"; then
+    DEBS=( $(find "${WORKSPACE}/output/" -name '*.deb' -print) )
+    ./smoke/run_smoke.sh "${DEBS[0]}"
+fi
+
```
andrewegel, line 173: Move this into the Jenkinsfile.

smoke/deb/Dockerfile
```diff
@@ -0,0 +1,20 @@
...
+
+RUN apt-get update && apt-get install -y wget gnupg software-properties-common openjdk-8-jdk curl jq
+
+ENV JAVA_HOME=/usr/lib/jvm/java-8-openjdk-amd64
+
+RUN wget -qO - https://packages.confluent.io/deb/6.1/archive.key | apt-key add -
+
+RUN add-apt-repository "deb [arch=amd64] https://packages.confluent.io/deb/6.1 stable main"
+
+RUN apt-get update && apt-get install -y confluent-kafka
```
andrewegel, line 16: Are you perpetually testing against CP 6.1? This should probably be an argument into the derived container.

andrewegel:
General comment is to run all these new shell scripts through shellcheck: https://www.shellcheck.net/ and address all raised issues.

smoke/run_smoke.sh
```diff
@@ -0,0 +1,16 @@
...
+
+run_package_test() {
+    local smoke_test_dir=$1
+    local package_file=$2
+    local tmp_dir=$(mktemp -d -t ksqldb-smoke)
+    echo "Created smoke directory for test $tmp_dir"
+    cp smoke/common/* "$tmp_dir"
+    cp "$smoke_test_dir"/* "$tmp_dir"
+    cp "$package_file" "$tmp_dir"
+    docker build $tmp_dir
```
andrewegel, line 13: You build thats fine, but you also will need to clean up docker containers after you're done - The job of cleaning up containers should probably be done in Jenkinsfile in a post = {} section to ensure everything is cleaned.

vvcephei (commented):
Thanks for this, @AlanConfluent !

The approach looks good. Modulo @andrewegel 's feedback, I'm +1.

Did/can you test this test by running the Jenkins job with an intentionally broken `deb` package? I.e., maybe you can do something like edit the Makefile to remove `ksql-server-start` and verify that the Jenkins file really fails and notifies us properly?

We've had multiple occasions in Streams benchmarks where the Jenkinsfile appears to work, but it turns out not to actually fail the build when it should.

build-packages.sh
AlanConfluent: Done

AlanConfluent:
> General comment is to run all these new shell scripts through shellcheck: https://www.shellcheck.net/ and address all raised issues.

@andrewegel 

Did it.  Everything lints nicely.

AlanConfluent:
> Thanks for this, @AlanConfluent !
> 
> The approach looks good. Modulo @andrewegel 's feedback, I'm +1.
> 
> Did/can you test this test by running the Jenkins job with an intentionally broken `deb` package? I.e., maybe you can do something like edit the Makefile to remove `ksql-server-start` and verify that the Jenkins file really fails and notifies us properly?

@vvcephei 

I stuck in a fake failure case to the test and it definitely failed the build: https://jenkins.confluent.io/job/confluentinc-pr/job/ksql/job/PR-7332/23/

AlanConfluent:
Please take another look @andrewegel and @vvcephei

Jenkinsfile
```diff
@@ -264,8 +264,17 @@ def job = {
...
                                 """
                             }
+                            // Run smoke tests for the packages produced above.  This utilizes
+                            // docker and must be wrapped in a withDockerServer
+                            sh """
+                                echo "Package Smoke tests"
+                                # The last argument is the public CP version to use for dependencies
+                                # such as Zookeeper and Kafka when testing ksqlDB. 
+                                ${env.WORKSPACE}/smoke/run_smoke.sh ${env.WORKSPACE} ./output/ "6.1"
+                            """
```
andrewegel, line 277: NIT: Not a blocker for my approval, but logically you could instead do this in a different stage called `Test KSQL-DB` (currently in `Build KSQL-DB`), but that would add a lot more lines of code to re-use the deeply nested `with*` directives above. Again not a blocker for me, more of a visual thing when looking at the stages of the build.

andrewegel (approved):

Jenkinsfile
AlanConfluent: Ok, made this change.  It actually wasn't bad at all since I didn't need most of those with blocks, just a couple.

PR #7335, opened 2021-04-01
feat: enable variable substitution for java client

jzaralim:
### Description 
Adds the functions `define(variable, value)`, `undefine(variable)` and `showVariables()`, and passes session variables to requests.

### Testing done 
integration + unit tests

### Reviewer checklist

### DIFF
--- a/docs/developer-guide/ksqldb-clients/java-client.md
+++ b/docs/developer-guide/ksqldb-clients/java-client.md
@@ -777,7 +777,6 @@ undefines a session variable, and [`getVariables()`](/api/io/confluent/ksql/api/
 and their values. Substitution is supported for the following functions:
 * `streamQuery`
 * `executeQuery`
-* `terminatePushQuery`
 * `executeStatement`
 * `describeSource`
 * `createConnector`
--- a/docs/developer-guide/ksqldb-clients/java-client.md
+++ b/docs/developer-guide/ksqldb-clients/java-client.md
@@ -33,6 +33,7 @@ Use the Java client to:
 - [Describe specific streams and tables](#describe-source)
 - [Get metadata about the ksqlDB cluster](#server-info)
 - [Manage, list and describe connectors](#connector-operations)
+- [Use variable substitution](#variable-substitution)
 
 Get started below or skip to the end for full-fledged [examples](#tutorial-examples).
 
@@ -765,6 +766,35 @@ System.out.println(description.name()
   + " It reads/writes to " + description.sources().size() + " ksqlDB sources"
   + " and uses " + description.topics().size() + " topics."
 );
+```
+
+Use variable substitution<a name="variable-substitution"></a>
+---------------------------------------------------------------
+
+Starting with ksqlDB 0.18.0, users can define session variables by calling the `define()` method and
+reference them in other fuctions by wrapping the variable name in `${}`. The `undefine()` method
+undefines a session variable, and `getVariables()` returns a map of the currently defined variables
+and their values.
+
+### Example Usage ###
+Define a new variable:
+```java
+client.define("topic", "stream-topic");
+```
+
+Use a variable in `executeStatement`:
+```java
+client.executeStatement("CREATE STREAM S (NAME STRING, AGE INTEGER) WITH (kafka_topic='${topic}', value_format='json');");
+```
+
+Undefine a variable:
+```java
+client.undefine("topic");
+```
+
+Get all variables:
+```java
+Map<String, Object> variables = client.getVariables();
 ```
 
 Tutorial Examples<a name="tutorial-examples"></a>
--- a/docs/developer-guide/ksqldb-clients/java-client.md
+++ b/docs/developer-guide/ksqldb-clients/java-client.md
@@ -33,7 +33,7 @@ Use the Java client to:
 - [Describe specific streams and tables](#describe-source)
 - [Get metadata about the ksqlDB cluster](#server-info)
 - [Manage, list and describe connectors](#connector-operations)
-- [Use variable substitution](#variable-substitution)
+- [Define variables for substitution](#variable-substitution)
 
 Get started below or skip to the end for full-fledged [examples](#tutorial-examples).
 
@@ -768,13 +768,21 @@ System.out.println(description.name()
 );
 ```
 
-Use variable substitution<a name="variable-substitution"></a>
+Define variables for substitution<a name="variable-substitution"></a>
 ---------------------------------------------------------------
 
-Starting with ksqlDB 0.18.0, users can define session variables by calling the `define()` method and
-reference them in other fuctions by wrapping the variable name in `${}`. The `undefine()` method
-undefines a session variable, and `getVariables()` returns a map of the currently defined variables
-and their values.
+Starting with ksqlDB 0.18.0, users can define session variables by calling the [`define()`](/api/io/confluent/ksql/api/client/Client.html#define(java.lang.String,boolean,java.lang.Object)) method and
+reference them in other functions by wrapping the variable name in `${}`. The [`undefine()`](/api/io/confluent/ksql/api/client/Client.html#undefine(java.lang.String)) method
+undefines a session variable, and [`getVariables()`](/api/io/confluent/ksql/api/client/Client.html#getVariables()) returns a map of the currently defined variables
+and their values. Substitution is supported for the following functions:
+* `streamQuery`
+* `executeQuery`
+* `terminatePushQuery`
+* `executeStatement`
+* `describeSource`
+* `createConnector`
+* `dropConnector`
+* `describeConnector`
 
 ### Example Usage ###
 Define a new variable:
--- a/ksqldb-api-client/src/main/java/io/confluent/ksql/api/client/Client.java
+++ b/ksqldb-api-client/src/main/java/io/confluent/ksql/api/client/Client.java
@@ -284,7 +284,7 @@ public interface Client {
    * Define a session variable which can be referenced in sql commands by wrapping the variable name
    * with {@code ${}}.
    *
-   * If the variable is already defined, the value will be overridden.
+   * If the variable is already defined, the existing value will be overridden.
    *
    * @param variable the name of the variable
    * @param value the value represented by the variable
@@ -295,7 +295,7 @@ public interface Client {
    * Undefine a session variable. {@code '${<variable name>}'} will no longer be replaced in other
    * functions.
    *
-   * If the variable is not defined, nothing will happen.
+   * If the variable is not defined, then this method call is a no-op.
    *
    * @param variable the name of the variable to undefine
    */
--- a/ksqldb-api-client/src/main/java/io/confluent/ksql/api/client/Client.java
+++ b/ksqldb-api-client/src/main/java/io/confluent/ksql/api/client/Client.java
@@ -281,7 +281,10 @@ public interface Client {
   CompletableFuture<ConnectorDescription> describeConnector(String connectorName);
 
   /**
-   * Define a session variable which can be referenced in sql commands by wrapping it with ${}
+   * Define a session variable which can be referenced in sql commands by wrapping the variable name
+   * with {@code ${}}.
+   *
+   * If the variable is already defined, the value will be overridden.
    *
    * @param variable the name of the variable
    * @param value the value represented by the variable
@@ -289,12 +292,17 @@ public interface Client {
   void define(String variable, Object value);
 
   /**
-   * Undefine a variable.
+   * Undefine a session variable. {@code '${<variable name>}'} will no longer be replaced in other
+   * functions.
+   *
+   * If the variable is not defined, nothing will happen.
+   *
+   * @param variable the name of the variable to undefine
    */
   void undefine(String variable);
 
   /**
-   * @return a map of variables and values
+   * @return a map of the session variables and values used for variable substitution.
    */
   Map<String, Object> getVariables();
 
--- a/ksqldb-api-client/src/main/java/io/confluent/ksql/api/client/Client.java
+++ b/ksqldb-api-client/src/main/java/io/confluent/ksql/api/client/Client.java
@@ -281,6 +281,24 @@ public interface Client {
   CompletableFuture<ConnectorDescription> describeConnector(String connectorName);
 
   /**
+   * Define a session variable which can be referenced in sql commands by wrapping it with ${}
+   *
+   * @param variable the name of the variable
+   * @param value the value represented by the variable
+   */
+  void define(String variable, Object value);
+
+  /**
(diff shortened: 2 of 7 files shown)


JimGalasyn (approved):
LGTM!

ksqldb-api-client/src/main/java/io/confluent/ksql/api/client/Client.java
```diff
@@ -280,6 +280,24 @@
    */
   CompletableFuture<ConnectorDescription> describeConnector(String connectorName);
 
+  /**
+   * Define a session variable which can be referenced in sql commands by wrapping it with ${}
```
vcrfxia, line 284:
```suggestion
   * Define a session variable which can be referenced in sql commands by wrapping the variable name with {@code ${}}
```
We should check that the braces render correctly in the javadoc when generated by generating the HTML docs and seeing what it looks like (or opting for Google search instead).

ksqldb-api-client/src/main/java/io/confluent/ksql/api/client/Client.java
```diff
@@ -280,6 +280,24 @@
...
+   */
+  void define(String variable, Object value);
+
+  /**
+   * Undefine a variable.
+   */
+  void undefine(String variable);
+
+  /**
+   * @return a map of variables and values
```
vcrfxia, line 297: Clarify that these variables are local session variables for purposes of variable substitution.

ksqldb-api-client/src/main/java/io/confluent/ksql/api/client/Client.java
```diff
@@ -280,6 +280,24 @@
...
+  /**
+   * Define a session variable which can be referenced in sql commands by wrapping it with ${}
+   *
+   * @param variable the name of the variable
+   * @param value the value represented by the variable
+   */
+  void define(String variable, Object value);
+
+  /**
+   * Undefine a variable.
```
vcrfxia, line 292: See above. (Meta-note: The purpose of these javadocs is so that a user unfamiliar with the client can understand what the methods do and how to use them at a glance. Better to be more verbose than less :) )

ksqldb-api-client/src/main/java/io/confluent/ksql/api/client/Client.java
```diff
@@ -280,6 +280,24 @@
    */
   CompletableFuture<ConnectorDescription> describeConnector(String connectorName);
 
+  /**
+   * Define a session variable which can be referenced in sql commands by wrapping it with ${}
+   *
```
vcrfxia, line 285: Clarify the behavior when the variable already exists (in this javadoc). The current behavior is that the value will be overridden. Is that the canonical choice, rather than throwing an error? Would be good to have someone from the product team sanity check these new interfaces and their expected behaviors.

Similarly, undefine currently does not throw an error if the variable name does not exist. (Clarify in javadoc and sanity check expected behavior.)

ksqldb-api-client/src/test/java/io/confluent/ksql/api/client/ClientTest.java
```diff
@@ -1501,6 +1501,45 @@ public void shouldDropConnector() {
...
+            CommandStatus.Status.SUCCESS,
+            "Success",
+            Optional.of(new QueryId("CSAS_0"))
+        ),
+        0L
+    );
+    testEndpoints.setKsqlEndpointResponse(Collections.singletonList(entity));
+
+    // When:
+    javaClient.executeStatement("CSAS;").get();
```
vcrfxia, line 1537: Let's add equivalent tests for `streamQuery()` and `executeQuery()` as well.

ksqldb-api-client/src/main/java/io/confluent/ksql/api/client/Client.java
```diff
@@ -280,6 +280,24 @@
...
+   * @param variable the name of the variable
+   * @param value the value represented by the variable
+   */
+  void define(String variable, Object value);
```
vcrfxia, line 289: Can we add an integration test for a non-string value? It's not clear to me from the docs whether variable substitution is even supported for non-string values. This interface definition suggests that is, and we should have corresponding test coverage.

docs/developer-guide/ksqldb-clients/java-client.md
```diff
@@ -33,6 +33,7 @@ Use the Java client to:
 - [Describe specific streams and tables](#describe-source)
 - [Get metadata about the ksqlDB cluster](#server-info)
 - [Manage, list and describe connectors](#connector-operations)
+- [Use variable substitution](#variable-substitution)
```
vcrfxia, line 36: nit: how about
```suggestion
- [Define variables for substitution](#variable-substitution)
```
?

docs/developer-guide/ksqldb-clients/java-client.md
```diff
@@ -767,6 +768,35 @@ System.out.println(description.name()
...
 ```
 
+Use variable substitution<a name="variable-substitution"></a>
+---------------------------------------------------------------
+
+Starting with ksqlDB 0.18.0, users can define session variables by calling the `define()` method and
+reference them in other fuctions by wrapping the variable name in `${}`. The `undefine()` method
+undefines a session variable, and `getVariables()` returns a map of the currently defined variables
+and their values.
+
```
vcrfxia, line 778: These docs need to specify which client methods support variable substitution. Otherwise, users might be confused that variable substitution isn't working for, e.g., the `describeSource()` method. Let's sanity-check the desired behavior here with product as well. I can see a strong case for supporting substitution for all methods, including `describeSource()`.

vcrfxia (commented):
Thanks @jzaralim -- exciting stuff! LGTM overall with comments inline.

docs/developer-guide/ksqldb-clients/java-client.md
```diff
@@ -768,13 +768,21 @@ System.out.println(description.name()
...
-reference them in other fuctions by wrapping the variable name in `${}`. The `undefine()` method
-undefines a session variable, and `getVariables()` returns a map of the currently defined variables
-and their values.
+Starting with ksqlDB 0.18.0, users can define session variables by calling the [`define()`](/api/io/confluent/ksql/api/client/Client.html#define(java.lang.String,boolean,java.lang.Object)) method and
+reference them in other functions by wrapping the variable name in `${}`. The [`undefine()`](/api/io/confluent/ksql/api/client/Client.html#undefine(java.lang.String)) method
+undefines a session variable, and [`getVariables()`](/api/io/confluent/ksql/api/client/Client.html#getVariables()) returns a map of the currently defined variables
+and their values. Substitution is supported for the following functions:
+* `streamQuery`
+* `executeQuery`
+* `terminatePushQuery`
```
vcrfxia, line 780: Does the close query endpoint actually support variable substitution? I didn't think so -- confused how this could work if not.

ksqldb-api-client/src/main/java/io/confluent/ksql/api/client/Client.java
```diff
@@ -281,20 +281,28 @@
   CompletableFuture<ConnectorDescription> describeConnector(String connectorName);
 
   /**
-   * Define a session variable which can be referenced in sql commands by wrapping it with ${}
+   * Define a session variable which can be referenced in sql commands by wrapping the variable name
+   * with {@code ${}}.
```
vcrfxia, line 285: A quick google search suggests that this won't work -- the inner curly brace causes the code format to terminate prematurely, and needs to be escaped.

ksqldb-api-client/src/main/java/io/confluent/ksql/api/client/Client.java
```diff
@@ -281,20 +281,28 @@
...
+   *
+   * If the variable is already defined, the value will be overridden.
```
vcrfxia, line 287:
```suggestion
   * If the variable is already defined, the existing value will be overridden.
```

ksqldb-api-client/src/main/java/io/confluent/ksql/api/client/Client.java
```diff
@@ -281,20 +281,28 @@
...
    *
    * @param variable the name of the variable
    * @param value the value represented by the variable
    */
   void define(String variable, Object value);
 
   /**
-   * Undefine a variable.
+   * Undefine a session variable. {@code '${<variable name>}'} will no longer be replaced in other
```
vcrfxia, line 295: Same issue with javadoc formatting and curly braces as above.

ksqldb-api-client/src/main/java/io/confluent/ksql/api/client/Client.java
```diff
@@ -281,20 +281,28 @@
...
+   * functions.
+   *
+   * If the variable is not defined, nothing will happen.
```
vcrfxia, line 298:
```suggestion
   * If the variable is not defined, then this method call is a no-op.
```

ksqldb-api-client/src/main/java/io/confluent/ksql/api/client/impl/ClientImpl.java
```diff
@@ -196,7 +196,7 @@ public BatchedQueryResult executeQuery(
 
     makePostRequest(
         CLOSE_QUERY_ENDPOINT,
-        new JsonObject().put("queryId", queryId),
+        new JsonObject().put("queryId", queryId).put("sessionVariables", sessionVariables),
```
vcrfxia, line 199: Does this work? Does the close query endpoint actually support variable substitution?

ksqldb-api-client/src/main/java/io/confluent/ksql/api/client/impl/ClientImpl.java
```diff
@@ -358,7 +360,9 @@ public BatchedQueryResult executeQuery(
 
     makePostRequest(
         KSQL_ENDPOINT,
-        new JsonObject().put("ksql", "drop connector " + name + ";"),
+        new JsonObject()
+            .put("ksql", "drop connector " + name + ";")
+            .put("sessionVariables", sessionVariables),
```
vcrfxia, line 365: We also need this for create connector, right? I don't see the change in this commit.

ksqldb-api-client/src/test/java/io/confluent/ksql/api/client/ClientTest.java
```diff
@@ -1540,6 +1540,30 @@ public void shouldSendSessionVariablesToEndpoint() throws Exception {
...
+    // Given
+    javaClient.define("a", "a");
+
+    // When
+    javaClient.streamQuery("query;").get();
+
+    // Then
+    assertThat(testEndpoints.getLastSessionVariables(), is(new JsonObject().put("a", "a")));
+  }
+
```
vcrfxia, line 1566: Let's also add unit test coverage for the additional client methods that we're supporting variable substitution for (create connector, describe source, etc). Unless I'm reading the PR wrong, create connector isn't actually covered right now. Hopefully a new unit test would catch that.

ksqldb-api-client/src/main/java/io/confluent/ksql/api/client/Client.java
jzaralim: I think this might have been fixed recently - it showed up properly when I generated the javadocs locally:
Screen Shot 2021-04-05 at 2 10 11 AM

ksqldb-api-client/src/main/java/io/confluent/ksql/api/client/impl/ClientImpl.java
jzaralim: That was in the previous commit.

docs/developer-guide/ksqldb-clients/java-client.md
jzaralim: My bad, forgot it was a different endpoint. Removed.

ksqldb-api-client/src/main/java/io/confluent/ksql/api/client/impl/ConnectorCommandResponseHandler.java
```diff
@@ -33,7 +33,8 @@ static void handleCreateConnectorResponse(
       final JsonObject connectorInfoEntity,
       final CompletableFuture<Void> cf
   ) {
-    if (connectorInfoEntity.getString("@type").equals("connector_info")) {
+    if (connectorInfoEntity.containsKey("statementText")
+        && connectorInfoEntity.containsKey("info")) {
```
jzaralim, line 37: Changed the logic for this because it was failing in the unit tests.

vcrfxia: To check my understanding: you're saying that this works in integration tests because of how the response entity is serialized and deserialized, but doesn't work in unit tests since those don't go through the serialization/deserialization path? 

I'm fine with switching away from checking for `@type`. However, let's reuse `AdminResponseHandlers#isCreateConnectorResponse()` and similar for the others, rather than duplicating the logic in two places.

ksqldb-api-client/src/main/java/io/confluent/ksql/api/client/Client.java
vcrfxia: Oh okay, good to know. Thanks for checking!

vcrfxia (approved):
Thanks @jzaralim ! One last comment inline but LGTM otherwise :)

docs/developer-guide/ksqldb-clients/java-client.md
```diff
@@ -767,6 +768,42 @@ System.out.println(description.name()
...
+reference them in other functions by wrapping the variable name in `${}`. The [`undefine()`](/api/io/confluent/ksql/api/client/Client.html#undefine(java.lang.String)) method
+undefines a session variable, and [`getVariables()`](/api/io/confluent/ksql/api/client/Client.html#getVariables()) returns a map of the currently defined variables
+and their values. Substitution is supported for the following functions:
+* `streamQuery`
+* `executeQuery`
+* `executeStatement`
+* `describeSource`
+* `createConnector`
+* `dropConnector`
+* `describeConnector`
```
vcrfxia, line 784: Just to confirm: we're planning to add support for `insertInto()` and `streamInserts()` in a future PR, right?

jzaralim: Yup, that will be added after substitution is supported for the REST endpoint.

ksqldb-api-client/src/main/java/io/confluent/ksql/api/client/impl/ConnectorCommandResponseHandler.java
jzaralim: Exactly

Issue #4397, opened 2020-01-28
Support window bounds columns on windowed group by


big-andy-coates:
QTT test:

```json
{
      "name": "non-windowed stream source using window bounds columns",
      "statements": [
        "CREATE STREAM TEST (ROWKEY INT KEY, IGNORED INT) WITH (kafka_topic='test_topic', value_format='JSON');",
        "CREATE TABLE S2 as SELECT count(1) as count, min(windowstart) as wstart, min(windowend) as wend FROM test WINDOW TUMBLING (SIZE 1 SECOND) group by ROWKEY;"
      ],
      "inputs": [
        {"topic": "test_topic", "key": 0, "value": {}, "timestamp": 10345},
        {"topic": "test_topic", "key": 0, "value": {}, "timestamp": 13251}
      ],
      "outputs": [
        {"topic": "S2", "key": 0,"value": {"COUNT":1}, "timestamp": 10345, "window": {"start": 10000, "end": 11000, "type": "time"}},
        {"topic": "S2", "key": 0,"value": {"COUNT":1}, "timestamp": 13251, "window": {"start": 13000, "end": 14000, "type": "time"}}
      ],
      "post": {
        "sources": [
          {
            "name": "S2",
            "type": "stream",
            "keyFormat": {"format": "KAFKA", "windowType": "HOPPING", "windowSize": 30000},
            "schema": "ROWKEY INT KEY, ID BIGINT, VALUE BIGINT, KEY INT"
          }
        ]
      }
    }
```

big-andy-coates:
While it should be possible to make `WINDOWSTART` and `WINDOWEND` available within:

* **non-aggregate select expressions**
* **having expressions**

The following places can not so easily be supported:

* **aggregate select expressions**, e.g. `min(windowStart)`: Streams passes the underlying key, not the `Windowed` key to the `Aggregator` on windowed aggregations. (The `aggregate` calls on `TimeWindowedKStream` and `SessionWindowedKStream` take an aggregator with `K` not `Windowed<K>`).  **Fixing this requires a Streams change**
** where expressions.  Where expressions are applied _before_ the aggregation, so the window bounds are not known.  I see two possible fixes for this:
   * support a `filter` method on `SessionWindoedKStream` and `TimeWindowedKeyStream` that allows access to the window bounds.  **This requires a Streams change**.
[...]

PR #7336, opened 2021-04-01
fix: append additional fields to row values in `KsqlMaterialization.java` to fix pull query filtering

### LINKED ISSUE #7312: Pull queries return 0 rows when there is a key filter node in the persistent query topology
**Describe the bug**
Pull queries return 0 rows when there is a key filter node in the persistent query topology

**To Reproduce**
1. The version of KSQL: `0.17`
2. Any SQL statements you ran:
```
ksql> CREATE TABLE INPUT (ID STRING PRIMARY KEY, GRADE STRING, RANK INT) WITH (kafka_topic='test_topic', value_format='JSON', partitions=1);
ksql> CREATE TABLE NEW_INPUT AS SELECT ID, CONCAT(GRADE, '+'), RANK * 2 AS DOUBLE_RANK FROM INPUT where ID='11';

ksql> insert into INPUT(ID, GRADE, RANK) VALUES ('11', 'A', 1);
ksql> insert into INPUT(ID, GRADE, RANK) VALUES ('10', 'B', 2);
ksql> insert into INPUT(ID, GRADE, RANK) VALUES ('12', 'C', 3);

ksql> SELECT * FROM NEW_INPUT;
```

**Expected behavior**
```
+--------------------------------------------------+--------------------------------------------------+--------------------------------------------------+
|ID                                                |KSQL_COL_0                                        |DOUBLE_RANK                                       |
+--------------------------------------------------+--------------------------------------------------+--------------------------------------------------+
|11                                                |A+                                                |2                                                 |
```
**Actual behaviour**
No rows returned:
```
+--------------------------------------------------+--------------------------------------------------+--------------------------------------------------+
|ID                                                |KSQL_COL_0                                        |DOUBLE_RANK                                       |
+--------------------------------------------------+--------------------------------------------------+--------------------------------------------------+

```

**Additional context**

cprasad1:
### Description 
Fixes #7312 and other similar situations by appending additional fields to row values in `KsqlMaterialization.java`

### Testing done 
- RQTT
- Unit Testing

### Reviewer checklist

### DIFF
--- a/ksqldb-functional-tests/src/test/resources/rest-query-validation-tests/pull-queries-against-ctas-materialized.json
+++ b/ksqldb-functional-tests/src/test/resources/rest-query-validation-tests/pull-queries-against-ctas-materialized.json
@@ -205,6 +205,101 @@
         "message": "The `USERS` table isn't queryable. To derive a queryable table, you can do 'CREATE TABLE QUERYABLE_USERS AS SELECT * FROM USERS'. See https://cnfl.io/queries for more info.",
         "status": 400
       }
+    },
+    {
+      "name": "persistent query with KEY filter and projection +++ pull query table scan and single key lookup",
+      "statements": [
+        "CREATE TABLE INPUT (ID STRING PRIMARY KEY, GRADE STRING, RANK INT) WITH (kafka_topic='test_topic', value_format='JSON');",
+        "CREATE TABLE NEW_INPUT AS SELECT ID, GRADE, RANK * 2 AS DOUBLE_RANK FROM INPUT WHERE ID='11';",
+        "SELECT * FROM NEW_INPUT;",
+        "SELECT ID, DOUBLE_RANK FROM NEW_INPUT;",
+        "SELECT * FROM NEW_INPUT WHERE ID='11';",
+        "SELECT ID, DOUBLE_RANK FROM NEW_INPUT WHERE ID='11';",
+        "SELECT * FROM NEW_INPUT WHERE ID='12';",
+        "SELECT ID, DOUBLE_RANK FROM NEW_INPUT WHERE ID='12';"
+      ],
+      "inputs": [
+        {"topic": "test_topic", "timestamp": 12346, "key": "11", "value": {"GRADE": "A", "RANK": 1}},
+        {"topic": "test_topic", "timestamp": 12345, "key": "10", "value": {"GRADE": "B", "RANK": 2}},
+        {"topic": "test_topic", "timestamp": 12389, "key": "12", "value": {"GRADE": "C", "RANK": 3}}
+      ],
+      "responses": [
+        {"admin": {"@type": "currentStatus"}},
+        {"admin": {"@type": "currentStatus"}},
+        {"query": [
+          {"header":{"schema":"`ID` STRING KEY, `GRADE` STRING, `DOUBLE_RANK` INTEGER"}},
+          {"row":{"columns":["11", "A", 2]}}
+        ]},
+        {"query": [
+          {"header":{"schema":"`ID` STRING KEY, `DOUBLE_RANK` INTEGER"}},
+          {"row":{"columns":["11", 2]}}
+        ]},
+        {"query": [
+          {"header":{"schema":"`ID` STRING KEY, `GRADE` STRING, `DOUBLE_RANK` INTEGER"}},
+          {"row":{"columns":["11", "A", 2]}}
+        ]},
+        {"query": [
+          {"header":{"schema":"`ID` STRING KEY, `DOUBLE_RANK` INTEGER"}},
+          {"row":{"columns":["11", 2]}}
+        ]},
+        {"query": [
+          {"header":{"schema":"`ID` STRING KEY, `GRADE` STRING, `DOUBLE_RANK` INTEGER"}}
+        ]},
+        {"query": [
+          {"header":{"schema":"`ID` STRING KEY, `DOUBLE_RANK` INTEGER"}}
+        ]}
+      ]
+    },
+    {
+      "name": "persistent query with NON KEY filter and projection +++ pull query table scan and single key lookup",
+      "statements": [
+        "CREATE TABLE INPUT (ID STRING PRIMARY KEY, GRADE STRING, RANK INT) WITH (kafka_topic='test_topic', value_format='JSON');",
+        "CREATE TABLE NEW_INPUT AS SELECT ID, GRADE, RANK * 2 AS DOUBLE_RANK FROM INPUT WHERE GRADE='A';",
+        "SELECT * FROM NEW_INPUT;",
+        "SELECT ID, DOUBLE_RANK FROM NEW_INPUT;",
+        "SELECT * FROM NEW_INPUT WHERE ID='11';",
+        "SELECT ID, DOUBLE_RANK FROM NEW_INPUT WHERE ID='11';",
+        "SELECT * FROM NEW_INPUT WHERE ID='12';",
+        "SELECT ID, DOUBLE_RANK FROM NEW_INPUT WHERE ID='12';"
+      ],
+      "inputs": [
+        {"topic": "test_topic", "timestamp": 12346, "key": "11", "value": {"GRADE": "A", "RANK": 1}},
+        {"topic": "test_topic", "timestamp": 12345, "key": "10", "value": {"GRADE": "B", "RANK": 2}},
+        {"topic": "test_topic", "timestamp": 12389, "key": "12", "value": {"GRADE": "C", "RANK": 3}},
+        {"topic": "test_topic", "timestamp": 12397, "key": "14", "value": {"GRADE": "A", "RANK": 4}},
+        {"topic": "test_topic", "timestamp": 12478, "key": "15", "value": {"GRADE": "B", "RANK": 5}},
+        {"topic": "test_topic", "timestamp": 12378, "key": "16", "value": {"GRADE": "A", "RANK": 6}}
+      ],
+      "responses": [
+        {"admin": {"@type": "currentStatus"}},
+        {"admin": {"@type": "currentStatus"}},
+        {"query": [
+          {"header":{"schema":"`ID` STRING KEY, `GRADE` STRING, `DOUBLE_RANK` INTEGER"}},
+          {"row":{"columns":["11", "A", 2]}},
+          {"row":{"columns":["16", "A", 12]}},
+          {"row":{"columns":["14", "A", 8]}}
+        ]},
+        {"query": [
+          {"header":{"schema":"`ID` STRING KEY, `DOUBLE_RANK` INTEGER"}},
+          {"row":{"columns":["11", 2]}},
+          {"row":{"columns":["16", 12]}},
+          {"row":{"columns":["14", 8]}}
+        ]},
+        {"query": [
+          {"header":{"schema":"`ID` STRING KEY, `GRADE` STRING, `DOUBLE_RANK` INTEGER"}},
+          {"row":{"columns":["11", "A", 2]}}
+        ]},
+        {"query": [
+          {"header":{"schema":"`ID` STRING KEY, `DOUBLE_RANK` INTEGER"}},
+          {"row":{"columns":["11", 2]}}
+        ]},
+        {"query": [
+          {"header":{"schema":"`ID` STRING KEY, `GRADE` STRING, `DOUBLE_RANK` INTEGER"}}
+        ]},
+        {"query": [
+          {"header":{"schema":"`ID` STRING KEY, `DOUBLE_RANK` INTEGER"}}
+        ]}
+      ]
     }
   ]
 }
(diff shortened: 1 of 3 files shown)


ksqldb-streams/src/test/java/io/confluent/ksql/execution/streams/materialization/KsqlMaterializationTest.java
```diff
@@ -137,10 +94,69 @@
...
+            A_ROWTIME
+    );
+
+    WINDOWED_ROW2 = WindowedRow.of(
+            SCHEMA,
+            new Windowed<>(A_KEY2, STREAM_WINDOW),
+            A_VALUE2,
+            A_ROWTIME
+    );
+
```
cprasad1, line 155: These were pulled into the setup layer because `KsqlMaterialization.getIntermediateRow` mutates some of the variables

ksqldb-streams/src/main/java/io/confluent/ksql/execution/streams/materialization/KsqlMaterialization.java
```diff
@@ -185,16 +185,38 @@ public MaterializedWindowedTable windowed() {
...
+            return filterAndTransform(row.windowedKey(), getIntermediateRow(row), row.rowTime())
                 .map(v -> row.withValue(v, schema()));
           })
           .filter(Optional::isPresent)
           .map(Optional::get)
           .iterator();
     }
   }
+
+  public static GenericRow getIntermediateRow(final TableRow row) {
```
cprasad1, line 199: This method can be broken down into 2 methods: One for Regular rows, and the other for Windowed rows. Thoughts?

ksqldb-streams/src/main/java/io/confluent/ksql/execution/streams/materialization/KsqlMaterialization.java
```diff
@@ -174,7 +174,7 @@ public MaterializedWindowedTable windowed() {
       final Builder<WindowedRow> builder = ImmutableList.builder();
 
       for (final WindowedRow row : result) {
-        filterAndTransform(row.windowedKey(), row.value(), row.rowTime())
+        filterAndTransform(row.windowedKey(), getIntermediateRow(row), row.rowTime())
```
cprasad1, line 177: Unsure if we even need to append the extra columns for windowed rows. Windowed state stores might need more testing

ksqldb-functional-tests/src/test/resources/rest-query-validation-tests/pull-queries-against-ctas-materialized.json
```diff
@@ -205,6 +205,451 @@
...
+          {"header":{"schema":"`GRADE` STRING KEY, `NUM_STUDENTS` BIGINT"}},
+          {"row":{"columns":["F",4]}}
+
+        ]},
+        {"query": [
+          {"header":{"schema":"`GRADE` STRING KEY"}},
+          {"row":{"columns":["F"]}}
+
+        ]}
+      ]
```
cprasad1, line 652: Currently, there are no additional tests for Windowed materialization. Unsure if we need them right now

ksqldb-streams/src/main/java/io/confluent/ksql/execution/streams/materialization/KsqlMaterialization.java
guozhangwang: Could you elaborate a bit more here? Is that because the windowed key may have the needed columns for filtering?

guozhangwang (commented):
Just wondering what's the difference for windowed stores, for my own education. Otherwise the fix lgtm.

ksqldb-streams/src/main/java/io/confluent/ksql/execution/streams/materialization/KsqlMaterialization.java
cprasad1: @guozhangwang my understanding is that windowed tables always need an aggregation, so they have a state store backing the aggregated table. If that is the case, then we don't need to append this extra metadata as `KsqlMaterialization` is sophisticated enough to handle those cases (we have test cases for that). I noticed that we follow a similar pattern in `ProjectOperator` and `SelectOperator` of generating intermediate rows on which filters and transformations can be applied. I added these fields as a hedge against potential cases that we might miss (obviously it comes at a cost). That being said, I have a couple of questions for you:
1) Are there any type of Windowed tables that are not queryable today that we want to be able to query?
2) Can Windowed tables be derived without doing any aggregations? (Specifically, `GROUP BY` aggregations)

ksqldb-streams/src/main/java/io/confluent/ksql/execution/streams/materialization/KsqlMaterialization.java
AlanConfluent: I don't think it's a big deal since it can be figured out by the row itself.  They also share other fields you'd have to otherwise duplicate.

AlanConfluent, line 199: It would be nice to just call https://github.com/confluentinc/ksql/blob/master/ksqldb-engine/src/main/java/io/confluent/ksql/physical/pull/operators/PullPhysicalOperatorUtil.java since this is effectively part of the physical layer for pull queries.

ksqldb-streams/src/test/java/io/confluent/ksql/execution/streams/materialization/KsqlMaterializationTest.java
```diff
@@ -53,74 +38,46 @@
...
-      A_ROWTIME
-  );
-
-  private static final WindowedRow WINDOWED_ROW2 = WindowedRow.of(
-      SCHEMA,
-      new Windowed<>(A_KEY2, STREAM_WINDOW),
-      A_VALUE2,
-      A_ROWTIME
-  );
+  private LogicalSchema SCHEMA;
```
AlanConfluent, line 64: If they're not static, they should have lowercase names.

ksqldb-streams/src/test/java/io/confluent/ksql/execution/streams/materialization/KsqlMaterializationTest.java
```diff
@@ -137,10 +94,69 @@
...
+            new Windowed<>(A_KEY2, STREAM_WINDOW),
+            A_VALUE2,
+            A_ROWTIME
+    );
+
     materialization = new KsqlMaterialization(
         inner,
         SCHEMA,
-        ImmutableList.of(project, filter)
+        ImmutableList.of(filter, project)
```
AlanConfluent, line 159: Did you swap these because this is a more realistic ordering?

ksqldb-functional-tests/src/test/resources/rest-query-validation-tests/pull-queries-against-ctas-materialized.json
AlanConfluent: You could always tested a windowed case like 
```
"CREATE TABLE AGGREGATE AS SELECT ID, COUNT(1) AS COUNT FROM INPUT WINDOW TUMBLING(SIZE 1 SECOND) GROUP BY ID HAVING COUNT(1) > 2;",
```
We don't have a lot of tests with the having clause for pull queries and I think it might trigger the `KsqlMaterialization` transform logic to apply the filtering.  Since they can apply for group bys I assume they work for windowed tables as well.  Normal where clauses presumably don't touch `KsqlMaterialization` when a group by is in place since the materialization happens after the filter has been applied, so that might not need additional testing.

But in general, you're right that a lot of windowed logic has already been tested fairly well in https://github.com/confluentinc/ksql/blob/master/ksqldb-functional-tests/src/test/resources/rest-query-validation-tests/pull-queries-against-materialized-aggregates.json since windows require a group by.

ksqldb-streams/src/main/java/io/confluent/ksql/execution/streams/materialization/KsqlMaterialization.java
AlanConfluent: I haven't tested it by try a windowed query with group by and having clause.  I think that might trigger `KsqlMaterialization`.  In that case, presumably you could say `HAVING WINDOWSTART > 20` and therefore need to have those columns.

guozhangwang: +1 to @AlanConfluent , I think today if you have a `WINDOW BY + GROUP BY + HAVING` you would have a first windowed table from `windowBy+groupBy` aggregations, and then a second windowed table from `having` as a filtering condition. So in that sense not all windowed tables should be generated with aggregations, they can also be generated from table statless operators from other existing windowed tables.

ksqldb-streams/src/test/java/io/confluent/ksql/execution/streams/materialization/KsqlMaterializationTest.java
cprasad1: yes

ksqldb-functional-tests/src/test/resources/rest-query-validation-tests/pull-queries-against-ctas-materialized.json
AlanConfluent: These are good cases.  It would be good to do a windowed table + group by + having clause mentioning windowstart.  That's one last case I don't see.

ksqldb-streams/src/main/java/io/confluent/ksql/execution/streams/materialization/KsqlMaterialization.java
AlanConfluent: It would be good to have this as a RQTT test where the having clause mentions windowstart or windowend

ksqldb-functional-tests/src/test/resources/rest-query-validation-tests/pull-queries-against-ctas-materialized.json
cprasad1: we don't allow `WINDOWSTART` that at the moment, so it's not a testable case. The specific error message is `Window bounds column WINDOWSTART can only be used in the SELECT clause of windowed aggregations and can not be passed to aggregate functions.`

ksqldb-streams/src/main/java/io/confluent/ksql/execution/streams/materialization/KsqlMaterialization.java
AlanConfluent, line 199: If we don't always want to unconditionally add these fields in the future, we should add a comment saying that we need to address this in the future to decide when we need to add these additional fields.

AlanConfluent (approved):
Nice PR and good test cases

ksqldb-functional-tests/src/test/resources/rest-query-validation-tests/pull-queries-against-ctas-materialized.json
cprasad1: Related: https://github.com/confluentinc/ksql/issues/4397

guozhangwang (approved):
Thanks for the added testing coverage!

ksqldb-functional-tests/src/test/resources/rest-query-validation-tests/pull-queries-against-ctas-materialized.json
```diff
@@ -205,6 +205,613 @@
...
+        {"query": [
+          {"header":{"schema":"`ID` STRING KEY, `GRADE` STRING, `DOUBLE_RANK` INTEGER"}}
+        ]},
+        {"query": [
+          {"header":{"schema":"`ID` STRING KEY, `DOUBLE_RANK` INTEGER"}}
+        ]}
+      ]
+    },
+    {
+      "name": "persistent query with KEY filter and projection +++ pull query table scan and single key lookup ***FAILURE***",
```
guozhangwang, line 254: nit: should we explicitly state the the error root cause here? Since otherwise the name is exactly the same as above except we say it is a `***FAILURE***` case.

ksqldb-functional-tests/src/test/resources/rest-query-validation-tests/pull-queries-against-ctas-materialized.json
guozhangwang: I'm still wondering, if @AlanConfluent 's query is possible:

```
CREATE TABLE AGGREGATE AS SELECT ID, COUNT(1) AS COUNT FROM INPUT WINDOW TUMBLING(SIZE 1 SECOND) GROUP BY ID HAVING COUNT(1) > 2;
```

note it does not try to group by window-start/end, while we should still be able to pull query it with conditions on other columns?

ksqldb-functional-tests/src/test/resources/rest-query-validation-tests/pull-queries-against-ctas-materialized.json
```diff
@@ -205,6 +205,613 @@
...
+          {"row":{"columns":["F"]}}
+
+        ]}
+      ]
+    },
+    {
+      "name": "windowed - select star with HAVING filter",
+      "statements": [
+        "CREATE STREAM INPUT (ID STRING KEY, IGNORED INT) WITH (kafka_topic='test_topic', value_format='JSON');",
+        "CREATE TABLE AGGREGATE AS SELECT ID, COUNT(1) AS COUNT FROM INPUT WINDOW TUMBLING(SIZE 1 SECOND) GROUP BY ID HAVING COUNT(1) > 1;",
```
cprasad1, line 658: @guozhangwang is this the test similar to what you are interested in?

guozhangwang: Ah yes, thanks!

Issue #7242, opened 2021-03-16
Changing KEY_FORMAT doesn't work in CTAS with GROUP BY STRUCT


mikebin:
**Describe the bug**
Using `WITH (KEY_FORMAT='<format>')` seems to be ignored in an aggregation CTAS that performs a `GROUP BY` on a `STRUCT`, when the source stream has `KEY_FORMAT='KAFKA'` (the default).

**To Reproduce**
Version: 0.15.0

```sql
create stream s (f1 string, f2 string, f3 string) with (kafka_topic='s', value_format='json', partitions=1);

create table t with (key_format='json') as select struct(f1:=f1, f2:=f2) as k, as_value(struct(f1:=f1, f2:=f2)) as key, count(*) as total from s group by struct(f1:=f1, f2:=f2);
```

**Expected behavior**
Table successfully created with JSON key struct.

**Actual behaviour**
```
Key format does not support schema.
format: KAFKA
schema: Persistence{columns=[`KSQL_COL_0` STRUCT<`F1` STRING, `F2` STRING> KEY], features=[]}
reason: The 'KAFKA' format does not support type 'STRUCT'
Caused by: The 'KAFKA' format does not support type 'STRUCT'
```

**Additional context**
```sql
--Works fine for CTAS with multiple individual fields instead of a STRUCT
create table t1 with (key_format='json') as select f1, f2, as_value(f1) as f1v, as_value(f2) as f2v, count(*) as cnt from s group by f1, f2;

---Also works with CSAS using a STRUCT key
create stream s1 with (key_format='json') as select struct(f1:=f1, f2:=f2) as k, as_value(struct(f1:=f1, f2:=f2)) as key from s partition by struct(f1:=f1, f2:=f2); 
```

vcrfxia:
Slight clarification on this bug: the bug is only present when converting from a key format that does not support structs (e.g., the `KAFKA` format) to a key format that does support structs (e.g., `JSON`). Until this bug is fixed, a workaround is to copy the source stream into another stream with a key format that supports structs, and then run the aggregation query to create the table.

Issue #7066, opened 2021-02-22
CodeGen could be optimized when it is applied more frequently


guozhangwang:
**Is your feature request related to a problem? Please describe.**
While reviewing https://github.com/confluentinc/ksql/pull/6966 I found there are two inefficiencies with the codegen traversals today:

1) The first pass (CodeGen#Visitor) and the second pass (SqlToJavaVisitor#Formatter) had duplicated type resolution paths, e.g. for functional calls we try to resolve the argument types in both passes. We could, instead, do such resolution once in the first pass, cache the information on the node and then reuse it in the second.

2) Both passes leverage on a shared utility of `ExpressionTypeManager`, which again had some duplicated resolution logic with `CodeGenRunner` and `SqlToJavaVisitor`. Using it as a magical blackbox on the callers helps development convenience but may incur some unnecessary overhead.

We do not have time breakdown right now, but from pull queries we do see code-gen phase taking a long portion of time, and I suspect things like this contributes to its latency.

At the moment this seems not an urgent need since 1) we are moving away from code-gen for pull queries, and 2) we expect submission rate for persistent / transient queries to be small. So code-gen is not on our critical path. But moving forward as we potentially combine interpretation and code gen for pull queries, and allow more programmatical ways to create scalable transient push queries, we may need to pay attention to the time elapsed on code gen.

guozhangwang:
@vvcephei is this ticket triaged?

PR #7337, opened 2021-04-01
fix: source key format compatibility with non primitive types

abbccdda:
### Description 
This patch tries to fix the wrong behavior we spotted in https://github.com/confluentinc/ksql/issues/7242, where the conversion to struct key format should work when the source stream key format is KafkaFormat.

In addition, we expanded the fix and addressed the potential conversion failure for DelimitedFormat case as well.

### Testing done 
Added unit tests and validation tests for KafkaFormat, DelimitedFormat and NoneFormat conversions.

### Reviewer checklist


ksqldb-engine/src/main/java/io/confluent/ksql/serde/SerdeFeaturesFactory.java
```diff
@@ -258,6 +282,7 @@ private static void validateExplicitValueWrapping(
 
   private static boolean formatSupportsMultipleColumns(final KeyFormat format) {
     return !format.getFormat().equals(KafkaFormat.NAME)
-        && !format.getFormat().equals(NoneFormat.NAME);
+        && !format.getFormat().equals(NoneFormat.NAME)
+        && !format.getFormat().equals(DelimitedFormat.NAME);
```
abbccdda, line 286: We also need to check delimited type as well

ksqldb-engine/src/main/java/io/confluent/ksql/structured/SchemaKStream.java
```diff
@@ -392,6 +399,10 @@ public SchemaKGroupedStream groupBy(
     );
   }
 
+  protected List<SqlType> toSqlTypes(List<Expression> expressions) {
+    return expressions.stream().map(typeManager::getExpressionSqlType).collect(Collectors.toList());
```
guozhangwang, line 403: Hmm.. I'm wondering if we do need to re-parse the expression again to get the sql types (I filed https://github.com/confluentinc/ksql/issues/7066 before feeling that we are doing multiple duplicated passes today). If we have to, and this is not on the critical paths, maybe it's fine. cc @agavra @AlanConfluent could you share your thoughts?

ksqldb-engine/src/test/java/io/confluent/ksql/serde/SerdeFeaturesFactoryTest.java
```diff
@@ -390,10 +411,55 @@ public void shouldNotConvertFormatForMulticolKeysWhenSanitizingIfDisallowed() {
...
-    final KeyFormat sanitized = SerdeFeaturesFactory.sanitizeKeyFormat(format, 2, false);
+    final KeyFormat sanitized = SerdeFeaturesFactory.sanitizeKeyFormat(format, MULTI_SQL_TYPES, false);
 
     // Then:
     assertThat(sanitized.getFormatInfo(), equalTo(FormatInfo.of(KafkaFormat.NAME)));
     assertThat(sanitized.getFeatures(), equalTo(SerdeFeatures.of()));
   }
+
+  @Test
+  public void shouldConvertKafkaFormatForSingleKeyWithNonPrimitiveType() {
```
guozhangwang, line 422: What about `shouldConvertKafkaFormatForSingleKeyWithPrimitiveType`, is that covered elsewhere? Ditto beow.

ksqldb-functional-tests/src/main/java/io/confluent/ksql/test/tools/TestExecutor.java
```diff
@@ -362,7 +362,7 @@ private void verifyOutput(final TestCase testCase) {
     try {
       key = topicInfo.getKeySerializer().serialize(rec.topic(), rec.key());
     } catch (final Exception e) {
-      throw new AssertionError("Failed to serialize value: " + e.getMessage()
+      throw new AssertionError("Failed to serialize key: " + e.getMessage()
```
guozhangwang, line 365: Nice catch.

ksqldb-engine/src/main/java/io/confluent/ksql/serde/SerdeFeaturesFactory.java
vcrfxia: This change is not correct. The delimited format _does_ support multiple columns, just not complex types (struct, array, map).

ksqldb-engine/src/main/java/io/confluent/ksql/serde/SerdeFeaturesFactory.java
```diff
@@ -150,6 +161,20 @@ private static KeyFormat sanitizeKeyFormatForMultipleColumns(
...
+   * Check whether given sql type set needs a compound key support. It returns true
+   * when one of the conditions is met:
+   * 1. given sql types have more than one entry
+   * 2. it is a singleton non-primitive sql type
+   *
+   * @param sqlTypes the sql types to be tested
+   * @return true if given sql types do not need a compound key type support.
+   */
+  private static boolean needCompoundKeySupport(List<SqlType> sqlTypes) {
+    return sqlTypes.size() > 1 || (sqlTypes.size() == 1 &&
```
vcrfxia, line 174: I don't think it makes sense to combine these two conditions. As per my comment above, there are two different properties that are separate: whether a format supports multiple columns (e.g., two string columns) and whether a format supports complex-types (e.g., struct or array). The two properties are different as the delimited format supports the former but not the latter. We need two separate checks in the methods in this file: (1) the original check for multiple columns (which is correct and should not need to be changed), and (2) a new check for whether specific column types are supported (what's currently missing, and why the struct type is causing a bug).

ksqldb-engine/src/main/java/io/confluent/ksql/structured/SchemaKStream.java
vcrfxia: Not on the CC list but my two cents anyway 😄 : I think this is fine since, as you pointed out, this is not on the critical path. This code path is only hit once while validating, starting, or re-starting the query.

ksqldb-engine/src/main/java/io/confluent/ksql/serde/SerdeFeaturesFactory.java
```diff
@@ -150,6 +161,20 @@ private static KeyFormat sanitizeKeyFormatForMultipleColumns(
...
+        !(sqlTypes.get(0) instanceof SqlPrimitiveType));
```
vcrfxia, line 175: It's also not enough to check for primitive vs non-primitive types. This will work for the delimited format since the delimited format supports exactly primitive types and no non-primitive types, but this will not work for the Kafka format, which not only doesn't support non-primitive types, but also does not support some primitive types such as DECIMAL. We will need more custom logic to encode which types are supported by which formats. We should consider adding this into the Format interface, rather than encoding all the logic into this specific class.

vcrfxia (commented):
Thanks for tackling this, @abbccdda -- definitely a tricky first bug! Comments inline. We'll want additional test coverage as well for the different cases (e.g., converting away from Kafka key format for queries that repartition by a decimal column).

ksqldb-engine/src/main/java/io/confluent/ksql/structured/SchemaKStream.java
agavra: +1 to @vcrfxia's comment - it might be worth it in the future but not at the expense of a bigger PR here 😉

ksqldb-functional-tests/src/test/resources/query-validation-tests/group-by.json
```diff
@@ -572,6 +572,83 @@
         ]
       }
     },
+    {
+      "name": "multiple expressions - KAFKA to struct key format",
```
agavra, line 576: unless we already have some, can we add negative tests of changing from `JSON` to `KAFKA` in a way that doesn't work? and similarly, trying to have delimited with multi-columns where they are no all primitives

agavra (commented):
Thanks for the pr @abbccdda! Some comments inline (don't feel the need to wait for my second pass if @vcrfxia approves)

ksqldb-engine/src/main/java/io/confluent/ksql/serde/SerdeFeaturesFactory.java
```diff
@@ -150,6 +166,37 @@ private static KeyFormat sanitizeKeyFormatForMultipleColumns(
...
+        return isSupportedByDelimitedFormat(sqlTypes);
+      case NoneFormat.NAME:
+        return false;
+      default:
+        return true;
+    }
+  }
+
+  private static boolean isSupportedByKafkaFormat(final List<SqlType> sqlTypes) {
+    return sqlTypes.size() <= 1
```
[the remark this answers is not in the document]
abbccdda: I'm not sure about that. The previous logic is definitely covering it but seems not reliable:
```
 if (!allowKeyFormatChangeToSupportMultipleKeys
        || numKeyColumns <= 1
        || formatSupportsMultipleColumns(keyFormat))
```
If `numKeyColumns` is 1, it seems we shouldn't use NONE format.

ksqldb-engine/src/main/java/io/confluent/ksql/serde/SerdeFeaturesFactory.java
```diff
@@ -150,6 +166,37 @@ private static KeyFormat sanitizeKeyFormatForMultipleColumns(
...
+        && sqlTypes.get(0) instanceof SqlPrimitiveType
```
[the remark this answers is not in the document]
abbccdda: Just extra safety check since sql type and base type are loosely coupled.

ksqldb-engine/src/main/java/io/confluent/ksql/serde/SerdeFeaturesFactory.java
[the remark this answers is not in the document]
guozhangwang: I think the previous conditions a bit convoluted with None / Kafka, and for KafkFormat specifically we could exclude `sqlTypes.size() == 0`.

ksqldb-engine/src/test/java/io/confluent/ksql/serde/SerdeFeaturesFactoryTest.java
```diff
@@ -390,10 +398,101 @@ public void shouldNotConvertFormatForMulticolKeysWhenSanitizingIfDisallowed() {
...
+    // When:
+    final KeyFormat sanitized = SerdeFeaturesFactory.sanitizeKeyFormat(format, ImmutableList.of(SqlPrimitiveType.of(SqlBaseType.BOOLEAN)), true);
+
+    // Then:
+    assertThat(sanitized.getFormatInfo(), equalTo(FormatInfo.of(JsonFormat.NAME)));
+    assertThat(sanitized.getFeatures(), equalTo(SerdeFeatures.of(SerdeFeature.UNWRAP_SINGLES)));
+  }
+
+  @Test
+  public void shouldConvertDelimitedFormatForSingleKeyWithNonPrimitiveType() {
```
guozhangwang, line 439: Just curious: why should we convert for single column delimited format? I thought for either non-primitive or primitive types, single column delimited format does not need to be converted to json.

abbccdda: delimited format does not support compound type for either single or multi-key case IMHO.

ksqldb-engine/src/main/java/io/confluent/ksql/serde/SerdeFeaturesFactory.java
```diff
@@ -112,34 +129,33 @@ public static SerdeFeatures buildValueFeatures(
...
       final KeyFormat keyFormat,
-      final int numKeyColumns,
+      final List<SqlType> newKeyColumnSqlTypes,
       final boolean allowKeyFormatChangeToSupportMultipleKeys
   ) {
     return sanitizeKeyFormatWrapping(
         sanitizeKeyFormatForMultipleColumns(
             keyFormat,
-            numKeyColumns,
+            newKeyColumnSqlTypes,
```
vcrfxia, line 146: nit: can we add a third type of sanitization, rather than combining the logic for multiple key columns and specific key column types into a single method? The two are logically different and can be handled independently from each other. I think this will make the logic in this file easier to read.

ksqldb-engine/src/main/java/io/confluent/ksql/serde/SerdeFeaturesFactory.java
```diff
@@ -150,6 +166,37 @@ private static KeyFormat sanitizeKeyFormatForMultipleColumns(
...
   }
 
+  /**
+   * Check whether given sql type set could be supported by given key format.
+   *
+   * @param keyFormat the key format to be checked
+   * @param sqlTypes the sql types to be tested
+   * @return true if the out sql types are supported by key format
+   */
+  private static boolean keyFormatConversionSupported(final KeyFormat keyFormat,
```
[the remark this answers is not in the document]
vcrfxia: +1 to moving the logic for which types are supported into the Formats themselves, rather than having them here.

ksqldb-engine/src/test/java/io/confluent/ksql/serde/SerdeFeaturesFactoryTest.java
```diff
@@ -390,10 +398,101 @@ public void shouldNotConvertFormatForMulticolKeysWhenSanitizingIfDisallowed() {
...
-    final KeyFormat sanitized = SerdeFeaturesFactory.sanitizeKeyFormat(format, 2, false);
+    final KeyFormat sanitized = SerdeFeaturesFactory.sanitizeKeyFormat(format, MULTI_SQL_TYPES, false);
 
     // Then:
     assertThat(sanitized.getFormatInfo(), equalTo(FormatInfo.of(KafkaFormat.NAME)));
     assertThat(sanitized.getFeatures(), equalTo(SerdeFeatures.of()));
   }
+
+  @Test
+  public void shouldConvertKafkaFormatForSingleKeyWithPrimitiveType() {
```
vcrfxia, line 409:
```suggestion
  public void shouldConvertKafkaFormatForSingleKeyWithNonPrimitiveType() {
```

ksqldb-engine/src/test/java/io/confluent/ksql/serde/SerdeFeaturesFactoryTest.java
vcrfxia: +1 to @abbccdda 's rationale -- the delimited format does not support STRUCT, MAP, or ARRAY.

ksqldb-functional-tests/src/test/resources/query-validation-tests/group-by.json
```diff
@@ -572,6 +572,83 @@
...
+        {"topic": "OUTPUT", "key": {"F1": 2, "F2": 4}, "value": {"KEY": {"F1":2,"F2":4}, "TOTAL": 1}},
+        {"topic": "OUTPUT", "key": {"F1": 1, "F2": null}, "value": {"KEY": {"F1":1,"F2":null}, "TOTAL": 1}},
+        {"topic": "OUTPUT", "key": {"F1": null, "F2": 1}, "value": {"KEY": {"F1":null,"F2":1}, "TOTAL": 1}},
+        {"topic": "OUTPUT", "key": {"F1": 1, "F2": 2}, "value": {"KEY": {"F1":1,"F2":2}, "TOTAL": 2}},
+        {"topic": "OUTPUT", "key": {"F1": 2, "F2": 4}, "value": {"KEY": {"F1":2,"F2":4}, "TOTAL": 2}},
+        {"topic": "OUTPUT", "key": {"F1": 2, "F2": 1}, "value": {"KEY": {"F1":2,"F2":1}, "TOTAL": 1}}
+      ]
+    },
+    {
+      "name": "multiple expressions - delimited key format to multi-column primitive format",
```
vcrfxia, line 602: What is the purpose of adding this test? In the context of this PR, we should add a test to verify that no key format change is required in this case -- this test should pass without the `(KEY_FORMAT='JSON')` in the CTAS statement.

ksqldb-functional-tests/src/test/resources/query-validation-tests/group-by.json
vcrfxia: +1 to the additional tests @agavra suggested. Let's also add one for converting from KAFKA key format to an unsupported primitive type such as DECIMAL (similar to the relevant unit test you've already added above).

ksqldb-engine/src/test/java/io/confluent/ksql/serde/SerdeFeaturesFactoryTest.java
guozhangwang: Yeah I know; my question was primarily that, for a single column delimited format (which is a primitive type), why do we still need to convert to a json format, than just keeping it as a primitive typed format?

vcrfxia: Are you asking about this specific unit test or something else? In this specific unit test, there is a single, non-primitive type key (struct type), which is why the key format must be converted from DELIMITED to JSON.

guozhangwang: Sorry I should clarify myself a bit, my original confusions should be separated to two: 1) why delimited single-column non-primitive key should be converted to Json, and that's now cleared since DELIMITED does not support non-primitive types, 2) why delimited single-column primitive key should be converted to Json, and it seems we never convert in this case actually and I was just confused by not seeing a `shouldNotConvertDelimitedFormatForSingleKeyWithPrimitiveType` case.

abbccdda: @guozhangwang for 2) I will add this as a new test case

ksqldb-functional-tests/src/test/resources/query-validation-tests/group-by.json
abbccdda: The original bug statement does contain `WITH (KEY_FORMAT='JSON')`. Without it, the original test would also fail, see `multiple expressions - KAFKA to non-KAFKA key format` and remove the `WITH (KEY_FORMAT='JSON')` to reproduce.

I think we are inherently not converting the source format for key construction, but let me check.

vcrfxia: > The original bug statement does contain WITH (KEY_FORMAT='JSON'). Without it, the existing test would also fail

I agree that the test in the original bug statement fails without `WITH (KEY_FORMAT='JSON')` in the CTAS statement, but for this particular test (delimited key format to multi-column primitive format), the test should not fail if `WITH (KEY_FORMAT='JSON')` is removed, since the sink schema is supported by the delimited format. It would be good to verify that this test still passes without the `WITH (KEY_FORMAT='JSON')`. I also don't think we need to test the version with `WITH (KEY_FORMAT='JSON')` (for this test) since that would no longer be testing the bug in this PR but simply testing the functionality of converting key formats in general, which should already be tested elsewhere.

ksqldb-engine/src/main/java/io/confluent/ksql/serde/SerdeFeaturesFactory.java
```diff
@@ -112,37 +116,59 @@ public static SerdeFeatures buildValueFeatures(
...
    * @param allowKeyFormatChangeToSupportMultipleKeys safeguard to prevent changing key formats in
    *                                                  unexpected ways. if false, no format change
    *                                                  will take place
    * @return the new key format to use
    */
   public static KeyFormat sanitizeKeyFormat(
       final KeyFormat keyFormat,
-      final int numKeyColumns,
+      final List<SqlType> newKeyColumnSqlTypes,
       final boolean allowKeyFormatChangeToSupportMultipleKeys
```
vcrfxia, line 128: Let's rename this to `allowKeyFormatChangeToSupportNewKeySchema` now that this flag is used to guard key format changes for key column types as well.

ksqldb-engine/src/main/java/io/confluent/ksql/serde/SerdeFeaturesFactory.java
```diff
@@ -112,37 +116,59 @@ public static SerdeFeatures buildValueFeatures(
...
+        newKeyColumnSqlTypes.size() == 1
     );
   }
 
+  /**
+   * Check whether given key format supports multiple columns.
+   *
+   * @param keyFormat the key format to be checked
+   * @param numKeyColumns number of output key columns
+   * @return true if multiple column key types are supported by key format
```
vcrfxia, line 147: This javadoc looks incorrect?

ksqldb-engine/src/main/java/io/confluent/ksql/serde/SerdeFeaturesFactory.java
```diff
@@ -112,37 +116,59 @@ public static SerdeFeatures buildValueFeatures(
...
     }
+    return convertToJsonFormat(keyFormat);
+  }
+
+  /**
+   * Check whether given sql type set could be supported by given key format.
+   *
+   * @param keyFormat the key format to be checked
+   * @param sqlTypes the sql types to be tested
+   * @return true if the out sql types are supported by key format
```
vcrfxia, line 164: Same as above: this javadoc looks incorrect.

ksqldb-functional-tests/src/test/resources/query-validation-tests/group-by.json
```diff
@@ -572,6 +572,143 @@
...
+        {"topic": "OUTPUT", "key": {"F1": 1, "F2": 2}, "value": {"KEY": {"F1":1,"F2":2}, "TOTAL": 2}},
+        {"topic": "OUTPUT", "key": {"F1": 2, "F2": 4}, "value": {"KEY": {"F1":2,"F2":4}, "TOTAL": 2}},
+        {"topic": "OUTPUT", "key": {"F1": 2, "F2": 1}, "value": {"KEY": {"F1":2,"F2":1}, "TOTAL": 1}}
+      ]
+    },
+    {
+      "name": "multiple expressions - KAFKA to non-supported decimal key format",
+      "statements": [
+        "CREATE STREAM TEST (f1 INT KEY, f2 INT) WITH (kafka_topic='test_topic', value_format='JSON', partitions=1);",
+        "CREATE TABLE OUTPUT WITH (KEY_FORMAT='JSON') AS SELECT ROUND(f1) AS key, COUNT(*) AS total FROM TEST GROUP BY ROUND(f1);"
```
vcrfxia, line 605: The output schema of `ROUND(int)` is a long, not a decimal, so this test does not test what it claims to test.

You can add a post-test assertion to assert the expected schema: https://github.com/confluentinc/ksql/blob/3bf8531d0f67de2590ed8f945033b58f68e38a0f/ksqldb-functional-tests/src/test/resources/query-validation-tests/group-by.json#L28

ksqldb-serde/src/main/java/io/confluent/ksql/serde/KeyFormat.java
```diff
@@ -97,6 +106,30 @@ public String getFormat() {
     return format.getProperties();
   }
 
+  @JsonIgnore
+  public boolean supportKeyTypes(final List<SqlType> sqlTypes) {
```
vcrfxia, line 110: Rather than having this logic here, I think it'd be better to have it in the `Format` interface itself. We can add a new method `boolean supportsType(final SqlType sqlType)` and have each `Format` implementation define which types are supported by the particular format, and avoid this switch statement entirely.

vcrfxia (commented):
Thanks @abbccdda -- this is really close! Nice test coverage. Comments inline.

ksqldb-functional-tests/src/test/resources/query-validation-tests/group-by.json
abbccdda: I see, how do we create a decimal type then?

ksqldb-serde/src/main/java/io/confluent/ksql/serde/KeyFormat.java
abbccdda: I think the tricky scenario we hit here is that we are using `FormatInfo`, which contains a format string, not the actual `Format` type. I'm afraid we couldn't get rid of the switch condition, at a minimum we need a switch for converting string to `Format`

ksqldb-functional-tests/src/test/resources/query-validation-tests/group-by.json
vcrfxia: You can create a decimal-type column in the source stream and GROUP BY that column: https://docs.ksqldb.io/en/latest/reference/sql/data-types/#arbitrary-precision-types

ksqldb-serde/src/main/java/io/confluent/ksql/serde/KeyFormat.java
vcrfxia: `FormatFactory#of()` addresses this exact use case: https://github.com/confluentinc/ksql/blob/3bf8531d0f67de2590ed8f945033b58f68e38a0f/ksqldb-serde/src/main/java/io/confluent/ksql/serde/FormatFactory.java#L48

ksqldb-serde/src/main/java/io/confluent/ksql/serde/KeyFormat.java
```diff
@@ -24,8 +24,17 @@
 import com.google.errorprone.annotations.Immutable;
 import edu.umd.cs.findbugs.annotations.SuppressFBWarnings;
 import io.confluent.ksql.model.WindowType;
+import io.confluent.ksql.schema.ksql.SchemaConverters;
```
vcrfxia, line 27: These imports look unused? Checkstyle will fail. Looks like there are other unused imports in some other files too -- might want to run checkstyle locally in order to clean these up in order to get a clean Jenkins build sooner :)

ksqldb-serde/src/main/java/io/confluent/ksql/serde/kafka/KafkaFormat.java
```diff
@@ -59,4 +62,11 @@ public String name() {
 
     return KafkaSerdeFactory.createSerde(schema);
   }
+
+  @Override
+  public boolean supportKeyType(SqlType type) {
+    return type instanceof SqlPrimitiveType
```
vcrfxia, line 68: Is this first piece necessary?

ksqldb-functional-tests/src/test/resources/query-validation-tests/group-by.json
```diff
@@ -572,6 +572,153 @@
...
+        {"topic": "OUTPUT", "key": {"F1": 2, "F2": 4}, "value": {"KEY": {"F1":2,"F2":4}, "TOTAL": 1}},
+        {"topic": "OUTPUT", "key": {"F1": 1, "F2": null}, "value": {"KEY": {"F1":1,"F2":null}, "TOTAL": 1}},
+        {"topic": "OUTPUT", "key": {"F1": null, "F2": 1}, "value": {"KEY": {"F1":null,"F2":1}, "TOTAL": 1}},
+        {"topic": "OUTPUT", "key": {"F1": 1, "F2": 2}, "value": {"KEY": {"F1":1,"F2":2}, "TOTAL": 2}},
+        {"topic": "OUTPUT", "key": {"F1": 2, "F2": 4}, "value": {"KEY": {"F1":2,"F2":4}, "TOTAL": 2}},
+        {"topic": "OUTPUT", "key": {"F1": 2, "F2": 1}, "value": {"KEY": {"F1":2,"F2":1}, "TOTAL": 1}}
+      ]
+    },
+    {
+      "name": "multiple expressions - KAFKA to non-supported decimal key format",
```
vcrfxia, line 602: Can we also add a negative test for this case? Specifically, the same test case without `WITH (KEY_FORMAT='JSON')` should fail.

Let's also change one of the `WITH (KEY_FORMAT='JSON')` to `WITH (KEY_FORMAT='AVRO')` to guard against bugs where the explicit format is ignored in favor of the internal default.

Thanks!

ksqldb-serde/src/main/java/io/confluent/ksql/serde/KeyFormat.java
abbccdda: Sounds good!

ksqldb-serde/src/main/java/io/confluent/ksql/serde/kafka/KafkaFormat.java
abbccdda: I think so, just a double safety approach.

vcrfxia (approved):
Thanks @abbccdda -- LGTM!

PR #7343, opened 2021-04-05
test: retries for known flaky RestApiTest

agavra:
### Description 

Add a retry around a known flaky test to help stabilize master build. I know this isn't the best thing to do, but I think we have a pretty good understanding about why this test is flaky (cc @AlanConfluent I believe you investigated this, can you comment here so that we have history recorded?)

### Testing done 

I see the following in the logs:
```
[2021-04-05 16:31:00,102] WARN Retrying test after 3 SECONDS due to: 
Expected: is "{\"row\":{\"columns\":[1,\"USER_1\"]}}]"
     but: was "{\"errorMessage\":{\"@type\":\"generic_error\",\"error_code\":50000,\"message\":\"Unable to execute pull query \\"SELECT COUNT, USERID from AGG_TABLE WHERE USERID='USER_1';\\". Error executing query locally at node http://localhost:64740/: Failed to get value from materialized table\"}}]" (io.confluent.ksql.integration.Retry:135)
```

And it passes afterwards

### Reviewer checklist

### DIFF
--- a/ksqldb-rest-app/src/test/java/io/confluent/ksql/rest/integration/RestApiTest.java
+++ b/ksqldb-rest-app/src/test/java/io/confluent/ksql/rest/integration/RestApiTest.java
@@ -48,6 +48,7 @@ import com.google.common.collect.ImmutableMap;
 import io.confluent.common.utils.IntegrationTest;
 import io.confluent.ksql.api.utils.QueryResponse;
 import io.confluent.ksql.integration.IntegrationTestHarness;
+import io.confluent.ksql.integration.Retry;
 import io.confluent.ksql.rest.ApiJsonMapper;
 import io.confluent.ksql.rest.entity.CommandId;
 import io.confluent.ksql.rest.entity.CommandId.Action;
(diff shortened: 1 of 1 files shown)


mjsax (approved):

AlanConfluent:
My investigation another time was on a slightly different test I believe where periodically the failure was due to routing to a different host than the expected one and failing the host check.  This seems different since the failure is `Failed to get value from materialized table`.  I believe this means that the materialization just isn't ready just yet and that waiting longer would work.  I also think this will likely work fine too since it's unlikely to take so long on a subsequent run.

Is it possible to only apply this rule to the known flakey test?  I think the most troublesome one is `shouldExecutePullQueryOverRest`.  Just an idea.

agavra:
> Is it possible to only apply this rule to the known flakey test? I think the most troublesome one is shouldExecutePullQueryOverRest. Just an idea.

That isn't possible with JUnit rules as far as I'm aware - I can build a retry loop directly into the test, but I'm not sure that's cleaner (a little harder to audit where the rule is being used)

PR #7355, opened 2021-04-07
chore: remove old rest client from migrations tool

jzaralim:
### Description 
Switch to Java api to execute `CREATE CONNECTOR` and `DROP CONNECTOR`. 

### Testing done 
Updated unit tests, existing integration test still passes

### Reviewer checklist

### DIFF
--- a/ksqldb-tools/src/main/java/io/confluent/ksql/tools/migrations/commands/ApplyMigrationCommand.java
+++ b/ksqldb-tools/src/main/java/io/confluent/ksql/tools/migrations/commands/ApplyMigrationCommand.java
@@ -46,7 +46,6 @@ import io.confluent.ksql.tools.migrations.util.MigrationsUtil;
 import io.confluent.ksql.util.KsqlException;
 import io.confluent.ksql.util.RetryUtil;
 import java.time.Clock;
-import java.util.Arrays;
 import java.util.Collections;
 import java.util.HashMap;
 import java.util.List;
--- a/ksqldb-tools/src/main/java/io/confluent/ksql/tools/migrations/commands/ApplyMigrationCommand.java
+++ b/ksqldb-tools/src/main/java/io/confluent/ksql/tools/migrations/commands/ApplyMigrationCommand.java
@@ -15,6 +15,7 @@
 
 package io.confluent.ksql.tools.migrations.commands;
 
+import static io.confluent.ksql.tools.migrations.util.CommandParser.preserveCase;
 import static io.confluent.ksql.tools.migrations.util.MigrationsDirectoryUtil.getAllMigrations;
 import static io.confluent.ksql.tools.migrations.util.MigrationsDirectoryUtil.getMigrationForVersion;
 import static io.confluent.ksql.tools.migrations.util.MigrationsDirectoryUtil.getMigrationsDirFromConfigFile;
@@ -336,7 +337,7 @@ public class ApplyMigrationCommand extends BaseCommand {
       final List<FieldInfo> fields =
           ksqlClient.describeSource(((SqlInsertValues) command).getSourceName()).get().fields();
       ksqlClient.insertInto(
-          preserveCase(((SqlInsertValues) command).getSourceName()),
+          ((SqlInsertValues) command).getSourceName(),
           getRow(
               fields,
               ((SqlInsertValues) command).getColumns(),
@@ -489,9 +490,5 @@ public class ApplyMigrationCommand extends BaseCommand {
   ) {
     LOGGER.info("Validating current migration state before applying new migrations");
     return ValidateMigrationsCommand.validate(config, migrationsDir, ksqlClient);
-  }
-
-  private static String preserveCase(final String fieldOrSourceName) {
-    return "`" + fieldOrSourceName + "`";
   }
 }
--- a/ksqldb-tools/src/main/java/io/confluent/ksql/tools/migrations/commands/ApplyMigrationCommand.java
+++ b/ksqldb-tools/src/main/java/io/confluent/ksql/tools/migrations/commands/ApplyMigrationCommand.java
@@ -30,13 +30,12 @@ import io.confluent.ksql.api.client.FieldInfo;
 import io.confluent.ksql.api.client.KsqlObject;
 import io.confluent.ksql.execution.expression.tree.Expression;
 import io.confluent.ksql.rest.client.KsqlRestClient;
-import io.confluent.ksql.rest.client.RestResponse;
-import io.confluent.ksql.rest.entity.KsqlEntityList;
 import io.confluent.ksql.tools.migrations.MigrationConfig;
 import io.confluent.ksql.tools.migrations.MigrationException;
 import io.confluent.ksql.tools.migrations.util.CommandParser;
 import io.confluent.ksql.tools.migrations.util.CommandParser.SqlCommand;
-import io.confluent.ksql.tools.migrations.util.CommandParser.SqlConnectorStatement;
+import io.confluent.ksql.tools.migrations.util.CommandParser.SqlCreateConnectorStatement;
+import io.confluent.ksql.tools.migrations.util.CommandParser.SqlDropConnectorStatement;
 import io.confluent.ksql.tools.migrations.util.CommandParser.SqlInsertValues;
 import io.confluent.ksql.tools.migrations.util.CommandParser.SqlPropertyCommand;
 import io.confluent.ksql.tools.migrations.util.CommandParser.SqlStatement;
@@ -48,6 +47,7 @@ import io.confluent.ksql.tools.migrations.util.MigrationsUtil;
 import io.confluent.ksql.util.KsqlException;
 import io.confluent.ksql.util.RetryUtil;
 import java.time.Clock;
+import java.util.Arrays;
 import java.util.Collections;
 import java.util.HashMap;
 import java.util.List;
@@ -138,7 +138,6 @@ public class ApplyMigrationCommand extends BaseCommand {
     return command(
         config,
         MigrationsUtil::getKsqlClient,
-        MigrationsUtil::createRestClient,
         getMigrationsDirFromConfigFile(getConfigFile()),
         Clock.systemDefaultZone()
     );
@@ -149,7 +148,6 @@ public class ApplyMigrationCommand extends BaseCommand {
   int command(
       final MigrationConfig config,
       final Function<MigrationConfig, Client> clientSupplier,
-      final Function<MigrationConfig, KsqlRestClient> restClientSupplier,
       final String migrationsDir,
       final Clock clock
   ) {
@@ -158,7 +156,6 @@ public class ApplyMigrationCommand extends BaseCommand {
     final KsqlRestClient restClient;
     try {
       ksqlClient = clientSupplier.apply(config);
-      restClient = restClientSupplier.apply(config);
     } catch (MigrationException e) {
       LOGGER.error(e.getMessage());
       return 1;
@@ -177,13 +174,12 @@ public class ApplyMigrationCommand extends BaseCommand {
     boolean success;
     try {
       success = validateCurrentState(config, ksqlClient, migrationsDir)
-          && apply(config, ksqlClient, restClient, migrationsDir, clock);
+          && apply(config, ksqlClient, migrationsDir, clock);
     } catch (MigrationException e) {
       LOGGER.error(e.getMessage());
       success = false;
     } finally {
       ksqlClient.close();
-      restClient.close();
     }
 
     return success ? 0 : 1;
@@ -192,7 +188,6 @@ public class ApplyMigrationCommand extends BaseCommand {
   private boolean apply(
       final MigrationConfig config,
       final Client ksqlClient,
-      final KsqlRestClient restClient,
       final String migrationsDir,
       final Clock clock
   ) {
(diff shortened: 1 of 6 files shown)


ksqldb-tools/src/main/java/io/confluent/ksql/tools/migrations/util/CommandParser.java
```diff
@@ -52,6 +54,8 @@
       Pattern.compile("\\s*SET\\s+'((?:[^']*|(?:''))*)'\\s*=\\s*'((?:[^']*|(?:''))*)'\\s*;\\s*");
   private static final Pattern UNSET_PROPERTY =
       Pattern.compile("\\s*UNSET\\s+'((?:[^']*|(?:''))*)'\\s*;\\s*");
+  private static final Pattern DROP_CONNECTOR =
+      Pattern.compile("\\s*DROP\\s+CONNECTOR\\s+((?:[^']*|(?:''))*)\\s*;\\s*");
```
vcrfxia, line 58: Why is the connector name portion of this regex escaping single quotes? I wouldn't expect those to be valid within an identifier.

ksqldb-tools/src/main/java/io/confluent/ksql/tools/migrations/util/CommandParser.java
```diff
@@ -192,7 +197,9 @@ public static Object toFieldType(final Expression expressionValue) {
   /**
   * Determines the type of command a sql string is and returns a SqlCommand.
   */
+  // CHECKSTYLE_RULES.OFF: CyclomaticComplexity
```
vcrfxia, line 200: Rather than disabling this checkstyle rule, I think we'd have better readability by pulling out the logic for each case in this method into separate helper methods.

ksqldb-tools/src/main/java/io/confluent/ksql/tools/migrations/util/CommandParser.java
```diff
@@ -214,8 +221,35 @@ private static SqlCommand transformToSqlCommand(final String sql) {
...
+        );
+      case DROP_CONNECTOR:
+        final Matcher dropConnectorMatcher = DROP_CONNECTOR.matcher(sql);
+        if (!dropConnectorMatcher.matches()) {
+          throw new MigrationException("Invalid DROP CONNECTOR command: " + sql);
+        }
+        final String connectorName = dropConnectorMatcher.group(1);
+        return new SqlDropConnectorStatement(
+            sql,
+            connectorName.startsWith("`")
```
vcrfxia, line 249: See my comment about case-sensitivity in the Java client on your other PR. Also, even if we were to go that route, this logic here is not correct since case-sensitivity of identifiers can also be preserved with double quotes (not just backticks).

ksqldb-tools/src/main/java/io/confluent/ksql/tools/migrations/util/CommandParser.java
```diff
@@ -350,11 +386,54 @@ public String getSourceName() {
...
+    SqlCreateConnectorStatement(
+        final String command,
+        final String name,
+        final boolean isSource,
+        final Map<String, Object> properties
+    ) {
       super(command);
+      this.name = name;
+      this.isSource = isSource;
+      this.properties = properties;
```
vcrfxia, line 405: nit: validate that this is non-null, and similarly for the other params and SqlDropConnectorStatement below.

As an FYI, Java primitive types (including boolean) don't need the validation since they may not be null. (I notice that `SqlPropertyCommand` has a non-null check for the boolean `set` param but not the `Optional` `value` param -- those should be flipped.)

ksqldb-tools/src/test/java/io/confluent/ksql/tools/migrations/util/CommandParserTest.java
```diff
@@ -280,10 +281,21 @@ public void shouldParseConnectorStatements() {
...
+    assertThat(((SqlCreateConnectorStatement) commands.get(0)).getProperties().get("connector.class"), is("io.confluent.connect.jdbc.JdbcSourceConnector"));
+    assertThat(((SqlCreateConnectorStatement) commands.get(0)).getProperties().get("connection.url"), is("jdbc:postgresql://localhost:5432/my.db"));
+    assertThat(((SqlCreateConnectorStatement) commands.get(0)).getProperties().get("mode"), is("bulk"));
+    assertThat(((SqlCreateConnectorStatement) commands.get(0)).getProperties().get("topic.prefix"), is("jdbc-"));
+    assertThat(((SqlCreateConnectorStatement) commands.get(0)).getProperties().get("table.whitelist"), is("users"));
+    assertThat(((SqlCreateConnectorStatement) commands.get(0)).getProperties().get("key"), is("username"));
+    assertThat(commands.get(0).getCommand(), is(createConnector));
+    assertThat(commands.get(1), instanceOf(SqlDropConnectorStatement.class));
     assertThat(commands.get(1).getCommand(), is(dropConnector));
+    assertThat(((SqlDropConnectorStatement) commands.get(1)).getName(), is("jdbc-connector"));
```
vcrfxia, line 298: nit: can we break this up into separate tests for CREATE CONNECTOR and DROP CONNECTOR?

vcrfxia (commented):
Thanks @jzaralim ! Comments inline but LGTM overall.

ksqldb-tools/src/main/java/io/confluent/ksql/tools/migrations/util/CommandParser.java
jzaralim: Updated to pass quotes to the java client. I also added quotes to the SqlCreateConnector name, since the parser removes quotes. Is this something we would need to do for `INSERT INTO...VALUES` as well? I'm not sure how the endpoint deals with case-sensitive names.

ksqldb-tools/src/test/java/io/confluent/ksql/tools/migrations/util/CommandParserTest.java
```diff
@@ -273,17 +274,39 @@ public void shouldParseConnectorStatements() {
...
+    assertThat(((SqlCreateConnectorStatement) commands.get(0)).getProperties().get("mode"), is("bulk"));
+    assertThat(((SqlCreateConnectorStatement) commands.get(0)).getProperties().get("topic.prefix"), is("jdbc-"));
+    assertThat(((SqlCreateConnectorStatement) commands.get(0)).getProperties().get("table.whitelist"), is("users"));
+    assertThat(((SqlCreateConnectorStatement) commands.get(0)).getProperties().get("key"), is("username"));
+  }
+
+  @Test
+  public void shouldParseDropConnectorStatement() {
+    // Given:
+    final String dropConnector = "DROP CONNECTOR `jdbc-connector`;";
```
vcrfxia, line 299: Can we add some whitespace after the connector name to check that the regex properly leaves out whitespace between the connector name and the semicolon? It's not clear to me from reading the regex whether the greediness of the matching groups is such that this currently works or not.
```suggestion
    final String dropConnector = "DROP CONNECTOR `jdbc-connector` ;";
```

ksqldb-tools/src/main/java/io/confluent/ksql/tools/migrations/util/CommandParser.java
```diff
@@ -199,49 +204,92 @@ private static SqlCommand transformToSqlCommand(final String sql) {
...
-        }
-        return new SqlPropertyCommand(
-            sql, false, unsetPropertyMatcher.group(1), Optional.empty());
+        return getUnsetPropertyCommand(sql);
       default:
         throw new IllegalStateException();
     }
   }
 
+  private static SqlInsertValues getInsertValueStatement(final String sql) {
```
vcrfxia, line 223: super nit:
```suggestion
  private static SqlInsertValues getInsertValueStatement(final String sql) {
```

vcrfxia (approved):
Thanks @jzaralim ! LGTM with one question inline.

ksqldb-tools/src/main/java/io/confluent/ksql/tools/migrations/util/CommandParser.java
jzaralim: Do you mean `getInsertValuesStatement`?

ksqldb-tools/src/test/java/io/confluent/ksql/tools/migrations/util/CommandParserTest.java
jzaralim: Good catch, it did not leave out whitespace. Updated the regex so that it checks for either identifiers with no space and backquoted/double quoted identifiers with pretty much anything inside. I think that should cover all the cases.

ksqldb-tools/src/main/java/io/confluent/ksql/tools/migrations/util/CommandParser.java
jzaralim: False alarm - backquotes are added to the INSERT INTO source names at another point. I added a refactor to this commit so that the quotes are added in `CommandParser`, since the logic is used twice

ksqldb-tools/src/main/java/io/confluent/ksql/tools/migrations/util/CommandParser.java
vcrfxia: Yeah. Thanks for reading my mind 😁

ksqldb-tools/src/main/java/io/confluent/ksql/tools/migrations/util/CommandParser.java
vcrfxia: Thanks for checking. What about the INSERT VALUES field names? Those should have the same behavior as the stream name.

ksqldb-tools/src/test/java/io/confluent/ksql/tools/migrations/util/CommandParserTest.java
vcrfxia: Hm, ideally there'd be a way to tell the whitespace portion of the original regex (after the connector name and before the semicolon) to be greedy (alternatively, tell the connector name to be un-greedy) to avoid needing to hardcode into the regex the fact that connector names can be wrapped in backticks, double quotes, or neither. I think this should be possible but I haven't played around with it. If you give it a try and it doesn't work then this approach is fine too.

ksqldb-tools/src/main/java/io/confluent/ksql/tools/migrations/commands/ApplyMigrationCommand.java
```diff
@@ -15,6 +15,7 @@
 
 package io.confluent.ksql.tools.migrations.commands;
 
+import static io.confluent.ksql.tools.migrations.util.CommandParser.preserveCase;
```
vcrfxia, line 18: This import is unused, right?

ksqldb-tools/src/main/java/io/confluent/ksql/tools/migrations/util/CommandParser.java
```diff
@@ -343,6 +343,10 @@ private static void validateSupportedStatementType(final List<String> tokens) {
     }
   }
 
+  public static String preserveCase(final String fieldOrSourceName) {
```
vcrfxia, line 346: Does this need to be public?

ksqldb-tools/src/test/java/io/confluent/ksql/tools/migrations/util/CommandParserTest.java
```diff
@@ -296,7 +296,7 @@ public void shouldParseCreateConnectorStatement() {
   @Test
   public void shouldParseDropConnectorStatement() {
     // Given:
-    final String dropConnector = "DROP CONNECTOR `jdbc-connector`;";
+    final String dropConnector = "DROP CONNECTOR `jdbc-connector` ;";
```
vcrfxia, line 299: Let's add a comment so future readers know the motivation for including a space before the semicolon here, since the intention isn't obvious otherwise.

ksqldb-tools/src/test/java/io/confluent/ksql/tools/migrations/util/CommandParserTest.java
vcrfxia: If we leave the regex as is (with the fact that connector names can be wrapped in backticks, double quotes, or neither encapsulated in the regex), let's add a comment to explain the regex since it's really gnarly/difficult to parse otherwise.

ksqldb-tools/src/main/java/io/confluent/ksql/tools/migrations/commands/ApplyMigrationCommand.java
jzaralim: It's used to preserve case for field names in insert statements. This is being done here instead of in CommandParser, because the client is used to get field names.

ksqldb-tools/src/main/java/io/confluent/ksql/tools/migrations/util/CommandParser.java
jzaralim: See comment below on why `preserveCase` is public

ksqldb-tools/src/test/java/io/confluent/ksql/tools/migrations/util/CommandParserTest.java
jzaralim: :+1: that ended up working!

PR #7356, opened 2021-04-07
fix: add quotes to connector and source names in java client

jzaralim:
### Description 
The functions `describeConnector`, `createConnector`, `dropConnector` and `describeSource` did not add quotes to the sql commands sent to the rest endpoints, so case sensitive names weren't preserved. 

### Testing done 
updated unit tests, integration test still passes

### Reviewer checklist

### DIFF
--- a/ksqldb-api-client/src/main/java/io/confluent/ksql/api/client/impl/ClientImpl.java
+++ b/ksqldb-api-client/src/main/java/io/confluent/ksql/api/client/impl/ClientImpl.java
@@ -302,7 +302,7 @@ public class ClientImpl implements Client {
     makePostRequest(
         KSQL_ENDPOINT,
         new JsonObject()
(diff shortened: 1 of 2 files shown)


vcrfxia:
Hey @jzaralim , this functionality is intentional (as least as of today -- we could choose to explicitly change it if this is undesirable, but I think the change is more complicated than what's in this PR). Users can add quotes themselves before passing the strings to the methods, if they want to preserve case-sensitivity. Seems what we really want is (1) a product-level decision on the desired behavior, and (2) a docs update. WDYT? cc @colinhicks as well

jzaralim:
@vcrfxia That makes sense, I agree that this should be documented.  I'll close this if the consensus is to keep this behaviour.

vvcephei:

PR #7357, opened 2021-04-08
feat: implemented working query anonymizer

tolgadur:
Jira: [KCI-341](https://confluentinc.atlassian.net/browse/KCI-341)

### Description 
This is the first userstory in the anonymize query epic [KCI-340](https://confluentinc.atlassian.net/browse/KCI-340). Here, a class is defined that takes in a ksqldb query and returns the same statement with anonymised property names, column names, stream names etc. 

For example, the following query...
```
ALTER STREAM my_stream ADD COLUMN c3 INT, ADD COLUMN c4 INT;
```

...is anonymised to: 
```
ALTER STREAM stream1 (ADD COLUMN column1 INT, ADD COLUMN column2 INT);
```

The output in this story is all printed in the same line as discussed on [Jira](https://confluentinc.atlassian.net/browse/KCI-139?focusedCommentId=496131). The pretty printing according to specified rules from the product team is going to be done in [KCI-377](https://confluentinc.atlassian.net/browse/KCI-377).

### Testing done 
Unit- and approval testing to make sure anonymizer works as expected.

### Reviewer checklist


ksqldb-parser/pom.xml
```diff
@@ -84,6 +84,18 @@
             <version>${io.confluent.ksql.version}</version>
             <scope>test</scope>
         </dependency>
+      <dependency>
```
swist, line 87: Do you actually need this? If you know how to do it, we should depend on one version of approval tests throughout (via ksql-test-util or parent POM), if not let me know and I'll task myself up with that

ksqldb-parser/src/main/java/io/confluent/ksql/util/QueryAnonymizer.java
```diff
@@ -0,0 +1,776 @@
...
+import java.util.Hashtable;
+import java.util.List;
+import java.util.stream.Collectors;
+import org.antlr.v4.runtime.ParserRuleContext;
+import org.antlr.v4.runtime.tree.TerminalNode;
+import org.apache.commons.lang3.StringUtils;
+
+public class QueryAnonymizer {
+
+  public String anonymize(final String query) {
```
swist, line 84: Can you add an overload that also anonymizes from `ParserRuleContext` - just so that if we use it in situ we have a choice of passing the string or parseTree

ksqldb-parser/src/main/java/io/confluent/ksql/util/QueryAnonymizer.java
```diff
@@ -0,0 +1,776 @@
...
+    public String visitSingleStatement(final SingleStatementContext context) {
+      return visit(context.statement());
+    }
+
+    @Override
+    public String visitType(final TypeContext context) {
+      if (context.type().isEmpty()) {
+        return context.getText();
+      }
+      final List<String> typeList = new ArrayList<>();
```
swist, line 123: What happens if you mutate the nodes? Can you even do that? (not review, just curious)

swist (approved):
LGTM just two questions that would be good to have answers to before merging

ksqldb-parser/src/test/java/io/confluent/ksql/util/QueryAnonymizerTest.java
```diff
@@ -0,0 +1,209 @@
...
+  @Test
+  public void shouldAnonymizeInsertValuesCorrectly() {
+    final String output = anon.anonymize(
+        "INSERT INTO foo (ROWTIME, KEY_COL, COL_A) VALUES (1510923225000, 'key', 'A');");
+
+    Approvals.verify(output);
+  }
+
+  @Test
+  public void shouldAnonymizeDropStatementsCorrectly() {
```
lct45, line 199: might be good to add a UDF test here as well

lct45 (approved):
Overall LGTM - had a few questions about handling UDFs / inline functions

ksqldb-parser/src/main/java/io/confluent/ksql/util/QueryAnonymizer.java
```diff
@@ -0,0 +1,776 @@
...
+import java.util.List;
+import java.util.stream.Collectors;
+import org.antlr.v4.runtime.ParserRuleContext;
+import org.antlr.v4.runtime.tree.TerminalNode;
+import org.apache.commons.lang3.StringUtils;
+
+public class QueryAnonymizer {
+
+  public String anonymize(final String query) {
+    final ParserRuleContext tree = getParseTree(query);
```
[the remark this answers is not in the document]
tolgadur: I was originally getting the parsetree in a more cumbersome way but forgot the refactor. Thanks for pointing it out, I removed the additional private method :)

ksqldb-parser/src/main/java/io/confluent/ksql/util/QueryAnonymizer.java
```diff
@@ -0,0 +1,776 @@
...
+    private final Hashtable<String, String> anonTable = new Hashtable<>();
+
+    @Override
+    public String visitStatements(final StatementsContext context) {
+      final List<String> statementList = new ArrayList<>();
+      for (final SingleStatementContext stmtContext : context.singleStatement()) {
+        final String statement = visitSingleStatement(stmtContext);
+        statementList.add(statement);
+      }
+      return statementList.get(0);
```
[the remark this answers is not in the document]
tolgadur: When writing this I assumed that it will only ever be one statement but you are right, there might be instances where we anonymize multiple statements at once. I updated it to return a semicolon separated string will all anonymized statements instead :)

ksqldb-parser/src/main/java/io/confluent/ksql/util/QueryAnonymizer.java
```diff
@@ -0,0 +1,776 @@
...
+      if (context.EXTENDED() != null) {
+        stringBuilder.append(" EXTENDED");
+      }
+
+      return String.format("%s;", stringBuilder.toString());
+    }
+
+    @Override
+    public String visitDescribeFunction(final DescribeFunctionContext context) {
+      return "DESCRIBE FUNCTION function;";
```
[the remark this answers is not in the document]
tolgadur: Yes, that's right. I can add this if this is something we want though?

ksqldb-parser/pom.xml
tolgadur: You are right, I don't need it. I removed it now :)

ksqldb-parser/src/main/java/io/confluent/ksql/util/QueryAnonymizer.java
tolgadur: Yes that works.

ksqldb-parser/src/main/java/io/confluent/ksql/util/QueryAnonymizer.java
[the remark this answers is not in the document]
lct45: I think it's probably ok, depends on how much visibility product wants into functions

ksqldb-parser/src/main/java/io/confluent/ksql/util/QueryAnonymizer.java
```diff
@@ -0,0 +1,776 @@
...
+
+      stringBuilder.append("type");
+
+      return String.format("%s;", stringBuilder.toString());
+    }
+
+    private String getQuery(final QueryContext context, final boolean isStream) {
+      final StringBuilder stringBuilder = new StringBuilder();
+
+      // visit as select items
```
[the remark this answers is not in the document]
tolgadur: Good point, I hadn't thought of UDFs and as you pointed out they would have been anonymized as any other column names. I now implemented the UDF anonymization as well.

PR #7363, opened 2021-04-09
chore: remove deprecated members

vvcephei:
close(long, TimeUnit) has finally been removed,
so we can drop our tests that verified it was a no-op
in our testing subclasses.

### DIFF
--- a/ksqldb-engine/src/test/java/io/confluent/ksql/services/SandboxedAdminClientTest.java
+++ b/ksqldb-engine/src/test/java/io/confluent/ksql/services/SandboxedAdminClientTest.java
@@ -44,6 +44,7 @@ public final class SandboxedAdminClientTest {
       return TestMethods.builder(Admin.class)
           .ignore("close")
           .ignore("close", Duration.class)
+          .ignoreUnchecked("close", long.class, TimeUnit.class)
           .setDefault(ElectLeadersOptions.class, new ElectLeadersOptions())
           .setDefault(Optional.class, Optional.empty())
           .build();
--- a/ksqldb-engine/src/test/java/io/confluent/ksql/services/SandboxedAdminClientTest.java
+++ b/ksqldb-engine/src/test/java/io/confluent/ksql/services/SandboxedAdminClientTest.java
@@ -44,7 +44,6 @@ public final class SandboxedAdminClientTest {
       return TestMethods.builder(Admin.class)
           .ignore("close")
           .ignore("close", Duration.class)
-          .ignore("close", long.class, TimeUnit.class)
           .setDefault(ElectLeadersOptions.class, new ElectLeadersOptions())
           .setDefault(Optional.class, Optional.empty())
           .build();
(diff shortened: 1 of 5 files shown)


vvcephei:
I verified this change with `./mvnw clean test-compile`

agavra (approved):

mjsax (approved):

mjsax:
Why does the test fail:
```
java.lang.AssertionError: invalid test: unknown excluded method: close([long, class java.util.concurrent.TimeUnit])
	at io.confluent.ksql.services.SandboxedAdminClientTest$UnsupportedMethods.getMethodsToTest(SandboxedAdminClientTest.java:47)
```

vvcephei:
Thanks, @mjsax , I had missed that other test case. Since it's using reflection, it did not produce a compiler error.

I've removed the three additional lines that reflectively mention the dropped methods. However, there's an unfortunate circumstance that if we remove them, ksql will not compile on the old upstream, and if we keep them, ksql will not compile on the new upstream.

Therefore, this PR will not pass until the upstream kafka dependency is published, and I don't know if the upstream kafka dependency will be published until this PR is merged.

I might put in a hack to get past this deadlock.

ksqldb-engine/src/test/java/io/confluent/ksql/services/SandboxedAdminClientTest.java
```diff
@@ -44,7 +44,7 @@ private SandboxedAdminClientTest() {
       return TestMethods.builder(Admin.class)
           .ignore("close")
           .ignore("close", Duration.class)
-          .ignore("close", long.class, TimeUnit.class)
+          .ignoreUnchecked("close", long.class, TimeUnit.class)
```
vvcephei, line 47: instead of dropping the `ignore`, I'm just switching it to not check that the referenced method actually exists.

ksqldb-test-util/src/main/java/io/confluent/ksql/test/util/TestMethods.java
```diff
@@ -65,13 +67,57 @@ private TestMethods() {
...
       this.typeUnderTest = Objects.requireNonNull(typeUnderTest, "typeUnderTest");
     }
 
+    /**
+     * The purpose of this class is to provide a more general reference to the actual Method that we can
+     * use in the blacklist. Note that the equality algorithm is the same as Method's: class name, method name,
+     * and argument types. I expect this to be a valid equality algorithm for methods indefinitely because it's
+     * the same as the method overload resolution criteria specified in the Java language.
+     */
+    public static class MethodRef {
```
vvcephei, line 82: As the doc says, I'm just creating this little "bean" class so I can blacklist methods that don't actually exist on the declaring class (as well as ones that do). Regardless, the "contains"/"equality" criteria are the same as those of the real Method.

vvcephei (commented):
Ok, this was a little more involved than I'd like, but this will let our build pass both before and after the latest sync gets published, and it also gives us a tool to use for other deprecated methods that get dropped in the future.

vvcephei:
The tests haven't actually run on that last commit yet. Still waiting for them to finish on the commit before.

PR #7365, opened 2021-04-10
fix: add sessionVariables to /ksql endpoint validator

jzaralim:
### Description 
Statements like `CREATE STREAM ${variable} ...` were failing when sent to the `/ksql` endpoint even if the request included the needed session variables because  the request validator was parsing the stream without first replacing the variables.

### Testing done 
Updated an integration test to catch this scenario, added a unit test

### Reviewer checklist

### DIFF
--- a/ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/resources/KsqlResource.java
+++ b/ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/resources/KsqlResource.java
@@ -284,7 +284,8 @@ public class KsqlResource implements KsqlConfigurable {
               configProperties,
               localHost,
               localUrl,
-              requestConfig.getBoolean(KsqlRequestConfig.KSQL_REQUEST_INTERNAL_REQUEST)
+              requestConfig.getBoolean(KsqlRequestConfig.KSQL_REQUEST_INTERNAL_REQUEST),
+              request.getSessionVariables()
           ),
           request.getKsql()
       );
--- a/ksqldb-rest-app/src/test/java/io/confluent/ksql/rest/server/resources/KsqlResourceTest.java
+++ b/ksqldb-rest-app/src/test/java/io/confluent/ksql/rest/server/resources/KsqlResourceTest.java
@@ -1198,7 +1198,7 @@ public class KsqlResourceTest {
   }
 
   @Test
-  public void shouldSupportVariableSubstitutionWithInitialVariables() {
+  public void shouldSupportVariableSubstitutionWithVariablesInRequest() {
     // Given:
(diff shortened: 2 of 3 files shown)


ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/resources/KsqlResource.java
```diff
@@ -284,7 +284,8 @@ public EndpointResponse handleKsqlStatements(
               configProperties,
               localHost,
               localUrl,
-              requestConfig.getBoolean(KsqlRequestConfig.KSQL_REQUEST_INTERNAL_REQUEST)
+              requestConfig.getBoolean(KsqlRequestConfig.KSQL_REQUEST_INTERNAL_REQUEST),
+              request.getSessionVariables()
```
vcrfxia, line 288: Yikes, thanks for finding and fixing this bug! Unless SessionProperties (or any of its params) are mutable, it'd be good to pull out the creation of the SessionProperties to be shared between the validate() and execute() calls in this method, in order to guard against similar bugs in the future.

ksqldb-rest-app/src/test/java/io/confluent/ksql/rest/server/resources/KsqlResourceTest.java
```diff
@@ -1197,6 +1197,34 @@ public void shouldSupportVariableSubstitution() {
     assertThat(results.get(0).getStatementText(), is(csasSubstituted));
   }
 
+  @Test
+  public void shouldSupportVariableSubstitutionWithInitialVariables() {
```
vcrfxia, line 1201: nit: (I wasn't sure what "initial variables" meant until I looked at the other test for comparison)
```suggestion
  public void shouldSupportVariableSubstitutionWithVariablesInRequest() {
```

vcrfxia (approved):
Thanks @jzaralim -- LGTM with comments inline.

ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/resources/KsqlResource.java
jzaralim: One of the properties does actually get mutated unfortunately!

vcrfxia: Bummer (but thanks for checking)! Which one, out of curiosity?

jzaralim: At some point, some items get removed from`mutableScopedProperties`

PR #7366, opened 2021-04-10
feat(migrations): support DEFINE and UNDEFINE statements

jzaralim:
### Description 
^^^

Previously, the tool parsed each command, collected them and then executed them. Because the variables are tracked in the java client, but SET/UNSET/INSERT VALUES don't do variable substitution through the client, the flow was refactored to parse commands and then execute them one by one.

### Testing done 
unit + integration

### Reviewer checklist

### DIFF
--- a/docs/operate-and-deploy/migrations-tool.md
+++ b/docs/operate-and-deploy/migrations-tool.md
@@ -51,6 +51,9 @@ types of ksqlDB statements:
 * `DEFINE <variable>`
 * `UNDEFINE <variable>`
 
+Any properties or variables set using the `SET`, `UNSET`, `DEFINE` and `UNDEFINE` are applied in the 
+current migration file only. They do not carry over to the next migration file.
+
 Requirements and Installation
 -----------------------------
 
--- a/docs/operate-and-deploy/migrations-tool.md
+++ b/docs/operate-and-deploy/migrations-tool.md
@@ -48,6 +48,8 @@ types of ksqlDB statements:
 * `DROP TYPE`
 * `SET <property>`
 * `UNSET <property>`
+* `DEFINE <variable>`
+* `UNDEFINE <variable>`
 
 Requirements and Installation
 -----------------------------
--- a/docs/operate-and-deploy/migrations-tool.md
+++ b/docs/operate-and-deploy/migrations-tool.md
@@ -52,7 +52,8 @@ types of ksqlDB statements:
 * `UNDEFINE <variable>`
 
 Any properties or variables set using the `SET`, `UNSET`, `DEFINE` and `UNDEFINE` are applied in the 
-current migration file only. They do not carry over to the next migration file.
+current migration file only. They do not carry over to the next migration file, even if multiple
+migration files are applied as part of the same `ksql-migrations apply` command
 
 Requirements and Installation
 -----------------------------
--- a/ksqldb-tools/src/main/java/io/confluent/ksql/tools/migrations/commands/ApplyMigrationCommand.java
+++ b/ksqldb-tools/src/main/java/io/confluent/ksql/tools/migrations/commands/ApplyMigrationCommand.java
@@ -315,8 +315,10 @@ public class ApplyMigrationCommand extends BaseCommand {
   ) {
     final List<String> commands = CommandParser.splitSql(migrationFileContent);
 
-    executeCommands(commands, ksqlClient, config, executionStart, migration, clock, previous, true);
-    executeCommands(commands, ksqlClient, config, executionStart, migration, clock, previous, false);
+    executeCommands(
+        commands, ksqlClient, config, executionStart, migration, clock, previous, true);
+    executeCommands(
+        commands, ksqlClient, config, executionStart, migration, clock, previous, false);
   }
 
   private void executeCommands(
@@ -336,7 +338,11 @@ public class ApplyMigrationCommand extends BaseCommand {
         final Map<String, String> variables = ksqlClient.getVariables().entrySet()
             .stream().collect(Collectors.toMap(e -> e.getKey(), e -> e.getValue().toString()));
         executeCommand(
-            CommandParser.transformToSqlCommand(command, variables), ksqlClient, properties, verify);
+            CommandParser.transformToSqlCommand(command, variables),
+            ksqlClient,
+            properties,
+            verify
+        );
       } catch (InterruptedException | ExecutionException | MigrationException e) {
         final String action = verify ? "parse" : "execute";
         final String errorMsg = String.format(
@@ -358,9 +364,30 @@ public class ApplyMigrationCommand extends BaseCommand {
       final Map<String, Object> properties,
       final boolean defineUndefineOnly
   ) throws ExecutionException, InterruptedException {
-    if (command instanceof SqlStatement && !defineUndefineOnly) {
+    if (!defineUndefineOnly) {
+      executeServerCommands(command, ksqlClient, properties);
+    }
+    if (command instanceof SqlDefineVariableCommand) {
+      ksqlClient.define(
+          ((SqlDefineVariableCommand) command).getVariable(),
+          ((SqlDefineVariableCommand) command).getValue()
+      );
+    } else if (command instanceof SqlUndefineVariableCommand) {
+      ksqlClient.undefine(((SqlUndefineVariableCommand) command).getVariable());
+    }
+  }
+
+  /**
+  * Executes commands that get sent to the server (everything besides define/undefine commands)
+  */
+  private void executeServerCommands(
+      final SqlCommand command,
+      final Client ksqlClient,
+      final Map<String, Object> properties
+  ) throws ExecutionException, InterruptedException {
+    if (command instanceof SqlStatement) {
       ksqlClient.executeStatement(command.getCommand(), new HashMap<>(properties)).get();
-    } else if (command instanceof SqlInsertValues && !defineUndefineOnly) {
+    } else if (command instanceof SqlInsertValues) {
       final List<FieldInfo> fields =
           ksqlClient.describeSource(((SqlInsertValues) command).getSourceName()).get().fields();
       ksqlClient.insertInto(
@@ -369,15 +396,15 @@ public class ApplyMigrationCommand extends BaseCommand {
               fields,
               ((SqlInsertValues) command).getColumns(),
               ((SqlInsertValues) command).getValues())).get();
-    } else if (command instanceof SqlCreateConnectorStatement && !defineUndefineOnly) {
+    } else if (command instanceof SqlCreateConnectorStatement) {
       ksqlClient.createConnector(
           ((SqlCreateConnectorStatement) command).getName(),
           ((SqlCreateConnectorStatement) command).isSource(),
           ((SqlCreateConnectorStatement) command).getProperties()
       ).get();
-    } else if (command instanceof SqlDropConnectorStatement && !defineUndefineOnly) {
+    } else if (command instanceof SqlDropConnectorStatement) {
       ksqlClient.dropConnector(((SqlDropConnectorStatement) command).getName()).get();
-    } else if (command instanceof SqlPropertyCommand && !defineUndefineOnly) {
+    } else if (command instanceof SqlPropertyCommand) {
       if (((SqlPropertyCommand) command).isSetCommand()
           && ((SqlPropertyCommand) command).getValue().isPresent()) {
         properties.put(
@@ -387,13 +414,6 @@ public class ApplyMigrationCommand extends BaseCommand {
       } else {
         properties.remove(((SqlPropertyCommand) command).getProperty());
       }
-    } else if (command instanceof SqlDefineVariableCommand) {
-      ksqlClient.define(
-          ((SqlDefineVariableCommand) command).getVariable(),
-          ((SqlDefineVariableCommand) command).getValue()
-      );
-    } else if (command instanceof SqlUndefineVariableCommand) {
-      ksqlClient.undefine(((SqlUndefineVariableCommand) command).getVariable());
     }
   }
 
--- a/ksqldb-tools/src/main/java/io/confluent/ksql/tools/migrations/commands/ApplyMigrationCommand.java
+++ b/ksqldb-tools/src/main/java/io/confluent/ksql/tools/migrations/commands/ApplyMigrationCommand.java
@@ -321,6 +321,12 @@ public class ApplyMigrationCommand extends BaseCommand {
         commands, ksqlClient, config, executionStart, migration, clock, previous, false);
   }
 
+  /**
+   * If validateOnly is set to true, then this parses each of the commands but only executes
+   * DEFINE/UNDEFINE commands (variables are needed for parsing INSERT INTO... VALUES, SET/UNSET
+   * and DEFINE commands). If validateOnly is set to fales, then each command will execute after
+   * parsing.
+   */
   private void executeCommands(
       final List<String> commands,
       final Client ksqlClient,
@@ -329,7 +335,7 @@ public class ApplyMigrationCommand extends BaseCommand {
       final MigrationFile migration,
       final Clock clock,
       final String previous,
-      final boolean verify
+      final boolean validateOnly
   ) {
     cleanUpJavaClientVariables(ksqlClient);
     final Map<String, Object> properties = new HashMap<>();
@@ -341,10 +347,10 @@ public class ApplyMigrationCommand extends BaseCommand {
             CommandParser.transformToSqlCommand(command, variables),
             ksqlClient,
             properties,
-            verify
+            validateOnly
         );
       } catch (InterruptedException | ExecutionException | MigrationException e) {
-        final String action = verify ? "parse" : "execute";
+        final String action = validateOnly ? "parse" : "execute";
         final String errorMsg = String.format(
             "Failed to %s sql: %s. Error: %s", action, command, e.getMessage());
         updateState(config, ksqlClient, MigrationState.ERROR,
@@ -365,7 +371,7 @@ public class ApplyMigrationCommand extends BaseCommand {
       final boolean defineUndefineOnly
   ) throws ExecutionException, InterruptedException {
     if (!defineUndefineOnly) {
-      executeServerCommands(command, ksqlClient, properties);
+      executeNonVariableCommands(command, ksqlClient, properties);
     }
     if (command instanceof SqlDefineVariableCommand) {
       ksqlClient.define(
@@ -378,9 +384,9 @@ public class ApplyMigrationCommand extends BaseCommand {
   }
 
   /**
-  * Executes commands that get sent to the server (everything besides define/undefine commands)
-  */
-  private void executeServerCommands(
+   * Executes everything besides define/undefine commands
+   */
+  private void executeNonVariableCommands(
       final SqlCommand command,
       final Client ksqlClient,
       final Map<String, Object> properties
--- a/ksqldb-tools/src/main/java/io/confluent/ksql/tools/migrations/commands/ApplyMigrationCommand.java
+++ b/ksqldb-tools/src/main/java/io/confluent/ksql/tools/migrations/commands/ApplyMigrationCommand.java
@@ -313,18 +313,34 @@ public class ApplyMigrationCommand extends BaseCommand {
       final Clock clock,
       final String previous
   ) {
-    final Map<String, Object> properties = new HashMap<>();
     final List<String> commands = CommandParser.splitSql(migrationFileContent);
 
+    executeCommands(commands, ksqlClient, config, executionStart, migration, clock, previous, true);
+    executeCommands(commands, ksqlClient, config, executionStart, migration, clock, previous, false);
+  }
+
+  private void executeCommands(
+      final List<String> commands,
+      final Client ksqlClient,
+      final MigrationConfig config,
+      final String executionStart,
+      final MigrationFile migration,
+      final Clock clock,
+      final String previous,
+      final boolean verify
+  ) {
+    cleanUpJavaClientVariables(ksqlClient);
+    final Map<String, Object> properties = new HashMap<>();
     for (final String command : commands) {
       try {
         final Map<String, String> variables = ksqlClient.getVariables().entrySet()
             .stream().collect(Collectors.toMap(e -> e.getKey(), e -> e.getValue().toString()));
         executeCommand(
-            CommandParser.transformToSqlCommand(command, variables), ksqlClient, properties);
-      } catch (InterruptedException | ExecutionException e) {
+            CommandParser.transformToSqlCommand(command, variables), ksqlClient, properties, verify);
+      } catch (InterruptedException | ExecutionException | MigrationException e) {
+        final String action = verify ? "parse" : "execute";
         final String errorMsg = String.format(
-            "Failed to execute sql: %s. Error: %s", command, e.getMessage());
+            "Failed to %s sql: %s. Error: %s", action, command, e.getMessage());
         updateState(config, ksqlClient, MigrationState.ERROR,
             executionStart, migration, clock, previous, Optional.of(errorMsg));
         throw new MigrationException(errorMsg);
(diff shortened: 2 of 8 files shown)


ksqldb-tools/src/test/java/io/confluent/ksql/tools/migrations/MigrationsTest.java
```diff
@@ -237,19 +237,23 @@ private void shouldApplyMigrations() throws Exception {
...
             "CREATE STREAM `bar` AS SELECT CONCAT(A, 'woo''hoo') AS A FROM FOO;" +
             "UNSET 'ksql.output.topic.name.prefix';" +
             "CREATE STREAM CAR AS SELECT * FROM FOO;" +
             "DROP CONNECTOR D;" +
             "INSERT INTO `bar` SELECT A FROM CAR;" +
             "CREATE TYPE ADDRESS AS STRUCT<number INTEGER, street VARCHAR, city VARCHAR>;" +
-            "CREATE STREAM HOMES (ADDR ADDRESS) WITH (KAFKA_TOPIC='HOMES', PARTITIONS=1, VALUE_FORMAT='JSON');" +
-            "INSERT INTO HOMES VALUES (STRUCT(number := 123, street := 'sesame st', city := 'New York City'));" +
+            "DEFINE variable = 'HOMES';" +
+            "CREATE STREAM ${variable} (ADDR ADDRESS) WITH (KAFKA_TOPIC='${variable}', PARTITIONS=1, VALUE_FORMAT='JSON');" +
```
jzaralim, line 254: This line passing depends on https://github.com/confluentinc/ksql/pull/7365

JimGalasyn (approved):

docs/operate-and-deploy/migrations-tool.md
```diff
@@ -48,6 +48,8 @@ types of ksqlDB statements:
 * `DROP TYPE`
 * `SET <property>`
 * `UNSET <property>`
+* `DEFINE <variable>`
```
vcrfxia, line 51: Let's clarify the scope for these different commands. For example, `DEFINE` and `UNDEFINE` apply for the specific migrations file but not beyond that. It'd be good to clarify the scope for `SET` and `UNSET` too.

ksqldb-tools/src/main/java/io/confluent/ksql/tools/migrations/commands/ApplyMigrationCommand.java
```diff
@@ -312,13 +314,17 @@ private void executeCommands(
       final String previous
   ) {
     final Map<String, Object> properties = new HashMap<>();
-    final List<SqlCommand> commands = CommandParser.parse(migrationFileContent);
```
vcrfxia, line 315: Ideally we would still parse everything before starting to execute any commands, to avoid situations where a migrations file is partially executed because a later statement fails to parse. As discussed offline, we can achieve this by having a validation loop that is identical to the execution loop except only DEFINE and UNDEFINE statements are executed (i.e., sent to the java client for purposes of tracking variables, without anything actually being sent to the ksqlDB server).

ksqldb-tools/src/main/java/io/confluent/ksql/tools/migrations/commands/ApplyMigrationCommand.java
```diff
@@ -312,13 +314,17 @@ private void executeCommands(
...
-    for (final SqlCommand command : commands) {
+    final List<String> commands = CommandParser.splitSql(migrationFileContent);
+
+    for (final String command : commands) {
       try {
-        executeCommand(command, ksqlClient, properties);
+        final Map<String, String> variables = ksqlClient.getVariables().entrySet()
```
vcrfxia, line 321: We need to reset variables between migrations files, else a DEFINE statement in one file will carry over to the next file, which we don't want.

Let's also add test coverage for this.

ksqldb-tools/src/main/java/io/confluent/ksql/tools/migrations/util/CommandParser.java
```diff
@@ -56,6 +57,10 @@
       Pattern.compile("\\s*UNSET\\s+'((?:[^']*|(?:''))*)'\\s*;\\s*");
   private static final Pattern DROP_CONNECTOR =
       Pattern.compile("\\s*DROP\\s+CONNECTOR\\s+(.*?)\\s*;\\s*");
+  private static final Pattern DEFINE_VARIABLE = Pattern.compile(
+      "\\s*DEFINE\\s+([a-zA-Z_][a-zA-Z0-9_@]*)\\s*=\\s*'((?:[^']*|(?:''))*)'\\s*;\\s*");
```
vcrfxia, line 61: Just checking: DEFINE statements require that the values be strings, even though the java client and HTTP endpoints allow non-string values, right? It looks like the values are converted into strings by `SessionProperties` anyway. Is there a reason we didn't have the HTTP endpoints require string values to keep things simpler/more consistent? It's worth considering whether the java client should only accept string values as well.

ksqldb-tools/src/main/java/io/confluent/ksql/tools/migrations/util/CommandParser.java
```diff
@@ -56,6 +57,10 @@
...
+  private static final Pattern UNDEFINE_VARIABLE =
+      Pattern.compile("\\s*UNDEFINE\\s+([a-zA-Z_][a-zA-Z0-9_@]*)\\s*;\\s*");
```
vcrfxia, line 63: Where did this set of valid characters in the variable name (i.e., `[a-zA-Z_][a-zA-Z0-9_@]*`) come from?

ksqldb-tools/src/main/java/io/confluent/ksql/tools/migrations/util/CommandParser.java
```diff
@@ -197,34 +197,44 @@ public static Object toFieldType(final Expression expressionValue) {
...
       case STATEMENT:
         return new SqlStatement(sql);
       case SET_PROPERTY:
-        return getSetPropertyCommand(sql);
+        return getSetPropertyCommand(sql, variables);
       case UNSET_PROPERTY:
-        return getUnsetPropertyCommand(sql);
+        return getUnsetPropertyCommand(sql, variables);
+      case DEFINE_VARIABLE:
+        return getDefineVariableCommand(sql);
```
vcrfxia, line 220: This also needs to support variable substitution, right? In case users want to define a variable based on other variables. 

I'm not sure whether it makes sense to support variable substitution for the variable name in addition to for the variable value. It'd be good to check what the ksqlDB CLI's stance on this is, and add support if needed.

ksqldb-tools/src/main/java/io/confluent/ksql/tools/migrations/util/CommandParser.java
```diff
@@ -265,22 +275,51 @@ private static SqlDropConnectorStatement getDropConnectorStatement(final String
...
     return new SqlPropertyCommand(
-        sql, false, unsetPropertyMatcher.group(1), Optional.empty());
+        sql,
+        false,
+        VariableSubstitutor.substitute(unsetPropertyMatcher.group(1), variables),
+        Optional.empty());
+  }
+
+  private static SqlDefineVariableCommand getDefineVariableCommand(final String sql) {
+    final Matcher defineVariableMatcher = DEFINE_VARIABLE.matcher(sql);
```
vcrfxia, line 306: Haven't tried this so I might be mistaken but I think we've got a bug around casing here: `transformToSqlCommand()` uppercases the sql before applying the regex matchers but we don't do that here (we can't because we don't want to strings or quoted identifiers), which means this matcher might fail to find a match (e.g., if the user lowercases `DEFINE` in their sql command). I think this bug also affects the other statement types. 

If this is true, we definitely need to fix this, and we should add test coverage for it too.

ksqldb-tools/src/main/java/io/confluent/ksql/tools/migrations/util/CommandParser.java
```diff
@@ -485,4 +528,49 @@ public String getProperty() {
...
+   * Represents ksqlDB DEFINE commands.
+   */
+  public static class SqlDefineVariableCommand extends SqlCommand {
+    private final String variable;
+    private final Object value;
+
+    SqlDefineVariableCommand(
+        final String command,
+        final String variable,
+        final Object value
```
vcrfxia, line 542: This is always passed in as a string, right? If so we should simplify this by making that explicit.

ksqldb-tools/src/test/java/io/confluent/ksql/tools/migrations/util/CommandParserTest.java
```diff
@@ -124,10 +151,27 @@ public void shouldParseSetUnsetStatements() {
     assertTrue(!((SqlPropertyCommand) commands.get(1)).getValue().isPresent());
   }
 
+  @Test
+  public void shouldParseSetUnsetStatementsWithVariables() {
```
vcrfxia, line 155: nit: follow given/when/then pattern for improved readability, and consider breaking this into two tests if the fact that the two commands are parsed together is not relevant to this test.

ksqldb-tools/src/test/java/io/confluent/ksql/tools/migrations/util/CommandParserTest.java
```diff
@@ -337,60 +381,71 @@ public void shouldSplitCommandsWithComments() {
   }
 
   @Test
-  public void shouldThrowOnMalformedComment() {
+  public void shouldParseDefineStatement() {
+    // Given:
+    final String undefineVar = "DEFINE var = 'foo';";
```
vcrfxia, line 386: nit: variable name.

vcrfxia (commented):
Thanks @jzaralim -- this is exciting! Comments inline. Also I updated the PR title to reflect that this PR adds support for DEFINE and UNDEFINE statements in the migrations tool, but doesn't add the full variable substitution support (i.e., the `--define` option) specified in the KLIP.

ksqldb-tools/src/main/java/io/confluent/ksql/tools/migrations/util/CommandParser.java
jzaralim: The variable name follows the pattern for identifiers in the parser: https://github.com/confluentinc/ksql/blob/01bb840c0142d6e4a0c51b7988c0cb30b01f1a60/ksqldb-parser/src/main/antlr4/io/confluent/ksql/parser/SqlBase.g4#L582

ksqldb-tools/src/main/java/io/confluent/ksql/tools/migrations/util/CommandParser.java
jzaralim: It follows the same pattern as `SET`, which also requires strings in the statement but can be passed as object parameters. I can see two reasons this could be convenient:

1. Users using the endpoints don't need to treat `sessionProperties` and `sessionVariables` differently
2. Client users can make requests like `client.define("name", 5);` or `client.define("name", false);` without having to convert them to strings.

These are pretty minor reasons though. Swapping everything to strings is a pretty easy change.

ksqldb-tools/src/main/java/io/confluent/ksql/tools/migrations/util/CommandParser.java
jzaralim: Updated. Only the variable values can be replaced.

ksqldb-tools/src/main/java/io/confluent/ksql/tools/migrations/util/CommandParser.java
jzaralim: Added a fix

docs/operate-and-deploy/migrations-tool.md
```diff
@@ -51,6 +51,9 @@ types of ksqlDB statements:
 * `DEFINE <variable>`
 * `UNDEFINE <variable>`
 
+Any properties or variables set using the `SET`, `UNSET`, `DEFINE` and `UNDEFINE` are applied in the 
+current migration file only. They do not carry over to the next migration file.
```
vcrfxia, line 55:
```suggestion
current migration file only. They do not carry over to the next migration file, even if multiple migration files are applied as part of the same `ksql-migrations apply` command.
```

ksqldb-tools/src/main/java/io/confluent/ksql/tools/migrations/commands/ApplyMigrationCommand.java
```diff
@@ -313,26 +313,74 @@ private void executeCommands(
...
+
+  private void executeCommands(
+      final List<String> commands,
+      final Client ksqlClient,
+      final MigrationConfig config,
+      final String executionStart,
+      final MigrationFile migration,
+      final Clock clock,
+      final String previous,
+      final boolean verify
```
vcrfxia, line 332:
```suggestion
      final boolean validateOnly
```
to clarify that this boolean param being set to true performs less than when it is set to false, not to the other way around (as the name `verify` currently suggests).

ksqldb-tools/src/main/java/io/confluent/ksql/tools/migrations/commands/ApplyMigrationCommand.java
```diff
@@ -313,26 +313,74 @@ private void executeCommands(
       final Clock clock,
       final String previous
   ) {
-    final Map<String, Object> properties = new HashMap<>();
     final List<String> commands = CommandParser.splitSql(migrationFileContent);
 
+    executeCommands(
```
vcrfxia, line 318: Add a comment explaining what this does: it doesn't actually execute anything, just steps through the commands to ensure that they parse (before the next call which actually executes statements). It'd be good for the comment to explain why it's necessary to "execute" DEFINE and UNDEFINE statements as part of this parsing validation as well.

ksqldb-tools/src/main/java/io/confluent/ksql/tools/migrations/commands/ApplyMigrationCommand.java
```diff
@@ -313,26 +313,74 @@ private void executeCommands(
...
 
+  private void cleanUpJavaClientVariables(final Client ksqlClient) {
+    ksqlClient.getVariables().forEach((k, v) -> ksqlClient.undefine(k));
+  }
+
   private void executeCommand(
+      final SqlCommand command,
+      final Client ksqlClient,
+      final Map<String, Object> properties,
+      final boolean defineUndefineOnly
```
vcrfxia, line 365: As above, we need comments explaining this, both what it does and why it's useful. (Up to you where the comments go, no need to duplicate.)

ksqldb-tools/src/main/java/io/confluent/ksql/tools/migrations/commands/ApplyMigrationCommand.java
```diff
@@ -313,26 +313,74 @@ private void executeCommands(
...
+          ((SqlDefineVariableCommand) command).getVariable(),
+          ((SqlDefineVariableCommand) command).getValue()
+      );
+    } else if (command instanceof SqlUndefineVariableCommand) {
+      ksqlClient.undefine(((SqlUndefineVariableCommand) command).getVariable());
+    }
+  }
+
+  /**
+  * Executes commands that get sent to the server (everything besides define/undefine commands)
```
vcrfxia, line 381: I thought set/unset doesn't get sent to the server either?

ksqldb-tools/src/main/java/io/confluent/ksql/tools/migrations/util/CommandParser.java
vcrfxia: Thanks. Do you know whether this is a problem for 0.17 (e.g., for INSERT VALUES)? If so, we should backport a fix to 6.2.x (requires going through the CP blocker process). We should also consider getting a fix into 0.17 but I think that's unlikely at this point, in which case we'd want to document the limitation.

ksqldb-tools/src/main/java/io/confluent/ksql/tools/migrations/util/CommandParser.java
```diff
@@ -49,18 +48,22 @@
...
-      Pattern.compile("\\s*DROP\\s+CONNECTOR\\s+(.*?)\\s*;\\s*");
+      Pattern.compile("\\s*DROP\\s+CONNECTOR\\s+(.*?)\\s*;\\s*", Pattern.CASE_INSENSITIVE);
   private static final Pattern DEFINE_VARIABLE = Pattern.compile(
-      "\\s*DEFINE\\s+([a-zA-Z_][a-zA-Z0-9_@]*)\\s*=\\s*'((?:[^']*|(?:''))*)'\\s*;\\s*");
-  private static final Pattern UNDEFINE_VARIABLE =
-      Pattern.compile("\\s*UNDEFINE\\s+([a-zA-Z_][a-zA-Z0-9_@]*)\\s*;\\s*");
+      "\\s*DEFINE\\s+([a-zA-Z_][a-zA-Z0-9_@]*)\\s*=\\s*'((?:[^']*|(?:''))*)'\\s*;\\s*",
+      Pattern.CASE_INSENSITIVE);
+  private static final Pattern UNDEFINE_VARIABLE = Pattern.compile(
+      "\\s*UNDEFINE\\s+([a-zA-Z_][a-zA-Z0-9_@]*)\\s*;\\s*", Pattern.CASE_INSENSITIVE);
```
vcrfxia, line 66: Do we still need the `toUpperCase()` call in `transformToSqlCommand()` now that we have case-insensitivity?

ksqldb-tools/src/test/java/io/confluent/ksql/tools/migrations/commands/ApplyMigrationCommandTest.java
```diff
@@ -246,6 +245,34 @@ public void shouldApplyDefineUndefineCommands() throws Exception {
     inOrder.verifyNoMoreInteractions();
   }
 
+  @Test
+  public void shouldResetVariablesBetweenMigrations() throws Exception {
```
vcrfxia, line 249: Thanks for the extra test coverage! This is great. Could we also add an equivalent test to verify that SET/UNSET are also scoped to a single migrations file? (Feel free to defer to a follow-up PR if that's more convenient.)

vcrfxia (commented):
Thanks @jzaralim -- looking good! A couple important questions but nothing major in terms of changes. Looking forward to getting this merged :)

ksqldb-tools/src/main/java/io/confluent/ksql/tools/migrations/util/CommandParser.java
vcrfxia: OK. It's not great that we've got it in two places now (the parser and also the migrations tool) but there are disadvantages to having them be shared as well. How about we add a comment to this line in the migrations tool code to reference the SqlBase.g4 line (with a git permalink) so if the parser is changed in the future and something breaks, we'll at least know where to look to update this.

ksqldb-tools/src/main/java/io/confluent/ksql/tools/migrations/util/CommandParser.java
vcrfxia: That makes sense. Let's leave this as is!

ksqldb-tools/src/main/java/io/confluent/ksql/tools/migrations/util/CommandParser.java
jzaralim: Yup, it's used to determine the type of command

ksqldb-tools/src/main/java/io/confluent/ksql/tools/migrations/util/CommandParser.java
jzaralim: It affects SET and UNSET. Filed a blocker, but IMO because there is a clear workaround (tell users to capitalize SET and UNSET), it's not *terrible* if it doesn't get included.

ksqldb-tools/src/main/java/io/confluent/ksql/tools/migrations/util/CommandParser.java
vcrfxia: But why do we need to uppercase the command if the regex is case-insensitive now?

ksqldb-tools/src/main/java/io/confluent/ksql/tools/migrations/util/CommandParser.java
vcrfxia: Agreed it's not terrible if it's not included but I think it's still better to fix it before the 6.2 release. 0.17 probably isn't worth it, though.

ksqldb-tools/src/main/java/io/confluent/ksql/tools/migrations/util/CommandParser.java
jzaralim: The regexes are used to determine things like the name of a connector being dropped. There is some logic before that for determining the type of command or if the command is supported, which works by finding specific keywords in the sql string.

ksqldb-tools/src/main/java/io/confluent/ksql/tools/migrations/commands/ApplyMigrationCommand.java
```diff
@@ -311,22 +313,80 @@ private void executeCommands(
...
+    ksqlClient.getVariables().forEach((k, v) -> ksqlClient.undefine(k));
+  }
+
   private void executeCommand(
+      final SqlCommand command,
+      final Client ksqlClient,
+      final Map<String, Object> properties,
+      final boolean defineUndefineOnly
+  ) throws ExecutionException, InterruptedException {
+    if (!defineUndefineOnly) {
```
vcrfxia, line 373: nit: can we move this into an `else` block for the if-else stack below to improve readability?
```
if (command instanceof SqlDefineVariableCommand) {
  ...
} else if (command instanceof SqlUndefineVariableCommand) {
  ...
} else {
  if (!defineUndefineOnly) {
    executeNonVariableCommands(command, ksqlClient, properties);
  }
}
```

ksqldb-tools/src/main/java/io/confluent/ksql/tools/migrations/util/CommandParser.java
vcrfxia: Ah, got it. Would it be equivalent if we removed the `toUpperCase()` and instead replaced the `.equals()` for token comparison with `.equalsIgnoreCase()` instead? It's mildly less efficient but I think it improves readability by not separating the logic for case-insensitive comparison across two different places.

ksqldb-tools/src/main/java/io/confluent/ksql/tools/migrations/commands/ApplyMigrationCommand.java
```diff
@@ -311,22 +313,80 @@ private void executeCommands(
...
+    executeCommands(
+        commands, ksqlClient, config, executionStart, migration, clock, previous, true);
+    executeCommands(
+        commands, ksqlClient, config, executionStart, migration, clock, previous, false);
+  }
+
+  /**
+   * If validateOnly is set to true, then this parses each of the commands but only executes
+   * DEFINE/UNDEFINE commands (variables are needed for parsing INSERT INTO... VALUES, SET/UNSET
+   * and DEFINE commands). If validateOnly is set to fales, then each command will execute after
```
vcrfxia, line 327:
```suggestion
   * and DEFINE commands). If validateOnly is set to false, then each command will execute after
```

vcrfxia (approved):
LGTM! Thanks @jzaralim .

ksqldb-tools/src/main/java/io/confluent/ksql/tools/migrations/util/CommandParser.java
jzaralim: There are some places where it uses `contains` instead of `equals`, for example `tokens.contains(VALUES)`, so removing the `toUpperCase` would actually make it more confusing. I've added some comments to clarify what `tokens` is supposed to be.

vcrfxia: Sounds good. Thanks!

PR #7368, opened 2021-04-11
feat: Materialize Table-Table join results with new physical plans (Don't review)

cprasad1:
### Description 

### Testing done 

### Reviewer checklist


agavra:
Hey @cprasad1 - it's usually better to keep the old PR instead of opening a new one. That way when the commit is merged if we look at the git blame we can see the historical discussion. If people don't want to see the historical plans on the old PR they can just filter out that commit.

guozhangwang:
Could we remove the `ksql` team from the reviewers as well for such PRs so people won't be emailed?

PR #7371, opened 2021-04-12
chore: drop ignoreUnchecked for deprecated members

vvcephei:
Follow up to https://github.com/confluentinc/ksql/pull/7363 .

Once we feel all builds will see the upstream changes, we
no longer need to use ignoreUnchecked for the dropped close methods.


guozhangwang (approved):
Thanks!

PR #5769, opened 2020-07-06
fix: ensure null values cast to varchar/string remain null

stevenpyzhang:
### Description 
Fixes https://github.com/confluentinc/ksql/issues/4892
String.valueOf() will return the string "null" when a null value is passed in. Using Object.toString(obj, defaultValue) allows the null value to be returned when a null is passed in https://docs.oracle.com/javase/8/docs/api/java/util/Objects.html#toString-java.lang.Object-java.lang.String-

### Testing done 
Tested the scenario described in the ticket
Added new QTT case

### Reviewer checklist
- [ ] Ensure docs are updated if necessary. (eg. if a user visible feature is being added or changed).
- [ ] Ensure relevant issues are linked (description should include text like "Fixes #<issue number>")

PR #7373, opened 2021-04-12
fix: preserve the rest of a struct when one field has a processing error

### LINKED ISSUE #5980: When processing error occurs on nested field in a STRUCT, only set that field to null instead of entire STRUCT
Currently, when a processing error occurs on a nested field in a `STRUCT`, the entire encapsulating `STRUCT` is set to `null` in the output stream. Using an example where a date conversion fails for one of the fields in a struct type:

```sql
create stream s (val string) with (kafka_topic='s', value_format='json', partitions=1);
insert into s (val) values ('2020-08-11');
insert into s (val) values ('bad-date');

select struct(field1:=val, field2:=DATETOSTRING(STRINGTODATE(val, 'yyyy-MM-dd'), 'yyyyMMdd')) as body from s emit changes;
+--------------------------------------------------+
|BODY                                              |
+--------------------------------------------------+
|{FIELD1=2020-08-11, FIELD2=20200811}              |
|null                                              |
```

Only set the failing field(s) in the struct to `null`, and preserve other valid field(s). Using the example, above, instead of `null` for the entire `BODY` struct, return `null` for `field2`, and preserve `field1`:

```
{FIELD1=2020-08-11, FIELD2=null} 
```

jzaralim:
### Description 
Fixes #5980. This also applies to items in arrays and keys/values in maps.

### Testing done 
Added QTT tests

### Reviewer checklist


stevenpyzhang:
I think this should be considered a breaking change. Users would expect to lose the entire struct when a field has an error to null, but now it's being kept.

If you ran some CSAS/CTAS on a server without this change, process some records that cause a field processing error, stopped server, "upgraded" server to one with this change, and then processed the same records that ran into field errors, would the query still output null, or would we get this new behavior?

@jzaralim could you check what the behavior is here?

jzaralim:
@stevenpyzhang Tested it out, the old records remained as null, and new records follow the new behavior. It doesn't change the old records, but it's still a change in behaviour so I'm not sure if this is a breaking change or not.

```
|mmmmmmm              |null                                                                     |
|new record           |{NUM=44, AA=null}                                                        |
```

stevenpyzhang:
Sorry, breaking change isn't the right term here, I meant this change is backwards incompatible, and having query behavior change is something we should try to avoid. Ideally we keep the old behavior for now and trigger the new behavior with a config, see https://github.com/confluentinc/ksql/pull/5769 for an example of this.

stevenpyzhang (approved):
lgtm, can you verify one more time in the example above that queries issued on servers before this fix keep the old behavior of dropping the entire record even after upgrading to a server with the fix?

jzaralim:
Sure, here's an example:

```
-- on ksqlDB 6.2
SET 'auto.offset.reset'='earliest';
CREATE STREAM S (BDAY STRING) WITH (KAFKA_TOPIC='S', KEY_FORMAT='KAFKA', PARTITIONS=2, REPLICAS=1, VALUE_FORMAT='JSON');
CREATE STREAM OLD AS SELECT  BDAY, STRUCT(NUM:=44, B:=PARSE_TIMESTAMP(BDAY, 'yyyy-MM-dd')) FROM S;
INSERT INTO S VALUES ('abc');

-- current branch
SET 'auto.offset.reset'='earliest';
CREATE STREAM NEW AS SELECT  BDAY, STRUCT(NUM:=44, B:=PARSE_TIMESTAMP(BDAY, 'yyyy-MM-dd')) FROM S S;
INSERT INTO S VALUES ('def');

SELECT * FROM OLD EMIT CHANGES;
+-------------------------------------------------------+----------------------------------------------+
|BDAY                                                   |KSQL_COL_0                                    |
+-------------------------------------------------------+----------------------------------------------+
|abc                                                    |null                                          |
|def                                                    |null                                          |

SELECT * FROM NEW EMIT CHANGES;
+-------------------------------------------------------+----------------------------------------------+
|BDAY                                                   |KSQL_COL_0                                    |
+-------------------------------------------------------+----------------------------------------------+
|abc                                                    |{NUM=44, B=null}                              |
|def                                                    |{NUM=44, B=null}                              |
```

PR #7376, opened 2021-04-13
fix: source key format compatibility with non primitive types (#7337)

abbccdda:
This patch tries to fix the wrong behavior we spotted in #7242, where the conversion to struct key format should work when the source stream key format is KafkaFormat.

In addition, we expanded the fix and addressed the potential conversion failure for DelimitedFormat case as well.

Reviewers: Victoria Xia <victoria.xia@confluent.io>, Almog Gavra <agavra@confluent.io>, Guozhang Wang <wangguoz@gmail.com>

### Description 

### Testing done 

### Reviewer checklist


vcrfxia (approved):
Thanks @abbccdda . This is a direct backport of https://github.com/confluentinc/ksql/pull/7337, right? Looks like it was a clean backport (no changes or conflict resolution needed). LGTM if so!

PR #7383, opened 2021-04-14
add basic pojo support to ksqldb-client-api

survivant:
@gAmUssA 

I wanted to post "POJO" into a stream, but the current API only support KsqlObject.  ksqlclient-api  will do a post to the Rest API, so at the end the object KsqlObject will be converted to json anyway.  It's easier to use when you can use your Java model without conversion.

It's only a basic example, but it works.

I used this implementation with the project : https://github.com/gAmUssA/reactor-ksqldb/issues/6 

I added this class

````
@Data
@Builder
public class Shipment {
    int shipmentId;
    String cheese;
    String shipmentTimestamp;

}
````

with this testcase

````
List<Row> list = reactorClient
        .insertInto(
            SHIPMENTS_TOPIC_NAME,
            new KsqlObject()
                .put("shipmentId", 42)
                .put("cheese", "smile")
                .put("shipmentTimestamp", Instant.now().toString())
        )
        .log()
        .retryWhen(
            Retry.from(f -> f.take(1)
                .delayUntil(signal -> reactorClient
                    .executeStatement(String.format(CREATE_STREAM_STATEMENT, SHIPMENTS_TOPIC_NAME)))))
        .log("beforeBlock")
        .then(reactorClient.executeQueryFromBeginning("SELECT * FROM cheese_shipments EMIT CHANGES LIMIT 1;"))
        .block();

      Assertions.assertThat(list)
            .isNotNull()
            .first()
            .extracting(it -> it.getInteger("SHIPMENTID"))
            .isEqualTo(42);

      list = reactorClient
            .insertInto(
                    SHIPMENTS_TOPIC_NAME,
                    new Shipment.ShipmentBuilder()
                            .shipmentId(99)
                            .cheese("sad smile")
                            .shipmentTimestamp(Instant.now().toString())
                            .build()
            )
            .log()
            .then(reactorClient.executeQueryFromBeginning("SELECT * FROM cheese_shipments EMIT CHANGES LIMIT 2;"))
            .block();

    Assertions.assertThat(list)
        .isNotNull()
        .last()
        .extracting(it -> it.getInteger("SHIPMENTID"))
        .isEqualTo(99);
````

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -29,7 +29,7 @@
     <parent>
         <groupId>io.confluent</groupId>
         <artifactId>common</artifactId>
-        <version>7.6.0-725</version>
+        <version>7.6.0-726</version>
     </parent>
 
     <groupId>io.confluent.ksql</groupId>
@@ -142,7 +142,7 @@
         <git-commit-id-plugin.version>4.9.10</git-commit-id-plugin.version>
         <apache.io.version>2.7</apache.io.version>
         <io.confluent.ksql.version>7.6.0-0</io.confluent.ksql.version>
-        <io.confluent.schema-registry.version>7.6.0-254</io.confluent.schema-registry.version>
+        <io.confluent.schema-registry.version>7.6.0-255</io.confluent.schema-registry.version>
         <netty-tcnative-version>2.0.61.Final</netty-tcnative-version>
         <!-- We normally get this from common, but Vertx is built against this -->
         <!-- Note: `netty` depends on `tcnative` and if we bump `netty`

### CODE REVIEW

**Comment by vcrfxia** [comment]
Hey @survivant , thanks for the proof of concept! The new interface you've proposed definitely seems useful. (Incidentally, there was a similar request for being able to deserialize rows/objects returned from the client API into custom POJOs just a couple weeks ago: https://github.com/confluentinc/ksql/issues/7319.) Would you like to flesh out your proof of concept into a full PR to be merged? I can assist along the way. We'll want to add the equivalent new interface for the `streamInserts()` method in order to maintain parity between the two client methods for inserting rows, add test coverage, and also slip an example or two into the docs to improve discoverability of the new feature.

**Comment by survivant** [comment]
@vcrfxia the problem is that I don't know how to handle the serialization/deserialization of Pojo in Kafka.  If I had known, I'll have done a PR.  I have others ideas related to pojo support.  Like that one : https://github.com/gAmUssA/reactor-ksqldb/issues/7  

I think the pojo support should covert few areas.

**Comment by vcrfxia** [comment]
> the problem is that I don't know how to handle the serialization/deserialization of Pojo in Kafka.

Sorry, can you clarify how this is relevant? Were you thinking this is a blocker in the context of inserting POJOs as new rows via the client?

**Comment by survivant** [comment]
ya.. it's true that it's not different than other serialization.. sorry.  I had another thing in mind.   Back to this issue, here a little implementation that I did .  The serialization is done here
: requestBody.appendString(JsonObject.mapFrom(pojo).toString()).appendString("\n");

````

  @Override
  public CompletableFuture<Void> insertInto(final String streamName, final Object pojo) {
    final CompletableFuture<Void> cf = new CompletableFuture<>();

    final Buffer requestBody = Buffer.buffer();
    final JsonObject params = new JsonObject().put("target", streamName);
    requestBody.appendBuffer(params.toBuffer()).appendString("\n");
    requestBody.appendString(JsonObject.mapFrom(pojo).toString()).appendString("\n");

    makePostRequest(
            INSERTS_ENDPOINT,
            requestBody,
            cf,
            response -> handleStreamedResponse(response, cf,
                    (ctx, rp, fut, req) -> new InsertIntoResponseHandler(ctx, rp, fut))
    );

    return cf;
  }
````

I'll do a PR, it will be easier to review.

**Comment by survivant** [comment]
@vcrfxia here a basic example for streams too.  so far I was able to post using insertStream.  Fill free to changed anything, so it could fill your architecture.

File: ksqldb-api-client/src/main/java/io/confluent/ksql/api/client/impl/StreamInsertsSubscriber.java [hunk 1]
```diff
@@ -46,10 +47,10 @@ protected void afterSubscribe(final Subscription subscription) {
   }
 
   @Override
-  protected void handleValue(final KsqlObject row) {
+  protected void handleValue(final Object row) {
     httpRequest.writeCustomFrame(
         0, 0,
-        Buffer.buffer().appendString(row.toJsonString()).appendString("\n")
+        Buffer.buffer().appendString(JsonObject.mapFrom(row).toString()).appendString("\n")
```
vcrfxia, line 53: Thanks @survivant . This makes sense to me, though I don't think it works to make this change directly in StreamInsertsSubscriber, since `JsonObject.mapFrom(row).toString()` produces a different result from `row.toJsonString()` in the case where `row` is a `KsqlObject`. The former results in
```
{"empty":false,"map":{"k":"v"}}
```
while the latter produces the desired
```
{"k":"v"}
```
It's too bad that Java won't allow us to support separate client methods for the two types of inserts (`KsqlObject` and generic POJO). I think the best we can do is have a check here to see if `row` is an instance of a `KsqlObject` and perform the conversion to a string differently if so. Not the prettiest but it should work.

Do you want to drive this PR to completion by adding test coverage and the relevant documentation updates?

survivant: I see your point.  I can try do to another pass to make it cleaner.  I'll try my best to give testcases and documentation, but I'm involve in other opensource projects, so I don't have a lot of free time.

PR #7385, opened 2021-04-14
chore: log any errors that happen for pull queries

agavra:
### Description 

We recently noticed that a ksqlDB user was receiving erroneous responses to pull query requests. However, nothing relating to these errors was visible in the logs. We should ensure that we log any error a user receives in response to a command they've issued (whether from the UI or CLI) so that we can proactively investigate any unusual error activity.

### Testing done 

Without this change, the following exception was returned but never present in the logs. I also made sure that errors that contain the error message get logged properly:

```
[2021-04-16 09:01:47,141] ERROR Failure to execute pull query, not logging the error message since it contains the query string, which may contain sensitive information. If you see this LOG message, please submit a GitHub ticket and we will scrub the statement text from the error at io.confluent.ksql.physical.pull.HARouting.handlePullQuery(HARouting.java:121) (io.confluent.ksql.engine.EngineExecutor:220)
```

### Reviewer checklist

### DIFF
--- a/ksqldb-engine/src/main/java/io/confluent/ksql/engine/EngineExecutor.java
+++ b/ksqldb-engine/src/main/java/io/confluent/ksql/engine/EngineExecutor.java
@@ -232,7 +232,7 @@ final class EngineExecutor {
             pullPlannerOptions.debugString(),
             loc);
       } else {
-        LOG.warn("Failure to execute pull query. {} {}",
+        LOG.error("Failure to execute pull query. {} {}",
             routingOptions.debugString(),
             pullPlannerOptions.debugString(),
             e);
--- a/ksqldb-engine/src/main/java/io/confluent/ksql/engine/EngineExecutor.java
+++ b/ksqldb-engine/src/main/java/io/confluent/ksql/engine/EngineExecutor.java
@@ -17,6 +17,8 @@ package io.confluent.ksql.engine;
 
 import static io.confluent.ksql.metastore.model.DataSource.DataSourceType;
 
+import com.google.common.base.Throwables;
+import com.google.common.collect.Iterables;
 import io.confluent.ksql.KsqlExecutionContext.ExecuteResult;
 import io.confluent.ksql.analyzer.ImmutableAnalysis;
 import io.confluent.ksql.analyzer.QueryAnalyzer;
@@ -67,10 +69,13 @@ import io.confluent.ksql.util.PlanSummary;
 import io.confluent.ksql.util.TransientQueryMetadata;
 import java.util.Arrays;
 import java.util.Collections;
+import java.util.Locale;
 import java.util.Objects;
 import java.util.Optional;
 import java.util.Set;
 import java.util.stream.Collectors;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 /**
  * Executor of {@code PreparedStatement} within a specific {@code EngineContext} and using a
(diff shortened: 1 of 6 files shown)


ksqldb-engine/src/main/java/io/confluent/ksql/engine/EngineExecutor.java
```diff
@@ -205,6 +212,33 @@ PullQueryResult executePullQuery(
...
+            .getStackTrace()[0];
+        LOG.error("Failure to execute pull query {} {}, not logging the error message since it "
+            + "contains the query string, which may contain sensitive information. If you "
+            + "see this LOG message, please submit a GitHub ticket and we will scrub "
+            + "the statement text from the error at {}",
+            routingOptions.debugString(),
+            pullPlannerOptions.debugString(),
+            loc);
+      } else {
+        LOG.warn("Failure to execute pull query. {} {}",
```
guozhangwang, line 235: Why we want to log as WARN v.s. ERROR in the else condition? I feel they are at the same severity level and the only reason for this condition is to be cautious of not printing the query stmt.

guozhangwang (commented):
LGTM! Just one minor question.

ksqldb-engine/src/main/java/io/confluent/ksql/physical/pull/HARouting.java
```diff
@@ -118,9 +118,8 @@ public void close() {
     if (anyPartitionsEmpty) {
       LOG.debug("Unable to execute pull query: {}. All nodes are dead or exceed max allowed lag.",
                 statement.getStatementText());
-      throw new MaterializationException(String.format(
-          "Unable to execute pull query \"%s\". All nodes are dead or exceed max allowed lag.",
-          statement.getStatementText()));
+      throw new MaterializationException("Unable to execute pull query. "
```
guozhangwang, line 121: Nice cleanup and consolidation on the outer caller's trace.

ksqldb-engine/src/main/java/io/confluent/ksql/engine/EngineExecutor.java
agavra: Most of these errors are no actionable and just require retries, so I assumed they should be logged as `WARN`. The `ERROR` in the if condition is to flag with higher priority that there's something in our code that needs to be fixed. I'm happy logging them both as ERROR or WARN though, not a big deal.

ksqldb-engine/src/main/java/io/confluent/ksql/physical/pull/HARouting.java
agavra: :) thanks

ksqldb-engine/src/main/java/io/confluent/ksql/engine/EngineExecutor.java
guozhangwang: Hmm, maybe I missed something important here: I thought the only rationale for differentiating them is to not disclose the query stmt from the error message or stack trace.

Are we assuming that, inside the callees deep-down, we have a standard pattern that if the error is due to our code, then we include include the query stmt in the error message, otherwise we would not??

agavra: > Hmm, maybe I missed something important here: I thought the only rationale for differentiating them is to not disclose the query stmt from the error message or stack trace.

That was the case - I will go ahead and just make them both `ERROR`.

guozhangwang (approved):
LGTM!

PR #7390, opened 2021-04-16
feat: emit an error reason before closing websocket

swist:
### Description 
Reasons for closing websocket are limited to 123 bytes.
We can, however, emit a final message before closing the
socket - we're using that here to send a non-truncated
version of the error to display in the UI

### Reviewer checklist
- [] Ensure docs are updated if necessary. (eg. if a user visible feature is being added or changed).


tolgadur (approved):
lgtm

PR #7405, opened 2021-04-20
test: test excluding streams threads for uncaught exception handler

wcarlson5:
(cherry picked from commit 6123c42c02d10c41d66f8aa20048763188d08ac3)

Streams no longer sets the java Thread uncaught exception handler and instead uses their own. The issue is once a thread has been replaced it throws the exception to finish killing the thread. Ksql however sets the java Thread uncaught exception handler globally, this calls System.exit(1), perviously the streams handler would override this so the stream threads would be excluded but now that is not the case.

We ignore the calls for the stream threads

### Testing done 

### Reviewer checklist


wcarlson5:
Integration tests to follow

stevenpyzhang (approved):
lgtm if jenkins passes

ksqldb-rest-app/pom.xml
```diff
@@ -177,6 +177,13 @@
             <scope>test</scope>
         </dependency>
 
+        <dependency>
+            <groupId>com.github.stefanbirkner</groupId>
+            <artifactId>system-rules</artifactId>
+            <version>1.19.0</version>
```
agavra, line 183: nit: all dependencies should be declared in the parent pom and versions should be declared as properties

ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/util/KsqlUncaughtExceptionHandler.java
```diff
@@ -32,6 +33,9 @@ public KsqlUncaughtExceptionHandler(final Runnable flusher) {
 
   @SuppressFBWarnings
   public void uncaughtException(final Thread t, final Throwable e) {
+    if (t instanceof StreamThread) {
+      return;
```
agavra, line 37: we might want to create a testing-only constructor here where we can pipe down an action for this if branch. That way we can set a countdown latch or something and make sure that this branch is getting hit instead of relying on sleeps, which both will increase our test run time by a minute and also give us worse coverage

ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/util/KsqlUncaughtExceptionHandler.java
```diff
@@ -15,23 +15,41 @@
...
+    this(flusher, "");
+  }
+
+  @VisibleForTesting
+  public KsqlUncaughtExceptionHandler(
+      final Runnable flusher,
+      final String streamsThreadMessage
+  ) {
     this.flusher = flusher;
+    this.streamsThreadMessage = streamsThreadMessage;
```
agavra, line 42: Thanks for addressing the comments! Is there any reason not to use a `CountDownLatch` here instead of a message that we want to print out? I just think that'll be a little cleaner (avoids the assert eventually call and reading to stdout)

ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/util/KsqlUncaughtExceptionHandler.java
stevenpyzhang: @agavra I updated the test so that we keep trying to assert that a specific error message that's passed in through the uncaught handler constructor is printed to stderr, which is only done in this logic branch.

ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/util/KsqlUncaughtExceptionHandler.java
wcarlson5: @stevenpyzhang Wrote this test, not trying to take credit for his work. I just did the 3 line fix. But my local build it messed up. It might be easier for Steven to just push his changes and open a new pr at this point...

agavra (approved):
Thanks @stevenpyzhang and @wcarlson5 :)

guozhangwang (approved):
LGTM.

ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/util/KsqlUncaughtExceptionHandler.java
stevenpyzhang: it's latched now

stevenpyzhang:
waiting for build to pass will then merge

stevenpyzhang:
FYI, the test takes a while to wrap up in the after part where it stops the server, it spams the logs with ```[2021-04-22 19:06:10,223] WARN [Consumer clientId=_confluent-ksql-default_query_CSAS_S_1-2aca7365-9217-43ba-a8d7-cfdf406ad081-StreamThread-5-consumer, groupId=_confluent-ksql-default_query_CSAS_S_1] Error while fetching metadata with correlation id 2533 : {PAGEVIEW_TOPIC=UNKNOWN_TOPIC_OR_PARTITION} (org.apache.kafka.clients.NetworkClient:1100)``` during this time

PR #7415, opened 2021-04-21
build: use proper artifact version during release builds

vcrfxia:
### Description 

Same as https://github.com/confluentinc/ksql/pull/7406 but more targeted. This PR fixes the bug with the maven and docker artifact versions, and leaves the deb, rpm, and archive versions as is. @andrewegel is it correct for the archive version to be the save as the deb and rpm versions, or should it match the maven and docker versions instead?

### Reviewer checklist

### DIFF
--- a/build-packages.sh
+++ b/build-packages.sh
@@ -64,10 +64,12 @@ done
 if [[ "${FULL_VERSION}" =~ ^[0-9]+.[0-9]+.[0-9]+-rc[0-9]+$ ]]; then
     RELEASE="0.${FULL_VERSION##*-rc}"
     VERSION="${FULL_VERSION%%-rc*}"
+    MAVEN_ARTIFACT_VERSION="${FULL_VERSION}"
 elif [[ "${FULL_VERSION}" =~ ^[0-9]+.[0-9]+.[0-9]+$ ]]; then
     RELEASE=1
     VERSION="${FULL_VERSION}"
     FULL_VERSION="${FULL_VERSION}-${RELEASE}"
+    MAVEN_ARTIFACT_VERSION="${VERSION}"
 else
     error "Unknown version format '${FULL_VERSION}' "
 fi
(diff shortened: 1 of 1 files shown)


lct45:
Since I reverted 7406 this morning should Maven Settings, RPM, and Archive have `VERSION` instead of `FULL_VERSION`?

vcrfxia:
I don't understand your question. My branch is on top of your revert from this morning: https://github.com/vcrfxia/ksql/commits/vxia-ksql-db-release-rpm-version

lct45:
Ah sorry, I thought we wanted to revert the revert (so put back the changes from yesterday) and then make your additional tweaks so it works for both full releases and candidates.

I think your changes in this PR correctly fix the issues my patch, 7406, but since I undid the patch this morning I think we need to put back in a few of the other changes

vcrfxia:
I thought the original version (what this branch is built on) was already working for candidates, and we only need changes to get it working for release builds as well? The changes in this PR are targeted such that only the behavior of release builds are affected.

If you have ideas on how to get this working and want to own the fix (and testing), that'd actually be great. I just wanted to open this as an alternative way to approach the bugfix since I think it's more contained than trying to use some combination of VERSION and FULL_VERSION everywhere as the previous PR did. Feel free to subsume this PR with another one.

andrewegel:
> @andrewegel is it correct for the archive version to be the save as the deb and rpm versions, or should it match the maven and docker versions instead?

Its up to you really - version values in archives don't really have any meaning as they're not part of a package manager. My suggestion would be to drop anything after the `-` in the version field. 

Getting a PR build working (pass in real params against a sha that builds correctly) is going to inspire more confidence your modifications work here.

vcrfxia:
Thanks. Updated the PR to drop the `-1` from the tgz version (for release builds).

andrewegel (approved):
Green PR build with parameters passed in.

PR #7417, opened 2021-04-22
docs(klip-46): date and time data type support

jzaralim:
### Description 
KLIP for adding support for date and time types.

### Reviewer checklist

### DIFF
--- a/design-proposals/klip-46-date-and-time-data-type-support.md
+++ b/design-proposals/klip-46-date-and-time-data-type-support.md
@@ -199,17 +199,6 @@ Casting from DATE to TIMESTAMP adds `00:00:00.000` as the time portion to the DA
 SELECT CAST(d AS TIMESTAMP) FROM s;
 > 2007-01-1T00:00:00
 ```
-#### TIME AS TIMESTAMP
-Casting from TIME to TIMESTAMP adds `1970-01-01` as the date portion to the TIME:
-```roomsql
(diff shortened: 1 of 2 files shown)


spena (approved):
Looks pretty good! Thanks @jzaralim 
+1

design-proposals/klip-46-date-and-time-data-type-support.md
```diff
@@ -0,0 +1,297 @@
...
+```
+#### DATE AS TIMESTAMP
+Casting from DATE to TIMESTAMP adds `00:00:00.000` as the time portion to the DATE:
+```roomsql
+-- d is the date, '2007-01-01'
+SELECT CAST(d AS TIMESTAMP) FROM s;
+> 2007-01-1T00:00:00
+```
+#### TIME AS TIMESTAMP
+Casting from TIME to TIMESTAMP adds `1970-01-01` as the date portion to the TIME:
```
derekjn, line 203: I don't think we even need to support this. It wouldn't add value IMO, and would just create surface area for surprising behavior and/or silent mistakes.

derekjn (approved):

design-proposals/klip-46-date-and-time-data-type-support.md
jzaralim: Yeah that makes sense. Removed.

Issue #7358, opened 2021-04-08
PUSH queries get blocked when the Vert.x event-loop handling the request is running another push query


spena:
**Describe the bug**
The issue is with the number of PUSH queries that can be executed by KSQL. This issue was found by a user who said some KSQL queries were stuck and not responding. 

The problem is when running several PUSH queries, sometimes some queries do not respond. Consequent queries may or may not respond. Sometimes, the user could run 4 queries in a row, the 5th gets stuck, but the 6th and 7th queries could  be run without issues. Killing one query may or may not affect the 5th query. Sometimes, you kill 1 query, and the 5th is still stuck, but you kill another query, and the 5th query starts running and streaming data. This behavior is not consistent.

I found out the issue is related to how Vert.x handles the PUSH queries requests. Vert.x has two different types of threads, event loop threads and worker threads.

Event loop threads are configured by the `ksql.verticle.instances` configuration (default: 2 * CPU cores). An event loop is a thread that handles a request on the KSQL endpoints. Event loops do not block a request. An event loop thread quickly handles the request and respond to the user. If the request needs to do some heavy work, such as PUSH queries which stream data until terminated, then the event loop puts these requests into a queue, so they are handled by the worker threads.

```
public static final String VERTICLE_INSTANCES = KSQL_CONFIG_PREFIX + "verticle.instances";
  public static final int DEFAULT_VERTICLE_INSTANCES =
      2 * Runtime.getRuntime().availableProcessors();
  public static final String VERTICLE_INSTANCES_DOC =
      "The number of server verticle instances to start per listener. Usually you want at least "
          + "many instances as there are cores you want to use, as each instance is single "
          + "threaded.";
```

Worker threads are configured by the `ksql.worker.pool.size` configuration (default: 100). Worker threads can handle code that may block. So the reason of having 100 works by default. Worker threads are not in sync with event loop threads. An event loop thread puts a request into a queue and returns a response to the client. A worker thread later picks a request from the queue and executes it for as long as it needs to (PUSH queries for instance).

```
public static final String WORKER_POOL_SIZE = KSQL_CONFIG_PREFIX + "worker.pool.size";
  public static final String WORKER_POOL_DOC =
      "Max number of worker threads for executing blocking code";
  public static final int DEFAULT_WORKER_POOL_SIZE = 100;
```

[...]

PR #7420, opened 2021-04-22
fix: long-running queries shouldn't block the main event loop

vvcephei:
### Description 
Currently, due to https://github.com/confluentinc/ksql/issues/7358, long-running queries block the main server event loop, even though they execute on the worker pool instead of the event loop pool.
This patch corrects this flaw.

### Testing done 
Added integration tests to reproduce the bug (https://github.com/confluentinc/ksql/issues/7358).
Note that the newer HTTP2 API did not have the bug, but the integration test for the older API does hang and time out without the fix in place.

### Reviewer checklist

### DIFF
--- a/ksqldb-api-client/src/test/java/io/confluent/ksql/api/client/integration/ClientIntegrationTest.java
+++ b/ksqldb-api-client/src/test/java/io/confluent/ksql/api/client/integration/ClientIntegrationTest.java
@@ -21,10 +21,13 @@ import static io.confluent.ksql.test.util.AssertEventually.assertThatEventually;
 import static io.confluent.ksql.util.KsqlConfig.KSQL_DEFAULT_KEY_FORMAT_CONFIG;
 import static io.confluent.ksql.util.KsqlConfig.KSQL_STREAMS_PREFIX;
 import static org.hamcrest.MatcherAssert.assertThat;
+import static org.hamcrest.Matchers.allOf;
 import static org.hamcrest.Matchers.contains;
 import static org.hamcrest.Matchers.containsInAnyOrder;
 import static org.hamcrest.Matchers.containsString;
 import static org.hamcrest.Matchers.equalTo;
+import static org.hamcrest.Matchers.hasItem;
+import static org.hamcrest.Matchers.hasProperty;
 import static org.hamcrest.Matchers.hasSize;
 import static org.hamcrest.Matchers.instanceOf;
 import static org.hamcrest.Matchers.is;
@@ -69,8 +72,11 @@ import io.confluent.ksql.engine.KsqlEngine;
 import io.confluent.ksql.integration.IntegrationTestHarness;
 import io.confluent.ksql.integration.Retry;
 import io.confluent.ksql.name.ColumnName;
+import io.confluent.ksql.rest.client.RestResponse;
+import io.confluent.ksql.rest.client.StreamPublisher;
 import io.confluent.ksql.rest.entity.ConnectorList;
 import io.confluent.ksql.rest.entity.KsqlEntity;
+import io.confluent.ksql.rest.entity.StreamedRow;
 import io.confluent.ksql.rest.integration.RestIntegrationTestUtil;
 import io.confluent.ksql.rest.server.ConnectExecutable;
 import io.confluent.ksql.rest.server.TestKsqlRestApp;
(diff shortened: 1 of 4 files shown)


ksqldb-api-client/src/test/java/io/confluent/ksql/api/client/integration/ClientIntegrationTest.java
```diff
@@ -296,6 +310,30 @@ public void tearDown() {
     REST_APP.getServiceContext().close();
   }
 
+  @Test
+  public void shouldStreamMultiplePushQueries() throws Exception {
```
vvcephei, line 314: This works with and without my patch, but it doesn't hurt to have a test for it.

ksqldb-api-client/src/test/java/io/confluent/ksql/api/client/integration/ClientIntegrationTest.java
```diff
@@ -958,29 +996,24 @@ public void shouldListTopics() throws Exception {
   }
 
   @Test
-  public void shouldListQueries() {
+  public void shouldListQueries() throws ExecutionException, InterruptedException {
```
vvcephei, line 999: This test was flake-prone, since it assumes that it could serialize itself after other tests (i.e., wait until all other tests' queries have stopped) and that other tests wouldn't start new queries during its execution.

There doesn't seem to be a reason for this, so I changed the test just to verify the thing it wanted to verify, regardless of what other queries are running.

ksqldb-api-client/src/test/java/io/confluent/ksql/api/client/integration/RestClientIntegrationTest.java
```diff
@@ -0,0 +1,219 @@
...
+
+import static io.confluent.ksql.util.KsqlConfig.KSQL_DEFAULT_KEY_FORMAT_CONFIG;
+import static io.confluent.ksql.util.KsqlConfig.KSQL_STREAMS_PREFIX;
+import static org.hamcrest.MatcherAssert.assertThat;
+import static org.hamcrest.Matchers.*;
+
+/**
+ * This integration test is displaced from the rest-client package
+ * to make use of utilities that are not available there.
+ */
```
vvcephei, line 73: This is pretty obnoxious.

The integration test needs to start a REST server, but in the rest-client test module, we don't have a dependency on the rest-server test module. And we can't add one because rest-server test module already depends on rest-client main module.

I can think of two solutions that are potentially better than putting it here: create a new "integration-test" module that depends on all the other modules, or relocate this integration test to the rest-server test module.

ksqldb-rest-app/src/main/java/io/confluent/ksql/api/server/OldApiUtils.java
```diff
@@ -132,30 +132,34 @@ private static void streamEndpointResponse(final Server server,
...
-      } finally {
-        try {
-          ros.close();
-        } catch (IOException ignore) {
-          // Ignore - it might already be closed
-        }
-      }
-    }, vcf);
+      },
+      false /*if this is true, worker execution blocks the main event loop*/,
```
vvcephei, line 160: A lot of lines changed because I fixed the indentation, but this is what actually changed here.

AlanConfluent (approved):
It seems from the documentation that one worker task can block another if they are invoked from the same context. Not sure if that's what you're describing in the description (though it certainly would cause the bug referenced).  Anyway, in our world, worker invocations are all independent.

spena:
Hey @vvcephei, the PR looks good. Did you look into the other places where `executeBlocking` is used? I found these files that do not have the `false` flag either. I don't know if those could block the event loop too.

```
JaasAuthProvider.authenticate()
KsqlAuthorizationProviderHandler.handle()
```

spena (approved):
+1. Just look at the other 2 files that call executeBlocking() to see if they need the false too.

vvcephei:
Thanks, @spena ! 

I also saw those, but I had enough trouble with these integration tests that I lost my appetite to do it all in one chunk. I think I'll file some follow-on tickets to try and produce blocking in those contexts and fix it there. My instinct says that we probably _never_ want to block the main event loop on the worker pool. Otherwise, why would we bother delegating to the worker pool? If you think it's ok to just change those other two usages without tests, I can certainly do that in this PR.

Thanks,
-John

spena:
@vvcephei I think it's ok to change it without tests. We should not block the event loop as you said.

vvcephei:
Thanks, @spena ! Since I have a green build right now, I'll go ahead and capitalize on it by merging this PR. I'll follow up right away with a separate PR to change those other two methods.

Thanks for the reviews, @spena and @AlanConfluent !

PR #7424, opened 2021-04-23
feat: Adds ScalablePushRegistry and peeking ability in a persistent query

AlanConfluent:
### Description 
This is the first step toward scalable push queries. `ScalablePushRegistry` is held by a `PersistentQueryMetadata` and peeks at the rows passing through to the end of the topology.

It also introduces the `ProcessingQueue` which is the object registered for a given scalable push query with the `ScalablePushRegistry`, and where rows are offered.

Next steps will be to introduce the the physical operator which calls `register` and `unregister` and pipe it through to the physical plan.

### Testing done 
Ran unit tests.

### Reviewer checklist


ksqldb-engine/src/main/java/io/confluent/ksql/physical/scalablepush/ScalablePushRegistry.java
```diff
@@ -0,0 +1,188 @@
...
+ * eventually passed on to scalable push queries.
+ */
+public class ScalablePushRegistry {
+
+  private static final Logger LOG = LoggerFactory.getLogger(ScalablePushRegistry.class);
+
+  private final KStream<?, GenericRow> stream;
+  private final LogicalSchema logicalSchema;
+  private final PushLocator pushLocator;
+  private final Set<ProcessingQueue> processingQueues = new ConcurrentHashSet<>();
```
agavra, line 57: it's a little odd for us to use a `Set` here when `PorcessingQueue` doesn't implement `hashCode` or `equals` since we'll never be able to check set membership 😉 that's my long way of saying, can this just be a list? (that also allows us to play around with guaranteeing iteration order)

ksqldb-engine/src/main/java/io/confluent/ksql/physical/scalablepush/ScalablePushRegistry.java
```diff
@@ -0,0 +1,188 @@
...
+  @VisibleForTesting
+  int numRegistered() {
+    return processingQueues.size();
+  }
+
+  // Registers a peek processor which makes a copy of every row and passes it on to any registered
+  // queues.
+  @SuppressWarnings("unchecked")
+  private void registerPeek(final boolean windowed) {
+    final ProcessorSupplier<Object, GenericRow> peek = new Peek<>((key, value, timestamp) -> {
```
agavra, line 98: it _might_ be a little cleaner to have `ScalablePushRegistry` just implement `ProcessorSupplier<K, V>` and then call `stream.process(aScalablePushRegistry)` - it seems like, to me, passing the entire `KStream` into the constructor breaks some abstraction barriers

ksqldb-engine/src/main/java/io/confluent/ksql/physical/scalablepush/ScalablePushRegistry.java
```diff
@@ -0,0 +1,188 @@
...
+      for (ProcessingQueue queue : processingQueues) {
+        try {
+          final TableRow row;
+          if (!windowed) {
+            final GenericKey keyCopy = GenericKey.fromList(((GenericKey) key).values());
+            final GenericRow valueCopy = GenericRow.fromList(value.values());
+            row = Row.of(logicalSchema, keyCopy, valueCopy, timestamp);
```
agavra, line 105: can we do this outside of the for loop or am I missing something?

ksqldb-engine/src/main/java/io/confluent/ksql/physical/scalablepush/locator/AllHostsLocator.java
```diff
@@ -0,0 +1,136 @@
...
+    }
+
+    return currentQueries.stream()
+        .map(QueryMetadata::getAllMetadata)
+        .filter(Objects::nonNull)
+        .flatMap(Collection::stream)
+        .filter(streamsMetadata -> streamsMetadata != StreamsMetadata.NOT_AVAILABLE)
+        .map(StreamsMetadata::hostInfo)
+        .map(hi -> new Node(isLocalhost(hi), buildLocation(hi)))
+        .collect(Collectors.toCollection(LinkedHashSet::new));
```
agavra, line 62: out of curiosity, why `LinkedHashSet`? do we care about ordering? (if so we should probably comment about why)

ksqldb-engine/src/main/java/io/confluent/ksql/physical/scalablepush/locator/AllHostsLocator.java
```diff
@@ -0,0 +1,136 @@
...
+  }
+
+  private boolean isLocalhost(final HostInfo hostInfo) {
+    if (hostInfo.port() != localhost.getPort()) {
+      return false;
+    }
+
+    return hostInfo.host().equalsIgnoreCase(localhost.getHost())
+        || hostInfo.host().equalsIgnoreCase("localhost");
```
agavra, line 71: nit: is there an easily available constant for `"localhost"` somewhere?

ksqldb-engine/src/main/java/io/confluent/ksql/query/QueryExecutor.java
```diff
@@ -209,6 +234,12 @@ PersistentQueryMetadata buildPersistentQuery(
 
     final RuntimeBuildContext runtimeBuildContext = buildContext(applicationId, queryId);
     final Object result = buildQueryImplementation(physicalPlan, runtimeBuildContext);
+    // Must be created before the topology is created since the ScalablePushRegistry adds a peek
+    // to the topology.
```
agavra, line 238: this goes back to my other point about it being weird to register the peek call within the constructor

agavra (commented):
Quick pass - didn't review tests yet. Overall LGTM, some data structure questions... I'm excited to have this work finally kick off! By the way, we might want to make the design doc public so that we have something to start pointing people to.

ksqldb-engine/src/main/java/io/confluent/ksql/query/QueryExecutor.java
AlanConfluent: I agree.  Made the suggested change.  I still added some comments here to make it clear what's going on, but I changed the method to `applyScalablePushProcessor` so it should be clearer.

AlanConfluent:
> Quick pass - didn't review tests yet. Overall LGTM, some data structure questions... I'm excited to have this work finally kick off! By the way, we might want to make the design doc public so that we have something to start pointing people to.

Good idea.  Let's chat on the best way to do that.

ksqldb-engine/src/main/java/io/confluent/ksql/physical/scalablepush/ScalablePushRegistry.java
agavra: thanks for the detailed explanation - totally agree we shouldn't change the original code, it might have some real performance implications elsewhere.

ksqldb-engine/src/main/java/io/confluent/ksql/physical/scalablepush/ScalablePushRegistry.java
agavra: > Basically, I just used a Set so that I could have a constant time unregister rather than searching through a list.

I missed that - if that's the case then a reference set is a-ok

ksqldb-engine/src/main/java/io/confluent/ksql/physical/scalablepush/ProcessingQueue.java
```diff
@@ -0,0 +1,77 @@
...
+    this.rowQueue = new LinkedList<>();
+  }
+
+  public synchronized boolean offer(final TableRow tableRow) {
+    if (rowQueue.size() < queueSizeLimit && !droppedRows) {
+      rowQueue.offer(tableRow);
+      newRowCallback.run();
+      return true;
+    }
+    droppedRows = true;
```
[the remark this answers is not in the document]
agavra: sounds good, I didn't have that context that the query will fail elsewhere

ksqldb-engine/src/main/java/io/confluent/ksql/physical/scalablepush/ProcessingQueue.java
```diff
@@ -0,0 +1,77 @@
...
+  public ProcessingQueue() {
+    this(BLOCKING_QUEUE_CAPACITY);
+  }
+
+  public ProcessingQueue(final int queueSizeLimit) {
+    this.queueSizeLimit = queueSizeLimit;
+    this.rowQueue = new LinkedList<>();
+  }
+
+  public synchronized boolean offer(final TableRow tableRow) {
```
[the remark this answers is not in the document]
agavra: I was weighing between a mutex and some concurrent data structure - it seems like the contention would be pretty high (per event?). Probably better to be safe now and optimize later, but I would be interested to know the overhead spent on synchronization when we push the benchmark to it's limits 

(c.f. https://lmax-exchange.github.io/disruptor/)

ksqldb-engine/src/main/java/io/confluent/ksql/physical/scalablepush/ScalablePushRegistry.java
```diff
@@ -0,0 +1,177 @@
...
+  public void close() {
+    for (ProcessingQueue queue : processingQueues) {
+      queue.close();
+    }
+    processingQueues.clear();
+    closed = true;
+  }
+
+  public void register(final ProcessingQueue processingQueue) {
+    if (closed) {
```
agavra, line 78: I think there's a race condition here - if someone calls `close` and another calls `register` then close can be in the middle of iterating when we add another one to the set. I can't tell whether this is meant to be concurrent code or not, but the fact the `closed` is volatile makes me think it is. It seems to me like we'd be safer off just synchronizing everything

ksqldb-engine/src/main/java/io/confluent/ksql/physical/scalablepush/locator/PushLocator.java
```diff
@@ -0,0 +1,41 @@
...
+
+import java.net.URI;
+import java.util.List;
+
+/**
+ * The interface used for locating hosts for a given scalable push query.
+ */
+public interface PushLocator {
+
+  // Returns the list of nodes which should be contacted to respond to this query.
```
agavra, line 26:
```suggestion
  /** Returns the list of nodes which should be contacted to respond to this query. */
```

ksqldb-engine/src/test/java/io/confluent/ksql/physical/scalablepush/ScalablePushRegistryTest.java
```diff
@@ -0,0 +1,189 @@
...
+  @Test
+  public void shouldRegisterAndGetQueueOffer_nonWindowed() {
+    // Given:
+    ScalablePushRegistry registry = new ScalablePushRegistry(locator, SCHEMA, false);
+    when(genericKey.values()).thenAnswer(a -> KEY);
+    when(genericRow.values()).thenAnswer(a -> VALUE);
+    when(processorContext.timestamp()).thenReturn(TIMESTAMP);
+
+    // When:
+//    verify(stream).process(processorSupplier.capture());
```
agavra, line 83: why are these commented out?

agavra (approved):
lgtm :)

ksqldb-engine/src/main/java/io/confluent/ksql/physical/scalablepush/ScalablePushRegistry.java
AlanConfluent: The code that calls this effectively shouldn't ever register anything after calling close, so it's a bit funny to reason about what it means to be consistent, but I agree that it's confusing this way.  I added synchronized to protect the use of `closed` and changing the size of `processingQueues`.

I still use a ConcurrentHashMap for `processingQueues` so that I don't need to get a lock when processing rows.  It gives a weakly consistent view of the map while iterating, which should be performant.

ksqldb-engine/src/main/java/io/confluent/ksql/physical/scalablepush/ProcessingQueue.java
```diff
@@ -0,0 +1,77 @@
...
+
+import io.confluent.ksql.execution.streams.materialization.TableRow;
+import java.util.LinkedList;
+
+/**
+ * A queue for storing pre-processed rows for a given scalable push query request. This queue
+ * starts dropping rows if they're past the capacity, and keeps track so it can be reported to the
+ * request.
+ */
+public class ProcessingQueue {
```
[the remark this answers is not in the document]
cprasad1: Why not use a doubly linked list implementation of a Deque instead of an ArrayDeque? We can avoid rows getting dropped when we hit capacity using that as it can grow 'infinitely'. We can achieve constant time poll and offer methods using a DLL implementation as well

cprasad1 (approved):
LGTM. Just had one question about the choice of a data structure

ksqldb-engine/src/main/java/io/confluent/ksql/physical/scalablepush/ProcessingQueue.java
```diff
@@ -0,0 +1,112 @@
...
+package io.confluent.ksql.physical.scalablepush;
+
+import io.confluent.ksql.execution.streams.materialization.TableRow;
+import io.confluent.ksql.query.QueryId;
+import java.util.ArrayDeque;
+import java.util.Deque;
+
+/**
+ * A queue for storing pre-processed rows for a given scalable push query request. This queue
+ * starts dropping rows if they're past the capacity, and keeps track so it can be reported to the
```
guozhangwang, line 25: Regarding the error message for dropped records: since for now we are relying on the peek function, which is more of a `push` model than a `pull` model (if we fetch from the topics), and hence we cannot control how fast we want to return the data --- on the other hand though, we can make sure that the push query is always at the tail of the stream, though it could be lossy --- that means with large number of push queries, dropping rows may be a norm, and currently we would only report an error messaged as "dropped rows". This could be confusing and worrisome to users.

I'm wondering if we should consider, from easy to hard: 1) report how many records were dropped, 2) report the dropped records' corresponding offset on the source topic, 3) fail the query immediately if we ever drop.

ksqldb-engine/src/main/java/io/confluent/ksql/physical/scalablepush/ProcessingQueue.java
```diff
@@ -0,0 +1,112 @@
...
+
+  public synchronized boolean isClosed() {
+    return closed;
+  }
+
+  /**
+   * Sets a callback which is invoked every time a new row has been enqueued.
+   * @param newRowCallback The callback to invoke
+   */
+  public synchronized void setNewRowCallback(final Runnable newRowCallback) {
```
guozhangwang, line 98: From the other PR, it seems the `setNewRowCallback` call is only for triggering the poll first the first time, and after that we rely on the timer to `context.owner().setTimer(100, timerId -> context.runOnContext(v -> maybeNext(publisher)));` no matter if there's new row or not, is that intentional by design? If yes I'm wondering if we can just always rely on the `setNewRowCallback` for all triggering events: i.e. 1) whenever the `set` is called, we check immediately if there's any data available, and if yes executes the callback, and 2) whenever the callback is triggered, we reset it to null so that the callback would be triggered just once after it's set, 3) from caller `PushPhysicalPlan`, we call `setNewRowCallback` with the timer in `makeNext`.

ksqldb-engine/src/main/java/io/confluent/ksql/util/PersistentQueryMetadata.java
```diff
@@ -215,5 +220,13 @@ public synchronized void restart() {
    */
   public synchronized void stop() {
     doClose(false);
+    scalablePushRegistry.ifPresent(ScalablePushRegistry::close);
+  }
+
+  /**
+   * Gets the scalable push query registry, if one was created for this persistent query.
+   */
+  public Optional<ScalablePushRegistry> getScalablePushRegistry() {
```
guozhangwang, line 229: This is more or less for the other PR together: since I did not see the caller to construct this object I guess it is in other PRs, my question is though how do we find which `PersistentQueryMetadata` to pass in when we build the builder? That is because, it seems like the trace are:

`PushPhysicalPlanBuilder#persistentQueryMetadata -> PeekStreamOperator#scalablePushRegistry -> scalablePushRegistry#register`

And we would have one registry for each source stream/table only, where all registered push queries would be given the record from the `peek` call. This means we must be able to get the corresponding `persistentQueryMetadata` to pass to the `PushPhysicalPlanBuilder`, but at that time how we would know which source stream/table it is for?

ksqldb-engine/src/main/java/io/confluent/ksql/physical/scalablepush/ProcessingQueue.java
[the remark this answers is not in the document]
AlanConfluent: I think we want to have a capacity so that we don't store unbounded data for a slow client, so being able to grow isn't a feature here.

Often allocating/deallocating in larger chunks can lead to less memory fragmentation.  This is less of an issue in Java due to how most GC implementations work, but I took Almog's suggestion.  We'd have to profile it to really know what was going on.

ksqldb-engine/src/main/java/io/confluent/ksql/physical/scalablepush/ProcessingQueue.java
AlanConfluent: Yeah, I think you're right that technically the timer isn't required if we're guaranteed that this will always call back.

I don't want to rely solely on the timer because it will mean that there will be a certain waiting latency.  We can try to ensure that we schedule just one call to mabeNext after a given call to the callback here.  Let's take that discussion to the followup PR.

ksqldb-engine/src/main/java/io/confluent/ksql/util/PersistentQueryMetadata.java
AlanConfluent: It's actually in this PR.  There's a factory method called `ScalablePushRegistry.create`.

If you look in `QueryExecutor.applyScalablePushProcessor`, you can see that when the `PersistentQueryMetadata` is being constructed, that's when this is created, so we know exactly how the two match up.

ksqldb-engine/src/main/java/io/confluent/ksql/physical/scalablepush/ProcessingQueue.java
AlanConfluent: Well, if they push the server to the limit, it's possible that the queues couldn't be drained fast enough to keep up, though hopefully in normal operation, that's not the case unless the requester is slow.

I think that 3) is definitely the route we want to go.  I might not have been clear, but I meant to send an error message and fail the query. For the near term, if you're too slow in reading rows, your request will be dropped.  Doing 2) would be easy though I'm not sure how useful it would be.  1) would be good in a situation where we allowed requests to catch up starting at offset X, but this is much more advanced for the moment.

ksqldb-engine/src/main/java/io/confluent/ksql/physical/scalablepush/ProcessingQueue.java
guozhangwang: Sounds good! Thanks.

guozhangwang (approved):

ksqldb-engine/src/main/java/io/confluent/ksql/query/QueryExecutor.java
```diff
@@ -188,14 +190,39 @@ TransientQueryMetadata buildTransientQuery(
...
+    }
+    final KStream<?, GenericRow> stream;
+    if (result instanceof KTableHolder) {
+      stream = ((KTableHolder<?>) result).getTable().toStream();
+    } else {
+      stream = ((KStreamHolder<?>) result).getStream();
+    }
+    final Optional<ScalablePushRegistry> registry = ScalablePushRegistry.create(schema,
+        allPersistentQueries, windowed, streamsProperties);
+    registry.ifPresent(r -> stream.process(registry.get()));
```
guozhangwang, line 213: nit: could this just be `registry.ifPresent(r -> stream.process(r));`?

ksqldb-engine/src/main/java/io/confluent/ksql/util/PersistentQueryMetadata.java
guozhangwang: Okay, I guess the confusion from my side is that I did not see where the constructor of `PushPhysicalPlanBuilder` is called in either PR, and hence hard to guess which `persistentQueryMetadata` parameter would be passed in. If you have that constructor called somewhere could you point me to it?

PR #7425, opened 2021-04-23
feat: made class to compute essential meta-data

tolgadur:
### Description 
We want to be able to filter the anonymized queries according to things such as physical cluster, organization and time created. For this we need to get/generate the relevant meta-data. The organization and physical cluster id is picked up from the control plane via a metric set [here](https://github.com/confluentinc/cc-spec-ksql/pull/233).

### Testing done 
Unit tests to make sure metadata gets computed as expected.

### References
jira: [KCI-342](https://confluentinc.atlassian.net/browse/KCI-342)
cc-structs-ksql update: [PR-223](https://github.com/confluentinc/cc-spec-ksql/pull/233)
cc-scheduler-service update: [PR-2240](https://github.com/confluentinc/cc-scheduler-service/pull/2240)
protobuff update: [PR-1507](https://github.com/confluentinc/cc-structs/pull/1507)

### Reviewer checklist

### DIFF
--- a/ksqldb-rest-app/src/main/java/io/confluent/ksql/api/auth/JaasAuthProvider.java
+++ b/ksqldb-rest-app/src/main/java/io/confluent/ksql/api/auth/JaasAuthProvider.java
@@ -96,6 +96,7 @@ public class JaasAuthProvider implements AuthProvider {
 
     server.getWorkerExecutor().executeBlocking(
         p -> getUser(contextName, username, password, allowedRoles, p),
+        false,
         resultHandler
     );
   }
--- a/ksqldb-rest-app/src/main/java/io/confluent/ksql/api/auth/KsqlAuthorizationProviderHandler.java
+++ b/ksqldb-rest-app/src/main/java/io/confluent/ksql/api/auth/KsqlAuthorizationProviderHandler.java
@@ -57,6 +57,7 @@ public class KsqlAuthorizationProviderHandler implements Handler<RoutingContext>
 
     workerExecutor.<Void>executeBlocking(
         promise -> authorize(promise, routingContext),
+        false,
         ar -> handleAuthorizeResult(ar, routingContext));
   }
 
--- a/ksqldb-rest-app/src/test/java/io/confluent/ksql/api/auth/JaasAuthProviderTest.java
+++ b/ksqldb-rest-app/src/test/java/io/confluent/ksql/api/auth/JaasAuthProviderTest.java
@@ -19,6 +19,7 @@ import static org.hamcrest.MatcherAssert.assertThat;
 import static org.hamcrest.Matchers.instanceOf;
 import static org.hamcrest.Matchers.is;
 import static org.mockito.ArgumentMatchers.any;
+import static org.mockito.ArgumentMatchers.eq;
 import static org.mockito.Mockito.doAnswer;
 import static org.mockito.Mockito.mock;
 import static org.mockito.Mockito.verify;
@@ -253,11 +254,11 @@ public class JaasAuthProviderTest {
     when(server.getWorkerExecutor()).thenReturn(worker);
     doAnswer(invocation -> {
       final Handler<Promise<User>> blockingCodeHandler = invocation.getArgument(0);
-      final Handler<AsyncResult<User>> resultHandler = invocation.getArgument(1);
+      final Handler<AsyncResult<User>> resultHandler = invocation.getArgument(2);
       final Promise<User> promise = Promise.promise();
       promise.future().onComplete(resultHandler);
       blockingCodeHandler.handle(promise);
       return null;
-    }).when(worker).executeBlocking(any(), any());
+    }).when(worker).executeBlocking(any(), eq(false), any());
   }
 }


ksqldb-parser/src/main/java/io/confluent/ksql/util/QueryMetaDataCollector.java
```diff
@@ -0,0 +1,107 @@
...
+import java.util.HashMap;
+import java.util.Hashtable;
+import java.util.Map;
+import java.util.Objects;
+import java.util.UUID;
+import org.joda.time.DateTime;
+
+public final class QueryMetaDataCollector {
+  private static final Hashtable<QueryIdKey, String> queryIdTable = new Hashtable<>();
+  private static final Hashtable<String, String> structurallySimilarTable = new Hashtable<>();
```
swist, line 27: Don't need these two

ksqldb-parser/src/main/java/io/confluent/ksql/util/QueryMetaDataCollector.java
```diff
@@ -0,0 +1,107 @@
...
+
+    // get metadata from control plane
+    final Map<String, String> metaData = new HashMap<>();
+    final String clusterNamespace = config
+        .getString(KsqlConfig.KSQL_CCLOUD_QUERYANONYMIZER_CLUSTER_NAMESPACE);
+
+    // check if metadata contains org id and cluster id
+    if (clusterNamespace.length() == 0) {
+      throw new Exception("Organization Id or Cluster Id could not be found");
+    }
```
swist, line 45: Don't throw, it will just means that all queries are treated as if they come from the same place (which makes sense for something like C3 deployment)

swist: We should think about getting C3 to understand these - you could default this config to be the KsqlConfig.KSQL_CLUSTER_NAME iirc

ksqldb-parser/src/main/java/io/confluent/ksql/util/QueryMetaDataCollector.java
```diff
@@ -0,0 +1,107 @@
...
+import java.util.UUID;
+import org.joda.time.DateTime;
+
+public final class QueryMetaDataCollector {
+  private static final Hashtable<QueryIdKey, String> queryIdTable = new Hashtable<>();
+  private static final Hashtable<String, String> structurallySimilarTable = new Hashtable<>();
+
+  private QueryMetaDataCollector() {}
+
+  public static Map<String, String> getMetaData(
```
swist, line 31: This looks like something that builds metadata rather than `gets` it - you're not retrieving a record, it's a constructor. Naming is important, especially so in OOP - you're giving a constructor pattern here so name it constructor-like

swist, line 31: Also, make it ImmutableMap

ksqldb-parser/src/main/java/io/confluent/ksql/util/QueryMetaDataCollector.java
```diff
@@ -0,0 +1,107 @@
...
+    metaData.put("cluster_namespace", clusterNamespace);
+    metaData.put("id", getQueryId(nonAnonQuery, clusterNamespace));
+    metaData.put("structurally_similar_id", getStructurallySimilarId(anonQuery));
+    metaData.put("time_of_creation", DateTime.now().toString());
+
+    return metaData;
+  }
+
+  private static String getStructurallySimilarId(final String query) {
+    if (!structurallySimilarTable.containsKey(query)) {
```
swist, line 56: This could be just ```
return UUID.uuidV5(clusterNamespace, query)
```

https://en.wikipedia.org/wiki/Universally_unique_identifier#Versions_3_and_5_(namespace_name-based)

This might require you to actually turn the UUID to one of the compatible formats, but that just means changing it in a few places

File: ksqldb-parser/src/main/java/io/confluent/ksql/util/QueryMetaDataCollector.java [hunk 5]
```diff
@@ -0,0 +1,107 @@
...
+      structurallySimilarTable.put(query, UUID.randomUUID().toString());
+    }
+    return structurallySimilarTable.get(query);
+  }
+
+  private static String getQueryId(final String query, final String clusterNamespace) {
```
swist, line 62: Then you don't need this

ksqldb-parser/src/main/java/io/confluent/ksql/util/QueryMetaDataCollector.java
```diff
@@ -0,0 +1,107 @@
...
+      queryIdTable.put(queryIdKey, UUID.randomUUID().toString());
+    }
+    return queryIdTable.get(queryIdKey);
+  }
+
+  private static String getGenericQueryForm(final String query) {
+    return query.replaceAll("[\\n\\t ]", "");
+  }
+
+  private static final class QueryIdKey {
```
swist, line 76: And the UUID5 basically replaces this guy

swist (changes requested):
As discussed on slack, you want a determinisic, namespace UUID5 instead - it makes it much easier to reason about and it reduced the amount of code and testing you need to do

swist:
Another thing to note is that it's probably not in the correct classpath - I wouldn't say that this is related to the parser. Looks more like a ksqldb-rest-app concern

PR #7427, opened 2021-04-23
fix: worker-poll tasks shouldn't block main loop

vvcephei:
### Description 
Fixes a call to execute a task on the worker pool that still may block main event-loop
execution. See https://github.com/confluentinc/ksql/issues/7358 for more information.

### Testing done 
Just ran the tests.

### Reviewer checklist


guozhangwang (approved):
LGTM! And I also feel that we can just change it without adding more clunking tests.

vvcephei:
Thanks, @guozhangwang ! It looks like I need a small tweak to the matcher in the JaasProviderTest. I'll fix it and then merge once there's a green build.

PR #7446, opened 2021-04-28
chore: add MDC tag for queryId

wcarlson5:
add a query Id field to the logs

### Reviewer checklist


guozhangwang (approved):

ksqldb-engine/src/main/java/io/confluent/ksql/util/QueryMetadata.java
```diff
@@ -165,6 +166,7 @@ public long read() {
 
   public void initialize() {
     // initialize the first KafkaStreams
+    MDC.put("queryId", queryId.toString());
```
rodesai, line 169: I think the MDC is thread-local - so we'd need this to be done from the streams thread itself.

wcarlson5: https://logging.apache.org/log4j/1.2/apidocs/org/apache/log4j/MDC.html

It is but `The MDC is managed on a per thread basis. A child thread automatically inherits a copy of the mapped diagnostic context of its parent.`

PR #7452, opened 2021-04-29
feat: Allow ksqlDB to detect FK-join table-table join condition

mjsax:
### Description 
First PR to add FK-join support. This PR only aims to detect the FK-join condition, and throws an informative error message for now. This PR also includes error message improvements for existing joins.

I am not happy with everything is this PR yet, but wanted to open it to get early feedback. In particular, we cannot handle the case when an expression is use in the join condition yet. -- I am also not sure if there might be some existing helper method that I could re-use to simplify the code.

### Testing done 
Added new QueryTranslationTest cases.

### Reviewer checklist


ksqldb-engine/src/main/java/io/confluent/ksql/planner/LogicalPlanner.java
```diff
@@ -575,7 +579,107 @@ private JoinNode buildJoin(final Join root, final String prefix, final boolean i
...
+        // stream-table join detected
+
+        if (root.getInfo().getType().equals(JoinType.OUTER)) {
+          throw new KsqlException("Invalid join type:"
+              + " full-outer join not supported for stream-table joins.");
+        }
+
+        // added outer check to make `multi-joins.json` pass
+        // the test uses expressions -> need to follow up on this issue
+        if (root.getInfo().getRightJoinExpression() instanceof ColumnReferenceExp) {
```
mjsax, line 595: Need to fix this, to be able to process `BinaryArithmeticExpressions` and maybe others?

ksqldb-engine/src/main/java/io/confluent/ksql/planner/LogicalPlanner.java
```diff
@@ -575,7 +579,107 @@ private JoinNode buildJoin(final Join root, final String prefix, final boolean i
...
+      }
+    }
+  }
+
+  private boolean joinOnNonKeyAttribute(final ColumnReferenceExp joinExpression,
+                                        final PlanNode node) {
+    final String joinAttributeName = joinExpression.getColumnName().text();
+    final List<DataSourceNode> dataSourceNodes = node.getSourceNodes().collect(Collectors.toList());
+    final String singleAttributeKeyName;
+    if (dataSourceNodes.size() == 1) {
```
mjsax, line 640: This is the binary-join case, for which `PlanNode` access a stream/table directly (ie, no nested join in PlanNode)

ksqldb-engine/src/main/java/io/confluent/ksql/planner/plan/JoinNode.java
```diff
@@ -296,8 +296,9 @@ private static SourceName getSourceName(final PlanNode node) {
         final DataSourceType rightType) {
 
       return joinerMap.getOrDefault(new Pair<>(leftType, rightType), () -> {
-        throw new KsqlException("Join between invalid operands requested: left type: "
-            + leftType + ", right type: " + rightType);
+        throw new KsqlException(
```
mjsax, line 299: We have a check in the new `verifyJoin` method and thus should never hit this line any longer, ie, the user should never see this error message again.

ksqldb-functional-tests/src/test/resources/query-validation-tests/joins.json
```diff
@@ -1232,18 +1232,6 @@
         ]
       }
     },
-    {
-      "name": "to table using something other than key column",
```
mjsax, line 1236: This is covered in the newly added `stream-table-join.json` now

ksqldb-functional-tests/src/test/resources/query-validation-tests/joins.json
```diff
@@ -1703,18 +1691,6 @@
         ]
       }
     },
-    {
-      "name": "on non-key table column",
```
mjsax, line 1707: As above (note that this test seems to be redundant to the above one anyway?)

ksqldb-functional-tests/src/test/resources/query-validation-tests/joins.json
```diff
@@ -3096,7 +3072,7 @@
       "statements": [
         "CREATE TABLE T1 (ID INT PRIMARY KEY, VAL STRING) WITH (kafka_topic='t1', key_format='AVRO', value_format='JSON');",
         "CREATE TABLE T2 (ID BIGINT PRIMARY KEY, FOO INT) WITH (kafka_topic='t2', key_format='AVRO', value_format='JSON');",
-        "CREATE STREAM OUTPUT AS SELECT T2.ID, T1.VAL FROM T1 JOIN T2 ON T1.ID = T2.ID;"
+        "CREATE TABLE OUTPUT AS SELECT T2.ID, T1.VAL FROM T1 JOIN T2 ON T1.ID = T2.ID;"
```
mjsax, line 3075: Side fix (was not an issue, as we fail before we verify the query output type)

ksqldb-functional-tests/src/test/resources/query-validation-tests/joins.json
```diff
@@ -3234,18 +3210,6 @@
         "type": "io.confluent.ksql.util.KsqlStatementException",
         "message": "Implicit repartitioning of windowed sources is not supported. See https://github.com/confluentinc/ksql/issues/4385."
       }
-    },
-    {
-      "name": "stream-table key-to-key with necessary repartition - SR-enabled key format",
```
mjsax, line 3239: This test also seems redundant to the two from above?

ksqldb-functional-tests/src/test/resources/query-validation-tests/multi-joins.json
```diff
@@ -222,19 +222,6 @@
         "message": "The query used to build `OUTPUT` must include the join expression ROWKEY in its projection."
       }
     },
-    {
-      "name": "full-full missing join expression in projection",
```
mjsax, line 226: This test was invalid to begin with, because stream-table full outer joins are not supported.

ksqldb-engine/src/main/java/io/confluent/ksql/planner/LogicalPlanner.java
vcrfxia: `joinOnNonKeyAttribute()` should always return true if the expression is not a `ColumnReferenceExp`, right? Why don't we just pass the entire expression in and let `joinOnNonKeyAttribute()` perform the check?

ksqldb-engine/src/main/java/io/confluent/ksql/planner/LogicalPlanner.java
```diff
@@ -575,7 +579,107 @@ private JoinNode buildJoin(final Join root, final String prefix, final boolean i
...
+      singleAttributeKeyName = dataSourceNodes.get(0).getSchema().key().get(0).name().text();
```
vcrfxia, line 641: Why are we assuming there's a single key? Shouldn't we check whether there are multiple, and return true if so?

ksqldb-engine/src/main/java/io/confluent/ksql/planner/LogicalPlanner.java
vcrfxia: Do we know that we've already performed the self-join check by this point? If not, is it possible that we are looking at a PlanNode representing a join, rather than a single-data source? A more reliable way to perform this check would be to check whether the PlanNode is a JoinNode or a DataSourceNode.

ksqldb-engine/src/main/java/io/confluent/ksql/planner/LogicalPlanner.java
```diff
@@ -575,7 +579,107 @@ private JoinNode buildJoin(final Join root, final String prefix, final boolean i
...
+    } else {
+      final List<DataSourceNode> qualifiedNodes;
+
+      if (joinExpression.maybeQualifier().isPresent()) {
```
vcrfxia, line 645: Why are we manipulating qualifiers here? All we want is to know whether we're joining on the key column of the node, right? Why don't we check that more directly, similar to https://github.com/confluentinc/ksql/blob/5bb1d2acb16d423966693ce4eae83f76fb9f1687/ksqldb-engine/src/main/java/io/confluent/ksql/planner/plan/JoinNode.java#L229

ksqldb-functional-tests/src/test/resources/query-validation-tests/joins.json
vcrfxia: This test is not redundant from the other two. We have special logic that permits repartitioning tables for joins on sources with SR-enabled key formats (in order to ensure copartitioning in light of potentially different schema IDs). This test ensures that this feature does not allow changing the primary key column.

vcrfxia (commented):
Thanks @mjsax . Initial questions inline.

ksqldb-engine/src/main/java/io/confluent/ksql/planner/LogicalPlanner.java
mjsax: We don't support joins on structured keys yet, and if one tries to do this, we should not reach this point in the code but fail earlier.

ksqldb-engine/src/main/java/io/confluent/ksql/planner/LogicalPlanner.java
mjsax: My understanding was different. If we do `t1 JOIN t2 ON t1.key = t2.key + 1` it's still a PK-join not a FK-join right?However, ``1 JOIN t2 ON t1.a = t2.key + 1` would be a FK-joins. For both cases, it seems we would repartition `t2` onto new key `key + 1`?

I am not sure what expression we support atm and what parsing on the expression we would need to do to cover all cases? I _would_ assume that the expression can only contain a single "field/columnName"? So maybe it's sufficient to parse the full expression-tree and extract the single field?

ksqldb-engine/src/main/java/io/confluent/ksql/planner/LogicalPlanner.java
mjsax: The self-join check happens before IIRC -- we have test in place about it that still pass so it seems we don't break anything at this point. Can try out to check the node type as suggested.

ksqldb-engine/src/main/java/io/confluent/ksql/planner/LogicalPlanner.java
mjsax: We don't manipulating qualifiers: if a qualifier is given, we limit the check for this single stream/table instead of the need to check all input schemas to find the attribute.

ksqldb-functional-tests/src/test/resources/query-validation-tests/joins.json
mjsax: If I understand you comment correctly, the test does not test what it is supposed to test? With this PR, we would fail in `verifyJoin` already as we try to join on a non-key table attribute (`T.VAL`) that is always disallowed.

ksqldb-engine/src/main/java/io/confluent/ksql/planner/LogicalPlanner.java
guozhangwang: If it is a `ColumnReferenceExp` then I think it would not be arithmetic, like `t2.key + 1` atm (just reading through the sqlBase here).

ksqldb-engine/src/main/java/io/confluent/ksql/planner/LogicalPlanner.java
```diff
@@ -575,7 +579,107 @@ private JoinNode buildJoin(final Join root, final String prefix, final boolean i
...
+      singleAttributeKeyName = qualifiedNodes.get(0).getSchema().key().get(0).name().text();
+    }
+
+
+    return !joinAttributeName.equals(singleAttributeKeyName);
+  }
+
+
   private JoinKey buildJoinKey(final Join join) {
+    //
```
guozhangwang, line 682: nit: remove

ksqldb-engine/src/main/java/io/confluent/ksql/planner/LogicalPlanner.java
guozhangwang: Maybe adding a TODO in case we forget to update it when adding multi-key join support.

ksqldb-engine/src/main/java/io/confluent/ksql/planner/LogicalPlanner.java
mjsax: Yes. If it's a `ColumnReferenceExp` we do the check, but if not we skip it, but IMHO we need to do the check even if it's not a `ColumnReferenceExp`. (For now, I just put this guard in place to avoid that `multi-joins.json` pass, but it's just a workaround to make it pass -- we need to properly address those cases.)

ksqldb-engine/src/main/java/io/confluent/ksql/planner/plan/JoinNode.java
spena: If users will never hit it, whey do we throw it?

mjsax: We could remove it but would hit a NPE (if there is a bug) for this case -- thus, it seems better to just throw if we hit a bug instead of crashing with NPE?

It also seems safer to just keep the exception to void introducing a regression by accident (ie, if we miss a case and user could still hit this, even if they should not).

Not sure how this is handled in ksqlDB normally?

ksqldb-engine/src/main/java/io/confluent/ksql/planner/LogicalPlanner.java
mjsax: Did look into this, but we cannot do the `instanceof` check, because the `JoinNode` is "hidden" by the `ProJoinRepartitionNode` (and not exposed).

ksqldb-engine/src/main/java/io/confluent/ksql/planner/LogicalPlanner.java
```diff
@@ -575,6 +579,113 @@ private JoinNode buildJoin(final Join root, final String prefix, final boolean i
...
+            + " table-table joins require to join on the primary key of the right input table."
+        );
+      }
+
+      if (joinOnNonKeyAttribute(root.getInfo().getLeftJoinExpression(), left)) {
+        // after we lift the n-way join restriction, we should be able to support FK-joins
+        // at any level in the join tree, even after we add right-deep/bushy join tree support,
+        // because a FK-join output table has the same PK as its left input table
+        throw new KsqlException("Invalid join condition:"
+            + " foreign-key table-table joins are not supported.");
```
vcrfxia, line 620: nit: (to clarify that there is nothing fundamentally unsupported about this, just that we haven't added support yet)
```suggestion
            + " foreign-key table-table joins are not yet supported.");
```

ksqldb-engine/src/main/java/io/confluent/ksql/planner/LogicalPlanner.java
```diff
@@ -575,6 +579,113 @@ private JoinNode buildJoin(final Join root, final String prefix, final boolean i
...
+      final DataSourceNode qualifiedNode;
+
+      if (simpleJoinExpression.maybeQualifier().isPresent()) {
+        final String qualifier = simpleJoinExpression.maybeQualifier().get().text();
+        final List<DataSourceNode> allNodes = dataSourceNodes.stream()
+            .filter(n -> n.getDataSource().getName().text().equals(qualifier))
+            .collect(Collectors.toList());
+
+        if (allNodes.size() != 1) {
+          // double check this
```
vcrfxia, line 647: nit: remove comment.

ksqldb-engine/src/main/java/io/confluent/ksql/planner/LogicalPlanner.java
```diff
@@ -575,6 +579,113 @@ private JoinNode buildJoin(final Join root, final String prefix, final boolean i
...
+          throw new KsqlException(String.format(
+              "Join qualifier '%s' could not be resolved (either not found or not unique).",
+              qualifier
+          ));
+        }
+        qualifiedNode = allNodes.get(0);
+      } else {
+        final List<DataSourceNode> allNodes = dataSourceNodes.stream()
+            .filter(n -> n.getSchema().columns().stream()
```
vcrfxia, line 656: Why not use `n.getSchema().findColumn()` to simplify this?

ksqldb-engine/src/main/java/io/confluent/ksql/planner/LogicalPlanner.java
```diff
@@ -575,6 +579,113 @@ private JoinNode buildJoin(final Join root, final String prefix, final boolean i
...
+                .map(column -> column.name().text()).collect(Collectors.toList())
+                .contains(joinAttributeName))
+            .collect(Collectors.toList());
+
+        if (allNodes.size() != 1) {
+          // double check this
```
vcrfxia, line 662: Remove comment.

ksqldb-engine/src/main/java/io/confluent/ksql/planner/LogicalPlanner.java
```diff
@@ -575,6 +579,113 @@ private JoinNode buildJoin(final Join root, final String prefix, final boolean i
...
+          throw new KsqlException(String.format(
+              "Join identifier '%s' could not be resolved (either not found or not unique).",
+              joinAttributeName
+          ));
+        }
+        qualifiedNode = allNodes.get(0);
+      }
+
+      // we only support joins on single attributes so the key has a single field
```
vcrfxia, line 671: This comment does not make sense to me. The fact that we only support joins on single attributes does not mean a user will not try to issue a join on a single attribute that is only one of multiple key columns for a source. Do we have test coverage for this case?

ksqldb-engine/src/main/java/io/confluent/ksql/planner/LogicalPlanner.java
```diff
@@ -575,6 +579,113 @@ private JoinNode buildJoin(final Join root, final String prefix, final boolean i
...
+      // - we only support joins on single attributes so the key has a single field
+      singleAttributeKeyName = dataSourceNodes.get(0).getSchema().key().get(0).name().text();
+    }
+
+
+    return !joinAttributeName.equals(singleAttributeKeyName);
+  }
+
+  private boolean isInnerNode(final PlanNode node) {
+    return node.getSourceNodes().count() > 1;
```
vcrfxia, line 685: I'm still uncomfortable with the fact that this logic is brittle. If one day we decide to allow self-joins, then this breaks and it's very plausible that the person implementing self-joins won't realize they'll have broken the logic for detecting FK joins. Let's continue the discussion in https://github.com/confluentinc/ksql/pull/7452#discussion_r625262567.

ksqldb-functional-tests/src/test/resources/query-validation-tests/fk-join.json
```diff
@@ -0,0 +1,136 @@
...
+        "CREATE TABLE right_table (id3 BIGINT PRIMARY KEY, f3 BIGINT) WITH (kafka_topic='right_topic', format='JSON');",
+        "CREATE TABLE output AS SELECT id1, id2, id3, f1, f2, f3 FROM left_table JOIN middle_table ON f1 = id2 JOIN right_table ON id2 = id3;"
+      ],
+      "expectedException": {
+        "type": "io.confluent.ksql.util.KsqlStatementException",
+        "message": "Invalid join condition: foreign-key table-table joins are not supported."
+      }
+    }
+  ]
+}
```
vcrfxia, line 136: Thanks for the test coverage! Can we verify that we either already have tests for, or add tests for the following:
- ambiguous join attribute
- invalid data source qualifier
- invalid data source alias
- positive test cases for qualifiers and aliases (hopefully these already exist)

vcrfxia (commented):
Thanks @mjsax -- LGTM overall. Some minor comments/questions inline.

ksqldb-engine/src/main/java/io/confluent/ksql/planner/LogicalPlanner.java
vcrfxia: Why can't we check for `instanceOf` `JoiningNode` instead of `JoinNode`?

ksqldb-functional-tests/src/test/resources/query-validation-tests/joins.json
vcrfxia: With the changes in this PR, yeah. I'm fine removing this test from here. It'd be nice to update the new test for this case (that we only allow joins on the primary key of the table) to use an SR-enabled key format as an extra safe-guard.

ksqldb-engine/src/main/java/io/confluent/ksql/planner/LogicalPlanner.java
mjsax: Well, we want to test the "child" of `node`, not `node` itself, ie, `((PreJoinRepartitionNode)node).joiningNode`.

We could of course expose it, but would it be any better?

Something like adding `PreJoinRepartitionNode#isInnerJoinNode()`.

vcrfxia: `node.getSource() instanceOf JoiningNode`?

mjsax: Hmm... we have only a `PlanNode` at hand, but `getSource()` is a method on `SingleSourcePlanNode` (that extends `PlanNode`).

Also, digging a little more, it seems we could have either a `JoinNode`, or `PreJoinRepartitionNode`, or `PreJoinProjectionNode` here and it's unclear how many layer to go down to find a `JoinNode`.

Maybe better to hold off for now and keep the simple check as-is. We can still refactor in a follow up if we think its required.

ksqldb-engine/src/main/java/io/confluent/ksql/planner/LogicalPlanner.java
mjsax: Well, we add them shortly anyway...

ksqldb-engine/src/main/java/io/confluent/ksql/planner/LogicalPlanner.java
mjsax: From my understanding, as we have multiple source node, we first need to identify the correct source node base on the qualifier from the `joinExpression`, and can match the column name afterward . The `ColumnName` itself is just a plain `String` without qualifier and thus if we use `n.getSchema().findColumn()` we might find the column multiple times and/or in the wrong data source schema?

ksqldb-engine/src/main/java/io/confluent/ksql/planner/LogicalPlanner.java
mjsax: I can add a test -- for this case, we have a check in the `Analyzer` and thus would never reach this point in the code.

ksqldb-engine/src/main/java/io/confluent/ksql/planner/LogicalPlanner.java
mjsax: About self-joins: I actually think they won't break this logic. We want to figure out if the join happens on the PK, and if there is self-join we may have just one data source, but it still seems to be correct to execute the `else` of this statement to figure out correctly if the join happens on the PK or not.

ksqldb-functional-tests/src/test/resources/query-validation-tests/fk-join.json
mjsax: Seems there is only tests for "field does not exist" in `joins.json` -- happy to add those test, but seems orthogonal to this PR to close test gaps?

ksqldb-engine/src/main/java/io/confluent/ksql/planner/LogicalPlanner.java
vcrfxia: I don't understand. Why is 
```
.filter(n -> n.getSchema().columns().stream()
    .map(column -> column.name().text()).collect(Collectors.toList())
    .contains(joinAttributeName))
```
(as is currently written in the PR) different from
```
.filter(n -> n.getSchema().findColumn(ColumnName.of(joinAttributeName)).isPresent())
```
?

ksqldb-engine/src/main/java/io/confluent/ksql/planner/LogicalPlanner.java
vcrfxia: A self-join can change the primary key of the result. For example, a table-table outer self-join would result in a new, system-generated primary key. If this output is then used in another join, that join should check that we are joining on the PK of the output of the join, not the original PK, right? (Or is that not supported, in which case this isn't a concern from a correctness point of view; it'd just be confusing since the method name would be a bit misleading in certain cases.)

ksqldb-functional-tests/src/test/resources/query-validation-tests/fk-join.json
vcrfxia: Agreed that these tests should have already existed before this PR, but the fact that this PR changes logic that could impact these test cases means I'd feel more confident in the correctness of this PR if we had the test coverage. Feel free to add additional test coverage in a follow-up PR -- in general, these types of requests are not blocking, especially if there is other work blocked behind the functionality of the PR (which there is in this case).

ksqldb-engine/src/main/java/io/confluent/ksql/planner/LogicalPlanner.java
mjsax: Good point! Damn you, outer-joins...

ksqldb-engine/src/main/java/io/confluent/ksql/planner/LogicalPlanner.java
mjsax: My bad... I was looking into the wrong part of the code...

ksqldb-engine/src/main/java/io/confluent/ksql/planner/LogicalPlanner.java
```diff
@@ -575,6 +579,113 @@ private JoinNode buildJoin(final Join root, final String prefix, final boolean i
...
+  }
+
+  private boolean joinOnNonKeyAttribute(final Expression joinExpression,
+                                        final PlanNode node) {
+    if (!(joinExpression instanceof ColumnReferenceExp)) {
+      return true;
+    }
+    final ColumnReferenceExp simpleJoinExpression = (ColumnReferenceExp) joinExpression;
+
+    final String joinAttributeName = simpleJoinExpression.getColumnName().text();
```
agavra, line 632: can we avoid using `.text()`? making all the Strings type-safe was a goal of wrapping them all in classes and we really shouldn't be comparing `ColumnName` to anything other than that (same for `SourceName` below)

ksqldb-engine/src/main/java/io/confluent/ksql/planner/LogicalPlanner.java
```diff
@@ -575,6 +579,113 @@ private JoinNode buildJoin(final Join root, final String prefix, final boolean i
...
+        // after we lift the n-way join restriction, we should be able to support FK-joins
+        // at any level in the join tree, even after we add right-deep/bushy join tree support,
+        // because a FK-join output table has the same PK as its left input table
+        throw new KsqlException("Invalid join condition:"
+            + " foreign-key table-table joins are not supported.");
+      }
+    }
+  }
+
+  private boolean joinOnNonKeyAttribute(final Expression joinExpression,
```
agavra, line 625: without looking into the code, it feels to me like this should be part of `JoinTree` logic (and we should output this information in `JoinTree.Node`). Ideally, anything that has to do with "resolving" a Join should be outside of LogicalPlanner since this class is already getting a little unwieldly before this change - and also I think we can leverage the recursion used there to make this a little cleaner

ksqldb-engine/src/main/java/io/confluent/ksql/planner/LogicalPlanner.java
```diff
@@ -575,6 +579,113 @@ private JoinNode buildJoin(final Join root, final String prefix, final boolean i
...
+            .collect(Collectors.toList());
+
+        if (allNodes.size() != 1) {
+          // double check this
+          throw new KsqlException(String.format(
+              "Join qualifier '%s' could not be resolved (either not found or not unique).",
+              qualifier
+          ));
+        }
+        qualifiedNode = allNodes.get(0);
```
agavra, line 653: nit: `Iterables.getOnlyElement` is useful to make sure this survives refactoring

agavra (commented):
Some comments while I'm looking at it :) didn't review the tests or really the algorithms part of it, just the code structure

ksqldb-engine/src/main/java/io/confluent/ksql/planner/LogicalPlanner.java
```diff
@@ -575,6 +579,113 @@ private JoinNode buildJoin(final Join root, final String prefix, final boolean i
     );
   }
 
+  private void verifyJoin(final Join root, final PlanNode left, final PlanNode right) {
+    if (left.getNodeOutputType() == DataSourceType.KSTREAM) {
+      if (right.getNodeOutputType() == DataSourceType.KTABLE) {
+
+        // stream-table join detected
+
+        if (root.getInfo().getType().equals(JoinType.OUTER)) {
```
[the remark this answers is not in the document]
mjsax: I think there are two places where we could check: in the `Analyzer` or here. However, it think that the LogicalPlanner may be the better place.  From my understanding, the `Analyzer` won't even have enough context information?

ksqldb-engine/src/main/java/io/confluent/ksql/planner/LogicalPlanner.java
```diff
@@ -575,6 +579,113 @@ private JoinNode buildJoin(final Join root, final String prefix, final boolean i
...
+          throw new KsqlException("Invalid join type:"
+              + " full-outer join not supported for stream-table joins.");
+        }
+
+        if (joinOnNonKeyAttribute(root.getInfo().getRightJoinExpression(), right)) {
+          throw new KsqlException("Invalid join condition:"
+              + " stream-table joins require to join on the table's primary key.");
+        }
```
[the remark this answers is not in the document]
mjsax: If we have `ON stream.attributes = table.someAttribute` this conditions verifies `table.someAttribute` and would throw, because we can only use `table.key` for stream-table join. It does not matter here, if we use the stream-side key or some other attribute.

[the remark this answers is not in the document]
agavra: got it! (same comment as below then, let's call out what the key is and what we tried to join on)

ksqldb-engine/src/main/java/io/confluent/ksql/planner/LogicalPlanner.java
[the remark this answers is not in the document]
agavra: I guess my question was, where does it fail today?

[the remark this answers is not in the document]
mjsax: Needed to dig it out first... Currently, it fails later when we build the physical plan, ie, when we call `StreamToTableJoiner#join()`.

However, it seems much more reasonable to through in the logical planner... We should always ensure that we build a valid logical plan IMHO, and building the physical plan should _never_ fail. Or is there any reason why we would need to fail when building the physical plan? Cannot think of any reason.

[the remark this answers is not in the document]
agavra: I think I disagree here - the limitation is at the physical layer, not logical, so it's not unreasonable to throw at that level (imagine we had different back ends, say a pull query executor that just uses a network shuffle - would this limitation apply there?)

ksqldb-engine/src/main/java/io/confluent/ksql/planner/LogicalPlanner.java
guozhangwang: +10 here.

ksqldb-engine/src/main/java/io/confluent/ksql/planner/LogicalPlanner.java
```diff
@@ -575,6 +579,113 @@ private JoinNode buildJoin(final Join root, final String prefix, final boolean i
...
+    }
+    final ColumnReferenceExp simpleJoinExpression = (ColumnReferenceExp) joinExpression;
+
+    final String joinAttributeName = simpleJoinExpression.getColumnName().text();
+    final List<DataSourceNode> dataSourceNodes = node.getSourceNodes().collect(Collectors.toList());
+    final String singleAttributeKeyName;
+
+    // n-way join sub-tree (ie, not a leaf)
+    if (isInnerNode(node)) {
+      final DataSourceNode qualifiedNode;
```
guozhangwang, line 638: For my own understanding: is `DataSourceNode` the only node type whose `getSources()` return empty today?

ksqldb-engine/src/main/java/io/confluent/ksql/planner/LogicalPlanner.java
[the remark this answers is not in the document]
mjsax: For a stream-table join, a OUTER join is not defined, so it is really a logical issue. (If we could define it, KafkaStreams would have implemented it).

ksqldb-engine/src/main/java/io/confluent/ksql/planner/LogicalPlanner.java
mjsax: Yes. (Double checked the code).

ksqldb-engine/src/main/java/io/confluent/ksql/planner/LogicalPlanner.java
mjsax: I agree that we might want to do some refactoring, but it should not be part of this PR IMHO.

In particular, the logical planner should use a visitor pattern IMHO to build the logical plan. Just pushing out "random methods" into "random helper" classed, does make `LogicalPlanner` shorter, but does not really improve the code structure IMHO.

agavra: to defend my original comment 😝 I don't think `JoinTree` is a random helper, it already has the logic to extract keys and traverse N-way join trees. That's why I thought this would fit in there. Up to you if you feel there's a better way and when you want to do it (which PR)

ksqldb-engine/src/main/java/io/confluent/ksql/planner/LogicalPlanner.java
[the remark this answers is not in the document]
agavra: makes sense then!

ksqldb-engine/src/main/java/io/confluent/ksql/planner/LogicalPlanner.java
mjsax: I keep it in mind but don't want to hold off this PR any longer, as it blocks follow up PRs for FK-joins.

ksqldb-engine/src/main/java/io/confluent/ksql/planner/LogicalPlanner.java
```diff
@@ -575,6 +581,154 @@ private JoinNode buildJoin(final Join root, final String prefix, final boolean i
...
+            ));
+      }
+
+      // table-table join detected
+
+      if (joinOnNonKeyAttribute(rightExpression, rightNode)) {
+        throw new KsqlException(String.format(
+                "Invalid join condition:"
+                    + " table-table joins require to join on the primary key of the right input"
+                    + " table: %s = %s.",
```
vcrfxia, line 635: This error message seems confusing. It's printing what the user passed in after the colon, so a user might think that they're already passing in a valid join condition even though that's not the case. Ideally we'd clarify by printing what the "primary key of the right input" is but that's more work. A minor clarification in the meantime could be the following:
```suggestion
                    + " table. Got: %s = %s.",
```

ksqldb-engine/src/main/java/io/confluent/ksql/planner/LogicalPlanner.java
```diff
@@ -575,6 +581,154 @@ private JoinNode buildJoin(final Join root, final String prefix, final boolean i
...
+                  joinInfo.getLeftSource().getDataSource().getName().text(),
+                  joinType,
+                  joinInfo.getRightSource().getDataSource().getName().text()
+              ));
+        }
+
+        if (joinOnNonKeyAttribute(rightExpression, rightNode)) {
+          throw new KsqlException(String.format(
+                  "Invalid join condition:"
+                      + " stream-table joins require to join on the table's primary key: %s = %s.",
```
vcrfxia, line 609: Same comment as above (below).

vcrfxia (approved):
Thanks @mjsax -- great test coverage! Couple minor points inline but don't let those block you from merging this since this PR is blocking other work.

ksqldb-engine/src/main/java/io/confluent/ksql/planner/LogicalPlanner.java
mjsax: > "primary key of the right input

I tried that but it's not so simple, because there might be "tree"...

PR #7453, opened 2021-04-29
Kci 347/query anonymizer interface

swist:
### Description 
This is the KSQL side of [KCI-347](https://confluentinc.atlassian.net/browse/KCI-347). 
The idea is to expose a query logger which will take queries passed into it and anonymize them. It has a pretty rigid interface to avoid mistaking it for a regular logger, it lets you pass in a message and a query. You can configure it to pipe it to whatever appended you want

### Testing done 
Unit tested

### Reviewer checklist


ksqldb-parser/src/main/java/io/confluent/ksql/util/QueryGuid.java
```diff
@@ -21,27 +21,27 @@
...
-  private final String anonQueryUuid;
+  private final String queryGuid;
+  private final String structuralGuid;
   private final LocalDateTime timeOfCreation;
 
   public QueryGuid(final String namespace, final String nonAnonQuery, final String anonQuery) {
     this.clusterNamespace = namespace;
-    this.queryUuid = computeQueryId(nonAnonQuery, clusterNamespace);
-    this.anonQueryUuid = computeQueryId(anonQuery, "");
+    this.queryGuid = computeQueryId(nonAnonQuery, clusterNamespace);
```
rodesai, line 30: nit: name this `computeGuid` - query ID typically has a different meaning throughout the codebase

ksqldb-parser/src/main/java/io/confluent/ksql/util/QueryGuid.java
```diff
@@ -21,27 +21,27 @@
...
+    this.structuralGuid = computeQueryId(anonQuery, "");
```
rodesai, line 31: can we get rid of `getClusterNamespace`? it seems to be an internal detail and is only called from tests

ksqldb-rest-app/src/main/java/io/confluent/ksql/logging/query/QueryLogger.java
```diff
@@ -0,0 +1,107 @@
...
+
+  public static Logger getLogger() {
+    return logger;
+  }
+
+  public static void configure(final KsqlConfig config) {
+    rewriteAppender.setRewritePolicy(new QueryAnonymizingRewritePolicy(config));
+  }
+
+  private static ImmutableMap<String, Object> buildPayload(
```
rodesai, line 58: Seems like it would be better to use a stronger type (via a POJO) than a map here. I know we have tests to cover usage of this map (so the guarantees the compiler gives you aren't wins from a correctness pov per se), but it also makes things less error-prone for future devs. Plus it's more idiomatic.

ksqldb-rest-app/src/main/java/io/confluent/ksql/logging/query/QueryAnonymizingRewritePolicy.java
```diff
@@ -0,0 +1,91 @@
...
+        Stream.concat(oldMessage.entrySet().stream(), newMessage.entrySet().stream())
+            .collect(
+                toImmutableMap(Map.Entry::getKey, Map.Entry::getValue, (left, right) -> right));
+
+    return new LoggingEvent(
+        source.getFQNOfLoggerClass(),
+        source.getLogger() != null ? source.getLogger() : Logger.getLogger(source.getLoggerName()),
+        source.getTimeStamp(),
+        source.getLevel(),
+        combined,
```
rodesai, line 84: Might make sense to use an object that has a sane `toString` method in case someone wants to use this with an off-the-shelf appender.

ksqldb-rest-app/src/main/java/io/confluent/ksql/logging/query/QueryAnonymizingRewritePolicy.java
```diff
@@ -0,0 +1,91 @@
...
+
+    final QueryGuid queryGuids = buildGuids(query, anonymizedQuery);
+    final ImmutableMap<Object, Object> newMessage =
+        ImmutableMap.builder()
+            .put("query", anonymizeQueries ? anonymizedQuery : query)
+            .put("queryGUID", queryGuids.getQueryGuid())
+            .put("structuralGUID", queryGuids.getStructuralGuid())
+            .build();
+
+    final ImmutableMap<Object, Object> combined =
```
rodesai, line 74: won't this throw on the duplicate "query" key?

ksqldb-rest-app/src/test/java/io/confluent/ksql/logging/query/QueryAnonymizingRewritePolicyTest.java
```diff
@@ -0,0 +1,67 @@
...
+        .thenReturn("cathouse.org.meowcluster");
+
+
+    final RewriteAppender rewriteAppender = new RewriteAppender();
+    rewriteAppender.setRewritePolicy(new QueryAnonymizingRewritePolicy(config));
+    rewriteAppender.addAppender(testAppender);
+    LogManager.getLogger(QueryAnonymizingRewritePolicyTest.class).addAppender(rewriteAppender);
+
+  }
+
```
rodesai, line 46: add a case for when the anonymizer is disabled?

ksqldb-rest-app/src/main/java/io/confluent/ksql/logging/query/QueryAnonymizingRewritePolicy.java
rodesai: nvm, turns out `ImmutableMap.toString` _is_ sane!

ksqldb-rest-app/src/main/java/io/confluent/ksql/logging/query/QueryAnonymizingRewritePolicy.java
```diff
@@ -0,0 +1,91 @@
...
+        (ImmutableMap<String, String>) source.getMessage();
+    final String query = oldMessage.get("query");
+
+    final Map props = new HashMap(source.getProperties());
+
+    String anonymizedQuery;
+    try {
+      anonymizedQuery = anonymizer.anonymize(query);
+    } catch (ParsingException e) {
+      anonymizedQuery = "";
```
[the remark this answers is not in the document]
swist: Hmm I was trying to figure out the best semantic here. We certainly don't want to put the raw query content of things that failed to parse in here, because we might break our promise of not leaking PII. Perhaps it would make sense to add a log statement here? Otherwise I think silently failing should be good enough

ksqldb-rest-app/src/main/java/io/confluent/ksql/logging/query/QueryLogger.java
```diff
@@ -0,0 +1,107 @@
...
+import java.util.Enumeration;
+import org.antlr.v4.runtime.tree.ParseTree;
+import org.apache.log4j.Appender;
+import org.apache.log4j.LogManager;
+import org.apache.log4j.Logger;
+import org.apache.log4j.rewrite.RewriteAppender;
+
+public final class QueryLogger {
+  private static final Logger logger = LogManager.getLogger(QueryLogger.class);
+  private static final RewriteAppender rewriteAppender = new RewriteAppender();
```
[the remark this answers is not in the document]
swist: These are statically initialised - we just want one logger and one appender per entire application, this is just a utility class that configures these and gives them a facade

ksqldb-parser/src/main/java/io/confluent/ksql/util/QueryGuid.java
```diff
@@ -21,27 +21,27 @@
 
 public final class QueryGuid {
   private final String clusterNamespace;
-  private final String queryUuid;
-  private final String anonQueryUuid;
+  private final String queryGuid;
```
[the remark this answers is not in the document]
swist: Guid implies that these are globally unique identifiers for a concrete thing, also class is supposed to carry guids

ksqldb-common/src/main/java/io/confluent/ksql/util/KsqlConfig.java
```diff
@@ -154,9 +154,14 @@
   public static final String KSQL_WRAP_SINGLE_VALUES =
       "ksql.persistence.wrap.single.values";
 
+  public static final String KSQL_CCLOUD_QUERYANONYMIZER_ENABLED =
+      "ksql.ccloud.queryanonymizer.logs_enabled";
```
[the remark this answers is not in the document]
swist: Will do!

rodesai (approved):
LGTM!

lct45 (approved):
thanks @swist, lgtm

PR #7458, opened 2021-04-30
fix: Fixes org.mock-server to be 5.11.2 to utilize newer netty

AlanConfluent:
### Description 
Fixes test and the build.

### Testing done 
Ran test

### Reviewer checklist

### DIFF
--- a/ksqldb-version-metrics-client/pom.xml
+++ b/ksqldb-version-metrics-client/pom.xml
@@ -58,7 +58,7 @@
         <dependency>
             <groupId>org.mock-server</groupId>
             <artifactId>mockserver-netty</artifactId>
-            <version>3.10.4</version>
+            <version>5.11.2</version>
             <scope>test</scope>
             <exclusions>
                 <exclusion>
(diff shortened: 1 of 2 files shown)


agavra (approved):
Thanks again for the sluething! The size of the PR never does justice to the size of the investigation 😆

ksqldb-version-metrics-client/pom.xml
```diff
@@ -58,7 +58,7 @@
         <dependency>
             <groupId>org.mock-server</groupId>
             <artifactId>mockserver-netty</artifactId>
-            <version>3.10.4</version>
+            <version>5.11.2</version>
```
mjsax, line 61: Why we we set this version explicitly and not inherit from top level pom?

Also, as we do set the version explicitly, how could it happen that the build broke? Seems we pulled in a different version somehow during runtime? Is there something wrong with our setup?

AlanConfluent: My understanding is that another library from common, netty, overrode the version relied by on this library, to be a newer incompatible version. I effectively am upping the version of this library to now allow for a newer version of netty.

PR #7460, opened 2021-04-30
chore: refactor QueryMetadata and PersistentQueryMetadata

wcarlson5:
In order to make bin packed Queries easier to implement I refactored the current version.

Now I can just implement the PersistantQueryMetadata interface and drop it in place. This will make the feature flag logic and change over much easier.

### Reviewer checklist

### DIFF
--- a/ksqldb-engine/src/main/java/io/confluent/ksql/query/QueryRegistryImpl.java
+++ b/ksqldb-engine/src/main/java/io/confluent/ksql/query/QueryRegistryImpl.java
@@ -30,8 +30,10 @@ import io.confluent.ksql.schema.ksql.LogicalSchema;
 import io.confluent.ksql.serde.WindowInfo;
 import io.confluent.ksql.services.ServiceContext;
 import io.confluent.ksql.util.PersistentQueryMetadata;
+import io.confluent.ksql.util.PersistentQueryMetadataImpl;
 import io.confluent.ksql.util.QueryMetadata;
-import io.confluent.ksql.util.SandboxedPersistentQueryMetadata;
+import io.confluent.ksql.util.QueryMetadataImpl;
+import io.confluent.ksql.util.SandboxedPersistentQueryMetadataImpl;
 import io.confluent.ksql.util.SandboxedTransientQueryMetadata;
 import io.confluent.ksql.util.TransientQueryMetadata;
 import java.util.Collection;
@@ -84,10 +86,10 @@ public class QueryRegistryImpl implements QueryRegistry {
     insertQueries = new ConcurrentHashMap<>();
     original.allLiveQueries.forEach(query -> {
       if (query instanceof PersistentQueryMetadata) {
-        final PersistentQueryMetadata sandboxed = SandboxedPersistentQueryMetadata.of(
-            (PersistentQueryMetadata) query,
+        final PersistentQueryMetadata sandboxed = SandboxedPersistentQueryMetadataImpl.of(
+            (PersistentQueryMetadataImpl) query,
             new ListenerImpl()
-        );
+                                                                                         );
         persistentQueries.put(sandboxed.getQueryId(), sandboxed);
         allLiveQueries.add(sandboxed);
       } else {
@@ -332,7 +334,7 @@ public class QueryRegistryImpl implements QueryRegistry {
     );
   }
 
-  private class ListenerImpl implements QueryMetadata.Listener {
+  private class ListenerImpl implements QueryMetadataImpl.Listener {
     @Override
     public void onError(final QueryMetadata queryMetadata, final QueryError error) {
       eventListeners.forEach(l -> l.onError(queryMetadata, error));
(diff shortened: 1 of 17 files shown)


ableegoldman (commented):
Hey @wcarlson5 , I'm kind of struggling to see the bigger picture here. I think it would help if you could lay out a brief summary of the current hierarchy, and then expand on that to explain the motivation behind going from an abstract class to an interface+class.

I definitely thought that sounded like a good idea when we discussed it earlier, but after poking around the QueryMetadata & children code a bit I'm starting to feel that it might only be appropriate when we have the time to take on a larger refactoring/tech debt cleanup (which we currently do not).

I think we can focus on some specific smaller improvements for now and still achieve the desired outcome of just dropping in a new kind of PersistentQueryMetadata that uses the NamedTopology internally. I'm not sure an interface is the best/only way to do this, it seems to me that it just adds another layer (since you now have the interface and the implementing class), whereas the same thing can be achieved leaving QueryMetadata as an abstract class and promoting PersistentQueryMetadata to an abstract class as well, then extending that. You'll still have one additional class because you then need to extend the PersistentQueryMetadata abstract class, but it's only one class. Plus, if the two PersistentQueryMetadata classes end up sharing a lot of code (which I'm guessing they will, at least the boilerplate stuff) then an abstract class will make your life much easier than an interface.

Just recording my thoughts as I'm looking over this PR and the existing code. We can sync on this too

ksqldb-engine/src/main/java/io/confluent/ksql/query/QueryRegistryImpl.java
```diff
@@ -332,7 +334,7 @@ QueryExecutor create(
     );
   }
 
-  private class ListenerImpl implements QueryMetadata.Listener {
+  private class ListenerImpl implements QueryMetadataImpl.Listener {
```
guozhangwang, line 337: Should we still implement `QueryMetadata.Listener` here?

ksqldb-engine/src/main/java/io/confluent/ksql/util/PersistentQueryMetadata.java
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright 2018 Confluent Inc.
```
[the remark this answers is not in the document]
guozhangwang: Usually it reflects when the file is created.

ksqldb-engine/src/main/java/io/confluent/ksql/util/QueryMetadataImpl.java
```diff
@@ -0,0 +1,469 @@
...
+  public boolean hasEverBeenStarted() {
+    return everStarted;
+  }
+
+  public QueryId getQueryId() {
+    return queryId;
+  }
+
+  public KsqlQueryType getQueryType() {
+    return KsqlQueryType.PERSISTENT;
```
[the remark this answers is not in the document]
guozhangwang: I feel it's better to keep `QueryMetadataImpl` as abstract since we have another `TransientQueryMetadataImpl` already anyways.

ksqldb-engine/src/test/java/io/confluent/ksql/query/QueryExecutorTest.java
```diff
@@ -178,7 +179,7 @@
   @Mock
   private SessionConfig config;
   @Mock
-  private QueryMetadata.Listener queryListener;
+  private QueryMetadataImpl.ListenerImpl queryListener;
```
guozhangwang, line 182: Why we need to declare as `QueryMetadataImpl.ListenerImpl` here now?

guozhangwang (approved):
Since this is refactoring, I'm assuming that there's no logic change at all (regret that github diff is not smart enough to reflect this..), LMK if there's any piggy-backed logic changes.

Only nit comments, please feel free to merge afterwards.

ksqldb-engine/src/main/java/io/confluent/ksql/util/PersistentQueryMetadata.java
```diff
@@ -15,205 +15,50 @@
...
-  }
+      final QueryContext.Stacker contextStacker);
+
+  void restart();
+
+  void stop();
+
+  StreamsUncaughtExceptionHandler.StreamThreadExceptionResponse uncaughtHandler(
+      final Throwable error
+                                                                                         );
```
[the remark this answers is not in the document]
wcarlson5: I think you are right :)

104 lines so no

ksqldb-engine/src/test/java/io/confluent/ksql/query/QueryExecutorTest.java
wcarlson5: I just removed `ListenerImpl` so no

ksqldb-engine/src/main/java/io/confluent/ksql/util/QueryMetadataImpl.java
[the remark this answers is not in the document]
ableegoldman: > I feel it's better to keep QueryMetadataImpl as abstract

+1

ksqldb-engine/src/main/java/io/confluent/ksql/util/PersistentQueryMetadataImpl.java
```diff
@@ -0,0 +1,220 @@
+/*
+ * Copyright 2018 Confluent Inc.
```
ableegoldman, line 2: Since you're adding a new file here, this should be 2021.

ksqldb-engine/src/main/java/io/confluent/ksql/util/QueryMetadataImpl.java
```diff
@@ -0,0 +1,452 @@
+/*
+ * Copyright 2018 Confluent Inc.
```
ableegoldman, line 2: ditto here

PR #7463, opened 2021-05-03
chore: bump version

wcarlson5:
update dependency version

### Reviewer checklist

### DIFF
--- a/ksql-version-metrics-client/pom.xml
+++ b/ksql-version-metrics-client/pom.xml
@@ -42,6 +42,7 @@
         <dependency>
             <groupId>com.fasterxml.jackson.core</groupId>
             <artifactId>jackson-databind</artifactId>
+            <version>2.12.3</version>
         </dependency>
 
         <dependency>
--- a/pom.xml
+++ b/pom.xml
@@ -96,7 +96,7 @@
         <antlr.version>4.7.1</antlr.version>
         <csv.version>1.4</csv.version>
         <lang3.version>3.3.1</lang3.version>
-        <guava.version>24.1.1-jre</guava.version>
+        <guava.version>30.1.1-jre</guava.version>
         <inject.version>1</inject.version>
         <janino.version>3.0.7</janino.version>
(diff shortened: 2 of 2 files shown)


spena (approved):

ableegoldman (approved):
LGTM

ableegoldman:
@wcarlson5 the semantic check thing failed, can you fix that? After that, can this be merged? Final question: can/should we pint merge this up to master? It seems very odd to me that the 5.3.x branch is already on a newer version of guava than master (see [this](https://github.com/confluentinc/ksql/pull/7501/files#r631333485))

wcarlson5:
@ableegoldman I also had to updated one more dependency I am not sure if this still needs to be pint merged. I thought the guava was updated in 6.2 as well

ksql-version-metrics-client/pom.xml
```diff
@@ -42,6 +42,7 @@
         <dependency>
             <groupId>com.fasterxml.jackson.core</groupId>
             <artifactId>jackson-databind</artifactId>
+            <version>2.12.3</version>
```
ableegoldman, line 45: Interesting, this didn't have a version before? I wonder if it was somehow just fetching the version (or the dependency itself )as a transitive dependency since many of the other artifacts we pull in also use jackson -- did you test which jackson version was ultimately being pulled in when ksql is compiled?

wcarlson5: yeah, it was the same that they detected that had a CVE. I don't know enough about mvn to explain that tho

ableegoldman: Fair enough. Although I guess the flaw with the test I brought up is that we could be pulling in the CVE version just because whatever dependency we're getting jackson from hasn't yet upgraded their version of jackson to fix the CVE. I'm also not an maven expert, but maybe there's a way to detect which dependency is responsible for pulling in this version of jackson?
Who are the maven experts here...I want to say, maybe @agavra or @vvcephei will know?

PR #7464, opened 2021-05-04
chore: update QTT framework & tests to support improvements on left/outer joins

spena:
### Description 
^ It also fixes the ksql/master branch

### Testing done 

### Reviewer checklist

### DIFF
--- a/ksqldb-common/src/main/java/io/confluent/ksql/util/KsqlConfig.java
+++ b/ksqldb-common/src/main/java/io/confluent/ksql/util/KsqlConfig.java
@@ -16,6 +16,7 @@
 package io.confluent.ksql.util;
 
 import static io.confluent.ksql.configdef.ConfigValidators.zeroOrPositive;
+import static org.apache.kafka.streams.StreamsConfig.InternalConfig.ENABLE_KSTREAMS_OUTER_JOIN_SPURIOUS_RESULTS_FIX;
 
 import com.google.common.base.Splitter;
 import com.google.common.collect.ImmutableList;
@@ -543,7 +544,7 @@ public class KsqlConfig extends AbstractConfig {
             StreamsConfig.OPTIMIZE),
           // Disable KAFKA-10847 fix on older topologies
           new CompatibilityBreakingStreamsConfig(
-            StreamsConfig.InternalConfig.ENABLE_KSTREAMS_OUTER_JOIN_SPURIOUS_RESULTS_FIX,
+            ENABLE_KSTREAMS_OUTER_JOIN_SPURIOUS_RESULTS_FIX,
             false,
             true)
   );
@@ -565,8 +566,7 @@ public class KsqlConfig extends AbstractConfig {
     }
 
     private boolean isInternal(final String name) {
-      return
-          name.equals(StreamsConfig.InternalConfig.ENABLE_KSTREAMS_OUTER_JOIN_SPURIOUS_RESULTS_FIX);
+      return name.equals(ENABLE_KSTREAMS_OUTER_JOIN_SPURIOUS_RESULTS_FIX);
     }
 
 
--- a/ksqldb-common/src/main/java/io/confluent/ksql/util/KsqlConfig.java
+++ b/ksqldb-common/src/main/java/io/confluent/ksql/util/KsqlConfig.java
@@ -540,7 +540,12 @@ public class KsqlConfig extends AbstractConfig {
           new CompatibilityBreakingStreamsConfig(
             StreamsConfig.TOPOLOGY_OPTIMIZATION_CONFIG,
             StreamsConfig.OPTIMIZE,
-            StreamsConfig.OPTIMIZE)
+            StreamsConfig.OPTIMIZE),
+          // Disable KAFKA-10847 fix on older topologies
+          new CompatibilityBreakingStreamsConfig(
+            StreamsConfig.InternalConfig.ENABLE_KSTREAMS_OUTER_JOIN_SPURIOUS_RESULTS_FIX,
+            false,
+            true)
   );
 
   private static final class CompatibilityBreakingStreamsConfig {
@@ -551,13 +556,19 @@ public class KsqlConfig extends AbstractConfig {
     CompatibilityBreakingStreamsConfig(final String name, final Object defaultValueLegacy,
         final Object defaultValueCurrent) {
       this.name = Objects.requireNonNull(name);
-      if (!StreamsConfig.configDef().names().contains(name)) {
+      if (!StreamsConfig.configDef().names().contains(name) && !isInternal(name)) {
         throw new IllegalArgumentException(
             String.format("%s is not a valid streams config", name));
       }
       this.defaultValueLegacy = defaultValueLegacy;
       this.defaultValueCurrent = defaultValueCurrent;
     }
+
+    private boolean isInternal(final String name) {
+      return
+          name.equals(StreamsConfig.InternalConfig.ENABLE_KSTREAMS_OUTER_JOIN_SPURIOUS_RESULTS_FIX);
+    }
+
 
     String getName() {
       return this.name;
--- a/ksqldb-execution/src/main/java/io/confluent/ksql/schema/query/QuerySchemas.java
+++ b/ksqldb-execution/src/main/java/io/confluent/ksql/schema/query/QuerySchemas.java
@@ -35,8 +35,6 @@ import java.util.Map;
 import java.util.Objects;
 import java.util.Optional;
 import java.util.Set;
-import org.slf4j.Logger;
-import org.slf4j.LoggerFactory;
 
 /**
  * Pojo for holding data about the schemas and formats in use at the different stages within a
@@ -55,8 +53,6 @@ import org.slf4j.LoggerFactory;
  * between releases.
  */
 public final class QuerySchemas {
-  private static final Logger LOG = LoggerFactory.getLogger(QuerySchemas.class);
-
   private static final String EMPTY_TOPIC_NAME = "";
 
   private static final ValueFormat NONE_VALUE_FORMAT = ValueFormat.of(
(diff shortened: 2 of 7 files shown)


ksqldb-execution/src/main/java/io/confluent/ksql/schema/query/QuerySchemas.java
```diff
@@ -129,24 +144,35 @@ public SchemaInfo getTopicInfo(final String topicName) {
...
     }
 
     final SchemaInfo keyInfo = loggerToSchemas.get(Iterables.getOnlyElement(keyLoggerNames));
-    final SchemaInfo valueInfo = loggerToSchemas.get(Iterables.getOnlyElement(valueTopicNames));
+
+    // If valueTopicsNames has multiple logger names, then takes only the first logger found.
+    // Multiple loggers may be caused by a topic that has multiple value formats, i.e. the shared
+    // outer store from left/outer joins supports multiple value formats. Currently, the
+    // getTopicInfo() method is only called by ksqldb-functional-tests, which uses the same
+    // value format on both left/right joins.
```
mjsax, line 171: > which uses the same value format on both left/right joins.

As discussed in person. I this actually correct, ie, can we rely on it? Do we have test for mix value formats for stream-stream joins?

ksqldb-functional-tests/src/main/java/io/confluent/ksql/test/tools/TopicInfoCache.java
```diff
@@ -102,7 +102,16 @@ public TopicInfoCache(
...
 
   public TopicInfo get(final String topicName) {
-    return cache.getUnchecked(topicName);
+    final TopicInfo topicInfo = cache.getUnchecked(topicName);
+
+    // Value formats can not be determined if a null (tombstone) value was produced in the topic.
+    // If the valueFormat is not determined, then attempt to refresh the entry cached to get the
+    // right format in case a non-null value is already produced.
+    if (topicInfo.getValueFormat().getFormat().equals(FormatFactory.NONE.name())) {
+      cache.refresh(topicName);
```
mjsax, line 111: For my own education: what is this `cache` thing? And won't we need to get `TopicInfo` from the cache again, after we refreshed it? What does `refresh()` do internally (ie, how does a refresh work)?

ksqldb-common/src/main/java/io/confluent/ksql/util/KsqlConfig.java
```diff
@@ -549,14 +554,20 @@ void defineCurrent(final ConfigDef configDef) {
...
         throw new IllegalArgumentException(
             String.format("%s is not a valid streams config", name));
       }
       this.defaultValueLegacy = defaultValueLegacy;
       this.defaultValueCurrent = defaultValueCurrent;
     }
 
+    private boolean isInternal(final String name) {
+      return
+          name.equals(StreamsConfig.InternalConfig.ENABLE_KSTREAMS_OUTER_JOIN_SPURIOUS_RESULTS_FIX);
```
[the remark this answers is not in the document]
spena: It was more than 100 columns. I just imported the static InternalConfig to have all in one line. The null check is already done in the constructor.

ksqldb-execution/src/main/java/io/confluent/ksql/schema/query/QuerySchemas.java
```diff
@@ -129,24 +144,35 @@ public SchemaInfo getTopicInfo(final String topicName) {
...
           + "topic: " + topicName
           + "loggers: " + keyLoggerNames
       );
     }
 
     final Set<String> valueTopicNames = kvLoggerNames.getOrDefault(false, ImmutableSet.of());
-    if (valueTopicNames.size() != 1) {
-      throw new IllegalStateException("Multiple value logger names registered for topic."
+    if (valueTopicNames.size() > 1) {
+      LOG.debug("Multiple value logger names registered for topic."
```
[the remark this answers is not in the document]
spena: I changed this code. If the multiple values do not have the same value format, then it will throw an exception. I left a comment about the zero cases too. I cannot throw with zero cases.

ksqldb-execution/src/main/java/io/confluent/ksql/schema/query/QuerySchemas.java
spena: I'm not sure if add a test for this because it should be supported. But the QTT is the only client using the `getTopicInfo` and had the issue with the multiple value formats. I added code just to verify multiple values have the same format. I don't think the `getTopicInfo` will be called by anyone else. 

We should refactor the code, though, so it is only used by testing.

ksqldb-functional-tests/src/main/java/io/confluent/ksql/test/tools/TopicInfoCache.java
spena: The `cache.getUnchecked(topicName)` will call the `load(final String topicName)` internally if no items exist in the cache. After that is called, consecutive calls to `getUnchecked` will use the item in the cache, and `load` will not be called any more until the item has expired. 

The problem I am fixing is that sometimes, the first record found in the shared outer changelog topic has a null value. This causes the `getTopicInfo` to return a `NO_VALUE_SCHEMA_FOUND`. When that info is obtained from the `load` function, then it is kept in the cache for some time. I know that the null value is just something that may not happen too often, so a new record in the chanelog topic should allow the TrackSerde to detect the value format. However, the cache will not have the new format until the item expires, so in cases of the NONE value found, I refresh the cache so it triggers the `load` method again in case the value format was detected.

ksqldb-execution/src/main/java/io/confluent/ksql/schema/query/QuerySchemas.java
```diff
@@ -118,6 +132,60 @@ public synchronized void trackSerdeOp(
...
+        Iterables.getFirst(topicNames, EMPTY_TOPIC_NAME),
+        NO_VALUE_SCHEMA_FOUND);
+
+    if (topicNames.size() > 1) {
+      final Optional<ValueFormat> firstValueFormat = valueInfo.valueFormat();
+
+      // Multiple value topics names may be detected. This happens with left/outer joins. The left
+      // and right streams share a state store which share one changelog topic with different
+      // value formats. The track code, though, is only used for QTT which always executes joins
+      // with same value formats. We only make sure all formats are the same so we return one
```
mjsax, line 170: > is only used for QTT which always executes joins with same value formats

This seems to be the way the test are effectively written but not a fundamental limitation. At it seems QTT should be able to test all valid queries and joining two streams with different value formats seems to be a valid query that we support atm, and should keep supporting?

mjsax (approved):
From my side LGTM. Might be good to get a second review from @agavra or @vcrfxia though, to verify.

ksqldb-execution/src/main/java/io/confluent/ksql/schema/query/QuerySchemas.java
agavra: +1 seems like something we should support and test. if there's a bug here then there might be real bugs in the code somewhere. out of scope for this PR but can you create a ticket tracking it?

ksqldb-functional-tests/src/main/java/io/confluent/ksql/test/tools/TopicInfoCache.java
agavra: Is that the only time that it can be `None`? Otherwise we risk always refreshing the cache and getting the same result (I think so, but worth double checking)

agavra (approved):
LGTM! thanks

ksqldb-execution/src/main/java/io/confluent/ksql/schema/query/QuerySchemas.java
vcrfxia: +1 to the confusion above. Is it true that this code only affects the intermediate topic value formats written into the historic plan spec files, and that the lack of support for differing value formats in stream-stream joins introduced here is only limited to what gets written into the spec, rather than our ability to run such tests and verify inputs and outputs (assuming we were to remove the error thrown on line 177 below)?

ksqldb-execution/src/main/java/io/confluent/ksql/schema/query/QuerySchemas.java
```diff
@@ -118,6 +132,60 @@ public synchronized void trackSerdeOp(
...
+      // SchemaInfo.
+      for (final String name : topicNames) {
+        final Optional<ValueFormat> nextValueFormat = loggerToSchemas.getOrDefault(
+            name, NO_VALUE_SCHEMA_FOUND).valueFormat();
+
+        if (!firstValueFormat.equals(nextValueFormat)) {
+          throw new IllegalStateException("Multiple value logger names registered for topic."
```
vcrfxia, line 177: Assuming my understanding above is correct, can we update this error message to be more clear about what the problem is? If this error is hit, that means that someone wrote a stream-stream join test with differing value formats, and that this is not currently supported by the test framework. It'd also be good to link to the issue that @agavra proposed we file to track a fix for this in the error message, as a way to provide additional context. (Without this clarification, I'd be really confused if I wrote a test I expected to pass and then hit this.)

vcrfxia (commented):
Thanks @spena . Can we update the PR description to summarize what this PR is about? It'd be great to link to the Streams change that the changes in this PR are addressing as I don't think this PR is understandable by anyone without the relevant context.

ksqldb-execution/src/main/java/io/confluent/ksql/schema/query/QuerySchemas.java
mjsax: I am still wondering how complex it would be to lift this restriction in this PR. If too complex, I am also happy if we address it in a follow up PR, but given the new stream-stream-join requirement, it might be worth to actually add a stream-stream join QTT test with different value format now...

vcrfxia:
We've decided to disable the stream-stream left/outer join enhancements in ksqlDB until the grace period is configurable, so this PR is not needed to unblock master anymore. (https://github.com/confluentinc/ksql/pull/7569 will be merged instead.) The changes in this PR are still useful for when we eventually do want to enable the enhancements since we'll need the updates to the QTT framework in this PR at that time. @spena do you prefer to leave this PR open and update it at that time, or close it in favor of opening a new one later? I'll leave that up to you.

spena:
This PR was implemented in https://github.com/confluentinc/ksql/pull/7695

PR #7465, opened 2021-05-04
chore: Change pull query tag names to be shorter and less redundant

AlanConfluent:
### Description 
Changes tag names for pull query metrics.

### Testing done 
Ran unit tests.

### Reviewer checklist

### DIFF
--- a/ksqldb-common/src/main/java/io/confluent/ksql/util/KsqlConstants.java
+++ b/ksqldb-common/src/main/java/io/confluent/ksql/util/KsqlConstants.java
@@ -44,8 +44,8 @@ public final class KsqlConstants {
 
   public static final String KSQL_SERVICE_ID_METRICS_TAG = "ksql_service_id";
   public static final String KSQL_PULL_QUERY_SOURCE_TAG = "ksql_pull_query_source";
-  public static final String KSQL_PULL_QUERY_PLAN_TYPE_TAG = "query_plan_type";
-  public static final String KSQL_PULL_QUERY_ROUTING_TYPE_TAG = "query_routing_type";
+  public static final String KSQL_QUERY_PLAN_TYPE_TAG = "query_plan_type";
+  public static final String KSQL_QUERY_ROUTING_TYPE_TAG = "query_routing_type";
 
   public enum KsqlQueryType {
     PERSISTENT,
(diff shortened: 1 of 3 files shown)


ksqldb-common/src/main/java/io/confluent/ksql/util/KsqlConstants.java
```diff
@@ -44,8 +44,8 @@ private KsqlConstants() {
 
   public static final String KSQL_SERVICE_ID_METRICS_TAG = "ksql_service_id";
   public static final String KSQL_PULL_QUERY_SOURCE_TAG = "ksql_pull_query_source";
-  public static final String KSQL_PULL_QUERY_PLAN_TYPE_TAG = "ksql_pull_query_plan_type";
-  public static final String KSQL_PULL_QUERY_ROUTING_TYPE_TAG = "ksql_pull_query_routing_type";
+  public static final String KSQL_PULL_QUERY_PLAN_TYPE_TAG = "query_plan_type";
+  public static final String KSQL_PULL_QUERY_ROUTING_TYPE_TAG = "query_routing_type";
```
agavra, line 48: is there any backwards compatibility concerns here or have these not been published yet?

AlanConfluent: They've only been published on devel, and I don't believe that's a problem.  The metrics pipeline ignores metrics with tags it doesn't recognize.

agavra (approved):

ksqldb-common/src/main/java/io/confluent/ksql/util/KsqlConstants.java
vvcephei: Are we staking a claim to the word "query" to mean "pull query"? I'm wondering if "pull_routing_type" is more identifiable.

vvcephei (approved):
Just one question.

guozhangwang (approved):
Change LGTM, will chat with Alan offline for me to understand the rationale.

AlanConfluent:
@vvcephei 
> Are we staking a claim to the word "query" to mean "pull query"? I'm wondering if "pull_routing_type" is more identifiable.
I don't think so.  These can be included on any metrics I believe.  We could, for example use them as part of push queries. 
 That said, I'm not sure if it would be confusing to reuse it in this way if they have different value spaces under each metric type.  It seems ok to me to treat them as more generic extensions of the metric name.

AlanConfluent:
> @vvcephei
> 
> > Are we staking a claim to the word "query" to mean "pull query"? I'm wondering if "pull_routing_type" is more identifiable.
> 
> I don't think so. These can be included on any metrics I believe. We could, for example use them as part of push queries.
> That said, I'm not sure if it would be confusing to reuse it in this way if they have different value spaces under each metric type. It seems ok to me to treat them as more generic extensions of the metric name.

In fact pulling out a query_query = (pull|push) was another explicit recommendation, though I'm not sure if I like that vs just putting it in the metric name, they they aren't necessarily the same metrics.

Issue #5673, opened 2020-06-24
Unreadable state store when running a query with a source table with a schema-registry schema


rodesai:
We recently discovered an issue in streams that impacts ksql (https://issues.apache.org/jira/browse/KAFKA-10179). KS can optimize materialization of a topology's source table by using the source topic as the table's change-log rather than creating a new change-log topic. This becomes an issue during restore, because the restore just copies the source topic data into rocksdb. This is a problem for schemas registered with schema-registry. The data in the source topic has a schema id that is registered under the source topic's schema. However, the deserializer for the state store is provided a subject for the store, which doesn't have that schema registered, and deserialization fails.

Streams has proposed a fix in KAFKA-10179. However, the proposed fix - to associate the source topic name with the changelog state store - is not going to work for ksql. In ksql, the schema of the data going into the changelog may be different from the source topic’s schema (e.g. we may only be using some of the source’s fields), and we can’t register that schema with the source topic’s subject in the schema registry. We need to do our own work to solve the following:

How do we want to proceed for new queries? We could use the optimization, but then we must read and materialize the source using the source’s schema, and project out the needed fields in a downstream transformation. Alternatively, we could disable the optimization (provided streams gives us a way of doing this).

How do we want to handle existing queries? Existing queries may be in a corrupted state (e.g. corrupted state stores). How do we handle this? Some options:
- try and detect this state and mark the queries as failed
- detect this from the deserializer and drop corrupted rows
- play some games with the deserializer to look for schemas in both subjects during deserialization

agavra:
Steps to reproduce:

```sql
create table t_source (id int primary key, `col1` int) with (kafka_topic='t', value_format='avro', partitions=1);
create table t (id int primary key) with (kafka_topic='t', value_format='avro');
create stream s (id int key, col1 int) with (kafka_topic='s', value_format='avro', partitions=1);
insert into t_source  VALUES (1, 1);
insert into s (id, col1) VALUES (1, 1);
SET 'auto.offset.reset'='earliest';
create stream j as select t.*, s.* from s join t on s.col1 = t.id;

--- at this point you need to delete your streams data dir and
--- restart the ksql application

insert into s (id, col1) VALUES (1, 1);
```

The following topics have been created:
```
ksql> SHOW TOPICS;

 Kafka Topic                 | Partitions | Partition Replicas
---------------------------------------------------------------
 J                           | 1          | 1
 default_ksql_processing_log | 1          | 1
 s                           | 1          | 1
[...]

Issue #3585, opened 2019-10-15
KSQL should not delete the fact topic not created by KSQL


brianwhu:
When a KSQL stream or table is registered from an existing Kafka topic with CREATE STREAM/TABLE statement, the topic does not belong to the stream or table, and therefore should be left alone even when the stream or table is dropped with the "DELETE TOPIC" clause.

On v5.2.1, the cli tool always deletes the associated topic when a stream or a table is dropped. This has the potential to cause serious data loss.

agavra:
Hello @brianwhu - thanks for pointing this out! We've given this a lot of thought and decided a while back that this behavior is expected and should remain supported. If the user explicitly mentions `DELETE TOPIC`, then we should respect this explicit command; we don't differentiate between topics that were created by KSQL and those that weren't. This becomes especially important when you look at more recent features in 5.3 that allow KSQL to create a topic in a `CREATE STREAM/TABLE` statement if it doesn't exist, and then populate it using `INSERT INTO ... VALUES`.

I do, though, understand your concern. I think the proper solution to this problem is ACLs on topics that are important - any production topic that should not be "deletable" by KSQL should be guarded by the proper permissions.

Note that in 5.3.x we made this behavior much safer (see #2680)  so that at least we won't keep deleting the topic every time we recover...

I'm closing out this ticket to keep our backlog clean, but feel free to continue the conversation!

brianwhu:
Thanks for sharing your thoughts and reasoning for the current design. I see your points, particularly the one on `DELETE TOPIC` being explicitly expressed by a user. I do want to share a couple of brief points in favor of my position, just in case you want to give it some more thoughts.

1. In principle, CREATE and DROP are mirror operations where DROP removes what CREATE brings about, possibly including implicitly created objects but nothing more. If KSQL knows it didn't create a topic it really has no business in destroying it, no matter what the user asks. Not deleting those topics is actually more consistent with the CREATE/DROP contract all SQL users have come to expect.

2. The use cases where the behavior in question is expected and natural are probably in the minority. I'd argue that when a user issues `DROP TABLE/STREAM ... DELETE TOPIC` where the topic is preexisting, the `DELETE `clause is most likely a mistake. It's probably better to do less and give user the option.

Note: I do see that `DELETE TOPIC` is quite safe and valuable when the topics are auto-generated by KSQL.

agavra:
[...]

PR #7474, opened 2021-05-04
Draft add source table materialization

abbccdda:
### Description 

### Testing done 

### Reviewer checklist


design-proposals/klip-47-add-source-stream-table-semantic.md
```diff
@@ -0,0 +1,122 @@
...
+* We will not add role-based write permission for source stream/table in v1, even though it seems to be useful.
+
+## Design
+
+### Semantic Change
+We shall first add optional keyword `SOURCE` into the KSQL codebase, by adding the new syntax to `sqlBase.g4` for:
+* createStream
+* createStreamAs
+* createTable
+* createTableAs
```
guozhangwang, line 52: I'd suggest not piggy-back on existing logical node, but rather create separate createSourceStream and createSourceTable, since in the long run the createStream/Table would have different semantics, and createStreamAs/TableAs would be replaced too.

design-proposals/klip-47-add-source-stream-table-semantic.md
```diff
@@ -0,0 +1,122 @@
...
+```
+Error: insertion into source stream/table is not allowed
+```   
+
+* In addition, the `CREATE SOURCE TABLE` will materialize the input source topic into a stream 
+instance which would generate a RocksDB instance for interactive queries.
+
+
+## What is in scope
+* Add `CREATE SOURCE STREAM/TABLE` syntax to KSQL
```
guozhangwang, line 36: Would we support CREATE OR UPDATE syntax as well? If yes maybe better to clarify too.

guozhangwang (commented):
cc @mjsax @agavra for another look.

design-proposals/klip-47-add-source-stream-table-semantic.md
```diff
@@ -0,0 +1,122 @@
...
+are populated by the upstream Kafka topic, making the data ownership more clear._
+           
+## Motivation and background
+
+Users of KsqlDB have no clear ownership definition around tables or streams they created. In a “normal” database, 
+when a user creates a table the DB creates the corresponding files on disk and owns the table. 
+User may write data into those tables via `INSERT INTO` statements and/or modify/delete existing data 
+via `UPDATE` and `DELETE` statements. However, the table will be not changed by any other means. 
+
+For ksqlDB, the setup is quite differently. Specifically, user could create the database based on
```
mjsax, line 18: `could create the database` sounds confusing? What do you mean by it?

design-proposals/klip-47-add-source-stream-table-semantic.md
```diff
@@ -0,0 +1,122 @@
...
+           
+## Motivation and background
+
+Users of KsqlDB have no clear ownership definition around tables or streams they created. In a “normal” database, 
+when a user creates a table the DB creates the corresponding files on disk and owns the table. 
+User may write data into those tables via `INSERT INTO` statements and/or modify/delete existing data 
+via `UPDATE` and `DELETE` statements. However, the table will be not changed by any other means. 
+
+For ksqlDB, the setup is quite differently. Specifically, user could create the database based on 
+the topics stored in Kafka as the source of truth. The defined database could be mutable by
```
mjsax, line 19: The source of truth would always be Kafka topics for ksqlDB -- the point is _who_ owns those topics, and who is writing into them, right? If there is an upstream CDC pipeline that populates the topic, the CDC producer "owns" the topic and ksqlDB should only read it, but not write into it.

design-proposals/klip-47-add-source-stream-table-semantic.md
```diff
@@ -0,0 +1,122 @@
...
+
+* Syntax to add read-only stream/table as `CREATE SOURCE STREAM/TABLE` to make the data source as read-only. Any 
+insertion command will be rejected, with an example error response as:
+
+```
+Error: insertion into source stream/table is not allowed
+```   
+
+* In addition, the `CREATE SOURCE TABLE` will materialize the input source topic into a stream 
+instance which would generate a RocksDB instance for interactive queries.
```
mjsax, line 32: "stream instance" might be confusing -- I understand that you refer to Kafka Stream, but not everybody might have enough context.

It might be worth to point out, that `CREATE SOURCE TABLE` is not just a metadata operation like `CREATE TABLE`, `CREATE STREAM`, and `CREATE SOURCE STREAM`, but effectively deploys a "persistent query" (for a lack of a better term) in the ksqlDB servers.

design-proposals/klip-47-add-source-stream-table-semantic.md
mjsax: Atm, we support `CREATE OR REPLACE` but it's for queries... (if `CSAS` and `CTAS`). I don't believe we support to modify (base) streams/tables created via `CS` or `CT`. Thus we would not need to support it I guess? \cc @agavra

agavra: we do support it for base streams/tables :) it just updates the schema. you can also use `ALTER`

design-proposals/klip-47-add-source-stream-table-semantic.md
mjsax: If am fine either way, but the `SOURCE` keyword only applies to `createStream ` and `createTable`, but not to `createStreamAs` or `createTableAs`.

mjsax: Does this section go to much into implementation details?

design-proposals/klip-47-add-source-stream-table-semantic.md
```diff
@@ -0,0 +1,122 @@
...
+In addition, when user calls `DESCRIBE ... EXTENDED`, the read-only attribute shall be displayed for stream/table. 
+
+### Make Source Stream/Table Read-only
+We plan to add restriction check into the `DistributingExecutor#throwIfInsertOnReadOnlyTopic` or create a similar function, 
+which could verify whether the given stream/table is read-only. If so, the insertion will be rejected.
+
+### Source Table Materialization
+In the call `EngineExecutor#execute`, we will check whether given plan has a `queryPlan` before deciding to proceed 
+creating a persistent query. To bypass this check, we need to provide a concrete query plan within the input plan. We would 
+inject a dummy query plan in `EngineExecutor#plan` phase, when the statement is a CreateTable statement with `SOURCE` tag.
```
mjsax, line 68: Not sure if a dummy query plan would be a "clean" solution? \cc @AlanConfluent @vvcephei

design-proposals/klip-47-add-source-stream-table-semantic.md
```diff
@@ -0,0 +1,122 @@
...
+
+In addition, when user tries to show all the running queries, the materialization query should also be displayed as it takes 
+part of the computation resource. This would be a special type of running query without `SELECT` clause.
```
mjsax, line 71: This is an interesting aspect. I agree to the sentiment, but we should not display it as a "query" to the user IMHO. That might be confusing. It should be clear for the user that it's a running "source table program" (for the lack of a better term).

design-proposals/klip-47-add-source-stream-table-semantic.md
```diff
@@ -0,0 +1,122 @@
...
+
+Here comes a precaution we might be aware. The original `CREATE TABLE` command does not need materialization, so there is 
+no enforcement for primary key. However, in order to let pull query run on the materialized state, the store must be keyed so that
```
mjsax, line 74: > so there is no enforcement for primary key

From my understanding defining a PK is enforced for existing `CT` statement.

mjsax: Follow up question for @agavra @vcrfxia: How strict are we about verifying the physical schema of the key compare to the defined PK? For values, I know that we effectively just ignore additional data (ie, if the AVRO record has 10 fields, and a user only does a `CS` picking 4 field it is fine), and for missing fields we just put `NULL`.

However, for keys it seems we might break co-partitioning if we are not strict about it? Would we drop all records on read if the define PK does not exactly match the physical schema of the message key?

design-proposals/klip-47-add-source-stream-table-semantic.md
```diff
@@ -0,0 +1,122 @@
...
+
+In long-term, we don't want to bloat `AstBuilder` with all the logical constraint checks. In the short-term, the complexity 
+of #1 is low, and the expected code change to enforce this constraint should be low, so we would pick #1. 
+
+## Public APIS
+* Create source table/stream, which makes the data source read-only:
+
+```roomsql
+CREATE SOURCE STREAM stream_name (COL1 INT, COL2 STRING) AS ...
+CREATE SOURCE TABLE table_name (ID INT KEY, COL1 INT) AS ...
```
mjsax, line 91: For `TABLES` the correct syntax is `PRIMARY KEY` (not just `KEY`).

mjsax: There is also no `AS...`, because we consider `CS/CT` not `CSAS/CTAS` -- there wont be a query, but only a `WITH` clause that define the input topic.

design-proposals/klip-47-add-source-stream-table-semantic.md
```diff
@@ -0,0 +1,122 @@
...
+
+## LOEs and Delivery Milestones
+The estimated work time will be spent as follows (to implement and review):
+1. Add `SOURCE` syntax for both stream/table takes a week
+2. Make changes in Metastore and CreateSource for read-only sources takes a week to review
+3. Reject insertion to `SOURCE` table/stream takes a week
+4. Display read-only attributes on `DESCRIBE ... EXTENDED` 3 days 
+5. Source table materialization takes a week
+6. Documentation changes takes 3 days
+7. Integration testing takes 2 weeks
```
mjsax, line 108: I think the changes to make pull queries work are missing in this list?

design-proposals/klip-47-add-source-stream-table-semantic.md
```diff
@@ -0,0 +1,122 @@
...
+There will be 3 milestones, first is the completion of syntax addition (#1), which marks the start 
+point of splitting the work between engineers. Second milestone will be the feature complete ($2~5), and 
+the last one will be documentation and tests complete.
+
+## Documentation Updates
+* Add a new doc page as `docs/developer-guide/ksqldb-reference/source-table.md`
+* Add a new doc page as `docs/developer-guide/ksqldb-reference/source-stream.md`
+
+## Compatibility Implications
+This is a new feature in CTAS and CSAS, which has no backward compatibility issue.
```
mjsax, line 119: > in CTAS and CSAS

It's not relate to `CSAS/CTAS` but to `CS/CT`

design-proposals/README.md
```diff
@@ -86,5 +86,6 @@ Next KLIP number: **47**
...
 | [KLIP-45: Materializing Tables for Query](klip-45-materializing-tables-for-query.md)                                             | Proposal |                   |            | [Discussion](https://github.com/confluentinc/ksql/pull/6997) |
-| KLIP-46: DATE and TIME data type support                                                                                         | Proposal |        0.20.0     |            | [Discussion](https://github.com/confluentinc/ksql/pull/7417) |
+| [KLIP-46: DATE and TIME data type support](klip-46-date-and-time-data-type-support.md)                                           | Proposal |      0.20.0       |            | [Discussion](https://github.com/confluentinc/ksql/pull/7417) |
+| [KLIP-47: Add source stream/table semantic](klip-47-add-source-stream-table-semantic.md)                                         | Proposal |                   |            | [Discussion](https://github.com/confluentinc/ksql/pull/7474) |
```
vcrfxia, line 90: KLIP 47 is already taken. Follow the proposal intent procedure to lock in a number before opening the KLIP: https://github.com/confluentinc/ksql/blob/master/design-proposals/README.md#proposal-intent

design-proposals/klip-47-add-source-stream-table-semantic.md
```diff
@@ -0,0 +1,122 @@
...
+
+In addition, when user calls `DESCRIBE ... EXTENDED`, the read-only attribute shall be displayed for stream/table. 
+
+### Make Source Stream/Table Read-only
+We plan to add restriction check into the `DistributingExecutor#throwIfInsertOnReadOnlyTopic` or create a similar function, 
+which could verify whether the given stream/table is read-only. If so, the insertion will be rejected.
+
+### Source Table Materialization
+In the call `EngineExecutor#execute`, we will check whether given plan has a `queryPlan` before deciding to proceed 
+creating a persistent query. To bypass this check, we need to provide a concrete query plan within the input plan. We would
```
vcrfxia, line 67: Does creating a persistent query mean we're creating not just a RocksDB for the materialization but also a new sink topic? If so, what will the sink topic be named? How will we protect against users deleting the new sink topic out of confusion, and causing the query to fail as a result?

design-proposals/klip-47-add-source-stream-table-semantic.md
vcrfxia: Piggybacking off Matthias's comment, will we allow users to interact with this query? For example, will it expose a query ID like other queries? How will we prevent users from terminating the query if so?

design-proposals/klip-47-add-source-stream-table-semantic.md
vcrfxia: No, we don't do any such verification today. Keys and values are deserialized in the same way.

Why is co-partitioning a concern with pull queries? The CREATE SOURCE TABLE statement is going to have to copy the data into a new state store to enable pull queries anyway. As long as ksqlDB controls the key serializer used for converting the record into bytes for the state store, then finding the proper partition at the time of issue a pull query for a specific key will not be an issue. 

OTOH, if the proposal is to directly copy bytes from the Kafka key and record into RocksDB, then serialization discrepancies at the time of pull queries being issued is indeed a concern. That would be a concern not only if the specified key is only a subset of the actual key, but also for Schema-Registry-enabled key formats in case the schema ID is different.

design-proposals/klip-47-add-source-stream-table-semantic.md
```diff
@@ -0,0 +1,122 @@
...
+CREATE SOURCE STREAM stream_name (COL1 INT, COL2 STRING) AS ...
+CREATE SOURCE TABLE table_name (ID INT KEY, COL1 INT) AS ...
+```
+
+## Test plan
+The new semantics will be tested as:
+
+* Unit testing for modular changes
+* QTTs with the new source table/stream syntax, including insertion failure and successful materialization
+
```
vcrfxia, line 99: We should also have tests for `DESCRIBE` on the sources, and tests to ensure that users are able to drop source streams/tables without a hitch. (Watch out for the case where, today, we disallow dropping sources if there are queries reading from them.)

design-proposals/klip-47-add-source-stream-table-semantic.md
agavra: I don't know this off the top of my head - but we use the same serde for keys/values so I suspect we're not strict about it. The only difference is that we pass in `isKey=true` to the (connect) serdes so that might change the way the handle it but I think it's unlikely.

We probably do want to be strict here.

design-proposals/klip-47-add-source-stream-table-semantic.md
abbccdda: What I mean is table, let me change it.

design-proposals/klip-47-add-source-stream-table-semantic.md
mjsax: There should not be a sink topic. I guess the term "persistent query" is tricky -- we basically need to run a Kafka Streams program and "persistent query == Kafka Streams" program, even if this Kafka Streams program does not really run a _query_... Guess we don't have a good term for this special case.

design-proposals/klip-47-add-source-stream-table-semantic.md
mjsax: What's the impact on queries that use those tables? -- How should it work for pull queries for source tables? Do we require a default value or would be just return NULL?

design-proposals/klip-47-add-source-stream-table-semantic.md
mjsax: I don't think that there would be query ID -- the underlying program should be bound to the table IMHO, so user can `DROP` the table (and this will terminate the program, too).

design-proposals/klip-47-add-source-stream-table-semantic.md
mjsax: I think, without a repartitioning step, co-partitioning cannot be guaranteed, and it sound like as if ksqlDB would no re-partition? -- I though that if we run a source table program, we would just read from the topic and put into RocksDB...

Also for right now, when we run a persistent query, would we repartition? Assume you do a table-table join, do we repartition both input topic before the join?

design-proposals/klip-47-add-source-stream-table-semantic.md
abbccdda: That makes sense.

design-proposals/README.md
abbccdda: Sounds good, created PR: https://github.com/confluentinc/ksql/pull/7496/

design-proposals/klip-47-add-source-stream-table-semantic.md
abbccdda: +1 for Matthias, we don't plan to add a sink topic.

design-proposals/klip-47-add-source-stream-table-semantic.md
agavra: > What's the impact on queries that use those tables?

There is no impact, a query could not have selected the field so it is not piped through. A `SELECT *` is resolved at query-time, so it doesn't 'magically' update when you alter a source.

> How should it work for pull queries for source tables?

Same as with any other query - if you issue the pull query after the field was added you would get it, otherwise you wouldn't.

design-proposals/klip-49-add-source-stream-table-semantic.md
```diff
@@ -0,0 +1,129 @@
...
+* Make source stream/table immutable from KSQL by rejecting insertion requests
+ 
+
+## What is not in scope
+* We will not add role-based write permission for source stream/table in v1, even though it seems to be useful.
+
+## Design
+
+### Semantic Change
+We shall first add optional keyword `SOURCE` into the KSQL codebase, by adding the new syntax to `sqlBase.g4` for:
```
agavra, line 49: nit: no need to change this for this KLIP, but going forward I think we don't need to go into implementation details this deep - it distracts a bit from the design aspect of the KLIP (these kind of changes are better reviewed in PRs)

design-proposals/klip-49-add-source-stream-table-semantic.md
```diff
@@ -0,0 +1,129 @@
...
+Additionally, when user calls `DESCRIBE ... EXTENDED`, the read-only attribute shall be displayed for stream/table. 
+
+### Make Source Stream/Table Read-only
+We plan to add restriction check into the `DistributingExecutor#throwIfInsertOnReadOnlyTopic` or create a similar function, 
+which could verify whether the given stream/table is read-only. If so, the insertion will be rejected.
+
+### Source Table Materialization
+In the call `EngineExecutor#execute`, we will check whether given plan has a `queryPlan` before deciding to proceed 
+creating a persistent query. To bypass this check, we need to provide a concrete query plan within the input plan. We would 
+inject a dummy query plan in `EngineExecutor#plan` phase, when the statement is a CreateTable statement with `SOURCE` tag.
```
agavra, line 71: I don't think this is a "dummy" query plan - it's going to spin up a real query 😄 we should make sure to summarize the offline discussion about counting this toward the number of persistent queries

design-proposals/klip-49-add-source-stream-table-semantic.md
```diff
@@ -0,0 +1,129 @@
...
+
+In addition, when user tries to show all the running queries, the materialization query should also be displayed as it takes 
+part of the computation resource. This would be a special type of running query without `SELECT` clause or a query id, as it 
+should be only associated with source table, and user could choose to `DROP` the entire table if they don't want to use resource 
+to do the materialization.
```
agavra, line 76: what happens if they try to `TERMINATE` just the materializing query without dropping the table? Should we support this?

design-proposals/klip-49-add-source-stream-table-semantic.md
```diff
@@ -0,0 +1,129 @@
...
+
+Here comes a precaution we might be aware. The original `CREATE TABLE` command does not need materialization, so there is 
+no enforcement for primary key. However, in order to let pull query run on the materialized state, the store must be keyed so that
```
agavra, line 79: Are you sure about this? I thought that we do require that a `CREATE TABLE` has at least one primary key column - what we don't ensure is that the actual data that is sent matches the said schema (otherwise it can't be used for anything downstream really)

design-proposals/klip-49-add-source-stream-table-semantic.md
```diff
@@ -0,0 +1,129 @@
...
+
+## Documentation Updates
+* Add a new doc page as `docs/developer-guide/ksqldb-reference/source-table.md`
+* Add a new doc page as `docs/developer-guide/ksqldb-reference/source-stream.md`
+
+## Compatibility Implications
+This is a new feature in CT and CS, which has no backward compatibility issue. 
+
+## Security Implications
+None
```
agavra, line 129: I think it might make sense to explicitly call out that this is _not_ a security feature. For the scope of this PR it seems that you would still be allowed to `CREATE TABLE` over an existing topic (not `SOURCE TABLE`) and that we're not deprecating that functionality yet. Is that correct?

agavra (commented):
We should make sure to loop in the discussions that are ongoing offline to this KLIP. Otherwise here's a quick pass.

design-proposals/klip-47-add-source-stream-table-semantic.md
agavra: +1 on too much implementation detail

design-proposals/klip-47-add-source-stream-table-semantic.md
agavra: I think this is a good solution, except for that it's not really a "dummy" - it's a real query!

design-proposals/README.md
agavra: 🐎

design-proposals/klip-47-add-source-stream-table-semantic.md
agavra: we might also want to be explicit here about changelog topics - I suspect we _do_ want to create one here?

design-proposals/klip-47-add-source-stream-table-semantic.md
mjsax: Not sure if we should call it a query, because there is no `SELECT...` statement. It's of course some KafkaStreams program that we need to run a deploy, but I don't think that the term "persistent query" or "query plan" really applies? -- If we have a `CST` statement, there is no reason to build a "logical query plan" (and later when we have an optimizer pipe it through the optimizer). Lastly, persistent queries always have a sink node to write the result into an output  topic, but we don't need/want a sink node for a `CST` "program".

We discussed this question in person yesterday and it was unclear if we should re-use the exiting "query planning codepath" and extend the exiting code to be able to handle this "special plan" (what seems to be your proposal), or if we should add a new codepath and process `CST` statements independently (similar to the dedicated code path we have for `CT / CS` and pull queries).

Might be worth to keep discussing in more detail which option might be the better one. It's still a little unclear to me. \cc @vcrfxia

design-proposals/klip-47-add-source-stream-table-semantic.md
mjsax: I am not 100% sure if we would need a changelog topic... If we assume that the input data is partitioned correctly (we did discuss the issue if users specify a "partial key"...) we could restore from the input topic. Of course, we would require that the input topic is configured with log compaction enabled.

Again, it seems we need to figure out what assumption about "valid" vs "invalid" patterns we want to apply/enforce to make a decision on this question.

Also note, if people use the "source table" in a query, we won't be able to share the state store anyway atm, but each query would read the input topic by itself, so we can design the `CST` "program" independent (for now). Of course, we also need to make sure that pull queries work correctly.

agavra: I'm worried about not having a changelog here for the same reason as https://github.com/confluentinc/ksql/issues/5673

design-proposals/klip-49-add-source-stream-table-semantic.md
```diff
@@ -0,0 +1,129 @@
...
+To address these gaps, we would like to add:
+
+* Syntax to add read-only stream/table as `CREATE SOURCE STREAM/TABLE` to make the data source as read-only. Any 
+insertion command will be rejected, with an example error response as:
+
+```
+Error: insertion into source stream/table is not allowed
+```   
+
+* In addition, the `CREATE SOURCE TABLE` will become a persistent query instead of just a metadata operation like
```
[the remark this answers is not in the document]
abbccdda: I think the reasoning for only materializing `SOURCE TABLE` is that we don't want to have all `TABLE` type materialized immediately, but instead taking a first opt-in approach and see user's feedback.

design-proposals/klip-49-add-source-stream-table-semantic.md
abbccdda: Not sure I follow you here, are you asking whether it is allowed to have multiple source/non-source table being created on top of the same topic?

design-proposals/klip-49-add-source-stream-table-semantic.md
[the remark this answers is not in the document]
mjsax: It would be a "massive" (breaking?) change to let `CT` start to materialize as people expect it to be a light-weight metadata operation. -- In the long run, my personal proposal is, that we would materialize all tables. But there are also concerns raised by others (ie, cost) that propose to make is configurable via `WITH` clause.

Doing this experiment that is backward compatible should give us more data points to see how we want to move forward.

design-proposals/klip-49-add-source-stream-table-semantic.md
mjsax: > it's going to spin up a real query

I tend to disagree (at least from a terminology point of view). A "query" should have a `SELECT *` statements, but a `CT` and `CST` both don't have a SELECT but are DDL statements... Maybe we need to come up with a new term for this kind of KafkaStreams program that we deploy in the background.

design-proposals/klip-49-add-source-stream-table-semantic.md
mjsax: This should not be allows. There won't even be a query-id, so user cannot use TERMINATE to begin with. There should only be a `DROP TABLE` statement that terminate the KS program and delete the table metadata from the catalog.

design-proposals/klip-49-add-source-stream-table-semantic.md
agavra: yup, that's essentially what I meant

design-proposals/klip-49-add-source-stream-table-semantic.md
abbccdda: That sounds right, I think primary key is required for CT

design-proposals/klip-49-add-source-stream-table-semantic.md
```diff
@@ -0,0 +1,129 @@
...
+* In addition, the `CREATE SOURCE TABLE` will become a persistent query instead of just a metadata operation like 
+`CREATE TABLE`, `CREATE STREAM`, or `CREATE SOURCE STREAM`. Source table will materialize the input topic data as 
+a RocksDB instance for interactive queries in the KsqlDB servers.
+
+## What is in scope
+* Add `CREATE SOURCE STREAM/TABLE` syntax to KSQL
+* `DESCRIBE ... EXTENDED` should show source streams and tables as read-only  
+* Add materialization of the `source table` to make it pull-query accessible
+* Make source stream/table immutable from KSQL by rejecting insertion requests
+
```
mjsax, line 41: We should add that `DROP` statements should not support the `DELETE TOPIC` clause for source streams/tables.

cf https://github.com/confluentinc/ksql/issues/3585

design-proposals/klip-49-add-source-stream-table-semantic.md
abbccdda: That's a good question, IMHO, if we eventually have a full RBAC-like model for KSQL, it is logistic to create multiple tables on the same topic with different user roles, where the `admin` could create the table with write access, while a `consumer` role should be only allowed to create read-only table. I could mention explicitly here that we don't want to enforce such restriction at the moment for no obvious benefit.

design-proposals/klip-49-add-source-stream-table-semantic.md
```diff
@@ -0,0 +1,129 @@
...
+Error: insertion into source stream/table is not allowed
+```   
+
+* In addition, the `CREATE SOURCE TABLE` will become a persistent query instead of just a metadata operation like 
+`CREATE TABLE`, `CREATE STREAM`, or `CREATE SOURCE STREAM`. Source table will materialize the input topic data as 
+a RocksDB instance for interactive queries in the KsqlDB servers.
+
+## What is in scope
+* Add `CREATE SOURCE STREAM/TABLE` syntax to KSQL
+* `DESCRIBE ... EXTENDED` should show source streams and tables as read-only
```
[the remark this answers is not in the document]
abbccdda: Make sense, I think it's not.

design-proposals/klip-49-add-source-stream-table-semantic.md
guozhangwang: +1

design-proposals/klip-49-add-source-stream-table-semantic.md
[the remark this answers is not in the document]
agavra: Can we add this discussion to the KLIP if it's not already there?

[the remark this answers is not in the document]
abbccdda: Sounds good

design-proposals/klip-49-add-source-stream-table-semantic.md
abbccdda: I have mentioned this in the detailed design for source table materialization, where we do reject termination of the source table query. The slight change would be still giving the query a query.id, which is just not possible to be terminated directly, since I feel it might not be the best solution since we introduce the overhead for telling users what this special type of query is and why we need to hide its query.id.

guozhangwang (approved):
I was discussing the other day with @vvcephei about always materializing and counting as a persistent query for the new DDL, and would that surprise many users.

I think for the moment, if we message it just as "this is a syntax sugar as for

CREATE TABLE t1 WITH ... ;
CREATE TABLE t2 AS SELECT * from T1 .. (with materialization);

But note that you would not be able to INSERT into it, which kinda makes sense since it is a source table/stream.
"

Then people would be more acceptive for it to always materialize and count as persistent queries. In the future as we have higher persistent query limit and introduce materialized views and complete the ownership model, we can nudge our messaging to focus on the `ownership` side, and perhaps also allow users to opt-out of materialization.

guozhangwang:
Just to clarify: I was not suggesting that we should not move forward :slightly_smiling_face: was just suggesting that if after we release it out, users complain about enforcing the materialization / count as persistent query, then we can at least argue for the moment that it is a syntax sugar.

Regarding whether should we always materialize or not **in the long run** though: I think if a data is cdc’ed source, then it should always be declared as a table , and 90% of the time that table would be involved in some join and hence need to be materialized anyways, but if it was for the other 10% (say, the table is only for aggregation, or for non-stateful filter), we should allow advanced users to opt-out of materialization for reducing footprint

mjsax (approved):

design-proposals/klip-49-add-source-stream-table-semantic.md
```diff
@@ -0,0 +1,161 @@
...
+```roomsql
+CREATE SOURCE STREAM stream_name (COL1 INT, COL2 STRING) WITH ...
+CREATE SOURCE TABLE table_name (ID INT PRIMARY KEY, COL1 INT) WITH ...
+```
+
+## Test plan
+The new semantics will be tested as:
+
+* Unit testing for modular changes
+* QTTs with the new source table/stream syntax, including insertion failure, `DROP ... DELETE TOPIC` command integrity
```
vcrfxia, line 124: QTTs don't support `DROP STREAM/TABLE`. We'll have to write separate integration tests (separate from the QTT framework for this). 

(QTTs also don't support INSERT VALUES but RQTT does so we're fine on that one.)

design-proposals/klip-49-add-source-stream-table-semantic.md
```diff
@@ -0,0 +1,161 @@
...
+* Tests for successful materialization and pull query
+* Tests for query termination restriction
+* Tests for `DESCRIBE` sources
```
vcrfxia, line 127: We should also test/verify the output of `LIST QUERIES;` to ensure that the output makes clear that the source table materialization comes with a query, but that this query cannot/should not be terminated.

design-proposals/klip-49-add-source-stream-table-semantic.md
```diff
@@ -0,0 +1,161 @@
...
+stateful job that comes with overhead, which should definitely be exposed to end users. However, any exposed query 
+could be potentially terminated directly by the end user, which comes with a consistency problem with source table 
+semantic agreement. If a source table query could be terminated independently by the user, the original source table 
+would be dangling and not available for pull query anymore. To address this problem, we have two approaches:
+ 
+ 1. The KSQL server rejects the termination of a source table query and informs end user to try deleting the original source table.
+ 2. The KSQL server will verify the deletion of a query, and try deleting the associated source table with it.
+ 
+Both approaches are valid here to maintain the consistency, but considering when dropping a source table, the active 
+materialization query would be terminated, I'm inclined to suggest taking approach #1 here for simplicity, and avoid
```
vcrfxia, line 93: So just to clarify: the plan is to show the "query" associated with source table materialization as part of `LIST QUERIES;`, but to reject an attempt to terminate the query? Previously we'd discussed not showing a query ID for these queries at all, so users cannot even try to terminate them. I have a slight preference for the latter but am fine either way.

vcrfxia (approved):
LGTM with a few minor points inline. Thanks, all!

design-proposals/klip-49-add-source-stream-table-semantic.md
jzaralim: There's a drop_source.json file and an insert-values.json file in the test folder that contain the commands, so we should be fine with QTTs

vcrfxia: `insert-values.json` is the RQTT file I was referring to, yeah.

I didn't know about `drop_source.json`. I don't think `DROP ... DELETE TOPIC;` is going to work, though. (The fact that QTT is being used to test `DROP STREAM/TABLE` statements at all is sort of a hack since that's not what it's meant to do. The tests in `drop_source.json` have dummy CSAS statements for this reason.)

PR #7459, opened 2021-04-30
chore: Override dependencies: netty -> 4.1.49.Final, jersey-common -> 2.31

AlanConfluent:
### Description 
This is effectively a rollback to our previous version of netty since the update to 4.1.63.Final in common is breaking master.  Also pins jersey-common and fixes a small bug breaking master

### Testing done 
_Describe the testing strategy. Unit and integration tests are expected for any behavior changes._

### Reviewer checklist
- [ ] Ensure docs are updated if necessary. (eg. if a user visible feature is being added or changed).
- [ ] Ensure relevant issues are linked (description should include text like "Fixes #<issue number>")

PR #7475, opened 2021-05-05
fix: AdminResponseHandler should treat empty string as Optional.empty()

mjsax:
follow up to https://github.com/confluentinc/ksql/pull/7455#discussion_r625322594

Running `ClientTest` locally it actually failed without this fix. It's unclear to me, why it did not fail on #7455 (or did we miss it, and merge after failing Jenkins??)

### DIFF
--- a/ksqldb-api-client/src/main/java/io/confluent/ksql/api/client/impl/AdminResponseHandlers.java
+++ b/ksqldb-api-client/src/main/java/io/confluent/ksql/api/client/impl/AdminResponseHandlers.java
@@ -335,7 +335,9 @@ final class AdminResponseHandlers {
           source.getString("valueFormat"),
           formatQueries(source.getJsonArray("readQueries")),
           formatQueries(source.getJsonArray("writeQueries")),
-          Optional.ofNullable(source.getString("timestamp")),
+          source.getString("timestamp").isEmpty()
+              ? Optional.empty()
+              : Optional.of(source.getString("timestamp")),
           Optional.ofNullable(source.getString("windowType")),
           source.getString("statement"),
           source.getJsonArray("sourceConstraints").stream()


ksqldb-api-client/src/main/java/io/confluent/ksql/api/client/impl/AdminResponseHandlers.java
```diff
@@ -335,7 +335,9 @@ static boolean isConnectErrorResponse(final JsonObject ksqlEntity) {
           source.getString("valueFormat"),
           formatQueries(source.getJsonArray("readQueries")),
           formatQueries(source.getJsonArray("writeQueries")),
-          Optional.ofNullable(source.getString("timestamp")),
+          source.getString("timestamp").isEmpty()
```
spena, line 338: Can't getString("timesetamp" be null?

vvcephei (approved):
Thanks for the quick revert, @mjsax !

ksqldb-api-client/src/main/java/io/confluent/ksql/api/client/impl/AdminResponseHandlers.java
mjsax: Not sure... I just re-established the old code... \cc @vcrfxia do you know?

mjsax:
Seems there is some overlap to https://github.com/confluentinc/ksql/pull/7459/files#r626723704

Maybe we don't need to merge this one.

ksqldb-api-client/src/main/java/io/confluent/ksql/api/client/impl/AdminResponseHandlers.java
vcrfxia: No, it cannot be null: https://github.com/confluentinc/ksql/blob/05f5f617dbc30296c048ae0a421a1a9d97b5fead/ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/entity/SourceDescriptionFactory.java#L94

vcrfxia: Ideally this code would've been more defensive but the real fix we want is https://github.com/confluentinc/ksql/issues/6042 anyway.

Issue #6374, opened 2020-10-07
Split physical from logical schema


agavra:
Today, our logical schema has "key" and "value" columns. There is a 1:1 mapping between these logical columns and their physical location. This can get confusing when the kafka record key is a composition of multiple logical columns:

```sql
CREATE STREAM one (a INT KEY, b INT);
CREATE TABLE two AS SELECT a, b, count(*) GROUP BY a, b;
```

In this example, the physical Kakfa record key would be some structured concatenation of `a` and `b`, but that concatenation does not exist in the SQL schema. The logical columns that make up the key (`a` and `b`) exist in the SQL schema.

After #6371, we will be able to access `a` and `b` independently by copying them into the value for easy-access. Once we support structured keys (#824) we will be able to optimize away this wasted storage, and instead "extract" them from the key. In order to accomplish this, we need to be able to allow lookups for columns in either the key or the value, depending on the physical serialization.

PR #7477, opened 2021-05-05
fix: multi-column keys are broken in some scenarios when rearranged

agavra:
### Description 

There are a few situations where key columns are re-arranged (differ from the source/partition by). You can reproduce like this:
```sql
CREATE STREAM inputStream (id INT KEY, a INT, b INT, c INT) WITH (kafka_topic='input', partitions='4');

INSERT INTO input VALUES (1, 11, 21, 31);
INSERT INTO input VALUES (2, 12, 22, 32);
INSERT INTO input VALUES (1, 11, 21, 31);

CRATE STRAM repartitionedStream AS SELECT * FROM inputStream PARTITION BY c+5, b;
SELECT * FROM repartitionedStream EMIT CHANGES;

+-----------------+-----------------+-----------------+-----------------+-----------------+
|B                |KSQL_COL_0       |A                |C                |ID               |
+-----------------+-----------------+-----------------+-----------------+-----------------+
|36               |21               |11               |31               |1                |
|37               |22               |12               |32               |2                |
|36               |21               |11               |31               |1                |
```

This is caused because of a mismatch in the source schema and projection schema. This PR fixes this issue in two places:

1. in `PARTITION BY`, a `SELECT *` is resolved using the source schema and doesn't take into account the ordering of the keys selected in the `PARTITION BY` clause
2. in any `SELECT` (projection) clause where the keys are not ordered the same as the source, this PR just reorders it for the user so that it matches the source schema

### Testing done 

QTT tests

### Reviewer checklist

### DIFF
--- a/ksqldb-engine/src/main/java/io/confluent/ksql/planner/plan/PlanNode.java
+++ b/ksqldb-engine/src/main/java/io/confluent/ksql/planner/plan/PlanNode.java
@@ -17,7 +17,9 @@ package io.confluent.ksql.planner.plan;
 
 import static java.util.Objects.requireNonNull;
 
+import com.google.common.collect.ImmutableMap;
 import com.google.common.collect.Iterables;
+import com.google.common.collect.Maps;
 import com.google.common.collect.Streams;
 import io.confluent.ksql.execution.expression.tree.ColumnReferenceExp;
 import io.confluent.ksql.execution.expression.tree.Expression;
@@ -35,6 +37,7 @@ import io.confluent.ksql.structured.SchemaKStream;
 import io.confluent.ksql.util.GrammaticalJoiner;
 import io.confluent.ksql.util.KsqlException;
 import java.util.List;
+import java.util.Objects;
 import java.util.Optional;
 import java.util.Set;
 import java.util.stream.Collectors;
@@ -192,10 +195,13 @@ public abstract class PlanNode {
       final LogicalSchema schema
   ) {
     // When doing a `select *` key columns should be at the front of the column list
-    // but are added at the back during processing for performance reasons.
+    // but are added at the back during processing for performance reasons. Furthermore,
+    // the keys should be selected in the same order as they appear in the source.
     // Switch them around here:
-    final Stream<Column> keys = columns.stream()
-        .filter(c -> schema.isKeyColumn(c.name()));
+    final ImmutableMap<ColumnName, Column> columnsByName = Maps.uniqueIndex(columns, Column::name);
+    final Stream<Column> keys = schema.key().stream()
+        .map(key -> columnsByName.get(key.name()))
+        .filter(Objects::nonNull);
 
     final Stream<Column> windowBounds = columns.stream()
         .filter(c -> SystemColumns.isWindowBound(c.name()));
--- a/ksqldb-engine/src/main/java/io/confluent/ksql/planner/plan/SelectionUtil.java
+++ b/ksqldb-engine/src/main/java/io/confluent/ksql/planner/plan/SelectionUtil.java
@@ -51,7 +51,7 @@ public final class SelectionUtil {
    *
    * Essentially, we need to build a logical schema that mirrors the physical
    * schema until https://github.com/confluentinc/ksql/issues/6374 is addressed.
-   * That means that the keys must be ordered in the same was as the parent schema
+   * That means that the keys must be ordered in the same way as the parent schema
    * (e.g. if the source schema was K1 INT KEY, K2 INT KEY and the projection is
    * SELECT K2, K1 this method will produce an output schema that is K1, K2
    * despite the way that the keys were ordered in the projection) - see
@@ -60,7 +60,7 @@ public final class SelectionUtil {
    * But we cannot simply select all the keys and then the values, we must maintain
    * the interleaving of key and values because transient queries return all columns
    * to the user as "value columns". If someone issues a SELECT VALUE, * FROM FOO
-   * it is expected that X shows up _before_ the key fields. This means we need to
+   * it is expected that VALUE shows up _before_ the key fields. This means we need to
    * reorder the key columns within the list of projections without affecting the
    * relative order the keys/values.
    *
@@ -117,12 +117,12 @@ public final class SelectionUtil {
     int currKeyIdx = 0;
     for (final SelectExpression select : projection) {
       if (keyExpressionMembership.contains(select)) {
+        while (keyExpressions.get(currKeyIdx).isEmpty()) {
+          currKeyIdx++;
+        }
         final SelectExpression keyExp = keyExpressions.get(currKeyIdx).remove(0);
         final SqlType type = expressionTypeManager.getExpressionSqlType(keyExp.getExpression());
         builder.keyColumn(keyExp.getAlias(), type);
-        if (keyExpressions.get(currKeyIdx).isEmpty()) {
-          currKeyIdx++;
-        }
       } else {
         final Expression expression = select.getExpression();
         final SqlType type = expressionTypeManager.getExpressionSqlType(expression);
--- a/ksqldb-engine/src/main/java/io/confluent/ksql/planner/plan/SelectionUtil.java
+++ b/ksqldb-engine/src/main/java/io/confluent/ksql/planner/plan/SelectionUtil.java
@@ -15,7 +15,6 @@
 
 package io.confluent.ksql.planner.plan;
 
-import com.google.common.collect.ImmutableMap;
 import io.confluent.ksql.execution.expression.tree.ColumnReferenceExp;
 import io.confluent.ksql.execution.expression.tree.Expression;
 import io.confluent.ksql.execution.expression.tree.UnqualifiedColumnReferenceExp;
@@ -27,11 +26,15 @@ import io.confluent.ksql.parser.tree.AllColumns;
 import io.confluent.ksql.parser.tree.SelectItem;
 import io.confluent.ksql.parser.tree.SingleColumn;
 import io.confluent.ksql.schema.ksql.Column;
+import io.confluent.ksql.schema.ksql.Column.Namespace;
 import io.confluent.ksql.schema.ksql.LogicalSchema;
 import io.confluent.ksql.schema.ksql.LogicalSchema.Builder;
 import io.confluent.ksql.schema.ksql.types.SqlType;
+import java.util.ArrayList;
+import java.util.HashSet;
 import java.util.List;
 import java.util.Optional;
+import java.util.Set;
 import java.util.stream.Collectors;
 import java.util.stream.IntStream;
 import java.util.stream.Stream;
@@ -41,6 +44,39 @@ public final class SelectionUtil {
   private SelectionUtil() {
   }
 
+  /*
+   * The algorithm behind this method feels unnecessarily complicated and is begging
+   * for someone to come along and improve it, but until that time here is
+   * a description of what's going on.
+   *
+   * Essentially, we need to build a logical schema that mirrors the physical
+   * schema until https://github.com/confluentinc/ksql/issues/6374 is addressed.
+   * That means that the keys must be ordered in the same was as the parent schema
+   * (e.g. if the source schema was K1 INT KEY, K2 INT KEY and the projection is
+   * SELECT K2, K1 this method will produce an output schema that is K1, K2
+   * despite the way that the keys were ordered in the projection) - see
+   * https://github.com/confluentinc/ksql/pull/7477 for context on the bug.
+   *
+   * But we cannot simply select all the keys and then the values, we must maintain
+   * the interleaving of key and values because transient queries return all columns
+   * to the user as "value columns". If someone issues a SELECT VALUE, * FROM FOO
+   * it is expected that X shows up _before_ the key fields. This means we need to
+   * reorder the key columns within the list of projections without affecting the
+   * relative order the keys/values.
+   *
+   * To spice things up even further, there's the possibility that the same key is
+   * aliased multiple times (SELECT K1 AS X, K2 AS Y FROM ...), which is not supported
+   * but is verified later when building the final projection - so we maintain it here.
+   *
+   * Now on to the algorithm itself: we make two passes through the list of projections.
+   * The first pass builds a mapping from source key to all the projections for that key.
+   * We will use this mapping to sort the keys in the second pass. This mapping is two
+   * dimensional to address the possibility of the same key with multiple aliases.
+   *
+   * The second pass goes through the list of projections again and builds the logical schema,
+   * but this time if we encounter a projection that references a key column, we instead take
+   * it from the list we built in the first pass (in order defined by the parent schema).
+   */
   public static LogicalSchema buildProjectionSchema(
       final LogicalSchema parentSchema,
       final List<SelectExpression> projection,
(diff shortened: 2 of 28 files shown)


ksqldb-engine/src/main/java/io/confluent/ksql/planner/plan/PlanNode.java
```diff
@@ -192,10 +195,13 @@ static void throwKeysNotIncludedError(
       final LogicalSchema schema
   ) {
     // When doing a `select *` key columns should be at the front of the column list
-    // but are added at the back during processing for performance reasons.
-    // Switch them around here:
-    final Stream<Column> keys = columns.stream()
-        .filter(c -> schema.isKeyColumn(c.name()));
+    // but are added at the back during processing for performance reasons. Furthermore,
```
mjsax, line 198: For my own education. What does this comment mean: `but are added at the back during processing for performance reasons` -- what is the perf impact and why?

ksqldb-engine/src/main/java/io/confluent/ksql/planner/plan/PlanNode.java
```diff
@@ -192,10 +195,13 @@ static void throwKeysNotIncludedError(
...
+    // they should be selected in the order of the key, if that is not the same as the
+    // ordering within the value. Switch them around here:
```
mjsax, line 200: `the ordering within the value` sound confusing.

ksqldb-engine/src/main/java/io/confluent/ksql/planner/plan/PlanNode.java
```diff
@@ -192,10 +195,13 @@ static void throwKeysNotIncludedError(
       final LogicalSchema schema
```
mjsax, line 195: To better understand the PR. What is `columns` and what is `schema` exactly? Is `columns` the list of selected columns from the `SELECT` clause (in the corresponding order) ? Is `schema` the "physical" layout of the input (what is `schema` for a join query?)

ksqldb-engine/src/main/java/io/confluent/ksql/planner/plan/PlanNode.java
```diff
@@ -192,10 +195,13 @@ static void throwKeysNotIncludedError(
...
     // When doing a `select *` key columns should be at the front of the column list
-    // but are added at the back during processing for performance reasons.
-    // Switch them around here:
-    final Stream<Column> keys = columns.stream()
-        .filter(c -> schema.isKeyColumn(c.name()));
+    // but are added at the back during processing for performance reasons. Furthermore,
+    // they should be selected in the order of the key, if that is not the same as the
+    // ordering within the value. Switch them around here:
+    final ImmutableMap<ColumnName, Column> columnsByName = Maps.uniqueIndex(columns, Column::name);
+    final Stream<Column> keys = schema.key().stream()
```
mjsax, line 202: The main change (for the fix) seems to be to use `schema` instead of `columns` to find the keys?

ksqldb-engine/src/main/java/io/confluent/ksql/planner/plan/PlanNode.java
```diff
@@ -192,10 +195,13 @@ static void throwKeysNotIncludedError(
...
+        .map(key -> columnsByName.get(key.name()))
+        .filter(Objects::nonNull);
```
mjsax, line 204: Wondering why we need this filter?

ksqldb-engine/src/main/java/io/confluent/ksql/planner/plan/SelectionUtil.java
```diff
@@ -65,13 +71,27 @@ public static LogicalSchema buildProjectionSchema(
...
       } else {
         builder.valueColumn(select.getAlias(), expressionType);
       }
     }
 
+    // keys need to be added in the same order as the source
+    for (final Column key : parentSchema.key()) {
+      final List<Pair<ColumnName, SqlType>> aliasAndTypes = sourceKeyToAliasAndType.get(key.name());
+      for (final Pair<ColumnName, SqlType> aliasAndType : aliasAndTypes) {
+        if (aliasAndType != null) {
```
mjsax, line 87: How could `aliasAndType` be `null` ?

ksqldb-engine/src/main/java/io/confluent/ksql/planner/plan/PlanNode.java
agavra: I think the context here is that it avoids shuffling data (adding at the end of an array is easier than shuffling the whole array)

ksqldb-engine/src/main/java/io/confluent/ksql/planner/plan/SelectionUtil.java
```diff
@@ -65,13 +71,27 @@ public static LogicalSchema buildProjectionSchema(
...
+          // can be null if the key was not selected - this will cause a failure
```
mjsax, line 88: Do you refer to `aliasAndType` -- I don't see how it could be `null`? If could be that the `Pair` elements are `null` though?

I think we can remove the ` if (aliasAndType != null) {` check?

ksqldb-engine/src/main/java/io/confluent/ksql/planner/plan/SelectionUtil.java
```diff
@@ -65,13 +71,27 @@ public static LogicalSchema buildProjectionSchema(
...
+          // down the line but here we just ignore it here
```
mjsax, line 89: Why can it case an error? And how would it surface to the user? -- Not sure what we don't want to catch this error right here?

ksqldb-engine/src/main/java/io/confluent/ksql/planner/plan/SelectionUtil.java
```diff
@@ -53,7 +55,11 @@ public static LogicalSchema buildProjectionSchema(
 
     final Builder builder = LogicalSchema.builder();
 
-    final ImmutableMap.Builder<ColumnName, SqlType> keys = ImmutableMap.builder();
+    // it is invalid to have the source key aliased multiple times in the value,
+    // but we will throw a better error at the end when we call builder.keyColumn
```
mjsax, line 59: `aliased multiple times in the value` and `we call builder.keyColumn`: do you mean `builder.valueColumn`?

ksqldb-engine/src/main/java/io/confluent/ksql/planner/plan/PlanNode.java
agavra: yeah I read that back and totally agree - I'll reword

ksqldb-functional-tests/src/test/resources/query-validation-tests/partition-by.json
```diff
@@ -1050,6 +1050,30 @@
         ]
       }
     },
+    {
+      "name": "multiple columns - select star - reorder columns",
+      "statements": [
+        "CREATE STREAM INPUT (NAME STRING, ID INT, AGE INT) with (kafka_topic='input', format='JSON');",
+        "CREATE STREAM OUTPUT AS select * from INPUT partition by AGE, ID;"
```
mjsax, line 1057: meta question: The lower/upper case style is confusing to me.

Should we not upper case all keywords, and lower case all identifiers? This is all a wild mix making it hard to read.

```
CREATE STREAM input (name STRING, id INT, age INT) with (kafka_topic='input', format='JSON');
CREATE STREAM output AS SELECT * FROM input PARTITION BY age, id;
```

ksqldb-engine/src/main/java/io/confluent/ksql/planner/plan/PlanNode.java
agavra: honestly, this code is super weird to me; I was thinking the same thing, but erred against changing it. It's only ever called like this: `orderColumns(getSchema().value(), getSchema());` so really, it could be done by just passing in one  schema and then ensuring it's properly ordered.

ksqldb-engine/src/main/java/io/confluent/ksql/planner/plan/PlanNode.java
agavra: spot on, that's the first fix (the second fix is in the `SelectionUtil`

ksqldb-engine/src/main/java/io/confluent/ksql/planner/plan/PlanNode.java
agavra: um, we might not need this one here after digging into it more - but from the API of this method, there's nothing prevent `columns` from not covering everything in `schema`, which would mean `columnsByName.get(key.name())` would return empty. Without refactoring this method signature I think it's better to be defensive here.

ksqldb-engine/src/main/java/io/confluent/ksql/planner/plan/SelectionUtil.java
agavra: no, I meant `aliased multiple times in the sink`. I'll fix that

ksqldb-engine/src/main/java/io/confluent/ksql/planner/plan/SelectionUtil.java
agavra: you're required to select all keys, if you don't it'll complain :) I don't think we should add unnecessary redundant checks in places they don't belong. there's no reason why this specific code needs to know that you're required to select all keys. The code that requires that should enforce it.

ksqldb-functional-tests/src/test/resources/query-validation-tests/partition-by.json
mjsax: Can we replicate this test (or can you point out existing tests) for:
 - input stream with existing multi-column key plus reordering: `PARTITION BY key2,key1`
 - `PARTITION BY key, nonkey`
 - `PARTITION BY nonKey, key`
 - all of the above replicated `SELECT columnNames` (instead of `*`) with all 4 reordering combination (either in SELECT or PARTITION BY); seems two are already covered
 - mixing key and values in select: `SELECT key2, nonKey2, key1, nonKey1...`

ksqldb-engine/src/main/java/io/confluent/ksql/planner/plan/SelectionUtil.java
agavra: big oops! I did a last minute change here and meant to have the null check on `aliasAndTypes`. It should fail the tests with an NPE

agavra: actually, now that i've changed to `MultiMap` it'll never return `null`, just an empty list :) so we're all good to remove the null check

ksqldb-functional-tests/src/test/resources/query-validation-tests/partition-by.json
agavra: * `"name": "multiple key columns - reordered",` already exists
* `"name": "multiple columns - some key some value"` already exists
* I'll add that
* `multiple key columns` tests this
* the one I mentioned above tests this as well

As far as the casing style I just copied existing tests - happy to change the new ones but I'm not going to go and change the old ones for this PR

ksqldb-engine/src/main/java/io/confluent/ksql/planner/plan/SelectionUtil.java
```diff
@@ -41,6 +44,39 @@
...
   }
 
+  /*
+   * The algorithm behind this method feels unnecessarily complicated and is begging
+   * for someone to come along and improve it, but until that time here is
+   * a description of what's going on.
+   *
+   * Essentially, we need to build a logical schema that mirrors the physical
+   * schema until https://github.com/confluentinc/ksql/issues/6374 is addressed.
+   * That means that the keys must be ordered in the same was as the parent schema
```
vcrfxia, line 54:
```suggestion
   * That means that the keys must be ordered in the same way as the parent schema
```

ksqldb-engine/src/main/java/io/confluent/ksql/planner/plan/SelectionUtil.java
```diff
@@ -41,6 +44,39 @@
...
+   * (e.g. if the source schema was K1 INT KEY, K2 INT KEY and the projection is
+   * SELECT K2, K1 this method will produce an output schema that is K1, K2
+   * despite the way that the keys were ordered in the projection) - see
+   * https://github.com/confluentinc/ksql/pull/7477 for context on the bug.
+   *
+   * But we cannot simply select all the keys and then the values, we must maintain
+   * the interleaving of key and values because transient queries return all columns
+   * to the user as "value columns". If someone issues a SELECT VALUE, * FROM FOO
+   * it is expected that X shows up _before_ the key fields. This means we need to
```
vcrfxia, line 63:
```suggestion
   * it is expected that VALUE shows up _before_ the key fields. This means we need to
```

ksqldb-engine/src/main/java/io/confluent/ksql/planner/plan/SelectionUtil.java
```diff
@@ -41,6 +44,39 @@
...
+   *
+   * Now on to the algorithm itself: we make two passes through the list of projections.
+   * The first pass builds a mapping from source key to all the projections for that key.
+   * We will use this mapping to sort the keys in the second pass. This mapping is two
+   * dimensional to address the possibility of the same key with multiple aliases.
+   *
+   * The second pass goes through the list of projections again and builds the logical schema,
+   * but this time if we encounter a projection that references a key column, we instead take
+   * it from the list we built in the first pass (in order defined by the parent schema).
+   */
```
vcrfxia, line 79: Thanks for the detailed comment! Super helpful.

ksqldb-engine/src/main/java/io/confluent/ksql/planner/plan/SelectionUtil.java
```diff
@@ -51,24 +87,46 @@ public static LogicalSchema buildProjectionSchema(
...
+    // second pass, which iterates the projections but ignores any key expressions,
+    // instead taking them from the ordered keyExpressions list
+    final Builder builder = LogicalSchema.builder();
+    int currKeyIdx = 0;
+    for (final SelectExpression select : projection) {
+      if (keyExpressionMembership.contains(select)) {
+        final SelectExpression keyExp = keyExpressions.get(currKeyIdx).remove(0);
+        final SqlType type = expressionTypeManager.getExpressionSqlType(keyExp.getExpression());
+        builder.keyColumn(keyExp.getAlias(), type);
+        if (keyExpressions.get(currKeyIdx).isEmpty()) {
```
vcrfxia, line 123: What happens if a key isn't selected? (Should this be a while-loop rather than a simple if-check?  If so, we also need to update `currKeyIdx = 0;` on line 117.)

ksqldb-functional-tests/src/test/resources/query-validation-tests/partition-by.json
```diff
@@ -1100,6 +1172,110 @@
...
+        {"topic": "OUTPUT", "key": {"ID": 10, "NAME": null}, "value": {"AGE": 30}},
+        {"topic": "OUTPUT", "key": {"ID": null, "NAME": "bob"}, "value": null},
+        {"topic": "OUTPUT", "key": {"ID": null, "NAME": null}, "value": null}
+      ],
+      "post": {
+        "sources": [
+          {"name": "OUTPUT", "type": "stream", "schema": "NAME STRING KEY, ID INT KEY, AGE INT"}
+        ]
+      }
+    },
```
vcrfxia, line 1278: Thanks for the test coverage! This is great. Do we also need to audit the test coverage for GROUP BY?

vcrfxia (approved):
Thanks @agavra and @mjsax for fixing (and finding) this bug! Are we planning to get this into 0.18 as well?

ksqldb-engine/src/main/java/io/confluent/ksql/planner/plan/SelectionUtil.java
agavra: good call - made this a while loop and moved it up (before the remove) so no need to change line 117

ksqldb-functional-tests/src/test/resources/query-validation-tests/partition-by.json
agavra: We probably should, though since `GROUP BY` doesn't use the same code we should do that separately (Matthias tested it locally and GROUP BYs have this "rearrange" behavior)

ksqldb-engine/src/main/java/io/confluent/ksql/planner/plan/PlanNode.java
guozhangwang: Reading further below, I feel this logic need to be kept as-is even when we have https://github.com/confluentinc/ksql/issues/6374 in place with a clear separation (I'm assuming `schema` references the physical schema of the source, and `columns` references the logical schema mapped from the phyiscal schema by then).

guozhangwang (approved):
The new algorithm looks good to me :) I think for https://github.com/confluentinc/ksql/issues/6374 itself we probably need to rely on this two-phase procedure anyways to separate logical `columns` from phyiscal `schema`, along with the maintained mapping between the two.

ksqldb-engine/src/main/java/io/confluent/ksql/planner/plan/PlanNode.java
mjsax: > I think it's better to be defensive here.

I guess that is the question? We know that it should never be `null`, so should not fail fast to expose a bug?

ksqldb-functional-tests/src/test/resources/query-validation-tests/partition-by.json
mjsax: > 'm not going to go and change the old ones for this PR

Sure :) (did not expect that)

pom.xml
```diff
@@ -136,7 +136,7 @@
         <scala.version>2.13.2</scala.version>
         <apache.io.version>2.6</apache.io.version>
         <io.confluent.ksql.version>6.2.0-0</io.confluent.ksql.version>
-        <io.confluent.schema-registry.version>${confluent.version.range}</io.confluent.schema-registry.version>
+        <io.confluent.schema-registry.version>6.2.0-685</io.confluent.schema-registry.version>
```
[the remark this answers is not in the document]
agavra: oops! I didn't mean to check that in... but without this things take way too long locally...

PR #7482, opened 2021-05-06
feat: bin packed queries POC [WIP]

wcarlson5:
Idea for bin packing in ksql

### Reviewer checklist


ksqldb-engine/src/main/java/io/confluent/ksql/util/BinPackedStreamsMetadata.java
```diff
@@ -0,0 +1,185 @@
...
+    kafkaStreams.addTopology(queryMap.get(queryId.toString()));
+  }
+
+  public List<QueryError> getQueryErrors(final QueryId queryId) {
+    return queryErrors.toImmutableList();
+  }
+
+  public Map<String, Map<Integer, LagInfo>> allLocalStorePartitionLags(final QueryId queryId) {
+    try {
+      return kafkaStreams.allLocalStorePartitionLags();
```
wcarlson5, line 171: Maybe I can filter for the queries but am not sure

ksqldb-engine/src/main/java/io/confluent/ksql/util/BinPackedStreamsMetadata.java
```diff
@@ -0,0 +1,185 @@
...
+    for (Topology topology: queryMap.values()) {
+      newKafkaStreams.addTopology(topology);
+    }
+    newKafkaStreams.start();
+    kafkaStreams.close();
+    kafkaStreams = newKafkaStreams;
+  }
+
+  public boolean isError(final QueryId queryId) {
+    return !queryErrors.toImmutableList().isEmpty();
```
wcarlson5, line 153: Even now the old checks if the query is in error state, will fix

wcarlson5 (commented):
@ableegoldman @rodesai @guozhangwang I took a stab at bin packing queries. This is about what I think it should look like. let me know your thoughts. There are obviously a lot of open questions still

ksqldb-engine/src/main/java/io/confluent/ksql/util/BinPackedStreamsMetadata.java
```diff
@@ -0,0 +1,199 @@
...
+                       .collect(Collectors.toSet());
+  }
+
+  public void addToplogy(final Topology topology) {
+    queryMap.put(topology.getName(), topology);
+    kafkaStreams.addTopology(topology);
+  }
+
+  public void stop(final QueryId queryId) {
+    queryMap.remove(queryId.toString());
```
guozhangwang, line 141: I feel that interchangingly use `queryId.toString()` and `topology.getName()` may confused future readers that do not know we always treat them the same. Maybe just stick with one and even better just have one map. See my other comment above.

ksqldb-engine/src/main/java/io/confluent/ksql/util/BinPackedStreamsMetadata.java
```diff
@@ -0,0 +1,199 @@
...
+public class BinPackedStreamsMetadata {
+
+  private static final Logger LOG = LoggerFactory.getLogger(BinPackedStreamsMetadata.class);
+
+  private final KafkaStreamsBuilder kafkaStreamsBuilder;
+  private QueryErrorClassifier errorClassifier ;
+  private KafkaStreams kafkaStreams;
+  private ImmutableMap<String, Object> streamsProperties;
+  private final QueryMetadataImpl.TimeBoundedQueue queryErrors;
+  private final Map<String, Topology> queryMap;
```
guozhangwang, line 54: Could we just maintain a single map with a consolidated class here? E.g. if we have a `BinPackedQuery` which contains `QueryId, Topology, BinPackedPersistentQueryMetadataImpl`. Then we can just maintain a single map here which is `<QueryId, BinPackedQuery>`. Then in the `BinPackedPersistentQueryMetadataImpl` above it only need to remember its `QueryId` and call `startQuery(queryId)` upon initialization.

ksqldb-engine/src/main/java/io/confluent/ksql/util/BinPackedStreamsMetadata.java
```diff
@@ -0,0 +1,199 @@
...
+
+  public Set<StreamsTaskMetadata> getTaskMetadata() {
+    return kafkaStreams.localThreadsMetadata()
+                       .stream()
+                       .flatMap(t -> t.activeTasks().stream())
+                       .map(StreamsTaskMetadata::fromStreamsTaskMetadata)
+                       .collect(Collectors.toSet());
+  }
+
+  public void addToplogy(final Topology topology) {
```
guozhangwang, line 135: Maybe we can just call this `start(queryId)`?

ksqldb-engine/src/main/java/io/confluent/ksql/util/BinPackedPersistentQueryMetadataImpl.java
```diff
@@ -0,0 +1,322 @@
...
+  private final Set<SourceName> sourceNames;
+  private final LogicalSchema logicalSchema;
+  private final QueryId queryId;
+  private final DataSource sinkDataSource;
+  private final ProcessingLogger processingLogger;
+  private final ExecutionStep<?> physicalPlan;
+  private final PhysicalSchema resultSchema;
+  private final Listener listener;
+
+  private static final Ticker CURRENT_TIME_MILLIS_TICKER = new Ticker() {
```
guozhangwang, line 75: Seems not used?

ksqldb-engine/src/main/java/io/confluent/ksql/util/BinPackedStreamsMetadata.java
guozhangwang: My previous comment may be a bis unclear, I'm just wondering why we cannot just call the other

```
public void start(final QueryId queryId) {
    kafkaStreams.addTopology(queryMap.get(queryId.toString()));
  }
```

whenever we need to call this? again I'm feeling that with a single consolidated map we should just need to `start/addTopology` once, together.

ksqldb-engine/src/main/java/io/confluent/ksql/util/BinPackedPersistentQueryMetadataImpl.java
```diff
@@ -0,0 +1,322 @@
...
+
+  private static final Logger LOG = LoggerFactory
+      .getLogger(BinPackedPersistentQueryMetadataImpl.class);
+
+  private final String statementString;
+  private final String executionPlan;
+  private final String queryApplicationId;
+  private final Topology topology;
+  private final BinPackedStreamsMetadata binPackedStreamsMetadata;
+  private final QuerySchemas schemas;
```
guozhangwang, line 64: For my education: what's the difference between QuerySchemas and LogicalSchema here?

ksqldb-engine/src/main/java/io/confluent/ksql/util/BinPackedStreamsMetadata.java
```diff
@@ -0,0 +1,199 @@
...
+  protected StreamsUncaughtExceptionHandler.StreamThreadExceptionResponse uncaughtHandler(
+      final Throwable e
+                                                                                         ) {
+    QueryError.Type errorType = QueryError.Type.UNKNOWN;
+    try {
+      errorType = errorClassifier.classify(e);
+    } catch (final Exception classificationException) {
+      LOG.error("Error classifying unhandled exception", classificationException);
+    } finally {
+      // If error classification throws then we consider the error to be an UNKNOWN error.
```
guozhangwang, line 90: A bit orthogonal to binpacking but more for error handling: when the thrown error is a runtime IllegalState/ArgumentException, what error type do we classify to? And do we still handle as `REPLACE_THREAD`?

ksqldb-engine/src/main/java/io/confluent/ksql/util/BinPackedStreamsMetadata.java
```diff
@@ -0,0 +1,199 @@
...
+
+  public void stop(final QueryId queryId) {
+    queryMap.remove(queryId.toString());
+    metadata.remove(queryId);
+    listeners.remove(queryId);
+    kafkaStreams.removeTopology(queryId.toString());
+  }
+
+  public void restart(final QueryId queryId) {
+    final KafkaStreams newKafkaStreams = kafkaStreamsBuilder.build(null, streamsProperties);
```
guozhangwang, line 148: Hmm.. do we really not care which `queryId` to restart, or is it just the scope of V1? I thought we would also leverage `removeTopology`, i.e. `removeTopology` and then `addTopology`, right?

ksqldb-engine/src/main/java/io/confluent/ksql/util/BinPackedStreamsMetadata.java
guozhangwang: Yeah I think we should filter for the queries --- from the queryId we get the named topology and all its contained taskIds, and filter based on the taskID -> partitions.

ksqldb-engine/src/main/java/io/confluent/ksql/util/BinPackedPersistentQueryMetadataImpl.java
wcarlson5: It is not

ksqldb-engine/src/main/java/io/confluent/ksql/util/BinPackedStreamsMetadata.java
wcarlson5: I  can probably make this simpler, and add some more getters to the `BinPackedPersistentQueryMetadataImpl` to fill in the gaps.

ksqldb-engine/src/main/java/io/confluent/ksql/util/BinPackedStreamsMetadata.java
wcarlson5: ah that might work, if you can't tell I added the consolidated map after I had designed it and haven't worked out all the ways I can use it best yet :)

ksqldb-engine/src/main/java/io/confluent/ksql/util/BinPackedStreamsMetadata.java
wcarlson5: Right now we always replace the thread. With bin packing we might able to remove the problematic topology on non recoverable errors once we have classification on that axis

ksqldb-engine/src/main/java/io/confluent/ksql/util/BinPackedStreamsMetadata.java
wcarlson5: They should both be the same but I see what you are saying, if I expand the map I might be able to only use the queryID string as the key and that should simplify things

ksqldb-engine/src/main/java/io/confluent/ksql/util/BinPackedStreamsMetadata.java
wcarlson5: I think the purpose of restart would be to get a clean `kafkaStreams` environment. There should be defiantly be a way to do that. We can also do what you are describing by jsut stoping then starting the query

ksqldb-engine/src/main/java/io/confluent/ksql/util/BinPackedStreamsMetadata.java
wcarlson5: We will have to add the tasks to the queryId I think but that might be something we do later

wcarlson5 (commented):
@guozhangwang thanks for the review! I will make some changes for our sync on Tuesday.

But do you think it's fair to say that this is a variable direction? it seemed most of your concerns were implementation based not design issues.

ksqldb-engine/src/main/java/io/confluent/ksql/util/BinPackedStreamsMetadata.java
guozhangwang: Sounds good.

We had some discussions on AK regarding these types of exceptions, since they indicate some bugs in the KS code. In the future maybe we can improve KS to not pass this to exceptional handler at all and just fail-stop immediately, but for now we'd need KSQL layer to handle them as fatal.

guozhangwang:
Yeah on the meta level I think this is a viable direction :)

ksqldb-engine/src/main/java/io/confluent/ksql/util/BinPackedStreamsMetadata.java
```diff
@@ -0,0 +1,199 @@
...
+import java.time.Duration;
+import java.util.Collection;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.Objects;
+import java.util.Set;
+import java.util.stream.Collectors;
+
+public class BinPackedStreamsMetadata {
```
rodesai, line 45: nit: rename something like `SharedKafkaStreamsRuntime` or `KafkaStreamsQueryRuntime`. bin-packed is more about the top-level algorithm for assigning queries to these - eventually we don't want to do bin-packing and just run everything in one runtime. And metadata is just a weird word we're using in `PersistentQueryMetadata`/`TransientQueryMetadata` - those objects are literally the query entities in our model, not really metadata. We should rename those too, but no need to retain the weird convention.

ksqldb-engine/src/main/java/io/confluent/ksql/query/QueryExecutor.java
```diff
@@ -261,30 +274,70 @@ PersistentQueryMetadata buildPersistentQuery(
...
+          ksqlConfig.getLong(KsqlConfig.KSQL_QUERY_RETRY_BACKOFF_INITIAL_MS),
+          ksqlConfig.getLong(KsqlConfig.KSQL_QUERY_RETRY_BACKOFF_MAX_MS),
+          listener,
+          scalablePushRegistry
+      );
+    }
+  }
+
+  private BinPackedStreamsMetadata getStream(final Set<SourceName> sources) {
+    for (BinPackedStreamsMetadata binPackedStreamsMetadata: streams) {
```
rodesai, line 330: I don't think we should be computing this here. It's a little brittle - if we change the selection algorithm or change the data structure we use to store the runtimes so that the traversal is in the wrong order we might pick different runtimes across restarts. Instead, I think we should compute the target runtime when building `QueryPlan` (see `EngineExecutor.plan`) and then we can store its name in the final plan that gets serialized to the command topic.

ksqldb-engine/src/main/java/io/confluent/ksql/util/BinPackedStreamsMetadata.java
```diff
@@ -0,0 +1,199 @@
...
+
+  public Map<String, Object> getStreamProperties() {
+    return streamsProperties;
+  }
+
+  public List<QueryError> getQueryErrors() {
+    return queryErrors.toImmutableList();
+  }
+
+  public Set<SourceName> getMetadata() {
```
rodesai, line 191: why the name `getMetadata`? maybe `getSources` or `getSourceTopics`?

ksqldb-engine/src/main/java/io/confluent/ksql/util/BinPackedStreamsMetadata.java
```diff
@@ -0,0 +1,199 @@
...
+    queryMap = new HashMap<>();
+    metadata = new HashMap<>();
+    listeners = new HashMap<>();
+  }
+
+  public void addQuery(final QueryErrorClassifier errorClassifier,
+                       final Map<String, Object> streamsProperties,
+                       final BinPackedPersistentQueryMetadataImpl binPackedPersistentQueryMetadata,
+                       final QueryId queryId) {
+    this.errorClassifier = errorClassifier;
```
rodesai, line 73: would be good to log the mapping here

ksqldb-engine/src/main/java/io/confluent/ksql/util/BinPackedStreamsMetadata.java
```diff
@@ -0,0 +1,199 @@
...
+    listeners.remove(queryId);
+    kafkaStreams.removeTopology(queryId.toString());
+  }
+
+  public void restart(final QueryId queryId) {
+    final KafkaStreams newKafkaStreams = kafkaStreamsBuilder.build(null, streamsProperties);
+    for (Topology topology: queryMap.values()) {
+      newKafkaStreams.addTopology(topology);
+    }
+    newKafkaStreams.start();
```
rodesai, line 152: would it make sense to reverse the order here? especially if someone has static membership configured

rodesai:
@wcarlson5 should we designate this wip or a draft? before we merge this we should at least have the feature flag and unit+integration tests

ksqldb-engine/src/main/java/io/confluent/ksql/util/BinPackedPersistentQueryMetadataImpl.java
wcarlson5: that is a good question. the logical schema is used by the abstract class `QueryMetadata` but this class is doing that work as well. I think I might be able to simplify here a bit

ksqldb-engine/src/main/java/io/confluent/ksql/query/QueryExecutor.java
```diff
@@ -261,30 +274,69 @@ PersistentQueryMetadata buildPersistentQuery(
...
+          getUncaughtExceptionProcessingLogger(queryId),
+          ksqlConfig.getLong(KsqlConfig.KSQL_QUERY_RETRY_BACKOFF_INITIAL_MS),
+          ksqlConfig.getLong(KsqlConfig.KSQL_QUERY_RETRY_BACKOFF_MAX_MS),
+          listener,
+          scalablePushRegistry
+      );
+    }
+  }
+
+  private SharedKafkaStreamsRuntime getStream(final Set<SourceName> sources) {
```
wcarlson5, line 328: @rodesai I agree that we should move this. I don't think we need to do that right now though. I would like to keep it as simple as possible and I think once we get it all in one runtime we will not even need this. If there is an intermediate step I propose we move it then

wcarlson5 (commented):
@guozhangwang @rodesai Thanks for the review I think I got to all your comments. I will do a bit of renaming in another PR soon but I am going to start testing with local builds

ksqldb-engine/src/main/java/io/confluent/ksql/util/BinPackedStreamsMetadata.java
ableegoldman: I may not have added it in the initial skeleton API, but it's in the Pt. 1 PR. I'm hoping we can have that merged soon. But yeah, for restarted queries we should use the collection-based API rather than adding them one-at-a-time

guozhangwang:
@rodesai has a good point that we should push the decision on query assignment earlier at `QueryPlan` time before writing to the cmd topic, instead of doing it until `QueryExecutor`. Admittedly would be a relatively major refactoring as we need the query planner to remember the current assignment, and to make deterministic decisions, but I think that is worthy the complexity.

wcarlson5:
@guozhangwang I agree that is probably a good idea eventually, I am just not sure that it is a good idea to do right now. I think that maybe we should put that in V1 instead of V0. Since we are not writing to the cmd topic we don't have to worry about compatibility yet, If we were moving the other way I would be more onboard with doing it now. 

 I also don't want to complicate this PR more than necessary. I think that this the minimum needed to get it working for V0. As it is it would be quite easy to strip this implementation out and replace it for a later version.

ableegoldman:
Yeah I agree with Walker that we should punt on that for this PR as our current goal is just to get the POC up and soaking, not fully optimized. Though I also agree that's something we should consider as followup work -- @wcarlson5 can you file a ticket under the V1 milestone for this?

wcarlson5:
done https://confluentinc.atlassian.net/browse/KCI-629 :)

guozhangwang:
Would anything from V0 be released in a near-term version? If yes do we have a smooth upgrade path in mind when such information was only later added to the cmd topic?

ksqldb-engine/src/main/java/io/confluent/ksql/util/PersistentQueriesInSharedRuntimesImpl.java
```diff
@@ -0,0 +1,320 @@
...
+
+  @Override
+  public Optional<Materialization> getMaterialization(
+      final QueryId queryId,
+      final QueryContext.Stacker contextStacker) {
+    return null;
+  }
+
+  @Override
+  public void restart() {
```
guozhangwang, line 175: Just FYI, in https://github.com/confluentinc/ksql/pull/7627/files we removed the `restart()` func -- I think this is not required for binpacking either but in case it does, @wcarlson5 please raise it to the other side for attention.

wcarlson5: I don't think its required, but it is the only way to get a fresh streams object with bin-packing so it might be nice to have. I am fine either way but @rodesai might have a perfernce.

wcarlson5:
Closing in favor of https://github.com/confluentinc/ksql/pull/7721

PR #7483, opened 2021-05-06
fix: change the isError to not use the state

wcarlson5:
since the state ERROR changed definitions the current isError is no longer really relevant. However it actually seems that isError is not used so I will just remove it._

### Reviewer checklist

### DIFF
--- a/ksqldb-engine/src/main/java/io/confluent/ksql/util/QueryMetadata.java
+++ b/ksqldb-engine/src/main/java/io/confluent/ksql/util/QueryMetadata.java
@@ -43,8 +43,6 @@ public interface QueryMetadata {
 
   KafkaStreams.State getState();
 
-  boolean isError();
-
   String getExecutionPlan();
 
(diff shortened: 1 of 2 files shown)


ableegoldman (approved):
Makes sense to me

wcarlson5:
@guozhangwang since the state ERROR changed definitions the current isError is no longer really relevant. However it actually seems that isError is not used so I will just remove it

ableegoldman:
+1 on putting what you just said in the PR description. Filling out the description just makes it easier to understand what the PR is about at a glance, which is useful eg if we ever have to skim over PRs looking for possible regressions. It's a good idea to always do, even if the changes in the PR are extremely obvious to you

wcarlson5:
good call. TBH I was going to check with some more testing to see if I needed to remove it (which I can). I probably should have made it as a draft PR first. But thanks for the quick reviews!

PR #7484, opened 2021-05-06
docs: intent for KLIP-48: Schemas

derekjn:
Intent to add KLIP-48: Schemas


agavra:
Very cryptic. Excited for it :)

PR #7485, opened 2021-05-06
docs: KLIP-48: Schemas

derekjn:
Discussion for KLIP-48, which proposes the introduction of a schemas (namespace) abstraction.

### DIFF
--- a/design-proposals/klip-48-schemas.md
+++ b/design-proposals/klip-48-schemas.md
@@ -1,0 +1,115 @@
+# KLIP-41: Schemas
+
+**Authors**: @derekjn
+**Release Target**: N/A |
+**Status**: _In Discussion_ |
+**Discussion**: --
+
+## Motivation and background
+
+Schemas are a widely employed abstraction across most SQL-based systems. They ultimately provide a layer of hierarchy underneath which logically related groups of objects (i.e. tables, streams, types, etc.) may be organized. In other words, **schemas are SQL’s notion of a namespace**. Similarly to namespaces in other contexts, SQL schemas introduce a level of indirection preceding a given object’s identifier. For example, if an object `my_table` is defined within the schema `my_schema`, the fully qualified object name becomes `my_schema.my_table`
+
+In addition to providing a means to logically organize groups of database objects, schemas also provide basic operational benefits. The most obvious of these benefits is the ability to drop an entire schema in a manner that recursively drops all objects underneath that schema. It is easy to imagine additional benefits that schemas have the potential to provide for ksqlDB users longer term: performing basic operational tasks such as starting and stopping groups of persistent queries, applying configuration properties to an entire schema of objects, and more.
+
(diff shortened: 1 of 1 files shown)


design-proposals/klip-48-schemas.md
```diff
@@ -0,0 +1,115 @@
...
+
+ksqlDB schemas will be primarily leveraged via SQL syntax, with some optional configuration. Each of the new syntax constructs associated with ksqlDB schemas are proposed as follows:
+
+### Creating schemas
+
+```sql
+CREATE SCHEMA my_schema;
+```
+
+This statement will create a new, empty schema. Once a schema is created, it may be used to qualify any newly created objects:
```
mjsax, line 46: From my understanding, in most systems it is not possible to create an object outside (ie, without) a schema, and if no schema is specified is a CREATE, the default schema is use. I also believe (not 100%) that the default schema cannot be dropped?

This part seems important with regard to upgrading, because it would imply that all existing objects would be migrated into the default schema.

design-proposals/klip-48-schemas.md
```diff
@@ -0,0 +1,115 @@
...
+ALTER TYPE my_type SET SCHEMA my_schema;
+ALTER CONNECTOR my_connector SET SCHEMA my_schema;
+```
+
+### Configuration
+
+As a matter of convenience and brevity, we should provide a means for users to specify a default schema. If a default schema is given, any references to an object that don’t explicitly include a schema qualifier should be interpreted as targeting the default schema:
+
+```sql
+SET 'ksql.default.schema' = 'my_schema';
```
mjsax, line 92: For a server config, this might be fine, but user should be able to switch the current schema they are operation in, and it might be better to have a dedicated `SET SCHEMA <schemaName>` command?

- https://www.ibm.com/docs/en/db2/11.1?topic=statements-set-schema
- https://docs.oracle.com/cd/B28359_01/server.111/b28310/general009.htm#ADMIN02101
- https://www.postgresql.org/docs/9.1/sql-set.html

design-proposals/klip-48-schemas.md
```diff
@@ -0,0 +1,115 @@
...
+```
+
+## Documentation updates
+
+* New documentation will be required to generally describe ksqlDB schemas and their associated syntax.
+* Any existing documentation pertaining to naming objects will need to be updated to account for the option of qualifying object names with a schema.
+
+## Compatibility implications
+
+* Once users upgrade from a pre-schemas version of ksqlDB to a version that supports schemas, how will ksqlDB interpret the older version's objects' schemas?
```
mjsax, line 114: Cf my comment above: it might be best to have a default schema and migrate existing objects into this default schema when upgrading).

design-proposals/klip-48-schemas.md
```diff
@@ -0,0 +1,115 @@
+# KLIP-41: Schemas
```
colinhicks, line 1:
```suggestion
# KLIP-48: Schemas
```

PR #7491, opened 2021-05-10
chore: bypass pre-join repartition for foreign key joins

vcrfxia:
### Description 

This PR refactors the logic in the logical planner for building join source nodes to first build a raw source node (either a JoinNode or a DataSourceNode) and then to add potential PreJoinRepartition or PreJoinProject nodes only after the join has been validated, since part of the validation is detecting whether the join is a foreign key join or not, and we do not want to add pre-join repartition nodes in the case of foreign key joins. This PR also adds a feature flag for foreign key joins while the feature is still under development.

### Testing done 

This PR is mostly a refactor; I'll ensure that existing unit and QTT tests pass prior to merging. Test coverage for the new functionality (involving foreign key joins themselves) will be added once more of the relevant functionality is wired in.

### Reviewer checklist

### DIFF
--- a/ksqldb-common/src/main/java/io/confluent/ksql/util/KsqlConfig.java
+++ b/ksqldb-common/src/main/java/io/confluent/ksql/util/KsqlConfig.java
@@ -131,6 +131,8 @@ public class KsqlConfig extends AbstractConfig {
       + "'CREATE STREAM S AS ...' will create a topic 'thing-S', where as the statement "
       + "'CREATE STREAM S WITH(KAFKA_TOPIC = 'foo') AS ...' will create a topic 'foo'.";
 
+  public static final String KSQL_FOREIGN_KEY_JOINS_ENABLED = "ksql.joins.foreign.key.enable";
+
   public static final String KSQL_ACTIVE_PERSISTENT_QUERY_LIMIT_CONFIG =
       "ksql.query.persistent.active.limit";
   private static final int KSQL_ACTIVE_PERSISTENT_QUERY_LIMIT_DEFAULT = Integer.MAX_VALUE;
@@ -782,6 +784,12 @@ public class KsqlConfig extends AbstractConfig {
             KSQL_ACTIVE_PERSISTENT_QUERY_LIMIT_DEFAULT,
             Importance.MEDIUM,
             KSQL_ACTIVE_PERSISTENT_QUERY_LIMIT_DOC
+        ).define(
+            KSQL_FOREIGN_KEY_JOINS_ENABLED,
+            Type.BOOLEAN,
+            false,
+            Importance.MEDIUM,
+            "Feature flag for foreign key joins, currently under development."
         ).define(
             KSQL_SHUTDOWN_TIMEOUT_MS_CONFIG,
             Type.LONG,
--- a/ksqldb-engine/src/main/java/io/confluent/ksql/planner/LogicalPlanner.java
+++ b/ksqldb-engine/src/main/java/io/confluent/ksql/planner/LogicalPlanner.java
@@ -463,27 +463,35 @@ public class LogicalPlanner {
   private PlanNode prepareSourceForJoin(
       final DataSourceNode sourceNode,
       final String side,
-      final Expression joinExpression
+      final Expression joinExpression,
+      final boolean isForeignKeyJoin
   ) {
-    // it is always safe to build the repartition node - this operation will be
-    // a no-op if a repartition is not required. if the source is a table, and
-    // a repartition is needed, then an exception will be thrown
-    final VisitParentExpressionVisitor<Optional<Expression>, Context<Void>> rewriter =
-        new VisitParentExpressionVisitor<Optional<Expression>, Context<Void>>(Optional.empty()) {
-          @Override
-          public Optional<Expression> visitQualifiedColumnReference(
-              final QualifiedColumnReferenceExp node,
-              final Context<Void> ctx
-          ) {
-            return Optional.of(new UnqualifiedColumnReferenceExp(node.getColumnName()));
-          }
-        };
+    final PlanNode preProjectNode;
+    if (isForeignKeyJoin) {
+      // we do not need to repartition for foreign key joins, as FK joins do not
+      // have co-partitioning requirements
+      preProjectNode = sourceNode;
+    } else {
+      // it is always safe to build the repartition node - this operation will be
+      // a no-op if a repartition is not required. if the source is a table, and
+      // a repartition is needed, then an exception will be thrown
+      final VisitParentExpressionVisitor<Optional<Expression>, Context<Void>> rewriter =
+          new VisitParentExpressionVisitor<Optional<Expression>, Context<Void>>(Optional.empty()) {
+            @Override
+            public Optional<Expression> visitQualifiedColumnReference(
+                final QualifiedColumnReferenceExp node,
+                final Context<Void> ctx
+            ) {
+              return Optional.of(new UnqualifiedColumnReferenceExp(node.getColumnName()));
+            }
+          };
 
-    final PlanNode repartition =
-        buildInternalRepartitionNode(sourceNode, side, joinExpression, rewriter::process);
+      preProjectNode =
+          buildInternalRepartitionNode(sourceNode, side, joinExpression, rewriter::process);
+    }
 
     return buildInternalProjectNode(
-        repartition,
+        preProjectNode,
         "PrependAlias" + side,
         sourceNode.getAlias()
     );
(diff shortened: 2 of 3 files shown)


ksqldb-engine/src/main/java/io/confluent/ksql/planner/LogicalPlanner.java
```diff
@@ -581,7 +627,10 @@ private JoinNode buildJoin(final Join root, final String prefix, final boolean i
     );
   }
 
-  private void verifyJoin(final JoinInfo joinInfo,
+  /**
+   * @return whether this is a foreign key join or not
```
vcrfxia, line 631: I don't love this but it seems better to have this method both perform verification and detect whether the join is an FK join or not than to have a second method duplicate the work in this one. I considered adding a join type enum rather than simply returning a boolean but we already have JoinType defined to mean something else: https://github.com/confluentinc/ksql/blob/89c1588440c26b6467ad58026ec664fbd0b6b85b/ksqldb-engine/src/main/java/io/confluent/ksql/planner/plan/JoinNode.java#L69-L71

Open to suggestions for how to make this feel less hacky.

mjsax: IMHO, it make actually a lot of sense to do is this way. We don't need to over-engineer for now and a boolean flag is just fine. After some more pending refactoring, we might get into a cleaner state. For not it's good as-is IMHO.

ksqldb-engine/src/main/java/io/confluent/ksql/planner/LogicalPlanner.java
```diff
@@ -581,7 +627,10 @@ private JoinNode buildJoin(final Join root, final String prefix, final boolean i
...
+   */
+  private boolean verifyJoin(final JoinInfo joinInfo,
                           final PlanNode leftNode,
```
mjsax, line 634: nit: fix indention

ksqldb-common/src/main/java/io/confluent/ksql/util/KsqlConfig.java
```diff
@@ -782,6 +784,12 @@ private static ConfigDef buildConfigDef(final ConfigGeneration generation) {
             KSQL_ACTIVE_PERSISTENT_QUERY_LIMIT_DEFAULT,
             Importance.MEDIUM,
             KSQL_ACTIVE_PERSISTENT_QUERY_LIMIT_DOC
+        ).define(
+            KSQL_FOREIGN_KEY_JOINS_ENABLED,
+            Type.BOOLEAN,
+            false,
```
mjsax, line 790: Not sure if this should be `true` for now? And we only disable by default is we don't make it?

ksqldb-functional-tests/src/test/java/io/confluent/ksql/test/planned/PlannedTestUtils.java
```diff
@@ -47,7 +49,8 @@ public static boolean isPlannedTestCase(final TestCase testCase) {
 
   public static boolean isNotExcluded(final TestCase testCase) {
     // Place temporary logic here to exclude test cases based on feature flags, etc.
-    return true;
+    final Map<String, Object> props = testCase.properties();
+    return !(boolean) props.getOrDefault(KsqlConfig.KSQL_FOREIGN_KEY_JOINS_ENABLED, false);
```
mjsax, line 53: For my own education: What does this do and why do we need it?

ksqldb-engine/src/main/java/io/confluent/ksql/planner/LogicalPlanner.java
```diff
@@ -535,40 +567,54 @@ private PlanNode buildSourceNode(final boolean isWindowed) {
...
 
-    verifyJoin(root.getInfo(), left, right);
+    final boolean isForeignKeyJoin =
+        verifyJoin(root.getInfo(), preRepartitionLeft, preRepartitionRight);
 
     final boolean finalJoin = prefix.isEmpty();
 
     final JoinKey joinKey = buildJoinKey(root);
 
+    final PlanNode left = prepareSourceForJoin(
```
mjsax, line 603: Wondering if we should pass the `isForeignKeyJoin` into this method, or do the check right here instead?
```
final PlanNode left;
final PlanNode right;
if (isForeignKeyJoin) {
  left = preRepartitionLeft;
  right = preRepartitionRight;
} else {
  left = prepareSourceForJoin(...);
  right = left = prepareSourceForJoin(...);
}
```

The idea is to keep the logic "contained". I seems `prepareSourceForJoin` does not really need to know anything about FK-joins? Atm, it feels like as if we have a "leaky" abstraction.

mjsax (commented):
Just some nits and questions. Overall LGTM.

ksqldb-common/src/main/java/io/confluent/ksql/util/KsqlConfig.java
vcrfxia: Feature flags guarding new functionality should default to false (especially for incompatible changes -- not as important for this particular feature but still nice to follow the same pattern) else we'd have to update/regenerate QTT tests while the feature is still a work-in-progress and likely to change.

ksqldb-engine/src/main/java/io/confluent/ksql/planner/LogicalPlanner.java
vcrfxia: If we have a foreign key join, then we don't want PreJoinRepartition nodes but we do still want PreJoinProject nodes on top of DataSourceNodes (the non-nested-join case). Therefore the logic in the foreign key join case is not as simple as in your sample code above, and it makes sense to continue wrapping it in a helper method. (Also helps de-dup code from the nested-join case since the logic for building PreJoinProject nodes is the same in both cases.)

ksqldb-functional-tests/src/test/java/io/confluent/ksql/test/planned/PlannedTestUtils.java
mjsax: I see -- that was my follow up question: if we disable the feature by default, how can we test it, ie, how can we enable it for QTT tests?

ksqldb-engine/src/main/java/io/confluent/ksql/planner/LogicalPlanner.java
mjsax: I see. I guess, it's good enough for now.

But it seems that `PreJoinProject` is actually an optimization, and thus should (mid term) not be part of the logical planner. The logical planner should generate a straight-forward plan and just verify if the query is valid and can be planned. This logical plan would be the input to an optimizer that modify the plan into an optimized (logical) plan, including projection push-down. More follow up refactoring for the future :)

ksqldb-engine/src/main/java/io/confluent/ksql/planner/LogicalPlanner.java
```diff
@@ -460,49 +460,81 @@ private FlatMapNode buildFlatMapNode(final PlanNode sourcePlanNode) {
...
+  ) {
+    final PlanNode preProjectNode;
+    if (isForeignKeyJoin) {
+      // we do not need to repartition for foreign key joins, as FK joins do not
+      // have co-partitioning requirements
+      preProjectNode = sourceNode;
+    } else {
+      // it is always safe to build the repartition node - this operation will be
+      // a no-op if a repartition is not required. if the source is a table, and
+      // a repartition is needed, then an exception will be thrown
```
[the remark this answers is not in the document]
vcrfxia: ksqlDB does not allow repartitioning tables as a safeguard to prevent users from unintentionally changing their table semantics, since ordering of records across partitions is not preserved. Users can work around this limitation by declaring a stream over the topic and repartitioning the stream if they wish.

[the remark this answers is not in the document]
mjsax: I would not say it's about "ordering" but about fundamental semantics: `PARTITION BY` changes the primary key in ksqlDB (it has slightly different semantics than in an RDBMS), similar to `GROUP BY` and for `GROUP BY` for all non-grouping columns an aggregation function is applied to "merge" values (of a column across all rows of a group) into a single value. For `PARTITION BY` there is no such mechanism: multiple rows could map to the same "new primary key" and thus the system cannot compute a "target row" because it's not well defined.

Of course we could pick a "random" one and then the ordering statement from Victoria applies: if we pick a random one, it would be non-deterministic and we should not allow non-deterministic operations. (That is also my concern with regard to `LATEST_BY_KEY` that allows users to repartition the table using a `GROUP BY` in a basically deterministic manner.)

PR #7499, opened 2021-05-11
chore: update joins to allow lack of pre-join repartition (MINOR)

vcrfxia:
### Description 

Minor follow-up to https://github.com/confluentinc/ksql/pull/7491. This PR only affects foreign key joins (still under development, hidden behind a feature flag). Because https://github.com/confluentinc/ksql/pull/7491 removed the PreJoinRepartition node for foreign key joins, it's now possible that a PreJoinProject node is immediately preceded by a DataSourceNode rather than another JoiningNode. This PR updates PreJoinProject to accommodate this rather than the current behavior which is to throw an error.

### Testing done 

Local QTT with feature flag enabled. (QTT for foreign key join still does not pass since the feature isn't complete yet, but the previous error that was thrown in PreJoinProject is no longer thrown.)

### Reviewer checklist

### DIFF
--- a/ksqldb-engine/src/main/java/io/confluent/ksql/planner/plan/PreJoinProjectNode.java
+++ b/ksqldb-engine/src/main/java/io/confluent/ksql/planner/plan/PreJoinProjectNode.java
@@ -17,6 +17,7 @@ package io.confluent.ksql.planner.plan;
 
 import com.google.common.collect.ImmutableBiMap;
 import com.google.common.collect.ImmutableList;
+import com.google.common.collect.Iterators;
 import io.confluent.ksql.execution.expression.tree.ColumnReferenceExp;
 import io.confluent.ksql.execution.expression.tree.UnqualifiedColumnReferenceExp;
 import io.confluent.ksql.execution.plan.SelectExpression;
@@ -44,7 +45,7 @@ public class PreJoinProjectNode extends ProjectNode implements JoiningNode {
   private final ImmutableList<SelectExpression> selectExpressions;
   private final ImmutableBiMap<ColumnName, ColumnName> aliases;
   private final LogicalSchema schema;
-  private final JoiningNode joiningSource;
+  private final Optional<JoiningNode> joiningSource;
 
   public PreJoinProjectNode(
       final PlanNodeId id,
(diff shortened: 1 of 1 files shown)


spena (approved):
LGTM

guozhangwang (approved):

ksqldb-engine/src/main/java/io/confluent/ksql/planner/plan/PreJoinProjectNode.java
```diff
@@ -73,12 +83,16 @@ public LogicalSchema getSchema() {
...
+      return joiningSource.get().getPreferredKeyFormat();
+    }
+
+    return Optional.of(getSourceKeyFormat());
   }
 
   @Override
   public void setKeyFormat(final KeyFormat format) {
-    joiningSource.setKeyFormat(format);
+    joiningSource.ifPresent(source -> source.setKeyFormat(format));
```
mjsax, line 95: Not sure if I understand this change? If `source` is not `JoiningNode` but a `DataSourceNode`, why can we make this call a no-op?

ksqldb-engine/src/main/java/io/confluent/ksql/planner/plan/PreJoinProjectNode.java
```diff
@@ -109,6 +123,12 @@ public void setKeyFormat(final KeyFormat format) {
     return super.validateColumns(builder.build());
   }
 
+  // Only safe to call this if joiningSource is empty.
+  private KeyFormat getSourceKeyFormat() {
```
mjsax, line 127: Would it be simpler to embed this code within `getPreferredKeyFormat`? It's not called from multiple places so what is the advantage of putting it into a method (and add a comment when it's safe to call?)

vcrfxia: Sure, I don't feel strongly.

ksqldb-engine/src/main/java/io/confluent/ksql/planner/plan/PreJoinProjectNode.java
mjsax: Wondering if we should throw an exception for this case instead of a no-op? Or could it happen that `setKeyFormat` is called for this case? (Seems, it should never be called for this case?)

vcrfxia: It will be called for this case. `setKeyFormat()` is called on all JoiningNodes.

PR #7501, opened 2021-05-12
fix: updated guave version

tolgadur:
### Description 
This is to update the guava version for the detected [CVE](https://confluentinc.atlassian.net/browse/KCI-405?page=com.atlassian.jira.plugin.system.issuetabpanels%3Aall-tabpanel).

### Testing done 
Made sure it builts.

### Reviewer checklist

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -94,7 +94,7 @@
         <antlr.version>4.7</antlr.version>
         <csv.version>1.4</csv.version>
         <lang3.version>3.3.1</lang3.version>
-        <guava.version>24.0-jre</guava.version>
+        <guava.version>24.1.1-jre</guava.version>
         <inject.version>1</inject.version>
         <janino.version>3.0.7</janino.version>
         <jline.version>3.11.0</jline.version>
@@ -268,6 +268,12 @@
                 <groupId>org.apache.avro</groupId>
                 <artifactId>avro</artifactId>
                 <version>${avro.version}</version>
+                <exclusions>
+                    <exclusion>
+                        <groupId>com.google.guava</groupId>
+                        <artifactId>guava</artifactId>
+                    </exclusion>
+                </exclusions>
             </dependency>
 
             <dependency>


pom.xml
```diff
@@ -94,7 +94,7 @@
         <antlr.version>4.7.1</antlr.version>
         <csv.version>1.4</csv.version>
         <lang3.version>3.3.1</lang3.version>
-        <guava.version>24.0-jre</guava.version>
+        <guava.version>24.1.1-jre</guava.version>
```
ableegoldman, line 97: We have to bump this up further to a 30.x version on the 5.3.x branch, so we should probably do the same here (or we just pint merge the [5.3.x PR](https://github.com/confluentinc/ksql/pull/7463/files) to sync the version all the way up to master)
@wcarlson5 do you think your PR is ready to be merged, or are you still investigating the other CVEs?

(also, why is 5.3.x already on 24.1.1, but master is only on 24.0 🤔 )

wcarlson5: I still have other CVEs I need to take care of but I don't think there is any reason to fix them all in the same PR. If you want to pint merge i don't see any problems as long as it works

tolgadur: I updated it to 30.0-jre.

ableegoldman: Oh, I just realized this PR was against 5.2.x -- not sure why I thought it was master 🤦‍♀️ 

Anyways I think we should be good if we just bump it to `30.1.1-jre` -- then we can pint merge this PR instead of the 5.3.x one to get the `guava` version settled all the way up to master

pom.xml
```diff
@@ -268,6 +268,12 @@
                 <groupId>org.apache.avro</groupId>
                 <artifactId>avro</artifactId>
                 <version>${avro.version}</version>
+                <exclusions>
+                    <exclusion>
+                        <groupId>com.google.guava</groupId>
+                        <artifactId>guava</artifactId>
```
ableegoldman, line 274: Just curious, why do we need this? Also do you know where/what `{avro.version}` is defined as, I don't see it in the list of properties with all the other versions above

swist: avro.version comes from `confluent.common`
I asked tolga to cherry-pick this upgrade from 5.3.x because I suspected that naive version bump might break the old version (look at the original PR linked #2544)

ableegoldman: I think the bigger question here is why do we have different versions of the same dependencies scattered throughout the Confluent codebase. We should just update it in `confluentinc/common` and pull it in that way, so we don't end up with potential runtime errors when various confluent repos start pulling in different versions of shared dependencies and overriding each other.

Note that `guava` is already on `30.1.1-jre` on the 5.2.x branch of `confluentinc/common` (cc @wcarlson5 )

wcarlson5: Interesting. for 5.3.x We are planning on removing the version override and inheriting from the parent (which is also 30.1.1)

ableegoldman: Yeah, we should do the same thing on 5.2.x as we have for 5.3.x, unless there is something in 5.2.x which is not compatible with the newer version of guava. Is that the case or was this a preemptive exclusion? Imo we should at least try this first and see if things break. It was not clear to me whether or not this exclusion was necessary back in [#2544](https://github.com/confluentinc/ksql/pull/2544) either

swist: OK let's split this up. For now we know we can merge this with just cursory "does it build" testing and it will be fine. 
Then any changes removing the overrides we can pint merge when they land in 5.3.x

swist (approved):

pom.xml
swist: This also means that if someone is on 5.3.x they won't need to bump guava in order to get CVE fixes

PR #6641, opened 2020-11-18
feat: add partitions to PRINT TOPIC output

agavra:
### Description 

Simple change to add partitions to the output of print topic

### Testing done 

Unit testing & local testing

### Reviewer checklist
- [ ] Ensure docs are updated if necessary. (eg. if a user visible feature is being added or changed).
- [ ] Ensure relevant issues are linked (description should include text like "Fixes #<issue number>")

PR #7505, opened 2021-05-12
docs(klip-50): partition and offset in ksqlDB

vcrfxia:
### Description 

KLIP 50: proposal for exposing partition and offset in ksqlDB. Here's a rendered version for easier viewing: https://github.com/vcrfxia/ksql/blob/klip-50/design-proposals/klip-50-partition-and-offset-in-ksqldb.md

### Testing done 

Docs-only change.

### Reviewer checklist

### DIFF
--- a/design-proposals/klip-50-partition-and-offset-in-ksqldb.md
+++ b/design-proposals/klip-50-partition-and-offset-in-ksqldb.md
@@ -1,0 +1,253 @@
+# KLIP 50 - Partition and Offset in ksqlDB
+
+**Author**: Victoria Xia (@vcrfxia) | 
+**Release Target**: 0.23.0 | 
+**Status**: _In Discussion_ | 
+**Discussion**: https://github.com/confluentinc/ksql/pull/7505
+
+**tl;dr:** _To aid in development and debugging of ksqlDB applications, ksqlDB should
+			allow access to the underlying Kafka topic partition and offset of messages
+			during processing. These metadata may also be used to skip select messages
+			("poison pills") during processing._
+
+## Motivation and background
+
+As described in the relevant [Github issue](https://github.com/confluentinc/ksql/issues/976),
+it's helpful when developing and troubleshooting ksqlDB queries, or when using ksqlDB to
+inspect other anomalies in a Kafka topic, to be able to target individual Kafka records.
+Today this can be approximated by filtering on the message timestamp via the `ROWTIME`
+pseudocolumn in ksqlDB, but this is only approximate as there could be multiple messages
+with the same timestamp. In contrast, the combination of topic partition and offset is
+guaranteed to uniquely identify a specific message.
+
+## What is in scope
+
+* Syntax for accessing Kafka topic partition and offset
+* Desired query semantics (persistent and transient)
+* Desired pull query semantics (specifically, that these metadata will not be available 
+  by default and must instead be copied into the message value to be accessible)
+* Other semantics, such as whether these metadata will be included in source descriptions
+  or `PRINT TOPIC` commands, or allowed in `INSERT VALUES` statements
+* Compatibility implications for existing ksqlDB queries
+* High-level implementation strategy
+
+## What is not in scope
+
+* [Exposing Kafka message headers in ksqlDB](https://github.com/confluentinc/ksql/issues/1940).
+  We would like to do this in the future but we'll have a separate KLIP for this as it's a
+  larger discussion.
+
+## Value/Return
+
+ksqlDB will expose message partition and offset as pseudocolumns similar to the `ROWTIME`
+pseudocolumn that exists today. Users will be able to select the partition and offset in
+transient queries, persistent queries, and also use the metadata in filters and UDFs.
+(See ["Public APIS"](#public-apis) below for example syntax.)
+
+This will allow users to more easily inspect data and debug ksqlDB queries:
+```
+SELECT *, ROWPARTITION, ROWOFFSET FROM my_stream EMIT CHANGES;
+```
+experiment with processing only a subset of messages:
+```
+SELECT * FROM my_stream WHERE ROWPARTITION = 0 AND ROWOFFSET > 12 EMIT CHANGES;
+```
+and skip select messages ("poison pills") during processing:
+```
+CREATE STREAM my_cleaned_stream AS SELECT * FROM my_stream WHERE NOT (ROWPARTITION = 0 AND ROWOFFSET = 42);
+```
+
+## Public APIS
+
+We will add two new [pseudocolumns](https://github.com/confluentinc/ksql/blob/master/design-proposals/klip-14-rowtime-as-pseudocolumn.md)
+to ksqlDB: `ROWPARTITION` (integer) and `ROWOFFSET` (bigint). We choose these names to be 
+consistent with the existing `ROWTIME` pseudocolumn, and to minimize the chance of conflict 
+with users' existing data columns for compatibility reasons. See 
+["Alternatives Considered"](#alternatives-considered) for discussion on possible alternatives.
+
+### Query Semantics
+
+The behavior of `ROWPARTITION` and `ROWOFFSET` in persistent queries and push queries
+will be analogous to those of the [existing `ROWTIME` pseudocolumn](https://github.com/confluentinc/ksql/blob/master/design-proposals/klip-14-rowtime-as-pseudocolumn.md#queries).
+In summary:
+
+* `SELECT *` projections will not include `ROWPARTITION` and `ROWOFFSET`
+* Users may choose to explicitly add `ROWPARTITION` or `ROWOFFSET` in the selection,
+  which will cause ksqlDB to copy the selected pseudocolumn(s) into the output _value_
+* `ROWPARTITION` and `ROWOFFSET` will become reserved column names, and users will not
+  be allowed to use these names either as source column names or as output column aliases.
+  (Existing persistent queries that already use these names will continue to run undisrupted.
+  See ["Compatibility Implications"](#compatibility-implications) for more.)
+
+Consistent with `ROWTIME`, the new `ROWPARTITION` and `ROWOFFSET` pseudocolumns will 
+also be allowed in `WHERE`, `GROUP BY`, and `PARTITION BY` clauses, though use cases
+for the latter two are probably limited. `ROWPARTITION` and `ROWOFFSET` will also be
+allowed in UDFs.
+
+### Pull Queries
+
+Today, users are able to select the `ROWTIME` pseudocolumn as part of pull queries:
+```
+SELECT *, ROWTIME FROM my_table WHERE my_key='foo';
+```
+For multiple reasons, we propose that the new `ROWPARTITION` and `ROWOFFSET` pseudocolumns
+should diverge from this behavior. In other words, users will NOT be allowed to select
+`ROWPARTITION` or `ROWOFFSET` as part of a pull query.
+
+The first reason is that selecting `ROWPARTITION` or `ROWOFFSET` for a table
+is semantically dubious, especially for tables that are the result of other
+computation such as an aggregation over a stream. Should `ROWPARTITION` and `ROWOFFSET`
+in such a case represent the partition and offset of the message in the source stream
+that resulted in the aggregate value, or should they represent the partition and offset
+of the aggregate value in the output table? Note that timestamps don't have this
+ambiguity as the timestamp of the two records mentioned are the same.
+
+The second reason has to do with implications on state store size. ksqlDB is able to 
+support `ROWTIME` in pull queries because the state stores from which pull queries are 
+served includes the timestamp for the latest table row (for each key). In contrast, 
+message partition and offset are not saved in the state stores. As a result, supporting 
+`ROWPARTITION` and `ROWOFFSET` in pull queries would require adding this information
+to state stores. (I think there might be a way to get the partition by other means,
+from the replica queried in order to serve the pull query, but offset information
+definitely wouldn't be available.)
+
+If a user really wants to make partition and/or offset information available for pull
+queries, they can choose to explicitly include the source partition and offset as
+part of the query that materializes their table and query for partition/offset as
+value columns as part of their pull query.
+
+### Other semantics
+
+Similar to `ROWTIME`, the new `ROWPARTITION` and `ROWOFFSET` pseudocolumns will not 
+be included in the list of columns returned as part of a source description.
+
+`ROWPARTITION` and `ROWOFFSET` will also not be added to the output of `PRINT TOPIC` commands.
+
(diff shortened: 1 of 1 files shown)


design-proposals/klip-50-partition-and-offset-in-ksqldb.md
```diff
@@ -0,0 +1,253 @@
...
+
+We will add two new [pseudocolumns](https://github.com/confluentinc/ksql/blob/master/design-proposals/klip-14-rowtime-as-pseudocolumn.md)
+to ksqlDB: `ROWPARTITION` (integer) and `ROWOFFSET` (bigint). We choose these names to be 
+consistent with the existing `ROWTIME` pseudocolumn, and to minimize the chance of conflict 
+with users' existing data columns for compatibility reasons. See 
+["Alternatives Considered"](#alternatives-considered) for discussion on possible alternatives.
+
+### Query Semantics
+
+The behavior of `ROWPARTITION` and `ROWOFFSET` in persistent queries and push queries
```
guozhangwang, line 70: From the section below, seems you were proposing that we do not add these two to the push queries, right? If yes maybe better to remove `and push queries` here.

design-proposals/klip-50-partition-and-offset-in-ksqldb.md
```diff
@@ -0,0 +1,253 @@
...
+```
+SELECT *, ROWTIME FROM my_table WHERE my_key='foo';
+```
+For multiple reasons, we propose that the new `ROWPARTITION` and `ROWOFFSET` pseudocolumns
+should diverge from this behavior. In other words, users will NOT be allowed to select
+`ROWPARTITION` or `ROWOFFSET` as part of a pull query.
+
+The first reason is that selecting `ROWPARTITION` or `ROWOFFSET` for a table
+is semantically dubious, especially for tables that are the result of other
+computation such as an aggregation over a stream. Should `ROWPARTITION` and `ROWOFFSET`
```
guozhangwang, line 99: This argument may apply to persistent CSAS queries as well: should the partition/offset come from the resulted stream, or from the source stream record that generated the result record. Personally I feel just stick to say "it is always the resulted stream/table" is fine, but that also means we do not provide any way still for users to check the partition/offset of the source topics.

design-proposals/klip-50-partition-and-offset-in-ksqldb.md
```diff
@@ -0,0 +1,253 @@
...
+`ROWPARTITION` and `ROWOFFSET` in pull queries, while `ROWTIME` will still be allowed.
+
+However, directly adding `ROWPARTITION` and `ROWOFFSET` as new pseudocolumns would break 
+compatibility for certain existing queries since, in some situations, 
+[ksqlDB writes pseudocolumns into repartition and changelog topics](https://github.com/confluentinc/ksql/issues/7489). 
+This is problematic because adding new pseudocolumns would change the schema of these 
+intermediate topics (and state stores) and break running queries as a result.
+
+In light of the above, we have two approaches for implementation:
+1. Update ksqlDB to never write pseudocolumns into repartition and changelog topics
```
guozhangwang, line 145: I thought the ultimate solution would be https://github.com/confluentinc/ksql/issues/6374. More specifically, since the partition/offset are only for the sink topics, so conceptually intermediate topics would never need to encode the partition/offset at all assuming we have clear separation between physical and logical schemas, no?

design-proposals/klip-50-partition-and-offset-in-ksqldb.md
```diff
@@ -0,0 +1,253 @@
...
+## Value/Return
+
+ksqlDB will expose message partition and offset as pseudocolumns similar to the `ROWTIME`
+pseudocolumn that exists today. Users will be able to select the partition and offset in
+transient queries, persistent queries, and also use the metadata in filters and UDFs.
+(See ["Public APIS"](#public-apis) below for example syntax.)
+
+This will allow users to more easily inspect data and debug ksqlDB queries:
+```
+SELECT *, ROWPARTITION, ROWOFFSET FROM my_stream EMIT CHANGES;
```
agavra, line 49: nit: thoughts on `ROW_PARTITION` or just `PARTITION`?

design-proposals/klip-50-partition-and-offset-in-ksqldb.md
```diff
@@ -0,0 +1,253 @@
...
+
+* `SELECT *` projections will not include `ROWPARTITION` and `ROWOFFSET`
+* Users may choose to explicitly add `ROWPARTITION` or `ROWOFFSET` in the selection,
+  which will cause ksqlDB to copy the selected pseudocolumn(s) into the output _value_
+* `ROWPARTITION` and `ROWOFFSET` will become reserved column names, and users will not
+  be allowed to use these names either as source column names or as output column aliases.
+  (Existing persistent queries that already use these names will continue to run undisrupted.
+  See ["Compatibility Implications"](#compatibility-implications) for more.)
+
+Consistent with `ROWTIME`, the new `ROWPARTITION` and `ROWOFFSET` pseudocolumns will
```
agavra, line 82: should we discuss `INSERT INTO` and `INSERT VALUES` here as well? (for the former, what happens if you `INSERT INTO` into a stream that selected partition/offset (I don't think there's a problem, you might just produce garbage data) and for the latter it's probably OK to prohibit inserting those fields, unlike `ROWTIME`)

EDIT: I see you discuss this below for `INSERT VALUES` :)

design-proposals/klip-50-partition-and-offset-in-ksqldb.md
agavra, line 70: can we be extra clear about which partition/offset it chooses when there are repartition topics involved? (I assume it would be the source)

design-proposals/klip-50-partition-and-offset-in-ksqldb.md
```diff
@@ -0,0 +1,253 @@
...
+queries, they can choose to explicitly include the source partition and offset as
+part of the query that materializes their table and query for partition/offset as
+value columns as part of their pull query.
+
+### Other semantics
+
+Similar to `ROWTIME`, the new `ROWPARTITION` and `ROWOFFSET` pseudocolumns will not 
+be included in the list of columns returned as part of a source description.
+
+`ROWPARTITION` and `ROWOFFSET` will also not be added to the output of `PRINT TOPIC` commands.
```
agavra, line 124: partition is already there (https://github.com/confluentinc/ksql/pull/6641) - it might make sense to add offset as well, even if it isn't part of the "print" schema

design-proposals/klip-50-partition-and-offset-in-ksqldb.md
```diff
@@ -0,0 +1,253 @@
...
+
+In light of the above, we have two approaches for implementation:
+1. Update ksqlDB to never write pseudocolumns into repartition and changelog topics 
+   (unless they are needed as part of the query downstream), i.e., address 
+   https://github.com/confluentinc/ksql/issues/7489. Then we can freely add new 
+   pseudocolumns without breaking compatibility for existing queries.
+2. Defer on updating ksqlDB to not write pseudocolumns into intermediate topics and 
+   instead modify the stream and table source execution steps to include something 
+   to indicate the set of pseudocolumns that should be included with the source. 
+   Specifically, we can add a "pseudocolumn version number" field into the execution
```
agavra, line 152: instead of a version number, why not just list out all the psuedocolumns to include? magic versions numbers can be tough to reason about

design-proposals/klip-50-partition-and-offset-in-ksqldb.md
```diff
@@ -0,0 +1,253 @@
...
+2. Defer on updating ksqlDB to not write pseudocolumns into intermediate topics and 
+   instead modify the stream and table source execution steps to include something 
+   to indicate the set of pseudocolumns that should be included with the source. 
+   Specifically, we can add a "pseudocolumn version number" field into the execution 
+   step. By increasing the version number when new pseudocolumns are added, old queries 
+   will continue to run with the old set of pseudocolumns and therefore maintain 
+   compatibility, while new queries will have new pseudocolumns.
+
+Given that we'd like to start work on exposing partition and offset within the next month,
+my preference is for Option 2. The drawback is that intermediate topics for pre-join
```
agavra, line 158: +1

agavra (approved):
Excellent KLIP - thanks @vcrfxia :)

design-proposals/klip-50-partition-and-offset-in-ksqldb.md
agavra: Just commented the same thing, but I felt that it should come from the source not the result topic partition/offset 😆 - cc @colinhicks for some product insight 🔮

design-proposals/klip-50-partition-and-offset-in-ksqldb.md
vcrfxia: Fascinating, and good to know! I'll move this to be out of scope for this KLIP since it's a bit orthogonal, but I agree it could be nice to add.

design-proposals/klip-50-partition-and-offset-in-ksqldb.md
vcrfxia: > since the partition/offset are only for the sink topics

As discussed above, the intention was for `ROWPARTITION` and `ROWOFFSET` to represent the source partition and offset, rather than the sink partition and offset. Let's resolve that discussion first, and return to this if it's still relevant.

design-proposals/klip-50-partition-and-offset-in-ksqldb.md
vcrfxia: I considered this but felt that a version number would allow for more evolvability in the future. For example, if we wanted to change the implementation of a pseudocolumn, that would be possible with a version number but not a list of column names. (A bit contrived but maybe we'll introduce `ROWHEADERS` in the future with type `ARRAY<STRUCT<key STRING, header BYTES>>` and later want to change it to a different type.)

I don't feel strongly though. Happy to keep it readable if introducing a version number feels like over-engineering.

design-proposals/klip-50-partition-and-offset-in-ksqldb.md
colinhicks: I also agree they should be relative to the source.

design-proposals/klip-50-partition-and-offset-in-ksqldb.md
```diff
@@ -0,0 +1,253 @@
...
+message partition and offset are not saved in the state stores. As a result, supporting 
+`ROWPARTITION` and `ROWOFFSET` in pull queries would require adding this information
+to state stores. (I think there might be a way to get the partition by other means,
+from the replica queried in order to serve the pull query, but offset information
+definitely wouldn't be available.)
+
+If a user really wants to make partition and/or offset information available for pull
+queries, they can choose to explicitly include the source partition and offset as
+part of the query that materializes their table and query for partition/offset as
+value columns as part of their pull query.
```
colinhicks, line 117: I think we should plan to illustrate this in the documentation with an example materializing query. Might also make sense to add the example query in the KLIP as well.

vcrfxia: Sure, added an example to the KLIP. Out of curiosity, what use cases do you envision for including partition/offset as part of a pull query? It seems less natural to be interested in partition and offset for table records compared to messages in a stream.

colinhicks: Yeah, I agree it's less natural. I could see it being educationally useful for folks checking their understanding about how streams and tables relate, for example to see how the aggregate row corresponds to the source stream's partition and offset. Thanks for adding it.

guozhangwang:
LGTM.

design-proposals/klip-50-partition-and-offset-in-ksqldb.md
guozhangwang: When punctuation is triggered, we may send records to downstream which do not have partition/offset/etc metadata since they are not triggered by data, but by time. Which means, we may need to be able to set some sentinel values in the corresponding columns for records generated by punctuations.

Yes vector clocks are for the context of punctuations, they can cover both data-driven and time-driven (a.k.a. punctuation) events, but are flushed across chains of query topologies with the source topic metadata (partition/offset) in it. E.g. say we have a chain of queries:

```
source -> query1 -> topic1 -> query2 -> topic2 -> query3 -> sink
```

Then the metadata would always be representing the `source` topic, flowing through topic1/2 and sink. For this KLIP we may need the metadata to be representing the `source`, `topic1`, and `topic2` respectively for query1/2/3, so for now we can just use the context metadata to achieve this. I'm just thinking time-driven results maybe we can do piggyback on the same mechanism (a.k.a. control messages) to do that in the longer term.

design-proposals/klip-50-partition-and-offset-in-ksqldb.md
agavra: partition offset for pull queries would be super useful for debugging purposes (answering "why is a certain piece of data there?")

design-proposals/klip-50-partition-and-offset-in-ksqldb.md
```diff
@@ -0,0 +1,272 @@
...
+ambiguity as the timestamp of the two records mentioned are the same.
+
+The second reason has to do with implications on state store size. ksqlDB is able to 
+support `ROWTIME` in pull queries because the state stores from which pull queries are 
+served includes the timestamp for the latest table row (for each key). In contrast, 
+message partition and offset are not saved in the state stores. As a result, supporting 
+`ROWPARTITION` and `ROWOFFSET` in pull queries would require adding this information
+to state stores. (I think there might be a way to get the partition by other means,
+from the replica queried in order to serve the pull query, but offset information
+definitely wouldn't be available.)
```
mjsax, line 124: I actually think this also applies to partitions. If we have two input records with the same grouping attributes, but from different input partitions (and we do a repartitioning), it's not defined which partition we should write into the result table (as we don't want to use the partition of the repartition topic).

design-proposals/klip-50-partition-and-offset-in-ksqldb.md
```diff
@@ -0,0 +1,272 @@
...
+
+If a user really wants to make partition and/or offset information available for pull
+queries, they can choose to explicitly include the source partition and offset as
+part of the query that materializes their table and query for partition/offset as
+value columns as part of their pull query:
```
mjsax, line 129: We need to make sure to treat both columns as any other columns (ie, for aggregations, they can only be included if they are part of the `GROUP BY` clause, or an aggregation function is applied to them).

design-proposals/klip-50-partition-and-offset-in-ksqldb.md
mjsax: ksqlDB and KS DSL don't use punctuations -- so we should not worry about it? -- I also think that vector clocks won't help, as we need the partition/offset information on a per record basis while vector-clock data would only be written periodically.

PR #7511, opened 2021-05-12
chore: disallow FK joins in n-way joins (MINOR)

vcrfxia:
### Description 

Three minor changes in this PR:
- disallows foreign key joins in n-way joins, for this first iteration of FK joins. Note that it's fine if a chain of n-way joins starts with an FK join (as processed from left to right), but not if an FK join is encountered later on in the chain.
- adds versions of the existing FK joins tests (all negative test cases, as the functionality is not yet complete) with the feature flag enabled
- updates the existing negative test case for FK joins as part of an n-way join to have the FK join as the second join in the chain, rather than the first, in light of the note above

### Testing done 

Tests pass.

### Reviewer checklist

### DIFF
--- a/ksqldb-functional-tests/src/test/resources/query-validation-tests/fk-join.json
+++ b/ksqldb-functional-tests/src/test/resources/query-validation-tests/fk-join.json
@@ -101,7 +101,9 @@
       "comments": [
         "After we add support for binary FK-joins, this test should still fail, but with a different error:",
         "The new error should point out that FK-joins cannot be used in n-way joins yet.",
-        "Note: this case should be supported after we lift the general restriction to not support FK-joins in n-way joins."
+        "Note: this case should be supported after we lift the general restriction to not support FK-joins in n-way joins.",
+        "TODO: this test should be deleted once the feature is ungated and the feature flag is removed",
+        "as it duplicates another test below (with the feature flag enabled)"
       ],
       "statements": [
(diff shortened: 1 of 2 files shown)


guozhangwang (approved):
LGTM!

ksqldb-functional-tests/src/test/resources/query-validation-tests/fk-join.json
```diff
@@ -101,7 +101,9 @@
       "comments": [
         "After we add support for binary FK-joins, this test should still fail, but with a different error:",
         "The new error should point out that FK-joins cannot be used in n-way joins yet.",
-        "Note: this case should be supported after we lift the general restriction to not support FK-joins in n-way joins."
+        "Note: this case should be supported after we lift the general restriction to not support FK-joins in n-way joins.",
+        "TODO: this test should be deleted once the feature is ungated and the feature flag is removed",
```
mjsax, line 105: Should we no remove the other test after removing the feature flag, as we don't need to enable the feature in tests any longer?

mjsax (approved):

ksqldb-functional-tests/src/test/resources/query-validation-tests/fk-join.json
vcrfxia: No, we'll need to remove this one because the error message is different and the other copy of the test has the correct error message.

PR #7517, opened 2021-05-12
feat: build physical plan for FK table-table joins

mjsax:
### Description 
Add step to add FK-join to the physical plan.

### Testing done 
Added some unit testing. Later we need to extend QTT tests when we actually use this code (that is currently "deal code").

### Reviewer checklist

### DIFF
--- a/ksqldb-execution/src/main/java/io/confluent/ksql/execution/plan/ForeignKeyTableTableJoin.java
+++ b/ksqldb-execution/src/main/java/io/confluent/ksql/execution/plan/ForeignKeyTableTableJoin.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2019 Confluent Inc.
+ * Copyright 2021 Confluent Inc.
  *
  * Licensed under the Confluent Community License; you may not use this file
  * except in compliance with the License.  You may obtain a copy of the License at
--- a/ksqldb-execution/src/main/java/io/confluent/ksql/execution/plan/ForeignKeyTableTableJoin.java
+++ b/ksqldb-execution/src/main/java/io/confluent/ksql/execution/plan/ForeignKeyTableTableJoin.java
@@ -16,7 +16,9 @@ package io.confluent.ksql.execution.plan;
 
 import static java.util.Objects.requireNonNull;
 
+import com.fasterxml.jackson.annotation.JsonCreator;
 import com.fasterxml.jackson.annotation.JsonIgnore;
+import com.fasterxml.jackson.annotation.JsonProperty;
 import com.google.common.collect.ImmutableList;
 import com.google.errorprone.annotations.Immutable;
 import io.confluent.ksql.name.ColumnName;
@@ -33,12 +35,13 @@ public class ForeignKeyTableTableJoin<KLeftT, KRightT>
   private final ExecutionStep<KTableHolder<KLeftT>> leftSource;
   private final ExecutionStep<KTableHolder<KRightT>> rightSource;
 
+  @JsonCreator
   public ForeignKeyTableTableJoin(
-      final ExecutionStepPropertiesV1 props,
-      final JoinType joinType,
-      final ColumnName leftJoinColumnName,
-      final ExecutionStep<KTableHolder<KLeftT>> leftSource,
-      final ExecutionStep<KTableHolder<KRightT>> rightSource
+      @JsonProperty(value = "properties", required = true) final ExecutionStepPropertiesV1 props,
+      @JsonProperty(value = "joinType", required = true) final JoinType joinType,
+      @JsonProperty(value = "leftJoinColumnName", required = true) final ColumnName leftJoinColumnName,
+      @JsonProperty(value = "leftSource", required = true) final ExecutionStep<KTableHolder<KLeftT>> leftSource,
+      @JsonProperty(value = "rightSource", required = true) final ExecutionStep<KTableHolder<KRightT>> rightSource
   ) {
     this.properties = requireNonNull(props, "props");
     this.joinType = requireNonNull(joinType, "joinType");
--- a/ksqldb-execution/src/main/java/io/confluent/ksql/execution/plan/ForeignKeyTableTableJoin.java
+++ b/ksqldb-execution/src/main/java/io/confluent/ksql/execution/plan/ForeignKeyTableTableJoin.java
@@ -37,11 +37,16 @@ public class ForeignKeyTableTableJoin<KLeftT, KRightT>
 
   @JsonCreator
   public ForeignKeyTableTableJoin(
-      @JsonProperty(value = "properties", required = true) final ExecutionStepPropertiesV1 props,
-      @JsonProperty(value = "joinType", required = true) final JoinType joinType,
-      @JsonProperty(value = "leftJoinColumnName", required = true) final ColumnName leftJoinColumnName,
-      @JsonProperty(value = "leftSource", required = true) final ExecutionStep<KTableHolder<KLeftT>> leftSource,
-      @JsonProperty(value = "rightSource", required = true) final ExecutionStep<KTableHolder<KRightT>> rightSource
+      @JsonProperty(value = "properties", required = true)
+      final ExecutionStepPropertiesV1 props,
+      @JsonProperty(value = "joinType", required = true)
+      final JoinType joinType,
+      @JsonProperty(value = "leftJoinColumnName", required = true)
+      final ColumnName leftJoinColumnName,
+      @JsonProperty(value = "leftSource", required = true)
+      final ExecutionStep<KTableHolder<KLeftT>> leftSource,
+      @JsonProperty(value = "rightSource", required = true)
+      final ExecutionStep<KTableHolder<KRightT>> rightSource
   ) {
     this.properties = requireNonNull(props, "props");
     this.joinType = requireNonNull(joinType, "joinType");
--- a/ksqldb-execution/src/main/java/io/confluent/ksql/execution/plan/ForeignKeyTableTableJoin.java
+++ b/ksqldb-execution/src/main/java/io/confluent/ksql/execution/plan/ForeignKeyTableTableJoin.java
@@ -1,0 +1,109 @@
+/*
+ * Copyright 2019 Confluent Inc.
+ *
+ * Licensed under the Confluent Community License; you may not use this file
+ * except in compliance with the License.  You may obtain a copy of the License at
+ *
+ * http://www.confluent.io/confluent-community-license
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OF ANY KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations under the License.
+ */
+
+package io.confluent.ksql.execution.plan;
+
+import static java.util.Objects.requireNonNull;
+
+import com.fasterxml.jackson.annotation.JsonIgnore;
+import com.google.common.collect.ImmutableList;
+import com.google.errorprone.annotations.Immutable;
+import io.confluent.ksql.name.ColumnName;
+import java.util.List;
+import java.util.Objects;
+
+@Immutable
+public class ForeignKeyTableTableJoin<KLeftT, KRightT>
+    implements ExecutionStep<KTableHolder<KLeftT>> {
+
+  private final ExecutionStepPropertiesV1 properties;
+  private final JoinType joinType;
+  private final ColumnName leftJoinColumnName;
+  private final ExecutionStep<KTableHolder<KLeftT>> leftSource;
+  private final ExecutionStep<KTableHolder<KRightT>> rightSource;
+
+  public ForeignKeyTableTableJoin(
+      final ExecutionStepPropertiesV1 props,
+      final JoinType joinType,
+      final ColumnName leftJoinColumnName,
+      final ExecutionStep<KTableHolder<KLeftT>> leftSource,
+      final ExecutionStep<KTableHolder<KRightT>> rightSource
+  ) {
+    this.properties = requireNonNull(props, "props");
+    this.joinType = requireNonNull(joinType, "joinType");
+    if (joinType == JoinType.OUTER) {
+      throw new IllegalArgumentException("OUTER join not supported.");
+    }
+    this.leftJoinColumnName = requireNonNull(leftJoinColumnName, "leftJoinColumnName");
+    this.leftSource = requireNonNull(leftSource, "leftSource");
+    this.rightSource = requireNonNull(rightSource, "rightSource");
+  }
+
+  @Override
+  public ExecutionStepPropertiesV1 getProperties() {
+    return properties;
+  }
+
+  @Override
+  @JsonIgnore
+  public List<ExecutionStep<?>> getSources() {
+    return ImmutableList.of(leftSource, rightSource);
+  }
+
+  public ExecutionStep<KTableHolder<KLeftT>> getLeftSource() {
+    return leftSource;
+  }
+
+  public ExecutionStep<KTableHolder<KRightT>> getRightSource() {
+    return rightSource;
+  }
+
+  public JoinType getJoinType() {
+    return joinType;
+  }
+
+  public ColumnName getLeftJoinColumnName() {
+    return leftJoinColumnName;
+  }
+
+  @Override
+  public KTableHolder<KLeftT> build(final PlanBuilder builder, final PlanInfo info) {
+    return builder.visitForeignKeyTableTableJoin(this, info);
+  }
+
+  @Override
+  public PlanInfo extractPlanInfo(final PlanInfoExtractor extractor) {
+    return extractor.visitForeignKeyTableTableJoin(this);
+  }
+
+  @Override
+  public boolean equals(final Object o) {
+    if (this == o) {
+      return true;
+    }
+    if (o == null || getClass() != o.getClass()) {
+      return false;
+    }
+    final ForeignKeyTableTableJoin<?, ?> that = (ForeignKeyTableTableJoin<?, ?>) o;
+    return Objects.equals(properties, that.properties)
+        && joinType == that.joinType
+        && Objects.equals(leftSource, that.leftSource)
+        && Objects.equals(rightSource, that.rightSource);
+  }
+
+  @Override
+  public int hashCode() {
+    return Objects.hash(properties, joinType, leftSource, rightSource);
+  }
+}
--- a/ksqldb-execution/src/main/java/io/confluent/ksql/execution/plan/ForeignKeyTableTableJoin.java
+++ b/ksqldb-execution/src/main/java/io/confluent/ksql/execution/plan/ForeignKeyTableTableJoin.java
@@ -98,12 +98,13 @@ public class ForeignKeyTableTableJoin<KLeftT, KRightT>
     final ForeignKeyTableTableJoin<?, ?> that = (ForeignKeyTableTableJoin<?, ?>) o;
     return Objects.equals(properties, that.properties)
         && joinType == that.joinType
+        && Objects.equals(leftJoinColumnName, that.leftJoinColumnName)
         && Objects.equals(leftSource, that.leftSource)
         && Objects.equals(rightSource, that.rightSource);
   }
 
   @Override
   public int hashCode() {
-    return Objects.hash(properties, joinType, leftSource, rightSource);
+    return Objects.hash(properties, joinType, leftJoinColumnName, leftSource, rightSource);
   }
 }
--- a/ksqldb-streams/src/main/java/io/confluent/ksql/execution/streams/ExecutionStepFactory.java
+++ b/ksqldb-streams/src/main/java/io/confluent/ksql/execution/streams/ExecutionStepFactory.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2019 Confluent Inc.
+ * Copyright 2021 Confluent Inc.
  *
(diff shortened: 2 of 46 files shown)


ksqldb-streams/src/test/java/io/confluent/ksql/execution/streams/JoinParamsFactoryTest.java
```diff
@@ -28,15 +28,13 @@
       .valueColumn(ColumnName.of("R_K"), SqlTypes.STRING) // Copy of key in value
       .build();
 
-  private JoinParams joinParams;
```
mjsax, line 31: side cleanup to inline variable.

ksqldb-streams/src/test/java/io/confluent/ksql/execution/streams/TableTableJoinBuilderTest.java
```diff
@@ -193,7 +193,12 @@ public void shouldReturnCorrectSchema() {
     // Then:
     assertThat(
         result.getSchema(),
-        is(JoinParamsFactory.create(R_KEY, LEFT_SCHEMA, RIGHT_SCHEMA).getSchema())
```
mjsax, line 196: Side cleanup: We should not use `JoinParamsFactory` to compute the expected result, because `JoinParamsFactory` is use to compute the actual `result` within `join.build(...)`. Otherwise we have a cyclic tes "dependency"...

Same below.

ksqldb-streams/src/main/java/io/confluent/ksql/execution/streams/ForeignKeyTableTableJoinBuilder.java
```diff
@@ -0,0 +1,66 @@
...
+public final class ForeignKeyTableTableJoinBuilder {
+
+  private ForeignKeyTableTableJoinBuilder() {
+  }
+
+  public static <KLeftT, KRightT> KTableHolder<KLeftT> build(
+      final KTableHolder<KLeftT> left,
+      final KTableHolder<KRightT> right,
+      final ForeignKeyTableTableJoin<KLeftT, KRightT> join
+  ) {
```
spena, line 33: What if we add info to the current `TableTableJoin` where we could check if the join is on a foreign key? Maybe we can reuse the `TableTableJoinBuilder`, and just change the impl. to call a different builder for foregign keys and another that has the current code.

ksqldb-execution/src/main/java/io/confluent/ksql/execution/plan/ForeignKeyTableTableJoin.java
```diff
@@ -0,0 +1,110 @@
...
+  private final ExecutionStep<KTableHolder<KLeftT>> leftSource;
+  private final ExecutionStep<KTableHolder<KRightT>> rightSource;
+
+  public ForeignKeyTableTableJoin(
+      final ExecutionStepPropertiesV1 props,
+      final JoinType joinType,
+      final ColumnName leftJoinColumnName,
+      final ExecutionStep<KTableHolder<KLeftT>> leftSource,
+      final ExecutionStep<KTableHolder<KRightT>> rightSource
+  ) {
```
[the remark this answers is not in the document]
vcrfxia: > I'm not sure if we should have two classes for each time of join. They have repeated code except the extra left column name.

I think it makes sense to have a separate execution step for FK joins since the implementation is different. Execution steps are meant to be fine-grained so we can evolve them separately as needed.

> Also, I'm not sure if the TableTableJoin info is written in the plan. I see that class has a @JsonCreator and it is also referenced in PlanSummary.java as .put(TableTableJoin.class, "JOIN"). Did you verify this part?

PlanSummary.java, ExecutionStep.java, and StepSchemaResolver.java all need to be updated to include the new execution step. Let's add these changes into this PR so we don't hit confusing failures down the line when wiring up the functionality.

> The TableTableJoin accepts an optional keyColumn. Perhaps we can add another parameter to accept an optional nonKeyColumn too?

Unless I'm mistaken, the optional keyColumn in TableTableJoin specifies the name of the _output_ key column. That is not necessary for FK joins because the output key always matches the left source key.

ksqldb-streams/src/main/java/io/confluent/ksql/execution/streams/ExecutionStepFactory.java
```diff
@@ -320,6 +321,23 @@ public static WindowedTableSource tableSourceWindowed(
     );
   }
 
+  public static <KLeftT, KRightT> ForeignKeyTableTableJoin<KLeftT, KRightT>
+      foreignKeyTableTableJoin(final QueryContext.Stacker stacker,
+                               final JoinType joinType,
+                               final ColumnName leftJoinColumnName,
```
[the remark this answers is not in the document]
vcrfxia: If we're going to have a separate execution step, we should have a separate method for creating the execution step. (See above comment.)

ksqldb-streams/src/main/java/io/confluent/ksql/execution/streams/ForeignKeyJoinParamsFactory.java
```diff
@@ -0,0 +1,63 @@
...
+        createKeyExtractor(leftSchema, leftJoinColumnName),
+        new KsqlValueJoiner(leftSchema.value().size(), rightSchema.value().size(), 0),
+        createSchema(leftSchema, rightSchema)
+    );
+  }
+
+  public static LogicalSchema createSchema(
+      final LogicalSchema leftSchema,
+      final LogicalSchema rightSchema
+  ) {
```
vcrfxia, line 44: We need to validate that the join expression types match. If we follow the pattern of the existing JoinParamsFactory, we can do that here. 

JoinParamsFactory also validates that the join expressions represent the complete key, which is a requirement for the right side of FK joins as well. I think we don't need this extra validation here in light of the recent updates to LogicalPlanner to check this condition up front, but it may make sense to have a duplicate check as a second guard in the physical layer here as well.

ksqldb-streams/src/test/java/io/confluent/ksql/execution/streams/ForeignKeyTableTableJoinBuilderTest.java
```diff
@@ -0,0 +1,315 @@
...
+        eq(new KsqlKeyExtractor<>(1)),
+        eq(new KsqlValueJoiner(LEFT_SCHEMA.value().size(), RIGHT_SCHEMA.value().size(), 0))
+    );
+    verifyNoMoreInteractions(leftKTable, rightKTable, resultKTable);
+    assertThat(result.getTable(), is(resultKTable));
+    assertThat(result.getExecutionKeyFactory(), is(executionKeyFactory));
+  }
+
+  @Test
+  public void shouldDoLeftJoinOnKey() {
```
vcrfxia, line 139: What is the purpose of this test? Isn't this a primary key join, not a foreign key join?

ksqldb-streams/src/test/java/io/confluent/ksql/execution/streams/ForeignKeyTableTableJoinBuilderTest.java
```diff
@@ -0,0 +1,315 @@
...
+        eq(new KsqlKeyExtractor<>(1)),
+        eq(new KsqlValueJoiner(LEFT_SCHEMA.value().size(), RIGHT_SCHEMA.value().size(), 0))
+    );
+    verifyNoMoreInteractions(leftKTable, rightKTable, resultKTable);
+    assertThat(result.getTable(), is(resultKTable));
+    assertThat(result.getExecutionKeyFactory(), is(executionKeyFactory));
+  }
+
+  @Test
+  public void shouldDoInnerJoinOnKey() {
```
vcrfxia, line 196: Same question as above: What is the purpose of this test? Isn't this a primary key join, not a foreign key join?

ksqldb-streams/src/test/java/io/confluent/ksql/execution/streams/ForeignKeyTableTableJoinBuilderTest.java
```diff
@@ -0,0 +1,315 @@
...
+        result.getSchema(),
+        is(LogicalSchema.builder()
+            .keyColumns(LEFT_SCHEMA_MULTI_KEY.key())
+            .valueColumns(LEFT_SCHEMA_MULTI_KEY.value())
+            .valueColumns(RIGHT_SCHEMA.value())
+            .build())
+    );
+  }
+
+  // not sure what "legacy schema" is and if we need to support it?
```
vcrfxia, line 272: I don't think this is necessary. `LEGACY_KEY_COL` refers to the old default output key column name for joins. For FK joins, the output key columns always match the left source key columns.

ksqldb-execution/src/main/java/io/confluent/ksql/execution/plan/ForeignKeyTableTableJoin.java
```diff
@@ -0,0 +1,110 @@
+/*
+ * Copyright 2019 Confluent Inc.
```
vcrfxia, line 2: nit: update year (and similarly for all the other new files in this PR)

vcrfxia (commented):
Thanks @mjsax . LGTM minus the comments inline.

ksqldb-streams/src/main/java/io/confluent/ksql/execution/streams/ForeignKeyJoinParamsFactory.java
```diff
@@ -0,0 +1,63 @@
...
+    final Builder builder = LogicalSchema.builder()
+        .keyColumns(leftSchema.key())
```
vcrfxia, line 46: It occurs to me now that these column names might be aliased, which we probably don't want. Can we verify? Can also defer to a later PR when we wire everything together but my current suspicion is that this will be a problem.

ksqldb-streams/src/main/java/io/confluent/ksql/execution/streams/ForeignKeyJoinParamsFactory.java
```diff
@@ -0,0 +1,63 @@
...
+        .valueColumns(rightSchema.value());
+
+    return builder.build();
+  }
+
+  private static <KRightT> KsqlKeyExtractor<KRightT> createKeyExtractor(
+      final LogicalSchema leftSchema,
+      final ColumnName leftJoinColumnName) {
+
+    final Optional<Column> leftJoinColumn = leftSchema.findValueColumn(leftJoinColumnName);
```
guozhangwang, line 57: Just for my understanding, is it possible that due to alias we would not find a match here? I.e. if we do

```
... SELECT A.a AS a1, ... WHERE A.a = B.b
```

would we try to find `a` in a schema with `a1` then?

ksqldb-streams/src/main/java/io/confluent/ksql/execution/streams/ForeignKeyTableTableJoinBuilder.java
```diff
@@ -0,0 +1,66 @@
...
+            right.getTable(),
+            joinParams.getKeyExtractor(),
+            joinParams.getJoiner()
+        );
+        break;
+      default:
+        throw new IllegalStateException("invalid join type");
+    }
+
+    return KTableHolder.unmaterialized(
```
guozhangwang, line 60: Is it intentional to not materialize the join results?

ksqldb-streams/src/main/java/io/confluent/ksql/execution/streams/ForeignKeyTableTableJoinBuilder.java
```diff
@@ -0,0 +1,66 @@
...
+        break;
+      case INNER:
+        result = left.getTable().join(
+            right.getTable(),
+            joinParams.getKeyExtractor(),
+            joinParams.getJoiner()
+        );
+        break;
+      default:
+        throw new IllegalStateException("invalid join type");
```
guozhangwang, line 57: nit: also include the type enum?

ksqldb-streams/src/main/java/io/confluent/ksql/execution/streams/KsqlKeyExtractor.java
```diff
@@ -0,0 +1,66 @@
...
+
+package io.confluent.ksql.execution.streams;
+
+import static com.google.common.base.Preconditions.checkArgument;
+
+import io.confluent.ksql.GenericRow;
+import java.util.Objects;
+import java.util.function.Function;
+
+public final class KsqlKeyExtractor<KRightT> implements Function<GenericRow, KRightT> {
```
guozhangwang, line 24: Just a question here: I saw different Extractor classes in the repo, and some with a `Ksql` prefix in its name and some do not. Is there a convention for the naming at the moment? cc @agavra @vcrfxia

guozhangwang (commented):
Made a pass over the non-testing part.

ksqldb-execution/src/main/java/io/confluent/ksql/execution/plan/ForeignKeyTableTableJoin.java
mjsax: What about older files? How do we handle this?

ksqldb-streams/src/main/java/io/confluent/ksql/execution/streams/ForeignKeyJoinParamsFactory.java
mjsax: > We need to validate that the join expression types match. If we follow the pattern of the existing JoinParamsFactory, we can do that here.

Yes, but this should be done in the logical planner IMHO.

Happy to add a second guard.

ksqldb-streams/src/main/java/io/confluent/ksql/execution/streams/ForeignKeyTableTableJoinBuilder.java
mjsax: We could, but it seems cleaner to have a dedicated node for FK-joins in the physical plan, as mentioned by Victory above.

ksqldb-streams/src/test/java/io/confluent/ksql/execution/streams/ForeignKeyTableTableJoinBuilderTest.java
mjsax: I guess you are right -- seems I overshot on corner case testing. Even if we could compile a PK_PK-join as a FK-join and it should still work -- it would be less efficient though I guess.

This actually raises a question: if we have two tables with different partition count and/or different key format, could we implement a PK-PK-join without repartitioning one input table (or even rejecting the join do the missing co-partitioning), but compiling it down as FK-join?

Thus, I am wondering if we should just keep this test, because from a physical plan point of view, this should work, even if we don't use it when translating the logical plan into a physical plan atm -- but maybe we would do this in the future?

ksqldb-streams/src/main/java/io/confluent/ksql/execution/streams/ForeignKeyTableTableJoinBuilder.java
mjsax: It's was kinda intentional, because `TableTableJoinBuilder()` does the same thing. Not sure what the status of materializing table-table joins result is. \cc @cprasad1

ksqldb-streams/src/main/java/io/confluent/ksql/execution/streams/ForeignKeyJoinParamsFactory.java
mjsax: Not 100% sure atm -- `leftSchema.key()`returns `Column` that only has a `ColumnName` but no "alias".

But yes, we need to keep aliases in mind... I think it might be best to defer to later.

ksqldb-streams/src/main/java/io/confluent/ksql/execution/streams/ForeignKeyJoinParamsFactory.java
mjsax: Cf https://github.com/confluentinc/ksql/pull/7517/files#r632182509

ksqldb-streams/src/main/java/io/confluent/ksql/execution/streams/KsqlKeyExtractor.java
agavra: 🔥 ⚠️ - I really don't like `Ksql` prefix. I don't think we have a convention, so it's really just do what you like.

ksqldb-streams/src/main/java/io/confluent/ksql/execution/streams/ForeignKeyTableTableJoinBuilder.java
cprasad1: We skip materializing the Table-Table join results immediately. We materialize at the `TableSelectBuilder` which comes after the `TableTableJoinBuilder()`. It makes sense to follow the same pattern and skip materializing here.

cprasad1: Specifically, we materialize here: https://github.com/confluentinc/ksql/blob/100767d73dfe013a41745ba540532b717261eeb8/ksqldb-streams/src/main/java/io/confluent/ksql/execution/streams/TableSelectBuilder.java#L114

ksqldb-streams/src/main/java/io/confluent/ksql/execution/streams/ForeignKeyJoinParamsFactory.java
vcrfxia: Where are we currently validating that the expression types match? I don't see the check in either the logical or the physical layer right now.

ksqldb-streams/src/test/java/io/confluent/ksql/execution/streams/ForeignKeyTableTableJoinBuilderTest.java
vcrfxia: > if we have two tables with different partition count and/or different key format, could we implement a PK-PK-join without repartitioning one input table (or even rejecting the join do the missing co-partitioning), but compiling it down as FK-join?

I suppose we could but it seems odd to do this by default, as it allows users to inadvertently write queries with poorer performance than expected. I'm not convinced we need these unit tests until (and if) we decide to do this one day, but I don't feel strongly. If we're going to leave them in, can we at least add a comment explaining that these are actually primary key joins which won't reach this code today, but we've left it in the future? Otherwise future readers may be similarly confused.

vcrfxia (approved):
LGTM besides the couple minor comments I just added. Thanks @mjsax !

ksqldb-streams/src/main/java/io/confluent/ksql/execution/streams/ForeignKeyJoinParamsFactory.java
mjsax: For table-table joins, it happens in `JoinParamsFactory.throwOnKeyMismatch` (but I think it's wrong to do this check when building the physical plan, but it should be done earlier.

Thus, I did not add it to `ForeignKeyJoinParamsFactory` and I think @spena current WIP PR should add this check. (And we should move the existing check for PK-joins in a follow up PR into the logical planner.)

ksqldb-streams/src/main/java/io/confluent/ksql/execution/streams/ForeignKeyTableTableJoinBuilder.java
guozhangwang: Got it, thanks @cprasad1 for confirming!

ksqldb-streams/src/test/java/io/confluent/ksql/execution/streams/ForeignKeyTableTableJoinBuilderTest.java
mjsax: Agreed. Also not sure _if_ we should do it, but from a unit-test point of view, it seem ok to keep the tests, as unit testing the physical plan should not know anything about the logical plan layer.

Will add a comment to clarify.

ksqldb-streams/src/main/java/io/confluent/ksql/execution/streams/ForeignKeyJoinParamsFactory.java
vcrfxia: `JoinParamsFactory.throwOnKeyMismatch()` is called from `JoinParamsFactory.createSchema()`, which is called from `JoinNode.buildJoinSchema()`, which is called as part of attempting to build the join from the logical planner. In other words, this check _is_ already happening in the logical planner (for table-table joins today). Where are you proposing that the check should happen instead?

mjsax: That is quite convoluted code than, because `JoinParamsFactory` is in package `io.confluent.ksql.execution.streams` -- the logical planner should not use helpers from the physical planner IMHO. Maybe the solution is to move the whole factory class into the logical planner package?

vcrfxia: The method is also called from the physical layer when executing the execution step. Ideally this method would be in a separate package clearly meant to be shared between the two, rather than entirely in one or the other.

mjsax: Interesting. Think we don't need to resolve this in this PR. We only need to make sure that @spena PR's adds the type check.

PR #7521, opened 2021-05-13
chore: ksqlDB java client can extend Closeable

wicknicks:
Signed-off-by: Arjun Satish <arjun@confluent.io>

### Description 
ksqldb java client can extend the java.io.closeable interface. 

### Testing done 
clean run of unit tests.

### Reviewer checklist

### DIFF
--- a/ksqldb-api-client/src/main/java/io/confluent/ksql/api/client/Client.java
+++ b/ksqldb-api-client/src/main/java/io/confluent/ksql/api/client/Client.java
@@ -17,6 +17,7 @@ package io.confluent.ksql.api.client;
 
 import io.confluent.ksql.api.client.impl.ClientImpl;
 import io.vertx.core.Vertx;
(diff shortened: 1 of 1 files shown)


vcrfxia (commented):
Thanks @wicknicks . Can we remove the definition of `void close();` in the Client interface, as it's now redundant? LGTM otherwise.

wicknicks:
@vcrfxia the [close() from Closable](https://docs.oracle.com/javase/8/docs/api/java/io/Closeable.html#close--) throws an IOException. I think we want to keep the definition to indicate this method does not throw that exception (will help with backward compatibility).

vcrfxia:
Ah got'cha. This approach sounds good then. Will merge once the PR builder passes.

vcrfxia (approved):

guozhangwang (commented):
LGTM~

PR #7522, opened 2021-05-13
chore: Update JoinNode with FK column name

spena:
### Description 
This PR passes the FK column name detected by the `LogicalPlanner` to the `JoinNode` for further processing. Later on the physical plan can get the FK information from the `JoinNode` to pass it to the right join method.

As part of the PR, I modified the `buildJoinKey` to prevent setting the internal `keyColumn` to a FK join. The resulted `JoinKey` must have only a primary key, which is later used to build the resulted schema with the desired key.

### Testing done 
Updated unit tests

### Reviewer checklist

### DIFF
--- a/ksqldb-engine/src/main/java/io/confluent/ksql/planner/JoinTree.java
+++ b/ksqldb-engine/src/main/java/io/confluent/ksql/planner/JoinTree.java
@@ -23,9 +23,7 @@ import io.confluent.ksql.analyzer.Analysis.AliasedDataSource;
 import io.confluent.ksql.analyzer.Analysis.JoinInfo;
 import io.confluent.ksql.execution.expression.tree.Expression;
 import io.confluent.ksql.execution.expression.tree.QualifiedColumnReferenceExp;
-import io.confluent.ksql.name.ColumnName;
 import io.confluent.ksql.planner.plan.JoinNode.JoinType;
-import io.confluent.ksql.schema.ksql.Column;
 import io.confluent.ksql.schema.utils.FormatOptions;
 import io.confluent.ksql.util.KsqlException;
 import java.util.List;
@@ -264,31 +262,15 @@ final class JoinTree {
       return Streams
           .concat(
               left.viableKeyColumns().stream(),
-              emptyIfNonKeyColumn(info.getLeftJoinExpression(), info.getLeftSource()),
+              Stream.of(info.getLeftJoinExpression()),
               right.viableKeyColumns().stream(),
-              emptyIfNonKeyColumn(info.getRightJoinExpression(), info.getRightSource())
+              Stream.of(info.getRightJoinExpression())
           )
           .filter(e -> e instanceof QualifiedColumnReferenceExp)
           .map(QualifiedColumnReferenceExp.class::cast)
           .distinct()
           .filter(equiv::contains)
           .collect(Collectors.toList());
-    }
-
-    private Stream<Expression> emptyIfNonKeyColumn(final Expression expression,
-                                                   final AliasedDataSource aliasedDataSource) {
-      if (!(expression instanceof QualifiedColumnReferenceExp)) {
-        return Stream.empty();
-      }
-
-      final QualifiedColumnReferenceExp columnReference = (QualifiedColumnReferenceExp) expression;
-      final ColumnName columnName =  columnReference.getColumnName();
-      final List<Column> keyColumns = aliasedDataSource.getDataSource().getSchema().key();
-
-      final boolean isKeyColumn = keyColumns.stream().map(Column::name)
-          .anyMatch(key -> key.equals(columnName));
-
-      return isKeyColumn ? Stream.of(expression) : Stream.empty();
     }
 
     @Override
--- a/ksqldb-engine/src/main/java/io/confluent/ksql/planner/JoinTree.java
+++ b/ksqldb-engine/src/main/java/io/confluent/ksql/planner/JoinTree.java
@@ -23,7 +23,9 @@ import io.confluent.ksql.analyzer.Analysis.AliasedDataSource;
 import io.confluent.ksql.analyzer.Analysis.JoinInfo;
 import io.confluent.ksql.execution.expression.tree.Expression;
 import io.confluent.ksql.execution.expression.tree.QualifiedColumnReferenceExp;
+import io.confluent.ksql.name.ColumnName;
 import io.confluent.ksql.planner.plan.JoinNode.JoinType;
+import io.confluent.ksql.schema.ksql.Column;
 import io.confluent.ksql.schema.utils.FormatOptions;
 import io.confluent.ksql.util.KsqlException;
 import java.util.List;
@@ -262,15 +264,31 @@ final class JoinTree {
       return Streams
           .concat(
               left.viableKeyColumns().stream(),
-              Stream.of(info.getLeftJoinExpression()),
+              emptyIfNonKeyColumn(info.getLeftJoinExpression(), info.getLeftSource()),
               right.viableKeyColumns().stream(),
-              Stream.of(info.getRightJoinExpression())
+              emptyIfNonKeyColumn(info.getRightJoinExpression(), info.getRightSource())
           )
           .filter(e -> e instanceof QualifiedColumnReferenceExp)
           .map(QualifiedColumnReferenceExp.class::cast)
           .distinct()
           .filter(equiv::contains)
           .collect(Collectors.toList());
+    }
+
+    private Stream<Expression> emptyIfNonKeyColumn(final Expression expression,
+                                                   final AliasedDataSource aliasedDataSource) {
+      if (!(expression instanceof QualifiedColumnReferenceExp)) {
+        return Stream.empty();
+      }
+
+      final QualifiedColumnReferenceExp columnReference = (QualifiedColumnReferenceExp) expression;
+      final ColumnName columnName =  columnReference.getColumnName();
+      final List<Column> keyColumns = aliasedDataSource.getDataSource().getSchema().key();
+
+      final boolean isKeyColumn = keyColumns.stream().map(Column::name)
+          .anyMatch(key -> key.equals(columnName));
+
+      return isKeyColumn ? Stream.of(expression) : Stream.empty();
     }
 
     @Override
--- a/ksqldb-engine/src/main/java/io/confluent/ksql/planner/LogicalPlanner.java
+++ b/ksqldb-engine/src/main/java/io/confluent/ksql/planner/LogicalPlanner.java
@@ -819,17 +819,12 @@ public class LogicalPlanner {
             .map(c -> new QualifiedColumnReferenceExp(alias, c.name()))
             .collect(Collectors.toList());
 
-    final QualifiedColumnReferenceExp keyColumn = leftSourceKeys.get(0);
-
-    final ColumnName keyColumnName = ColumnNames
-        .generatedJoinColumnAlias(keyColumn.getQualifier(), keyColumn.getColumnName());
-
     final ColumnName foreignKeyColumnName = columnRef.maybeQualifier().isPresent()
         ? ColumnNames.generatedJoinColumnAlias(
             columnRef.maybeQualifier().get(), columnRef.getColumnName())
         : columnRef.getColumnName();
 
-    return JoinKey.foreignKeyColumn(keyColumnName, foreignKeyColumnName, leftSourceKeys);
+    return JoinKey.foreignKeyColumn(foreignKeyColumnName, leftSourceKeys);
   }
 
   private JoinKey buildJoinKey(final Join join) {
--- a/ksqldb-engine/src/main/java/io/confluent/ksql/planner/LogicalPlanner.java
+++ b/ksqldb-engine/src/main/java/io/confluent/ksql/planner/LogicalPlanner.java
@@ -593,12 +593,17 @@ public class LogicalPlanner {
       );
     }
 
-    final Optional<ColumnName> fkColumnName =
+    final Optional<ColumnReferenceExp> fkColumnName =
         verifyJoin(root.getInfo(), preRepartitionLeft, preRepartitionRight);
 
     final boolean finalJoin = prefix.isEmpty();
 
-    final JoinKey joinKey = buildJoinKey(root, fkColumnName);
+    final JoinKey joinKey;
+    if (fkColumnName.isPresent()) {
+      joinKey = buildForeignJoinKey(root, fkColumnName.get());
+    } else {
+      joinKey = buildJoinKey(root);
+    }
 
     final PlanNode left = prepareSourceForJoin(
         root.getLeft(),
@@ -619,7 +624,6 @@ public class LogicalPlanner {
         new PlanNodeId(prefix + "Join"),
         root.getInfo().getType(),
         joinKey.rewriteWith(refRewriter::process),
-        fkColumnName,
         finalJoin,
         left,
         right,
@@ -629,10 +633,10 @@ public class LogicalPlanner {
   }
 
   /**
-   * @return the foreign key ColumnName if this is a foreign key join
+   * @return the foreign key column if this is a foreign key join
    */
   // CHECKSTYLE_RULES.OFF: CyclomaticComplexity
-  private Optional<ColumnName> verifyJoin(
+  private Optional<ColumnReferenceExp> verifyJoin(
       final JoinInfo joinInfo,
       final PlanNode leftNode,
       final PlanNode rightNode
@@ -718,7 +722,7 @@ public class LogicalPlanner {
             ));
           }
 
-          return Optional.of(((ColumnReferenceExp) leftExpression).getColumnName());
+          return Optional.of(((ColumnReferenceExp) leftExpression));
         } else {
           throw new KsqlException(String.format(
               "Invalid join condition:"
@@ -807,7 +811,28 @@ public class LogicalPlanner {
     throw new IllegalStateException("Unknown node type: " + node.getClass().getName());
   }
 
-  private JoinKey buildJoinKey(final Join join, final Optional<ColumnName> fkColumnName) {
+  private JoinKey buildForeignJoinKey(final Join join, final ColumnReferenceExp columnRef) {
+    final AliasedDataSource leftSource = join.getInfo().getLeftSource();
+    final SourceName alias = leftSource.getAlias();
+    final List<QualifiedColumnReferenceExp> leftSourceKeys =
+        leftSource.getDataSource().getSchema().key().stream()
+            .map(c -> new QualifiedColumnReferenceExp(alias, c.name()))
+            .collect(Collectors.toList());
+
+    final QualifiedColumnReferenceExp keyColumn = leftSourceKeys.get(0);
+
+    final ColumnName keyColumnName = ColumnNames
+        .generatedJoinColumnAlias(keyColumn.getQualifier(), keyColumn.getColumnName());
+
+    final ColumnName foreignKeyColumnName = columnRef.maybeQualifier().isPresent()
+        ? ColumnNames.generatedJoinColumnAlias(
+            columnRef.maybeQualifier().get(), columnRef.getColumnName())
+        : columnRef.getColumnName();
+
+    return JoinKey.foreignKeyColumn(keyColumnName, foreignKeyColumnName, leftSourceKeys);
+  }
+
+  private JoinKey buildJoinKey(final Join join) {
     final List<QualifiedColumnReferenceExp> viableKeyColumns = join.viableKeyColumns();
     if (viableKeyColumns.isEmpty()) {
       return JoinKey.syntheticColumn();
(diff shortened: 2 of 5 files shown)


ksqldb-engine/src/main/java/io/confluent/ksql/planner/plan/JoinNode.java
```diff
@@ -138,6 +141,10 @@ public void resolveKeyFormats() {
     setKeyFormat(joinKeyFormat);
   }
 
+  public Optional<ColumnName> getLeftForeignKeyColumn() {
```
spena, line 144: This will be called later by the `join()` method to pass the FK column to it.

ksqldb-engine/src/main/java/io/confluent/ksql/planner/LogicalPlanner.java
```diff
@@ -600,6 +600,11 @@ private JoinNode buildJoin(final Join root, final String prefix, final boolean i
 
     final JoinKey joinKey = buildJoinKey(root);
 
+    Optional<ColumnName> leftJoinNonKeyColumn = Optional.empty();
+    if (isForeignKeyJoin) {
+      leftJoinNonKeyColumn = getLeftJoinNonKeyColumn(root);
```
mjsax, line 605: Could we embed this into `verifyJoin` and change it's return type from `boolean` to `Optional<ColumnName>` ? If the optional is empty, it's a PK join, otherwise it's a FK-join? (Or would this be too messy?)

ksqldb-engine/src/main/java/io/confluent/ksql/planner/LogicalPlanner.java
```diff
@@ -627,6 +633,23 @@ private JoinNode buildJoin(final Join root, final String prefix, final boolean i
     );
   }
 
+  private Optional<ColumnName> getLeftJoinNonKeyColumn(final Join join) {
+    // This method is called after the verifyJoin() has detected a FK join. This code does not
+    // check for values are in the right order. It is assumed verifyJoin() already checked them
+    // and failed in case they are not valid.
```
mjsax, line 639: Looking into `verifyJoin` we also only check if `joinOnNonKeyAttribute(rightExpression, rightNode)` and `joinOnNonKeyAttribute(leftExpression, leftNode)` -- guess we will need to make this more flexible...

Overall, it seems that `getLeftJoinNonKeyColumn` is somewhat similar what adds to my previous comment to push this inside `verifyJoin` (or in particular `joinOnNonKeyAttribute`) ?

ksqldb-engine/src/main/java/io/confluent/ksql/planner/plan/JoinNode.java
```diff
@@ -111,9 +113,10 @@ public JoinNode(
         Optional.empty()
     );
 
-    this.schema = buildJoinSchema(joinKey, left, right);
     this.joinType = requireNonNull(joinType, "joinType");
     this.joinKey = requireNonNull(joinKey, "joinKey");
+    this.schema = buildJoinSchema(joinKey, left, right);
```
mjsax, line 118: nit: Why did you move this line?

ksqldb-engine/src/main/java/io/confluent/ksql/planner/JoinTree.java
```diff
@@ -262,9 +264,9 @@ public String debugString(final int indent) {
       return Streams
           .concat(
               left.viableKeyColumns().stream(),
-              Stream.of(info.getLeftJoinExpression()),
+              emptyIfNonKeyColumn(info.getLeftJoinExpression(), info.getLeftSource()),
```
vcrfxia, line 267: I don't think we should be updating `viableKeyColumns()` at all. I think we should change the logic in `buildJoinKey()` to have a separate pathway for foreign key joins, since the key columns for FK joins are always simply the key columns of the left source.

However, we do need to update `joinEquivalenceSet()` to reflect that the join expressions (definitely at least the left one) should not be added to the equivalence set in the case of FK joins. Otherwise the detection of PK vs FK joins downstream (for n-way joins) will be messed up, since the logic will mistakenly think that the output is partitioned on the foreign key column, which it's not.

ksqldb-engine/src/main/java/io/confluent/ksql/planner/LogicalPlanner.java
```diff
@@ -627,6 +633,23 @@ private JoinNode buildJoin(final Join root, final String prefix, final boolean i
...
+
+    // What if expression is not ColumnReferenceExp?
```
vcrfxia, line 641: We should throw for now, with a helpful error message. We can update this later when we want to support expressions (rather than limit FK joins to individual columns).

ksqldb-engine/src/main/java/io/confluent/ksql/planner/LogicalPlanner.java
```diff
@@ -627,6 +633,23 @@ private JoinNode buildJoin(final Join root, final String prefix, final boolean i
...
+    final ColumnReferenceExp columnReference =
+        (ColumnReferenceExp) join.getInfo().getLeftJoinExpression();
+
+    final Optional<SourceName> qualifier = columnReference.maybeQualifier();
+    final ColumnName columnName = columnReference.getColumnName();
+
+    return Optional.of(qualifier.isPresent()
+        ? ColumnNames.generatedJoinColumnAlias(qualifier.get(), columnName)
```
vcrfxia, line 649: Why are we doing this? The column name needs to appear in the left source schema else the code in the physical later will fail. Unless I'm misunderstanding what's happening here, we should not be rewriting the column name.

ksqldb-engine/src/main/java/io/confluent/ksql/planner/LogicalPlanner.java
vcrfxia: If my understanding in my comment above regarding the use of qualifiers in `getLeftJoinNonKeyColumn()` is correct and that code is indeed not needed, then `getLeftJoinNonKeyColumn()` becomes very simple and my vote is also to embed it into `verifyJoin()` as Matthias has suggested.

ksqldb-engine/src/main/java/io/confluent/ksql/planner/JoinTree.java
spena: I looked at the changes, but I don't see the difference with ignoring FK expressions between `viableKeyColumns()` and `joinEquivalenceSet()`. The `joinEquivalenceSet()` returns columns and expressions. Then, the `viableKeyColumns()` filters the result to keep only columns. I just added the extra filter to remove FK columns. If I update the `joinEquivalenceSet()` to not add FK columns, then it won't return the expressions ,which is what `viableKeyColumns()` is filtering out.

I thought `viableKeyColumns()` was the method to do it 'cause it returns keys. Not necessary a foreign-key, but a primary key. A key that can be used to repartition only (perhaps we can call it viablePrimaryKeyColumns?). Also, it was easier to check if the column was a key in the schema. In `buildJoinKey`, I would need to look for the source that matches the alias + table name, then get the schema. Not a big change, but an extra thing to do.

ksqldb-engine/src/main/java/io/confluent/ksql/planner/LogicalPlanner.java
spena: Done. It was easier because now I return `return Optional.of(((ColumnReferenceExp) leftExpression).getColumnName());` in the `verifyJoin`. I'd like to rename that method now, but I'm not sure what name to use. It verifies a join is correct, either PK or FK, but also returns the FK column name? what name to use? Maybe we can do the rename later.

ksqldb-engine/src/main/java/io/confluent/ksql/planner/LogicalPlanner.java
spena: Just moved it to the verifyJoin.

ksqldb-engine/src/main/java/io/confluent/ksql/planner/plan/JoinNode.java
spena: I moved it after the `joinKey` was check for null, otherwise we'll get an NPE inside the `buildJoinSchema`.

ksqldb-engine/src/main/java/io/confluent/ksql/planner/LogicalPlanner.java
spena: Added the exception inside the `verifyJoin`.

ksqldb-engine/src/main/java/io/confluent/ksql/planner/LogicalPlanner.java
spena: Old code. I removed this.

ksqldb-engine/src/main/java/io/confluent/ksql/planner/JoinTree.java
spena: Actually, I just found out the issue with this code. The `EndToEndIntegrationTest.shouldSupportDroppingAndRecreatingJoinQuery` was failing because this query:
```
final String createStreamStatement = format(
        "create stream cart_event_product as "
        + "select pv.userid, pv.pageid, u.gender "
        + "from %s pv left join %s u on pv.userid=u.userid;",
        PAGE_VIEW_STREAM, USER_TABLE);
```

It is a stream-table join. The `pv.userid` is in the join expression and in the projection. The `u.userid` is not in the projection. The `pv.userid` is a non-key in the stream, but the plan treats the `pv.userid` as the key because it has the same type as the `u.userid` which is a key. In my code, I am removing `pv.userid` because it is not a key, and `u.userid` is also removed because it is not in the projection, so the test fails with an error that the key must be in the projection.

I fixed with your suggestion. I am not touching the `viablekeyColumns` anymore. I just added a filter in the `buildJoinKey` to ignore the fk key found in the join expression.

vcrfxia (commented):
Hey @spena , thanks for the updates but these changes still don't make sense to me. Don't we need to update `JoinKey` to accommodate the fact that for foreign key joins, the join expressions are no the resulting key of the join? More specifically, `JoinNode#buildJoinSchema()` needs to be updated. Currently, the `JoinKey` is inserted as the key for the join result, which is incorrect in the case of foreign key joins. Also, `JoinNode#validateKeyPresent()` needs to be updated for the same reason. Currently, the code validates that the `JoinKey` is present in the selection else an error is thrown. For foreign key joins, it is not a requirement that the join expressions are part of the selection, and it _is_ a requirement that the left source key(s) are present. 

I haven't looked too closely into how to address this yet but I think a reasonable approach is to rename `JoinKey` to `JoinExpressions` to clarify that even for FK joins, this object represents the join expressions rather than the resulting key of the join, and update the other methods (mentioned above) to properly account for the fact that the resulting key of the join is different. We'll probably want to restructure some of the methods in `JoinKey` if we do this, since a number of those methods are only relevant for primary key joins.

A similar approach is to change the interfaces of `JoinKey` to more closely match the methods that are needed by other classes and have two implementations of the interface, one for primary key joins and another for foreign key joins. If we do this, then the optional column name for `leftForeignKeyColumn` can be wrapped into the `FKJoinKey` rather than being separate, and we can continue passing a single object into `JoinNode` from the `LogicalPlanner`.

ksqldb-engine/src/main/java/io/confluent/ksql/planner/plan/JoinNode.java
```diff
@@ -668,4 +694,64 @@ public JoinKey rewriteWith(
       return this;
     }
   }
+
+  private static final class ForeignJoinKey implements JoinKey {
+    private final ColumnName keyColumn;
```
vcrfxia, line 699: What is the purpose of this key column? It looks like the only place it's being used is in `resolveKeyName()`. This method should never be called for foreign key joins, since we don't need to resolve a key name -- the key column names are exactly the names of the key columns for the left source.

ksqldb-engine/src/main/java/io/confluent/ksql/planner/LogicalPlanner.java
```diff
@@ -799,6 +811,27 @@ private boolean isInnerNode(final PlanNode node) {
...
+    final SourceName alias = leftSource.getAlias();
+    final List<QualifiedColumnReferenceExp> leftSourceKeys =
+        leftSource.getDataSource().getSchema().key().stream()
+            .map(c -> new QualifiedColumnReferenceExp(alias, c.name()))
+            .collect(Collectors.toList());
+
+    final QualifiedColumnReferenceExp keyColumn = leftSourceKeys.get(0);
+
+    final ColumnName keyColumnName = ColumnNames
+        .generatedJoinColumnAlias(keyColumn.getQualifier(), keyColumn.getColumnName());
```
vcrfxia, line 825: See my comment below about the purpose of `keyColumn` in `ForeignJoinKey`. I think this code can be removed.

ksqldb-engine/src/main/java/io/confluent/ksql/planner/LogicalPlanner.java
```diff
@@ -799,6 +811,27 @@ private boolean isInnerNode(final PlanNode node) {
...
+
+    final ColumnName foreignKeyColumnName = columnRef.maybeQualifier().isPresent()
+        ? ColumnNames.generatedJoinColumnAlias(
```
vcrfxia, line 828: Why are we rewriting the column name here? This column name needs to be a column in the left source's schema, else the physical plan generation will fail.

ksqldb-engine/src/main/java/io/confluent/ksql/planner/plan/JoinNode.java
```diff
@@ -668,4 +694,64 @@ public JoinKey rewriteWith(
...
+      return true;
+    }
+
+    @Override
+    public List<? extends Expression> getOriginalViableKeys(final LogicalSchema schema) {
+      return leftSourceKeyColumns;
+    }
+
+    @Override
+    public List<? extends Expression> getAllViableKeys(final LogicalSchema schema) {
```
vcrfxia, line 734: As discussed offline, I still don't think we should be calling this method (or the one above) from this class at all. I would prefer to throw an error if someone tries to do this, but for the sake of getting this feature in, I'm OK with merging this if it works and we have plans to clean it up later.

ksqldb-engine/src/test/java/io/confluent/ksql/planner/plan/JoinNodeTest.java
```diff
@@ -322,7 +322,7 @@ public void shouldPerformStreamToStreamOuterJoin() {
     setupStream(right, rightSchemaKStream);
 
     final JoinNode joinNode =
-        new JoinNode(nodeId, OUTER, joinKey, true, left, right, WITHIN_EXPRESSION, "KAFKA");
+        new JoinNode(nodeId, OUTER, joinKey,          true, left, right, WITHIN_EXPRESSION, "KAFKA");
```
vcrfxia, line 325: nit: white space changes in this file.

ksqldb-engine/src/main/java/io/confluent/ksql/planner/LogicalPlanner.java
spena: While debugging, I found the `columnRef.getColumnName()` did not contain the column alias. I had to generate the column again using the qualifier + column name. But now I think about it, if we know this FK column is from the left source, then there is no need to specify the alias, right? Is that what you mean that it is not necessary?

ksqldb-engine/src/main/java/io/confluent/ksql/planner/LogicalPlanner.java
```diff
@@ -710,7 +715,14 @@ private boolean verifyJoin(
             ));
           }
 
-          return true;
+          if (!(leftExpression instanceof ColumnReferenceExp)) {
```
mjsax, line 718: We should change the error message: `Invalid join condition: foreign-key table-table joins don't support expressions in the join condition yet.` (conceptually, we can support expression for the left hand side).

However the `rightExpression` _must_ refer to the PK of the right input table and we should add a corresponding check. (I will add a check for right hand side via #7543 )

ksqldb-engine/src/main/java/io/confluent/ksql/planner/LogicalPlanner.java
```diff
@@ -710,7 +715,14 @@ private boolean verifyJoin(
...
+            throw new KsqlException(String.format(
+                "Invalid join condition: foreign-key table-table joins with expressions "
+                    + "are not supported. Got %s = %s."
```
mjsax, line 721: You need to pass in `leftExpression` and `rightExpression` into `format()` method

ksqldb-engine/src/main/java/io/confluent/ksql/planner/LogicalPlanner.java
```diff
@@ -593,26 +593,31 @@ private JoinNode buildJoin(final Join root, final String prefix, final boolean i
...
-    final boolean isForeignKeyJoin =
+    final Optional<ColumnReferenceExp> fkColumnName =
         verifyJoin(root.getInfo(), preRepartitionLeft, preRepartitionRight);
 
     final boolean finalJoin = prefix.isEmpty();
 
-    final JoinKey joinKey = buildJoinKey(root);
+    final JoinKey joinKey;
+    if (fkColumnName.isPresent()) {
+      joinKey = buildForeignJoinKey(root, fkColumnName.get());
```
mjsax, line 603: Why do we need to pass `root` and `fkColumnName`? The output key should be the key of the left input, so it seems to be sufficient to pass `root.getLeft()` into this method?

ksqldb-engine/src/main/java/io/confluent/ksql/planner/LogicalPlanner.java
```diff
@@ -799,6 +811,22 @@ private boolean isInnerNode(final PlanNode node) {
     throw new IllegalStateException("Unknown node type: " + node.getClass().getName());
   }
 
+  private JoinKey buildForeignJoinKey(final Join join, final ColumnReferenceExp columnRef) {
```
mjsax, line 814: This method seems over complicated, and I am not even sure why we need it.

Seems a simple call to `JoinKey.foreignKeyColumn(leftSourceKeys);` should be sufficient? (or even push the extraction of the key columns into `foreignKeyColumn` and just pass in the "left input"?

ksqldb-engine/src/main/java/io/confluent/ksql/planner/plan/JoinNode.java
```diff
@@ -111,9 +112,11 @@ public JoinNode(
         Optional.empty()
     );
 
-    this.schema = buildJoinSchema(joinKey, left, right);
     this.joinType = requireNonNull(joinType, "joinType");
     this.joinKey = requireNonNull(joinKey, "joinKey");
+    this.schema = joinKey.isForeignKey()
```
mjsax, line 117: I think we need to refactor this later: If the purpose of `JoinKey` is to define the key of the result, it should not know anything about PK vs FK joins. (Guess it's good enough for now, to get the PR merged).

ksqldb-engine/src/main/java/io/confluent/ksql/planner/plan/JoinNode.java
```diff
@@ -668,4 +693,60 @@ public JoinKey rewriteWith(
       return this;
     }
   }
+
+  private static final class ForeignJoinKey implements JoinKey {
+    private final ColumnName foreignKeyColumn;
```
mjsax, line 698: Why do we include this variable? The purpose of `JoinKey` is to define the key of the result table, and thus it seem it does not need to know anything about the foreign key?

ksqldb-engine/src/main/java/io/confluent/ksql/planner/LogicalPlanner.java
```diff
@@ -710,7 +715,14 @@ private boolean verifyJoin(
             ));
           }
 
-          return true;
+          if (!(leftExpression instanceof ColumnReferenceExp)) {
+            throw new KsqlException(String.format(
+                "Invalid join condition: foreign-key table-table joins with expressions "
```
mjsax, line 720: We should change the error message: `Invalid join condition: foreign-key table-table joins don't support expressions on the join condition yet.` (conceptually, we can support expression for the left hand side).

Only `rightExpression` _must_ refer to the PK of the right input table and we have already a check for this in place.

ksqldb-engine/src/main/java/io/confluent/ksql/planner/LogicalPlanner.java
vcrfxia: Not only is it unnecessary, I think it is also incorrect. `generatedJoinColumnAlias()` actually changes the column name to append the qualifier in front of the column name. If the column name is `FOO` and the qualifier is `Q`, then this code means we'll be looking for `Q_FOO` in the left source schema, which does not exist. Only `FOO` exists.

vcrfxia:
There are a few unaddressed issues/comments on this PR still but I'm going to merge it anyway to unblock other work. The open issues can be addressed in https://github.com/confluentinc/ksql/pull/7543 or another follow-up.

ksqldb-engine/src/main/java/io/confluent/ksql/planner/LogicalPlanner.java
mjsax: I tried to address all open question in #7543 expect this one. Not 100% sure if I understand the scope yet.

A simple rewrite did not work, as we use `QualifiedColumnReferenceExp` while `leftSchema.key()` returns a list of `Column` object... Not sure how to translated between both and why we even use `QualifiedColumnReferenceExp` to begin wtih.

PR #7523, opened 2021-05-13
Run query anon logger

tolgadur:
Jira: [KCI-484](https://confluentinc.atlassian.net/browse/KCI-484)

### Description 
This is calling the [QueryLogger](https://github.com/confluentinc/ksql/pull/7453) from the QueryAnonymization epic whenever:
- a persistent query is created
- a persistent query is terminated
- a running query throws an exception
- a transient query is submitted 

### Testing done 
Unit testing and made sure it runs correctly.

### Reviewer checklist


ksqldb-engine/src/main/java/io/confluent/ksql/util/QueryMetadataImpl.java
```diff
@@ -174,6 +175,7 @@ public void initialize() {
   ) {
     QueryError.Type errorType = Type.UNKNOWN;
     try {
+      QueryLogger.error("Query exception", this.statementString);
```
swist, line 178: This is not a very good error message. Something like

```
Uncaught exception in query {e} 
```

(pass the exception and make it clear it's an uncaught one)

swist (changes requested):
You haven't moved the corresponding tests

ksqldb-engine/pom.xml
```diff
@@ -175,6 +175,24 @@
...
+            <artifactId>confluent-log4j-extensions</artifactId>
+            <version>7.0.0-399</version>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>io.confluent</groupId>
+            <artifactId>confluent-log4j-extensions</artifactId>
+            <version>7.0.0-399</version>
+            <scope>test</scope>
+        </dependency>
```
swist, line 195: This shouldn't be here

swist (approved):

PR #7525, opened 2021-05-14
chore: update year in license header

mjsax:
### Description 
Update the license header yeat to 2021.

### Testing done 
n/a

### Reviewer checklist

### DIFF
--- a/ksqldb-api-client/src/main/java/io/confluent/ksql/api/client/AcksPublisher.java
+++ b/ksqldb-api-client/src/main/java/io/confluent/ksql/api/client/AcksPublisher.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2020 Confluent Inc.
+ * Copyright 2021 Confluent Inc.
  *
  * Licensed under the Confluent Community License (the "License"); you may not use
  * this file except in compliance with the License.  You may obtain a copy of the
--- a/ksqldb-api-client/src/main/java/io/confluent/ksql/api/client/BatchedQueryResult.java
+++ b/ksqldb-api-client/src/main/java/io/confluent/ksql/api/client/BatchedQueryResult.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2020 Confluent Inc.
+ * Copyright 2021 Confluent Inc.
  *
  * Licensed under the Confluent Community License (the "License"); you may not use
  * this file except in compliance with the License.  You may obtain a copy of the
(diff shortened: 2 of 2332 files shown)


agavra (approved):

ksqldb-api-client/src/main/java/io/confluent/ksql/api/client/AcksPublisher.java
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright 2020 Confluent Inc.
```
ableegoldman, line 2: It's my understanding that the year in the license header of individual files is supposed to be the year the files were created, not the current year. We had this discussion on some other PR recently and that was the consensus, but of course, democracy can still be wrong 😉

mjsax: I am not sure actually, but I was told _all_ files should be updated and reflect the current year. 🤷

ableegoldman: Hm, not sure who we should ask to confirm this -- but to start with, I'm like 99% sure that you should _not_ update the date for all files, it would only be those files which were actually modified this year, as the copyright year is supposed to refer to the date of "publication". The point is to give you a sense of how long the work's copyright will last.

(For example, for individual authors it's something like "70 years after the author's death" -- in that example, updating the copyright date wouldn't matter anyways) 

From there it gets a bit more fuzzy, probably because copyright law was written way back in the day, before things like code or computers existed; and like most of the US laws, it was not kept up to date with current technology. The [US Copyright Office](https://www.copyright.gov/title17/92chap4.html)
says it should reflect "The year of first publication".  I'm a little less sure of this, but I interpret this to mean it should refer to the date of file creation. From what I understand, if you do want to extend the copyright because the file has been updated with new or modified code, then you change the date to represent a range of years from creation to last modification. 

We could just do that here, but it's going to be annoying to try and remember to update the range to the current year in every file that you touch, if said file hasn't yet been updated that year. So I would propose to just stick with the date of creation, as that is still legally correct. And I'm sure Confluent has some specific copyright clauses that assert how this works for our source code, so we're not going to lose copyright rights just because we didn't bump the year in some source code files. Plus, I'm also pretty sure that headers in source code are not legally binding copyright notices, so if we get it wrong it's just misleading and kind of rude, but not illegal. In fact they're technically not mandatory in the first place, unless maybe the registered copyright has some specific rules to enforce it.

All that said, I'm not a lawyer, so if we're really concerned about this then we should reach out to the legal team.

PR #7526, opened 2021-05-14
test: don't allow FK-outer joins and add QTT cases for FK-join with flipped join condition

mjsax:
### Description 
Add a check to disallow OUTER join for FK-joins.

Add new tests for FK-joins with "flipped" join condition. Luckily `JoinTree.build()` take care of the order by calling `join.flip()` if necessary, and thus we don't need to write additional code for FK-joins for this case. However, we should have the corresponding tests.

### Testing done 
Only added tests for binary joins. Not sure if we should add tests for n-way joins at this point, as don't plan to support n-way joins yet.

### Reviewer checklist

### DIFF
--- a/ksqldb-functional-tests/src/test/resources/query-validation-tests/fk-join.json
+++ b/ksqldb-functional-tests/src/test/resources/query-validation-tests/fk-join.json
@@ -13,11 +13,35 @@
       }
     },
     {
+      "name": "Should fail on right non-key attribute for inner-join -- revers join condition order",
+      "statements": [
+        "CREATE TABLE left_table (id1 BIGINT PRIMARY KEY, f1 BIGINT) WITH (kafka_topic='left_topic', format='JSON');",
+        "CREATE TABLE right_table (id2 BIGINT PRIMARY KEY, f2 BIGINT) WITH (kafka_topic='right_topic', format='JSON');",
+        "CREATE TABLE output AS SELECT id1, id2, f1, f2 FROM left_table JOIN right_table ON f2 = id1;"
+      ],
+      "expectedException": {
+        "type": "io.confluent.ksql.util.KsqlStatementException",
+        "message": "Invalid join condition: table-table joins require to join on the primary key of the right input table. Got LEFT_TABLE.ID1 = RIGHT_TABLE.F2."
+      }
+    },
+    {
       "name": "Should fail on right non-key attribute for left-join",
       "statements": [
         "CREATE TABLE left_table (id1 BIGINT PRIMARY KEY, f1 BIGINT) WITH (kafka_topic='left_topic', format='JSON');",
         "CREATE TABLE right_table (id2 BIGINT PRIMARY KEY, f2 BIGINT) WITH (kafka_topic='right_topic', format='JSON');",
         "CREATE TABLE output AS SELECT id1, id2, f1, f2 FROM left_table LEFT JOIN right_table ON id1 = f2;"
+      ],
+      "expectedException": {
+        "type": "io.confluent.ksql.util.KsqlStatementException",
+        "message": "Invalid join condition: table-table joins require to join on the primary key of the right input table. Got LEFT_TABLE.ID1 = RIGHT_TABLE.F2."
+      }
+    },
+    {
+      "name": "Should fail on right non-key attribute for left-join -- revers join condition order",
+      "statements": [
+        "CREATE TABLE left_table (id1 BIGINT PRIMARY KEY, f1 BIGINT) WITH (kafka_topic='left_topic', format='JSON');",
+        "CREATE TABLE right_table (id2 BIGINT PRIMARY KEY, f2 BIGINT) WITH (kafka_topic='right_topic', format='JSON');",
+        "CREATE TABLE output AS SELECT id1, id2, f1, f2 FROM left_table LEFT JOIN right_table ON f2 = id1;"
       ],
       "expectedException": {
         "type": "io.confluent.ksql.util.KsqlStatementException",
@@ -37,11 +61,35 @@
       }
     },
     {
+      "name": "Should fail on right non-key attribute for outer-join -- revers join condition order",
+      "statements": [
+        "CREATE TABLE left_table (id1 BIGINT PRIMARY KEY, f1 BIGINT) WITH (kafka_topic='left_topic', format='JSON');",
+        "CREATE TABLE right_table (id2 BIGINT PRIMARY KEY, f2 BIGINT) WITH (kafka_topic='right_topic', format='JSON');",
+        "CREATE TABLE output AS SELECT id1, id2, f1, f2 FROM left_table FULL OUTER JOIN right_table ON f2 = id1;"
+      ],
+      "expectedException": {
+        "type": "io.confluent.ksql.util.KsqlStatementException",
+        "message": "Invalid join condition: table-table joins require to join on the primary key of the right input table. Got LEFT_TABLE.ID1 = RIGHT_TABLE.F2."
+      }
+    },
+    {
       "name": "Should fail on left non-key attribute for inner-join",
       "statements": [
         "CREATE TABLE left_table (id1 BIGINT PRIMARY KEY, f1 BIGINT) WITH (kafka_topic='left_topic', format='JSON');",
         "CREATE TABLE right_table (id2 BIGINT PRIMARY KEY, f2 BIGINT) WITH (kafka_topic='right_topic', format='JSON');",
         "CREATE TABLE output AS SELECT id1, id2, f1, f2 FROM left_table JOIN right_table ON f1 = id2;"
+      ],
+      "expectedException": {
+        "type": "io.confluent.ksql.util.KsqlStatementException",
+        "message": "Invalid join condition: foreign-key table-table joins are not supported. Got LEFT_TABLE.F1 = RIGHT_TABLE.ID2."
+      }
+    },
+    {
+      "name": "Should fail on left non-key attribute for inner-join -- revers join condition order",
+      "statements": [
+        "CREATE TABLE left_table (id1 BIGINT PRIMARY KEY, f1 BIGINT) WITH (kafka_topic='left_topic', format='JSON');",
+        "CREATE TABLE right_table (id2 BIGINT PRIMARY KEY, f2 BIGINT) WITH (kafka_topic='right_topic', format='JSON');",
+        "CREATE TABLE output AS SELECT id1, id2, f1, f2 FROM left_table JOIN right_table ON id2 = f1;"
       ],
       "expectedException": {
         "type": "io.confluent.ksql.util.KsqlStatementException",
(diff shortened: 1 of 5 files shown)


ksqldb-functional-tests/src/test/resources/query-validation-tests/fk-join.json
```diff
@@ -12,6 +12,18 @@
         "message": "Invalid join condition: table-table joins require to join on the primary key of the right input table. Got LEFT_TABLE.ID1 = RIGHT_TABLE.F2."
       }
     },
+    {
+      "name": "Should fail on right non-key attribute for inner-join -- revers join condition order",
```
vcrfxia, line 16: I don't think we need copies of these tests with the feature flag disabled. If the purpose of the test is to check that table-table joins are only allowed on primary keys, then tests for that already exist. If the purpose is to check that the behavior does not change if the join condition is flipped, then that belongs in the regular `joins.json` file since it's not related to foreign key joins.

I think adding duplicates of all the new FK join tests will unnecessarily blow up this file and create more busywork for us (harder reviews, more work when we ungate the feature flag) than it's worth.

ksqldb-functional-tests/src/test/resources/query-validation-tests/fk-join.json
```diff
@@ -155,6 +251,25 @@
         "message": "Invalid join condition: table-table joins require to join on the primary key of the right input table. Got LEFT_TABLE.ID1 = RIGHT_TABLE.F2."
       }
     },
+    {
+      "name": "Should fail on right non-key attribute for inner-join (reverse join condition order) - feature flag enabled",
+      "properties": {
+        "ksql.joins.foreign.key.enable": true
+      },
+      "comments": [
```
vcrfxia, line 259: If you go with my suggestion above to remove the copies of these tests without the feature flag enabled, then we can also remove these comments. When we ungate the feature flag, we'll simply remove the `properties` block rather than deleting the test. You can also update the test name to remove the `- feature flag enabled` portion now so we don't have to update the test name later when ungating the feature flag.

ksqldb-functional-tests/src/test/resources/query-validation-tests/fk-join.json
```diff
@@ -155,6 +251,25 @@
...
+        "as it duplicates another test above (without the feature flag)"
+      ],
+      "statements": [
+        "CREATE TABLE left_table (id1 BIGINT PRIMARY KEY, f1 BIGINT) WITH (kafka_topic='left_topic', format='JSON');",
+        "CREATE TABLE right_table (id2 BIGINT PRIMARY KEY, f2 BIGINT) WITH (kafka_topic='right_topic', format='JSON');",
+        "CREATE TABLE output AS SELECT id1, id2, f1, f2 FROM left_table JOIN right_table ON f2 = id1;"
+      ],
+      "expectedException": {
+        "type": "io.confluent.ksql.util.KsqlStatementException",
+        "message": "Invalid join condition: table-table joins require to join on the primary key of the right input table. Got LEFT_TABLE.ID1 = RIGHT_TABLE.F2."
```
vcrfxia, line 270: I think it'd be marginally better if the error message showed the join expressions in the same order as in the user's query but given the structure of the code right now this is somewhat difficult to achieve so I think this is fine.

ksqldb-functional-tests/src/test/resources/query-validation-tests/fk-join.json
```diff
@@ -193,6 +327,25 @@
         "message": "Invalid join condition: table-table joins require to join on the primary key of the right input table. Got LEFT_TABLE.ID1 = RIGHT_TABLE.F2."
       }
     },
+    {
+      "name": "Should fail on right non-key attribute for outer-join (reverse join condition order) - feature flag enabled",
```
vcrfxia, line 331: Why do we need this test? FK joins aren't supported for outer joins at all.

vcrfxia (commented):
Thanks @mjsax . Comments inline.

ksqldb-functional-tests/src/test/resources/query-validation-tests/fk-join.json
mjsax: What is your concrete proposal? I think we should have some of those tests, to make sure that the condition is correctly identified even if the join condition is flipped. And yes, we should have corresponding tests in `joins.json` in addition.

ksqldb-functional-tests/src/test/resources/query-validation-tests/fk-join.json
mjsax: I had a similar though... But I think we could fix it: JoinTree could set a flag "flipped" that we preserve that allows us to flip back in the error message. Let me add this.

ksqldb-functional-tests/src/test/resources/query-validation-tests/fk-join.json
mjsax: Ah. Good catch. Seems I did to much "blind" c&p :)

ksqldb-functional-tests/src/test/resources/query-validation-tests/fk-join.json
vcrfxia: My concrete proposal is:
- delete all the new tests in this PR that do not have the feature flag enabled
- remove the comments from the copies of the tests with the feature flag enabled about deleting the test when ungating from the feature flag
- remove `- feature flag enabled` from the names of the copies of the tests that have the feature flag enabled

If we're planning to support flipped join conditions out of the box (which it sounds like we are), then we should add positive test coverage to ensure this is properly supported. I will add that into https://github.com/confluentinc/ksql/pull/7528.

mjsax: I thought about this a little bit. I agree that those this test (and it's "siblings") should not be part of fk-join test. However, `joins.json` is very long and hard to keep an overview what we test and what not. Will add `table-table-join.json` and move these 10 test over. Want to keep the reverse order as I think we should test that the condition is detected for both cases (and also that we "flip" left/right condition in the error message).

I would hope we can cleanup `join.json` eventually, but it's a little tricky to "move" test as we also need to migrate old plans accordingly. Plan tp can dig into this as a follow up.

ksqldb-functional-tests/src/test/resources/query-validation-tests/fk-join.json
```diff
@@ -1,99 +1,105 @@
 {
   "tests": [
     {
-      "name": "Should fail on right non-key attribute for inner-join",
```
mjsax, line 4: This diff is confusion... The test (and the next 3 tests) got moved to `table-table-join.json`

ksqldb-functional-tests/src/test/resources/query-validation-tests/fk-join.json
```diff
@@ -1,99 +1,105 @@
...
       ],
       "expectedException": {
         "type": "io.confluent.ksql.util.KsqlStatementException",
-        "message": "Invalid join condition: foreign-key table-table joins are not supported. Got LEFT_TABLE.F1 = RIGHT_TABLE.ID2."
+        "message": "Invalid join condition: foreign-key table-table joins are not supported. Got RIGHT_TABLE.ID2 = LEFT_TABLE.F1."
       }
     },
     {
-      "name": "Should fail on right non-key attribute for inner-join with qualifiers",
+      "name": "Should fail on outer-join",
```
mjsax, line 76: new test

ksqldb-functional-tests/src/test/resources/query-validation-tests/fk-join.json
```diff
@@ -1,99 +1,105 @@
 {
   "tests": [
     {
-      "name": "Should fail on right non-key attribute for inner-join",
+      "name": "Should fail on left non-key attribute for inner-join",
```
mjsax, line 4: existing test, just bad diff

ksqldb-functional-tests/src/test/resources/query-validation-tests/fk-join.json
```diff
@@ -1,99 +1,105 @@
...
       ],
       "expectedException": {
         "type": "io.confluent.ksql.util.KsqlStatementException",
-        "message": "Invalid join condition: table-table joins require to join on the primary key of the right input table. Got LEFT_TABLE.ID1 = RIGHT_TABLE.F2."
+        "message": "Invalid join condition: foreign-key table-table joins are not supported. Got LEFT_TABLE.F1 = RIGHT_TABLE.ID2."
       }
     },
     {
-      "name": "Should fail on right non-key attribute for left-join",
+      "name": "Should fail on left non-key attribute for inner-join -- revers join condition order",
```
mjsax, line 16: new test: duplicates the above existing test.

ksqldb-functional-tests/src/test/resources/query-validation-tests/fk-join.json
```diff
@@ -1,99 +1,105 @@
...
       ],
       "expectedException": {
         "type": "io.confluent.ksql.util.KsqlStatementException",
-        "message": "Invalid join condition: table-table joins require to join on the primary key of the right input table. Got LEFT_TABLE.ID1 = RIGHT_TABLE.F2."
+        "message": "Invalid join condition: foreign-key table-table joins are not supported. Got LEFT_TABLE.F1 = RIGHT_TABLE.ID2."
       }
     },
     {
-      "name": "Should fail on left non-key attribute for inner-join",
+      "name": "Should fail on left non-key attribute for left-join -- revers join condition order",
```
mjsax, line 40: new test: duplicates the above existing test.

ksqldb-functional-tests/src/test/resources/query-validation-tests/fk-join.json
```diff
@@ -1,99 +1,105 @@
...
+        "CREATE TABLE output AS SELECT id1, id2, f1, f2 FROM left_table FULL OUTER JOIN right_table ON f1 = id2;"
       ],
       "expectedException": {
         "type": "io.confluent.ksql.util.KsqlStatementException",
         "message": "Invalid join condition: foreign-key table-table joins are not supported. Got LEFT_TABLE.F1 = RIGHT_TABLE.ID2."
       }
     },
     {
-      "name": "Should fail on left non-key attribute for outer-join",
+      "name": "Should fail on left non-key attribute for outer-join -- revers join condition order",
```
mjsax, line 64: new test: duplicates the above existing test.

ksqldb-functional-tests/src/test/resources/query-validation-tests/fk-join.json
```diff
@@ -1,99 +1,105 @@
...
+        "CREATE TABLE output AS SELECT id1, id2, f1, f2 FROM left_table FULL OUTER JOIN right_table ON id2 = f1;"
       ],
       "expectedException": {
         "type": "io.confluent.ksql.util.KsqlStatementException",
-        "message": "Invalid join condition: foreign-key table-table joins are not supported. Got LEFT_TABLE.F1 = RIGHT_TABLE.ID2."
+        "message": "Invalid join condition: foreign-key table-table joins are not supported. Got RIGHT_TABLE.ID2 = LEFT_TABLE.F1."
       }
     },
     {
-      "name": "Should fail on right non-key attribute for inner-join with qualifiers",
```
mjsax, line 76: This test (and the next tests) got moved to `table-table-join.json`

ksqldb-functional-tests/src/test/resources/query-validation-tests/fk-join.json
```diff
@@ -1,99 +1,105 @@
...
       ],
       "expectedException": {
         "type": "io.confluent.ksql.util.KsqlStatementException",
-        "message": "Invalid join condition: table-table joins require to join on the primary key of the right input table. Got LEFT_TABLE.ID1 = RIGHT_TABLE.F2."
+        "message": "Invalid join type: full-outer join not supported for foreign-key table-table join. Got LEFT_TABLE [FULL] OUTER JOIN RIGHT_TABLE."
       }
     },
     {
-      "name": "Should fail on right non-key attribute for inner-join with alias",
+      "name": "Should fail on outer-join -- reverse join condition",
```
mjsax, line 91: new test (duplicate of the previous new test)

ksqldb-functional-tests/src/test/resources/query-validation-tests/table-table-join.json
```diff
@@ -0,0 +1,124 @@
+{
+  "tests": [
+    {
+      "name": "Should fail on right non-key attribute for inner-join",
```
mjsax, line 4: moved from `fk-join.json`

ksqldb-functional-tests/src/test/resources/query-validation-tests/table-table-join.json
```diff
@@ -0,0 +1,124 @@
...
+        "CREATE TABLE right_table (id2 BIGINT PRIMARY KEY, f2 BIGINT) WITH (kafka_topic='right_topic', format='JSON');",
+        "CREATE TABLE output AS SELECT id1, id2, f1, f2 FROM left_table JOIN right_table ON id1 = f2;"
+      ],
+      "expectedException": {
+        "type": "io.confluent.ksql.util.KsqlStatementException",
+        "message": "Invalid join condition: table-table joins require to join on the primary key of the right input table. Got LEFT_TABLE.ID1 = RIGHT_TABLE.F2."
+      }
+    },
+    {
+      "name": "Should fail on right non-key attribute for inner-join -- revers join condition order",
```
mjsax, line 16: duplicate the test (similar pattern below)

vcrfxia (approved):
I forgot this PR was still open... we're going to have a lot of merge conflicts with https://github.com/confluentinc/ksql/pull/7528 and https://github.com/confluentinc/ksql/pull/7537

Didn't look at the latest changes too carefully but LGTM based on the comments that the latest revision claims to have addressed.

PR #7528, opened 2021-05-14
test: adds tests for foreign key joins

vcrfxia:
### Description 

Many of these new tests do not pass yet as the feature is still under development, but I'm opening this PR now to get early eyes on the tests themselves (to check we're all on the same page about the expected behavior of the joins -- I'm a bit iffy on the initial positive test cases in particular) and on the test coverage.

Tests for FK joins as part of n-way joins will be added separately.

### Testing done 

:)

### Reviewer checklist

### DIFF
--- a/ksqldb-functional-tests/src/test/resources/query-validation-tests/fk-join.json
+++ b/ksqldb-functional-tests/src/test/resources/query-validation-tests/fk-join.json
@@ -295,7 +295,7 @@
       ],
       "post": {
         "sources": [
-          {"name": "OUTPUT", "type": "table", "schema": "L_ID BIGINT KEY, `R_ID` STRING, `NAME` STRING, `F1` STRING"}
+          {"name": "OUTPUT", "type": "table", "schema": "L_ID BIGINT KEY, `R_ID` BIGINT, `NAME` STRING, `F1` STRING"}
         ]
       }
     },
@@ -330,7 +330,7 @@
       ],
       "post": {
         "sources": [
-          {"name": "OUTPUT", "type": "table", "schema": "L_ID BIGINT KEY, `R_ID` STRING, `NAME` STRING, `F1` STRING"}
+          {"name": "OUTPUT", "type": "table", "schema": "L_ID BIGINT KEY, `R_ID` BIGINT, `NAME` STRING, `F1` STRING"}
         ]
       }
     },
@@ -611,7 +611,7 @@
       ],
       "post": {
         "sources": [
-          {"name": "OUTPUT", "type": "table", "schema": "L_ID BIGINT KEY, L_ID_2 BIGINT KEY, `R_ID` STRING, `NAME` STRING, `F1` STRING"}
+          {"name": "OUTPUT", "type": "table", "schema": "L_ID BIGINT KEY, L_ID_2 BIGINT KEY, `R_ID` BIGINT, `NAME` STRING, `F1` STRING"}
         ]
       }
     },
@@ -646,12 +646,12 @@
       ],
       "post": {
         "sources": [
-          {"name": "OUTPUT", "type": "table", "schema": "L_ID BIGINT KEY, L_ID_2 BIGINT KEY, `R_ID` STRING, `NAME` STRING, `F1` STRING"}
+          {"name": "OUTPUT", "type": "table", "schema": "L_ID BIGINT KEY, L_ID_2 BIGINT KEY, `R_ID` BIGINT, `NAME` STRING, `F1` STRING"}
         ]
       }
     },
     {
-      "name": "fails on partial left source key",
+      "name": "fails on partial left source key in projection",
       "properties": {
         "ksql.joins.foreign.key.enable": true
       },
@@ -706,7 +706,7 @@
       ],
       "post": {
         "sources": [
-          {"name": "OUTPUT", "type": "table", "schema": "L_ID BIGINT KEY, `R_ID` STRING, `NAME` STRING, `F1` STRING"}
+          {"name": "OUTPUT", "type": "table", "schema": "L_ID BIGINT KEY, `R_ID` BIGINT, `NAME` STRING, `F1` STRING"}
         ]
       }
     },
--- a/ksqldb-functional-tests/src/test/resources/query-validation-tests/fk-join.json
+++ b/ksqldb-functional-tests/src/test/resources/query-validation-tests/fk-join.json
@@ -232,7 +232,7 @@
       }
     },
     {
-      "name": "Should fail on n-way join (fk outer) - feature flag enabled",
+      "name": "Should fail on n-way join - feature flag enabled",
       "properties": {
         "ksql.joins.foreign.key.enable": true
       },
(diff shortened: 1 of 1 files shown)


ksqldb-functional-tests/src/test/resources/query-validation-tests/fk-join.json
```diff
@@ -268,6 +268,417 @@
...
+        {"topic": "OUTPUT", "key": 0, "value": null, "timestamp": 19000}
+      ],
+      "post": {
+        "sources": [
+          {"name": "OUTPUT", "type": "table", "schema": "LT_ID BIGINT KEY, `RT_ID` STRING, `NAME` STRING, `F1` STRING"}
+        ]
+      }
+    },
+    {
+      "name": "without join expression",
```
vcrfxia, line 403: This and associated tests will not pass until the logic in `JoinNode#validateKeyPresent()` is updated.

ksqldb-functional-tests/src/test/resources/query-validation-tests/fk-join.json
```diff
@@ -268,6 +268,417 @@
...
+        "ksql.joins.foreign.key.enable": true
+      },
+      "statements": [
+        "CREATE TABLE left_table (L_ID BIGINT PRIMARY KEY, NAME varchar, VALUE bigint) WITH (kafka_topic='left_topic', value_format='JSON');",
+        "CREATE TABLE right_table (R_ID BIGINT PRIMARY KEY, F1 varchar, F2 bigint) WITH (kafka_topic='right_topic', value_format='JSON');",
+        "CREATE TABLE OUTPUT as SELECT l_id, name, f1 FROM left_table JOIN right_table ON value + 1 = r_id;"
+      ],
+      "expectedException": {
+        "type": "io.confluent.ksql.util.KsqlStatementException",
+        "message": "TODO: some sensible error message here"
```
vcrfxia, line 459: This error will be thrown as part of https://github.com/confluentinc/ksql/pull/7522.

ksqldb-functional-tests/src/test/resources/query-validation-tests/fk-join.json
```diff
@@ -268,6 +268,417 @@
...
+        "ksql.joins.foreign.key.enable": true
+      },
+      "statements": [
+        "CREATE TABLE left_table (L_ID BIGINT PRIMARY KEY, NAME varchar, VALUE bigint) WITH (kafka_topic='left_topic', value_format='JSON');",
+        "CREATE TABLE right_table (R_ID BIGINT PRIMARY KEY, F1 varchar, F2 bigint) WITH (kafka_topic='right_topic', format='JSON');",
+        "CREATE TABLE OUTPUT as SELECT l_id, r_id, name, f1 FROM left_table FULL OUTER JOIN right_table ON value = r_id;"
+      ],
+      "expectedException": {
+        "type": "io.confluent.ksql.util.KsqlStatementException",
+        "message": "TODO: some sensible error message here"
```
vcrfxia, line 474: This error message is being added in https://github.com/confluentinc/ksql/pull/7517.

ksqldb-functional-tests/src/test/resources/query-validation-tests/fk-join.json
```diff
@@ -268,6 +268,417 @@
...
+      }
+    },
+    {
+      "name": "partition count mismatch",
```
vcrfxia, line 478: This and the key format/feature mismatch tests will not pass until a follow-up PR is opened to relax these restrictions for FK joins.

ksqldb-functional-tests/src/test/resources/query-validation-tests/fk-join.json
```diff
@@ -268,6 +268,417 @@
...
+        "ksql.joins.foreign.key.enable": true
+      },
+      "statements": [
+        "CREATE TABLE left_table (L_ID BIGINT PRIMARY KEY, NAME varchar, VALUE bigint) WITH (kafka_topic='left_topic', value_format='JSON');",
+        "CREATE TABLE right_table (R_ID BIGINT PRIMARY KEY, F1 varchar, F2 bigint) WITH (kafka_topic='right_topic', value_format='JSON');",
+        "CREATE TABLE OUTPUT as SELECT l_id, r_id, name, f1 FROM left_table JOIN right_table ON name = r_id;"
+      ],
+      "expectedException": {
+        "type": "io.confluent.ksql.util.KsqlStatementException",
+        "message": "TODO: some sensible error message here"
```
vcrfxia, line 680: This error message will be added in https://github.com/confluentinc/ksql/pull/7517 (or possible in https://github.com/confluentinc/ksql/pull/7522, based on the outcome of the discussion in https://github.com/confluentinc/ksql/pull/7517#discussion_r631984712).

ksqldb-functional-tests/src/test/resources/query-validation-tests/fk-join.json
guozhangwang: Just for my understanding: the only difference here is that we do not include r_id in `SELECT` clause here and hence validation would fail, but after the update this should still be accepted right?

ksqldb-functional-tests/src/test/resources/query-validation-tests/fk-join.json
```diff
@@ -268,6 +268,482 @@
...
+          "blacklist": ".*-repartition"
+        }
+      }
+    },
+    {
+      "name": "key format mismatch",
+      "properties": {
+        "ksql.joins.foreign.key.enable": true
+      },
+      "statements": [
```
guozhangwang, line 512: What's the difference of this one and the first "successful" one? Seems no key format mismatch existed?

ksqldb-functional-tests/src/test/resources/query-validation-tests/fk-join.json
```diff
@@ -268,6 +268,482 @@
...
+        }
+      }
+    },
+    {
+      "name": "should fail on partial right source key",
+      "properties": {
+        "ksql.joins.foreign.key.enable": true
+      },
+      "statements": [
+        "CREATE TABLE left_table (L_ID BIGINT PRIMARY KEY, NAME varchar, VALUE bigint) WITH (kafka_topic='left_topic', value_format='JSON');",
```
guozhangwang, line 579: For my education: 1) do we only allow either `format` or `key_format/value_format` to be specified, and would fail if both of them are specified? 2) in this case where only `value_format` is specified, what's the default value for `key_format`?

ksqldb-functional-tests/src/test/resources/query-validation-tests/fk-join.json
```diff
@@ -268,6 +268,482 @@
...
+        {"topic": "OUTPUT", "key": {"L_ID": 0, "L_ID_2": 100}, "value": null, "timestamp": 19000}
+      ],
+      "post": {
+        "sources": [
+          {"name": "OUTPUT", "type": "table", "schema": "L_ID BIGINT KEY, L_ID_2 BIGINT KEY, `R_ID` STRING, `NAME` STRING, `F1` STRING"}
+        ]
+      }
+    },
+    {
+      "name": "fails on partial left source key",
```
guozhangwang, line 654: Why we should fail on partial left source key? Also it seems the query below is not on partial left key, but still on value column?

ksqldb-functional-tests/src/test/resources/query-validation-tests/fk-join.json
```diff
@@ -268,6 +268,482 @@
...
+    },
+    {
+      "name": "flipped join condition - with aliases",
+      "properties": {
+        "ksql.joins.foreign.key.enable": true
+      },
+      "statements": [
+        "CREATE TABLE left_table (ID BIGINT PRIMARY KEY, NAME varchar, VALUE bigint) WITH (kafka_topic='left_topic', value_format='JSON');",
+        "CREATE TABLE right_table (ID BIGINT PRIMARY KEY, F1 varchar, F2 bigint) WITH (kafka_topic='right_topic', value_format='JSON');",
+        "CREATE TABLE OUTPUT as SELECT lt.id, rt.id, name, rt.f1 FROM left_table lt LEFT JOIN right_table rt ON rt.id = lt.value;"
```
guozhangwang, line 721: nit: maybe use `OUTER JOIN` for this case, to improve coverage?

ksqldb-functional-tests/src/test/resources/query-validation-tests/fk-join.json
vcrfxia: > 1) do we only allow either format or key_format/value_format to be specified, and would fail if both of them are specified?

Yes, precisely.

> 2) in this case where only value_format is specified, what's the default value for key_format?

The default key format is `KAFKA`. This default was chosen in order to maintain backwards compatibility from versions of ksqlDB before other key formats were introduced.

ksqldb-functional-tests/src/test/resources/query-validation-tests/fk-join.json
vcrfxia: The failure is that the left source has two key columns, but only one of them is being selected in the projection. I realize the naming of this test is confusing since for most of the other test names, the name refers to the join expression rather than the projection. I will update the name of this test to be `fails on partial left source key in projection` to clarify (though hopefully the error message, once populated, will also make this more clear).

ksqldb-functional-tests/src/test/resources/query-validation-tests/fk-join.json
vcrfxia: Foreign key outer joins are not supported. (This is a limitation of Kafka Streams, not ksqlDB.)

ksqldb-functional-tests/src/test/resources/query-validation-tests/fk-join.json
```diff
@@ -268,6 +268,482 @@
...
+        {"topic": "left_topic", "key": 1, "value": null, "timestamp": 18000}
+      ],
+      "outputs": [
+        {"topic": "OUTPUT", "key": 1, "value": {"R_ID": 0, "NAME": "zero", "F1": "blah"}, "timestamp": 10000},
+        {"topic": "OUTPUT", "key": 10, "value": {"R_ID": 0, "NAME": "bar", "F1": "a"}, "timestamp": 16000},
+        {"topic": "OUTPUT", "key": 1, "value": null, "timestamp": 18000}
+      ],
+      "post": {
+        "sources": [
+          {"name": "OUTPUT", "type": "table", "schema": "L_ID BIGINT KEY, `R_ID` STRING, `NAME` STRING, `F1` STRING"}
```
spena, line 298: Shouldn't the resulted `R_ID STRING` be `R_ID BIGINT`?

vcrfxia: Yes, good catch. Looks like I made an initial copy-paste error and then copy-pasted it throughout to the other tests as well. Will update all of them.

ksqldb-functional-tests/src/test/resources/query-validation-tests/fk-join.json
guozhangwang: Thanks! Yeah I was wondering about it about `join`, did not realize it was actually for `select`'s projection.

ksqldb-functional-tests/src/test/resources/query-validation-tests/fk-join.json
guozhangwang: Ah right, I forgot.

guozhangwang (approved):
I know we cannot merge yet, but the added tests LGTM.

ksqldb-functional-tests/src/test/resources/query-validation-tests/fk-join.json
```diff
@@ -268,6 +268,482 @@
         "type": "io.confluent.ksql.util.KsqlStatementException",
         "message": "Invalid join condition: foreign-key table-table joins are not supported as part of n-way joins. Got LEFT_TABLE.F1 = RIGHT_TABLE.ID3."
       }
+    },
+    {
+      "name": "value column - inner join",
```
mjsax, line 273: What does "value column" mean? Maybe just `Should compute inner join`

ksqldb-functional-tests/src/test/resources/query-validation-tests/fk-join.json
```diff
@@ -268,6 +268,482 @@
...
+      "properties": {
+        "ksql.joins.foreign.key.enable": true
+      },
+      "statements": [
+        "CREATE TABLE left_table (L_ID BIGINT PRIMARY KEY, NAME varchar, VALUE bigint) WITH (kafka_topic='left_topic', value_format='JSON');",
```
mjsax, line 278: As mentioned elsewhere: the formatting make it really hard to read. We should use all-upper-case only for _keywords_, an should use all-lower for identifiers... (especially the wild mix of all-lower / all-upper is super confusing)

mjsax, line 278: nit: instead of `value` call it `foreign_key` ?

ksqldb-functional-tests/src/test/resources/query-validation-tests/fk-join.json
```diff
@@ -268,6 +268,482 @@
...
+        "CREATE TABLE right_table (R_ID BIGINT PRIMARY KEY, F1 varchar, F2 bigint) WITH (kafka_topic='right_topic', value_format='JSON');",
+        "CREATE TABLE OUTPUT as SELECT l_id, r_id, name, f1 FROM left_table JOIN right_table ON value = r_id;"
```
mjsax, line 280: Based on our other discussion, putting `r_id` into SELECT should fail is we don't use `AS_VALUE(r_id)` ?

ksqldb-functional-tests/src/test/resources/query-validation-tests/fk-join.json
```diff
@@ -268,6 +268,482 @@
...
+      ],
+      "inputs": [
+        {"topic": "right_topic", "key": 0, "value": {"F1": "blah", "F2": 4}, "timestamp": 0},
+        {"topic": "left_topic", "key": 1, "value": {"NAME": "zero", "VALUE": 0}, "timestamp": 10000},
+        {"topic": "left_topic", "key": 0, "value": {"NAME": "foo", "VALUE": 100}, "timestamp": 13000},
+        {"topic": "right_topic", "key": 0, "value": {"F1": "a", "F2": 10}, "timestamp": 15000},
```
mjsax, line 286: Seems this record should trigger a result for (left input) `key = 1` ?

ksqldb-functional-tests/src/test/resources/query-validation-tests/fk-join.json
```diff
@@ -268,6 +268,482 @@
...
+        {"topic": "left_topic", "key": 10, "value": {"NAME": "bar", "VALUE": 0}, "timestamp": 16000},
+        {"topic": "right_topic", "key": 0, "value": null, "timestamp": 17000},
```
mjsax, line 288: This should delete both result record.

ksqldb-functional-tests/src/test/resources/query-validation-tests/fk-join.json
```diff
@@ -268,6 +268,482 @@
...
+        {"topic": "left_topic", "key": 1, "value": null, "timestamp": 18000}
```
mjsax, line 289: Should not produce any result

ksqldb-functional-tests/src/test/resources/query-validation-tests/fk-join.json
```diff
@@ -268,6 +268,482 @@
...
+      },
+      "statements": [
+        "CREATE TABLE left_table (L_ID BIGINT PRIMARY KEY, NAME varchar, VALUE bigint) WITH (kafka_topic='left_topic', value_format='JSON');",
+        "CREATE TABLE right_table (R_ID BIGINT PRIMARY KEY, F1 varchar, F2 bigint) WITH (kafka_topic='right_topic', value_format='JSON');",
+        "CREATE TABLE OUTPUT as SELECT l_id, r_id, name, f1 FROM left_table LEFT JOIN right_table ON value = r_id;"
+      ],
+      "inputs": [
+        {"topic": "left_topic", "key": 1, "value": {"NAME": "zero", "VALUE": 0}, "timestamp": 0},
+        {"topic": "right_topic", "key": 0, "value": {"F1": "blah", "F2": 4}, "timestamp": 10000},
+        {"topic": "left_topic", "key": 1, "value": {"NAME": "zero", "VALUE": 0}, "timestamp": 11000},
```
mjsax, line 315: This should trigger a new output record as we emit on update (not on change)?

ksqldb-functional-tests/src/test/resources/query-validation-tests/fk-join.json
```diff
@@ -268,6 +268,482 @@
...
+        {"topic": "left_topic", "key": 0, "value": {"NAME": "foo", "VALUE": 100}, "timestamp": 13000},
+        {"topic": "right_topic", "key": 0, "value": {"F1": "a", "F2": 10}, "timestamp": 15000},
```
mjsax, line 317: This should produce an output record for (left input) `key=1` ?

ksqldb-functional-tests/src/test/resources/query-validation-tests/fk-join.json
```diff
@@ -268,6 +268,482 @@
...
+        {"topic": "left_topic", "key": 10, "value": {"NAME": "bar", "VALUE": 0}, "timestamp": 16000},
+        {"topic": "right_topic", "key": 0, "value": null, "timestamp": 17000},
```
mjsax, line 319: This should not produce a tombstone but two left join results for (left input) `key=1` and `key-10`

ksqldb-functional-tests/src/test/resources/query-validation-tests/fk-join.json
```diff
@@ -268,6 +268,482 @@
...
+        {"topic": "OUTPUT", "key": 0, "value": null, "timestamp": 19000}
+      ],
+      "post": {
+        "sources": [
+          {"name": "OUTPUT", "type": "table", "schema": "LT_ID BIGINT KEY, `RT_ID` STRING, `NAME` STRING, `F1` STRING"}
+        ]
+      }
+    },
+    {
+      "name": "without join expression",
```
mjsax, line 403: The name is rather confusing: "Should allow to omit right input key columns in SELECT"

ksqldb-functional-tests/src/test/resources/query-validation-tests/fk-join.json
```diff
@@ -268,6 +268,482 @@
...
+        {"topic": "OUTPUT", "key": 1, "value": null, "timestamp": 18000}
+      ],
+      "post": {
+        "sources": [
+          {"name": "OUTPUT", "type": "table", "schema": "L_ID BIGINT KEY, `NAME` STRING, `F1` STRING"}
+        ]
+      }
+    },
+    {
+      "name": "fails without left source key",
```
mjsax, line 433: Should fail if left-input key column is missing in SELECT"

(we should duplicate this test for a multi-column left-input key, and only have a partial key in the SELECT)

ksqldb-functional-tests/src/test/resources/query-validation-tests/fk-join.json
```diff
@@ -268,6 +268,482 @@
...
+        "CREATE TABLE right_table (R_ID BIGINT PRIMARY KEY, F1 varchar, F2 bigint) WITH (kafka_topic='right_topic', value_format='JSON');",
+        "CREATE TABLE OUTPUT as SELECT r_id, name, f1 FROM left_table JOIN right_table ON value = r_id;"
+      ],
+      "expectedException": {
+        "type": "io.confluent.ksql.util.KsqlStatementException",
+        "message": "TODO: some sensible error message here"
+      }
+    },
+    {
+      "name": "non-column expression",
```
mjsax, line 448: "Should not allow expressions in join-condition for right input"

ksqldb-functional-tests/src/test/resources/query-validation-tests/fk-join.json
```diff
@@ -268,6 +268,482 @@
...
+      "properties": {
+        "ksql.joins.foreign.key.enable": true
+      },
+      "statements": [
+        "CREATE TABLE left_table (L_ID BIGINT PRIMARY KEY, NAME varchar, VALUE bigint) WITH (kafka_topic='left_topic', value_format='JSON');",
+        "CREATE TABLE right_table (R_ID BIGINT PRIMARY KEY, F1 varchar, F2 bigint) WITH (kafka_topic='right_topic', value_format='JSON');",
+        "CREATE TABLE OUTPUT as SELECT l_id, name, f1 FROM left_table JOIN right_table ON value + 1 = r_id;"
```
mjsax, line 455: The condition seems fine: we should (long term) support expression on the left-hand side FK -- we only need to disallow it for the right-hand side PK.

ksqldb-functional-tests/src/test/resources/query-validation-tests/fk-join.json
```diff
@@ -268,6 +268,482 @@
...
+      ],
+      "expectedException": {
+        "type": "io.confluent.ksql.util.KsqlStatementException",
+        "message": "TODO: some sensible error message here"
+      }
+    },
+    {
+      "name": "should fail on outer join",
```
mjsax, line 463: Added this test in #7526 already. (including the corresponding check and error message)

ksqldb-functional-tests/src/test/resources/query-validation-tests/fk-join.json
mjsax: Actually not, because we did not have the feature flag yet. But added in #7526

ksqldb-functional-tests/src/test/resources/query-validation-tests/fk-join.json
mjsax: Can we split those out into a separate PR? So we could merge all the other tests from above before we have the PR in place that does lift these restrictions?

mjsax: Also: "Should allow to join tables with different number of partitions."

ksqldb-functional-tests/src/test/resources/query-validation-tests/fk-join.json
```diff
@@ -268,6 +268,482 @@
...
+        {"topic": "OUTPUT", "key": 1, "value": null, "timestamp": 18000}
+      ],
+      "post": {
+        "topics": {
+          "blacklist": ".*-repartition"
+        }
+      }
+    },
+    {
+      "name": "key format mismatch",
```
mjsax, line 508: "Should allow to join tables with different key format"

(Btw: we should also test for the value format, as we extract the left hand side FK from the value, and it must be correctly serialized into the "subscription topic" using the right hand side key format)

ksqldb-functional-tests/src/test/resources/query-validation-tests/fk-join.json
```diff
@@ -268,6 +268,482 @@
...
+        "sources": [
+          {"name": "OUTPUT", "type": "table", "keyFormat": {"format": "KAFKA"}}
+        ],
+        "topics": {
+          "blacklist": ".*-repartition"
+        }
+      }
+    },
+    {
+      "name": "default key feature mismatch",
```
mjsax, line 541: What does this mean?

ksqldb-functional-tests/src/test/resources/query-validation-tests/fk-join.json
```diff
@@ -268,6 +268,482 @@
...
+        "sources": [
+          {"name": "OUTPUT", "type": "table", "keyFormat": {"format": "DELIMITED"}}
+        ],
+        "topics": {
+          "blacklist": ".*-repartition"
+        }
+      }
+    },
+    {
+      "name": "should fail on partial right source key",
```
mjsax, line 574: Should not allow to join against multi-key right table.

ksqldb-functional-tests/src/test/resources/query-validation-tests/fk-join.json
```diff
@@ -268,6 +268,482 @@
...
+        "CREATE TABLE right_table (R_ID BIGINT PRIMARY KEY, R_ID_2 BIGINT PRIMARY KEY, F1 varchar, F2 bigint) WITH (kafka_topic='right_topic', format='JSON');",
+        "CREATE TABLE OUTPUT as SELECT l_id, name, f1 FROM left_table JOIN right_table ON value = r_id;"
+      ],
+      "expectedException": {
+        "type": "io.confluent.ksql.util.KsqlStatementException",
+        "message": "Invalid join condition: table-table joins require to join on the primary key of the right input table. Got LEFT_TABLE.VALUE = RIGHT_TABLE.R_ID."
+      }
+    },
+    {
+      "name": "key column - inner join",
```
mjsax, line 589: ?

ksqldb-functional-tests/src/test/resources/query-validation-tests/fk-join.json
```diff
@@ -268,6 +268,482 @@
...
+        {"topic": "OUTPUT", "key": {"L_ID": 1, "L_ID_2": 0}, "value": null, "timestamp": 18000}
+      ],
+      "post": {
+        "sources": [
+          {"name": "OUTPUT", "type": "table", "schema": "L_ID BIGINT KEY, L_ID_2 BIGINT KEY, `R_ID` BIGINT, `NAME` STRING, `F1` STRING"}
+        ]
+      }
+    },
+    {
+      "name": "key column - left join",
```
mjsax, line 619: ?

ksqldb-functional-tests/src/test/resources/query-validation-tests/fk-join.json
```diff
@@ -268,6 +268,482 @@
...
+        {"topic": "OUTPUT", "key": {"L_ID": 0, "L_ID_2": 100}, "value": null, "timestamp": 19000}
+      ],
+      "post": {
+        "sources": [
+          {"name": "OUTPUT", "type": "table", "schema": "L_ID BIGINT KEY, L_ID_2 BIGINT KEY, `R_ID` BIGINT, `NAME` STRING, `F1` STRING"}
+        ]
+      }
+    },
+    {
+      "name": "fails on partial left source key in projection",
```
mjsax, line 654: Seems this is the missing test I mentioned above -- would be good to "group tests".

ksqldb-functional-tests/src/test/resources/query-validation-tests/fk-join.json
```diff
@@ -268,6 +268,482 @@
...
+        "CREATE TABLE right_table (R_ID BIGINT PRIMARY KEY, F1 varchar, F2 bigint) WITH (kafka_topic='right_topic', format='JSON');",
+        "CREATE TABLE OUTPUT as SELECT l_id, r_id, name, f1 FROM left_table JOIN right_table ON value = r_id;"
+      ],
+      "expectedException": {
+        "type": "io.confluent.ksql.util.KsqlStatementException",
+        "message": "TODO: some sensible error message here"
+      }
+    },
+    {
+      "name": "fails on type mismatch",
```
mjsax, line 669: Should fail for invalid join condition (type mismatch).

ksqldb-functional-tests/src/test/resources/query-validation-tests/fk-join.json
```diff
@@ -268,6 +268,482 @@
...
+        "CREATE TABLE right_table (R_ID BIGINT PRIMARY KEY, F1 varchar, F2 bigint) WITH (kafka_topic='right_topic', value_format='JSON');",
+        "CREATE TABLE OUTPUT as SELECT l_id, r_id, name, f1 FROM left_table JOIN right_table ON name = r_id;"
+      ],
+      "expectedException": {
+        "type": "io.confluent.ksql.util.KsqlStatementException",
+        "message": "TODO: some sensible error message here"
+      }
+    },
+    {
+      "name": "flipped join condition",
```
mjsax, line 684: If this test inner-join with flipped join condition, please move to inner-join with "regular" join condition

ksqldb-functional-tests/src/test/resources/query-validation-tests/fk-join.json
```diff
@@ -268,6 +268,482 @@
...
+        {"topic": "OUTPUT", "key": 1, "value": null, "timestamp": 18000}
+      ],
+      "post": {
+        "sources": [
+          {"name": "OUTPUT", "type": "table", "schema": "L_ID BIGINT KEY, `R_ID` BIGINT, `NAME` STRING, `F1` STRING"}
+        ]
+      }
+    },
+    {
+      "name": "flipped join condition - with aliases",
```
mjsax, line 714: as above.

ksqldb-functional-tests/src/test/resources/query-validation-tests/fk-join.json
```diff
@@ -232,7 +232,7 @@
       }
     },
     {
-      "name": "Should fail on n-way join (fk outer) - feature flag enabled",
+      "name": "Should fail on n-way join - feature flag enabled",
```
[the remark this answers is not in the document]
vcrfxia: This test is renamed in https://github.com/confluentinc/ksql/pull/7537 to be more clear.

ksqldb-functional-tests/src/test/resources/query-validation-tests/fk-join.json
vcrfxia: It means we're joining on a value column from the left source, rather than a key column. See the tests titled "partial key column" for comparison.

ksqldb-functional-tests/src/test/resources/query-validation-tests/fk-join.json
vcrfxia: I'll keep this in mind for new tests but I don't think it's worth going back and updating this for existing tests. If we really want to, we can change this after we ship the feature.

ksqldb-functional-tests/src/test/resources/query-validation-tests/fk-join.json
vcrfxia: This is fixed in https://github.com/confluentinc/ksql/pull/7537.

ksqldb-functional-tests/src/test/resources/query-validation-tests/fk-join.json
vcrfxia: Is that how FK joins work? If so I'll have to go back through and update a bunch of these tests. I didn't realize that joins could be triggered by the right side.

ksqldb-functional-tests/src/test/resources/query-validation-tests/fk-join.json
vcrfxia: You mean no-op tombstones are silently discarded? Is that true just for FK joins or in general?

ksqldb-functional-tests/src/test/resources/query-validation-tests/fk-join.json
vcrfxia: There's already an output record for this. I'm not sure what update to the expected output you're proposing.

ksqldb-functional-tests/src/test/resources/query-validation-tests/fk-join.json
vcrfxia: The name has been updated in https://github.com/confluentinc/ksql/pull/7537.

ksqldb-functional-tests/src/test/resources/query-validation-tests/fk-join.json
vcrfxia: The test you proposed already exists below. See "fails on partial left source key in projection".

ksqldb-functional-tests/src/test/resources/query-validation-tests/fk-join.json
vcrfxia: What do you mean "for right input"? The expression is coming from the left side, no?

ksqldb-functional-tests/src/test/resources/query-validation-tests/fk-join.json
vcrfxia: I agree we should support this eventually but we don't support it today, and therefore we should have a test to ensure that the user gets a sane error message. I can add a comment explaining that we plan to support this in the future.

ksqldb-functional-tests/src/test/resources/query-validation-tests/fk-join.json
vcrfxia: This PR is going to be wrecked by merge conflicts from https://github.com/confluentinc/ksql/pull/7526. I'm thinking it might make sense to just discard this PR and copy the new tests into separate PRs as we add support the different tests.

ksqldb-functional-tests/src/test/resources/query-validation-tests/fk-join.json
vcrfxia: Good point! I'll add some tests for that once I figure out how to handle the fact that this PR is going to get wrecked by merge conflicts from https://github.com/confluentinc/ksql/pull/7526

ksqldb-functional-tests/src/test/resources/query-validation-tests/fk-join.json
vcrfxia: See above. This is in contrast to the "value column" tests. These tests ensure that we can perform FK joins on key columns too.

ksqldb-functional-tests/src/test/resources/query-validation-tests/fk-join.json
vcrfxia: See above.

ksqldb-functional-tests/src/test/resources/query-validation-tests/fk-join.json
vcrfxia: The tests are grouped based on source. This test requires the left source to have multiple key columns, as do the two above.

ksqldb-functional-tests/src/test/resources/query-validation-tests/fk-join.json
```diff
@@ -268,6 +268,482 @@
...
+        {"topic": "left_topic", "key": 1, "value": null, "timestamp": 18000}
+      ],
+      "outputs": [
+        {"topic": "OUTPUT", "key": 1, "value": {"R_ID": 0, "NAME": "zero", "F1": "blah"}, "timestamp": 10000},
+        {"topic": "OUTPUT", "key": 10, "value": {"R_ID": 0, "NAME": "bar", "F1": "a"}, "timestamp": 16000},
+        {"topic": "OUTPUT", "key": 1, "value": null, "timestamp": 18000}
+      ],
+      "post": {
+        "sources": [
+          {"name": "OUTPUT", "type": "table", "schema": "L_ID BIGINT KEY, `R_ID` BIGINT, `NAME` STRING, `F1` STRING"}
```
mjsax, line 298: Why do we use 
```
`R_ID` and not just R_ID ?
```

vcrfxia:
Closing in favor of https://github.com/confluentinc/ksql/pull/7537, since that PR subsumes this one.

ksqldb-functional-tests/src/test/resources/query-validation-tests/fk-join.json
vcrfxia: Huh, interesting. I saw other tests had quoted value column names in the expected schema and assumed it was necessary, but a quick test shows it isn't so I guess we can remove them 🤷

PR #7533, opened 2021-05-17
fix: npe on transient query close (#7530)

rodesai:
Sometimes when queries close the closing thread times out and leaves behind
the cleanup thread in streams. Then, this thread calls the state change callback
which causes our metrics listener to throw an NPE. This patch changes the listener
to deal with this case by checking for null values

Cherry-pick #7530


vcrfxia (approved):
LGTM

PR #7534, opened 2021-05-17
Use Java's Base64 instead of Jersey's.

dimitarndimitrov:
### Description 

This is a partial backport of @niteshmor's https://github.com/confluentinc/ksql/pull/6702/ back to `5.2.x`. It's only partial because only 1 of the 4 main changes actually apply here.
- In `5.2.x`, the 2 changed POMs don't have the dependency from which Jetty is excluded, and the integration test doesn't exist.

Just like [its sibling change in the blueway repo](https://github.com/confluentinc/blueway/pull/2258), this change is needed because of the CVE fixes in confluentinc/rest-utils#252 and confluentinc/rest-utils#253 - the `rest-utils` changes will be breaking without this.

The change is also missing from 5.3.x, but it's probably best to merge it with `-s ours` from here up and to post a separate PR with a full cherry-pick of the #6702 changes into `5.3.x` (which again should be merged up with `-s ours`, as the original PR landed in `5.4.x` and above).

### Testing done 

- `mvn clean package -DskipTests` for the build
- `mvn -f ksql-cli/pom.xml -Dit.test=BasicAuthFunctionalTest test-compile failsafe:integration-test` for the test

### Reviewer checklist

### DIFF
--- a/ksql-cli/src/test/java/io/confluent/ksql/cli/BasicAuthFunctionalTest.java
+++ b/ksql-cli/src/test/java/io/confluent/ksql/cli/BasicAuthFunctionalTest.java
@@ -31,8 +31,10 @@ import io.confluent.ksql.test.util.EmbeddedSingleNodeKafkaCluster;
 import io.confluent.rest.RestConfig;
 import java.io.IOException;
 import java.net.URI;
+import java.nio.charset.Charset;
 import java.nio.file.Files;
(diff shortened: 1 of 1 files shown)


niteshmor (approved):
Backporting sounds good. Thanks for the PR

agavra (approved):

PR #7535, opened 2021-05-17
Backport "fix: use Java's Base64 instead of jersey's (#6702)"

dimitarndimitrov:
### Description 

This is a cherry-pick of @niteshmor's https://github.com/confluentinc/ksql/pull/6702 back to `5.3.x`. Unlike the similar change in `5.2.x` (see #7534), this is almost identical to Nitesh's original change.

Just like its sibling changes for `5.2.x` and [in the blueway repo](https://github.com/confluentinc/blueway/pull/2258), this change is needed because of the CVE fixes in confluentinc/rest-utils#252 and confluentinc/rest-utils#253 - the rest-utils changes will be breaking without this.

### Testing done 

- `mvn clean package -DskipTests` for the build
- `mvn -f ksql-rest-app/pom.xml -Dit.test=ClusterTerminationTest,KsqlResourceFunctionalTest,RestApiTest test-compile failsafe:integration-test` for the test utility

### Reviewer checklist

### DIFF
--- a/ksql-engine/pom.xml
+++ b/ksql-engine/pom.xml
@@ -144,6 +144,16 @@
             <scope>test</scope>
         </dependency>
 
+        <!-- Transitive dependency of wiremock-jre8 that's excluded in ksql-parent pom.xml
+             because it brings an older version of jetty than we'd like -->
+        <dependency>
+            <groupId>org.eclipse.jetty</groupId>
+            <artifactId>jetty-webapp</artifactId>
+            <!-- jetty.version definition comes from rest-utils -->
+            <version>${jetty.version}</version>
+            <scope>test</scope>
+        </dependency>
+
     </dependencies>
 
     <build>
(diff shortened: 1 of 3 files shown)


ksql-engine/pom.xml
```diff
@@ -144,6 +144,16 @@
...
+        <!-- Transitive dependency of wiremock-jre8 that's excluded in ksql-parent pom.xml
+             because it brings an older version of jetty than we'd like -->
+        <dependency>
+            <groupId>org.eclipse.jetty</groupId>
+            <artifactId>jetty-webapp</artifactId>
+            <!-- jetty.version definition comes from rest-utils -->
+            <version>${jetty.version}</version>
+            <scope>test</scope>
+        </dependency>
+
```
niteshmor, line 156: This seems unrelated to the Base64 change. I take it the build was already failing, and this was needed to make it pass? 
I am wondering whether this should be a separate pull request. I'll let reviewers from KSQL side make the call. (cc @agavra )

niteshmor (approved):
Looks good, but see the comment about the `pom.xml` change.

ksql-engine/pom.xml
agavra: it's fine to keep it in this PR if the build is otherwise broken. affects test dependencies only so if the build is green I'm happy with it.

agavra (approved):

Issue #6372, opened 2020-10-07
Consider lifting the requirement that a key component is selected


agavra:
Today, we require that a key is selected:
```sql
ksql> CREATE STREAM x AS SELECT col1 FROM a;
Key missing from projection.
The query used to build `X` must include the key column ID in its projection.
```

This requirement is a fallout from the decision that we should always give a name to the key, since there may be multiple columns in the key. Since we wanted every column in the key to have a name, every column in the key must also be selected. This is best described by example:

```sql
CREATE STREAM one (a INT KEY, b INT);
CREATE TABLE two AS SELECT count(*) GROUP BY a, b;
```

With this query, the key for table `two` is not a column in the input, giving us three options: (1) we could decide not to name it (2) we could, knowing the implementation, magically copy a and b into the result schema and set that as the key or (3) we could force the user to select a and b, thereby knowing that all components of the key are in the projection schema. We chose to implement (3).

NOTE: to preempt discussion about whether such a query is meaningful without selecting the keys, imagine a UDF that generated a more readable value from a and b: `SELECT better_name(a, b), count(*) GROUP BY a, b;`. This query would have the same problems as above.

**I think we should reconsider choice (3) and instead go down the path of (1).**  If the key is not named, then the key will be inaccessible until we implement `ROWID` (TODO: insert ticket here).

The solution of requiring keys means that queries that are standard SQL fail in ksqlDB with somewhat poor user experience (as evidenced by the standard select statement at the top of this issue describing current behavior).

Furthermore, it leaks details of our key implementation to the user (who should not need to know why we need a and b in the projection). Consider the following examples with expression GROUP BY and foreign key joins:

```sql
CREATE TABLE one (a INT PRIMARY KEY, b INT);
CREATE TABLE two (c INT PRIMARY KEY, d INT, e INT);
CREATE TABLE joined AS SELECT a, b, d, e FROM one JOIN two ON one.a = two.d;
```

In this example, the user wants to join two tables together but doesn’t care about the key of table two (the column c shows up nowhere in the joined statement). The implementation, however, requires that the key of the output table is the concatenation of the two input table primary keys. Making users select c to satisfy our implementation requirements is very icky… this would be the case even if we had multi-column keys implemented.

PR #7537, opened 2021-05-17
test: add test coverage for FK joins

vcrfxia:
### Description 

Many of the new tests are disabled as they will not pass until the rest of the functionality for foreign key joins is wired in. (If curious, see inline comments plus inline comments in https://github.com/confluentinc/ksql/pull/7528 for details on why.)

### Testing done 

😎 

### Reviewer checklist

### DIFF
--- a/ksqldb-functional-tests/src/test/resources/query-validation-tests/fk-join.json
+++ b/ksqldb-functional-tests/src/test/resources/query-validation-tests/fk-join.json
@@ -79,9 +79,9 @@
         "ksql.joins.foreign.key.enable": true
       },
       "statements": [
-        "CREATE TABLE left_table (L_ID BIGINT PRIMARY KEY, NAME varchar, VALUE bigint) WITH (kafka_topic='left_topic', value_format='JSON');",
-        "CREATE TABLE right_table (R_ID BIGINT PRIMARY KEY, F1 varchar, F2 bigint) WITH (kafka_topic='right_topic', value_format='JSON');",
-        "CREATE TABLE OUTPUT as SELECT l_id, r_id, name, f1 FROM left_table JOIN right_table ON value = r_id;"
+        "CREATE TABLE left_table (l_id BIGINT PRIMARY KEY, name VARCHAR, value BIGINT) WITH (kafka_topic='left_topic', value_format='JSON');",
+        "CREATE TABLE right_table (r_id BIGINT PRIMARY KEY, f1 VARCHAR, f2 BIGINT) WITH (kafka_topic='right_topic', value_format='JSON');",
+        "CREATE TABLE output AS SELECT l_id, r_id, name, f1 FROM left_table JOIN right_table ON value = r_id;"
       ],
       "inputs": [
         {"topic": "right_topic", "key": 0, "value": {"F1": "blah", "F2": 4}, "timestamp": 0},
@@ -101,7 +101,7 @@
       ],
       "post": {
         "sources": [
-          {"name": "OUTPUT", "type": "table", "schema": "L_ID BIGINT KEY, `R_ID` BIGINT, `NAME` STRING, `F1` STRING"}
+          {"name": "OUTPUT", "type": "table", "schema": "L_ID BIGINT KEY, R_ID BIGINT, NAME STRING, F1 STRING"}
         ]
       }
     },
@@ -112,9 +112,9 @@
         "ksql.joins.foreign.key.enable": true
       },
       "statements": [
-        "CREATE TABLE left_table (L_ID BIGINT PRIMARY KEY, NAME varchar, VALUE bigint) WITH (kafka_topic='left_topic', value_format='JSON');",
-        "CREATE TABLE right_table (R_ID BIGINT PRIMARY KEY, F1 varchar, F2 bigint) WITH (kafka_topic='right_topic', value_format='JSON');",
-        "CREATE TABLE OUTPUT as SELECT l_id, r_id, name, f1 FROM left_table LEFT JOIN right_table ON value = r_id;"
+        "CREATE TABLE left_table (l_id BIGINT PRIMARY KEY, name VARCHAR, value BIGINT) WITH (kafka_topic='left_topic', value_format='JSON');",
+        "CREATE TABLE right_table (r_id BIGINT PRIMARY KEY, f1 VARCHAR, f2 BIGINT) WITH (kafka_topic='right_topic', value_format='JSON');",
+        "CREATE TABLE output AS SELECT l_id, r_id, name, f1 FROM left_table LEFT JOIN right_table ON value = r_id;"
       ],
       "inputs": [
         {"topic": "left_topic", "key": 1, "value": {"NAME": "zero", "VALUE": 0}, "timestamp": 0},
@@ -140,7 +140,7 @@
       ],
       "post": {
         "sources": [
-          {"name": "OUTPUT", "type": "table", "schema": "L_ID BIGINT KEY, `R_ID` BIGINT, `NAME` STRING, `F1` STRING"}
+          {"name": "OUTPUT", "type": "table", "schema": "L_ID BIGINT KEY, R_ID BIGINT, NAME STRING, F1 STRING"}
         ]
       }
     },
@@ -151,9 +151,9 @@
         "ksql.joins.foreign.key.enable": true
       },
       "statements": [
-        "CREATE TABLE left_table (ID BIGINT PRIMARY KEY, NAME varchar, VALUE bigint) WITH (kafka_topic='left_topic', value_format='JSON');",
-        "CREATE TABLE right_table (ID BIGINT PRIMARY KEY, F1 varchar, F2 bigint) WITH (kafka_topic='right_topic', value_format='JSON');",
-        "CREATE TABLE OUTPUT as SELECT lt.id, rt.id as rt_id_alias, name as name_alias, rt.f1 FROM left_table lt JOIN right_table rt ON lt.value = rt.id;"
+        "CREATE TABLE left_table (id BIGINT PRIMARY KEY, name VARCHAR, value BIGINT) WITH (kafka_topic='left_topic', value_format='JSON');",
+        "CREATE TABLE right_table (id BIGINT PRIMARY KEY, f1 VARCHAR, f2 BIGINT) WITH (kafka_topic='right_topic', value_format='JSON');",
+        "CREATE TABLE output AS SELECT lt.id, rt.id AS rt_id_alias, name AS name_alias, rt.f1 FROM left_table AS lt JOIN right_table AS rt ON lt.value = rt.id;"
       ],
       "inputs": [
         {"topic": "right_topic", "key": 0, "value": {"F1": "blah", "F2": 4}, "timestamp": 0},
@@ -171,7 +171,7 @@
       ],
       "post": {
         "sources": [
-          {"name": "OUTPUT", "type": "table", "schema": "LT_ID BIGINT KEY, `RT_ID_ALIAS` STRING, `NAME_ALIAS` STRING, `F1` STRING"}
+          {"name": "OUTPUT", "type": "table", "schema": "LT_ID BIGINT KEY, RT_ID_ALIAS STRING, NAME_ALIAS STRING, F1 STRING"}
         ]
       }
     },
@@ -182,9 +182,9 @@
         "ksql.joins.foreign.key.enable": true
       },
       "statements": [
-        "CREATE TABLE left_table (ID BIGINT PRIMARY KEY, NAME varchar, VALUE bigint) WITH (kafka_topic='left_topic', value_format='JSON');",
-        "CREATE TABLE right_table (ID BIGINT PRIMARY KEY, F1 varchar, F2 bigint) WITH (kafka_topic='right_topic', value_format='JSON');",
-        "CREATE TABLE OUTPUT as SELECT lt.id, rt.id, name, rt.f1 FROM left_table lt LEFT JOIN right_table rt ON lt.value = rt.id;"
+        "CREATE TABLE left_table (id BIGINT PRIMARY KEY, name VARCHAR, value BIGINT) WITH (kafka_topic='left_topic', value_format='JSON');",
+        "CREATE TABLE right_table (id BIGINT PRIMARY KEY, f1 VARCHAR, f2 BIGINT) WITH (kafka_topic='right_topic', value_format='JSON');",
+        "CREATE TABLE output AS SELECT lt.id, rt.id, name, rt.f1 FROM left_table AS lt LEFT JOIN right_table AS rt ON lt.value = rt.id;"
       ],
       "inputs": [
         {"topic": "left_topic", "key": 1, "value": {"NAME": "zero", "VALUE": 0}, "timestamp": 0},
@@ -210,7 +210,7 @@
       ],
       "post": {
         "sources": [
-          {"name": "OUTPUT", "type": "table", "schema": "LT_ID BIGINT KEY, `RT_ID` STRING, `NAME` STRING, `F1` STRING"}
+          {"name": "OUTPUT", "type": "table", "schema": "LT_ID BIGINT KEY, RT_ID STRING, NAME STRING, F1 STRING"}
         ]
       }
     },
@@ -221,9 +221,9 @@
         "ksql.joins.foreign.key.enable": true
       },
       "statements": [
-        "CREATE TABLE left_table (L_ID BIGINT PRIMARY KEY, NAME varchar, VALUE bigint) WITH (kafka_topic='left_topic', value_format='JSON');",
-        "CREATE TABLE right_table (R_ID BIGINT PRIMARY KEY, F1 varchar, F2 bigint) WITH (kafka_topic='right_topic', value_format='JSON');",
-        "CREATE TABLE OUTPUT as SELECT l_id, name, f1 FROM left_table JOIN right_table ON value = r_id;"
+        "CREATE TABLE left_table (l_id BIGINT PRIMARY KEY, name VARCHAR, value BIGINT) WITH (kafka_topic='left_topic', value_format='JSON');",
+        "CREATE TABLE right_table (r_id BIGINT PRIMARY KEY, f1 VARCHAR, f2 BIGINT) WITH (kafka_topic='right_topic', value_format='JSON');",
+        "CREATE TABLE output AS SELECT l_id, name, f1 FROM left_table JOIN right_table ON value = r_id;"
       ],
       "inputs": [
         {"topic": "right_topic", "key": 0, "value": {"F1": "blah", "F2": 4}, "timestamp": 0},
@@ -241,7 +241,7 @@
       ],
       "post": {
         "sources": [
-          {"name": "OUTPUT", "type": "table", "schema": "L_ID BIGINT KEY, `NAME` STRING, `F1` STRING"}
+          {"name": "OUTPUT", "type": "table", "schema": "L_ID BIGINT KEY, NAME STRING, F1 STRING"}
         ]
       }
     },
@@ -251,9 +251,9 @@
         "ksql.joins.foreign.key.enable": true
       },
       "statements": [
-        "CREATE TABLE left_table (L_ID BIGINT PRIMARY KEY, NAME varchar, VALUE bigint) WITH (kafka_topic='left_topic', value_format='JSON');",
-        "CREATE TABLE right_table (R_ID BIGINT PRIMARY KEY, F1 varchar, F2 bigint) WITH (kafka_topic='right_topic', value_format='JSON');",
-        "CREATE TABLE OUTPUT as SELECT name, f1 FROM left_table JOIN right_table ON value = r_id;"
+        "CREATE TABLE left_table (l_id BIGINT PRIMARY KEY, name VARCHAR, value BIGINT) WITH (kafka_topic='left_topic', value_format='JSON');",
+        "CREATE TABLE right_table (r_id BIGINT PRIMARY KEY, f1 VARCHAR, f2 BIGINT) WITH (kafka_topic='right_topic', value_format='JSON');",
+        "CREATE TABLE output AS SELECT name, f1 FROM left_table JOIN right_table ON value = r_id;"
       ],
       "expectedException": {
         "type": "io.confluent.ksql.util.KsqlStatementException",
@@ -266,9 +266,9 @@
         "ksql.joins.foreign.key.enable": true
       },
       "statements": [
-        "CREATE TABLE left_table (L_ID BIGINT PRIMARY KEY, NAME varchar, VALUE bigint) WITH (kafka_topic='left_topic', value_format='JSON');",
-        "CREATE TABLE right_table (R_ID BIGINT PRIMARY KEY, F1 varchar, F2 bigint) WITH (kafka_topic='right_topic', value_format='JSON');",
-        "CREATE TABLE OUTPUT as SELECT l_id, l_id as this_is_a_duplicate, name, f1 FROM left_table JOIN right_table ON value = r_id;"
+        "CREATE TABLE left_table (l_id BIGINT PRIMARY KEY, name VARCHAR, value BIGINT) WITH (kafka_topic='left_topic', value_format='JSON');",
+        "CREATE TABLE right_table (r_id BIGINT PRIMARY KEY, f1 VARCHAR, f2 BIGINT) WITH (kafka_topic='right_topic', value_format='JSON');",
+        "CREATE TABLE output AS SELECT l_id, l_id as this_is_a_duplicate, name, f1 FROM left_table JOIN right_table ON value = r_id;"
       ],
       "expectedException": {
         "type": "io.confluent.ksql.util.KsqlStatementException",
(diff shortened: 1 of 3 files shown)


ksqldb-functional-tests/src/test/resources/query-validation-tests/fk-join.json
```diff
@@ -186,87 +184,502 @@
...
       ],
       "expectedException": {
         "type": "io.confluent.ksql.util.KsqlStatementException",
-        "message": "Invalid join condition: table-table joins require to join on the primary key of the right input table. Got LEFT_TABLE.ID1 = RIGHT_TABLE.F2."
+        "message": "TODO: some sensible error message here"
       }
     },
     {
-      "name": "Should fail on right non-key attribute for inner-join with alias - feature flag enabled",
+      "name": "fails with unexpected right source key in projection",
```
vcrfxia, line 370: This new test explains the changes to remove `r_id` from the tests above.

ksqldb-functional-tests/src/test/resources/query-validation-tests/fk-multi-join.json
```diff
@@ -0,0 +1,294 @@
...
+      },
+      "statements": [
+        "CREATE TABLE left_table (id1 BIGINT PRIMARY KEY, f1 BIGINT) WITH (kafka_topic='left_topic', format='JSON');",
+        "CREATE TABLE middle_table (id2 BIGINT PRIMARY KEY, f2 BIGINT) WITH (kafka_topic='middle_topic', format='JSON');",
+        "CREATE TABLE right_table (id3 BIGINT PRIMARY KEY, f3 BIGINT) WITH (kafka_topic='right_topic', format='JSON');",
+        "CREATE TABLE output AS SELECT f1, f2, f3 FROM left_table JOIN middle_table ON f1 = id2 JOIN right_table ON id1 = id3;"
+      ],
+      "expectedException": {
+        "type": "io.confluent.ksql.util.KsqlStatementException",
+        "message": "TODO: sensible error message here"
```
vcrfxia, line 154: This error message is being added as part of https://github.com/confluentinc/ksql/pull/7522.

ksqldb-functional-tests/src/test/resources/query-validation-tests/fk-multi-join.json
```diff
@@ -0,0 +1,294 @@
...
+      },
+      "statements": [
+        "CREATE TABLE left_table (id1 BIGINT PRIMARY KEY, f1 BIGINT) WITH (kafka_topic='left_topic', format='JSON');",
+        "CREATE TABLE middle_table (id2 BIGINT PRIMARY KEY, f2 BIGINT, other STRING) WITH (kafka_topic='middle_topic', format='JSON');",
+        "CREATE TABLE right_table (id3 BIGINT PRIMARY KEY, f3 BIGINT) WITH (kafka_topic='right_topic', format='JSON');",
+        "CREATE TABLE output AS SELECT id1, id3, f1, f2, f3 FROM left_table JOIN middle_table ON f1 = id2 LEFT JOIN right_table ON id1 = id3;"
+      ],
+      "expectedException": {
+        "type": "io.confluent.ksql.util.KsqlStatementException",
+        "message": "TODO: sensible error message here"
```
vcrfxia, line 260: This test fails until we add the necessary check to ensure that the only key columns selected are the left source key columns.

ksqldb-functional-tests/src/test/resources/query-validation-tests/fk-multi-join.json
```diff
@@ -0,0 +1,294 @@
...
+    {
+      "name": "Should allow fk join at start of n-way join - with aliases",
+      "properties": {
+        "ksql.joins.foreign.key.enable": true
+      },
+      "statements": [
+        "CREATE TABLE left_table (id1 BIGINT PRIMARY KEY, f1 BIGINT) WITH (kafka_topic='left_topic', format='JSON');",
+        "CREATE TABLE middle_table (id2 BIGINT PRIMARY KEY, f2 BIGINT, other STRING) WITH (kafka_topic='middle_topic', format='JSON');",
+        "CREATE TABLE right_table (id3 BIGINT PRIMARY KEY, f3 BIGINT) WITH (kafka_topic='right_topic', format='JSON');",
+        "CREATE TABLE output AS SELECT id1, as_value(id2) as id2, f1, mt.f2, f3 FROM left_table lt JOIN middle_table mt ON f1 = mt.id2 LEFT JOIN right_table rt ON lt.id1 = id3;"
```
vcrfxia, line 272: This test is currently failing with a bizarre error about the qualifier `LT` being invalid, which seems like a bug.

ksqldb-functional-tests/src/test/resources/query-validation-tests/fk-join.json
```diff
@@ -186,87 +184,502 @@
...
-        "TODO: this test should be deleted once the feature is ungated and the feature flag is removed",
-        "as it duplicates another test above (without the feature flag)"
+      "statements": [
+        "CREATE TABLE left_table (L_ID BIGINT PRIMARY KEY, NAME varchar, VALUE bigint) WITH (kafka_topic='left_topic', value_format='JSON');",
+        "CREATE TABLE right_table (R_ID BIGINT PRIMARY KEY, F1 varchar, F2 bigint) WITH (kafka_topic='right_topic', value_format='JSON');",
+        "CREATE TABLE OUTPUT as SELECT l_id, r_id, name, f1 FROM left_table JOIN right_table ON value = r_id;"
       ],
+      "expectedException": {
+        "type": "io.confluent.ksql.util.KsqlStatementException",
+        "message": "TODO: some sensible error message here"
```
vcrfxia, line 381: This test fails until we add the necessary check to ensure that the only key columns selected are the left source key columns.

ksqldb-functional-tests/src/test/resources/query-validation-tests/fk-join.json
```diff
@@ -186,87 +184,502 @@
...
-      "name": "Should fail on right non-key attribute for inner-join with qualifiers - feature flag enabled",
+      "name": "value column - inner join",
       "properties": {
         "ksql.joins.foreign.key.enable": true
       },
-      "comments": [
-        "TODO: this test should be deleted once the feature is ungated and the feature flag is removed",
-        "as it duplicates another test above (without the feature flag)"
+      "statements": [
+        "CREATE TABLE left_table (L_ID BIGINT PRIMARY KEY, NAME varchar, VALUE bigint) WITH (kafka_topic='left_topic', value_format='JSON');",
```
mjsax, line 200: nit: the formatting if rather confusing. Can we use all-caps for keywords and lower cases for everything else? Would be much easier to read...
`CREATE TABLE left_table (l_id BIGINT PRIMARY KEY, name VARCHAR, value BIGINT)`

Also applies to all other newly added tests.

ksqldb-functional-tests/src/test/resources/query-validation-tests/fk-join.json
```diff
@@ -5,7 +5,7 @@
       "statements": [
         "CREATE TABLE left_table (id1 BIGINT PRIMARY KEY, f1 BIGINT) WITH (kafka_topic='left_topic', format='JSON');",
         "CREATE TABLE right_table (id2 BIGINT PRIMARY KEY, f2 BIGINT) WITH (kafka_topic='right_topic', format='JSON');",
-        "CREATE TABLE output AS SELECT id1, id2, f1, f2 FROM left_table JOIN right_table ON id1 = f2;"
+        "CREATE TABLE output AS SELECT id1, f1, f2 FROM left_table JOIN right_table ON id1 = f2;"
```
mjsax, line 8: Should we rewrite those tests as `SELECT *` instead? What is the recommended test pattern?

Btw: #7526 move those cases to a new file.

ksqldb-functional-tests/src/test/resources/query-validation-tests/fk-multi-join.json
```diff
@@ -0,0 +1,294 @@
+{
```
mjsax, line 1: multi-joins is not a good name... Can we call it `fk-n-way-joins` instead?

I hope we can rename existing `mutli-joins.json`, too (after I figured out how to migrate generated history plans)

ksqldb-functional-tests/src/test/resources/query-validation-tests/fk-join.json
```diff
@@ -186,87 +184,502 @@
...
+    },
+    {
+      "name": "value column - inner join - with aliases",
+      "properties": {
+        "ksql.joins.foreign.key.enable": true
+      },
+      "statements": [
+        "CREATE TABLE left_table (ID BIGINT PRIMARY KEY, NAME varchar, VALUE bigint) WITH (kafka_topic='left_topic', value_format='JSON');",
+        "CREATE TABLE right_table (ID BIGINT PRIMARY KEY, F1 varchar, F2 bigint) WITH (kafka_topic='right_topic', value_format='JSON');",
+        "CREATE TABLE OUTPUT as SELECT lt.id, as_value(rt.id) as rt_id, name, rt.f1 FROM left_table lt JOIN right_table rt ON lt.value = rt.id;"
```
mjsax, line 267: `left_table lt JOIN right_table rt`

Can we please always use `AS` -> `left_table AS lt JOIN right_table AS rt`

We want to make `AS` mandatory and thus can avoid to break a lot a test in the future.

ksqldb-functional-tests/src/test/resources/query-validation-tests/fk-join.json
```diff
@@ -53,131 +53,557 @@
...
+    {
+      "enabled": false,
+      "name": "value column - left join",
+      "properties": {
+        "ksql.joins.foreign.key.enable": true
+      },
+      "statements": [
+        "CREATE TABLE left_table (L_ID BIGINT PRIMARY KEY, NAME varchar, VALUE bigint) WITH (kafka_topic='left_topic', value_format='JSON');",
+        "CREATE TABLE right_table (R_ID BIGINT PRIMARY KEY, F1 varchar, F2 bigint) WITH (kafka_topic='right_topic', value_format='JSON');",
+        "CREATE TABLE OUTPUT as SELECT l_id, as_value(r_id) as r_id, name, f1 FROM left_table LEFT JOIN right_table ON value = r_id;"
```
guozhangwang, line 117: This is just a random thought while reading this test, and is not really for the PR itself: I feel that now with PK-join and FK-join together, the current PK "inheritance" semantics is a bit odd. More specifically, with a CTAS statement from a join, seems our scenarios are:

1) `SELECT left.key, right.key` will throw an error from either PK and FK, however conceptually for PK-join it is okay to have both, which would always be the same.

2) `SELECT *` will only set `left.key` as the key for the resulted table, and the `right.key` would be moved to values.

3) If user do want to have both `left.key` and `right.key` they need to call this `as_value()` function to cast the `right.key` to a value. This also means that for FK-join, the foreign-key must be on the left-hand side, and ksql would not be able to auto-switch the left/right if we have `left JOIN right on left.key = right.fk-value`.

I'm feeling that we should try to be more flexible about inheriting the key columns from a join statement, instead of enforcing users to make `as_value()` calls. Say in the above case, would it be better if:

1) `SELECT left.key, right.key` would be allowed, and we would look at the join condition to determine if the `left.key`, or `right.key` (for FK-join) or both together (for PK-join) would be the new table's PK.

2) `SELECT *` would similarly depend on the join conditions, to set either `left.key`, or `right.key` or both together as the new table's PK.

guozhangwang (commented):
I just had a irrelevant though about our PK definition from a join statement; the added test coverage looks great.

ksqldb-functional-tests/src/test/resources/query-validation-tests/fk-join.json
vcrfxia: Hi @guozhangwang , thanks for raising this. I played around with existing table-table joins some more and discovered some discrepancies in my understanding of how ksqlDB handles keys in join projections, your understanding, and what actually happens. Let me start with some background on my understanding, with the caveat that some of my understanding turns out to be incorrect.

ksqlDB currently requires that users select all key columns in the projection of their persistent queries, across the board. If a key column is left out, they will receive an error. If a key column is duplicated (e.g., `select k, k as key_alias, ...`) they will receive an error, since this indicates that they are trying to change the key column(s), which is not allowed. If they want a key column to appear in the result value, they can use `AS_VALUE()` to achieve this. If a user truly wishes to change the key column(s) of the output, they can use `PARTITION BY`. There has been [discussion](https://github.com/confluentinc/ksql/issues/6372) around lifting these stringent requirements for having keys appear exactly as expected in the `SELECT` clause of the query, but it has thus far been inconclusive.

If you extend these principles to joins, then key selection for primary key (non-outer) joins should have the following semantics. Suppose the join is `left JOIN right ON left.k = right.k`:
* If the projection includes `left.k` but not `right.k`, this should succeed.
* If the projection includes `right.k` but not `left.k`, this should succeed.
* If the projection includes both `left.k` and `right.k`, then this should fail because it is attempting to include two key columns in the projection.
* If the projection includes multiple copies of either `left.k` or `right.k`, then this should fail because it is attempting to include multiple key columns in the projection.
* If the projection includes `left.k` and `right.*`, then this should succeed with no copies of `right.k` in the resulting value columns, since the user did not explicitly request this with `AS_VALUE()`.
* If the projection includes `right.k` and `left.*`, then this should succeed with no copies of `left.k` in the resulting value columns, since the user did not explicitly request this with `AS_VALUE()`.
* If the right includes both ` no explicit key columns

vcrfxia: We can see from the unexpected behavior (unexpected in light ksqlDB's approach of requiring users to be strict about selected keys, selecting no more and no less, and using `AS_VALUE()` in order to explicitly move them to the value if desired) that there are many inconsistencies around how this is handled today. IMO discussing what the behavior should be (e.g., whether we want to lift the stringent requirements around key selection, not just for joins but for all queries) is a separate question from the immediate decision at hand which is what we want the behavior for FK joins to be, as we are introducing FK joins for the first time. I think that the existing inconsistencies that have been pointed out here should be considered bugs and that we should introduce FK joins with "proper" semantics and restrictions, in light of ksqlDB's overall approach today, even if that means we're inconsistent between the behavior of FK and PK joins. Then we can revamp all these semantics in the future as part of a larger project.

cc @mjsax @agavra @colinhicks who might have other opinions.

vcrfxia: On second thought, I have a hunch for where this odd behavior is coming from in the ksqlDB code base and suspect it's likely easier to propagate the existing behavior (where right key columns are magically copied into the join result value) for FK joins than to try to tighten up the semantics. If this is the case, I think we should go with this option since (1) it's actually much less weird for an FK join to copy right key columns into the result value, since otherwise the right key columns would not be represented in the result, and (2) for the sake of shipping FK joins sooner rather than later, since users who don't want right key columns in the result value can choose to leave them out via explicit selection.

I'll dig into the implementation a bit more and update this test suite to reflect the outcome.

agavra: I was going to say that this odd behavior is not, all things considered, that odd. We can say that in a join the left key is always taken as the key, unless you select only the right key. I believe this was by design from Andy when he originally introduced this feature. We might be able to dig this up in the KLIP somewhere.

For FK joins, I'd recommend just sticking to the existing behavior as I don't think it's that important TBH to nail this. It might be a quirk, but as long as it can be clearly explained by one principle ("in any situation where you select both keys, the right gets copied into the value") it's not that urgent to fix.

guozhangwang: Thanks for the very nice summary @vcrfxia . So if we follow the direction to stick with the current semantics, I'd just want to clarify the following for added FK-join:

0. we would not allow `JOIN right ON left.k == right.v` but only `JOIN right ON left.v == right.k`. I.e. users need to make sure the FK is always on left-hand side.
1. If the projection includes `left.k` but not `right.k`, this should succeed with `left.k` as PK.
2. If the projection includes `right.k` but not `left.k`, this should fail.
3. If the projection includes both `left.k` and `right.k`, then this should fail; users need to move `right.k` to values via "as_value".
4. If the projection includes multiple copies of either left.k or right.k, then this should fail.
5. If the projection includes `left.k` and `right.*`, then this should succeed with `left.k` as PK and `right.k` in values automatically.
6. If the projection includes `right.k` and `left.*`, then this should fail; users need to move `right.k` to values via "as_value".
7. If the projection includes both `left.*` and `right.*` and no explicit key columns, then this should succeed with `left.k` as PK and `right.k` in values automatically.
8. If the projection contains a single (unqualified) * and no explicit key columns, then this should succeed with `left.k` as PK and `right.k` in values automatically.

Does that sounds right?

ksqldb-functional-tests/src/test/resources/query-validation-tests/fk-join.json
vcrfxia: Went through and updating the casing of all statements to be consistent.

ksqldb-functional-tests/src/test/resources/query-validation-tests/fk-join.json
vcrfxia: Added explicit tests for `SELECT *`, including with different combinations of qualifiers and aliases. I think it's fine to have the default tests use explicit selects rather than `SELECT *` as long as both are covered.

ksqldb-functional-tests/src/test/resources/query-validation-tests/fk-join.json
vcrfxia: I updated the tests to have `AS` here as requested, though FWIW I don't think it hurts to use the other form in tests since we currently support both?

ksqldb-functional-tests/src/test/resources/query-validation-tests/fk-join.json
vcrfxia: Confirmed the suspicion above that consistent behavior with PK joins is what the current (ongoing) FK join implementation uses, so we'll go with this for the upcoming release. Updated this test suite to reflect the changes. This PR is now ready for final review and to be merged cc @mjsax .

guozhangwang: Thanks @vcrfxia , could you verify that the point.0 above is true as well ?

vcrfxia: @guozhangwang : 
0. This is correct. ksqlDB today is not set up to reverse the order of joins (for example, this is why only LEFT JOIN and not RIGHT JOIN is supported).
1. Correct.
2. Correct.
3. Incorrect, assuming that by "current semantics" you mean what ksqlDB currently does for PK joins which is to allow this and put `right.k` into the value by default. We've decided to mirror the existing behavior for PK joins for FK joins as well, so I will answer the rest of these questions assuming we're talking about this (though I still contend that the existing behavior for PK joins is buggy / inconsistent with how ksqlDB treats keys in projections for the non-join case 🙂)
4. The query will fail if there are multiple copies of `left.k`, but the query will not fail if there are multiple copies of `right.k` (assuming the different copies have been properly aliased to have different names, as ksqlDB does not allow multiple columns with the same name).
5. Correct.
6. Incorrect. The will succeed with `right.k` moved to the value automatically.
7. Correct. 
8. Correct.

guozhangwang: For 4/6, I'm stating that just based on the current added tests, where if user do not manually specify `as_value` it would fail. I'm fine with changing it to be aligned with the PK-join semantics, just clarifying that in this case we would need to change the tests here.

ksqldb-functional-tests/src/test/resources/query-validation-tests/fk-join.json
```diff
@@ -53,131 +53,651 @@
...
+        "CREATE TABLE output AS SELECT l_id, r_id, name, f1 FROM left_table JOIN right_table ON value = r_id;"
+      ],
+      "inputs": [
+        {"topic": "right_topic", "key": 0, "value": {"F1": "blah", "F2": 4}, "timestamp": 0},
+        {"topic": "left_topic", "key": 1, "value": {"NAME": "zero", "VALUE": 0}, "timestamp": 10000},
+        {"topic": "left_topic", "key": 0, "value": {"NAME": "foo", "VALUE": 100}, "timestamp": 13000},
+        {"topic": "right_topic", "key": 0, "value": {"F1": "a", "F2": 10}, "timestamp": 15000},
+        {"topic": "left_topic", "key": 10, "value": {"NAME": "bar", "VALUE": 0}, "timestamp": 16000},
+        {"topic": "right_topic", "key": 0, "value": null, "timestamp": 17000},
+        {"topic": "left_topic", "key": 1, "value": null, "timestamp": 18000}
```
mjsax, line 93: Just tried this test on my feature branch, and this record does actually produce one more tombstone (for this case, we don't know that the result was already deleted, because the delete was trigger by the right-hand side).

ksqldb-functional-tests/src/test/resources/query-validation-tests/fk-join.json
vcrfxia: Which tests are you referring to? I updated the PR before my previous replies, wondering if it's possible you were looking at an older version.

ksqldb-functional-tests/src/test/resources/query-validation-tests/fk-join.json
```diff
@@ -53,131 +53,667 @@
...
+    {
+      "enabled": false,
+      "name": "qualified select * - left",
+      "properties": {
+        "ksql.joins.foreign.key.enable": true
+      },
+      "statements": [
+        "CREATE TABLE left_table (l_id BIGINT PRIMARY KEY, name VARCHAR, value BIGINT) WITH (kafka_topic='left_topic', value_format='JSON');",
+        "CREATE TABLE right_table (r_id BIGINT PRIMARY KEY, f1 VARCHAR, f2 BIGINT) WITH (kafka_topic='right_topic', value_format='JSON');",
+        "CREATE TABLE output AS SELECT left_table.*, right_table.f1 FROM left_table JOIN right_table ON value = r_id;"
```
mjsax, line 321: why do we include `right_table.f1`

ksqldb-functional-tests/src/test/resources/query-validation-tests/fk-join.json
```diff
@@ -53,131 +53,667 @@
...
+          {"name": "OUTPUT", "type": "table", "keyFormat": {"format": "KAFKA"}}
+        ],
+        "topics": {
+          "blacklist": ".*-repartition"
+        }
+      }
+    },
+    {
+      "enabled": false,
+      "name": "Should allow (default) key feature mismatch",
```
mjsax, line 489: Not sure what this means?

mjsax (approved):
I think we can merge this. We can address the question in follow up PRs.

ksqldb-functional-tests/src/test/resources/query-validation-tests/fk-join.json
vcrfxia: I answered in https://github.com/confluentinc/ksql/pull/7528#discussion_r634037193.

ksqldb-functional-tests/src/test/resources/query-validation-tests/fk-join.json
vcrfxia: No particular reason; just to check that we can include additional columns besides the qualified select *.

ksqldb-functional-tests/src/test/resources/query-validation-tests/fk-join.json
```diff
@@ -53,131 +53,667 @@
...
       "properties": {
         "ksql.joins.foreign.key.enable": true
       },
       "statements": [
-        "CREATE TABLE left_table (id1 BIGINT PRIMARY KEY, f1 BIGINT) WITH (kafka_topic='left_topic', format='JSON');",
-        "CREATE TABLE right_table (id2 BIGINT PRIMARY KEY, f2 BIGINT) WITH (kafka_topic='right_topic', format='JSON');",
-        "CREATE TABLE output AS SELECT id1, id2, f1, f2 FROM left_table FULL OUTER JOIN right_table ON f1 = id2;"
+        "CREATE TABLE left_table (l_id BIGINT PRIMARY KEY, name VARCHAR, value BIGINT) WITH (kafka_topic='left_topic', value_format='JSON');",
+        "CREATE TABLE right_table (r_id BIGINT PRIMARY KEY, f1 VARCHAR, f2 BIGINT) WITH (kafka_topic='right_topic', value_format='JSON');",
+        "CREATE TABLE output AS SELECT l_id, r_id, name, f1 FROM left_table JOIN right_table ON value = r_id;"
```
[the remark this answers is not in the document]
vcrfxia: I don't think it's a big deal. There are tests in this file both with and without `r_id`, since the query should work with both.

ksqldb-functional-tests/src/test/resources/query-validation-tests/fk-join.json
guozhangwang: Obviously I did not get the latest commit :) Sorry

PR #7542, opened 2021-05-17
chore: KREST-1030 suppress deprecated usage of SslContextFactory

dimitarndimitrov:
### Description 
~Replaced a usage of the default `SslContextFactory` ctor in order to avoid a warning that breaks the build once Jetty is bumped so that the ctor is deprecated.~
I have replaced the fix with a warning suppression, so that this change can be merged proactively. The previous fix is better in theory, but it can be merged only after the `rest-utils` changes have been merged (so it also breaks CI right now).
- This is not a problem on `5.3.x` and above, so this change needs to land solely here. We should merge it up with `pint merge -s ours ksql`. **I actually don't have the permissions to do that, so I'll have to ask the reviewers to help me with this**.
- I didn't cherry-pick the changes fixing this in `5.3.x` as they are slightly messy (there are 2 commits for this, and there's a redundant suppression of deprecation there - see https://github.com/confluentinc/ksql/blob/b860520484e4677602d94a39e78e044db2a89184/ksql-cli/src/test/java/io/confluent/ksql/cli/SslFunctionalTest.java).

### Testing done 
With the parent pom.xml edited to have explicit Jetty and Jersey dependencies with the versions bumped in https://github.com/confluentinc/rest-utils/pull/252, ensured that a local build passes successfully (it failed without this change with the error seen in https://jenkins.confluent.io/job/system-test-confluent-platform-branch-builder/5502/console).
### Reviewer checklist

### DIFF
--- a/ksql-cli/src/test/java/io/confluent/ksql/cli/SslFunctionalTest.java
+++ b/ksql-cli/src/test/java/io/confluent/ksql/cli/SslFunctionalTest.java
@@ -47,7 +47,6 @@ import org.eclipse.jetty.client.HttpClient;
 import org.eclipse.jetty.http.HttpStatus;
 import org.eclipse.jetty.http.HttpStatus.Code;
 import org.eclipse.jetty.util.ssl.SslContextFactory;
-import org.eclipse.jetty.util.ssl.SslContextFactory.Server;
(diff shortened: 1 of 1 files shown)


vcrfxia (approved):
LGTM

Issue #3060, opened 2019-07-10
KAFKA VALUE_FORMAT to support JOINs and GROUP BY


big-andy-coates:
The initial implementation of the KAFKA format does not support joins and aggregations because the format of the source topics is used to create repartition and changelog topics. Such topics have the ROWTIME and ROWKEY columns copied into the value and the KAFKA format only supports single fields.

This was deemed acceptable as the KAFKA format is primarily intended as a key format, not a value format. However, we should look to address this limitation.

Possible solutions are:
1. Use a single internal format for internal topics, rather than the source format. 
2. Do not copy the ROWTIME and ROWKEY into the value of internal topics.  This would also save space.

PR #7543, opened 2021-05-17
feat: add FK-join logical to physical plan translation

mjsax:
### Description 
Depend on #7522. Translated the logical FK-join into the physical plan.
-> still missing: add type check for join condition

### Testing done 
tbd

### Reviewer checklist

### DIFF
--- a/ksqldb-engine/src/main/java/io/confluent/ksql/analyzer/Analysis.java
+++ b/ksqldb-engine/src/main/java/io/confluent/ksql/analyzer/Analysis.java
@@ -437,6 +437,14 @@ public class Analysis implements ImmutableAnalysis {
       return rightJoinExpression;
     }
 
+    public Expression getFlippedLeftJoinExpression() {
+      return flippedJoinCondition ? rightJoinExpression : leftJoinExpression;
+    }
+
+    public Expression getFlippedRightJoinExpression() {
+      return flippedJoinCondition ? leftJoinExpression : rightJoinExpression;
+    }
+
     public JoinType getType() {
       return type;
     }
--- a/ksqldb-engine/src/main/java/io/confluent/ksql/planner/LogicalPlanner.java
+++ b/ksqldb-engine/src/main/java/io/confluent/ksql/planner/LogicalPlanner.java
@@ -596,14 +596,9 @@ public class LogicalPlanner {
     final Optional<ColumnReferenceExp> fkColumnName =
         verifyJoin(root.getInfo(), preRepartitionLeft, preRepartitionRight);
 
-    final boolean finalJoin = prefix.isEmpty();
-
-    final JoinKey joinKey;
-    if (fkColumnName.isPresent()) {
-      joinKey = buildForeignJoinKey(root, fkColumnName.get());
-    } else {
-      joinKey = buildJoinKey(root);
-    }
+    final JoinKey joinKey = fkColumnName
+        .map(columnReferenceExp -> buildForeignJoinKey(root, fkColumnName.get()))
+        .orElseGet(() -> buildJoinKey(root));
 
     final PlanNode left = prepareSourceForJoin(
         root.getLeft(),
@@ -624,7 +619,7 @@ public class LogicalPlanner {
         new PlanNodeId(prefix + "Join"),
         root.getInfo().getType(),
         joinKey.rewriteWith(refRewriter::process),
-        finalJoin,
+        prefix.isEmpty(),
         left,
         right,
         root.getInfo().getWithinExpression(),
@@ -635,41 +630,18 @@ public class LogicalPlanner {
   /**
    * @return the foreign key column if this is a foreign key join
    */
-  // CHECKSTYLE_RULES.OFF: CyclomaticComplexity|NPathComplexity
   private Optional<ColumnReferenceExp> verifyJoin(
       final JoinInfo joinInfo,
       final PlanNode leftNode,
       final PlanNode rightNode
   ) {
-    // CHECKSTYLE_RULES.ON: CyclomaticComplexity|NPathComplexity
     final JoinType joinType = joinInfo.getType();
     final Expression leftExpression = joinInfo.getLeftJoinExpression();
     final Expression rightExpression = joinInfo.getRightJoinExpression();
 
     if (leftNode.getNodeOutputType() == DataSourceType.KSTREAM) {
       if (rightNode.getNodeOutputType() == DataSourceType.KTABLE) {
-
-        // stream-table join detected
-
-        if (joinType.equals(JoinType.OUTER)) {
-          throw new KsqlException(String.format(
-                  "Invalid join type: "
-                      + "full-outer join not supported for stream-table join. Got %s %s %s.",
-                  joinInfo.getLeftSource().getDataSource().getName().text(),
-                  joinType,
-                  joinInfo.getRightSource().getDataSource().getName().text()
-              ));
-        }
-
-        if (joinOnNonKeyAttribute(rightExpression, rightNode)) {
-          throw new KsqlException(String.format(
-                  "Invalid join condition:"
-                      + " stream-table joins require to join on the table's primary key."
-                      + " Got %s = %s.",
-                  joinInfo.hasFlippedJoinCondition() ? rightExpression : leftExpression,
-                  joinInfo.hasFlippedJoinCondition() ? leftExpression : rightExpression
-              ));
-        }
+        verifyStreamTableJoin(joinInfo, rightNode);
       }
 
       // stream-stream join detected: nothing to verify
@@ -692,64 +664,135 @@ public class LogicalPlanner {
                 "Invalid join condition:"
                     + " table-table joins require to join on the primary key of the right input"
                     + " table. Got %s = %s.",
-                joinInfo.hasFlippedJoinCondition() ? rightExpression : leftExpression,
-                joinInfo.hasFlippedJoinCondition() ? leftExpression : rightExpression
+                joinInfo.getFlippedLeftJoinExpression(),
+                joinInfo.getFlippedRightJoinExpression()
            ));
       }
 
       if (joinOnNonKeyAttribute(leftExpression, leftNode)) {
         // foreign key join detected
 
-        if (ksqlConfig.getBoolean(KsqlConfig.KSQL_FOREIGN_KEY_JOINS_ENABLED)) {
+        return verifyForeignKeyJoin(joinInfo, leftNode, rightNode);
+      } else {
+        // primary key join detected
 
-          if (joinInfo.getType().equals(JoinType.OUTER)) {
-            throw new KsqlException(String.format(
-                "Invalid join type:"
-                    + " full-outer join not supported for foreign-key table-table join."
-                    + " Got %s %s %s.",
-                joinInfo.getLeftSource().getDataSource().getName().text(),
-                joinType,
-                joinInfo.getRightSource().getDataSource().getName().text()
-            ));
-          }
+        final SqlType leftKeyType = Iterables.getOnlyElement(leftNode.getSchema().key()).type();
+        final SqlType rightKeyType = Iterables.getOnlyElement(rightNode.getSchema().key()).type();
 
-          // after we lift this n-way join restriction, we should be able to support FK-joins
-          // at any level in the join tree, even after we add right-deep/bushy join tree support,
-          // because a FK-join output table has the same PK as its left input table
-          if (!(leftNode instanceof DataSourceNode)
-              || !(rightNode instanceof DataSourceNode)) {
-            throw new KsqlException(String.format(
-                "Invalid join condition: foreign-key table-table joins are not "
-                    + "supported as part of n-way joins. Got %s = %s.",
-                joinInfo.hasFlippedJoinCondition() ? rightExpression : leftExpression,
-                joinInfo.hasFlippedJoinCondition() ? leftExpression : rightExpression
-            ));
-          }
-
-          if (!(leftExpression instanceof ColumnReferenceExp)) {
-            throw new KsqlException(String.format(
-                "Invalid join condition: foreign-key table-table joins with expressions "
-                    + "are not supported. Got %s = %s."
-            ));
-          }
-
-          return Optional.of(((ColumnReferenceExp) leftExpression));
-        } else {
-          throw new KsqlException(String.format(
-              "Invalid join condition:"
-                  + " foreign-key table-table joins are not supported. Got %s = %s.",
-              joinInfo.hasFlippedJoinCondition() ? rightExpression : leftExpression,
-              joinInfo.hasFlippedJoinCondition() ? leftExpression : rightExpression
-          ));
-        }
+        verifyJoinConditionTypes(
+            leftKeyType,
+            rightKeyType,
+            leftExpression,
+            rightExpression,
+            joinInfo.hasFlippedJoinCondition()
+        );
       }
     }
 
     return Optional.empty();
   }
 
-  private boolean joinOnNonKeyAttribute(final Expression joinExpression,
-                                        final PlanNode node) {
+  private static void verifyStreamTableJoin(final JoinInfo joinInfo, final PlanNode rightNode) {
+    final JoinType joinType = joinInfo.getType();
+    final Expression rightExpression = joinInfo.getRightJoinExpression();
+
+    if (joinType.equals(JoinType.OUTER)) {
+      throw new KsqlException(String.format(
+          "Invalid join type: "
+              + "full-outer join not supported for stream-table join. Got %s %s %s.",
+          joinInfo.getLeftSource().getDataSource().getName().text(),
+          joinType,
+          joinInfo.getRightSource().getDataSource().getName().text()
+      ));
+    }
+
+    if (joinOnNonKeyAttribute(rightExpression, rightNode)) {
+      throw new KsqlException(String.format(
+          "Invalid join condition:"
+              + " stream-table joins require to join on the table's primary key."
+              + " Got %s = %s.",
+          joinInfo.getFlippedLeftJoinExpression(),
+          joinInfo.getFlippedRightJoinExpression()
+      ));
+    }
+  }
+
+  private Optional<ColumnReferenceExp> verifyForeignKeyJoin(
+      final JoinInfo joinInfo,
+      final PlanNode leftNode,
+      final PlanNode rightNode
+  ) {
+    final JoinType joinType = joinInfo.getType();
+    final Expression leftExpression = joinInfo.getLeftJoinExpression();
+    final Expression rightExpression = joinInfo.getRightJoinExpression();
+
+    if (ksqlConfig.getBoolean(KsqlConfig.KSQL_FOREIGN_KEY_JOINS_ENABLED)) {
+
+      if (joinInfo.getType().equals(JoinType.OUTER)) {
+        throw new KsqlException(String.format(
+            "Invalid join type:"
+                + " full-outer join not supported for foreign-key table-table join."
+                + " Got %s %s %s.",
+            joinInfo.getLeftSource().getDataSource().getName().text(),
+            joinType,
+            joinInfo.getRightSource().getDataSource().getName().text()
+        ));
+      }
+
+      // after we lift this n-way join restriction, we should be able to support FK-joins
+      // at any level in the join tree, even after we add right-deep/bushy join tree support,
+      // because a FK-join output table has the same PK as its left input table
+      if (!(leftNode instanceof DataSourceNode)
+          || !(rightNode instanceof DataSourceNode)) {
+        throw new KsqlException(String.format(
+            "Invalid join condition:"
+                + " foreign-key table-table joins are not supported as part of n-way joins."
+                + " Got %s = %s.",
+            joinInfo.getFlippedLeftJoinExpression(),
+            joinInfo.getFlippedRightJoinExpression()
+        ));
+      }
+
+      if (!(leftExpression instanceof ColumnReferenceExp)) {
+        throw new KsqlException(String.format(
+            "Invalid join condition:"
+                + " foreign-key table-table joins with expressions are not supported yet."
+                + " Got %s = %s.",
+            joinInfo.getFlippedLeftJoinExpression(),
+            joinInfo.getFlippedRightJoinExpression()
+        ));
+      }
+
+      // we need to extend this to support expressions later on
+      final ColumnReferenceExp fkColumnReference = (ColumnReferenceExp) leftExpression;
+
+      final SqlType fkColumnType =
+          leftNode.getSchema().findColumn(fkColumnReference.getColumnName()).get().type();
+      final SqlType rightKeyType = Iterables.getOnlyElement(rightNode.getSchema().key()).type();
+
+      verifyJoinConditionTypes(
+          fkColumnType,
+          rightKeyType,
+          leftExpression,
+          rightExpression,
+          joinInfo.hasFlippedJoinCondition()
+      );
+
+      return Optional.of(fkColumnReference);
+    } else {
+      throw new KsqlException(String.format(
+          "Invalid join condition:"
+              + " foreign-key table-table joins are not supported. Got %s = %s.",
+          joinInfo.getFlippedLeftJoinExpression(),
+          joinInfo.getFlippedRightJoinExpression()
+      ));
+    }
+  }
+
+  private static boolean joinOnNonKeyAttribute(
+      final Expression joinExpression,
+      final PlanNode node
+  ) {
     if (!(joinExpression instanceof ColumnReferenceExp)) {
       return true;
     }
@@ -806,7 +849,7 @@ public class LogicalPlanner {
     return !joinAttributeName.equals(Iterables.getOnlyElement(keyColumns).name());
   }
 
-  private boolean isInnerNode(final PlanNode node) {
+  private static boolean isInnerNode(final PlanNode node) {
     if (node instanceof JoinNode) {
       return true;
     }
@@ -832,10 +875,29 @@ public class LogicalPlanner {
 
     final ColumnName foreignKeyColumnName = columnRef.maybeQualifier().isPresent()
         ? ColumnNames.generatedJoinColumnAlias(
-            columnRef.maybeQualifier().get(), columnRef.getColumnName())
+            columnRef.maybeQualifier().get(),
+            columnRef.getColumnName())
         : columnRef.getColumnName();
 
     return JoinKey.foreignKeyColumn(foreignKeyColumnName, leftSourceKeys);
+  }
+
+  private static void verifyJoinConditionTypes(
+      final SqlType leftType,
+      final SqlType rightType,
+      final Expression leftExpression,
+      final Expression rightExpression,
+      final boolean flipped
+  ) {
+    if (!leftType.equals(rightType)) {
+      throw new KsqlException(String.format(
+          "Invalid join condition: types don't match. Got %s{%s} = %s{%s}.",
+          flipped ? rightExpression : leftExpression,
+          flipped ? rightType : leftType,
+          flipped ? leftExpression : rightExpression,
+          flipped ? leftType : rightType
+      ));
+    }
   }
 
   private JoinKey buildJoinKey(final Join join) {
--- a/ksqldb-engine/src/main/java/io/confluent/ksql/planner/plan/JoinNode.java
+++ b/ksqldb-engine/src/main/java/io/confluent/ksql/planner/plan/JoinNode.java
@@ -214,15 +214,59 @@ public class JoinNode extends PlanNode implements JoiningNode {
   @Override
   void validateKeyPresent(final SourceName sinkName, final Projection projection) {
 
-    final boolean atLeastOneKey = joinKey.getAllViableKeys(schema).stream()
-        .anyMatch(projection::containsExpression);
+    if (joinKey.isForeignKey()) {
+      final DataSource leftInputTable = getLeftmostSourceNode().getDataSource();
+      final SourceName leftInputTableName = leftInputTable.getName();
+      final List<Column> leftInputTableKeys = leftInputTable.getSchema().key();
+      final List<Column> missingKeys =
+          leftInputTableKeys.stream().filter(
+              k -> !projection.containsExpression(
+                    new QualifiedColumnReferenceExp(leftInputTableName, k.name()))
+                  && !projection.containsExpression(new UnqualifiedColumnReferenceExp(
+                      ColumnNames.generatedJoinColumnAlias(leftInputTableName, k.name())
+                  ))
+          ).collect(Collectors.toList());
 
-    if (!atLeastOneKey) {
-      final boolean synthetic = joinKey.isSynthetic();
-      final List<? extends Expression> viable = joinKey.getOriginalViableKeys(schema);
+      if (!missingKeys.isEmpty()) {
+        throwMissingKeyColumnException(missingKeys, leftInputTableName);
+      }
+    } else {
+      final boolean atLeastOneKey = joinKey.getAllViableKeys(schema).stream()
+          .anyMatch(projection::containsExpression);
 
-      throwKeysNotIncludedError(sinkName, "join expression", viable, false, synthetic);
+      if (!atLeastOneKey) {
+        final boolean synthetic = joinKey.isSynthetic();
+        final List<? extends Expression> viable = joinKey.getOriginalViableKeys(schema);
+
+        throwKeysNotIncludedError(sinkName, "join expression", viable, false, synthetic);
+      }
     }
+  }
+
+  private static void throwMissingKeyColumnException(
+      final List<Column> missingKeys,
+      final SourceName leftInputTableName
+  ) {
+    final String columnString = missingKeys.size() == 1 ? "column" : "columns";
+    final String plainInputTableName = leftInputTableName.text();
+    // column names may be represented as `TABLENAME_COLUMNNAME`
+    final int tablePrefixLength = plainInputTableName.length() + 1;
+    throw new KsqlException(String.format(
+        "Primary key %s missing in projection."
+            + " For foreign-key table-table joins, the projection must include"
+            + " all primary key columns from the left input table (%s)."
+            + " Missing %s: %s.",
+        columnString,
+        leftInputTableName,
+        columnString,
+        missingKeys.stream()
+            .map(c -> c.name().text())
+            .map(name -> name.startsWith(plainInputTableName)
+                ? name.substring(tablePrefixLength)
+                : name)
+            .map(name -> String.format("`%s`", name))
+            .collect(Collectors.joining(", "))
+    ));
   }
 
   @Override
@@ -395,7 +439,7 @@ public class JoinNode extends PlanNode implements JoiningNode {
               contextStacker
           );
         case INNER:
-          return leftStream.join(
+          return leftStream.innerJoin(
               rightStream,
               joinNode.getKeyColumnName(),
               joinNode.withinExpression.get().joinWindow(),
@@ -440,7 +484,7 @@ public class JoinNode extends PlanNode implements JoiningNode {
           );
 
         case INNER:
-          return leftStream.join(
+          return leftStream.innerJoin(
               rightTable,
               joinNode.getKeyColumnName(),
               JoiningNode.getValueFormatForSource(joinNode.left).getFormatInfo(),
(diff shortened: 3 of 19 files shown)


ksqldb-engine/src/main/java/io/confluent/ksql/planner/plan/JoinNode.java
```diff
@@ -474,16 +474,37 @@ private StreamToTableJoiner(
...
+                null, // fk-join column
+                contextStacker
+            );
+          }
         case INNER:
-          return leftTable.join(
-              rightTable,
-              joinNode.getKeyColumnName(),
-              contextStacker);
+          if (true) { // check if PK or FK
```
spena, line 491: So you would call `joinKey.isForeignKey()`, and then get the fk column with `((ForeignJoinKey) joinKey).getForeignKeyColumn()`

mjsax: Correct (assuming we don't change #7522 that adds those methods)

ksqldb-engine/src/main/java/io/confluent/ksql/planner/LogicalPlanner.java
```diff
@@ -719,14 +730,22 @@ private boolean verifyJoin(
               rightExpression
           ));
         }
+      } else {
+        // primary key join detected
+        verifyIfKeyTypesMatch(
```
mjsax, line 735: Side cleanup: pull this check out of the physical planner into the logical planner.

Need to do something similar later for other joins (ie, stream-stream and stream-table).

ksqldb-engine/src/main/java/io/confluent/ksql/planner/LogicalPlanner.java
```diff
@@ -719,14 +730,22 @@ private boolean verifyJoin(
...
+        );
       }
     }
 
     return false;
   }
 
-  private boolean joinOnNonKeyAttribute(final Expression joinExpression,
-                                        final PlanNode node) {
+  private static boolean joinOnNonKeyAttribute(final Expression joinExpression,
```
mjsax, line 747: side cleanup

ksqldb-engine/src/main/java/io/confluent/ksql/planner/LogicalPlanner.java
```diff
@@ -631,11 +633,9 @@ private JoinNode buildJoin(final Join root, final String prefix, final boolean i
    * @return whether this is a foreign key join or not
    */
   // CHECKSTYLE_RULES.OFF: CyclomaticComplexity
-  private boolean verifyJoin(
-      final JoinInfo joinInfo,
-      final PlanNode leftNode,
-      final PlanNode rightNode
-  ) {
+  private boolean verifyJoin(final JoinInfo joinInfo,
```
vcrfxia, line 636: nit: the existing spacing pattern is consistent with what's used throughout most of the ksqlDB code base. This change is moving away from the norm.

ksqldb-engine/src/main/java/io/confluent/ksql/planner/LogicalPlanner.java
```diff
@@ -719,14 +738,27 @@ private boolean verifyJoin(
               rightExpression
           ));
         }
+      } else {
+        // primary key join detected
+
+        final SqlType leftKeyType = Iterables.getOnlyElement(leftNode.getSchema().key()).type();
+        final SqlType rightKeyType = Iterables.getOnlyElement(rightNode.getSchema().key()).type();
+
+        verifyJoinConditionTypes(
```
vcrfxia, line 747: nit: if we're adding this check for regular table-table joins in this PR as well, can we also add the check for stream-stream and stream-table joins here as well for consistency? Or at least file an issue so we don't forget about it and leave the inconsistency.

vcrfxia (commented):
LGTM assuming the PR is updated for the dependent PR and we add a couple QTTs that pass before merging. Thanks @mjsax !

ksqldb-engine/src/main/java/io/confluent/ksql/planner/LogicalPlanner.java
mjsax: Let's file a ticket. For stream-stream and stream-table is more complex as we support expression...

mjsax: Actually, we should do this when adding expression support for FK-joins. Added a comment to the exiting ticket.

ksqldb-engine/src/main/java/io/confluent/ksql/planner/LogicalPlanner.java
```diff
@@ -596,14 +596,9 @@ private JoinNode buildJoin(final Join root, final String prefix, final boolean i
...
-    final boolean finalJoin = prefix.isEmpty();
-
-    final JoinKey joinKey;
-    if (fkColumnName.isPresent()) {
-      joinKey = buildForeignJoinKey(root, fkColumnName.get());
-    } else {
-      joinKey = buildJoinKey(root);
-    }
+    final JoinKey joinKey = fkColumnName
+        .map(columnReferenceExp -> buildForeignJoinKey(root, fkColumnName.get()))
```
mjsax, line 600: Was not sure how to simplify (or even remove `buildForeignJoinKey()`...) cf https://github.com/confluentinc/ksql/pull/7522/files#r634680772

ksqldb-execution/src/main/java/io/confluent/ksql/execution/plan/TableSelect.java
```diff
@@ -34,14 +34,14 @@
   private final ExecutionStep<KTableHolder<K>> source;
   private final ImmutableList<ColumnName> keyColumnNames;
   private final ImmutableList<SelectExpression> selectExpressions;
-  private final Optional<Formats> internalFormats;
+  private final Formats internalFormats;
```
mjsax, line 37: Side cleanup: we always have a format set; no need to make it `Optional`

ksqldb-functional-tests/src/main/java/io/confluent/ksql/test/tools/TestExecutor.java
```diff
@@ -198,9 +198,9 @@ public void buildAndExecuteQuery(
           );
         }
 
-        topologyTestDriverContainer.getTopologyTestDriver()
-            .producedTopicNames()
-            .forEach(topicInfoCache::get);
+//        topologyTestDriverContainer.getTopologyTestDriver()
+//            .producedTopicNames()
+//            .forEach(topicInfoCache::get);
```
mjsax, line 203: FK-join creates a subscription and response topic, that the framework cannot handle yet... removed this now to be able to run tests

ksqldb-streams/src/main/java/io/confluent/ksql/execution/streams/KsqlKeyExtractor.java
```diff
@@ -37,7 +38,7 @@
   @SuppressWarnings("unchecked")
   @Override
   public KRightT apply(final GenericRow left) {
-    return (KRightT) left.get(leftJoinColumnIndex);
+    return (KRightT) GenericKey.genericKey(left.get(leftJoinColumnIndex));
```
mjsax, line 41: Not 100% sure we can always return a `GenericKey`?

ksqldb-streams/src/main/java/io/confluent/ksql/execution/streams/SourceBuilder.java
```diff
@@ -351,12 +351,15 @@ private static PhysicalSchema getPhysicalSchema(final SourceStep<?> streamSource
         // re-partitioned topic will be used for any subsequent state stores, in lieu
         // of the original source topic, thus avoiding the issues above.
         // See https://github.com/confluentinc/ksql/issues/6650
-        table = source.mapValues(row -> row);
+        table = source.mapValues(row -> row, Materialized.with(null, valueSerde));
```
mjsax, line 354: Ensure that we don't drop the value serde (similar below)

ksqldb-functional-tests/src/test/resources/query-validation-tests/fk-join.json
```diff
@@ -1,5 +1,23 @@
 {
   "tests": [
+    {
+      "name": "Inner join",
```
mjsax, line 4: This one passed now 🎉

ksqldb-execution/src/main/java/io/confluent/ksql/execution/plan/TableSelect.java
vcrfxia: This change needs to be backed out. This is not a backwards compatible evolution. There could be old versions written into the command topic.

ksqldb-functional-tests/src/main/java/io/confluent/ksql/test/tools/TestExecutor.java
vcrfxia: Are we planning to add support for this in this PR or a follow-up PR? If a follow-up PR, we should gate this with the feature flag instead (if possible) as this is an important piece of testing.

ksqldb-functional-tests/src/test/resources/query-validation-tests/fk-join.json
vcrfxia: Hooray!! Great stuff. What do you think about getting https://github.com/confluentinc/ksql/pull/7537 merged first and simply enabling select tests from that PR, rather than introducing new tests in this PR? That way we can avoid merge conflicts / duplicate tests.

The PR build passed on https://github.com/confluentinc/ksql/pull/7537 already. I still have more cleanup to do on the PR but we can merge first to unblock you and I can open a follow-up for the cleanup.

ksqldb-functional-tests/src/test/resources/query-validation-tests/table-table-join.json
```diff
@@ -1,5 +1,17 @@
 {
   "tests": [
+    {
+      "name": "foo",
```
vcrfxia, line 4: Is this meant to be checked in?

ksqldb-streams/src/main/java/io/confluent/ksql/execution/streams/KsqlKeyExtractor.java
vcrfxia: GenericKey might fail for windowed keys. This is a good reminder that we should add tests for that.

> It seems format KAFKA is not supported for joins

You mean as a value format? It's definitely supported as a key format.

ksqldb-streams/src/main/java/io/confluent/ksql/execution/streams/SourceBuilder.java
vcrfxia: I'm confused -- is this creating a new state store? We need to ensure that any changes here are backwards compatible with existing plans.

ksqldb-streams/src/main/java/io/confluent/ksql/execution/streams/TableSelectBuilder.java
```diff
@@ -73,27 +73,25 @@ private TableSelectBuilder() {
 
     final boolean forceMaterialize = !matBuilder.isPresent();
 
-    if (formats.isPresent() && forceMaterialize) {
```
vcrfxia, line 76: Why is this code being removed?

ksqldb-streams/src/main/java/io/confluent/ksql/execution/streams/TableSelectBuilder.java
```diff
@@ -121,7 +119,8 @@ private TableSelectBuilder() {
 
     final KTable<K, GenericRow> transFormedTable = table.getTable().transformValues(
         () -> new KsTransformer<>(selectMapper.getTransformer(logger)),
-            selectName
+        Materialized.with(keySerde, valSerde),
```
vcrfxia, line 122: Same question as above regarding whether this is adding a new state store, and the implications for backwards incompatibility. 

Also, what's the purpose of moving the building of `physicalSchema`, `keySerde` and `valueSerde` out of the if-statement if they're only used inside the if-statement anyway?

ksqldb-functional-tests/src/main/java/io/confluent/ksql/test/tools/TestExecutor.java
mjsax: Not sure how complex it will be? Personally happy to do it both ways...
 - either fix in this PR
 - or keep QTT test disabled to merge this PR and do a fix in a follow up PR
 
 What do you think is better?

ksqldb-streams/src/main/java/io/confluent/ksql/execution/streams/KsqlKeyExtractor.java
mjsax: Well, I tried to change the right-hand side format to `KAFKA` (to get PK format) -- but there was an exception that it's not supported. Even tries to set the format for the result table, but did not help. Maybe something we need to dig into...

ksqldb-streams/src/main/java/io/confluent/ksql/execution/streams/SourceBuilder.java
mjsax: Hmm... Good point. Need to think about it.

ksqldb-functional-tests/src/test/resources/query-validation-tests/table-table-join.json
mjsax: No :)

ksqldb-streams/src/main/java/io/confluent/ksql/execution/streams/KsqlKeyExtractor.java
vcrfxia: The `KAFKA` value format is not supported for joins: https://github.com/confluentinc/ksql/issues/3060. However, it is supported as a key format.

ksqldb-functional-tests/src/main/java/io/confluent/ksql/test/tools/TestExecutor.java
vcrfxia: I'm fine either way. If this PR is close to being merged, let's merge first to make it easier to parallelize subsequent work. We should address the open questions around the potential addition of extra state stores, and back out the changes to TableSelect but I think all the other comments/discussions can be addressed in follow-ups.

ksqldb-engine/src/main/java/io/confluent/ksql/analyzer/Analysis.java
```diff
@@ -437,6 +437,14 @@ public Expression getRightJoinExpression() {
       return rightJoinExpression;
     }
 
+    public Expression getFlippedLeftJoinExpression() {
+      return flippedJoinCondition ? rightJoinExpression : leftJoinExpression;
```
guozhangwang, line 441: I did not review the previous PR adding this, but regarding the flipped flag, if we stick with the current semantics it seems we would not allow such flipping in FK-join since we need to stick to which PK to inherit from the join (see https://github.com/confluentinc/ksql/pull/7537#discussion_r635482988 point 0), please correct me if my understanding is wrong?

ksqldb-engine/src/main/java/io/confluent/ksql/planner/LogicalPlanner.java
```diff
@@ -692,64 +664,135 @@ private JoinNode buildJoin(final Join root, final String prefix, final boolean i
...
+    return Optional.empty();
+  }
 
-          if (!(leftExpression instanceof ColumnReferenceExp)) {
-            throw new KsqlException(String.format(
-                "Invalid join condition: foreign-key table-table joins with expressions "
-                    + "are not supported. Got %s = %s."
-            ));
-          }
+  private static void verifyStreamTableJoin(final JoinInfo joinInfo, final PlanNode rightNode) {
```
guozhangwang, line 695: I'm assuming this diff does not include any logical change for stream-table join, just refactoring here. LMK if it contains any logical changes.

Also, I'm wondering why we can declare this function a static but cannot do the same for `verifyForeignKeyJoin`?

ksqldb-engine/src/main/java/io/confluent/ksql/planner/plan/JoinNode.java
```diff
@@ -722,12 +749,12 @@ public boolean isForeignKey() {
 
     @Override
     public List<? extends Expression> getOriginalViableKeys(final LogicalSchema schema) {
-      return leftSourceKeyColumns;
+      throw new UnsupportedOperationException("Should not be called with foreign key joins.");
```
guozhangwang, line 752: Good catch.

ksqldb-engine/src/main/java/io/confluent/ksql/planner/LogicalPlanner.java
```diff
@@ -832,12 +875,47 @@ private JoinKey buildForeignJoinKey(final Join join, final ColumnReferenceExp co
...
+        maybeFlippedLeftExpression = rightExpression;
+        maybeFlippedRightExpression = leftExpression;
+      } else {
+        maybeFlippedLeftType = leftType;
+        maybeFlippedRightType = rightType;
+        maybeFlippedLeftExpression = leftExpression;
+        maybeFlippedRightExpression = rightExpression;
+      }
+
+      throw new KsqlException(String.format(
```
guozhangwang, line 909: nit: just use a ternary operator inside `String.format()` here?

ksqldb-engine/src/main/java/io/confluent/ksql/structured/SchemaKTable.java
```diff
@@ -318,6 +319,52 @@ public SchemaKGroupedTable groupBy(
     );
   }
 
+  public <KRightT> SchemaKTable<K> foreignKeyJoin(
```
guozhangwang, line 322: nit meta: for function names, I'd suggest we use `XXInnerJoinYY` instead of `XXJoinYY` for all such cases since `JOIN` is just a syntax sugar for `INNER JOIN` here, and the function name would better be with the full name to clearly indicate their scope.

ksqldb-streams/src/main/java/io/confluent/ksql/execution/streams/SourceBuilder.java
guozhangwang: Before we fix https://issues.apache.org/jira/browse/KAFKA-12242, maybe we should just pass `valueSerde` along to downstream and then set when needed later.

ksqldb-engine/src/main/java/io/confluent/ksql/analyzer/Analysis.java
vcrfxia: @guozhangwang there is a distinction between flipping the left and right sources for the join (i.e., the difference between LEFT JOIN and RIGHT JOIN) versus flipping the order of the join expressions (i.e., the difference between `JOIN ON a=b` and `JOIN ON b=a`, without flipping the order of the join sources). ksqlDB does not allow the former but _does_ allow the latter.

guozhangwang: Got it, I think I was originally misunderstanding this, because the flipping's purpose is to make sure the new source of a multi-way join is always a source. Now I understand the flipping is only on the `JOIN`.

ksqldb-streams/src/main/java/io/confluent/ksql/execution/streams/SourceBuilder.java
mjsax: It should actually be fine -- as long as we don't set a store name, we won't force materialization. Will revert this anyway, and just reconstruct the Serde in the FK-join node instead.

ksqldb-streams/src/main/java/io/confluent/ksql/execution/streams/TableSelectBuilder.java
mjsax: Ok. Actually it does not work to move it into the join-builder and we want to set the serdes here, as a `TableSelect` will produce the input-tables for the join (at least in the examples I did -- I assume we can relay on this -- would be good to get confirmation).

ksqldb-engine/src/main/java/io/confluent/ksql/planner/LogicalPlanner.java
mjsax: Correct. I extracted those only to make checkstyle happy so we can remove the supressions ` CHECKSTYLE_RULES.OFF: CyclomaticComplexity|NPathComplexity`

`verifyForeignKeyJoin` accesses member variable `ksqlConfig` atm, so it cannot be `static.` If we remove the feature flag at some point in the future, we could make it `static`, too.

ksqldb-engine/src/main/java/io/confluent/ksql/planner/LogicalPlanner.java
mjsax: Not sure what you mean?

ksqldb-engine/src/main/java/io/confluent/ksql/planner/plan/JoinNode.java
mjsax: Was actually pointed out by @vcrfxia on the original PR that added it (we just merge the other PR as-is, and I address open comments on this PR to close those gaps).

ksqldb-engine/src/main/java/io/confluent/ksql/planner/plan/JoinNode.java
```diff
@@ -214,14 +214,51 @@ protected int getPartitions(final KafkaTopicClient kafkaTopicClient) {
   @Override
   void validateKeyPresent(final SourceName sinkName, final Projection projection) {
 
-    final boolean atLeastOneKey = joinKey.getAllViableKeys(schema).stream()
-        .anyMatch(projection::containsExpression);
+    if (joinKey.isForeignKey()) {
+      final DataSource leftInputTable = getLeftmostSourceNode().getDataSource();
```
mjsax, line 218: This check is new.

ksqldb-engine/src/main/java/io/confluent/ksql/planner/plan/KsqlStructuredDataOutputNode.java
```diff
@@ -35,7 +35,7 @@
 
   private final KsqlTopic ksqlTopic;
   private final boolean doCreateInto;
-  private final SourceName intoSourceName;
+  private final SourceName sinkName;
```
mjsax, line 38: Just some renaming... `intoSourcName` is a really terrible name...

ksqldb-engine/src/main/java/io/confluent/ksql/planner/plan/KsqlStructuredDataOutputNode.java
```diff
@@ -67,17 +67,13 @@ public KsqlTopic getKsqlTopic() {
     return ksqlTopic;
   }
 
-  public SourceName getIntoSourceName() {
```
mjsax, line 70: There is `getSinkName` to not sure why we would need this method?

ksqldb-functional-tests/src/test/resources/query-validation-tests/fk-join.json
```diff
@@ -1,5 +1,35 @@
 {
   "tests": [
+    {
+      "name": "Should fail if single result key column is missing for inner-join",
```
mjsax, line 4: new test for the new check

ksqldb-functional-tests/src/test/resources/query-validation-tests/fk-join.json
```diff
@@ -179,6 +209,69 @@
...
+        "CREATE TABLE right_table (id2 INT PRIMARY KEY, f2 BIGINT) WITH (kafka_topic='right_topic', format='JSON');",
+        "CREATE TABLE output AS SELECT * FROM left_table JOIN right_table ON f1 = id2;"
+      ],
+      "expectedException": {
+        "type": "io.confluent.ksql.util.KsqlStatementException",
+        "message": "Invalid join condition: types don't match. Got LEFT_TABLE.F1{BIGINT} = RIGHT_TABLE.ID2{INTEGER}."
+      }
+    },
+    {
+      "name": "Inner join",
```
mjsax, line 244: This test passed locally.

ksqldb-streams/src/main/java/io/confluent/ksql/execution/streams/TableSelectBuilder.java
mjsax: > out of the if-statement if they're only used inside the if-statement anyway?

Maybe it was not clear from the previous diff, but we did (and still do) use them outside :)

vcrfxia: Still looking into this in more detail but I don't understand why we can't pass in serdes in the join builder. I also don't understand why passing in serdes in TableSelectBuilder is sufficient for ensuring serdes are always available at the FK join step. I think you might see a serde failure if you uncomment one of the tests in my PR for adding an FK join at the start of an n-way join. Can we sync offline to make this discussion go faster?

ksqldb-functional-tests/src/test/resources/query-validation-tests/fk-join.json
vcrfxia: As discussed offline, it'd be great to get https://github.com/confluentinc/ksql/pull/7537 merged first so this PR can simply uncomment the tests that are now expected to pass, rather than adding any new tests and requiring us to resolve merge conflicts / the possibility of duplicate tests.

ksqldb-functional-tests/src/main/java/io/confluent/ksql/test/tools/TestExecutor.java
```diff
@@ -198,9 +198,9 @@ public void buildAndExecuteQuery(
           );
         }
 
-        topologyTestDriverContainer.getTopologyTestDriver()
-            .producedTopicNames()
-            .forEach(topicInfoCache::get);
+        //        topologyTestDriverContainer.getTopologyTestDriver()
```
vcrfxia, line 201: As discussed offline, we should not check in this PR will this commented out. We can place in a hack to bypass this for FK joins only and clean it up later in order to merge this PR sooner, but we cannot disable the check for existing tests as this is an important piece of coverage.

ksqldb-engine/src/test/java/io/confluent/ksql/planner/plan/AggregateNodeTest.java
```diff
@@ -306,7 +306,8 @@ public void shouldCreateLoggers() {
     assertThat(loggers, contains(
         "queryid.KsqlTopic.Source",
         "queryid.Aggregate.GroupBy",
-        "queryid.Aggregate.Aggregate.Materialize"
+        "queryid.Aggregate.Aggregate.Materialize",
+        "queryid.Aggregate.Project"
```
vcrfxia, line 310: I don't totally understand why this changed yet (I assume it has to do with the changes in TableSelectBuilder?) but at the surface this looks concerning. I wouldn't have expected this PR to affect aggregates at all.

ksqldb-engine/src/main/java/io/confluent/ksql/planner/plan/JoinNode.java
```diff
@@ -214,14 +214,51 @@ protected int getPartitions(final KafkaTopicClient kafkaTopicClient) {
...
+      final SourceName leftInputTableName = leftInputTable.getName();
+      final String plainInputTableName = leftInputTableName.text();
+      final List<Column> leftInputTableKeys = leftInputTable.getSchema().key();
+      final List<Column> missingKeys =
+          leftInputTableKeys.stream().filter(
+              k -> !projection.containsExpression(
+                    new QualifiedColumnReferenceExp(leftInputTableName, k.name()))
+                  && !projection.containsExpression(new UnqualifiedColumnReferenceExp(
+                      ColumnName.of(plainInputTableName + "_" + k.name().text())
```
vcrfxia, line 227: This is duplicating the logic in `ColumnNames.generatedJoinColumnAlias()`, right? It'd be good to call the helper method to keep the logic in one place instead.

ksqldb-engine/src/main/java/io/confluent/ksql/planner/plan/JoinNode.java
```diff
@@ -214,14 +214,51 @@ protected int getPartitions(final KafkaTopicClient kafkaTopicClient) {
...
+    if (joinKey.isForeignKey()) {
+      final DataSource leftInputTable = getLeftmostSourceNode().getDataSource();
+      final SourceName leftInputTableName = leftInputTable.getName();
+      final String plainInputTableName = leftInputTableName.text();
+      final List<Column> leftInputTableKeys = leftInputTable.getSchema().key();
+      final List<Column> missingKeys =
+          leftInputTableKeys.stream().filter(
+              k -> !projection.containsExpression(
+                    new QualifiedColumnReferenceExp(leftInputTableName, k.name()))
+                  && !projection.containsExpression(new UnqualifiedColumnReferenceExp(
```
vcrfxia, line 226: Why do we need to check for both the original column name and the rewritten column name here, in contrast to the validation for the non-FK-join case below that is able to directly call `projection.containsExpression()`? (I assume there's a good reason -- I just don't understand what it is yet.)

ksqldb-engine/src/main/java/io/confluent/ksql/planner/plan/JoinNode.java
```diff
@@ -214,14 +214,51 @@ protected int getPartitions(final KafkaTopicClient kafkaTopicClient) {
...
+                      ColumnName.of(plainInputTableName + "_" + k.name().text())
+                  ))
+          ).collect(Collectors.toList());
+
+      if (!missingKeys.isEmpty()) {
+        final String columnString = missingKeys.size() == 1 ? "column" : "columns";
+        // column names may be represented as `TABLENAME_COLUMNNAME`
+        final int tablePrefixLength = plainInputTableName.length() + 1;
+        throw new KsqlException(String.format(
+            "Primary key %s missing in projection."
```
vcrfxia, line 236: nit: can we use the same error message format as in `throwKeysNotIncludedError()` for consistency? Ideally we'd just call `throwKeysNotIncludedError()` (and update the method as needed to accommodate this case), or at minimum pull this code out into a new helper method (`throwKeysNotIncludedErrorForFKJoin()` or similar) to make this more readable.

pom.xml
```diff
@@ -155,7 +155,7 @@
             <dependency>
                 <groupId>org.apache.kafka</groupId>
                 <artifactId>kafka-streams</artifactId>
-                <version>${kafka.version}</version>
+                <version>3.0.0-SNAPSHOT</version>
```
vcrfxia, line 158: This does not look like it's meant to be checked in.

vcrfxia (commented):
Thanks @mjsax -- initial review comments inline. There are still a number of things I'm looking into to understand these changes more deeply. LMK if you want to sync offline to make this go faster.

ksqldb-engine/src/main/java/io/confluent/ksql/planner/plan/JoinNode.java
mjsax: IIRC, the issue is that we cannot rely on `containsExpression()` as it just returns true/false, while we actually want to compile a list if missing key column. Happy to improve this code as a follow up if possible.

As a matter of fact, I don't like this code myself, but tried to get something to just work.

ksqldb-engine/src/main/java/io/confluent/ksql/planner/plan/JoinNode.java
mjsax: It seems that generating an alias would generate new name (like prefix the name with `KSQL_COL`, but we what reconstruct on existing name, using the table name as prefix?

ksqldb-engine/src/main/java/io/confluent/ksql/planner/plan/JoinNode.java
mjsax: I can extract it in a helper, however, the other error message is quite different, given we only handle single-column-keys, and that we might generate a synthetic key column. Might be better to just keep separated for now.

ksqldb-engine/src/test/java/io/confluent/ksql/planner/plan/AggregateNodeTest.java
mjsax: In this test, we just parse an aggregation query that inserts a `TableSelect` upstream that we need to mock.

ksqldb-streams/src/main/java/io/confluent/ksql/execution/streams/TableSelectBuilder.java
mjsax: It was rather a limitation of QQT that tracks registered serdes -- it assigns a unique ID for each registered serde and assumes that each operator registers at most one (output) serde. When we try to register left-value-input and left-value-ouptut serde, it raises an error that we try to register two output serdes.

pom.xml
mjsax: Ups...

ksqldb-engine/src/main/java/io/confluent/ksql/planner/LogicalPlanner.java
guozhangwang: Oh I just meant maybe we do not need to declare these local fields, and just inside `String.format` we do:

```
...
flipped ? rightExpression: leftExpression,
flipped ? rightType: leftType,
flipped ? leftExpression: rightExpression,
flipped ? leftType: rightType,
```

ksqldb-engine/src/main/java/io/confluent/ksql/planner/plan/JoinNode.java
vcrfxia: No, if you look at the implementation for the method, you'll see that all it does is concatenate the source name with the column name, with an underscore in between. This is the same logic used throughout the rest of the code to rewrite join column names, which the logic here should be consistent with.

ksqldb-engine/src/main/java/io/confluent/ksql/planner/plan/JoinNode.java
vcrfxia: This piece of code is inside a call to `Stream::filter()`. Isn't a helper method that returns a boolean exactly what we want?

Agreed that we can improve this in a follow-up. I just didn't understand what was different about checking for columns within the projection for PK and FK joins, and wanted to understand to have greater confidence that we haven't introduced any bugs around how FK joins are renaming columns in different situations.

mjsax: > Isn't a helper method that returns a boolean exactly what we want?

Well, yes and no. For base functionally "yes". For a good error message "no" -> if we only get a `true/false` answer, we don't know _which_ key columns are missing.

vcrfxia: Aren't we already collecting a list of `missingKeys`? My question is about updating lines 222-229 to become something like
```
final List<Column> missingKeys =
          leftInputTableKeys.stream().filter(
                   k -> !projection.containsExpression( ... )
          ).collect(Collectors.toList());
```
and then we can use the collected list of missingKeys in the error message, the same way the PR currently is. My question is just about simplifying the boolean filter to be a single call to `projection.containsExpression()` rather than two. Is this not possible?

mjsax: Ah. I remember now... `leftInputTableKeys` is `List<Column>` while we would need `List<Expression>` to apply `projection.containsExpression()`

ksqldb-functional-tests/src/test/resources/query-validation-tests/fk-join.json
```diff
@@ -3,92 +3,197 @@
     {
       "name": "Should fail on left non-key attribute for inner-join",
       "statements": [
-        "CREATE TABLE left_table (id1 BIGINT PRIMARY KEY, f1 BIGINT) WITH (kafka_topic='left_topic', format='JSON');",
```
mjsax, line 6: I did some more side cleanup to have a unified test and naming pattern.

Moved all negative tests to the top, and grouped them on "generally not supported", invalid join condition, invalid projection.

ksqldb-engine/src/main/java/io/confluent/ksql/planner/plan/JoinNode.java
```diff
@@ -214,17 +214,61 @@ protected int getPartitions(final KafkaTopicClient kafkaTopicClient) {
...
+        final boolean synthetic = joinKey.isSynthetic();
+        final List<? extends Expression> viable = joinKey.getOriginalViableKeys(schema);
 
-      throwKeysNotIncludedError(sinkName, "join expression", viable, false, synthetic);
+        throwKeysNotIncludedError(sinkName, "join expression", viable, false, synthetic);
+      }
     }
   }
 
+  private static void throwMissingKeyColumnException(
```
vcrfxia, line 246: nit: can we rename this to include FK joins in the name? Otherwise it's not clear what the difference between this method and `throwKeysNotIncludedError()` is since they do the same thing.

ksqldb-functional-tests/src/test/resources/query-validation-tests/fk-join.json
vcrfxia: Can we keep large refactors like this in separate PRs next time please? These test changes are really hard to review now -- I can't tell which changes are part of the refactor versus actual changes as part of this PR.

I think they're all just refactors? Hard to tell since tests were moved and renamed in the same change.

vcrfxia (approved):
As mentioned inline, didn't review the test changes very carefully. LGTM for merging sooner to unblock follow-up PRs.

ksqldb-functional-tests/src/test/resources/query-validation-tests/fk-join.json
mjsax: This PR does not do any changes to test. (otherwise I would have pointed it out).

ksqldb-engine/src/main/java/io/confluent/ksql/planner/plan/JoinNode.java
mjsax: Will rename if a follow up PR (assuming the build passes). If the build fails, will do it right away with any required fixed to make Jenkins pass.

ksqldb-streams/src/main/java/io/confluent/ksql/execution/streams/SourceBuilder.java
guozhangwang: Sounds good.

guozhangwang:
I made a second pass on it and do not have further comments. LGTM.

mjsax:
Closing this PT in favor of #7570

PR #7544, opened 2021-05-18
fix: reject mismatched decimals from avro topics

### LINKED ISSUE #7170: avro serdes don't correctly handle precision/scale mismatch between ksql and registered schema
Our avro deserializer doesn't try to handle the case where the incoming precision/scale don't match the schema precision/scale - it just takes the deserialized BigDecimal and hands it off to the engine (the json deserializer does do this). Then, when we try to serialize the BigDecimal to write it (e.g. to a change-log), the converter throws because the scale is off.

The correct behavior would be to either:
- treat the schema change as a query-killing error and fail the query from the deserializer so the error is more clear
- don't treat the schema change as a query-killing error and try to coerce the values into the new schema. rows with values that can't be coerced can be discarded and we'd log a processing log msg

Not sure which would be better. We should also look at enabling users to evolve their ksql schemas to the correct version (in the event this happens due to a schema change) - I don't think we support that now in a way that would also update the running query

jzaralim:
### Description 
Fixes #7170 

If you run the following statements
```
CREATE STREAM A (dec DECIMAL(4,2)) WITH (topic='topic', value_format='AVRO');
CREATE STREAM B (dec DECIMAL(6,3)) WITH (topic='topic', value_format='AVRO');
CREATE STREAM C AS SELECT * FROM A;
INSERT INTO B VALUES (123.456);
```

You would end up with a `org.apache.kafka.common.errors.SerializationException` in the logs, instead of rejecting the value to begin with.  

And querying A would return the mismatched decimal
```
SELECT * FROM A EMIT CHANGES;
+---------+
|DEC      |
+---------+
|123.456  |

```

This PR changes the behaviour so that if there is a mismatching decimal, it first tries to convert the topic decimal to the ksql schema, and if that is impossible, then it skips that value and the logs a `org.apache.kafka.common.errors.SerializationException` in the streams log.

### Testing done 
Manually tested, added unit test

### Reviewer checklist

### DIFF
--- a/ksqldb-serde/src/main/java/io/confluent/ksql/serde/avro/AvroDataTranslator.java
+++ b/ksqldb-serde/src/main/java/io/confluent/ksql/serde/avro/AvroDataTranslator.java
@@ -120,8 +120,7 @@ public class AvroDataTranslator implements DataTranslator {
       case STRUCT:
         return convertStruct((Struct) object, schema);
       case BYTES:
-        return DecimalUtil.ensureFit(
-            (BigDecimal) object, DecimalUtil.precision(schema), DecimalUtil.scale(schema));
+        return DecimalUtil.ensureFit((BigDecimal) object, schema);
       default:
         return object;
     }
--- a/ksqldb-serde/src/main/java/io/confluent/ksql/serde/avro/AvroDataTranslator.java
+++ b/ksqldb-serde/src/main/java/io/confluent/ksql/serde/avro/AvroDataTranslator.java
@@ -17,6 +17,8 @@ package io.confluent.ksql.serde.avro;
 
 import io.confluent.ksql.serde.connect.ConnectDataTranslator;
 import io.confluent.ksql.serde.connect.DataTranslator;
+import io.confluent.ksql.util.DecimalUtil;
+import java.math.BigDecimal;
 import java.util.ArrayList;
 import java.util.HashMap;
 import java.util.Iterator;
@@ -117,7 +119,9 @@ public class AvroDataTranslator implements DataTranslator {
 
       case STRUCT:
         return convertStruct((Struct) object, schema);
-
+      case BYTES:
+        return DecimalUtil.ensureFit(
+            (BigDecimal) object, DecimalUtil.precision(schema), DecimalUtil.scale(schema));
       default:
         return object;
     }
--- a/ksqldb-serde/src/test/java/io/confluent/ksql/serde/avro/AvroDataTranslatorTest.java
+++ b/ksqldb-serde/src/test/java/io/confluent/ksql/serde/avro/AvroDataTranslatorTest.java
@@ -26,6 +26,7 @@ import static org.junit.Assert.assertThrows;
 import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableMap;
 import io.confluent.ksql.util.DecimalUtil;
+import io.confluent.ksql.util.KsqlException;
 import java.math.BigDecimal;
 import org.apache.kafka.connect.data.Schema;
 import org.apache.kafka.connect.data.SchemaBuilder;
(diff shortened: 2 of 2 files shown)


ksqldb-serde/src/main/java/io/confluent/ksql/serde/avro/AvroDataTranslator.java
```diff
@@ -117,7 +119,9 @@ private static Object replaceSchema(final Schema schema, final Object object) {
 
       case STRUCT:
         return convertStruct((Struct) object, schema);
-
+      case BYTES:
+        return DecimalUtil.ensureFit(
```
abbccdda, line 123: Why don't we just use `public static BigDecimal ensureFit(final BigDecimal value, final Schema schema)`?

ksqldb-serde/src/test/java/io/confluent/ksql/serde/avro/AvroDataTranslatorTest.java
```diff
@@ -288,4 +292,37 @@ public void shouldDropOptionalFromRootStructSchema() {
...
+
+    // When:
+    final AvroDataTranslator translator =
+        new AvroDataTranslator(ksqlSchema, AvroProperties.DEFAULT_AVRO_SCHEMA_FULL_NAME);
+
+    // Then:
+    final ArithmeticException e = assertThrows(
+        ArithmeticException.class,
+        () -> translator.toKsqlRow(topicSchema, new BigDecimal("123.456")));
+    assertThat(e.getMessage(), containsString("A field with precision 4 and scale 2 must round"));
```
abbccdda, line 310: Could we include more text here for verification to show the max value comparison, such like:
```
Numeric field overflow: A field with precision 4 and scale 2 must round to an absolute value less than 10^4
```

ksqldb-serde/src/test/java/io/confluent/ksql/serde/avro/AvroDataTranslatorTest.java
```diff
@@ -288,4 +292,37 @@ public void shouldDropOptionalFromRootStructSchema() {
...
+  }
+
+  @Test
+  public void shouldForceUnmatchingDecimalSchemaIfPossible() {
```
abbccdda, line 314: Maybe a dumb question, could we support a conversion from 1234.5 to a 5 precision and 0 scale decimal number? Are there any other invalid cases we could test out?

ksqldb-serde/src/main/java/io/confluent/ksql/serde/avro/AvroDataTranslator.java
```diff
@@ -117,7 +119,9 @@ private static Object replaceSchema(final Schema schema, final Object object) {
 
       case STRUCT:
         return convertStruct((Struct) object, schema);
-
+      case BYTES:
```
abbccdda, line 122: How did we reach the assumption that given type of `BYTES` would require a decimal conversion?

jzaralim: Decimals in Avro are represented as bytes. [This function](https://github.com/confluentinc/ksql/blob/a321a753a08354d10d4166366c3fe9b13517a258/ksqldb-common/src/main/java/io/confluent/ksql/util/DecimalUtil.java#L43) is called earlier on to reject any `BYTES` that don't represent decimals. Right now, we don't have a `BYTES` type in ksqlDB so all incoming bytes must be decimals, but this logic will change once we implement it.

ksqldb-serde/src/test/java/io/confluent/ksql/serde/avro/AvroDataTranslatorTest.java
jzaralim: No, we don't support rounding for conversions now. Added a test for that case.

abbccdda (approved):
LGTM, thanks for the PR! @jzaralim

PR #7547, opened 2021-05-18
chore: fix alias resolution in n-way join logic

mjsax:
### Description 
FK-join must resolve table alias correctly.

### Testing done 

### Reviewer checklist

### DIFF
--- a/ksqldb-engine/src/main/java/io/confluent/ksql/planner/LogicalPlanner.java
+++ b/ksqldb-engine/src/main/java/io/confluent/ksql/planner/LogicalPlanner.java
@@ -659,7 +659,7 @@ public class LogicalPlanner {
 
       // table-table join detected
 
-      if (joinOnNonKeyAttribute(rightExpression, rightNode)) {
+      if (joinOnNonKeyAttribute(rightExpression, rightNode, joinInfo.getRightSource())) {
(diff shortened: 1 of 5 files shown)


vcrfxia (commented):
Thanks @mjsax -- if I'm reading the fix in this PR correctly, this issue impacts not only FK joins but also table-table-table joins where one of the first two tables is aliased. Is that correct? We should add a QTT for this if so. (This would also mean it's a regression for 0.19 if this PR isn't merged before then. Let's make extra sure to not lose track of this.)

Also I renamed the PR title to be a bit more descriptive, and changed the PR type from `fix` to `chore` so that it doesn't show up in the changelog since the fix is only for a unreleased change.

guozhangwang (approved):
The fix lgtm modulo @vcrfxia 's comments.

vcrfxia (approved):
Thanks @mjsax -- LGTM, though the PR description could be more informative :)

guozhangwang:
> Thanks @mjsax -- LGTM, though the PR description could be more informative :)

+1 :) I was originally confused by this and https://github.com/confluentinc/ksql/pull/7592 since their titles are the same, now I understand the other is only for table-table-table while this is for fk-table-table.

PR #7550, opened 2021-05-18
docs: klip-47 - NON NULL columns and component types

stevenpyzhang:
### Description 

### Testing done 

### Reviewer checklist

### DIFF
--- a/design-proposals/klip-47-non-null-columns-and-component-types.md
+++ b/design-proposals/klip-47-non-null-columns-and-component-types.md
@@ -36,24 +36,31 @@ Producers outside ksqlDB are able to write into the topic. This means that the u
 have null column records in it. There's no way around this as ksqlDB can't prevent outside producers from writing into the Kafka topics.
 
 With that in mind, the `NOT NULL` constraint can be enforced within the ksqlDB service. When reading records from a source that has a `NOT NULL` 
-column constraint, invalid records will be dropped, and an error sent to the processing log. When attempting to write an invalid
-record to a sink, the record would also be dropped, and an error sent to the processing log. A `INSERT INTO VALUES` statement with
+column constraint, invalid records will be dropped, and an error sent to the processing log.
+```
+"recordProcessingError":{"errorMessage":"Found row with NULL value in NOT NULL column (<column_name>) when reading from <stream/table name>. The record will be dropped."
+```
+
+When attempting to write an invalid
+record to a sink, the record would also be dropped, and an error sent to the processing log.
+```
+"recordProcessingError":{"errorMessage":"Error writing row with NULL value in NOT NULL column (<column_name>) for <stream/table name>. The record will be dropped."
+```
+
+A `INSERT INTO VALUES` statement with
 a null column record would error out. There would be clear documentation for users that `NOT NULL` does not prevent invalid records in the 
 Kafka topic due to records from producers outside ksqlDB.
-
-If there are concerns that applying a constraint over existing data (declaring a stream/table over an existing topic) would lead to records getting dropped, 
-we could require a `DEFAULT` value be specified with `NOT NULL`. This is explained later in the KLIP and needs to be discussed further
- as it's not within scope of the original issue filed.
 
 ## What is in scope
 * Allowing users to specify `NOT NULL` for columns
 * Prevent reading records with a null column value from `NOT NULL` streams/tables
 * Prevent writing records with a null column value to `NOT NULL` streams/tables
 * Allow users to specify `NOT NULL` for array type, map key/value type, and struct field types
-* (maybe) Allow users to specify a default value when applying `NOT NULL`
+
+## What is not in scope
+* Allowing users to add `NOT NULL` to an existing column since we currently don't support modifying columns of existing stream/table, only adding columns.
 
 ## Design
-This design section is only focused on `NOT NULL` for now until we decide whether `DEFAULT` is required or not.
 
 We can add a new `optional` variable to the `SqlType` class to indicate whether the type is optional or not. A `SqlType` variable is in each `Column` object so 
 this would be a convenient way to store the `NOT NULL` constraint for each column. When doing code generation in `SqlToJavaVisitor`, we can then add null filtering code
@@ -105,6 +112,19 @@ ksql> select * from pageviews_no_null emit changes;
 |3              |user                                                            |page                                                            |
 ```
 
+If there was an aggregate column defined as `NOT NULL` using aggregate function `UDAF_TEST`
+and the output of `UDAF_TEST` was NULL for ID `2`, but not for ID `3` and `1`, the result in the table would be:
+```sql
+CREATE TABLE not_null_aggregate as SELECT ID, UDAF_TEST(value) as AGGR FROM test group by id;
+
+select * from not_null_aggregate emit changes;
++---------------+-------------------------------------------+
+|ID              |AGGR                                      |
++---------------+-------------------------------------------+
+|3               | 54                                       |
+|1               | 2                                        |
+```
+
 Attempting to insert invalid records would error out:
 ```sql
 ksql> INSERT INTO pageviews VALUES (5, NULL, 'page23');
@@ -128,32 +148,6 @@ Records that have column values such as
 
 Would be dropped when reading from the `valid_data` stream and trying to insert record with these column values would error out.
 
-## NOT NULL DEFAULT
-When `ALTER` is used to add a `NOT NULL` constraint in SQL to an existing table column, there can't be any `NULL` values 
-in that column, otherwise there would be rows that violate the `NOT NULL` constraint.
-
-In this case, the user needs to either `UPDATE` all existing `NULL` 
-values in that column or specify a `DEFAULT` value for that column. All existing rows with `NULL` would have the default value added to that column, 
-so that the NOT NULL constraint is not violated (This is only done if the `NOT NULL` constraint is used in conjunction with `DEFAULT`, if a 
-column with no constraints has a `DEFAULT` applied to it, the existing rows with null columns won't be updated).
-
-When `DEFAULT` is specified, instead of dropping the record, the default value would be applied to that record's column.
-```sql
-CREATE stream pageviews_default(viewtime BIGINT, userid VARCHAR NOT NULL DEFAULT("default_user"), pageid VARCHAR NOT NULL DEFAULT("default_page")) ...
-```
-
-A `SELECT` query would now return all the records
-```sql
-ksql> select * from pageviews_default emit changes;
-+---------------+----------------------------------------------------------------+----------------------------------------------------------------+
-|VIEWTIME       |USERID                                                          |PAGEID                                                          |
-+---------------+----------------------------------------------------------------+----------------------------------------------------------------+
-|3              |default_user                                                    |default_page                                                    |
-|3              |user                                                            |page                                                            |
-|2              |default_user                                                    |page1                                                           |
-|null           |user                                                            |page                                                            |
-```
-
 ## Test plan
 There will need to be tests for the following:
 * QTT for queries with `NOT NULL` stream/table sources
@@ -170,7 +164,6 @@ There will need to be tests for the following:
 * `NOT NULL` for Array type (includes updating integration with Schema Registry) - 3 days
 * `NOT NULL` for MAP key and value type (includes updating integration with Schema Registry) - 4 days
 * `NOT NULL` in Struct fields (includes updating integration with Schema Registry) - 4 days
-* (maybe) Implement `NOT NULL DEFAULT` - 1 week
 
 ## Documentation Updates
 Documentation on the `NOT NULL` syntax and how applying it doesn't mean the Kafka topic won't have records 
--- a/design-proposals/klip-47-non-null-columns-and-component-types.md
+++ b/design-proposals/klip-47-non-null-columns-and-component-types.md
@@ -1,0 +1,183 @@
+# KLIP 47 - NON NULL columns and component types
+
+**Author**: Steven Zhang (@stevenpyzhang) | 
+**Release Target**: TBD | 
+**Status**: _In Discussion_ | 
+**Discussion**: TBD-PR-LINK
+
+**tl;dr:** _NOT NULL increases data integrity by putting constraints on column values in streams and tables._
+           
+## Motivation and background
+https://github.com/confluentinc/ksql/issues/4436
+
+The SQL standard allows a column to have a `NOT NULL` constraint. `NOT NULL` indicates that
+`NULL` is not an accepted value for that column. This constraint prevents records with a `NULL` column value from being written into the table. 
+When a user reads from the table, its guaranteed that every table record will have a value in the `NOT NULL` column.
+
+Currently, all columns in KSQL are implicitly nullable. This KLIP would introduce `NOT NULL` syntax so users can apply it as a constraint to ksqlDB columns.
+Columns that don't have `NOT NULL` constraint default to being nullable, which is the current behavior. When a `NOT NULL` constraint is applied to a column, 
+records with null column values should not be written into constrained stream/table via `CSAS/CTAS/Insert Into` queries or `Insert Into Values` statements.
+When a user reads from a ksqlDB stream/table with a `NOT NULL` column, they should not read any records that violate the `NOT NULL` constraint.
+
+Being able to specify `NOT NULL` as part of a stream/table schema would also help improve the integration with Schema Registry.
+All ksqlDB columns must have an internal SQL type associated with it. When ksqlDB converts its internal SQL types to a Connect Schema, the schema is always optional (nullable) because there's no 
+existing mechanism/syntax to indicate the type is not nullable. This leads to some undesirable behavior when users use Schema Registry.
+For example, if a map key is nullable, Schema Registry will interpret the map schema as a list of key/value map entries. Since ksqlDB map keys
+are always nullable, ksqlDB will take schemas with maps and write them to Schema Registry as a list of key/value map entries, which is a confusing behavior for users.
+Having map keys that are `NOT NULL` would allow ksqlDB maps to be represented properly in Schema Registry.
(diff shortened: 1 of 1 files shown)


design-proposals/klip-47-non-null-columns-and-component-types.md
```diff
@@ -0,0 +1,183 @@
...
+### Data Ownership
+Since ksqlDB relies on Kafka to store the data it reads/writes, the owner of the data in a topic isn't solely the ksqlDB service.
+Producers outside ksqlDB are able to write into the topic. This means that the underlying topic for a `NOT NULL` stream/table can still
+have null column records in it. There's no way around this as ksqlDB can't prevent outside producers from writing into the Kafka topics.
+
+With that in mind, the `NOT NULL` constraint can be enforced within the ksqlDB service. When reading records from a source that has a `NOT NULL` 
+column constraint, invalid records will be dropped, and an error sent to the processing log. When attempting to write an invalid
+record to a sink, the record would also be dropped, and an error sent to the processing log. A `INSERT INTO VALUES` statement with
+a null column record would error out. There would be clear documentation for users that `NOT NULL` does not prevent invalid records in the 
+Kafka topic due to records from producers outside ksqlDB.
```
guozhangwang, line 42: nit: `in the source Kafka topic` as we do prevent it to be written to sink topic.

design-proposals/klip-47-non-null-columns-and-component-types.md
```diff
@@ -0,0 +1,183 @@
...
+```sql
+CREATE STREAM valid_data (arr ARRAY<BIGINT NOT NULL>, mapping MAP<VARCHAR NOT NULL, BIGINT NOT NULL>, valid_struct STRUCT<c DOUBLE NOT NULL, d BOOLEAN>) ...
+```
+Records that have column values such as
+* `{arr: [3,4,null] ... }`
+* `{mapping: {"key": null} ...}`
+* `{mapping: {null: 5} ...}`
+* `{valid_struct: {c: null, d: true} ...}`
+
+Would be dropped when reading from the `valid_data` stream and trying to insert record with these column values would error out.
```
guozhangwang, line 129: Maybe we can add another example here illustrating how `NOT NULL` would effect for sink topics. E.g. my understanding is that if we have a `CTAS` from aggregating a stream with the aggregating column defined as `NOT NULL`, and user's UDAF happens to set the return value as null, then when reading from this table we would not see that result from UDAF as well.

design-proposals/klip-47-non-null-columns-and-component-types.md
```diff
@@ -0,0 +1,183 @@
...
+For the rest of the KLIP, I'll refer to any stream/table with a `NOT NULL` column as a `NOT NULL` stream/table.
+
+### Data Ownership
+Since ksqlDB relies on Kafka to store the data it reads/writes, the owner of the data in a topic isn't solely the ksqlDB service.
+Producers outside ksqlDB are able to write into the topic. This means that the underlying topic for a `NOT NULL` stream/table can still
+have null column records in it. There's no way around this as ksqlDB can't prevent outside producers from writing into the Kafka topics.
+
+With that in mind, the `NOT NULL` constraint can be enforced within the ksqlDB service. When reading records from a source that has a `NOT NULL` 
+column constraint, invalid records will be dropped, and an error sent to the processing log. When attempting to write an invalid
+record to a sink, the record would also be dropped, and an error sent to the processing log. A `INSERT INTO VALUES` statement with
```
guozhangwang, line 40: What we would include in that processing log entry?

design-proposals/klip-47-non-null-columns-and-component-types.md
```diff
@@ -0,0 +1,183 @@
...
+in that column, otherwise there would be rows that violate the `NOT NULL` constraint.
+
+In this case, the user needs to either `UPDATE` all existing `NULL` 
+values in that column or specify a `DEFAULT` value for that column. All existing rows with `NULL` would have the default value added to that column, 
+so that the NOT NULL constraint is not violated (This is only done if the `NOT NULL` constraint is used in conjunction with `DEFAULT`, if a 
+column with no constraints has a `DEFAULT` applied to it, the existing rows with null columns won't be updated).
+
+When `DEFAULT` is specified, instead of dropping the record, the default value would be applied to that record's column.
+```sql
+CREATE stream pageviews_default(viewtime BIGINT, userid VARCHAR NOT NULL DEFAULT("default_user"), pageid VARCHAR NOT NULL DEFAULT("default_page")) ...
```
spena, line 142: Other DBs use `DEFAULT "default_user"` without parenthesis. Let's use the standard when defining default values.

design-proposals/klip-47-non-null-columns-and-component-types.md
```diff
@@ -0,0 +1,183 @@
...
+```
+Records that have column values such as
+* `{arr: [3,4,null] ... }`
+* `{mapping: {"key": null} ...}`
+* `{mapping: {null: 5} ...}`
+* `{valid_struct: {c: null, d: true} ...}`
+
+Would be dropped when reading from the `valid_data` stream and trying to insert record with these column values would error out.
+
+## NOT NULL DEFAULT
```
spena, line 131: Would the DEFAULT work only along NOT NULL? What about when inserting data and not specifying a column that has a default value, would that use the default value?

design-proposals/klip-47-non-null-columns-and-component-types.md
```diff
@@ -0,0 +1,183 @@
...
+
+For a stream/table created from a `CSAS/CTAS`, it would be possible for the user to enforce a `NOT NULL` constraint by casting the a `Select` value
+with a `NOT NULL` type.
+
+```sql
+CREATE STREAM valid_data as select CAST(viewtime + 5 as BIGINT NOT NULL) ...
+```
+The `CastEvaluator` class would be updated accordingly to generate the code that prevents null records from being written into the sink
+
+## Public APIS
```
spena, line 70: You mentioned the use of `ALTER` in the `NOT NULL DEFAULT` section. Are you going to support `ALTER` for defining NOT NULL?

design-proposals/klip-47-non-null-columns-and-component-types.md
```diff
@@ -0,0 +1,183 @@
...
+* Allowing users to specify `NOT NULL` for columns
+* Prevent reading records with a null column value from `NOT NULL` streams/tables
+* Prevent writing records with a null column value to `NOT NULL` streams/tables
+* Allow users to specify `NOT NULL` for array type, map key/value type, and struct field types
+* (maybe) Allow users to specify a default value when applying `NOT NULL`
+
+## Design
+This design section is only focused on `NOT NULL` for now until we decide whether `DEFAULT` is required or not.
+
+We can add a new `optional` variable to the `SqlType` class to indicate whether the type is optional or not. A `SqlType` variable is in each `Column` object so
```
spena, line 58: The `SqlType` just defines a type. I don't we should add column constraints here. What about the `Column` class? It seems the right class to add different column attributes, like not_null, default_value, etc.

design-proposals/klip-47-non-null-columns-and-component-types.md
```diff
@@ -0,0 +1,183 @@
...
+## Motivation and background
+https://github.com/confluentinc/ksql/issues/4436
+
+The SQL standard allows a column to have a `NOT NULL` constraint. `NOT NULL` indicates that
+`NULL` is not an accepted value for that column. This constraint prevents records with a `NULL` column value from being written into the table. 
+When a user reads from the table, its guaranteed that every table record will have a value in the `NOT NULL` column.
+
+Currently, all columns in KSQL are implicitly nullable. This KLIP would introduce `NOT NULL` syntax so users can apply it as a constraint to ksqlDB columns.
+Columns that don't have `NOT NULL` constraint default to being nullable, which is the current behavior. When a `NOT NULL` constraint is applied to a column, 
+records with null column values should not be written into constrained stream/table via `CSAS/CTAS/Insert Into` queries or `Insert Into Values` statements.
```
mjsax, line 19: I don't think we should include `CSAS/CTAS` as the result of a query cannot be constraints IMHO.

If people want to ensure that no `NULL` values are in the result, they need to specify corresponding filters via `WHERE` or `HAVING` to drop columns that have a `NULL` value.

design-proposals/klip-47-non-null-columns-and-component-types.md
mjsax: > When attempting to write an invalidrecord to a sink

This seems only to apply to `INSERT INTO AS SELECT` (might be worth to clarify).

design-proposals/klip-47-non-null-columns-and-component-types.md
mjsax: Sound like implementation details? Seems we can omit those in the KLIP?

design-proposals/klip-47-non-null-columns-and-component-types.md
```diff
@@ -0,0 +1,183 @@
...
+* (maybe) Allow users to specify a default value when applying `NOT NULL`
+
+## Design
+This design section is only focused on `NOT NULL` for now until we decide whether `DEFAULT` is required or not.
+
+We can add a new `optional` variable to the `SqlType` class to indicate whether the type is optional or not. A `SqlType` variable is in each `Column` object so 
+this would be a convenient way to store the `NOT NULL` constraint for each column. When doing code generation in `SqlToJavaVisitor`, we can then add null filtering code
+in `visitUnqualifiedColumnReference` if the `optional` is present in the `SqlType` of the column. The set of `SchemaConverters` would need to be updated to account for non-optional `SqlType` 
+
+For a stream/table created from a `CSAS/CTAS`, it would be possible for the user to enforce a `NOT NULL` constraint by casting the a `Select` value
```
mjsax, line 62: As mentioned above. I don't think we should include `CSAS/CTAS` but only `CS/CT` statements.

> by casting the a `Select` value with a `NOT NULL` type.

Don't understand this. I think people would need to either actively filter those record to not include them in the result, of use a `IFNULL` in the SELECT clause to replace any `NULL` with some non-`NULL` value.

design-proposals/klip-47-non-null-columns-and-component-types.md
```diff
@@ -0,0 +1,183 @@
...
+with a `NOT NULL` type.
+
+```sql
+CREATE STREAM valid_data as select CAST(viewtime + 5 as BIGINT NOT NULL) ...
```
mjsax, line 66: If `viewtime` is `NULL`, `viewtime + 5` is also `NULL` -- not sure what the purpose or semantics of this statement should be?

design-proposals/klip-47-non-null-columns-and-component-types.md
```diff
@@ -0,0 +1,183 @@
...
+
+ksql> INSERT INTO pageviews_no_null VALUES (NULL, 'user3', 'page23');
+Failed to insert values into 'PAGEVIEWS': null violates the NOT NULL constraint for field "VIEWTIME"
+```
+
+`NOT NULL` for arrays, maps, struct
+```sql
+CREATE STREAM valid_data (arr ARRAY<BIGINT NOT NULL>, mapping MAP<VARCHAR NOT NULL, BIGINT NOT NULL>, valid_struct STRUCT<c DOUBLE NOT NULL, d BOOLEAN>) ...
+```
+Records that have column values such as
```
mjsax, line 123: Just to make sure: We would drop the full record? (We won't only remove the corresponding array/map/struct entry?)

design-proposals/klip-47-non-null-columns-and-component-types.md
spena: Example:
```
ksql> create stream s1 (id int, name string default 'unamed') with ...
ksql> insert into s1(id) values (1);
ksql> select * from s1 emit changes:
id |  name
1  |  'unamed'
```

I know DEFAULT is not part of the scope, but wouldn't be easier to handle NOT NULL and DEFAULT separate than forcing to use them together?

guozhangwang: +1. I also feel it's better to decouple the NOT NULL from DEFAULT syntax. For this specific use case of `ALTER`, I'm a bit leaning towards not using ALTER to modify the semantics constraint at the moment; and if we do want to do this now, maybe we should use the syntax like `ADD CONSTRAINT`.

design-proposals/klip-47-non-null-columns-and-component-types.md
stevenpyzhang:
```
CREATE STREAM test (age BIGINT NOT NULL) ...
CREATE STREAM test1 as select * from test;
```
What I meant here was that theoretically the user could set up an outside producer to produce to the sink topic `test1` and that could lead to null column records in the sink topic in addition to them already being present in the source topic.

design-proposals/klip-47-non-null-columns-and-component-types.md
stevenpyzhang: I added another example with an aggregate function

stevenpyzhang (commented):
Thanks for the reviews @spena @guozhangwang @mjsax . I had gotten side tracked by something so it's been a while for me to update the KLIP.

I've removed the references to `DEFAULT` since I think the consensus is that they shouldn't have to be tied together. I had originally brought it up as a way to not drop records that violate a `NOT NULL` constraint, but it should be ok to drop these invalid records.

 The main contentious point right now is whether or not we allow users to apply constraint to the derived stream/table created by CSAS/CTAS. I think we should have some input from product here to see if that's functionality that users want.

design-proposals/klip-47-non-null-columns-and-component-types.md
mjsax: I guess this does back to my comment about `CSAS/CTAS` -- personally, I don't think it make sense... `NOT NULL` is a _constraint_ and not a "type extension" -- there is no such type as `BIGING NOT NULL`.

design-proposals/klip-47-non-null-columns-and-component-types.md
guozhangwang: Just my 2c: 

1) if we do not consider UDF/UDAF for now, then `NOT NULL` constraint should always be "inherited" in the derived table/streams since no built-in operators so far would map a not-null field into a null field in the derived entities.

2) in addition, for built-in operators, I think they should be able to auto handling null fields, so users should not need to worry if they need to explicitly specify "I want to omit those null-fielded records in the expression/etc". @stevenpyzhang please correct me if I'm wrong.

3) for built-in UDF/UDAF, I think user's should be responsible to make sure that their implementation can handle null fields, and decide whether the output could be null or not-null.

With all that, I feel we do not need to support adding constraint on the CT/SAS statements, since either it should be enforced at the complied code level, or it should not be required. Thoughts?

mjsax: > constraint should always be "inherited" in the derived table/streams since no built-in operators so far would map a not-null field into a null field in the derived entities

I don't think this hold true. For example, left/outer joins may produce `NULL` even if no input stream/table contains any `NULL`s.

However, I don't think this is the point I try to make: If you define a _materialized view_, you define a "dependency" from the inputs to the output and this dependency is described by the _semantics of the query_: **The output stream/table must always contain all rows that are the result of the query.** Or to rephrase: if you do a (transient) query you define a query that gives you a result that may or may not contain `NULL`s. The `CMV` prefix **only** tells the system to _store_ the result (and to also maintain it to be "in-sync" if the input changes); the `CMV` statement must **not modify** the result by any means IMHO. Thus, if you have for example a left outer join, and the result contains `NULL`, the result _MUST_ contain this row. Otherwise the contract of the MV definition is violated. -- To me, it seems to be _fundamentally_ incorrect to define a constraint on a _query result_...

mjsax: > constraint should always be "inherited" in the derived table/streams since no built-in operators so far would map a not-null field into a null field in the derived entities

I don't think this hold true. For example, left/outer joins may produce `NULL` even if no input stream/table contains any `NULL`s.

However, I don't think this is the point I try to make: If you define a _materialized view_, you define a "dependency" from the inputs to the output and this dependency is described by the _semantics of the query_: **The output stream/table must always contain all rows that are the result of the query.** Or to rephrase: if you do a (transient) query you define a query that gives you a result that may or may not contain `NULL`s. The `CMV` prefix **only** tells the system to _store_ the result (and to also maintain it to be "in-sync" if the input changes); the `CMV` statement must **not modify** the result by any means IMHO. Thus, if you have for example a left outer join, and the result contains `NULL`, the result _MUST_ contain this row. Otherwise the contract of the MV definition is violated. -- To me, it seems to be _fundamentally_ incorrect to define a constraint on a _query result_...

PeterLindner: @mjsax @guozhangwang if I understand correctly, you both are saying, that constraints shouldn't be set on derived streams/tables directly, but rather infered from the constraints of the source.
Given the following statement:
```sql
CREATE STREAM target 
AS SELECT col1, col2 
FROM source
WHERE col2 IS NOT NULL;
```
If col1 from source is not null, then col1 in target shouldn't be nullable as well and col2 in target is never nullable. Or:
```sql
CREATE STREAM target 
AS SELECT COALESCE(col1, col2) AS col3 
FROM source;
```
If col2 from source is not null, then col3 in target shouldn't be nullable. This means any function should declare whether it produces nullable or not nullable results (and under which circumstances).
Also some clauses may affect nullability, eg. inner join forces not null for join criteria, left/outer join forces nullable.

In general it should be possible to derive the constraints of the target stream/table from the constraints of the source and the statement which creates it. Would you agree wirh that?

mjsax:
I just stubbled across https://docs.databricks.com/data-engineering/delta-live-tables/delta-live-tables-expectations.html#delta-live-tables-data-quality-constraints

They use:
```
CONSTRAINT <nameMe> EXPECT (<column> IS NOT NULL) ON VIOLATION DROP ROW
```

Seems to be more explicit and more user friendly?

PR #7554, opened 2021-05-19
fix: better error message on illegal arithmetic with NULL values

cprasad1:
### Description 
This PR makes the error handling better when we try to do arithmetic with `NULL`. 
```
ksql> SELECT NULL + 1 FROM s1 EMIT CHANGES;
Error processing expression (null + 1). Arithmetic on types null and INTEGER are not supported.

ksql> SELECT MAP(1 := 'cat') + NULL FROM s1 EMIT CHANGES;
Error processing expression: (MAP(1:='cat') + null). Arithmetic on types MAP<INTEGER, STRING> and null are not supported.

ksql> SELECT Array[1,2,3] + NULL FROM s1 EMIT CHANGES;
Error processing expression: (ARRAY[1, 2, 3] + null). Arithmetic on types ARRAY<INTEGER> and null are not supported.

ksql> SELECT 5.0 / NULL FROM s1 EMIT CHANGES;
Error processing expression: (5.0 / null). Arithmetic on types DECIMAL(2, 1) and null are not supported.

ksql> SELECT NULL * NULL FROM s1 EMIT CHANGES;
Error processing expression: (null * null). Arithmetic on types null and null are not supported.
```

### Testing done 
- manual testing
- Unit testing
- RQTT

### Reviewer checklist

### DIFF
--- a/ksqldb-common/src/main/java/io/confluent/ksql/schema/Operator.java
+++ b/ksqldb-common/src/main/java/io/confluent/ksql/schema/Operator.java
@@ -31,7 +31,7 @@ public enum Operator {
     public SqlType resultType(final SqlType left, final SqlType right) {
       if (left == null || right == null) {
         throw new KsqlException(
-                String.format("Arithmetic on types %s and %s are not supported.", left, right));
+                String.format("Arithmetic on types %s and %s are not supported", left, right));
       }
       if (left.baseType() == SqlBaseType.STRING && right.baseType() == SqlBaseType.STRING) {
         return SqlTypes.STRING;
--- a/ksqldb-common/src/main/java/io/confluent/ksql/schema/Operator.java
+++ b/ksqldb-common/src/main/java/io/confluent/ksql/schema/Operator.java
@@ -30,7 +30,8 @@ public enum Operator {
     @Override
     public SqlType resultType(final SqlType left, final SqlType right) {
       if (left == null || right == null) {
-        throw new KsqlException("Arithmetic on types " + left + " and " + right + " are not supported.");
+        throw new KsqlException(
+                String.format("Arithmetic on types %s and %s are not supported.", left, right));
       }
       if (left.baseType() == SqlBaseType.STRING && right.baseType() == SqlBaseType.STRING) {
         return SqlTypes.STRING;
@@ -65,7 +66,8 @@ public enum Operator {
    */
   public SqlType resultType(final SqlType left, final SqlType right) {
     if (left == null || right == null) {
-      throw new KsqlException("Arithmetic on types " + left + " and " + right + " are not supported.");
+      throw new KsqlException(
+              String.format("Arithmetic on types %s and %s are not supported.", left, right));
     }
     if (left.baseType().isNumber() && right.baseType().isNumber()) {
       if (left.baseType().canImplicitlyCast(right.baseType())) {
--- a/ksqldb-common/src/main/java/io/confluent/ksql/schema/Operator.java
+++ b/ksqldb-common/src/main/java/io/confluent/ksql/schema/Operator.java
@@ -28,7 +28,7 @@ import java.util.function.BinaryOperator;
 public enum Operator {
   ADD("+", SqlDecimal::add) {
     @Override
-    public SqlType resultType(final SqlType left, final SqlType right) {
+    public SqlType resultType(final SqlType left, final SqlType right) throws KsqlException {
       checkForNullTypes(left, right);
 
       if (left.baseType() == SqlBaseType.STRING && right.baseType() == SqlBaseType.STRING) {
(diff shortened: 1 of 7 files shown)


ksqldb-functional-tests/src/test/resources/rest-query-validation-tests/push-queries.json
```diff
@@ -927,6 +927,66 @@
           {"finalMessage":"Limit Reached"}
         ]}
       ]
+    },
+    {
+      "name": "NULL Arithmetic Behavior - INTEGER addition",
```
agavra, line 932: instead of RQTT I think it makes sense to have this in our normal QTT tests `math.json` would be a good fit. They're much lighter weight and also have historical tests

ksqldb-common/src/main/java/io/confluent/ksql/schema/Operator.java
```diff
@@ -29,6 +29,10 @@
   ADD("+", SqlDecimal::add) {
     @Override
     public SqlType resultType(final SqlType left, final SqlType right) {
+      if (left == null || right == null) {
```
agavra, line 32: does the interpreter code use this as well? we should make sure we have tests for this with the interpreter enabled and disabled (cc @AlanConfluent can point you how to do that if you're lost)

ksqldb-common/src/main/java/io/confluent/ksql/schema/Operator.java
```diff
@@ -61,6 +65,10 @@ public String getSymbol() {
    * @return the result schema.
    */
   public SqlType resultType(final SqlType left, final SqlType right) {
+    if (left == null || right == null) {
```
agavra, line 68: nit: let's extract this into a single `checkForNull` method and reuse it above (makes sure the error messages are identical)

ksqldb-execution/src/main/java/io/confluent/ksql/execution/util/ExpressionTypeManager.java
```diff
@@ -153,14 +153,20 @@ public SqlType getExpressionSqlType(
...
       process(node.getRight(), context);
       final SqlType rightType = context.getSqlType();
 
-      final SqlType resultType = node.getOperator().resultType(leftType, rightType);
+      final SqlType resultType;
+      try {
+        resultType = node.getOperator().resultType(leftType, rightType);
+      } catch (KsqlException e) {
+        throw new KsqlException(String.format(
+                "Error processing expression: %s. %s", node.toString(), e.getMessage()));
```
agavra, line 168: nice! thanks for the error message improvement :) might be good to also include `e` as the cause so that we don't lose the stack trace

agavra (commented):
Nice! A few nits before merging

ksqldb-engine/src/test/java/io/confluent/ksql/execution/ExpressionEvaluatorParityTest.java
```diff
@@ -174,10 +174,24 @@ public void shouldDoArithmetic() throws Exception {
   @Test
   public void shouldDoArithmetic_nulls() throws Exception {
     ordersRow = GenericRow.genericRow(null, null, null, null, null, null, null, null, null);
-    assertOrdersError("1 + null",  compileTime("Unexpected error generating code for Test"),
-        compileTime("Unexpected error generating code for expression: (1 + null)"));
-    assertOrdersError("'a' + null",  compileTime("Unexpected error generating code for Test"),
-        compileTime("Unexpected error generating code for expression: ('a' + null)"));
+    assertOrdersError("1 + null",  compileTime("Error processing expression: (1 + null)."),
```
AlanConfluent, line 177: I think most of these should effectively have the same error message for each of the interpreter and compiler, since it's the same thing throwing the error. So you can use the same message, to make clear they are the same error.

AlanConfluent (approved):
Maybe it's outside the scope of this PR, but might be good to have tests for other expression types, e.g. `COL > 1 AND NULL` where we have a NULL literal used in other areas.  I know this is a lot of surface area potentially.

AlanConfluent:
LGTM

ksqldb-functional-tests/src/test/resources/rest-query-validation-tests/push-queries.json
cprasad1: the tests in https://github.com/confluentinc/ksql/blob/7523e1e5c2d850a1f0d210d1680e6307ef351697/ksqldb-engine/src/test/java/io/confluent/ksql/execution/ExpressionEvaluatorParityTest.java ensure that

ksqldb-common/src/main/java/io/confluent/ksql/schema/Operator.java
cprasad1: the tests in `ExpressionEvaluatorParityTest.java` take care of that

guozhangwang (approved):
LGTM!

agavra (approved):
Thanks @cprasad1!

PR #7555, opened 2021-05-19
experimental: add keepalive pings to push queries

vvcephei:
### Description 

### Testing done 

### Reviewer checklist

### DIFF
--- a/ksqldb-rest-app/src/main/java/io/confluent/ksql/api/server/DelimitedQueryStreamResponseWriter.java
+++ b/ksqldb-rest-app/src/main/java/io/confluent/ksql/api/server/DelimitedQueryStreamResponseWriter.java
@@ -54,7 +54,11 @@ public class DelimitedQueryStreamResponseWriter implements QueryStreamResponseWr
 
   @Override
   public QueryStreamResponseWriter writeRow(final GenericRow row) {
-    response.write(ServerUtils.serializeObject(row.values()).appendString("\n"));
+    if (row == null) {
+      response.write("\n");
+    } else {
+      response.write(ServerUtils.serializeObject(row.values()).appendString("\n"));
+    }
     return this;
   }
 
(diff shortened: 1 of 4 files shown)


ksqldb-rest-app/src/main/java/io/confluent/ksql/api/server/DelimitedQueryStreamResponseWriter.java
```diff
@@ -54,7 +54,11 @@ public QueryStreamResponseWriter writeMetadata(final QueryResponseMetadata metaD
 
   @Override
   public QueryStreamResponseWriter writeRow(final GenericRow row) {
-    response.write(ServerUtils.serializeObject(row.values()).appendString("\n"));
+    if (row == null) {
+      response.write("\n");
```
guozhangwang, line 58: Why start a new line if `row` is null? If we have less traffic would that cause a lot of new lines?

ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/resources/streaming/QueryStreamWriter.java
```diff
@@ -81,7 +82,7 @@ public void write(final OutputStream out) {
             disconnectCheckInterval,
             TimeUnit.MILLISECONDS
         );
-        if (row != null) {
+        if (row != null && (row.key() != null || row.value() != null)) {
```
AlanConfluent, line 85: Just curious, is this required?  It seems like you're just sending a null row, not null parts.

ksqldb-rest-app/src/main/java/io/confluent/ksql/api/server/DelimitedQueryStreamResponseWriter.java
```diff
@@ -54,7 +54,11 @@ public QueryStreamResponseWriter writeMetadata(final QueryResponseMetadata metaD
 
   @Override
   public QueryStreamResponseWriter writeRow(final GenericRow row) {
-    response.write(ServerUtils.serializeObject(row.values()).appendString("\n"));
+    if (row == null) {
```
AlanConfluent, line 57: Is this preferable vs a sentinel?

AlanConfluent:
WRT the write timeout value utilized here:
https://github.com/confluentinc/ksql/blob/master/ksqldb-rest-app/src/main/java/io/confluent/ksql/api/server/ResponseOutputStream.java#L110

This will still trigger if the client isn't actually reading the values we're sending, so your above solution will prevent the idle case, but not the client isn't consuming case.  I think the latter is fine (since it's somewhat of a client error), but just wanted to point that out since it was mentioned in the issue.

Issue #6970, opened 2021-02-08
Allow push queries to be streamed for more than ten minutes at a time


vcrfxia:
**Is your feature request related to a problem? Please describe.**

Currently, a single push query may be streamed for a maximum of ten minutes at a time, after which the server closes the connection. ksqlDB should provide a way for users to stream push queries for longer if desired, while still not allowing disconnected clients to hold server resources for unreasonable amounts of time.

**Describe the solution you'd like**

It may make sense to allow users to specify that certain queries should have longer timeouts.

Details here need to be thought out more. There are two places today where the 10-minute timeout is enforced on the server side:
- on the server itself: https://github.com/confluentinc/ksql/blob/9988e6af3266319e8caf978a904f5d55c8986623/ksqldb-rest-app/src/main/java/io/confluent/ksql/api/server/Server.java#L295
- on the query stream writer for push queries issued via the old HTTP streaming endpoint: https://github.com/confluentinc/ksql/blob/9988e6af3266319e8caf978a904f5d55c8986623/ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/resources/streaming/QueryStreamWriter.java#L42

**Additional context**

Example community request for this: https://groups.google.com/g/ksql-users/c/iB6mo1tHgn0/m/dEzpEDHwAAAJ

purplefox:
A common approach with clients to prevent idle connections being timed out would be to send a "ping" request (which does nothing) every few seconds.

vcrfxia:
A workaround that can be used in the meantime (courtesy of @AlexCasdu): https://groups.google.com/g/ksql-users/c/iB6mo1tHgn0/m/ARfID-RoAgAJ

agavra:
Moving to 0.19, we'll need to tackle this as part of #7203

vvcephei:
As far as I can tell from the code, the `WRITE_TIMEOUT_MS` is probably not involved here. That property looks like it's only used to put an upper bound on how long to wait for the client to read our buffered responses. As long as the reader is consuming responses, we shouldn't hit this timeout.

Based on the reports and the proposed workaround, the problem just seems to be the idle timeout.

We can certainly add a heartbeat mechanism to the Java client, but that doesn't solve the general problem for users of the server APIs. Each API caller will have to set up their own heartbeat mechanism.

[...]

PR #7556, opened 2021-05-19
feat: add idle timeout

vvcephei:
### Description 
Expose a config to set the server's idle timeout to work around https://github.com/confluentinc/ksql/issues/6970

### Testing done 
Set the configuration to 5 seconds, ran the server locally, and verified that when I create a transient query on the HTTP2 API, the connection gets closed after 5 seconds of inactivity. I also verified that if I cancel my query (`Ctrl-C` on my `curl` command), the transient query gets shut down on the server.

### Reviewer checklist

### DIFF
--- a/ksqldb-rest-app/src/main/java/io/confluent/ksql/api/server/Server.java
+++ b/ksqldb-rest-app/src/main/java/io/confluent/ksql/api/server/Server.java
@@ -107,6 +107,9 @@ public class Server {
     if (!deploymentIds.isEmpty()) {
       throw new IllegalStateException("Already started");
     }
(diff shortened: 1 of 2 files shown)


agavra (approved):
nice - I think having this config also gives us a workaround for customers that need it. I think we should have this regardless of whether or not we end up doing #7555 

if we are concerned with 24h, we can start the default at 1h and then allow customers to increase it.

guozhangwang (approved):
LGTM, thanks @vvcephei

guozhangwang:
> if we are concerned with 24h, we can start the default at 1h and then allow customers to increase it.

Personally I think 24h is okay :)

PR #7568, opened 2021-05-21
chore: update QTT framework & tests to support improvements on left/outer joins

swist:
### Description 
Trying to fix QTT
Supersedes #7464 and #7561

### Testing done 

### Reviewer checklist

### DIFF
--- a/ksqldb-functional-tests/src/test/resources/test-runner/correct/join/output.json
+++ b/ksqldb-functional-tests/src/test/resources/test-runner/correct/join/output.json
@@ -1,13 +1,7 @@
 {
   "outputs": [
-    {"topic": "LEFT_OUTER_JOIN", "key": 0, "value": {"NAME": "zero", "VALUE": 0, "F1": null, "F2": null}, "timestamp": 0},
     {"topic": "LEFT_OUTER_JOIN", "key": 0, "value": {"NAME": "zero", "VALUE": 0, "F1": "blah", "F2": 50}, "timestamp": 10000},
-    {"topic": "LEFT_OUTER_JOIN", "key": 10, "value": {"NAME": "100", "VALUE": 5, "F1": null, "F2": null}, "timestamp": 11000},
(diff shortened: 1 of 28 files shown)


ksqldb-functional-tests/src/test/resources/test-runner/correct/join/output.json
```diff
@@ -1,13 +1,7 @@
 {
   "outputs": [
-    {"topic": "LEFT_OUTER_JOIN", "key": 0, "value": {"NAME": "zero", "VALUE": 0, "F1": null, "F2": null}, "timestamp": 0},
```
vcrfxia, line 3: @spena can you comment on these changes? The testing tool no longer emits join results for left source inputs that don't have any matches on the left side, even though this is a stream-stream left join. This does not seem correct to me?

vcrfxia: If you want to play around with this yourself, `KsqlTestingToolTest#shouldRunCorrectTests()` fails without these changes.

spena: The results are correct. The new stream-stream left join semantics do not emit results that do not have a join until the window + grace period has expired or a join is found on the right topic. The output of this test is emitting only the joined records. There are changes we must do to test the left/outer joins, though. But that's something we can do in a follow-up in order to unblock master.

vcrfxia: What's the default grace period? Also, what changes are needed in order to begin testing left and outer stream-stream joins again? The fact that we only have test coverage for inner stream-stream joins right now is quite a large test gap. (Agreed that we should merge this PR first to unblock master, though.)

vcrfxia:
We've decided to disable the stream-stream left/outer join enhancements in ksqlDB until the grace period is configurable, so I'm closing this PR in favor of https://github.com/confluentinc/ksql/pull/7569 instead.

PR #7574, opened 2021-05-21
fix: allow for KSQL_GC_LOG_OPTS env variable to be passed through without …

thomas-tomlinson:
…changes if present

### Description 
Allow the KSQL_GC_LOG_OPTS environment variable to be passed through if present.

### Testing done 
Tested the behavior of both providing the KSQL_GC_LOG_OPTS variable, in which case the supplied configuration was used, and omitting it, which reverts to the original behavior.

### Reviewer checklist

### DIFF
--- a/bin/ksql-run-class
+++ b/bin/ksql-run-class
@@ -113,7 +113,9 @@ while [ $# -gt 0 ]; do
       shift 2
       ;;
     -loggc)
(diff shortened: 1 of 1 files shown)


thomas-tomlinson:
> [clabot:check]

[clabot:check]

PR #7576, opened 2021-05-23
chore: fix bugs in FK n-way joins

mjsax:
### Description 
FK-joins need to set their output-format to make n-way join works. Without this fix, we fail with a `ClassCastException` because downstream operators fall back to default ByteArraySerde as no `valueSerde` is set.

All but two n-way join QQT test pass now, but as long as QTT is not fixed, we cannot enable them yet. One test fails with missing result columns (we need to fix forward in a follow up PR). The other fails with an error resolving an alias (this is fixed via #7547)

### Testing done 
Tested locally using `fk-n-way-joins.json`

### Reviewer checklist

### DIFF
--- a/ksqldb-engine/src/main/java/io/confluent/ksql/planner/plan/JoinNode.java
+++ b/ksqldb-engine/src/main/java/io/confluent/ksql/planner/plan/JoinNode.java
@@ -521,13 +521,8 @@ public class JoinNode extends PlanNode implements JoiningNode {
 
       final JoinKey joinKey = joinNode.joinKey;
 
-      final FormatInfo leftValueFormatInfo = joinNode
-          .getLeft()
-          .getLeftmostSourceNode()
-          .getDataSource()
-          .getKsqlTopic()
-          .getValueFormat()
-          .getFormatInfo();
+      final FormatInfo valueFormatInfo = JoiningNode
+          .getValueFormatForSource(joinNode.getLeft()).getFormatInfo();
 
       switch (joinNode.joinType) {
         case LEFT:
@@ -536,7 +531,7 @@ public class JoinNode extends PlanNode implements JoiningNode {
                 rightTable,
                 ((ForeignJoinKey) joinKey).getForeignKeyColumn(),
                 contextStacker,
-                leftValueFormatInfo
+                valueFormatInfo
             );
           } else {
             return leftTable.leftJoin(
@@ -551,7 +546,7 @@ public class JoinNode extends PlanNode implements JoiningNode {
                 rightTable,
                 ((ForeignJoinKey) joinKey).getForeignKeyColumn(),
                 contextStacker,
-                leftValueFormatInfo
+                valueFormatInfo
             );
           } else {
             return leftTable.innerJoin(
--- a/ksqldb-engine/src/main/java/io/confluent/ksql/planner/plan/JoinNode.java
+++ b/ksqldb-engine/src/main/java/io/confluent/ksql/planner/plan/JoinNode.java
@@ -521,13 +521,22 @@ public class JoinNode extends PlanNode implements JoiningNode {
 
       final JoinKey joinKey = joinNode.joinKey;
 
+      final FormatInfo leftValueFormatInfo = joinNode
+          .getLeft()
+          .getLeftmostSourceNode()
+          .getDataSource()
+          .getKsqlTopic()
+          .getValueFormat()
+          .getFormatInfo();
+
       switch (joinNode.joinType) {
         case LEFT:
           if (joinKey.isForeignKey()) {
             return leftTable.foreignKeyLeftJoin(
                 rightTable,
                 ((ForeignJoinKey) joinKey).getForeignKeyColumn(),
-                contextStacker
+                contextStacker,
+                leftValueFormatInfo
             );
           } else {
             return leftTable.leftJoin(
@@ -541,7 +550,8 @@ public class JoinNode extends PlanNode implements JoiningNode {
             return leftTable.foreignKeyInnerJoin(
                 rightTable,
                 ((ForeignJoinKey) joinKey).getForeignKeyColumn(),
-                contextStacker
+                contextStacker,
+                leftValueFormatInfo
             );
           } else {
             return leftTable.innerJoin(
--- a/ksqldb-execution/src/main/java/io/confluent/ksql/execution/plan/ForeignKeyTableTableJoin.java
+++ b/ksqldb-execution/src/main/java/io/confluent/ksql/execution/plan/ForeignKeyTableTableJoin.java
@@ -32,7 +32,7 @@ public class ForeignKeyTableTableJoin<KLeftT, KRightT>
   private final ExecutionStepPropertiesV1 properties;
   private final JoinType joinType;
   private final ColumnName leftJoinColumnName;
-  private final Formats leftFormats;
+  private final Formats formats;
   private final ExecutionStep<KTableHolder<KLeftT>> leftSource;
   private final ExecutionStep<KTableHolder<KRightT>> rightSource;
 
@@ -44,8 +44,8 @@ public class ForeignKeyTableTableJoin<KLeftT, KRightT>
       final JoinType joinType,
       @JsonProperty(value = "leftJoinColumnName", required = true)
       final ColumnName leftJoinColumnName,
-      @JsonProperty(value = "leftFormats", required = true)
-      final Formats leftFormats,
+      @JsonProperty(value = "formats", required = true)
+      final Formats formats,
       @JsonProperty(value = "leftSource", required = true)
       final ExecutionStep<KTableHolder<KLeftT>> leftSource,
       @JsonProperty(value = "rightSource", required = true)
@@ -57,7 +57,7 @@ public class ForeignKeyTableTableJoin<KLeftT, KRightT>
       throw new IllegalArgumentException("OUTER join not supported.");
     }
     this.leftJoinColumnName = requireNonNull(leftJoinColumnName, "leftJoinColumnName");
-    this.leftFormats = requireNonNull(leftFormats, "leftFormats");
+    this.formats = requireNonNull(formats, "formats");
     this.leftSource = requireNonNull(leftSource, "leftSource");
     this.rightSource = requireNonNull(rightSource, "rightSource");
   }
@@ -89,8 +89,8 @@ public class ForeignKeyTableTableJoin<KLeftT, KRightT>
     return leftJoinColumnName;
   }
 
-  public Formats getLeftFormats() {
-    return leftFormats;
+  public Formats getFormats() {
+    return formats;
   }
 
   @Override
@@ -115,12 +115,13 @@ public class ForeignKeyTableTableJoin<KLeftT, KRightT>
     return Objects.equals(properties, that.properties)
         && joinType == that.joinType
         && Objects.equals(leftJoinColumnName, that.leftJoinColumnName)
+        && Objects.equals(formats, that.formats)
         && Objects.equals(leftSource, that.leftSource)
         && Objects.equals(rightSource, that.rightSource);
   }
 
   @Override
   public int hashCode() {
-    return Objects.hash(properties, joinType, leftJoinColumnName, leftSource, rightSource);
+    return Objects.hash(properties, joinType, leftJoinColumnName, formats, leftSource, rightSource);
   }
 }
--- a/ksqldb-execution/src/main/java/io/confluent/ksql/execution/plan/ForeignKeyTableTableJoin.java
+++ b/ksqldb-execution/src/main/java/io/confluent/ksql/execution/plan/ForeignKeyTableTableJoin.java
@@ -32,6 +32,7 @@ public class ForeignKeyTableTableJoin<KLeftT, KRightT>
   private final ExecutionStepPropertiesV1 properties;
   private final JoinType joinType;
   private final ColumnName leftJoinColumnName;
+  private final Formats leftFormats;
   private final ExecutionStep<KTableHolder<KLeftT>> leftSource;
   private final ExecutionStep<KTableHolder<KRightT>> rightSource;
 
@@ -43,6 +44,8 @@ public class ForeignKeyTableTableJoin<KLeftT, KRightT>
       final JoinType joinType,
       @JsonProperty(value = "leftJoinColumnName", required = true)
       final ColumnName leftJoinColumnName,
+      @JsonProperty(value = "leftFormats", required = true)
+      final Formats leftFormats,
       @JsonProperty(value = "leftSource", required = true)
       final ExecutionStep<KTableHolder<KLeftT>> leftSource,
       @JsonProperty(value = "rightSource", required = true)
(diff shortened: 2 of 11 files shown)


ksqldb-execution/src/test/java/io/confluent/ksql/execution/plan/ForeignKeyTableTableJoinTest.java
```diff
@@ -43,29 +44,31 @@
   private ExecutionStep<KTableHolder<Struct>> left2;
   @Mock
   private ExecutionStep<KTableHolder<Struct>> right2;
+  @Mock
+  private Formats formats;
```
mjsax, line 48: Not 100% if those changes are correct. While the unit test passed, I don't really understand what we actually test here and how `EqualsTester` and `addEqualityGroup()` actually work

ksqldb-functional-tests/src/test/resources/query-validation-tests/fk-n-way-join.json
```diff
@@ -66,28 +66,28 @@
         "ksql.joins.foreign.key.enable": true
       },
       "statements": [
-        "CREATE TABLE left_table (id1 BIGINT PRIMARY KEY, f1 BIGINT) WITH (kafka_topic='left_topic', format='JSON');",
-        "CREATE TABLE middle_table (id2 BIGINT PRIMARY KEY, f2 BIGINT, other STRING) WITH (kafka_topic='middle_topic', format='JSON');",
-        "CREATE TABLE right_table (id3 BIGINT PRIMARY KEY, f3 BIGINT) WITH (kafka_topic='right_topic', format='JSON');",
-        "CREATE TABLE output AS SELECT id1, id2, f1, f2, f3 FROM left_table JOIN middle_table ON f1 = id2 LEFT JOIN right_table ON id1 = id3;"
+        "CREATE TABLE left_table (l_id BIGINT PRIMARY KEY, foreign_key BIGINT) WITH (kafka_topic='left_topic', format='JSON');",
```
mjsax, line 69: Just some cleanup to align theses test to the easier to read naming schema that we also use in other tests).

ksqldb-functional-tests/src/test/resources/query-validation-tests/fk-n-way-join.json
```diff
@@ -66,28 +66,28 @@
...
+        {"topic": "left_topic", "key": 1, "value": {"FOREIGN_KEY": 8}, "timestamp": 16000},
         {"topic": "left_topic", "key": 1, "value": null, "timestamp": 18000}
       ],
       "outputs": [
-        {"topic": "OUTPUT", "key": 1, "value": {"ID2": 0, "F1": 0, "F2": 100, "F3": null}, "timestamp": 10000},
-        {"topic": "OUTPUT", "key": 1, "value": {"ID2": 0, "F1": 0, "F2": 100, "F3": 4}, "timestamp": 11000},
-        {"topic": "OUTPUT", "key": 1, "value": {"ID2": 0, "F1": 8, "F2": 10, "F3": 4}, "timestamp": 16000},
+        {"topic": "OUTPUT", "key": 1, "value": {"M_ID": 0, "FOREIGN_KEY": 0, "F2": 100, "F3": null}, "timestamp": 10000},
+        {"topic": "OUTPUT", "key": 1, "value": {"M_ID": 0, "FOREIGN_KEY": 0, "F2": 100, "F3": 4}, "timestamp": 11000},
+        {"topic": "OUTPUT", "key": 1, "value": {"M_ID": 8, "FOREIGN_KEY": 8, "F2": 10, "F3": 4}, "timestamp": 16000},
```
mjsax, line 85: Small fix: the expected `M_ID` should be `8` (not `0`) -- same fix for other test cases below.

ksqldb-functional-tests/src/test/resources/query-validation-tests/fk-n-way-join.json
```diff
@@ -165,28 +165,28 @@
...
```
mjsax, line 184: fix

ksqldb-functional-tests/src/test/resources/query-validation-tests/fk-n-way-join.json
```diff
@@ -197,28 +197,28 @@
...
-        {"topic": "OUTPUT", "key": 1, "value": {"LEFT_TABLE_F1": 0, "MIDDLE_TABLE_ID2": 0, "MIDDLE_TABLE_F2": 100, "RIGHT_TABLE_ID3": 1, "RIGHT_TABLE_F3": 4}, "timestamp": 11000},
-        {"topic": "OUTPUT", "key": 1, "value": {"LEFT_TABLE_F1": 8, "MIDDLE_TABLE_ID2": 0, "MIDDLE_TABLE_F2": 10, "RIGHT_TABLE_ID3": 1, "RIGHT_TABLE_F3": 4}, "timestamp": 16000},
+        {"topic": "OUTPUT", "key": 1, "value": {"LEFT_TABLE_FOREIGN_KEY": 0, "MIDDLE_TABLE_M_ID": 0, "MIDDLE_TABLE_F2": 100, "MIDDLE_TABLE_OTHER": "unused", "RIGHT_TABLE_R_ID": null, "RIGHT_TABLE_F3": null}, "timestamp": 10000},
+        {"topic": "OUTPUT", "key": 1, "value": {"LEFT_TABLE_FOREIGN_KEY": 0, "MIDDLE_TABLE_M_ID": 0, "MIDDLE_TABLE_F2": 100, "MIDDLE_TABLE_OTHER": "unused", "RIGHT_TABLE_R_ID": 1, "RIGHT_TABLE_F3": 4}, "timestamp": 11000},
+        {"topic": "OUTPUT", "key": 1, "value": {"LEFT_TABLE_FOREIGN_KEY": 8, "MIDDLE_TABLE_M_ID": 8, "MIDDLE_TABLE_F2": 10, "MIDDLE_TABLE_OTHER": "unused", "RIGHT_TABLE_R_ID": 1, "RIGHT_TABLE_F3": 4}, "timestamp": 16000},
```
mjsax, line 216: fix

ksqldb-functional-tests/src/test/resources/query-validation-tests/fk-n-way-join.json
```diff
@@ -261,28 +261,28 @@
...
+        {"topic": "left_topic", "key": 1, "value": {"FOREIGN_KEY": 8}, "timestamp": 16000},
         {"topic": "left_topic", "key": 1, "value": null, "timestamp": 18000}
       ],
       "outputs": [
-        {"topic": "OUTPUT", "key": 1, "value": {"ID2_ALIAS": 0, "F1": 0, "MT_F2_ALIAS": 100, "F3": null}, "timestamp": 10000},
-        {"topic": "OUTPUT", "key": 1, "value": {"ID2_ALIAS": 0, "F1": 0, "MT_F2_ALIAS": 100, "F3": 4}, "timestamp": 11000},
-        {"topic": "OUTPUT", "key": 1, "value": {"ID2_ALIAS": 0, "F1": 8, "MT_F2_ALIAS": 10, "F3": 4}, "timestamp": 16000},
+        {"topic": "OUTPUT", "key": 1, "value": {"M_ID_ALIAS": 0, "FOREIGN_KEY": 0, "MT_F2_ALIAS": 100, "F3": null}, "timestamp": 10000},
+        {"topic": "OUTPUT", "key": 1, "value": {"M_ID_ALIAS": 0, "FOREIGN_KEY": 0, "MT_F2_ALIAS": 100, "F3": 4}, "timestamp": 11000},
+        {"topic": "OUTPUT", "key": 1, "value": {"M_ID_ALIAS": 8, "FOREIGN_KEY": 8, "MT_F2_ALIAS": 10, "F3": 4}, "timestamp": 16000},
```
mjsax, line 280: fix

ksqldb-functional-tests/src/test/resources/query-validation-tests/fk-n-way-join.json
```diff
@@ -197,28 +197,28 @@
...
+        {"topic": "OUTPUT", "key": 1, "value": {"LEFT_TABLE_FOREIGN_KEY": 8, "MIDDLE_TABLE_M_ID": 8, "MIDDLE_TABLE_F2": 10, "MIDDLE_TABLE_OTHER": "unused", "RIGHT_TABLE_R_ID": 1, "RIGHT_TABLE_F3": 4}, "timestamp": 16000},
         {"topic": "OUTPUT", "key": 1, "value": null, "timestamp": 18000}
       ],
       "post": {
         "sources": [
-          {"name": "OUTPUT", "type": "table", "schema": "LEFT_TABLE_ID1 BIGINT KEY, LEFT_TABLE_F1 BIGINT, MIDDLE_TABLE_ID2 BIGINT, MIDDLE_TABLE_F2 BIGINT, RIGHT_TABLE_ID3 BIGINT, RIGHT_TABLE_F3 BIGINT"}
+          {"name": "OUTPUT", "type": "table", "schema": "LEFT_TABLE_L_ID BIGINT KEY, LEFT_TABLE_FOREIGN_KEY BIGINT, MIDDLE_TABLE_M_ID BIGINT, MIDDLE_TABLE_F2 BIGINT, MIDDLE_TABLE_OTHER VARCHAR, RIGHT_TABLE_R_ID BIGINT, RIGHT_TABLE_F3 BIGINT"}
         ]
       }
     },
```
mjsax, line 224: The next test `Should allow fk join at start of n-way join - qualified select *` is the one that fails with missing output columns. All column of the middle table, that are included via `middle.*` are missing.

ksqldb-functional-tests/src/test/resources/query-validation-tests/fk-n-way-join.json
```diff
@@ -261,28 +261,28 @@
         "ksql.joins.foreign.key.enable": true
       },
       "statements": [
-        "CREATE TABLE left_table (id1 BIGINT PRIMARY KEY, f1 BIGINT) WITH (kafka_topic='left_topic', format='JSON');",
```
mjsax, line 264: This test fails with an alias resolution error. Fixed via #7547.

ksqldb-engine/src/main/java/io/confluent/ksql/planner/plan/JoinNode.java
```diff
@@ -521,13 +521,22 @@ private StreamToTableJoiner(
...
+
       switch (joinNode.joinType) {
         case LEFT:
           if (joinKey.isForeignKey()) {
             return leftTable.foreignKeyLeftJoin(
                 rightTable,
                 ((ForeignJoinKey) joinKey).getForeignKeyColumn(),
-                contextStacker
+                contextStacker,
+                leftValueFormatInfo
```
vcrfxia, line 539: nit: use `JoiningNode.getValueFormatForSource(joinNode.left).getFormatInfo()` rather than duplicating the code above.

ksqldb-execution/src/main/java/io/confluent/ksql/execution/plan/ForeignKeyTableTableJoin.java
```diff
@@ -43,6 +44,8 @@ public ForeignKeyTableTableJoin(
       final JoinType joinType,
       @JsonProperty(value = "leftJoinColumnName", required = true)
       final ColumnName leftJoinColumnName,
+      @JsonProperty(value = "leftFormats", required = true)
```
vcrfxia, line 47: It's not important that these formats are from the left source, right? The purpose is to record the formats used as the output for this step. If this is true, let's rename this to `outputFormats` or `resultFormats` or something similar, to avoid confusion in the future.

ksqldb-execution/src/test/java/io/confluent/ksql/execution/plan/ForeignKeyTableTableJoinTest.java
vcrfxia: This change is not correct. The whole point of the test in this file is to ensure that no one adds a new field that should be considered in the `equals()` and `hashCode()` methods without properly updating the methods. We need to update those methods in ForeignKeyTableTableJoin to consider the new field, and we need to update this test to catch any future regressions here.

ksqldb-streams/src/main/java/io/confluent/ksql/execution/streams/ForeignKeyTableTableJoinBuilder.java
```diff
@@ -29,28 +34,45 @@ private ForeignKeyTableTableJoinBuilder() {
...
     switch (join.getJoinType()) {
-      case LEFT:
-        result = left.getTable().leftJoin(
+      case INNER:
+        result = left.getTable().join(
             right.getTable(),
             joinParams.getKeyExtractor(),
-            joinParams.getJoiner()
+            joinParams.getJoiner(),
+            Materialized.with(null, valSerde)
```
vcrfxia, line 67: Just checking: is it correct to leave the key serde as null and only pass in value serde? Does this reset any key serde information that is already present?

vcrfxia (commented):
Thanks for the changes @mjsax . I'm still not unclear on why these changes are needed, though. What's different about FK joins and PK table-table joins such that we don't need to set the result value serde on PK table-table joins for them to work as part of n-way joins, but we do for FK joins?

mjsax (commented):
>  I'm still not clear on why these changes are needed, though. What's different about FK joins and PK table-table joins such that we don't need to set the result value serde on PK table-table joins for them to work as part of n-way joins, but we do for FK joins?

@vcrfxia The difference is, that a FK-join result must be materialized when it is the input to a downstream join. Thus we need to know the value serde. For PK-joins, we don't need to materialize them though if they are the result to a downstream join, because we (ie, Kafka Streams) can dynamically recompute it base on the upstream state stores.

ksqldb-execution/src/test/java/io/confluent/ksql/execution/plan/ForeignKeyTableTableJoinTest.java
```diff
@@ -43,29 +44,36 @@
...
-            new ForeignKeyTableTableJoin<>(props2, INNER, JOIN_COLUMN_NAME, left1, right1)
+            new ForeignKeyTableTableJoin<>(props1, LEFT, JOIN_COLUMN_NAME, formats1, left1, right1)
         )
         .addEqualityGroup(
-            new ForeignKeyTableTableJoin<>(props1, LEFT, JOIN_COLUMN_NAME, left1, right1)
+            new ForeignKeyTableTableJoin<>(props1, INNER, JOIN_COLUMN_NAME_2, formats1, left1, right1)
         )
         .addEqualityGroup(
-            new ForeignKeyTableTableJoin<>(props1, INNER, JOIN_COLUMN_NAME_2, left1, right1)
+            new ForeignKeyTableTableJoin<>(props1, INNER, JOIN_COLUMN_NAME, formats2, left2, right1)
```
vcrfxia, line 70:
```suggestion
            new ForeignKeyTableTableJoin<>(props1, INNER, JOIN_COLUMN_NAME, formats2, left1, right1)
```

vcrfxia (approved):
Thanks @mjsax -- LGTM besides the comment inline, and the failing build.

ksqldb-execution/src/test/java/io/confluent/ksql/execution/plan/ForeignKeyTableTableJoinTest.java
mjsax: Good catch!

guozhangwang (commented):
Sorry for reviewing the PR late, made a pass and LGTM too.

PR #7578, opened 2021-05-24
chore: relax copartitioning requirements for FK joins + disallow on windowed tables

vcrfxia:
### Description 

Two minor fixes in this PR:
- Do not require copartitioning for foreign key joins. The existing QTTs for this (currently disabled) are now passing except for the requisite changes to the QTT framework (see https://github.com/confluentinc/ksql/pull/7576 for more).
- Disallow foreign key joins on windowed tables, since it doesn't make sense to extract a column/expression from the left source to match a windowed key from the right source. ksqlDB does not support persistent queries on windowed tables so this is only relevant for push queries, so I've added a corresponding RQTT.

### Testing done 

Added tests. (Copartitioning tests will be enabled in a future PR.)

### Reviewer checklist

### DIFF
--- a/ksqldb-engine/src/main/java/io/confluent/ksql/planner/LogicalPlanner.java
+++ b/ksqldb-engine/src/main/java/io/confluent/ksql/planner/LogicalPlanner.java
@@ -778,6 +778,12 @@ public class LogicalPlanner {
           joinInfo.hasFlippedJoinCondition()
       );
 
+      if (((DataSourceNode) rightNode).isWindowed()) {
+        throw new KsqlException(
+           "Foreign-key table-table joins are not supported on windowed tables."
+        );
+      }
+
       return Optional.of(fkColumnReference);
     } else {
       throw new KsqlException(String.format(
--- a/ksqldb-engine/src/main/java/io/confluent/ksql/planner/plan/JoinNode.java
+++ b/ksqldb-engine/src/main/java/io/confluent/ksql/planner/plan/JoinNode.java
@@ -176,7 +176,9 @@ public class JoinNode extends PlanNode implements JoiningNode {
   @Override
   public SchemaKStream<?> buildStream(final PlanBuildContext buildContext) {
 
-    ensureMatchingPartitionCounts(buildContext.getServiceContext().getTopicClient());
+    if (!joinKey.isForeignKey()) {
+      ensureMatchingPartitionCounts(buildContext.getServiceContext().getTopicClient());
+    }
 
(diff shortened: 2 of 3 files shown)


ksqldb-engine/src/main/java/io/confluent/ksql/planner/LogicalPlanner.java
```diff
@@ -778,6 +778,12 @@ private static void verifyStreamTableJoin(final JoinInfo joinInfo, final PlanNod
           joinInfo.hasFlippedJoinCondition()
       );
 
+      if (((DataSourceNode) rightNode).isWindowed()) {
+        throw new KsqlException(
+           "Foreign-key table-table joins are not supported on windowed tables."
```
mjsax, line 783: We don't support any type of join for windowed tables, right, thus I am wondering if this is the right place for this check? (Also, we would need to check both inputs, not just the right input I guess?)

ksqldb-engine/src/main/java/io/confluent/ksql/planner/plan/JoinNode.java
```diff
@@ -176,7 +176,9 @@ public PlanNode getRight() {
   @Override
   public SchemaKStream<?> buildStream(final PlanBuildContext buildContext) {
 
-    ensureMatchingPartitionCounts(buildContext.getServiceContext().getTopicClient());
+    if (!joinKey.isForeignKey()) {
```
mjsax, line 179: Can we add a QTT for this?

Btw: do we already have a QTT for different key formats that we should also support for FK-joins.

vcrfxia: See PR description -- we already have a QTT for this but we can't enable it until the QTT fixes are in: https://github.com/confluentinc/ksql/blob/master/ksqldb-functional-tests/src/test/resources/query-validation-tests/fk-join.json#L659

ksqldb-engine/src/main/java/io/confluent/ksql/planner/LogicalPlanner.java
vcrfxia: > We don't support any type of join for windowed tables, right, thus I am wondering if this is the right place for this check?

We do support PK joins on windowed tables (push queries only).

> Also, we would need to check both inputs, not just the right input I guess?

There's a separate check in the code to ensure that you cannot join a windowed source to a non-windowed source.

mjsax (approved):

PR #7579, opened 2021-05-24
chore: ignore FK internal topics from QTT tests

spena:
### Description 
The loop in `TestExecutor` that calls the `topicInfoCache.get()` does not do anything with the resulted topic info. Seesm it only checks the get() call works. This `get()` call is causing the errors with QTT when determining the key formats. This PR filters out those internal FK topics to allow QTT for FK joins to pass.

### Testing done 

### Reviewer checklist

### DIFF
--- a/ksqldb-functional-tests/src/main/java/io/confluent/ksql/test/tools/TestExecutor.java
+++ b/ksqldb-functional-tests/src/main/java/io/confluent/ksql/test/tools/TestExecutor.java
@@ -15,7 +15,6 @@
 
 package io.confluent.ksql.test.tools;
 
-import static io.confluent.ksql.test.tools.TopicInfoCache.TOPIC_PATTERN_PREFIX;
 import static java.util.Objects.requireNonNull;
 import static org.hamcrest.MatcherAssert.assertThat;
 import static org.hamcrest.Matchers.hasEntry;
@@ -75,10 +74,7 @@ import java.util.Optional;
 import java.util.OptionalInt;
 import java.util.Set;
 import java.util.function.Function;
-import java.util.function.Predicate;
-import java.util.regex.Pattern;
 import java.util.stream.Collectors;
-
 import org.apache.kafka.clients.consumer.ConsumerConfig;
 import org.apache.kafka.clients.producer.ProducerRecord;
 import org.apache.kafka.streams.StreamsConfig;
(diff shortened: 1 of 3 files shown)


ksqldb-functional-tests/src/main/java/io/confluent/ksql/test/tools/TestExecutor.java
```diff
@@ -132,6 +136,23 @@ public static TestExecutor create(
     );
   }
 
+  // These internal topics names do not need to be verified by test cases
+  private static final Set<Pattern> SKIP_INTERNAL_TOPIC_NAMES_PATTERNS = ImmutableSet.of(
```
vcrfxia, line 140: I don't think this belongs in `TestExecutor`. It should be up to the `TopicInfoCache` to say that topic info will not be cached for certain topics. How about we move this into `TopicInfoCache` and update `TopicInfoCache.get()` to return `Optional<TopicInfo>` where an empty optional will be returned if there's no info?

vcrfxia (commented):
Thanks @spena . `TopicInfoCache` is also used to collect topic info such as key and value format into historic QTT spec files to ensure backwards compatibility, right? (Here's an example of what I'm referring to: https://github.com/confluentinc/ksql/blob/master/ksqldb-functional-tests/src/test/resources/historical_plans/joins_-_non-KAFKA_key_format/7.0.0_1618294663143/spec.json#L146) We should have a long-term plan to include FK join internal topics as part of this backwards compatibility validation. Did you get a chance to think about this during your investigation?

ksqldb-functional-tests/src/main/java/io/confluent/ksql/test/tools/TestExecutor.java
spena: The cache is not the problem, but when getting the schema/serde info of the topic. Caching is secondary. I'm not sure about returning Optional for internal topics that we don't want. The `TopicInfoCache.get()` contract is to return info about a topic. If there is no info about a topic, the `get()` already throws an exception:
```
throw new TestFrameworkException("No information found for topic '"
            + topicName + "'. Available topics: " + cache.asMap().keySet());
```

So we'll have now an `Optional.empty()` and an exception that mean the same thing. If I return an Optional, then the caller needs to check if it is optional, then throw an exception. 

I agree the list of internal topics don't look that they fit in the TestExecutor, but the TestExecutor is the one that has a list of topics, so IMHO we should filter them out there.

ksqldb-functional-tests/src/main/java/io/confluent/ksql/test/tools/TopicInfoCache.java
```diff
@@ -101,8 +109,22 @@ public TopicInfoCache(
         .build(CacheLoader.from(this::load));
   }
 
-  public TopicInfo get(final String topicName) {
-    return cache.getUnchecked(topicName);
+  private boolean ignoreInternalTopic(final String topicName) {
```
vcrfxia, line 112: nit: move below public methods in file.

ksqldb-functional-tests/src/main/java/io/confluent/ksql/test/tools/TopicInfoCache.java
```diff
@@ -87,6 +88,13 @@
       )
   );
 
+  // Internal topic names that are ignored by the TopicInfoCache::get() method
+  private static final Set<Pattern> IGNORE_INTERNAL_TOPIC_NAMES_PATTERNS = ImmutableSet.of(
+      Pattern.compile(TOPIC_PATTERN_PREFIX + ".*-FK-JOIN-SUBSCRIPTION-REGISTRATION-\\d+-topic"),
```
vcrfxia, line 93: Can we add a quick comment into the code explaining why we're ignoring these internal topics for now, and that it's only temporary (until we have a long-term fix)? Ideally we'd file the information in a Github issue and simply put the link to the Github issue in the comment. Whichever you prefer.

vcrfxia (approved):
Thanks @spena -- LGTM as long as we ensure that we're tracking the follow-up for removing this hack and replacing it with the long-term solution we discussed instead. Thanks!

ksqldb-functional-tests/src/main/java/io/confluent/ksql/test/tools/TopicInfoCache.java
spena: Done.

ksqldb-functional-tests/src/main/java/io/confluent/ksql/test/tools/TopicInfoCache.java
spena: Done

mjsax (approved):

Issue #5852, opened 2020-07-21
ksqlDB silently ignores KEY qualifier in CREATE STREAM if enclosed in parentheses


mikebin:
**Describe the bug**
ksqlDB silently ignores values in parentheses after field declarations in `CREATE STREAM` (CS) and `CREATE TABLE` (CT) statements, including `KEY`

**To Reproduce**
Version: 0.10.1

```sql
create stream keyed (mykey varchar (key), f1 varchar) with (kafka_topic='keyed', value_format='json', partitions=1);
```

Stream is successfully created, but no key field (`key` was silently ignored because it was enclosed in parentheses):

```
 Field | Type
-------------------------
 MYKEY | VARCHAR(STRING)
 F1    | VARCHAR(STRING)
-------------------------
``` 

Without the parentheses, the key field is defined as expected:

```sql
create stream keyed (mykey varchar key, f1 varchar) with (kafka_topic='keyed', value_format='json', partitions=1);
 Field | Type
--------------------------------
 MYKEY | VARCHAR(STRING)  (key)
 F1    | VARCHAR(STRING)
--------------------------------
```

**Expected behavior**
Either an error message should be presented if the syntax is invalid, or the key should be applied even if the keyword `KEY` is enclosed in parentheses.

**Additional context**
ksqlDB seems to silently ignore _any_ identifier in parentheses after the field declaration.

This came up because of how the docs show `KEY` here: https://docs.ksqldb.io/en/latest/developer-guide/ksqldb-reference/create-stream/#synopsis. Submitted a separate docs ticket to make the syntax more clear.

fjbecerra:
can I have this ticket assigned, please?

agavra:
done! thanks @fjbecerra

PR #7582, opened 2021-05-25
fix: silently ignores KEY qualifier if enclosed in parentheses (#5852)

fjbecerra:
### Description 
There’s an ambiguity in the grammar declared in `SqlBase.g4`. This is happening at **tableElement** only for primitive types.
Ex.  `BOB STRING (KEY)`, evaluates **STRING(KEY)** as a type parameter. 
Screenshot 2021-05-25 at 07 34 41

One possible solution is support `(KEY)` by adding at **KEY** token another option as `'(KEY)'`, but this causes another ambiguity, since fields named as **KEY** are supported, causing ambiguity when using a function.
ex. `AS_VALUE(KEY)`.

My proposed solution validates the schema in code at the parsing, throwing an exception if any primitive type has been evaluated as type parameter.

### Testing done 
ksqldb-parser - AstBuilderTest
ksqldb-functional-test - key-schema.json

### Reviewer checklist

### DIFF
--- a/ksqldb-parser/src/main/antlr4/io/confluent/ksql/parser/SqlBase.g4
+++ b/ksqldb-parser/src/main/antlr4/io/confluent/ksql/parser/SqlBase.g4
@@ -333,7 +333,7 @@ type
     ;
 
 typeParameter
-    : INTEGER_VALUE | type
+    : INTEGER_VALUE | 'STRING'
     ;
 
 baseType
@@ -387,7 +387,7 @@ literal
 nonReserved
     : SHOW | TABLES | COLUMNS | COLUMN | PARTITIONS | FUNCTIONS | FUNCTION | SESSION
     | STRUCT | MAP | ARRAY | PARTITION
-    | INTEGER | DATE | TIME | TIMESTAMP | INTERVAL | ZONE
+    | INTEGER | DATE | TIME | TIMESTAMP | INTERVAL | ZONE | 'STRING'
     | YEAR | MONTH | DAY | HOUR | MINUTE | SECOND
     | EXPLAIN | ANALYZE | TYPE | TYPES
     | SET | RESET
--- a/ksqldb-functional-tests/src/test/resources/historical_plans/key-schemas_-_VARCHAR(STRING)_as_key_and_value/7.1.0_1627023646891/plan.json
+++ b/ksqldb-functional-tests/src/test/resources/historical_plans/key-schemas_-_VARCHAR(STRING)_as_key_and_value/7.1.0_1627023646891/plan.json
@@ -1,0 +1,154 @@
+{
+  "plan" : [ {
+    "@type" : "ksqlPlanV1",
+    "statementText" : "CREATE STREAM INPUT (K STRING KEY, V1 STRING) WITH (KAFKA_TOPIC='input', KEY_FORMAT='KAFKA', VALUE_FORMAT='JSON');",
+    "ddlCommand" : {
+      "@type" : "createStreamV1",
+      "sourceName" : "INPUT",
+      "schema" : "`K` STRING KEY, `V1` STRING",
+      "topicName" : "input",
+      "formats" : {
+        "keyFormat" : {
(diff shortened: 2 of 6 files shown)


spena:
Thanks @fjbecerra for the catch. I was looking into the `SqlBase.g4` syntax:
```
type
    : type ARRAY
    | ARRAY '<' type '>'
    | MAP '<' type ',' type '>'
    | STRUCT '<' (identifier type (',' identifier type)*)? '>'
    | DECIMAL '(' number ',' number ')'
    | baseType ('(' typeParameter (',' typeParameter)* ')')?
    ;

typeParameter
    : INTEGER_VALUE | type
    ;
```

Seems `typeParameter` is only used for the `baseType (...)` syntax. By the time that line is checked by antlr, then we are for sure not using non-primitive types, so `baseType` is used only for primitives. I think the only primitive type that accepts parenthesis is `VARCHAR(STRING)` (see `CliTest.java`). Do you think we can use `INTEGER_VALUE | 'STRING'` in the `typeParameter` instead `type` and instead of checking for invalid schema?

fjbecerra:
Thanks @spena.
Yes, that worked! I updated the PR, LMK what you think

ksqldb-parser/src/main/antlr4/io/confluent/ksql/parser/SqlBase.g4
```diff
@@ -333,7 +333,7 @@ type
     ;
 
 typeParameter
-    : INTEGER_VALUE | type
+    : INTEGER_VALUE | STRING
```
spena, line 336: I don't think the `STRING` token works here. See below:
```
STRING
    : '\'' ( ~'\'' | '\'\'' )* '\''
    ;
```

The `STRING` token will translate to any string that starts/ends with single quotes. So the `VARCHAR(STRING)` will fail. Could you add a test that verifies `VARCHAR(STRING)` works? I'm pretty sure that will fail, so you'll have to update the syntax.

fjbecerra: Yup, You’re right... I misread the first comment. 
It works with `'STRING'`, however breaks the syntax  with `STRING`.
Screenshot 2021-06-10 at 08 11 16

To solve it, `'STRING'`  has also to be added in `nonReserved`.

fjbecerra: @spena take it a look when you got a chance, thanks!

spena (approved):
LGTM

spena:
@fjbecerra Let's wait for tests to pass. Ping me whenever they are green so I can merge the PR.

spena:
@fjbecerra Could you rebase your branch? I retriggered Jenkins but the job still fails with some branch indexing errors.

fjbecerra:
@spena , rebased, thanks

spena:
@fjbecerra Sorry to ping you again. Could you rebase it again? Last errors were caused by some issues in master that were in your last rebase.

fjbecerra:
hey @spena , dw, rebased

spena:
@fjbecerra Sorry to ask you again, really. There was too much noise in the past weeks about versioning and other issues that caused your branch to fail. But I checked and master is green again. Could you rebase your branch?

fjbecerra:
😄  dw! rebased

spena:
One test is failing. Not sure if this is affecting compatibility?

fjbecerra:
it was complaining on QTT plans, [here](https://jenkins.confluent.io/job/Confluentinc%20Contributors/job/ksql/view/change-requests/job/PR-7582/6/testReport/io.confluent.ksql.test.planned/PlannedTestsUpToDateTest/shouldHaveLatestPlans_key_schemas___VARCHAR_STRING__as_key_and_value_/), by re-generating it manually the tests passed

spena:
Thanks @fjbecerra for the contribution. Sorry it took too much time to getting merged.

Issue #7584, opened 2021-05-25
Qualified select * does not work in n-way joins with 3 or more joins


vcrfxia:
**Describe the bug**

The following n-way join statements
```
      "statements": [
        "CREATE TABLE T1 (ID INT PRIMARY KEY, V0 bigint) WITH (kafka_topic='left', value_format='JSON');",
        "CREATE TABLE T2 (ID INT PRIMARY KEY, V0 bigint) WITH (kafka_topic='right', value_format='JSON');",
        "CREATE TABLE T3 (ID INT PRIMARY KEY, V0 bigint) WITH (kafka_topic='right2', value_format='JSON');",
        "CREATE TABLE T4 (ID INT PRIMARY KEY, V0 bigint) WITH (kafka_topic='right3', value_format='JSON');",
        "CREATE TABLE OUTPUT as SELECT T1.ID, T2.*, T3.V0, T4.* FROM T1 JOIN T2 ON T1.ID = T2.ID JOIN T3 ON T1.ID = T3.ID JOIN T4 ON T1.ID = T4.ID;"
      ]
```
should result in the `OUTPUT` table having schema
```
T1_ID INT PRIMARY KEY, T2_ID INTEGER, T2_V0 BIGINT, T3_V0 BIGINT, T4_ID INTEGER, T4_V0 BIGINT
```
but actually the columns `T2_ID INTEGER, T2_V0 BIGINT,` are missing, which is incorrect.

**To Reproduce**

This bug is present on the current ksqlDB master, and appears to have been present for at least a year, probably longer.

This QTT should pass but doesn't:
```
    {
      "name": "table-table-table-table - qualified select *",
      "statements": [
        "CREATE TABLE T1 (ID INT PRIMARY KEY, V0 bigint) WITH (kafka_topic='left', value_format='JSON');",
        "CREATE TABLE T2 (ID INT PRIMARY KEY, V0 bigint) WITH (kafka_topic='right', value_format='JSON');",
        "CREATE TABLE T3 (ID INT PRIMARY KEY, V0 bigint) WITH (kafka_topic='right2', value_format='JSON');",
        "CREATE TABLE T4 (ID INT PRIMARY KEY, V0 bigint) WITH (kafka_topic='right3', value_format='JSON');",
[...]

Issue #7583, opened 2021-05-25
Qualified select * does not work in n-way joins requiring intermediate repartitions


vcrfxia:
**Describe the bug**

The following n-way join statements
```
      "statements": [
        "CREATE STREAM left_stream (id1 BIGINT KEY, f1 BIGINT) WITH (kafka_topic='left_topic', format='JSON');",
        "CREATE TABLE middle_table (id2 BIGINT PRIMARY KEY, f2 BIGINT, other STRING) WITH (kafka_topic='middle_topic', format='JSON');",
        "CREATE TABLE right_table (id3 BIGINT PRIMARY KEY, f3 BIGINT) WITH (kafka_topic='right_topic', format='JSON');",
        "CREATE STREAM output AS SELECT id1, f1, middle_table.*, right_table.* FROM left_stream JOIN middle_table ON id1 = id2 LEFT JOIN right_table ON f1 = id3;"
      ]
```
should result in the `OUTPUT` stream having schema 
```
F1 BIGINT KEY, ID1 BIGINT, MIDDLE_TABLE_ID2 BIGINT, MIDDLE_TABLE_F2 BIGINT, MIDDLE_TABLE_OTHER STRING, RIGHT_TABLE_ID3 BIGINT, RIGHT_TABLE_F3 BIGINT
```
but actually the columns `MIDDLE_TABLE_ID2 BIGINT, MIDDLE_TABLE_F2 BIGINT, MIDDLE_TABLE_OTHER STRING` are missing, which is incorrect.

**To Reproduce**

This bug is present on the current ksqlDB master, and appears to have been present for at least a year, probably longer.

This QTT should pass but doesn't:
```
    {
      "name": "stream-table-table - qualified select *",
      "statements": [
        "CREATE STREAM left_stream (id1 BIGINT KEY, f1 BIGINT) WITH (kafka_topic='left_topic', format='JSON');",
        "CREATE TABLE middle_table (id2 BIGINT PRIMARY KEY, f2 BIGINT, other STRING) WITH (kafka_topic='middle_topic', format='JSON');",
        "CREATE TABLE right_table (id3 BIGINT PRIMARY KEY, f3 BIGINT) WITH (kafka_topic='right_topic', format='JSON');",
        "CREATE STREAM output AS SELECT id1, f1, middle_table.*, right_table.* FROM left_stream JOIN middle_table ON id1 = id2 LEFT JOIN right_table ON f1 = id3;"
      ],
      "inputs": [
[...]

PR #7585, opened 2021-05-25
fix: qualified select * now works in n-way joins with repartitions and multiple layers of nesting

vcrfxia:
### Description 

Fixes https://github.com/confluentinc/ksql/issues/7583.
Fixes https://github.com/confluentinc/ksql/issues/7584.

The reason for both bugs is because the current logic in `JoinNode#resolveSelectStar()` for finding data source names says "if the current node is a JoinNode, inspect its children, else use the current node". This is incorrect because

1. The current node might be neither a JoinNode nor a node corresponding to a single data source -- it might be a PreJoinRepartitionNode attached to a JoinNode. In this case, the current logic says to use the current node but the current node is not associated with any data source. As a result, some data sources are missing from the set of data source names considered when resolving the qualified select `*`.
2. Even if the current node is a JoinNode, its children might not correspond to specific data sources. This is the case if one or more children are themselves JoinNodes, i.e., the n-way join involves 3 or more joins. In this case, the current logic says to inspect the children of the JoinNode but not all children are associated with specific data sources, and again some data sources are missing from the set of data source names considered when resolving the qualified select `*`.

This PR fixes this class of bugs by updating the logic to properly find all source nodes for the n-way join.

This PR also includes minor fixes to one of the new foreign key join tests (currently disabled), the one that exposed this class of long-standing bugs.

### Testing done 

Added two QTTs that fail before the bug fix in this PR but pass with the fix.

### Reviewer checklist

### DIFF
--- a/ksqldb-engine/src/main/java/io/confluent/ksql/planner/plan/JoinNode.java
+++ b/ksqldb-engine/src/main/java/io/confluent/ksql/planner/plan/JoinNode.java
@@ -199,7 +199,7 @@ public class JoinNode extends PlanNode implements JoiningNode {
       final Optional<SourceName> sourceName
   ) {
     final Stream<ColumnName> names = Stream.of(left, right)
-        .flatMap(s -> s instanceof JoinNode ? s.getSources().stream() : Stream.of(s))
+        .flatMap(JoinNode::getPreJoinProjectDataSources)
         .filter(s -> !sourceName.isPresent() || sourceName.equals(s.getSourceName()))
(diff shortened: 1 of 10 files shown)


spena (approved):
LGTM. Code looks simple.

mjsax (approved):

guozhangwang (commented):
Slowly catching up all the PRs related to n-way joins here. Only reviewed the non-testing part (learned something new :)

PR #7588, opened 2021-05-26
test: add test for FK-join with format mismatch

mjsax:
### Description 
We should add a test that verifies that a format mismatch of left and right input does not trigger a repartitioning for FK-joins.

### Testing done 
Added QTT test.

### Reviewer checklist

### DIFF
--- a/ksqldb-functional-tests/src/test/resources/query-validation-tests/fk-join.json
+++ b/ksqldb-functional-tests/src/test/resources/query-validation-tests/fk-join.json
@@ -684,6 +684,60 @@
       ],
       "post": {
         "topics": {
+          "topics": [
+            {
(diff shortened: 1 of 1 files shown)


ksqldb-functional-tests/src/test/resources/query-validation-tests/fk-join.json
```diff
@@ -684,6 +684,60 @@
       ],
       "post": {
         "topics": {
+          "topics": [
```
mjsax, line 687: Just updating the post condition for this test in addition.

vcrfxia (approved):
Thanks @mjsax -- LGTM (assuming the tests pass ;) )

PR #7591, opened 2021-05-26
feat: ungate support for foreign key joins

vcrfxia:
### Description 

This PR removes the feature flag currently guarding foreign key joins. All but four tests have been enabled -- these tests are failing due to a bug in key column qualifiers (being fixed in a separate PR). A number of tests have had their expected output cleaned up as well.

For reviewers:
- Review the test fixes in the commit titled `test: fix expected outputs in tests` to confirm they are as expected
- There is a minor change in `chore: do not ignore fk join result changelog in qtt` to explicitly pass the key serde in addition to the value serde. Even though this is not strictly necessary, it's good practice and also expected by the QTT framework.
- The disabled tests in `chore: disable failing tests` are the ones that are failing due to a bug in key column qualifiers, as mentioned above.
- The actual code changes for removing the feature flag are in `feat: ungate support for foreign key joins` if you're curious, though this was all pretty mechanical.

### Testing done 

Build passes.

### Reviewer checklist


ksqldb-functional-tests/src/main/java/io/confluent/ksql/test/tools/TopicInfoCache.java
```diff
@@ -96,7 +96,8 @@
   private static final Set<Pattern> IGNORE_INTERNAL_TOPIC_NAMES_PATTERNS = ImmutableSet.of(
       Pattern.compile(TOPIC_PATTERN_PREFIX + ".*-FK-JOIN-SUBSCRIPTION-REGISTRATION-\\d+-topic"),
       Pattern.compile(TOPIC_PATTERN_PREFIX + ".*-FK-JOIN-SUBSCRIPTION-RESPONSE-\\d+-topic"),
-      Pattern.compile(TOPIC_PATTERN_PREFIX + ".*-FK-JOIN-SUBSCRIPTION-STATE-STORE-\\d+-changelog")
+      Pattern.compile(TOPIC_PATTERN_PREFIX + ".*-FK-JOIN-SUBSCRIPTION-STATE-STORE-\\d+-changelog"),
+      Pattern.compile(TOPIC_PATTERN_PREFIX + ".*-FK-JOIN-OUTPUT-STATE-STORE-\\d+-changelog")
```
mjsax, line 100: I don't think we should exclude this topic -- it's not a "internal" topic, but the changelog of the FK-join result if we materialize it.

vcrfxia: Fixed in the commit titled "chore: do not ignore fk join result changelog in qtt"

mjsax (approved):

PR #7592, opened 2021-05-26
chore: fix alias resolution in n-way join logic

mjsax:
### Description 
We need to use the alias name (instead of the table name) if an alias exist to verify if the key column is present in the projection.

### Testing done 
Updated existing QTT tests.

### Reviewer checklist

### DIFF
--- a/ksqldb-engine/src/main/java/io/confluent/ksql/planner/plan/JoinNode.java
+++ b/ksqldb-engine/src/main/java/io/confluent/ksql/planner/plan/JoinNode.java
@@ -217,9 +217,9 @@ public class JoinNode extends PlanNode implements JoiningNode {
   void validateKeyPresent(final SourceName sinkName, final Projection projection) {
 
     if (joinKey.isForeignKey()) {
-      final DataSource leftInputTable = getLeftmostSourceNode().getDataSource();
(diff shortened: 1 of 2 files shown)


ksqldb-engine/src/main/java/io/confluent/ksql/planner/plan/JoinNode.java
```diff
@@ -217,9 +217,9 @@ protected int getPartitions(final KafkaTopicClient kafkaTopicClient) {
   void validateKeyPresent(final SourceName sinkName, final Projection projection) {
 
     if (joinKey.isForeignKey()) {
-      final DataSource leftInputTable = getLeftmostSourceNode().getDataSource();
-      final SourceName leftInputTableName = leftInputTable.getName();
-      final List<Column> leftInputTableKeys = leftInputTable.getSchema().key();
+      final DataSourceNode leftInputTable = getLeftmostSourceNode();
+      final SourceName leftInputTableName = leftInputTable.getAlias();
```
mjsax, line 221: This is the actual fix: `getAlias` return the alias is there is one; otherwise the table name.

vcrfxia (approved):
LGTM -- thanks @mjsax .

guozhangwang (commented):
LGTM.

PR #7600, opened 2021-05-27
fix: update confusing names

wcarlson5:
rename the QueryMetadata classes to QueryEntity

### Reviewer checklist

### DIFF
--- a/design-proposals/klip-12-pull-high-availability.md
+++ b/design-proposals/klip-12-pull-high-availability.md
@@ -209,7 +209,7 @@ Map<KsqlNode, Boolean> aliveNodes;
 Map<KsqlServer, Map<String, Map<Integer, LagInfo>>> globalPositionsMap;
  
 // Fetch the metadata related to the key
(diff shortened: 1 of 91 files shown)


ableegoldman:
I'm +10000 on renaming "QueryMetadata", but "Entity" feels a bit vague...not sure I have a better suggestion though (and I'd still consider it an improvement). Just wondering if you had considered any other names?

wcarlson5:
@ableegoldman My first thought was to just drop the Metadata part but that was pretty weird. Open to suggestions tho!

rodesai:
How about:
`QueryMetadata` -> `ContinuousQuery`
`PersistentQueryMetadata` -> `PersistentQuery`
`TransientQueryMetadata` -> `TransientQuery`

guozhangwang:
I like @rodesai 's suggestion above --- without any history, I'd actually suggest `QueryMetadata -> PushQuery` as contrast to `PullQuery`, but since we've used that term before and is kinda deprecating that, it could be confusing to reuse :P I think `ContinuousQuery` is the next in line.

PR #7606, opened 2021-05-27
test: additional stream-stream-stream join tests (MINOR)

vcrfxia:
### Description 

Currently the only stream-stream-stream join test coverage we have is for inner joins. This PR adds two new tests: left-left and outer-outer. I considered adding left-outer and outer-left as well but I think the additional benefit beyond what's in the PR is minimal.

### Testing done 

Test-only change.

### Reviewer checklist

### DIFF
--- a/ksqldb-functional-tests/src/test/resources/query-validation-tests/multi-joins.json
+++ b/ksqldb-functional-tests/src/test/resources/query-validation-tests/multi-joins.json
@@ -852,10 +852,10 @@
     {
       "name": "stream-stream-stream - left-left",
       "statements": [
-        "CREATE STREAM S1 (ID INT KEY, V0 bigint) WITH (kafka_topic='left', value_format='JSON');",
-        "CREATE STREAM S2 (ID INT KEY, V0 bigint) WITH (kafka_topic='right', value_format='JSON');",
-        "CREATE STREAM S3 (ID INT KEY, V0 bigint) WITH (kafka_topic='right2', value_format='JSON');",
-        "CREATE STREAM OUTPUT as SELECT S2.ID, s1.V0, s2.V0, s3.V0 FROM S1 LEFT JOIN S2 WITHIN 5 seconds ON S1.ID = S2.ID LEFT JOIN S3 WITHIN 10 seconds ON S1.ID = S3.ID;"
+        "CREATE STREAM s1 (id INT KEY, v0 BIGINT) WITH (kafka_topic='left', value_format='JSON');",
+        "CREATE STREAM s2 (id INT KEY, v0 BIGINT) WITH (kafka_topic='right', value_format='JSON');",
+        "CREATE STREAM s3 (id INT KEY, v0 BIGINT) WITH (kafka_topic='right2', value_format='JSON');",
+        "CREATE STREAM output AS SELECT s2.id, s1.v0, s2.v0, s3.v0 FROM s1 LEFT JOIN s2 WITHIN 5 SECONDS ON s1.id = s2.id LEFT JOIN S3 WITHIN 10 SECONDS ON s1.id = s3.id;"
       ],
       "inputs": [
         {"topic": "left", "key": 0, "value": {"V0": 1}, "timestamp": 0},
@@ -880,10 +880,10 @@
     {
       "name": "stream-stream-stream - inner-inner - different windows",
       "statements": [
-        "CREATE STREAM S1 (ID INT KEY, V0 bigint) WITH (kafka_topic='left', value_format='JSON');",
-        "CREATE STREAM S2 (ID INT KEY, V0 bigint) WITH (kafka_topic='right', value_format='JSON');",
-        "CREATE STREAM S3 (ID INT KEY, V0 bigint) WITH (kafka_topic='right2', value_format='JSON');",
-        "CREATE STREAM OUTPUT as SELECT S2.ID, s1.V0, s2.V0, s3.V0 FROM S1 JOIN S2 WITHIN 5 seconds ON S1.ID = S2.ID JOIN S3 WITHIN 10 seconds ON S1.ID = S3.ID;"
+        "CREATE STREAM s1 (id INT KEY, v0 BIGINT) WITH (kafka_topic='left', value_format='JSON');",
+        "CREATE STREAM s2 (id INT KEY, v0 BIGINT) WITH (kafka_topic='right', value_format='JSON');",
+        "CREATE STREAM s3 (id INT KEY, v0 BIGINT) WITH (kafka_topic='right2', value_format='JSON');",
+        "CREATE STREAM output AS SELECT s2.id, s1.v0, s2.v0, s3.v0 FROM S1 JOIN S2 WITHIN 5 SECONDS ON s1.id = s2.id JOIN S3 WITHIN 10 SECONDS ON s1.id = s3.id;"
       ],
       "inputs": [
         {"topic": "left", "key": 0, "value": {"V0": 1}, "timestamp": 0},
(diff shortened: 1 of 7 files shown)


ksqldb-functional-tests/src/test/resources/query-validation-tests/multi-joins.json
```diff
@@ -813,6 +813,70 @@
...
+      "inputs": [
+        {"topic": "left", "key": 0, "value": {"V0": 1}, "timestamp": 0},
+        {"topic": "right", "key": 0, "value": {"V0": 2}, "timestamp": 1},
+        {"topic": "right2", "key": 0, "value": {"V0": 3}, "timestamp": 2},
+        {"topic": "left", "key": 0, "value": {"V0": 4}, "timestamp": 3}
+      ],
+      "outputs": [
+        {"topic": "OUTPUT", "key": 0, "value": {"S1_V0": 1, "S2_V0": null, "S3_V0": null}, "timestamp": 0},
+        {"topic": "OUTPUT", "key": 0, "value": {"S1_V0": 1, "S2_V0": 2, "S3_V0": null}, "timestamp": 1},
+        {"topic": "OUTPUT", "key": 0, "value": {"S1_V0": 1, "S2_V0": null, "S3_V0": 3}, "timestamp": 2},
```
vcrfxia, line 869: Hey @mjsax @guozhangwang is this output record expected?

ksqldb-functional-tests/src/test/resources/query-validation-tests/multi-joins.json
```diff
@@ -813,6 +813,70 @@
...
```
vcrfxia, line 833: Same question as above (below).

guozhangwang: Right now with 1) @spena 's fix and 2) https://confluentinc.atlassian.net/browse/KSE-326?page=com.atlassian.jira.plugin.system.issuetabpanels%3Aall-tabpanel, I think this is reasonable: after we have piped `2` from `right`, on the second sub-topology we would have buffered two records: `(1, null)`, and `(1, 2)` so when `3` from `right2` is received, two outputs, i.e. line 3/4 would be emitted.

In practice though, things could be more tricky than the test driver, since there's a repartition topic between `(left join right)` and `right2`, when a record from `right2` is received, the previous join result `(1, null)` and `(1, 2)` may or may not have piped from the repartition topic, so we may also see a `(null, null, 3)` be emitted first, also maybe we would not emit `(1, 2, null)` if the intermediate record at the repartition topic was fetched a bit late: in that case the output ordering would be `(1, null, null)` then `(1, null, 3)` and then `(1, 2, 3)`, etc.

Once both 1) and 2) above are fixed, we should only see two outputs, which is `(1,2,3)` and `(4,2,3)`.

ksqldb-functional-tests/src/test/resources/query-validation-tests/multi-joins.json
```diff
@@ -813,6 +813,70 @@
         ]
       }
     },
+    {
+      "name": "stream-stream-stream - outer-outer",
+      "statements": [
+        "CREATE STREAM S1 (ID INT KEY, V0 bigint) WITH (kafka_topic='left', value_format='JSON');",
```
mjsax, line 819: Why are names (`ID`, `V0`) is upper case, while keywords (`BIGINT`) are in lower case?

ksqldb-functional-tests/src/test/resources/query-validation-tests/multi-joins.json
```diff
@@ -813,6 +813,70 @@
...
+        "CREATE STREAM S2 (ID INT KEY, V0 bigint) WITH (kafka_topic='right', value_format='JSON');",
+        "CREATE STREAM S3 (ID INT KEY, V0 bigint) WITH (kafka_topic='right2', value_format='JSON');",
+        "CREATE STREAM OUTPUT as SELECT ROWKEY, s1.V0, s2.V0, s3.V0 FROM S1 FULL OUTER JOIN S2 WITHIN 5 seconds ON S1.ID = S2.ID FULL OUTER JOIN S3 WITHIN 10 seconds ON S1.ID = S3.ID;"
```
mjsax, line 822: Names should be all lower case, eg, `s1.v0` and `s1.id = s1.id` an `s1 FULL OUTER JOIN s2`

ksqldb-functional-tests/src/test/resources/query-validation-tests/multi-joins.json
```diff
@@ -813,6 +813,70 @@
         ]
       }
     },
+    {
+      "name": "stream-stream-stream - outer-outer",
+      "statements": [
+        "CREATE STREAM S1 (ID INT KEY, V0 bigint) WITH (kafka_topic='left', value_format='JSON');",
+        "CREATE STREAM S2 (ID INT KEY, V0 bigint) WITH (kafka_topic='right', value_format='JSON');",
+        "CREATE STREAM S3 (ID INT KEY, V0 bigint) WITH (kafka_topic='right2', value_format='JSON');",
```
mjsax, line 821: `right` and `right2` is confusing... Can we rename the topics to `left_topic`, `middle_topic` and `right_topic` (make reading the `"inputs"` section below easier.

ksqldb-functional-tests/src/test/resources/query-validation-tests/multi-joins.json
mjsax: > since there's a repartition topic between

Not necessarily. Also, this test should not use a repartition as we join on the input key on all 3 streams. But you are of course right, that in the most general case, there _could_ be a repartition topic.

ksqldb-functional-tests/src/test/resources/query-validation-tests/multi-joins.json
vcrfxia: Pretty much all the tests in this file are copy-pasted from each other, and they're all like this. I've updated the casing for the new tests, but updating the casing for all the existing teams is out of scope IMO.

ksqldb-functional-tests/src/test/resources/query-validation-tests/multi-joins.json
vcrfxia: See above -- copy-and-paste from basically all the tests in this file. But I've updated the new ones to your liking.

ksqldb-functional-tests/src/test/resources/query-validation-tests/multi-joins.json
vcrfxia: This follows the pattern of all the other tests in this file. For this comment I actually think consistency is more important for readability than the rename so I've left it as is. If we want to go back through and update all the tests in this file we can do that in one big batch down the road.

ksqldb-functional-tests/src/test/resources/query-validation-tests/multi-joins.json
vcrfxia: Thanks, both. I have filed https://github.com/confluentinc/ksql/issues/7623 to track the unnecessary repartition. (This is the same as KSE-326 mentioned above, for anyone following this thread.)

ksqldb-functional-tests/src/test/resources/query-validation-tests/multi-joins.json
mjsax: Absolutely. I did not ask to rewrite existing tests.

mjsax (approved):

vcrfxia:
Not sure why Jenkins status is showing as building -- the PR build from a few days ago is green and the current build link isn't loading. Going to merge.

PR #7614, opened 2021-05-31
test: fix shouldNotBeAbleToUseWssIfClientDoesNotTrustServerCert test

jzaralim:
### Description 
From https://kafka.apache.org/21/documentation.html#upgrade_200_notable: 

The default value for ssl.endpoint.identification.algorithm was changed to https, which performs hostname verification (man-in-the-middle attacks are possible otherwise). Set ssl.endpoint.identification.algorithm to an empty string to restore the previous behaviour.

### Reviewer checklist


vcrfxia (approved):
Thanks @jzaralim . LGTM except I'm confused why this change is only needed now. The release notes linked in the PR description are for 2.0.0, which means this change has been present ever since 5.2.x was created?

Also I updated the PR title from "fix" to "test" since this is a test-only change (and does not need to appear in the auto-generated changelog as a result).

andrewegel (approved):

Issue #7429, opened 2021-04-23
Rationalize use of Avro decimal precision vs. connect.decimal.precision


colinhicks:
ksqlDB currently requires that the schema corresponding to Avro decimal fields includes the `connect.decimal.precision` property. The precision-parsing logic throws an error when it's not provided:
https://github.com/confluentinc/ksql/blob/d15026fe7c8993673152d740d46b16253896d8e8/ksql-common/src/main/java/io/confluent/ksql/util/DecimalUtil.java#L102-L105

Whether this property is included in Decimal-carrying schema is inconsistent:
* When ksqlDB builds the relevant schema, it's [included](https://github.com/confluentinc/ksql/blob/d15026fe7c8993673152d740d46b16253896d8e8/ksql-common/src/main/java/io/confluent/ksql/util/DecimalUtil.java#L69-L74)
* With some connectors it is included when the schema is built, e.g. [Debezium](https://github.com/debezium/debezium/blob/ef0d018c79e5aa25077eadf4ebe35f2dc4bb37cc/debezium-core/src/main/java/io/debezium/data/SpecialValueDecimal.java#L160-L162)
* With some connectors it is not included, e.g. JDBCConnector

#### Motivating use case
Currently, ksqlDB throws an exception upon `CREATE {STREAM,TABLE}` when schema inference is used against a schema created by connectors that do not write the `connect.decimal.precision` property:
```
Unable to verify if the value schema for topic <name> is compatible with ksqlDB.
Reason: Invalid Decimal schema: precision parameter not found.
```

One option is to consider Schema Registry's handling as canonical.
* When converting a Connect schema to Avro, Schema Registry [uses the `connect.decimal.precision` property if it's provided, if not, it uses the default value, `64`](https://github.com/confluentinc/schema-registry/blob/c19da74f7c4326438dd96d022a7e3d38dd3c68af/avro-data/src/main/java/io/confluent/connect/avro/AvroData.java#L902-L908).
* When converting Avro to a Connect schema Schema Registry [includes `connect.decimal.precision` if a non-default precision is set](https://github.com/confluentinc/schema-registry/blob/c19da74f7c4326438dd96d022a7e3d38dd3c68af/avro-data/src/main/java/io/confluent/connect/avro/AvroData.java#L1548-L1550).

#### Suggested change
[...]

PR #7615, opened 2021-06-01
feat: use Connect default precision for Avro decimals if unspecified (MINOR)

vcrfxia:
### Description 

Fixes https://github.com/confluentinc/ksql/issues/7429.

This PR implements the suggested change in https://github.com/confluentinc/ksql/issues/7429 to have ksqlDB mirror the Schema Registry behavior of defaulting precision for Avro decimals to 64 if not explicitly specified in the schema. This PR does _not_ change how ksqlDB generates output schemas -- the precision is still always written into the schema explicitly, even if the default of 64 was used. cc @colinhicks on whether we also want to make this additional change or not (considering potential implications for backwards incompatibility).

### Testing done 

Added unit + integration tests.

### Reviewer checklist

### DIFF
--- a/ksqldb-common/src/main/java/io/confluent/ksql/util/DecimalUtil.java
+++ b/ksqldb-common/src/main/java/io/confluent/ksql/util/DecimalUtil.java
@@ -32,6 +32,7 @@ import org.apache.kafka.connect.data.SchemaBuilder;
 public final class DecimalUtil {
 
   public static final String PRECISION_FIELD = "connect.decimal.precision";
+  public static final int PRECISION_DEFAULT = 64;
 
   private DecimalUtil() {
   }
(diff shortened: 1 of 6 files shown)


colinhicks (approved):
Thanks! This SGTM:

> the precision is still always written into the schema explicitly, even if the default of 64 was used.

spena (approved):
LGTM. Just one quick question. If the schema is explicitly defined, then the precision is still required, right? I don't know if the `precision(schema)` method will allow users to use `CREATE STREAM ... (f DECIMAL)` without specifying precision and scale.

Reference about default values here - https://confluentinc.atlassian.net/wiki/spaces/KSQL/pages/820675165/One+Pager+-+KSQL+Decimal+Support#OnePager-KSQLDecimalSupport-SyntaxandCharacteristics

vcrfxia:
> If the schema is explicitly defined, then the precision is still required, right?

That's correct. This PR only affects (Avro) schema inference. If a user wishes to explicitly specify a schema, then both precision and scale must be provided as part of that schema.

PR #7617, opened 2021-06-01
fix: dont throw NPE on listing if register fails

rodesai:
couple fixes to query registration to avoid NPE on listing if register fails:
- notify listeners before registering the query. This ensures we don't register
  on failure.
- dont do initialize as separate step from query metadata construction. There's no
  reason not to just do it up front.

### Testing done 
Added a unit test. Also tested manually by instrumenting a throw from
`QueryStateMetricsReportingListener`

### DIFF
--- a/ksqldb-engine/src/main/java/io/confluent/ksql/query/QueryRegistryImpl.java
+++ b/ksqldb-engine/src/main/java/io/confluent/ksql/query/QueryRegistryImpl.java
@@ -278,8 +278,8 @@ public class QueryRegistryImpl implements QueryRegistry {
       }
     }
     allLiveQueries.add(query);
-    notifyCreate(serviceContext, metaStore, query);
     query.initialize();
+    notifyCreate(serviceContext, metaStore, query);
   }
 
   private void unregisterQuery(final QueryMetadata query) {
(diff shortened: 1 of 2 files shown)


wcarlson5 (approved):
LGTM

ksqldb-engine/src/main/java/io/confluent/ksql/query/QueryRegistryImpl.java
```diff
@@ -249,6 +249,7 @@ private void registerQuery(
       final QueryMetadata query,
       final boolean createAsQuery
   ) {
+    notifyCreate(serviceContext, metaStore, query);
```
guozhangwang, line 252: Just double checking here: do you think we would ever have any listeners that assumes the query is already in the registry map in its impl?

stevenpyzhang:
Should we target this PR against master first so that we get a PR build and then cherry-pick it to 0.18.0? I don't think one will trigger against `0.18.x-ksqldb`

rodesai:
> Should we target this PR against master first so that we get a PR build and then cherry-pick it to 0.18.0? I don't think one will trigger against 0.18.x-ksqldb

this patch isn't pick-able between master and 0.18.x due to a refactor with the query metadata stuff. Patch to master is coming.

ksqldb-engine/src/main/java/io/confluent/ksql/query/QueryRegistryImpl.java
rodesai: no, they shouldn't need to either since we pass the query in.

PR #7621, opened 2021-06-02
fix: npe on listing after notification error

rodesai:
Ensure that the query is initialized before notifying

### DIFF
--- a/ksqldb-engine/src/main/java/io/confluent/ksql/query/QueryRegistryImpl.java
+++ b/ksqldb-engine/src/main/java/io/confluent/ksql/query/QueryRegistryImpl.java
@@ -280,8 +280,8 @@ public class QueryRegistryImpl implements QueryRegistry {
       }
     }
     allLiveQueries.add(query);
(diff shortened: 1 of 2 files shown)


ksqldb-engine/src/main/java/io/confluent/ksql/query/QueryRegistryImpl.java
```diff
@@ -280,8 +280,8 @@ private void registerQuery(
       }
     }
     allLiveQueries.add(query);
-    notifyCreate(serviceContext, metaStore, query);
     query.initialize();
+    notifyCreate(serviceContext, metaStore, query);
```
lct45, line 284: Nice, this works better for the CSU metrics reporter as well

lct45 (approved):
LGTM

guozhangwang (approved):

rodesai:
closed in favor of https://github.com/confluentinc/ksql/pull/7629 (it's mostly the same, but is cherry-picked from 0.18.x and includes a null-check in the lsitener)

PR #7629, opened 2021-06-03
fix: npe on listing after notification error

rodesai:
Ensure that the query is initialized before notifying


wcarlson5 (approved):
lGTM still

guozhangwang (approved):

PR #7631, opened 2021-06-03
chore: fix test issue with SchemaRegisterInjectorTest

spena:
### Description 
The test `shouldRegisterDependenciesForProtobuf` was failing because the Protobuf schema registered didn't contain the references that the test expected. I'm unsure why the reference is not added when registering the schema in `SchemaRegisterInjector.registerSchema()`. Perhaps SR decided not to add it because it is duplicated in the protobuf schema already.

### Testing done 

### Reviewer checklist


wcarlson5 (commented):
This file hasn't changed since this test was made. Why did it start failing now?

wcarlson5 (approved):
Perfect thanks for the explanation! Lets get this merged so I stop getting emails about test failures :P

PR #7633, opened 2021-06-03
POC: Pull Queries over Streams

vvcephei:
### Description 

Not suitable for merging. Merely exploring what is required to build pull queries over streams
on top of transient queries. It seems like it'll be a moderate effort to enable this approach in
production-worthy code.

### Testing done 

I followed the [quickstart](https://ksqldb.io/quickstart.html) to create and populate a stream.

I ran the following queries and got the following (self-terminating) responses in the CLI:

```
ksql> SELECT * FROM riderLocations WHERE GEO_DISTANCE(latitude, longitude, 37.4133, -122.1162) <= 5;
+---------------------------------+---------------------------------+---------------------------------+
|PROFILEID                        |LATITUDE                         |LONGITUDE                        |
+---------------------------------+---------------------------------+---------------------------------+
|4ab5cbad                         |37.3952                          |-122.0813                        |
|8b6eae59                         |37.3944                          |-122.0813                        |
|4a7c7b41                         |37.4049                          |-122.0822                        |
Query terminated
ksql> SELECT * FROM riderLocations;

+---------------------------------+---------------------------------+---------------------------------+
|PROFILEID                        |LATITUDE                         |LONGITUDE                        |
+---------------------------------+---------------------------------+---------------------------------+
|c2309eec                         |37.7877                          |-122.4205                        |
|18f4ea86                         |37.3903                          |-122.0643                        |
|4ab5cbad                         |37.3952                          |-122.0813                        |
|8b6eae59                         |37.3944                          |-122.0813                        |
|4a7c7b41                         |37.4049                          |-122.0822                        |
|4ddad000                         |37.7857                          |-122.4011                        |
Query terminated
```

I also checked the same queries against the pull query API:
```
[john@arcturus ksql]$ http -v localhost:8088/query <<< '{"ksql":"SELECT * FROM riderLocations WHERE GEO_DISTANCE(latitude, longitude, 37.4133, -122.1162) <= 5;", "streamsProperties":{}}'
POST /query HTTP/1.1
Accept: application/json, */*;q=0.5
Accept-Encoding: gzip, deflate
Connection: keep-alive
Content-Length: 130
Content-Type: application/json
Host: localhost:8088
User-Agent: HTTPie/2.4.0

{
    "ksql": "SELECT * FROM riderLocations WHERE GEO_DISTANCE(latitude, longitude, 37.4133, -122.1162) <= 5;",
    "streamsProperties": {}
}


HTTP/1.1 200 OK
Transfer-Encoding: chunked
content-type: application/json

[{"header":{"queryId":"_confluent-ksql-default_transient_5215988584470098718_1622740716264","schema":"`PROFILEID` STRING, `LATITUDE` DOUBLE, `LONGITUDE` DOUBLE"}},
{"row":{"columns":["4ab5cbad",37.3952,-122.0813]}},
{"row":{"columns":["8b6eae59",37.3944,-122.0813]}},
{"row":{"columns":["4a7c7b41",37.4049,-122.0822]}},


[john@arcturus ksql]$ http -v localhost:8088/query <<< '{"ksql":"SELECT * FROM riderLocations;", "streamsProperties":{}}'
POST /query HTTP/1.1
Accept: application/json, */*;q=0.5
Accept-Encoding: gzip, deflate
Connection: keep-alive
Content-Length: 65
Content-Type: application/json
Host: localhost:8088
User-Agent: HTTPie/2.4.0

{
    "ksql": "SELECT * FROM riderLocations;",
    "streamsProperties": {}
}


HTTP/1.1 200 OK
Transfer-Encoding: chunked
content-type: application/json

[{"header":{"queryId":"_confluent-ksql-default_transient_3822094306738327733_1622740726737","schema":"`PROFILEID` STRING, `LATITUDE` DOUBLE, `LONGITUDE` DOUBLE"}},
{"row":{"columns":["c2309eec",37.7877,-122.4205]}},
{"row":{"columns":["18f4ea86",37.3903,-122.0643]}},
{"row":{"columns":["4ab5cbad",37.3952,-122.0813]}},
{"row":{"columns":["8b6eae59",37.3944,-122.0813]}},
{"row":{"columns":["4a7c7b41",37.4049,-122.0822]}},
{"row":{"columns":["4ddad000",37.7857,-122.4011]}},

[john@arcturus ksql]$ 
```

### Reviewer checklist


ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/resources/streaming/StreamedQueryResource.java
```diff
@@ -404,6 +510,201 @@ private EndpointResponse handlePullQuery(
     }
   }
 
+  // There's no good way to find out the current position of
+  // Streams's Consumers on specific partitions, so I made this
```
agavra, line 514: I may be missing something, but could we use `ProcessorContext#partition` and `ProcessorContext#offset` and add a `process` step into the topology at the end? Another benefit of that is that we could have it stop processing any partitions that have passed their end-offset directly in there and not rely on something asynchronous to shut it down all-or-nothing

ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/resources/streaming/StreamedQueryResource.java
```diff
@@ -404,6 +510,201 @@ private EndpointResponse handlePullQuery(
...
+      QueryCapacityUtil.throwTooManyActivePushQueriesException(
+              ksqlEngine,
+              ksqlRestConfig,
+              statement.getStatementText()
+      );
+    }
+
+    final ServiceContextOverride override = new ServiceContextOverride(serviceContext);
+
+    final TransientQueryMetadata query = ksqlEngine
```
agavra, line 667: I guess this approach would allow us to do aggregates and all those other fun things as well, right?

agavra (commented):
Some comments - but don't take me too seriously :)

ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/resources/streaming/StreamedQueryResource.java
```diff
@@ -404,6 +510,201 @@ private EndpointResponse handlePullQuery(
...
+      final CompletableFuture<Void> connectionClosedFuture,
+      final Map<TopicPartition, Long> endOffsets) {
+
+    // stream pull queries always start from earliest.
+    streamsProperties.put("auto.offset.reset", "earliest");
+
+    final ConfiguredStatement<Query> configured = ConfiguredStatement
+        .of(statement, SessionConfig.of(ksqlConfig, streamsProperties));
+
+    if (QueryCapacityUtil.exceedsPushQueryCapacity(ksqlEngine, ksqlRestConfig)) {
```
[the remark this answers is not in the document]
vvcephei: Thanks for pointing this out. I noticed it, but didn't give it much thought. I kind of do think it should be accounted in some way because as implemented, it's a full-blown Streams app for the duration of the query.

ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/resources/streaming/StreamedQueryResource.java
vvcephei: Yeah, I did it that way first. I'm not sure if it wound up in any of the in-progress commits in this branch.

One problem is that a processor doesn't have the power to stop its own task or request shutdown of the runtime or anything like that. The only thing it could do is just start to drop all records it receives, but that alone doesn't help the query to terminate.

What did kind of work for me was passing a semaphore in to a processor at the end of the topology, which used the context to detect when it was "done" (just as you describe), and count-down the semaphore. Then, there was a thread that I hackily started in the StreamedQueryResource that blocked on the semaphore and then shut down the app. The problem I faced is that a processor at the end of the topology actually doesn't see all the records (for example in the test query I did with a `WHERE ` cause), so it won't know when Streams as a whole has passed the desired end point.

My conclusion was that there's not really a great way to implement the desired logic in the form of a processor, at least not without a significant rewrite of the way KSQL constructs the Streams topology.

Another alternative I considered was to change the "push query queue" so that the transient query would not just pipe out the GenericRow, but also include other metadata like the partition and offset, so we could decide in the QueryStreamWriter directly when we are "done". But that's also vulnerable to the fact that if we filter out some records, the writer isn't guaranteed to see all the offsets we've processed.

ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/resources/streaming/StreamedQueryResource.java
vvcephei: Yes, but the we should give some thought to how to surface amendments to users.

vvcephei: Ah, looks like it will take more work to add it to the parser:
```
ksql> SELECT * FROM riderLocations WINDOW TUMBLING (SIZE 1 SECOND) GROUP BY PROFILEID;
Pull queries don't support WINDOW clauses. See https://cnfl.io/queries for more info.
Caused by: Pull queries don't support WINDOW clauses.
ksql> SELECT * FROM riderLocations GROUP BY PROFILEID;
Pull queries don't support GROUP BY clauses. See https://cnfl.io/queries for more info.
Caused by: Pull queries don't support GROUP BY clauses.

```

Probably for the best, since it will let us think about the ramifications.

ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/resources/streaming/StreamedQueryResource.java
vvcephei: Oh, boy. None of this was necessary. I was thinking that we couldn't use committed offsets, because "transient" queries wouldn't commit offsets... But of course transient queries are just regular Streams apps and must commit offsets, so we can use the committedOffsets API to know when Streams has passed the desired point, no magic required.

guozhangwang (commented):
I know our terms are now mixing more or less now, but I feel we should call it `push queries over streams` rather than pull queries? :P

Anyways, I would advocate we categorize queries that need a streaming API (websocket, or http 2) and in the old language end with `EMIT CHANGES` and in new language would start with `SUBSCRIBE TO ..` into the same bucket, and whereas pull queries would only for req/resp typed API and would be on tables only.

ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/resources/streaming/StreamedQueryResource.java
guozhangwang: Thanks for checking~

config/ksql-server.properties
```diff
@@ -65,7 +65,7 @@ ksql.logging.processing.stream.auto.create=true
 #------ Kafka -------
 
 # The set of Kafka brokers to bootstrap Kafka cluster information from:
-bootstrap.servers=localhost:9092
+bootstrap.servers=localhost:29092
```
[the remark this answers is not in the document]
vvcephei: Not intended for merging, this was just because I was testing using the docker-compose from the quickstart.

ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/resources/streaming/StreamedQueryResource.java
```diff
@@ -300,7 +312,94 @@ private EndpointResponse handleStatement(
...
+  }
+
+  @NotNull
+  private Map<TopicPartition, Long> getEndOffsets(final DataSource dataSource) {
+    final DefaultKafkaClientSupplier defaultKafkaClientSupplier = new DefaultKafkaClientSupplier();
+    // these are not the right configs to pass a consumer.
+    final Map<String, Object> props = ksqlConfig.getKsqlAdminClientConfigProps();
+    try (Admin admin = defaultKafkaClientSupplier
+            .getAdmin(props)) {
+      try (Consumer<byte[], byte[]> consumer =
```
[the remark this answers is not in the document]
vvcephei: Yeah, I thought it was a weird gap in the Admin interface, but I couldn't find it. Did I miss something?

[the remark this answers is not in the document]
guozhangwang:
```
Admin#listOffsets
```

would do.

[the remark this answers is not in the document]
vvcephei: Oh boy, I sure did miss something. Thanks!

Quick question. If I use `OffsetSpec.latest()` with the listOffsets request, does it account for transaction markers? I.e., can I rely on the offset it gives me back to be an offset of an actual record?

vvcephei:
Trying out a new process to get repeatable builds:

1. rebase onto the latest nanoversion tag: `git rebase -i v7.1.0-28`
2. set the project's version for development: `mvn versions:set -DnewVersion=7.1.0-SNAPSHOT versions:commit`
3. proceed with development

During development, I will only rebase onto subsequent nanoversion tags.
Before merging, I'll have to rebase on `master` to unpin all dependencies.

ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/resources/streaming/StreamedQueryResource.java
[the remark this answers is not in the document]
guozhangwang: That can be specified in `ListOffsetsOptions`, and via `listOffsets(Map<TopicPartition, OffsetSpec> topicPartitionOffsets, ListOffsetsOptions options)`

guozhangwang:
I took another quick pass on the PR and do not find any major issues. I think we can start testing it out in benchmarks/soak now to see how it looks like.

PR #7637, opened 2021-06-04
docs: add source generation to IDE instructions

vvcephei:
### Description 

Refine the IDE instructions based on feedback.

### Testing done 

1. Ran `./mvnw clean`, reproduced the problem with missing Lexer/Parser and KsqlVersionMetrics classes.
2. Ran the new commands, then verified I can build cleanly in IDEA.

### Reviewer checklist


guozhangwang (approved):
Verified locally that the added cmd `./mvnw --projects ksqldb-parser,ksqldb-version-metrics-client generate-sources` works on my IDEA.

PR #7643, opened 2021-06-07
Removing a duplicated clause.

krisajenkins:
The phrase `(UDF for short)` appears in two places. The second is redundant, and kinda confusing as it implies 'UDF' is short for 'scalar function'.

### Description 

### Testing done 

### Reviewer checklist
- [X] Ensure docs are updated if necessary. (eg. if a user visible feature is being added or changed).
- [X] Ensure relevant issues are linked (description should include text like "Fixes #<issue number>")


JimGalasyn (approved):
LGTM, thanks for the fix!

PR #7661, opened 2021-06-08
Scalable push queries end to end

AlanConfluent:
### Description 

This PR adds scalable push query support to the query endpoints.  What can be handled by scalable push queries is checked in `ScalablePushUtil`, which requires that the query:
- Be a push query with `EMIT CHANGES`
- Have no group by
- Have no window function
- Have no partition by
- Have a config or override that sets `auto.offset.reset` to `latest`.

The endpoints specifically supported are:
- Web socket - `/ws/query`
- Stream endpoint - `/query-stream`
- Old endpoint - `/query`

More tests are coming, namely RQTT and multi-node routing tests.  The main tests in this PR beyond unit tests are in `RestApiTest` which verifies that each of the endpoints work end to end.

### Testing done 
Unit tests and `RestApiTest`.

### Reviewer checklist


ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/resources/streaming/PushQueryPublisher.java
```diff
@@ -63,32 +70,64 @@
...
+      localCommands.ifPresent(lc -> lc.write((TransientQueryMetadata) queryMetadata));
+    }
+
     final PushQuerySubscription subscription =
         new PushQuerySubscription(exec, subscriber, queryMetadata);
 
-    localCommands.ifPresent(lc -> lc.write(queryMetadata));
-
-    log.info("Running query {}", queryMetadata.getQueryApplicationId());
+    log.info("Running query {}", queryMetadata.getQueryId().toString());
```
agavra, line 116: I started to follow the code here but it isn't super straightforward - what's the queryID of a SPQ? is it the same as the underlying persistent query? should we give it a meaningful name (especially if we want to have an admin api that allows users to terminate a single instance of an SPQ)?

agavra (approved):
LGTM! I struggle a bit to wrap my head around all the vertx publisher and async handle things, but it seems to work so I'm happy

AlanConfluent (commented):
@agavra Thanks for the review.  Still working on your feedback but pushing some of my comments.

ksqldb-engine/src/main/java/io/confluent/ksql/planner/plan/QueryProjectNode.java
```diff
@@ -167,6 +170,17 @@ private LogicalSchema buildOutputSchema(final MetaStore metaStore) {
 
       outputSchema = selectOutputSchema(metaStore, projects, isWindowed);
     }
+
+    if (isScalablePush) {
+      // Transient queries return key columns in the value, so the projection includes them, and
+      // the schema needs to include them too:
```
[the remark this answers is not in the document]
AlanConfluent: I would say that the primary reason is that we use the same physical layer to execute the operation as with pull queries (`ProjectOperator`), and this assumes it's using a `QueryProjectNode`.  I agree that this is a bit of redundant logic since this class exists largely because pull queries are not handled well by `FinalProjectNode` -- see comments here: https://github.com/confluentinc/ksql/pull/6684

For SPQs, it might be worth detangling these things.  We also make use of some additional things like the concept of an intermediate schema.  I think the intention was that we could push down projections and do other things that are normally done in a relational db, where each layer might have a slightly different schema required to do the computation.  This is a different concept from how persistent and transient queries work today (and we currently don't do the optimization since we're passing around a table row with the whole data anyway 🤦 ).  We'll have to eventually ask Vicky what the vision there is.  Anyway, I could see us eventually merging back `QueryProjectNode` and `FinalProjectNode`, or diverging and making `QueryProjectNode` more specialized for relational type operations.

ksqldb-engine/src/main/java/io/confluent/ksql/util/ScalablePushQueryMetadata.java
```diff
@@ -0,0 +1,116 @@
...
+    .exceptionally(future::completeExceptionally);
+  }
+
+  @Override
+  public void close() {
+    rowQueue.close();
+    final PushConnectionsHandle handle = handleRef.get();
+    if (handle != null) {
+      handle.close();
+    }
```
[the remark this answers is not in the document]
AlanConfluent: Yeah, good catch.  I introduced a start future which returns the handle, and then the callback for this future will close it, so it ensures that whenever it returns, it gets closed.

ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/resources/streaming/PushQueryPublisher.java
```diff
@@ -63,32 +70,64 @@
     this.exec = requireNonNull(exec, "exec");
     this.query = requireNonNull(query, "query");
     this.localCommands = requireNonNull(localCommands, "localCommands");
+    this.pushRouting = null;
+    this.isScalablePush = false;
+    this.context = null;
+  }
+
+  PushQueryPublisher(
```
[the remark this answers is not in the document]
AlanConfluent: Done

ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/resources/streaming/PushQueryPublisher.java
AlanConfluent: The query id is something like "SCALABLE_PUSH_QUERY_<TIMESTAMP>".  This will hopefully be used as you suggest.  We could potentially make the name more meaningful if we expose it in in this way.  I agree that adding the source makes sense. I can do that in a followup PR (maybe along with the admin API work) if that works.  What do you think?

ksqldb-common/src/main/java/io/confluent/ksql/util/KsqlRequestConfig.java
```diff
@@ -50,6 +50,12 @@
   private static final String KSQL_REQUEST_QUERY_PULL_PARTITIONS_DOC =
       "Indicates which partitions to limit pull queries to.";
 
+  public static final String KSQL_REQUEST_QUERY_PUSH_SKIP_FORWARDING =
+      "request.ksql.query.push.skip.forwarding";
```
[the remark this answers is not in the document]
agavra: ack, sgtm

ksqldb-engine/src/main/java/io/confluent/ksql/engine/ScalablePushQueryExecutionUtil.java
```diff
@@ -0,0 +1,58 @@
...
+
+    if (queries.isEmpty()) {
+      throw new IllegalStateException(
+          "Scalable push queries require a query that has a sink. "
+              + "Source Name: " + sourceName);
+    }
+
+    if (queries.size() > 1) {
+      throw new IllegalStateException(
+          "Scalable push queries on work on sources that have a single writer query. "
```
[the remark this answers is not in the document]
agavra: great - let's add a GH issue to track this 😈

ksqldb-engine/src/main/java/io/confluent/ksql/planner/LogicalPlanner.java
```diff
@@ -182,7 +182,10 @@ public OutputNode buildPersistentLogicalPlan() {
     return buildOutputNode(currentNode);
   }
 
-  public OutputNode buildPullLogicalPlan(final PullPlannerOptions pullPlannerOptions) {
+  public OutputNode buildPullLogicalPlan(
```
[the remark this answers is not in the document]
agavra: 1) yeah no renaming configs unfortunately :( 
2) don't feel great about that either, but I have no better suggestions haha so we can go with whatever you prefer

ksqldb-engine/src/main/java/io/confluent/ksql/planner/plan/QueryProjectNode.java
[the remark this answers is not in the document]
agavra: ack

ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/resources/streaming/PushQueryPublisher.java
agavra: yeah, we can have this in a separate PR (definitely not urgent) we can file a ticket to track that as well (could be a good newbie task)

ksqldb-engine/src/main/java/io/confluent/ksql/physical/scalablepush/ScalablePushRegistry.java
```diff
@@ -52,6 +52,7 @@
 
   private final PushLocator pushLocator;
   private final LogicalSchema logicalSchema;
+  private final boolean isTable;
```
guozhangwang, line 55: This is not for this line but while reviewing this class I found that in https://github.com/confluentinc/ksql/pull/7661/files#diff-350c1eaff8254e08b2c4db5fab86e15d71b9e6a1590a698e2adc0e89a1e7e1c2R161 below, within

```
    public void process(final Object key, final GenericRow value) {
```

we do not need to call `this.context.forward(key, value);` after `handleRow` since this should never have any children processors, and if it ever do, calling it may result in unwanted behavior.

ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/util/ScalablePushUtil.java
```diff
@@ -0,0 +1,109 @@
...
+    final Query query = (Query) statement;
+    final SourceFinder sourceFinder = new SourceFinder();
+    sourceFinder.process(query.getFrom(), null);
+    // It will be present if it's not a join, which we don't handle
+    if (!sourceFinder.getSourceName().isPresent()) {
+      return false;
+    }
+    // Find all of the writers to this particular source.
+    final SourceName sourceName = sourceFinder.getSourceName().get();
+    final Set<QueryId> queries = ksqlEngine.getQueriesWithSink(sourceName);
```
guozhangwang, line 61: nit: maybe rename to `upstreamQueries` to distinguish from `query`?

ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/util/ScalablePushUtil.java
```diff
@@ -0,0 +1,109 @@
...
+        && !query.getHaving().isPresent()
+        // Partition by is not supported
+        && !query.getPartitionBy().isPresent()
+        // There must be an EMIT CHANGES clause
+        && (query.getRefinement().isPresent()
+            && query.getRefinement().get().getOutputRefinement() == OutputRefinement.CHANGES)
+        // Must be reading from "latest"
+        && isLatest
+        // We only handle a single sink source at the moment from a CTAS/CSAS
+        && queries.size() == 1;
```
guozhangwang, line 83: Do we support the case if the upstream query is emitting a windowed table? For a windowed table, its `toStream` would result in a `KStream<Windowed<K>, V>` and the processor would be on `KeyValue<Windowed<K>, V>`. How is the key type being translated on the ksql layer?

ksqldb-engine/src/main/java/io/confluent/ksql/physical/scalablepush/ScalablePushRegistry.java
AlanConfluent: Ok, makes sense.  I didn't know if it was considered good practice to do this (even if you know there are no children). I removed that line.

ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/util/ScalablePushUtil.java
AlanConfluent: Done

ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/util/ScalablePushUtil.java
AlanConfluent: Yes, we do.  If you look in `ScalablePushRegistry.handleRow`, you can see that we take the key as an Object in the method, see if it's windowed, and if so, interprets the key as a `Windowed<GenericKey>`.

guozhangwang: kk, in that case if the push query filters on the key in `WHERE` clause, we'd need to first extract the `GenericKey` from `Windowed<GenericKey>` and then apply the operator right? Could you point me to the extraction logic just for my education? Tried to search it somewhere but did not found yet.

PR #7662, opened 2021-06-08
feat: Support GRACE PERIOD syntax on WITHIN clause with simple window

spena:
### Description 
Part of [KLIP-36](https://github.com/confluentinc/ksql/blob/master/design-proposals/klip-36-grace-stream-stream-joins.md), this PR adds support of the `GRACE PERIOD` syntax in the `WITHIN` clause for joins with simple windows. 

The new syntax is optional and it is used like this:
```
SELECT * FROM s1 JOIN s2 WITHIN 1 MINUTE GRACE PERIOD 30 SECONDS ON <condition>
SELECT * FROM s1 JOIN s2 WITHIN (1 MINUTE, 2 MINUTE) GRACE PERIOD 30 SECONDS ON <condition>
```

The PR also updates the `QueryAnonymizer` and its tests.

Follow-up PRs will include the following:
- Call the grace() API in kstreams
- QTT tests for joins

### Testing done 
Updated unit tests

### Reviewer checklist


ksqldb-parser/src/main/antlr4/io/confluent/ksql/parser/SqlBase.g4
```diff
@@ -228,8 +228,9 @@ joinWindow
     ;
 
 withinExpression
-    : '(' joinWindowSize ',' joinWindowSize ')' # joinWindowWithBeforeAndAfter
-    | joinWindowSize # singleJoinWindow
+    : '(' joinWindowSize ',' joinWindowSize ')'         # joinWindowWithBeforeAndAfter
```
mjsax, line 231: Where is the `BEFORE` and `AFTER` keyword?

mjsax, line 231: Do we need a 4th line for `before/after with grace` ?

ksqldb-engine/src/main/java/io/confluent/ksql/engine/rewrite/QueryAnonymizer.java
```diff
@@ -887,15 +887,31 @@ private static String visitWithinExpression(final WithinExpressionContext contex
...
+        final SqlBaseParser.JoinWindowWithGraceContext joinWindowWithGrace
+    ) {
+      return String.format("(SIZE %s %s, GRACE PERIOD %s %s) ",
+          "'0'",
+          joinWindowWithGrace.joinWindowSize().windowUnit().getText().toUpperCase(),
+          "'0'",
+          joinWindowWithGrace.gracePeriodClause().windowUnit().getText().toUpperCase());
+    }
+
     private static String getSizeAndUnitFromJoinWindowSize(
```
mjsax, line 915: Should we rename this method to `anonymizeSizeAndUnitFromJoinWindowSize` (also, should we remove `AndUnit` from the name, as we preserve the units? Or is the method incorrect and it _should_ indeed anonymize the unit, too)?

ksqldb-engine/src/main/java/io/confluent/ksql/engine/rewrite/QueryAnonymizer.java
```diff
@@ -887,15 +887,31 @@ private static String visitWithinExpression(final WithinExpressionContext contex
...
-            + "ie. \"WITHIN (10 seconds, 20 seconds)");
+            + "seconds\", a join window with before and after specified, "
+            + "ie. \"WITHIN (10 seconds, 20 seconds)\", or a join window with a grace period, "
+            + "ie. \"WITHIN (SIZE 10 seconds, GRACE PERIOD 5 seconds)\"");
       }
 
       return stringBuilder.toString();
     }
 
+    private static String anonymizeJoinWindowWithGrace(
```
mjsax, line 905: Should we unify this method with the existing `getSizeAndUnitFromJoinWindowSize` ?

ksqldb-parser/src/main/antlr4/io/confluent/ksql/parser/SqlBase.g4
spena: I'll add it in a follow-up PR (part of parallelized work).

spena: I'll add it in a follow-up PR (part of parallelized work).

ksqldb-engine/src/main/java/io/confluent/ksql/engine/rewrite/QueryAnonymizer.java
spena: I just renamed to `anonymizeJoinWindowSize`. The anonymizer is not changing the unit so I removed it.

ksqldb-engine/src/main/java/io/confluent/ksql/engine/rewrite/QueryAnonymizer.java
spena: I made some changes on this method.

ksqldb-engine/src/main/java/io/confluent/ksql/engine/rewrite/QueryAnonymizer.java
```diff
@@ -877,30 +877,42 @@ private static String visitWithinExpression(final WithinExpressionContext contex
...
+
+        stringBuilder.append(String.format("(SIZE %s, GRACE PERIOD %s)",
+            anonymizeJoinWindowSize(joinWithGracePeriod.joinWindowSize()),
+            anonymizeGracePeriod(joinWithGracePeriod.gracePeriodClause())));
       } else {
         throw new RuntimeException("Expecting either a single join window, ie \"WITHIN 10 "
-            + "seconds\", or a join window with before and after specified, "
-            + "ie. \"WITHIN (10 seconds, 20 seconds)");
+            + "seconds\", a join window with before and after specified, "
+            + "ie. \"WITHIN (10 seconds, 20 seconds)\", or a join window with a grace period, "
```
mjsax, line 899: Just realizing that this error message does not match our docs:
```
 [ WITHIN [(before TIMEUNIT, after TIMEUNIT) | N TIMEUNIT] ]
```

But the error message aligns to the grammar definition. Something seems to be off?

ksqldb-parser/src/main/java/io/confluent/ksql/parser/AstBuilder.java
```diff
@@ -578,24 +579,34 @@ private static Node visitWithinExpression(final SqlBaseParser.WithinExpressionCo
...
 
+        beforeSize = getSizeAndUnitFromJoinWindowSize(joinWithGracePeriod.joinWindowSize());
+        afterSize = beforeSize;
+        gracePeriod = gracePeriodClause(joinWithGracePeriod.gracePeriodClause());
       } else {
         throw new RuntimeException("Expecting either a single join window, ie \"WITHIN 10 "
-            + "seconds\", or a join window with before and after specified, "
-            + "ie. \"WITHIN (10 seconds, 20 seconds)");
+            + "seconds\", a join window with before and after specified, "
+            + "ie. \"WITHIN (10 seconds, 20 seconds)\", or a join window with a grace period, "
```
mjsax, line 600: As above. Seems `AFTER` and `BEFORE` keywords are missing?

mjsax:
Should we extend some existing QTT tests, to set a grace period and to test that "late" input is dropped for inner/left/outer stream-stream joins? 

@vcrfxia also just adds new n-way joins test that we might want to extend (https://github.com/confluentinc/ksql/pull/7606)

spena:
@mjsax I will follow-up the QTT tests in another PR. I will update and add some to test the "late" inputs. This PR only adds syntax. The grace period does not work yet. There's more code to add.

spena:
@mjsax I updated the PR with the new proposed syntax (including KLIP-36). The new syntax is:
```
# Normal
JOIN <stream> WITHIN <window> [GRACE PERIOD <grace-period>]

# Before/After
JOIN <stream> WITHIN (<before-window>, <after-window>) [GRACE PERIOD <grace-period>]
```

design-proposals/klip-36-grace-stream-stream-joins.md
```diff
@@ -97,14 +96,13 @@ the order was placed, and shipped within 2 hours of the payment being received.
...
+        INNER JOIN shipments s WITHIN 2 HOURS GRACE PERIOD 30 MINUTES ON s.id = o.id;
 ```
 
 ## Compatibility Implications
 
-- We should consider whether we want to support the older syntax or require all new joins
-    to use the `(SIZE <size>, GRACE PERIOD <size>)` syntax (noting that `GRACE PERIOD` will
-    be optional).
+None. The `GRACE PERIOD` will be optional, so users can continue using the current syntax without
+it.
```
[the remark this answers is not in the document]
spena: I have another follow-up PR to update the docs. I can update the KLIP (as it is a proposal), but I prefer to update the docs once the PR that enables the fix is merged.

spena:
@mjsax I updated the PR to support the GRACE on before/after window syntax. The unit tests are still failing, but that's related to another issue 'cause ksql/master fails with the same.

ksqldb-parser/src/main/antlr4/io/confluent/ksql/parser/SqlBase.g4
```diff
@@ -228,8 +228,8 @@ joinWindow
     ;
 
 withinExpression
-    : '(' joinWindowSize ',' joinWindowSize ')' # joinWindowWithBeforeAndAfter
-    | joinWindowSize # singleJoinWindow
+    : '(' joinWindowSize ',' joinWindowSize ')' (gracePeriodClause)?  # joinWindowWithBeforeAndAfter
```
vcrfxia, line 231: Are we planning to feature flag these changes in the event that not all the functionality makes it into the upcoming 0.20 release?

ksqldb-engine/src/main/java/io/confluent/ksql/engine/rewrite/QueryAnonymizer.java
```diff
@@ -877,30 +877,45 @@ private static String visitWithinExpression(final WithinExpressionContext contex
...
         throw new RuntimeException("Expecting either a single join window, ie \"WITHIN 10 "
-            + "seconds\", or a join window with before and after specified, "
-            + "ie. \"WITHIN (10 seconds, 20 seconds)");
+            + "seconds\" or \"WITHIN 10 seconds GRACE PERIOD 2 seconds\", or a join window with "
+            + "before and after specified, ie. \"WITHIN (10 seconds, 20 seconds)\" or "
+            + "WITHIN (10 seconds, 20 seconds) GRACE PERIOD 5 seconds");
       }
 
-      return stringBuilder.toString();
+      return stringBuilder.append(' ').toString();
```
vcrfxia, line 906: Is this change intentional? I assume it's fine if the tests below are passing, was just surprised to see it here.

ksqldb-parser/src/main/java/io/confluent/ksql/parser/tree/WithinExpression.java
```diff
@@ -33,32 +34,50 @@
...
   private final JoinWindows joinWindows;
+  private final Optional<WindowTimeClause> gracePeriod;
 
   public WithinExpression(final long size, final TimeUnit timeUnit) {
     this(size, size, timeUnit, timeUnit);
   }
 
+  public WithinExpression(final long size, final TimeUnit timeUnit,
+                          final WindowTimeClause gracePeriod) {
+    this(Optional.empty(), size, size, timeUnit, timeUnit, Optional.of(gracePeriod));
```
vcrfxia, line 45: nit: this is inconsistent with the constructor above. We should either update the constructor above so that all constructors call the one with all arguments, or we should update this constructor to parallel the one above by having this one call the one with `before` and `after` but without `location`.

ksqldb-parser/src/test/java/io/confluent/ksql/parser/tree/WithinExpressionTest.java
```diff
@@ -26,12 +31,45 @@
...
   public void shouldDisplayCorrectStringWithBeforeAndAfter() {
     final WithinExpression expression = new WithinExpression(30, 40, TimeUnit.MINUTES, TimeUnit.HOURS);
     assertEquals(" WITHIN (30 MINUTES, 40 HOURS)", expression.toString());
+    assertEquals(
+        JoinWindows.of(Duration.ofMinutes(30)).after(Duration.ofHours(40)),
+        expression.joinWindow());
+  }
+
+  @Test
+  public void shouldDisplayCorrectStringWithSingleGracePeriod() {
```
vcrfxia, line 49: nit:
```suggestion
  public void shouldDisplayCorrectStringWithGracePeriod() {
```

vcrfxia (approved):
Thanks @spena -- LGTM with comments inline.

ksqldb-parser/src/main/antlr4/io/confluent/ksql/parser/SqlBase.g4
mjsax: ksqlDB already has a feature flag to disable the stream-stream join fix -- I think it makes sense to use it to fail if `GRACE` is used and the feature was disabled. Do you want to get this pulled into this PR or a later PR?

ksqldb-parser/src/main/java/io/confluent/ksql/parser/tree/WithinExpression.java
```diff
@@ -107,13 +134,18 @@ public boolean equals(final Object o) {
...
 
   private JoinWindows createJoinWindows() {
     final JoinWindows joinWindow = JoinWindows
-        .of(Duration.ofMillis(beforeTimeUnit.toMillis(before)));
-    return joinWindow.after(Duration.ofMillis(afterTimeUnit.toMillis(after)));
+        .of(Duration.ofMillis(beforeTimeUnit.toMillis(before)))
+        .after(Duration.ofMillis(afterTimeUnit.toMillis(after)));
+
+    return (gracePeriod.isPresent())
+        ? joinWindow.grace(gracePeriod.get().toDuration())
```
mjsax, line 147: This is actually "incorrect". We should apply GRACE PERIOD only via the new `JoinWindows#ofSizeAndGrace()` and never call `grace()` directly. -- Of course, we first need to get the AK PR merged. 

Just wanted to point it out, that we need to change this in a follow up PR. (Wondering though, we we need to apply GRACE in this PR at all? Could we not also just ignore it when building the plan until we can do it properly? Or do we need it for some testing?)

ksqldb-parser/src/test/java/io/confluent/ksql/parser/KsqlParserTest.java
```diff
@@ -938,6 +939,37 @@ public void shouldSetWithinExpressionWithSingleWithin() {
     assertThat(withinExpression.getBefore(), is(10L));
     assertThat(withinExpression.getAfter(), is(10L));
     assertThat(withinExpression.getBeforeTimeUnit(), is(TimeUnit.SECONDS));
+    assertThat(withinExpression.getGrace(), is(Optional.empty()));
+    assertThat(join.getType(), is(JoinedSource.Type.INNER));
+  }
+
+  @Test
+  public void shouldSetWithinExpressionWithSingleWithinAndGracePeriod() {
```
mjsax, line 947: To follow up to @vcrfxia comment -- should we add a test that verifies that we fail if `GRACE` is specified but if the feature flag disables the fix? (not sure if a unit test or QTT test should cover it?)

ksqldb-parser/src/test/java/io/confluent/ksql/parser/KsqlParserTest.java
```diff
@@ -969,6 +1001,39 @@ public void shouldSetWithinExpressionWithBeforeAndAfter() {
     assertThat(withinExpression.getAfter(), is(20L));
     assertThat(withinExpression.getBeforeTimeUnit(), is(TimeUnit.SECONDS));
     assertThat(withinExpression.getAfterTimeUnit(), is(TimeUnit.MINUTES));
+    assertThat(withinExpression.getGrace(), is(Optional.empty()));
+    assertThat(join.getType(), is(JoinedSource.Type.INNER));
+  }
+
+  @Test
+  public void shouldSetWithinExpressionWithBeforeAndAfterAndGracePeriod() {
```
mjsax, line 1009: as above.

mjsax (approved):
Feel free to merge after `master` is fixed.

ksqldb-parser/src/main/antlr4/io/confluent/ksql/parser/SqlBase.g4
spena: I'll pull it into another PR

ksqldb-engine/src/main/java/io/confluent/ksql/engine/rewrite/QueryAnonymizer.java
spena: Yes. The previous string builders used to append the empty character at the end. To avoid duplicating the effort, I just moved that empty char to the end of the method.

ksqldb-parser/src/main/java/io/confluent/ksql/parser/tree/WithinExpression.java
spena: There's another jira is planned to call the new API, so a follow-up PR will include it once the AK PR is merged. Another thing is that this part of the code, which creates the `JoinWindows`, does not actually do anything with grace in the Kafka streams yet. Later on, another part of the code just gets the after/before windows from this JoinWindows, but not the grace period yet. This is just a preparation for the WithinExpression.

ksqldb-parser/src/test/java/io/confluent/ksql/parser/KsqlParserTest.java
spena: I'll add a test in a follow-up PR

ksqldb-parser/src/test/java/io/confluent/ksql/parser/KsqlParserTest.java
spena: I'll add a test in a follow-up PR

PR #7667, opened 2021-06-10
fix: make current java clients compatible with pre-0.15 servers

jzaralim:
### Description 
If `sourceConstraints` is not included in the server response, then default to an empty list in the java client.

### Testing done 
Added a unit test

### Reviewer checklist


vcrfxia (approved):
Thanks @jzaralim . It's unfortunate that we need this fix -- https://github.com/confluentinc/ksql/issues/6042 should address this going forward. Thanks for the fix in the meantime!

PR #7678, opened 2021-06-15
feat: call KStreams API grace() when GRACE PERIOD is used in joins

spena:
### Description 
Follow up on https://github.com/confluentinc/ksql/pull/7662. This PR makes the connection from the `GRACE PERIOD` syntax in stream-stream joins and the call to the KStreams grace API, which enables the use of grace in join windows.

This is the workflow that happens internally to help review the code:

----------------
When a join happens. The parsing creates a `WithinExpression` which has the windows and optional grace (updated in https://github.com/confluentinc/ksql/pull/7662).

This window + optional grace are obtained from the `JoinNode` and passed to the `SchemaKStream` class, which accepts
an optional grace. The `SchemaKStream` then passes the window + optional grace to the `ExecutionStepFactory` which
gets the before/after and grace values in milliseconds, and then passes them to the `StreamStreamJoin`.

The `StreamStreamJoin` accepts graceMillis as optional because seems it is needed for compatibility with an old plan written
in JSON?. Anyway, the grace must be optional to continue with the call to the `KSPlanBuilder.visitStreamStreamJoin()`,
which later ends up calling the `StreamStreamJoinBuilder.build()`. 

The `StreamStreamJoinBuilder` is the final class that builds the windows + grace and calls the Kafka streams API. The
grace is optional in this class in order to know whether to call the new KStreams API ofSizeAndGrace() or just call
the normal of/after calls.

The `ofSizeAndGrace()` is called is not done in this PR. A follow-up PR will be done after the Kafka PR that contains this new method is merged.

----------------

### Testing done 
Updated unit tests
Add QTT tests

### Reviewer checklist

### DIFF
--- a/ksqldb-engine/src/main/java/io/confluent/ksql/structured/SchemaKStream.java
+++ b/ksqldb-engine/src/main/java/io/confluent/ksql/structured/SchemaKStream.java
@@ -40,10 +40,10 @@ import io.confluent.ksql.execution.streams.ExecutionStepFactory;
 import io.confluent.ksql.execution.streams.StepSchemaResolver;
 import io.confluent.ksql.execution.timestamp.TimestampColumn;
 import io.confluent.ksql.execution.util.ExpressionTypeManager;
-import io.confluent.ksql.execution.windows.WindowTimeClause;
 import io.confluent.ksql.function.FunctionRegistry;
 import io.confluent.ksql.model.WindowType;
 import io.confluent.ksql.name.ColumnName;
+import io.confluent.ksql.parser.tree.WithinExpression;
 import io.confluent.ksql.planner.plan.PlanBuildContext;
 import io.confluent.ksql.schema.ksql.LogicalSchema;
 import io.confluent.ksql.schema.ksql.types.SqlType;
@@ -61,8 +61,6 @@ import java.util.List;
 import java.util.Objects;
 import java.util.Optional;
 import java.util.stream.Collectors;
-
-import org.apache.kafka.streams.kstream.JoinWindows;
 
 // CHECKSTYLE_RULES.OFF: ClassDataAbstractionCoupling
 @SuppressWarnings("OptionalUsedAsFieldOrParameterType")
@@ -196,8 +194,7 @@ public class SchemaKStream<K> {
   public SchemaKStream<K> leftJoin(
       final SchemaKStream<K> otherSchemaKStream,
       final ColumnName keyColName,
-      final JoinWindows joinWindows,
-      final Optional<WindowTimeClause> gracePeriod,
+      final WithinExpression withinExpression,
       final FormatInfo leftFormat,
       final FormatInfo rightFormat,
       final Stacker contextStacker
@@ -212,8 +209,8 @@ public class SchemaKStream<K> {
         InternalFormats.of(keyFormat, rightFormat),
         sourceStep,
         otherSchemaKStream.sourceStep,
-        joinWindows,
-        gracePeriod
+        withinExpression.joinWindow(),
+        withinExpression.getGrace()
     );
 
     return new SchemaKStream<>(
@@ -254,8 +251,7 @@ public class SchemaKStream<K> {
   public SchemaKStream<K> innerJoin(
       final SchemaKStream<K> otherSchemaKStream,
       final ColumnName keyColName,
-      final JoinWindows joinWindows,
-      final Optional<WindowTimeClause> gracePeriod,
+      final WithinExpression withinExpression,
       final FormatInfo leftFormat,
       final FormatInfo rightFormat,
       final Stacker contextStacker
@@ -270,8 +266,8 @@ public class SchemaKStream<K> {
         InternalFormats.of(keyFormat, rightFormat),
         sourceStep,
         otherSchemaKStream.sourceStep,
-        joinWindows,
-        gracePeriod
+        withinExpression.joinWindow(),
+        withinExpression.getGrace()
     );
 
     return new SchemaKStream<>(
@@ -286,8 +282,7 @@ public class SchemaKStream<K> {
   public SchemaKStream<K> outerJoin(
       final SchemaKStream<K> otherSchemaKStream,
       final ColumnName keyColName,
-      final JoinWindows joinWindows,
-      final Optional<WindowTimeClause> gracePeriod,
+      final WithinExpression withinExpression,
       final FormatInfo leftFormat,
       final FormatInfo rightFormat,
       final Stacker contextStacker
@@ -302,8 +297,8 @@ public class SchemaKStream<K> {
         InternalFormats.of(keyFormat, rightFormat),
         sourceStep,
         otherSchemaKStream.sourceStep,
-        joinWindows,
-        gracePeriod
+        withinExpression.joinWindow(),
+        withinExpression.getGrace()
     );
 
     return new SchemaKStream<>(
--- a/ksqldb-engine/src/main/java/io/confluent/ksql/structured/SchemaKStream.java
+++ b/ksqldb-engine/src/main/java/io/confluent/ksql/structured/SchemaKStream.java
@@ -40,6 +40,7 @@ import io.confluent.ksql.execution.streams.ExecutionStepFactory;
 import io.confluent.ksql.execution.streams.StepSchemaResolver;
 import io.confluent.ksql.execution.timestamp.TimestampColumn;
 import io.confluent.ksql.execution.util.ExpressionTypeManager;
+import io.confluent.ksql.execution.windows.WindowTimeClause;
 import io.confluent.ksql.function.FunctionRegistry;
 import io.confluent.ksql.model.WindowType;
 import io.confluent.ksql.name.ColumnName;
@@ -196,6 +197,7 @@ public class SchemaKStream<K> {
       final SchemaKStream<K> otherSchemaKStream,
       final ColumnName keyColName,
       final JoinWindows joinWindows,
+      final Optional<WindowTimeClause> gracePeriod,
       final FormatInfo leftFormat,
       final FormatInfo rightFormat,
       final Stacker contextStacker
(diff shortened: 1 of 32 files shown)


guozhangwang (commented):
Thank you so much for the great PR description!

I'm half way there and would try to complete by eow.

ksqldb-engine/src/main/java/io/confluent/ksql/structured/SchemaKStream.java
```diff
@@ -196,6 +197,7 @@ static Expression rewriteTimeComparisonForFilter(final Expression expression) {
       final SchemaKStream<K> otherSchemaKStream,
       final ColumnName keyColName,
       final JoinWindows joinWindows,
+      final Optional<WindowTimeClause> gracePeriod,
```
guozhangwang, line 200: Why can't we just retrieve the grace period from the JoinWindows? I.e. we just set the grace time in JoinWindows if `GRACE` specified, and then pass around that object only.

ksqldb-functional-tests/src/test/resources/query-validation-tests/joins.json
```diff
@@ -855,6 +855,81 @@
...
+        {"topic": "right_topic", "key": 2, "value": {"L2": "c"}, "timestamp": 90000},
+        {"topic": "left_topic", "key": 3, "value": {"L1": "D"}, "timestamp": 60000},
+        {"topic": "right_topic", "key": 3, "value": {"L2": "d"}, "timestamp": 60000}
+      ],
+      "outputs": [
+        {"topic": "LEFT_JOIN", "key": 0, "value": {"L1": "A", "L2": null}, "timestamp": 0},
+        {"topic": "LEFT_JOIN", "key": 0, "value": {"L1": "A", "L2": "a"}, "timestamp": 60000},
+        {"topic": "LEFT_JOIN", "key": 1, "value": {"L1": "B", "L2": null}, "timestamp": 330000},
+        {"topic": "LEFT_JOIN", "key": 2, "value": {"L1": "C", "L2": null}, "timestamp": 90000},
+        {"topic": "LEFT_JOIN", "key": 2, "value": {"L1": "C", "L2": "c"}, "timestamp": 90000},
```
guozhangwang, line 902: Hmm.. is the output expected? With our fix shouldn't we only output one record for key 2, with C-c, and one record for key 3, with D-d?

ksqldb-functional-tests/src/test/resources/query-validation-tests/joins.json
```diff
@@ -855,6 +855,81 @@
...
+        {"topic": "right_topic", "key": 3, "value": {"L2": "d"}, "timestamp": 60000},
+        {"topic": "left_topic", "key": 3, "value": {"L1": "D"}, "timestamp": 60000}
+      ],
+      "outputs": [
+        {"topic": "OUTER_JOIN", "key": 0, "value": {"T_ID": 0, "TT_ID": null, "L1": "A", "L2": null}, "timestamp": 0},
+        {"topic": "OUTER_JOIN", "key": 0, "value": {"T_ID": 0, "TT_ID": 0, "L1": "A", "L2": "a"}, "timestamp": 60000},
+        {"topic": "OUTER_JOIN", "key": 1, "value": {"T_ID": null, "TT_ID": 1, "L1": null, "L2": "b"}, "timestamp": 330000},
+        {"topic": "OUTER_JOIN", "key": 2, "value": {"T_ID": null, "TT_ID": 2, "L1": null, "L2": "c"}, "timestamp": 90000},
+        {"topic": "OUTER_JOIN", "key": 2, "value": {"T_ID": 2, "TT_ID": 2, "L1": "C", "L2": "c"}, "timestamp": 90000},
+        {"topic": "OUTER_JOIN", "key": 3, "value": {"T_ID": null, "TT_ID": 3, "L1": null, "L2": "d"}, "timestamp": 60000},
```
guozhangwang, line 929: Ditto above.

ksqldb-engine/src/main/java/io/confluent/ksql/structured/SchemaKStream.java
guozhangwang: EDIT: I knew this is only used for `WITH` and the `GRACE PERIOD` is out of the scope; I guess my meta question is: why can't we have the `GRACE PERIOD` captured as optional in `WithinExpression` as well?

guozhangwang (commented):
I have a major question about the added test: with the fix in place, the output seems still spurious? Otherwise, just minor questions.

ksqldb-engine/src/main/java/io/confluent/ksql/structured/SchemaKStream.java
spena: The `JoinWindows` has a default grace (24h) if it is not passed one. I cannot know if the grace was optional in the ksql syntax with that default. Say a user set GRACE PERIOD 24 HOURS, then `JoinWindows` will be set to 24h, which it could be the default (no fix enabled) or manually set (fix enabled).

The grace is optional in the `WithinExpression`. I could pass the `WithinExpression` all down to the `StreamStreamJoinBuilder` but I would need to add the `ksqldb-parser` dependency into the `ksqldb-streams` module.

ksqldb-functional-tests/src/test/resources/query-validation-tests/joins.json
spena: The fix is not enabled yet. These results are working with the spurious results bug.

ksqldb-functional-tests/src/test/resources/query-validation-tests/joins.json
spena: The fix is not enabled yet. These results are working with the spurious results bug.

guozhangwang: I see.

ksqldb-engine/src/main/java/io/confluent/ksql/structured/SchemaKStream.java
guozhangwang: Let's say in the future if we do not need to rely on GRACE to enable the fix or not, then would this code be changed again to only pass in the `JoinWindows` object still?

If yes, I'd suggest we just keep doing that right away: we could, e.g., a wild idea is to have an extended JoinWindows that adds a flag indicating if user specified grace, and then pass around that object from WithinExpression -> ExecutionStepFactory -> StreamStreamJoin, and let it to return a `graceSpecified` rather than `join.getGraceMillis()`, used by the `StreamStreamJoinBuilder`. In the future we can just remove the extended class and the `graceSpecified` from StreamStreamJoin. Does that work?

ksqldb-functional-tests/src/test/resources/query-validation-tests/joins.json
spena: We need to wait for the fix in kafka. But it should not block this PR 'cause enabling the fix will involve other changes in QTT. A follow-up PR for making changes in the QTT framework is here (which is required to enable the fix and update the tests) - https://github.com/confluentinc/ksql/pull/7695

ksqldb-engine/src/main/java/io/confluent/ksql/structured/SchemaKStream.java
spena: I could do that. But it would only be used to pass the window+graceSpecified from `SchemaKStream#leftJoin()` -> `ExecutionStepFactory#streamStreamJoin()`. After that, the rest should be passed as millis to the `StreamStreamJoin` because of compatibility. Look at the `StreamStreamJoin` class. It has a constructor that is used by a JSON mapper when reads a plan. That already accepts before/after millis. I cannot change that constructor to pass the `JoinWindows`.

I'm not sure if it is worth creating a class around the JoinWindows for one single change and then remove it. I think we can remove the gracePeriod later easier than removing a class.

ksqldb-functional-tests/src/test/resources/query-validation-tests/joins.json
guozhangwang: Sounds good.

ksqldb-engine/src/main/java/io/confluent/ksql/structured/SchemaKStream.java
guozhangwang: Okay, I think I'm convinced. Let's just keep it as is then.

guozhangwang (approved):

spena:
Thanks @guozhangwang. I just merged the PR.

ksqldb-streams/src/main/java/io/confluent/ksql/execution/streams/StreamStreamJoinBuilder.java
```diff
@@ -85,8 +85,16 @@ private StreamStreamJoinBuilder() {
...
 
-    final JoinWindows joinWindows =
+    JoinWindows joinWindows =
         JoinWindows.of(join.getBeforeMillis()).after(join.getAfterMillis());
+
+    // Grace, as optional, helps to identify if a user use the new GRACE PERIOD syntax in the
+    // join window or not. If used, then we'll call the new KStreams API ofSizeAndGrace() which
+    // enables the bugfix with left/outer joins (see KAFKA-10847).
+    if (join.getGraceMillis().isPresent()) {
+      joinWindows = joinWindows.grace(join.getGraceMillis().get());
```
mjsax, line 95: This is calling the old/exiting `grace()` and thus it does not enable this fix. I think we need to change this (starting in L88):
```
final JoinWindows joinWindows;
if (join.getGraceMillis().isPresent()) {
    joinWindows = JoinsWindow.ofTimeDifferenceAndGrace(
        join.getBeforeMillis(),
        join.getGraceMillis().get()
    ).after(join.getAfterMillis();
} else {
    joinWindows  = JoinWindows.of(join.getBeforeMillis()).after(join.getAfterMillis())
}
```

guozhangwang: I think #7695 would update this.

spena: Right. I need https://github.com/confluentinc/ksql/pull/7695 merged before I make the changes for the new `ofTimeDifferenceAndGrace`.

PR #7688, opened 2021-06-16
fix: stop users from applying the same migration more than once

jzaralim:
### Description 
Adds a check in the apply command that stops users from applying migrations that have already been applied.

### Testing done 
Unit tests, added to the integration test

### Reviewer checklist

### DIFF
--- a/ksqldb-tools/src/main/java/io/confluent/ksql/tools/migrations/commands/ApplyMigrationCommand.java
+++ b/ksqldb-tools/src/main/java/io/confluent/ksql/tools/migrations/commands/ApplyMigrationCommand.java
@@ -208,7 +208,7 @@ public class ApplyMigrationCommand extends BaseCommand {
     LOGGER.info("Loading migration files");
     final List<MigrationFile> migrations;
     try {
-      migrations = loadMigrationsToApply(migrationsDir, minimumVersion);
+      migrations = loadMigrationsToApply(migrationsDir, minimumVersion, config, ksqlClient);
     } catch (MigrationException e) {
       LOGGER.error(e.getMessage());
       return false;
@@ -232,7 +232,9 @@ public class ApplyMigrationCommand extends BaseCommand {
 
   private List<MigrationFile> loadMigrationsToApply(
       final String migrationsDir,
-      final int minimumVersion
+      final int minimumVersion,
+      final MigrationConfig config,
+      final Client ksqlClient
   ) {
     if (version > 0) {
       final Optional<MigrationFile> migration =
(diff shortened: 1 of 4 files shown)


ksqldb-tools/src/main/java/io/confluent/ksql/tools/migrations/commands/ApplyMigrationCommand.java
```diff
@@ -248,6 +253,10 @@ private boolean apply(
           if (migration.getVersion() < minimumVersion) {
             return false;
           }
+          if (MetadataUtil
+              .isVersionMigrated(String.valueOf(migration.getVersion()), config, ksqlClient)) {
+            return false;
```
jzaralim, line 258: Also filters here, in case someone has migrations 1, 2 and 3 and calls `apply -v 3`, `apply -v 1` and then `apply -a`

vcrfxia: I thought we decided not to allow users to apply migrations out of order? As in, once the user has run `apply -v 3` then they should not be allowed to run `apply -v 1`. Do we not currently have a check for this? If not it was probably missed as part of the same gap that missed applying a migration version that has already been applied, and we should conversation about whether we want to start enforcing this or not. (My vote is yes, but a separate PR is fine rather than including that change in this one.)

ksqldb-tools/src/test/java/io/confluent/ksql/tools/migrations/MigrationsTest.java
```diff
@@ -266,6 +266,7 @@ private void shouldApplyMigrations() throws Exception {
     assertThat(applyStatus, is(0));
 
     verifyMigrationsApplied();
+    assertThat(MIGRATIONS_CLI.parse("--config-file", configFilePath, "apply", "-v", "2").runCommand(), is(1));
```
vcrfxia, line 269: Can we add a comment to explain why this check is here? Really it should be its own test but duplicating the rest of this test just for the sake of testing this feels a bit overkill, so I'm fine with leaving it in this test but we should be explicit about what it's testing / why it's here.

ksqldb-tools/src/test/java/io/confluent/ksql/tools/migrations/commands/ApplyMigrationCommandTest.java
```diff
@@ -641,6 +653,54 @@ public void shouldApplyDropConnectorStatement() throws Exception {
...
+  @Test
+  public void shouldNotApplyAppliedMigration() throws Exception {
+    // Given:
+    command = PARSER.parse("-v", "1");
+    createMigrationFile(1, NAME, migrationsDir, COMMAND);
+    givenCurrentMigrationVersion("1");
+    givenAppliedMigration(1, NAME, MigrationState.MIGRATED);
+    final Row row = mock(Row.class);
+    when(row.getString(1)).thenReturn(MigrationState.MIGRATED.name());
+    when(stateQueryResult.get()).thenReturn(Collections.singletonList(row));
```
vcrfxia, line 665: Why is this not part of `givenAppliedMigration()`?

vcrfxia (commented):
Thanks @jzaralim ! LGTM with a few questions inline.

ksqldb-tools/src/main/java/io/confluent/ksql/tools/migrations/commands/ApplyMigrationCommand.java
jzaralim: We don't have that check yet. If we want to enforce in-order only, then this check actually becomes obsolete, since we would strictly require users to apply versions greater than what was last applied. I'll just close this PR in favour of that.

vcrfxia: Yes, I think we should enforce that users only apply migrations in increasing order (not necessarily consecutive). See previous discussion here: https://github.com/confluentinc/ksql/pull/7099#discussion_r584892827
If we add this check into the apply command but not anything to do with validation, then we should be fine from a backwards compatibility standpoint.

PR #7689, opened 2021-06-16
test: set test timeout for BaseApiTest (MINOR)

vcrfxia:
### Description 

We saw a flakey test failure on Jenkins where AuthTest.java somehow got stuck in a loop and caused the build to hit the four-hour timeout. This PR adds a test timeout to all the api tests (including AuthTest.java) to at least prevent bad test runs from timing out the entire build.

### Testing done 

Test-only change.

### Reviewer checklist


vvcephei (approved):
Thanks, @vcrfxia !

PR #7692, opened 2021-06-17
chore: cleanup pom files

mjsax:
### Description 

### Testing done 

### Reviewer checklist

### DIFF
--- a/ksqldb-common/pom.xml
+++ b/ksqldb-common/pom.xml
@@ -147,27 +147,11 @@
             <resource>
                 <directory>src/main/resources</directory>
                 <filtering>true</filtering>
-                <includes>
-                    <include>ksql-version.streamsProperties</include>
-                </includes>
-            </resource>
-
-            <resource>
-                <directory>src/main/resources</directory>
-                <filtering>true</filtering>
             </resource>
 
             <resource>
                 <directory>src/main/resources/ksqldb</directory>
                 <filtering>true</filtering>
-            </resource>
-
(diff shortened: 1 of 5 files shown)


ksqldb-engine/pom.xml
```diff
@@ -197,23 +197,5 @@
                 </configuration>
             </plugin>
         </plugins>
-
```
mjsax, line 200: This and the other 3 modules don't have any resource folder and don't need to get the version. Thus, it seems appropriate to just delete those definitions in the pom (might have be added via c&p?)

guozhangwang:
What this PR is for?

mjsax:
We saw some test failures (I could reproduce locally at some point but not any longer), for which `CliTest` failed because it could not get the version from `AppInfo` -- instead of providing the version string (ie, `"7.0.0-0"`) or similar, the variable in `version.properties` was not replaced and `AppInfo` returned string `"${project.version}"`.

It all still somewhat a mystery...

ksqldb-common/pom.xml
```diff
@@ -144,14 +144,6 @@
         </plugins>
 
         <resources>
-            <resource>
-                <directory>src/main/resources</directory>
-                <filtering>true</filtering>
-                <includes>
-                    <include>ksql-version.streamsProperties</include>
```
mjsax, line 151: This file does not exist any longer

agavra (approved):

PR #7693, opened 2021-06-17
fix: block out of order migrations in migrations tool

jzaralim:
### Description 
Adds a check to make sure the version specified by `apply -v <number>` is greater than the last applied version.

### Testing done 
Unit and integration tests

### Reviewer checklist

### DIFF
--- a/ksqldb-tools/src/main/java/io/confluent/ksql/tools/migrations/commands/ApplyMigrationCommand.java
+++ b/ksqldb-tools/src/main/java/io/confluent/ksql/tools/migrations/commands/ApplyMigrationCommand.java
@@ -201,14 +201,11 @@ public class ApplyMigrationCommand extends BaseCommand {
       final Clock clock
   ) {
     String previous = MetadataUtil.getLatestMigratedVersion(config, ksqlClient);
-    final int minimumVersion = previous.equals(MetadataUtil.NONE_VERSION)
-        ? 1
-        : Integer.parseInt(previous) + 1;
 
     LOGGER.info("Loading migration files");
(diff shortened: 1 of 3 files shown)


ksqldb-tools/src/main/java/io/confluent/ksql/tools/migrations/commands/ApplyMigrationCommand.java
```diff
@@ -240,6 +240,11 @@ private boolean apply(
       if (!migration.isPresent()) {
         throw new MigrationException("No migration file with version " + version + " exists.");
       }
+      if (version < minimumVersion) {
+        throw new MigrationException(
+            "Version must be newer than the last version migrated. Last version migrated was "
+                + (minimumVersion - 1));
```
vcrfxia, line 246: nit: it's not great that `apply()` gets the latest migrated version, transforms that into `minimumVersion` to pass to `loadMigrationsToApply()`, and then `loadMigrationsToApply()` undoes the transformation again in this error message. Maybe we can have `apply()` directly pass the last migrated version and move the logic for computing `minimumVersion` into `loadMigrationsToApply()` instead? Less coupling between the two methods that way.

vcrfxia (approved):
Thanks @jzaralim ! I like this solution -- very clean :)

ksqldb-tools/src/main/java/io/confluent/ksql/tools/migrations/commands/ApplyMigrationCommand.java
jzaralim: updated

PR #7694, opened 2021-06-17
chore: fix build for KafkaConfig deprecations (MINOR)

vcrfxia:
### Description 

The build is currently failing since ksqlDB's KafkaEmbedded class depends on deprecated methods of KafkaConfig that were recently removed in https://github.com/apache/kafka/pull/10872. This PR switches away from the methods to fix the build.

### Testing done 

Will wait for Jenkins to pass before merging.

### Reviewer checklist

### DIFF
--- a/ksqldb-test-util/src/main/java/io/confluent/ksql/test/util/KafkaEmbedded.java
+++ b/ksqldb-test-util/src/main/java/io/confluent/ksql/test/util/KafkaEmbedded.java
@@ -33,6 +33,7 @@ import java.util.Objects;
 import java.util.Properties;
 import java.util.Set;
 import java.util.function.Supplier;
+import kafka.cluster.EndPoint;
 import kafka.server.KafkaConfig;
 import kafka.server.KafkaServer;
 import kafka.utils.TestUtils;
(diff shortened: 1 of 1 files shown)


guozhangwang (approved):

wcarlson5 (approved):
If the tests are passing it looks good to me!

mjsax (approved):

ksqldb-test-util/src/main/java/io/confluent/ksql/test/util/KafkaEmbedded.java
```diff
@@ -104,7 +106,9 @@ String brokerList() {
    * @return the broker list
    */
   String brokerList(final SecurityProtocol securityProtocol) {
-    return kafka.config().hostName() + ":"
+    final EndPoint endPoint = kafka.advertisedListeners().head();
```
ijuma, line 109: This doesn't seem correct. You want to look through the listeners and find the right one. The endpoint port will be correct, so you don't have to call `kafka.boundPort` at all.

vcrfxia: Hey @ijuma thanks for chiming in. To clarify your suggestion: are you saying that `kafka.advertisedListeners()` should already contain endpoints for all the different security protocols, and so this method should simply iterate through them in order to find the right one, rather than trying to construct a new endpoint? If so I've made the change in https://github.com/confluentinc/ksql/pull/7731.

The old behavior was not incorrect for the tests in this codebase since the host was the same for the different endpoints in all cases where this method was called, but I have updated the logic according to your suggestion anyway so this method is more correct going forward.

ijuma: Yes, that's right. Thanks for the follow up PR.

PR #7695, opened 2021-06-17
chore: update QTT framework to support multiple key/value loggers with same schema

spena:
### Description 
This is an improvement from the PR https://github.com/confluentinc/ksql/pull/7464. This PR just works around the QTT framework to support multiple KV loggers detected on QTT changelog topics that use the same schema info. This is required by stream-stream left/outer joins because there is a shared state store that write multiple value formats. QTT uses joins with one format, so we won't have to support different schema yet.

### Testing done 

### Reviewer checklist

### DIFF
--- a/ksqldb-execution/src/main/java/io/confluent/ksql/schema/query/QuerySchemas.java
+++ b/ksqldb-execution/src/main/java/io/confluent/ksql/schema/query/QuerySchemas.java
@@ -18,14 +18,10 @@ package io.confluent.ksql.schema.query;
 import static java.util.Objects.requireNonNull;
 
 import com.google.common.collect.ImmutableSet;
-import com.google.common.collect.Iterables;
 import com.google.errorprone.annotations.Immutable;
 import io.confluent.ksql.execution.runtime.RuntimeBuildContext;
 import io.confluent.ksql.schema.ksql.LogicalSchema;
-import io.confluent.ksql.serde.FormatFactory;
-import io.confluent.ksql.serde.FormatInfo;
 import io.confluent.ksql.serde.KeyFormat;
-import io.confluent.ksql.serde.SerdeFeatures;
 import io.confluent.ksql.serde.ValueFormat;
 import java.util.Collections;
 import java.util.HashMap;
@@ -61,16 +57,8 @@ public final class QuerySchemas {
   // Maps topic name -> (map of key/value flag -> set of logger name prefixes)
   private final Map<String, Map<Boolean, Set<String>>> topicsToLoggers = new HashMap<>();
 
-  private static final boolean IS_KEY = true;
-  private static final boolean IS_VALUE = false;
-
-  private static final SchemaInfo EMPTY_SCHEMA = new SchemaInfo(
-      LogicalSchema.builder().build(), Optional.empty(), Optional.empty()
-  );
-
-  private static final ValueFormat NONE_VALUE_FORMAT = ValueFormat.of(
-      FormatInfo.of(FormatFactory.NONE.name()), SerdeFeatures.of()
-  );
+  private static final boolean IS_KEY_SCHEMA = true;
+  private static final boolean IS_VALUE_SCHEMA = false;
 
   /**
    * Called when creating a key serde.
@@ -137,79 +125,51 @@ public final class QuerySchemas {
     return Collections.unmodifiableMap(topicsToLoggers);
   }
 
-  public TopicFormatsInfo getTopicFormatsInfo(final String topicName) {
+  /**
+   * Returns all different schemas and serde formats that the {@code topicName} is using. The
+   * {@link MultiSchemaInfo} contains a list of key/value schemas and serde formats that were
+   * tracked for the specified topic.
+   * </p>
+   * Key and/or value schemas may be empty if no schemas were detected.
+   */
+  public MultiSchemaInfo getTopicInfo(final String topicName) {
+    final Set<SchemaInfo> keySchemasInfo = getTopicSchemas(topicName, IS_KEY_SCHEMA);
+    final Set<SchemaInfo> valueSchemasInfo = getTopicSchemas(topicName, IS_VALUE_SCHEMA);
+
+    return new MultiSchemaInfo(keySchemasInfo, valueSchemasInfo);
+  }
+
+  /**
+   * Returns a list of schemas and serde formats detected on the specified topicName.
+   * The isKeySchema is a boolean value that specifies whether to look at the key schema (True)
+   * or at the value schema (False).
+   * </p>
+   * Each schema information is tracked by the loggers the topicName has. The logger is
+   * a changelog topic. Some topics may have multiple schemas and serde formats if the topicName
+   * is used with stream-stream joins and foreign key joins.
+   */
+  private Set<SchemaInfo> getTopicSchemas(final String topicName,
+                                          final boolean isKeySchema) {
+    // Look at the different changelog topics that tne topicName uses. There might be multiple
+    // changelog topics if the topicName was used by joins, which contains internal state stores
+    // with a changelog per store.
     final Map<Boolean, Set<String>> kvLoggerNames = topicsToLoggers.get(topicName);
     if (kvLoggerNames == null) {
       throw new IllegalArgumentException("Unknown topic: " + topicName);
     }
 
-    final Map<SchemaInfo, Set<String>> keyLoggerSchemas = getLoggerSchemas(
-        kvLoggerNames,
-        IS_KEY
-    );
-
-    final Map<SchemaInfo, Set<String>> valueLoggerSchemas = getLoggerSchemas(
-        kvLoggerNames,
-        IS_VALUE
-    );
-
-    final Set<SchemaInfo> keySchemaInfo = Optional.of(keyLoggerSchemas.keySet())
-        .orElseGet(() -> ImmutableSet.of(EMPTY_SCHEMA));
-    final Set<SchemaInfo> valueSchemaInfo = Optional.of(valueLoggerSchemas.keySet())
-        .orElseGet(() -> ImmutableSet.of(EMPTY_SCHEMA));
-
-    final Set<KeyFormat> keyFormats = keySchemaInfo.stream()
-        .map(s -> s.keyFormat)
-        .filter(Optional::isPresent)
-        .map(Optional::get)
-        .collect(Collectors.toSet());
-
-    if (keyFormats.size() != 1) {
-      final String result = keyFormats.size() == 0 ? "Zero" : "Multiple";
-
-      throw new IllegalStateException(result + " key formats registered for topic."
-          + System.lineSeparator()
-          + "topic: " + topicName
-          + "formats: " + keyFormats.stream().map(KeyFormat::getFormat)
-          .sorted().collect(Collectors.toList())
-          + "loggers: " + keyLoggerSchemas.values()
-      );
-    }
-
-    final Set<ValueFormat> valueFormats = valueSchemaInfo.stream()
-        .map(s -> s.valueFormat)
-        .filter(Optional::isPresent)
-        .map(Optional::get)
-        .collect(Collectors.toSet());
-
-    if (valueFormats.size() > 1) {
-      throw new IllegalStateException("Multiple value formats registered for topic."
-          + System.lineSeparator()
-          + "topic: " + topicName
-          + "formats: " + valueFormats.stream().map(ValueFormat::getFormat)
-          .sorted().collect(Collectors.toList())
-          + "loggers: " + valueLoggerSchemas.values()
-      );
-    }
-
-    final KeyFormat keyFormat = Iterables.getOnlyElement(keyFormats);
-    final ValueFormat valueFormat = !valueFormats.isEmpty()
-        ? Iterables.getOnlyElement(valueFormats)
-        : NONE_VALUE_FORMAT;
-
-    return new TopicFormatsInfo(keyFormat, valueFormat);
-  }
-
-  private Map<SchemaInfo, Set<String>> getLoggerSchemas(
-      final Map<Boolean, Set<String>> kvLoggerNames, final boolean isKeyLogger) {
-
     final Map<SchemaInfo, Set<String>> schemaToLoggers = new HashMap<>();
-    for (final String loggerName : kvLoggerNames.getOrDefault(isKeyLogger, ImmutableSet.of())) {
-      final SchemaInfo keyInfo = loggerToSchemas.getOrDefault(loggerName, EMPTY_SCHEMA);
-      schemaToLoggers.computeIfAbsent(keyInfo, k -> new HashSet<>()).add(loggerName);
+
+    // Look at the changelogs linked to the key or value serde, and get all schema and formats
+    // detected.
+    for (final String loggerName : kvLoggerNames.getOrDefault(isKeySchema, ImmutableSet.of())) {
+      final SchemaInfo keyInfo = loggerToSchemas.get(loggerName);
+      if (keyInfo != null) {
+        schemaToLoggers.computeIfAbsent(keyInfo, k -> new HashSet<>()).add(loggerName);
+      }
     }
 
-    return schemaToLoggers;
+    return schemaToLoggers.keySet();
   }
 
   @Override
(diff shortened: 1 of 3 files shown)


ksqldb-execution/src/main/java/io/confluent/ksql/schema/query/QuerySchemas.java
```diff
@@ -118,47 +133,83 @@ public synchronized void trackSerdeOp(
...
 
-  /**
-   * Builds complete SchemaInfo for a topic:
-   */
-  public SchemaInfo getTopicInfo(final String topicName) {
+  Map<String, Map<Boolean, Set<String>>> getTopicsToLoggers() {
+    return Collections.unmodifiableMap(topicsToLoggers);
+  }
+
+  public TopicFormatsInfo getTopicFormatsInfo(final String topicName) {
```
spena, line 140: The only  =usage of `getTopicInfo()` when called was to obtain key/value formats. I renamed to reflect the right usage. Also, in case two loggers with value formats are detected, it is simpler to combine them if the loggers use the same key/value format. With the old method, the formats were the same but the schema was different.

vcrfxia: I don't love that we've added a new class that's pretty much identical to an existing one. The only difference between the existing SchemaInfo and the new TopicFormatsInfo is that there's no schema and the formats are no longer optional. If not for the latter change, SchemaInfo could extend TopicFormatsInfo which would be less duplication of code. 

The current merge() method for SchemaInfo doesn't actually verify that the schemas are the same, which admittedly is odd but I think it also means it's fine if the different loggers have different schemas? What's the problem with using the existing SchemaInfo rather than introducing a new class?

ksqldb-execution/src/main/java/io/confluent/ksql/schema/query/QuerySchemas.java
```diff
@@ -118,47 +133,83 @@ public synchronized void trackSerdeOp(
...
-      );
+    final KeyFormat keyFormat = Iterables.getOnlyElement(keyFormats);
+    final ValueFormat valueFormat = !valueFormats.isEmpty()
+        ? Iterables.getOnlyElement(valueFormats)
+        : NONE_VALUE_FORMAT;
+
+    return new TopicFormatsInfo(keyFormat, valueFormat);
+  }
+
+  private Map<SchemaInfo, Set<String>> getLoggerSchemas(
```
vcrfxia, line 203: Can we add a short javadoc to this? It wasn't clear to me from the name what the method does -- I had to read the code to find out.

ksqldb-functional-tests/src/main/java/io/confluent/ksql/test/tools/TopicInfoCache.java
```diff
@@ -118,7 +119,21 @@ public TopicInfoCache(
       return Optional.empty();
     }
 
-    return Optional.of(cache.getUnchecked(topicName));
+    final TopicInfo topicInfo = cache.getUnchecked(topicName);
+
+    // The problem I am fixing is that sometimes the first record found in a topic has a null value.
+    // This causes the QuerySchemas#getTopicInfo() to return a EMPTY_SCHEMA for the value.
+    // When that info is obtained from the load function, then it is kept in the cache for some
+    // time. I assume that the null value is just something that may not
```
vcrfxia, line 127: Thanks for the detailed comment. I'm still a bit confused about what problem this is solving, though. Is there a race condition of some sort? Is the issue nondeterministic or would tests fail deterministically if we didn't refresh the cache in this way?

ksqldb-execution/src/test/java/io/confluent/ksql/schema/query/QuerySchemasTest.java
```diff
@@ -0,0 +1,299 @@
...
+        // When
+        final QuerySchemas.TopicFormatsInfo info = querySchemas.getTopicFormatsInfo("topic1");
+
+        // Then
+        assertThat(info.keyFormat(), is(JSON_KEY_FORMAT));
+        assertThat(info.valueFormat(), is(JSON_VALUE_FORMAT));
+    }
+
+    @Test
+    public void shouldReturnFormatsInfoWithEmptyValueFormatIsNoValueSchemaIsFound() {
```
vcrfxia, line 181:
```suggestion
    public void shouldReturnFormatsInfoWithEmptyValueFormatIfNoValueSchemaIsFound() {
```

ksqldb-execution/src/test/java/io/confluent/ksql/schema/query/QuerySchemasTest.java
```diff
@@ -0,0 +1,299 @@
...
+        // When
+        final QuerySchemas.TopicFormatsInfo info = querySchemas.getTopicFormatsInfo("topic1");
+
+        // Then
+        assertThat(info.keyFormat(), is(JSON_KEY_FORMAT));
+        assertThat(info.valueFormat(), is(NONE_VALUE_FORMAT));
+    }
+
+    @Test
+    public void shouldThrowFormatsInfoIfNoKeySchemaIsFound() {
```
vcrfxia, line 195:
```suggestion
    public void shouldThrowIfNoKeySchemaIsFound() {
```

ksqldb-execution/src/test/java/io/confluent/ksql/schema/query/QuerySchemasTest.java
```diff
@@ -0,0 +1,299 @@
...
+        final Exception e = assertThrows(
+            IllegalArgumentException.class,
+            () -> querySchemas.getTopicFormatsInfo("t1"));
+
+        // Then
+        assertThat(e.getMessage(), is("Unknown topic: t1"));
+    }
+
+    @Test
+    public void shouldThrowIfMultipleKeyLoggersSchemasAreFound() {
```
vcrfxia, line 228:
```suggestion
    public void shouldThrowIfMultipleKeyFormatsAreFound() {
```

ksqldb-execution/src/test/java/io/confluent/ksql/schema/query/QuerySchemasTest.java
```diff
@@ -0,0 +1,299 @@
...
+        assertThat(e.getMessage(),
+            containsString("topic: topic1"));
+        assertThat(e.getMessage(),
+            containsString("formats: [AVRO, JSON]"));
+        assertThat(e.getMessage(),
+            containsString("loggers: [[K_LOGGER_1], [K_LOGGER_2]]"));
+    }
+
+    @Test
+    public void shouldThrowIfMultipleValueLoggersSchemasAreFound() {
```
vcrfxia, line 252:
```suggestion
    public void shouldThrowIfMultipleValueFormatsAreFound() {
```

ksqldb-execution/src/test/java/io/confluent/ksql/schema/query/QuerySchemasTest.java
```diff
@@ -0,0 +1,299 @@
...
+        assertThat(e.getMessage(),
+            containsString("formats: [AVRO, JSON]"));
+        assertThat(e.getMessage(),
+            containsString("loggers: [[V_LOGGER_1], [V_LOGGER_2]"));
+    }
+
+    @Test
+    public void shouldThrowIfKeySchemaIsNotFound() {
+        // Given
+        querySchemas.trackSerdeOp("topic1", IS_KEY, "K_LOGGER_1");
```
vcrfxia, line 280: I don't understand this test. Isn't this a key schema? The test's name suggests that it's testing the case where there are no key schemas, though.

Also, doesn't this test duplicate `shouldThrowFormatsInfoIfNoKeySchemaIsFound()` above?

ksqldb-execution/src/test/java/io/confluent/ksql/schema/query/QuerySchemasTest.java
```diff
@@ -0,0 +1,299 @@
+/*
+ * Copyright 2019 Confluent Inc.
```
vcrfxia, line 2: nit: year

vcrfxia (commented):
Thanks @spena . Left some questions inline. Let me know if you want to sync about these offline to make the next review go faster.

ksqldb-execution/src/main/java/io/confluent/ksql/schema/query/QuerySchemas.java
spena: I did another change here. i didn't like the new class either. I created a `MultiSchemaInfo` now which it just contains a list of SchemaInfo for keys and another for values. This `getTopicInfo` will not throw exceptions if the schemas and/or formats are different.

ksqldb-execution/src/main/java/io/confluent/ksql/schema/query/QuerySchemas.java
```diff
@@ -118,47 +133,83 @@ public synchronized void trackSerdeOp(
...
-    if (valueTopicNames.size() != 1) {
-      throw new IllegalStateException("Multiple value logger names registered for topic."
+    final Set<ValueFormat> valueFormats = valueSchemaInfo.stream()
+        .map(s -> s.valueFormat)
+        .filter(Optional::isPresent)
+        .map(Optional::get)
+        .collect(Collectors.toSet());
+
+    if (valueFormats.size() > 1) {
+      throw new IllegalStateException("Multiple value formats registered for topic."
```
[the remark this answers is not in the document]
spena: I added the link as a reference.

ksqldb-execution/src/main/java/io/confluent/ksql/schema/query/QuerySchemas.java
spena: Done.

ksqldb-functional-tests/src/main/java/io/confluent/ksql/test/tools/TopicInfoCache.java
spena: I just removed it. Seems there's no issue with it. I tested it and it worked without it.

spena:
Thanks @vcrfxia for the comments.I made the changes.

ksqldb-execution/src/main/java/io/confluent/ksql/schema/query/QuerySchemas.java
```diff
@@ -118,47 +121,55 @@ public synchronized void trackSerdeOp(
...
+    return Collections.unmodifiableMap(topicsToLoggers);
+  }
+
   /**
-   * Builds complete SchemaInfo for a topic:
+   * Returns all different schemas and serde formats that the {@code topicName} is using. The
+   * {@link MultiSchemaInfo} contains a list of key/value schemas and serde formats that were
+   * tracked for the specified topic.
+   * </p>
+   * Key and/or value schemas may be empty if no schemas were detected.
```
vcrfxia, line 133: Do you mean that the formats may be empty? Or are there situations where the schema might be empty too?

ksqldb-execution/src/main/java/io/confluent/ksql/schema/query/QuerySchemas.java
```diff
@@ -118,47 +121,55 @@ public synchronized void trackSerdeOp(
...
+    return new MultiSchemaInfo(keySchemasInfo, valueSchemasInfo);
+  }
+
+  /**
+   * Returns a list of schemas and serde formats detected on the specified topicName.
+   * The isKeySchema is a boolean value that specifies whether to look at the key schema (True)
+   * or at the value schema (False).
+   * </p>
+   * Each schema information is tracked by the loggers the topicName has. The logger is
+   * a changelog topic. Some topics may have multiple schemas and serde formats if the topicName
```
vcrfxia, line 148: Is it true that all loggers are changelog topics? That's news to me if so.

ksqldb-execution/src/main/java/io/confluent/ksql/schema/query/QuerySchemas.java
```diff
@@ -118,47 +121,55 @@ public synchronized void trackSerdeOp(
...
-          + System.lineSeparator()
-          + "topic: " + topicName
-          + "loggers: " + valueTopicNames
-      );
-    }
+    final Map<SchemaInfo, Set<String>> schemaToLoggers = new HashMap<>();
 
-    final SchemaInfo keyInfo = loggerToSchemas.get(Iterables.getOnlyElement(keyLoggerNames));
-    final SchemaInfo valueInfo = loggerToSchemas.get(Iterables.getOnlyElement(valueTopicNames));
-    if (keyInfo == null || valueInfo == null) {
```
vcrfxia, line 150: Was this check removed intentionally? It still seems good to throw if either the key or value info is missing.

ksqldb-execution/src/main/java/io/confluent/ksql/schema/query/QuerySchemas.java
```diff
@@ -118,47 +121,55 @@ public synchronized void trackSerdeOp(
...
-      throw new IllegalStateException("Incomplete schema info for topic."
-          + System.lineSeparator()
-          + "topic: " + topicName
-          + "keyInfo: " + keyInfo
-          + "valueInfo: " + valueInfo
-      );
+    // Look at the changelogs linked to the key or value serde, and get all schema and formats
+    // detected.
+    for (final String loggerName : kvLoggerNames.getOrDefault(isKeySchema, ImmutableSet.of())) {
+      final SchemaInfo keyInfo = loggerToSchemas.get(loggerName);
```
vcrfxia, line 166: Why is this called `keyInfo`? It might also represent info about the value instead, right?

ksqldb-execution/src/main/java/io/confluent/ksql/schema/query/QuerySchemas.java
```diff
@@ -199,6 +210,46 @@ private void trackSerde(
     });
   }
 
+  public static class MultiSchemaInfo {
+    private final Set<SchemaInfo> keySchemas;
+    private final Set<SchemaInfo> valueSchema;
```
vcrfxia, line 215: nit: inconsistent naming compared to `keySchemas` in the line above.

vcrfxia (approved):
Thanks @spena . Left a few more questions inline but LGTM otherwise. Giving the green tick now to unblock subsequent work.

ksqldb-execution/src/main/java/io/confluent/ksql/schema/query/QuerySchemas.java
spena: Both. The QTT can't track what the format and/or schema is because a NULL value inserted in the value schema side. The left/outer joins has an internal `put(value, null)` as the first value in some cases, so QTT fails because it does not know what the format is.

ksqldb-execution/src/main/java/io/confluent/ksql/schema/query/QuerySchemas.java
spena: I'm not 100%. At least while I was debugging I was seeing topic names that end up with changelog-topic. But I might have missed some or just pressing F9 to quickly while debugging.

ksqldb-execution/src/main/java/io/confluent/ksql/schema/query/QuerySchemas.java
spena: It's not necessary anymore. Now  the key and value info are a `Set` which maybe empty. The `getTopicSchemas` will never return null.
```
final Set<SchemaInfo> keySchemasInfo = getTopicSchemas(topicName, IS_KEY_SCHEMA);
final Set<SchemaInfo> valueSchemasInfo = getTopicSchemas(topicName, IS_VALUE_SCHEMA);
```

ksqldb-execution/src/main/java/io/confluent/ksql/schema/query/QuerySchemas.java
spena: I'll change it.

ksqldb-execution/src/main/java/io/confluent/ksql/schema/query/QuerySchemas.java
vcrfxia: Right but previously the test framework ensured that there is always key and value info for every topic, i.e., that you never have a topic with key info but no value info or vice versa. The version of the code does not enforce this anymore. Do you not think this is worth enforcing still?

spena: It is now enforced in the `TopicInfoCache`. It is now up to the caller to check now, which up to now only `TopicInfoCache` calls it. The idea now is that `getTopicInfo` just returns schema and formats found without throwing exceptions (except when the topic is unknown).

vcrfxia: Oh I see, it's being enforced indirectly through `TopicInfoCache` since `TopicInfoCache` requires that a key format and at least one value format exist, and those formats come from the schemas. I didn't realize the second part until now. Thanks for clarifying!

spena:
All tests passed. There's one check that failed related to creating, which I assume is related to docker?

Error in sh step, with arguments mvn --batch-mode -Pjenkins clean install dependency:analyze site validate -U -Ddocker.registry=368821881613.dkr.ecr.us-west-2.amazonaws.com/ -Ddocker.upstream-registry=368821881613.dkr.ecr.us-west-2.amazonaws.com/ -DBUILD_NUMBER=6 -DGIT_COMMIT=6018ff3f30 -Ddocker.tag=dev-PR-7695-6 -Ddocker.upstream-tag=master-latest .

I'll merge the PR as it looks good and the above issue is unrelated.

PR #7709, opened 2021-06-22
fix: enable time unit functions for interpreter

### LINKED ISSUE #7512: Timestamp arithmetic functions do not work with pull query interpreter
**Describe the bug**
The following query fails:
```
SELECT  TIMESTAMPADD(MINUTES, 15,  TS) FROM TableA WHERE DUMMY_FIELD=1;
Unexpected error generating code for expression: TIMESTAMPADD(MINUTES, 15, 'k')
```
This is because time units (MINUTE, SECOND etc) are not implemented in TermCompiler.java yet: https://github.com/confluentinc/ksql/blob/0aeb57f41f6df58fb4788ea1d12ea6a9b8fc5191/ksqldb-execution/src/main/java/io/confluent/ksql/execution/interpreter/TermCompiler.java#L297

The error message is also confusing, since it doesn't actually print out what the error is https://github.com/confluentinc/ksql/blob/0aeb57f41f6df58fb4788ea1d12ea6a9b8fc5191/ksqldb-execution/src/main/java/io/confluent/ksql/execution/interpreter/InterpretedExpressionFactory.java#L69

jzaralim:
### Description 
Fixes #7512. Implements `visitIntervalUnit ` in `TermCompiler`.

### Testing done 
Added a unit test

### Reviewer checklist

### DIFF
--- a/ksqldb-execution/src/main/java/io/confluent/ksql/execution/interpreter/terms/LiteralTerms.java
+++ b/ksqldb-execution/src/main/java/io/confluent/ksql/execution/interpreter/terms/LiteralTerms.java
@@ -20,6 +20,7 @@ import io.confluent.ksql.schema.ksql.types.SqlType;
 import io.confluent.ksql.schema.ksql.types.SqlTypes;
 import java.math.BigDecimal;
 import java.sql.Timestamp;
+import java.util.concurrent.TimeUnit;
 
 @SuppressWarnings("checkstyle:ClassDataAbstractionCoupling")
 public final class LiteralTerms {
@@ -52,6 +53,10 @@ public final class LiteralTerms {
 
   public static Term of(final Timestamp value) {
     return new TimestampTermImpl(value);
+  }
+
+  public static Term of(final TimeUnit value) {
+    return new IntervalUnitTermImpl(value);
   }
 
   public static NullTerm ofNull() {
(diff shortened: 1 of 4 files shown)


ksqldb-execution/src/main/java/io/confluent/ksql/execution/interpreter/terms/LiteralTerms.java
```diff
@@ -209,4 +214,23 @@ public SqlType getSqlType() {
...
+    }
+
+    @Override
+    public Object getValue(final TermEvaluationContext context) {
+      return value;
+    }
+
+    @Override
+    public SqlType getSqlType() {
+      return null;
```
AlanConfluent, line 233: Does this not have a SqlType?  It doesn't seem like it since it's kind of a modifier rather than a data value, but just checking

AlanConfluent (approved):
Looks good.  You can also add to `ExpressionEvaluatorParityTest` if you want to ensure they both return the same results for same expressions.

ksqldb-execution/src/main/java/io/confluent/ksql/execution/interpreter/terms/LiteralTerms.java
jzaralim: It does not. The only place it can exist is in a function (parser would reject everything else).

PR #7710, opened 2021-06-22
chore: bump protobuf version to 3.17.0

jzaralim:
### Description 
needed after this change in schema registry: https://github.com/confluentinc/schema-registry/pull/1865

### Testing done 
Builds locally

### Reviewer checklist
- [ ] Ensure docs are updated if necessary. (eg. if a user visible feature is being added or changed).
- [ ] Ensure relevant issues are linked (description should include text like "Fixes #<issue number>")

PR #7711, opened 2021-06-22
chore: bump protobuf version to 3.17.0

jzaralim:
### Description 
protobuf version was bumped in schema registry, so we're seeing test failures with errors such as

```
java.lang.NoSuchMethodError: com.google.protobuf.DescriptorProtos$FieldDescriptorProto$Builder.setProto3Optional(Z)
```

meaning that there are some functions in 3.17.0 that SR uses that the current version doesnt have. 

see #7710 

### Testing done 
Master build passes locally

### Reviewer checklist

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -104,7 +104,7 @@
         <csv.version>1.4</csv.version>
         <lang3.version>3.3.1</lang3.version>
         <guava.version>24.1.1-jre</guava.version>
-        <protobuf.version>3.11.1</protobuf.version>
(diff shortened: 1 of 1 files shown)


vcrfxia (approved):
Thanks @jzaralim ! Minor comment: It'd be good to add a note in the PR description about _why_ we also need to bump the version. Is it because the build is failing? Some tests are failing? Copying in the error message (if you have it handy) makes it searchable so if someone else is hitting the same issues/failures then they'll discover this PR :)

PR #7716, opened 2021-06-23
chore: make ksql.suppress.enabled immutable (MINOR)

vcrfxia:
### Description 

The `ksql.suppress.enabled` config controls whether `EMIT FINAL` is allowed on a particular ksqlDB cluster or not. Users should not be able to change the value configured for a given server as part of their query (or session) properties. This PR implements this fix.

### Testing done 

Manual sanity-check.

### Reviewer checklist


nateab (approved):
LGTM!

tolgadur (approved):

vcrfxia:
Closing this PR since the use case the PR was trying to solve was mitigated elsewhere. I don't think it makes sense to special-case `ksql.suppress.enabled` compared to the other feature flags. Maybe we should add all our feature flags to this list but that's a separate (larger) discussion.

PR #7717, opened 2021-06-23
fix: Fixes slow streams shut down bug which made tests flaky

AlanConfluent:
### Description 
Fixes flaky test `PullQueryRoutingFunctionalTest`.  The symptom of the flakiness was the failure:

```
io.confluent.ksql.rest.integration.PullQueryRoutingFunctionalTest.shouldFilterLaggyServers
Error Message
error msg: Timed out while waiting for a previous command to execute. command sequence number: 14
Stacktrace
java.lang.UnsupportedOperationException: error msg: Timed out while waiting for a previous command to execute. command sequence number: 14
	at io.confluent.ksql.rest.integration.PullQueryRoutingFunctionalTest.cleanUp(PullQueryRoutingFunctionalTest.java:287)
```

After a lot of investigation, I noticed that `SequenceNumberFutureStore.completeFuturesUpToAndIncludingSequenceNumber` was never being called in rare scenarios, causing the future to never complete.  It turns out that the `CommandStore` was being closed, which was due to the line here in `CommandRunner`: https://github.com/confluentinc/ksql/blob/master/ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/computation/CommandRunner.java#L469.

This was ultimately caused by an exception:
```
[2021-06-22 18:08:13,927] WARN query has not terminated even after close. This may happen when streams threads are hung. State: PENDING_SHUTDOWN (io.confluent.ksql.util.QueryMetadataImpl:315)
java.lang.IllegalStateException: Cannot clean up while running.
	at org.apache.kafka.streams.KafkaStreams.cleanUp(KafkaStreams.java:1439)
	at io.confluent.ksql.util.QueryMetadataImpl.doClose(QueryMetadataImpl.java:338)
	at io.confluent.ksql.util.QueryMetadataImpl.close(QueryMetadataImpl.java:329)
```

This PR avoids the cleanup if the `KafkaStreams` doesn't close correctly.

### Testing done 
Ran `PullQueryRoutingFunctionalTest` a zillion times both before the change, reproducing the bug, and after it, not reproducing it.

### Reviewer checklist

### DIFF
--- a/ksqldb-engine/src/main/java/io/confluent/ksql/util/QueryMetadataImpl.java
+++ b/ksqldb-engine/src/main/java/io/confluent/ksql/util/QueryMetadataImpl.java
@@ -337,11 +337,11 @@ public class QueryMetadataImpl implements QueryMetadata {
 
   void doClose(final boolean cleanUp) {
     closed = true;
-    boolean closed = closeKafkaStreams();
+    final boolean closedKafkaStreams = closeKafkaStreams();
 
-    if (cleanUp && closed) {
+    if (cleanUp && closedKafkaStreams) {
       kafkaStreams.cleanUp();
-    } else if (!closed) {
+    } else if (!closedKafkaStreams) {
       LOG.warn("Query has not successfully closed, skipping cleanup");
     }
   }
--- a/ksqldb-engine/src/main/java/io/confluent/ksql/util/QueryMetadataImpl.java
+++ b/ksqldb-engine/src/main/java/io/confluent/ksql/util/QueryMetadataImpl.java
@@ -308,7 +308,10 @@ public class QueryMetadataImpl implements QueryMetadata {
     kafkaStreams.setStateListener((b, a) -> listener.onStateChange(this, b, a));
   }
 
-  protected void closeKafkaStreams() {
+  /**
+   * @return If the close was successful
+   */
+  protected boolean closeKafkaStreams() {
     if (initialized) {
       kafkaStreams.close(closeTimeout);
       if (!getState().equals(State.NOT_RUNNING)) {
@@ -316,8 +319,10 @@ public class QueryMetadataImpl implements QueryMetadata {
             "query has not terminated even after close. "
                 + "This may happen when streams threads are hung. State: " + getState()
         );
+        return false;
       }
     }
+    return true;
   }
 
   /**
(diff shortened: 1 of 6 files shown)


agavra (approved):
nice! thanks @AlanConfluent. Should we add a unit test for `QuerymetadataImpl` as this seems like a bug that could hit us in production as well, right?

agavra:
Also do you think we can remove retries for https://github.com/confluentinc/ksql/pull/7343 or is that too risky 😉 ?

AlanConfluent:
> nice! thanks @AlanConfluent. Should we add a unit test for `QuerymetadataImpl` as this seems like a bug that could hit us in production as well, right?

Added a unit test for the non-cleanup case and there was already one for the cleanup case.

It could certainly hit production as well, so I agree, another test is good.

AlanConfluent:
> Also do you think we can remove retries for #7343 or is that too risky 😉 ?

From reading that one, I'm not sure that they are related since I believe it just takes time for things to get materialized.

I was looking at re-enabling PullQuerySingleNodeFunctionalTest.restoreAfterClearState since I think it might be related, but I want to give that one some additional testing to be sure.

vvcephei:
Thanks, @AlanConfluent !

Am I right in thinking that if we skip cleanup, then there's no other process that will clean up the state directory later?

AlanConfluent:
> Thanks, @AlanConfluent !
> 
> Am I right in thinking that if we skip cleanup, then there's no other process that will clean up the state directory later?

I believe that's correct.  Not sure if there's a better solution to this than to skip it entirely.  This happens quite rarely in practice from what I can tell, but when it does happen at the moment, it kills the command runner and still doesn't clean up the state store.  This at least avoids killing the command runner.  We could always increase the timeout on the close operation, though I don't know how often this is hit in practice.  At least we log a warning and can search for it.

vvcephei:
sgtm. Thanks, @AlanConfluent !

Do you think it would make sense to file a follow-on ticket to add a scheduled job to clean up after queries that are no longer running?

vvcephei (approved):

AlanConfluent:
> sgtm. Thanks, @AlanConfluent !
> 
> Do you think it would make sense to file a follow-on ticket to add a scheduled job to clean up after queries that are no longer running?

Yes, filed here: https://github.com/confluentinc/ksql/issues/7720

PR #7719, opened 2021-06-23
fix: fixed nondeterminism in UdfIndex

Sullivan-Patrick:
### Description 
When working on `LEAST` and `GREATEST` functions, I noticed there was some nondeterminism inside the `UdfIndex` class, specifically with the `order` field, which was intended to give comparably equal functions a final ordering based in the way they were added inside the Udf's class. This wasn't working in practice, and was leading to equally comparable methods that matched a query being selected arbitrarily. For example, when calling `LEAST(INT, LONG)`, several UDFs could be selected: `LEAST(LONG, LONG...)`, `LEAST(DOUBLE, DOUBLE...)`, `LEAST(DECIMAL, DECIMAL)`.

To fix this, `UdfIndex.Node.compare()` was changed to no longer consider `order`. As this change implies we are no longer able to select a matching function if multiple exist, `getFunction` now throws a `KsqlException` if there is more than one equally comparable method that would match the query.

A minor change that was made on top of this is adding to `UdfIndex.Node.compare()` the preference of picking methods with fewer generics as a final comparison.

BREAKING CHANGE: Existing queries that relied on vague implicit casting will not be started after an upgrade, and new queries that rely on vague implicit casting will be rejected.  For example, `foo(INT, INT)` will not be able to resolve against two underlying function signatures of `foo(BIGINT, BIGINT)` and `foo(DOUBLE, DOUBLE)`. Calling a function whose only parameter is variadic with an explicit null will also result in the call being rejected as vague.

### Testing done 
Added to UDFIndexTest negative tests which ensure that queries which would rely on vague implicit casting fail properly.

Some tests that relied on vague implicit casting may need to be removed, this is worth discussion.

### Reviewer checklist

### DIFF
--- a/ksqldb-common/src/main/java/io/confluent/ksql/function/UdfIndex.java
+++ b/ksqldb-common/src/main/java/io/confluent/ksql/function/UdfIndex.java
@@ -58,8 +58,8 @@ import org.slf4j.LoggerFactory;
  *       arguments.</li>
  *   <li>If two methods exist that match given the above rules, return the
  *       method with fewer generic arguments.</li>
- *   <li>If two methods exist that match given the above rules, an exception
- *       is thrown due to a vague function call</li>
+ *   <li>If two methods exist that match given the above rules, the function
+ *       call is ambiguous and an exception is thrown.</li>
  * </ul>
  */
 public class UdfIndex<T extends FunctionSignature> {
@@ -145,27 +145,27 @@ public class UdfIndex<T extends FunctionSignature> {
   }
 
   T getFunction(final List<SqlArgument> arguments) {
-    final List<Node> candidates = new ArrayList<>();
 
     // first try to get the candidates without any implicit casting
-    Optional<T> candidate = findMatchingCandidate(candidates, arguments, false);
+    Optional<T> candidate = findMatchingCandidate(arguments, false);
     if (candidate.isPresent()) {
       return candidate.get();
     } else if (!supportsImplicitCasts) {
       throw createNoMatchingFunctionException(arguments);
     }
-    candidates.clear();
 
-    //if non were found (candidate isn't present) try again with implicit casting
-    candidate = findMatchingCandidate(candidates, arguments, true);
+    // if none were found (candidate isn't present) try again with implicit casting
+    candidate = findMatchingCandidate(arguments, true);
     if (candidate.isPresent()) {
       return candidate.get();
     }
     throw createNoMatchingFunctionException(arguments);
   }
 
-  private Optional<T> findMatchingCandidate(final List<Node> candidates,
+  private Optional<T> findMatchingCandidate(
       final List<SqlArgument> arguments, final boolean allowCasts) {
+
+    final List<Node> candidates = new ArrayList<>();
 
     getCandidates(arguments, 0, root, candidates, new HashMap<>(), allowCasts);
     candidates.sort(Node::compare);
--- a/ksqldb-common/src/main/java/io/confluent/ksql/function/UdfIndex.java
+++ b/ksqldb-common/src/main/java/io/confluent/ksql/function/UdfIndex.java
@@ -56,8 +56,10 @@ import org.slf4j.LoggerFactory;
  *   <li>If two methods exist that match given the above rules, and both
  *       have variable arguments, return the method with the more non-variable
  *       arguments.</li>
- *   <li>If two methods exist that match only null values, return the one
- *       that was added first.</li>
+ *   <li>If two methods exist that match given the above rules, return the
+ *       method with fewer generic arguments.</li>
+ *   <li>If two methods exist that match given the above rules, an exception
+ *       is thrown due to a vague function call</li>
  * </ul>
  */
 public class UdfIndex<T extends FunctionSignature> {
@@ -112,7 +114,6 @@ public class UdfIndex<T extends FunctionSignature> {
       );
     }
 
-    final int order = allFunctions.size();
 
     Node curr = root;
     Node parent = curr;
@@ -125,14 +126,14 @@ public class UdfIndex<T extends FunctionSignature> {
     if (function.isVariadic()) {
       // first add the function to the parent to address the
       // case of empty varargs
-      parent.update(function, order);
+      parent.update(function);
 
       // then add a new child node with the parameter value type
       // and add this function to that node
       final ParamType varargSchema = Iterables.getLast(parameters);
       final Parameter vararg = new Parameter(varargSchema, true);
       final Node leaf = parent.children.computeIfAbsent(vararg, ignored -> new Node());
-      leaf.update(function, order);
+      leaf.update(function);
 
       // add a self referential loop for varargs so that we can
       // add as many of the same param at the end and still retrieve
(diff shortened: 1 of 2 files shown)


ksqldb-common/src/test/java/io/confluent/ksql/function/UdfIndexTest.java
```diff
@@ -562,35 +563,6 @@ public void shouldFindNonVarargWithPartialNullValues() {
...
-    givenFunctions(
-        function(EXPECTED, true, STRING_VARARGS)
-    );
-
-    // When:
-    final KsqlScalarFunction fun = udfIndex.getFunction(Arrays.asList(new SqlArgument[]{null}));
-
-    // Then:
-    assertThat(fun.name(), equalTo(EXPECTED));
-  }
```
Sullivan-Patrick, line 592: I'll need to figure out why this is failing, it's not immediately obvious to me.

ksqldb-common/src/main/java/io/confluent/ksql/function/UdfIndex.java
```diff
@@ -56,8 +56,10 @@
  *   <li>If two methods exist that match given the above rules, and both
  *       have variable arguments, return the method with the more non-variable
  *       arguments.</li>
- *   <li>If two methods exist that match only null values, return the one
- *       that was added first.</li>
+ *   <li>If two methods exist that match given the above rules, return the
+ *       method with fewer generic arguments.</li>
+ *   <li>If two methods exist that match given the above rules, an exception
+ *       is thrown due to a vague function call</li>
```
vcrfxia, line 62: Minor suggestion:
```suggestion
 *   <li>If two methods exist that match given the above rules, the function
 *       call is ambiguous and an exception is thrown</li>
```

ksqldb-common/src/main/java/io/confluent/ksql/function/UdfIndex.java
```diff
@@ -125,48 +126,58 @@ void addFunction(final T function) {
...
-
-    if (fun.isPresent()) {
-      return fun.get();
+    Optional<T> candidate = findMatchingCandidate(candidates, arguments, false);
+    if (candidate.isPresent()) {
+      return candidate.get();
     } else if (!supportsImplicitCasts) {
       throw createNoMatchingFunctionException(arguments);
     }
+    candidates.clear();
```
vcrfxia, line 157: If `candidates` is not used outside of `findMatchingCandidate()` and is also wiped between successive calls to `findMatchingCandidate()`, then it should just be created within `findMatchingCandidate()` rather than created outside and passed in.

ksqldb-common/src/main/java/io/confluent/ksql/function/UdfIndex.java
```diff
@@ -125,48 +126,58 @@ void addFunction(final T function) {
...
 
-    // if none were found (candidates is empty) try again with
-    // implicit casting
-    getCandidates(arguments, 0, root, candidates, new HashMap<>(), true);
-    return candidates
-        .stream()
-        .max(Node::compare)
-        .map(node -> node.value)
-        .orElseThrow(() -> createNoMatchingFunctionException(arguments));
+    //if non were found (candidate isn't present) try again with implicit casting
```
vcrfxia, line 159: nit: whitespace (though I'm surprised checkstyle didn't complain) + typo
```suggestion
    // if none were found (candidate isn't present) try again with implicit casting
```

ksqldb-common/src/main/java/io/confluent/ksql/function/UdfIndex.java
```diff
@@ -307,8 +316,15 @@ public String toString() {
     }
 
     int compare(final Node other) {
-      final int compare = compareFunctions.compare(value, other.value);
-      return compare == 0 ? -(order - other.order) : compare;
+      final int compareVal = compareFunctions.compare(value, other.value);
+      return compareVal == 0 ? countGenerics(other) - countGenerics(this) : compareVal;
```
vcrfxia, line 320: Nice! Much cleaner than the previous iteration.

ksqldb-common/src/test/java/io/confluent/ksql/function/UdfIndexTest.java
```diff
@@ -964,6 +936,60 @@ public void shouldThrowIfNoExactMatchAndImplicitCastDisabled() {
             + "(INTEGER)"));
   }
 
+  @Test
+  public void shouldThrowWhenGivenVagueImplicitCast() {
```
vcrfxia, line 940: Minor suggestion (slightly more concise):
```suggestion
  public void shouldThrowOnVagueImplicitCast() {
```

vcrfxia: Though I see that there's another negative test case for a different ambiguous implicit cast below. We could rename this one to `shouldThrowOnAmbiguousImplicitCastWithoutGenerics()` to contrast with the other one?

ksqldb-common/src/test/java/io/confluent/ksql/function/UdfIndexTest.java
```diff
@@ -964,6 +936,60 @@ public void shouldThrowIfNoExactMatchAndImplicitCastDisabled() {
...
+    // Given:
+    givenFunctions(
+        function(FIRST_FUNC, false, LONG, LONG),
+        function(SECOND_FUNC, false, DOUBLE, DOUBLE)
+    );
+
+    // When:
+    final Exception e = assertThrows(Exception.class,
```
vcrfxia, line 948: Is the type of exception thrown really just Exception rather than something more specific?

Same question for the other new negative test below.

ksqldb-common/src/test/java/io/confluent/ksql/function/UdfIndexTest.java
```diff
@@ -964,6 +936,60 @@ public void shouldThrowIfNoExactMatchAndImplicitCastDisabled() {
...
+    // When:
+    final KsqlScalarFunction fun = udfIndex.getFunction(ImmutableList
+        .of(SqlArgument.of(INTEGER), SqlArgument.of(INTEGER), SqlArgument.of(INTEGER)));
+
+    // Then:
+    assertThat(fun.name(), equalTo(EXPECTED));
+  }
+
+  @Test
+  public void shouldThrowOnComparablyEqualFunctionsWithSameGenericCount() {
```
vcrfxia, line 974: What does "comparably equal" mean in the context of this unit test name?

Another option for consideration: `shouldThrowOnAmbiguousImplicitCastWithGenerics()`.

vcrfxia (commented):
Thanks @Sullivan-Patrick ! LGTM besides the failing unit test and updating the "breaking change" description as we discussed offline. Great stuff.

ksqldb-common/src/test/java/io/confluent/ksql/function/UdfIndexTest.java
Sullivan-Patrick: I just noticed other tests used Exception. I could make it a KsqlException with no issues, seems to work fine.

ksqldb-common/src/test/java/io/confluent/ksql/function/UdfIndexTest.java
Sullivan-Patrick: It means that the two functions are equal according to `UdfIndex.Node.compare()`. I think I like your name better as it doesn't threaten to rely on internal details of the UdfIndex class.

ksqldb-common/src/test/java/io/confluent/ksql/function/UdfIndexTest.java
```diff
@@ -964,6 +936,60 @@ public void shouldThrowIfNoExactMatchAndImplicitCastDisabled() {
...
+    );
+
+    // When:
+    final KsqlException e = assertThrows(KsqlException.class,
+        () -> udfIndex
+            .getFunction(ImmutableList
+                .of(SqlArgument.of(INTEGER), SqlArgument.of(INTEGER), SqlArgument.of(INTEGER))));
+
+    // Then:
+    assertThat(e.getMessage(), containsString("Function 'name' does not accept parameters "
```
agavra, line 988: I think we should throw an error message that explicitly says something about ambiguous method resolution - you can take inspiration from Java's error message. Otherwise the users might be left scratching their head a bit

Sullivan-Patrick: I like this idea.

ksqldb-common/src/main/java/io/confluent/ksql/function/UdfIndex.java
```diff
@@ -125,48 +126,69 @@ void addFunction(final T function) {
...
+    final int len = candidates.size();
+    if (len == 1) {
+      return Optional.of(candidates.get(0).value);
+    } else if (len > 1) {
+      if (candidates.get(len - 1).compare(candidates.get(len - 2)) > 0) {
+        return Optional.of(candidates.get(len - 1).value);
+      }
+      throw new KsqlException("Function '" + udfName
+          + "' cannot be resolved due to ambiguous method parameters "
+          + getParamsAsString(arguments) + "."
```
vcrfxia, line 182: nit: we can make this error message even more useful by adding the workaround for the user, something like "Use CAST() to explicitly cast your parameters to one of the supported function calls." 

Also, we can pull the creation of this exception out into a helper method with an informative name to make the code slightly more readable (similar to the `createNoMatchingFunctionException()` helper method that exists today).

JimGalasyn (approved):
LGTM!

vcrfxia (commented):
Thanks @Sullivan-Patrick . This is great! One minor comment inline.

Also a suggestion to improve the "breaking change" message in your PR description (which is also great, btw): let's clarify what it means for "Functions that relied on vague implicit casting" to "no longer work." To "no longer work" means that any new queries will be rejected, and any existing queries will not be started after an upgrade.

LGTM otherwise!

PR #7721, opened 2021-06-23
feat: shared runtimes

wcarlson5:
A improved version of https://github.com/confluentinc/ksql/pull/7482 

This will cause ksql to use shared runtimes

### DIFF
--- a/ksqldb-common/src/main/java/io/confluent/ksql/util/KsqlConfig.java
+++ b/ksqldb-common/src/main/java/io/confluent/ksql/util/KsqlConfig.java
@@ -386,6 +386,13 @@ public class KsqlConfig extends AbstractConfig {
           + "if false, new lambda queries won't be processed but any existing lambda "
           + "queries are unaffected.";
 
+  public static final String KSQL_SHARED_RUNTIME_ENABLED = "ksql.SharedRuntime.enabled";
+  public static final Boolean KSQL_SHARED_RUNTIME_ENABLED_DEFAULT = false;
+  public static final String KSQL_SHARED_RUNTIME_ENABLED_DOC =
+          "Feature flag for sharing streams runtimes. Default is false. If false, persistent queries will use separate "
+                  + " runtimes, if true, new queries may share streams instances.";
+
+
   public static final String KSQL_SUPPRESS_BUFFER_SIZE_BYTES = "ksql.suppress.buffer.size.bytes";
   public static final Long KSQL_SUPPRESS_BUFFER_SIZE_BYTES_DEFAULT = -1L;
   public static final String KSQL_SUPPRESS_BUFFER_SIZE_BYTES_DOC =
@@ -999,6 +1006,12 @@ public class KsqlConfig extends AbstractConfig {
             KSQL_LAMBDAS_ENABLED_DEFAULT,
             Importance.LOW,
             KSQL_LAMBDAS_ENABLED_DOC
+        ).define(
+            KSQL_SHARED_RUNTIME_ENABLED,
+            Type.BOOLEAN,
+            KSQL_SHARED_RUNTIME_ENABLED_DEFAULT,
+            Importance.MEDIUM,
+            KSQL_SHARED_RUNTIME_ENABLED_DOC
         )
         .withClientSslSupport();
 
--- a/ksqldb-common/src/main/java/io/confluent/ksql/util/KsqlConfig.java
+++ b/ksqldb-common/src/main/java/io/confluent/ksql/util/KsqlConfig.java
@@ -393,7 +393,6 @@ public class KsqlConfig extends AbstractConfig {
           + "Default is false. If false, persistent queries will use separate "
           + " runtimes, if true, new queries may share streams instances.";
 
-
   public static final String KSQL_SUPPRESS_BUFFER_SIZE_BYTES = "ksql.suppress.buffer.size.bytes";
   public static final Long KSQL_SUPPRESS_BUFFER_SIZE_BYTES_DEFAULT = -1L;
   public static final String KSQL_SUPPRESS_BUFFER_SIZE_BYTES_DOC =
--- a/ksqldb-common/src/main/java/io/confluent/ksql/util/KsqlConfig.java
+++ b/ksqldb-common/src/main/java/io/confluent/ksql/util/KsqlConfig.java
@@ -389,8 +389,9 @@ public class KsqlConfig extends AbstractConfig {
   public static final String KSQL_SHARED_RUNTIME_ENABLED = "ksql.SharedRuntime.enabled";
   public static final Boolean KSQL_SHARED_RUNTIME_ENABLED_DEFAULT = false;
   public static final String KSQL_SHARED_RUNTIME_ENABLED_DOC =
-          "Feature flag for sharing streams runtimes. Default is false. If false, persistent queries will use separate "
-                  + " runtimes, if true, new queries may share streams instances.";
+      "Feature flag for sharing streams runtimes. "
+          + "Default is false. If false, persistent queries will use separate "
+          + " runtimes, if true, new queries may share streams instances.";
 
 
   public static final String KSQL_SUPPRESS_BUFFER_SIZE_BYTES = "ksql.suppress.buffer.size.bytes";
--- a/ksqldb-engine/src/main/java/io/confluent/ksql/query/QueryExecutor.java
+++ b/ksqldb-engine/src/main/java/io/confluent/ksql/query/QueryExecutor.java
@@ -81,6 +81,7 @@ import org.apache.kafka.streams.StreamsConfig;
 import org.apache.kafka.streams.Topology;
 import org.apache.kafka.streams.kstream.KStream;
 import org.apache.kafka.streams.kstream.KTable;
+import scala.collection.immutable.IntMap;
 
 // CHECKSTYLE_RULES.OFF: ClassDataAbstractionCoupling
 final class QueryExecutor {
@@ -145,13 +146,12 @@ final class QueryExecutor {
     );
     this.streams = new ArrayList<>();
 
-    for (int i = 0; i < 8; i++) {
-      BinPackedStreamsMetadata stream = new BinPackedStreamsMetadata(
-          kafkaStreamsBuilder,
-          config.getConfig(true).getInt(KsqlConfig.KSQL_QUERY_ERROR_MAX_QUEUE_SIZE)
-      );
-      streams.add(stream);
-    }
+    BinPackedStreamsMetadata stream = new BinPackedStreamsMetadata(
+        kafkaStreamsBuilder,
+        config.getConfig(true).getInt(KsqlConfig.KSQL_QUERY_ERROR_MAX_QUEUE_SIZE)
+    );
+    streams.add(stream);
+
   }
 
   TransientQueryMetadata buildTransientQuery(
@@ -284,7 +284,7 @@ final class QueryExecutor {
     final boolean useBinPacked = true;
     if (useBinPacked) {
       BinPackedStreamsMetadata binPackedStreamsMetadata = getStream(sources);
-      binPackedStreamsMetadata.addQuery(classifier, streamsProperties);
+      binPackedStreamsMetadata.addQuery(classifier, streamsProperties, sources);
       return new BinPackedPersistentQueryMetadataImpl(
           statementText,
           querySchema.logicalSchema(),
@@ -333,8 +333,17 @@ final class QueryExecutor {
   }
 
   private BinPackedStreamsMetadata getStream(final Set<SourceName> sources) {
-    //TODO: find streams app with no conflicting sources
-    return streams.get(0);
+    for (BinPackedStreamsMetadata binPackedStreamsMetadata: streams) {
+      if (binPackedStreamsMetadata.getSources().stream().noneMatch(sources::contains)) {
+        return binPackedStreamsMetadata;
+      }
+    }
+    BinPackedStreamsMetadata stream = new BinPackedStreamsMetadata(
+        kafkaStreamsBuilder,
+        config.getConfig(true).getInt(KsqlConfig.KSQL_QUERY_ERROR_MAX_QUEUE_SIZE)
+    );
+    streams.add(stream);
+    return stream;
   }
 
   private ProcessingLogger getUncaughtExceptionProcessingLogger(final QueryId queryId) {
--- a/ksqldb-engine/src/main/java/io/confluent/ksql/query/QueryExecutor.java
+++ b/ksqldb-engine/src/main/java/io/confluent/ksql/query/QueryExecutor.java
@@ -50,6 +50,8 @@ import io.confluent.ksql.schema.ksql.LogicalSchema;
 import io.confluent.ksql.schema.ksql.PhysicalSchema;
 import io.confluent.ksql.serde.WindowInfo;
 import io.confluent.ksql.services.ServiceContext;
+import io.confluent.ksql.util.BinPackedPersistentQueryMetadataImpl;
+import io.confluent.ksql.util.BinPackedStreamsMetadata;
 import io.confluent.ksql.util.KsqlConfig;
 import io.confluent.ksql.util.KsqlConstants;
 import io.confluent.ksql.util.KsqlException;
@@ -59,6 +61,9 @@ import io.confluent.ksql.util.PushQueryMetadata.ResultType;
 import io.confluent.ksql.util.QueryApplicationId;
 import io.confluent.ksql.util.QueryMetadata;
 import io.confluent.ksql.util.TransientQueryMetadata;
+import io.confluent.ksql.util.TransientQueryMetadata.ResultType;
+
+import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.HashMap;
 import java.util.LinkedList;
@@ -91,6 +96,7 @@ final class QueryExecutor {
   private final KafkaStreamsBuilder kafkaStreamsBuilder;
   private final StreamsBuilder streamsBuilder;
   private final MaterializationProviderBuilderFactory materializationProviderBuilderFactory;
+  private final List<BinPackedStreamsMetadata> streams;
 
   QueryExecutor(
       final SessionConfig config,
@@ -137,6 +143,15 @@ final class QueryExecutor {
         materializationProviderBuilderFactory,
         "materializationProviderBuilderFactory"
     );
+    this.streams = new ArrayList<>();
+
+    for (int i = 0; i < 8; i++) {
+      BinPackedStreamsMetadata stream = new BinPackedStreamsMetadata(
+          kafkaStreamsBuilder,
+          config.getConfig(true).getInt(KsqlConfig.KSQL_QUERY_ERROR_MAX_QUEUE_SIZE)
+      );
+      streams.add(stream);
+    }
   }
 
   TransientQueryMetadata buildTransientQuery(
@@ -266,31 +281,60 @@ final class QueryExecutor {
         .map(userErrorClassifiers::and)
         .orElse(userErrorClassifiers);
 
-    return new PersistentQueryMetadataImpl(
-        persistentQueryType,
-        statementText,
-        querySchema,
-        sources,
-        sinkDataSource,
-        planSummary,
-        queryId,
-        materializationProviderBuilder,
-        applicationId,
-        topology,
-        kafkaStreamsBuilder,
-        runtimeBuildContext.getSchemas(),
-        streamsProperties,
-        config.getOverrides(),
-        ksqlConfig.getLong(KSQL_SHUTDOWN_TIMEOUT_MS_CONFIG),
-        classifier,
-        physicalPlan,
-        ksqlConfig.getInt(KsqlConfig.KSQL_QUERY_ERROR_MAX_QUEUE_SIZE),
-        getUncaughtExceptionProcessingLogger(queryId),
-        ksqlConfig.getLong(KsqlConfig.KSQL_QUERY_RETRY_BACKOFF_INITIAL_MS),
-        ksqlConfig.getLong(KsqlConfig.KSQL_QUERY_RETRY_BACKOFF_MAX_MS),
-        listener,
-        scalablePushRegistry
-    );
+    final boolean useBinPacked = true;
+    if (useBinPacked) {
+      BinPackedStreamsMetadata binPackedStreamsMetadata = getStream(sources);
+      binPackedStreamsMetadata.addQuery(classifier, streamsProperties);
+      return new BinPackedPersistentQueryMetadataImpl(
+          statementText,
+          querySchema.logicalSchema(),
+          querySchema,
+          sources,
+          planSummary,
+          applicationId,
+          topology,
+          binPackedStreamsMetadata,
+          runtimeBuildContext.getSchemas(),
+          config.getOverrides(),
+          queryId,
+          materializationProviderBuilder,
+          physicalPlan,
+          getUncaughtExceptionProcessingLogger(queryId),
+          sinkDataSource
+      );
+    } else {
+
+      return new PersistentQueryMetadataImpl(
+          persistentQueryType,
+          statementText,
+          querySchema,
+          sources,
+          sinkDataSource,
+          planSummary,
+          queryId,
+          materializationProviderBuilder,
+          applicationId,
+          topology,
+          kafkaStreamsBuilder,
+          runtimeBuildContext.getSchemas(),
+          streamsProperties,
+          config.getOverrides(),
+          ksqlConfig.getLong(KSQL_SHUTDOWN_TIMEOUT_MS_CONFIG),
+          classifier,
+          physicalPlan,
+          ksqlConfig.getInt(KsqlConfig.KSQL_QUERY_ERROR_MAX_QUEUE_SIZE),
+          getUncaughtExceptionProcessingLogger(queryId),
+          ksqlConfig.getLong(KsqlConfig.KSQL_QUERY_RETRY_BACKOFF_INITIAL_MS),
+          ksqlConfig.getLong(KsqlConfig.KSQL_QUERY_RETRY_BACKOFF_MAX_MS),
+          listener,
+          scalablePushRegistry
+      );
+    }
+  }
+
+  private BinPackedStreamsMetadata getStream(final Set<SourceName> sources) {
+    //TODO: find streams app with no conflicting sources
+    return streams.get(0);
   }
 
   private ProcessingLogger getUncaughtExceptionProcessingLogger(final QueryId queryId) {
--- a/ksqldb-engine/src/main/java/io/confluent/ksql/query/QueryExecutor.java
+++ b/ksqldb-engine/src/main/java/io/confluent/ksql/query/QueryExecutor.java
@@ -146,15 +146,8 @@ final class QueryExecutor {
         materializationProviderBuilderFactory,
         "materializationProviderBuilderFactory"
     );
+
     this.streams = new ArrayList<>();
-
-    final SharedKafkaStreamsRuntime stream = new SharedKafkaStreamsRuntime(
-        kafkaStreamsBuilder,
-        config.getConfig(true).getInt(KsqlConfig.KSQL_QUERY_ERROR_MAX_QUEUE_SIZE),
-        buildStreamsProperties("applicationId", new QueryId(""))
-    );
-    streams.add(stream);
-
   }
 
   TransientQueryMetadata buildTransientQuery(
@@ -294,8 +287,7 @@ final class QueryExecutor {
         .map(userErrorClassifiers::and)
         .orElse(userErrorClassifiers);
 
-    final boolean useBinPacked = true;
-    if (useBinPacked) {
+    if (ksqlConfig.getBoolean(KsqlConfig.KSQL_SHARED_RUNTIME_ENABLED)) {
       final SharedKafkaStreamsRuntime sharedKafkaStreamsRuntime = getStream(sources);
       final PersistentQueriesInSharedRuntimesImpl binPackedPersistentQueryMetadata
               = new PersistentQueriesInSharedRuntimesImpl(
--- a/ksqldb-engine/src/main/java/io/confluent/ksql/query/QueryExecutor.java
+++ b/ksqldb-engine/src/main/java/io/confluent/ksql/query/QueryExecutor.java
@@ -149,7 +149,8 @@ final class QueryExecutor {
 
     final SharedKafkaStreamsRuntime stream = new SharedKafkaStreamsRuntime(
         kafkaStreamsBuilder,
-        config.getConfig(true).getInt(KsqlConfig.KSQL_QUERY_ERROR_MAX_QUEUE_SIZE)
+        config.getConfig(true).getInt(KsqlConfig.KSQL_QUERY_ERROR_MAX_QUEUE_SIZE),
+        buildStreamsProperties("applicationId", new QueryId(""))
     );
     streams.add(stream);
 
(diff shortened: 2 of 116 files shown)


wcarlson5:
I am opening up a new PR as we have settle on a design and are moving to start adapting tests

guozhangwang:
@wcarlson5 is this updated one ready for a quick look?

wcarlson5:
@guozhangwang I am just adding tests now so I shouldn't expect the code to change too much barring finding something majorly wrong. So yes please take a look if you can

ksqldb-engine/src/main/java/io/confluent/ksql/util/ValidationSharedKafkaStreamsRuntimeImpl.java
```diff
@@ -0,0 +1,155 @@
...
+import org.apache.kafka.streams.KafkaStreams;
+import org.apache.kafka.streams.LagInfo;
+import org.apache.kafka.streams.StreamsMetadata;
+import org.apache.kafka.streams.errors.StreamsException;
+import org.apache.kafka.streams.errors.StreamsUncaughtExceptionHandler;
+import org.apache.kafka.streams.processor.internals.namedtopology.KafkaStreamsNamedTopologyWrapper;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class ValidationSharedKafkaStreamsRuntimeImpl implements SharedKafkaStreamsRuntime {
```
guozhangwang, line 44: I'm wondering if we really need another impl here: for sandboxing, maybe we should just pass the parameter (btw if it is only for sandboxing, then maybe we can just inherit its name than using a new name `real`?), all the way down into the runtime, and then skip starting the KS accordingly?

guozhangwang (commented):
I'm less familiar with pt.3 https://github.com/apache/kafka/pull/10788/files (would need to review that again after pt.2 is merged, so maybe I would miss some parts. Would take another thorough pass after I complete reviewing pt.3.

ksqldb-engine/src/main/java/io/confluent/ksql/internal/QueryStateMetricsReportingListener.java
```diff
@@ -132,9 +132,12 @@ public void onDeregister(final QueryMetadata query) {
...
           Collections.singletonMap("status", queryApplicationId)
       );
-
-      this.metrics.addMetric(stateMetricName, (Gauge<String>) (config, now) -> state);
-      this.metrics.addMetric(errorMetricName, (Gauge<String>) (config, now) -> error);
+      try {
+        this.metrics.addMetric(stateMetricName, (Gauge<String>) (config, now) -> state);
+        this.metrics.addMetric(errorMetricName, (Gauge<String>) (config, now) -> error);
+      } catch (IllegalArgumentException e) {
+        //skip if the metric is already there
```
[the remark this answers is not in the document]
ableegoldman: Walker's mentioned this to me, I believe this is handling the metrics that are at the KS runtime level and thus no longer unique to each metric. 

I think we also discussed that silently swallowing any/every IllegalArgumentException is not very safe, so I'm guessing this is just a temporary workaround to unblock the tests?  In the end we should have separate handling for separate metrics types, ie query-level vs runtime-level, and avoid registering the same metric twice instead of catching and swallowing the exception.

If you want to do that in a followup PR that's ok with me, but can you (a) file a ticket for this, and (b) make sure we aren't accidentally swallowing a real IllegalArgumentException. It probably makes sense to tackle (b) by swallowing or rethrowing it based on the feature flag, so this doesn't affect the existing cases. Thoguhts?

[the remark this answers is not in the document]
ableegoldman: Note that this is the constructor for the '_Per Query_ Listener`, it probably shouldn't be called at all for metrics that aren't per query 🙂 . I think we probably need to break this up into a `PerQueryListener` which is keyed by the `queryId` (as. in the current map `perQuery`), and a `PerRuntimeListener` which is keyed by the `queryApplicationId` (in a new map, eg `perRuntime` or `perRuntimeMetrics)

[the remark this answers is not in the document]
wcarlson5: Yeah this is a temporary thing to unblock a lot of the tests. I think we can come back and rethink this when we take care of observe stuff. We shouldn't need to fix it for the POC really

ksqldb-engine/src/main/java/io/confluent/ksql/util/ValidationSharedKafkaStreamsRuntimeImpl.java
wcarlson5: This is a bit form the part 3 stuff. But basically for the validation we are adding the topology to a runtime that is not started. This way we can be sure its is going to work. I went through the other way it got complicated pretty quickly. IMHO it is much cleaner to to keep is seprate

guozhangwang: Let me think about this a bit more after I get hands dirty at the pt.3 PR on AK. What kind of validations are we actually checking for shared runtime in sandboxing here?

wcarlson5: thats a good question. @rodesai might be able to fill in more details but for sure we want to avoid duplicate topics. Perviously we used to make a query completely then close the runtime. That obviously won't work now. I am not sure how much it work it would take to revamp our validation flow so I would rather just keep it as close as we can an not block on it yet

ksqldb-engine/src/main/java/io/confluent/ksql/internal/QueryStateMetricsReportingListener.java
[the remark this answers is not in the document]
ableegoldman: I'm just a bit concerned about merging this unless we can be sure to get the followup done ASAP. Seems dangerous to silently swallow an exception like this -- maybe we could at least put it behind a feature flag for the time being? That way we can file a ticket for the V1 milestone

ksqldb-engine/src/main/java/io/confluent/ksql/util/ValidationSharedKafkaStreamsRuntimeImpl.java
ableegoldman: I think this is related to some of the rbac/authorization stuff we were discussing earlier, ie we try to start up a query as the user and if it's all good, then allow the "real" query to be started...is that right @rodesai ?

ksqldb-engine/src/main/java/io/confluent/ksql/internal/QueryStateMetricsReportingListener.java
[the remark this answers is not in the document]
ableegoldman: Just FYI, I filed https://confluentinc.atlassian.net/browse/KCI-812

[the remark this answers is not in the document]
wcarlson5: I can put it behind a feature for for sure

ksqldb-engine/src/main/java/io/confluent/ksql/internal/KsqlEngineMetrics.java
```diff
@@ -343,7 +343,11 @@ private void configureGaugeForState(
     final String description = String.format("Count of queries in %s state.", state.toString());
     final MetricName metricName = metrics.metricName(name, group, description, tags);
     final CountMetric countMetric = new CountMetric(metricName, gauge);
-    metrics.addMetric(metricName, gauge);
+    try {
+      metrics.addMetric(metricName, gauge);
+    } catch (IllegalArgumentException e) {
```
wcarlson5, line 348: This one I am not sure why we do it. It seems to be a duplicate of the other metric. They both report the state. It looks like some legacy stuff from 2018. It might be a good time to remove it?

ksqldb-engine/src/main/java/io/confluent/ksql/internal/QueryStateMetricsReportingListener.java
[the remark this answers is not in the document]
wcarlson5: With the new corruption detection these are actually still per-query metrics. So I changed it to use the queryId instead

wcarlson5:
@guozhangwang @ableegoldman @rodesai I cleaned up the hacks that I need to get this working as the streams were being developed and it should be ready for another review

ksqldb-engine/src/main/java/io/confluent/ksql/util/SharedKafkaStreamsRuntimeImpl.java
```diff
@@ -0,0 +1,225 @@
...
+      try {
+        kafkaStreams.removeNamedTopology(queryId.toString());
+      } catch (IllegalArgumentException e) {
+        //don't block
+      }
+    } else {
+      throw new IllegalStateException("Streams in not running but is in state"
+          + kafkaStreams.state());
+    }
+    kafkaStreams.cleanUpNamedTopology(queryId.toString());
```
guozhangwang, line 168: This function would check `getTopologyByName(name).isPresent()` which would return nothing since we already `metadata.remove(queryId.toString());` beforehand, right?

ksqldb-engine/src/main/java/io/confluent/ksql/util/ValidationSharedKafkaStreamsRuntimeImpl.java
guozhangwang: If the main goal is around rbac/authX then I feel we do not need a full fledged runtime impl to be used in the sandbox -- I am not 100% sure though..

ksqldb-engine/src/main/java/io/confluent/ksql/util/SharedKafkaStreamsRuntimeImpl.java
```diff
@@ -0,0 +1,225 @@
...
+                                       final int maxQueryErrorsQueueSize,
+                                       final Map<String, Object> streamsProperties) {
+    this.kafkaStreamsBuilder = kafkaStreamsBuilder;
+    kafkaStreams = kafkaStreamsBuilder.build(streamsProperties);
+    queryErrors
+        = new QueryMetadataImpl.TimeBoundedQueue(Duration.ofHours(1), maxQueryErrorsQueueSize);
+    this.streamsProperties = ImmutableMap.copyOf(streamsProperties);
+    metadata = new ConcurrentHashMap<>();
+    sources = new ConcurrentHashMap<>();
+    kafkaStreams.setUncaughtExceptionHandler(this::uncaughtHandler);
```
guozhangwang, line 65: Should we have a different exception handler which can tell which query inside the KS runtime was throwing the error, since I think it would be needed for the per-query healthcheck?

ksqldb-engine/src/main/java/io/confluent/ksql/util/SharedKafkaStreamsRuntimeImpl.java
```diff
@@ -0,0 +1,225 @@
...
+  }
+
+  public void addQuery(
+          final QueryErrorClassifier errorClassifier,
+          final Map<String, Object> streamsProperties,
+          final PersistentQueriesInSharedRuntimesImpl persistentQueriesInSharedRuntimesImpl,
+          final QueryId queryId) {
+    this.errorClassifier = errorClassifier;
+    if (metadata.containsKey(queryId.toString())) {
+      kafkaStreams.removeNamedTopology(queryId.toString());
```
guozhangwang, line 80: When this scenario can actually happen? And should we really just remove the old query silently?

ksqldb-engine/src/main/java/io/confluent/ksql/util/SharedKafkaStreamsRuntimeImpl.java
```diff
@@ -0,0 +1,225 @@
...
+  public Set<StreamsTaskMetadata> getTaskMetadata() {
+    return kafkaStreams.metadataForLocalThreads()
+                       .stream()
+                       .flatMap(t -> t.activeTasks().stream())
+                       .map(StreamsTaskMetadata::fromStreamsTaskMetadata)
+                       .collect(Collectors.toSet());
+  }
+
+  public void restart(final QueryId queryId) {
+    final KafkaStreamsNamedTopologyWrapper newKafkaStreams = kafkaStreamsBuilder
```
guozhangwang, line 141: Maybe we have discussed about this but I've forgot about our conclusions by then.. why we'd need to recreate a KS runtime and re-add the rest of the queries just to restart a new query? Could we just remove that query from the runtime and re-add it back, similarly to a `close(query)` followed by a `start(query)`?

ksqldb-engine/src/main/java/io/confluent/ksql/util/SharedKafkaStreamsRuntime.java
```diff
@@ -0,0 +1,76 @@
...
+import org.apache.kafka.streams.LagInfo;
+import org.apache.kafka.streams.StreamsMetadata;
+import org.apache.kafka.streams.errors.StreamsUncaughtExceptionHandler;
+
+
+public interface SharedKafkaStreamsRuntime {
+
+  void markSources(QueryId queryId, Set<SourceName> sourceNames);
+
+  void addQuery(
```
guozhangwang, line 37: Very nit: to align with the naming of `restart/close`, could we just call it `add`, or vice versa call others `start/restart/closeQuery`?

ksqldb-engine/src/main/java/io/confluent/ksql/util/SharedKafkaStreamsRuntime.java
```diff
@@ -0,0 +1,76 @@
...
+
+  Collection<StreamsMetadata> allMetadata();
+
+  Set<StreamsTaskMetadata> getTaskMetadata();
+
+  void restart(QueryId queryId);
+
+  boolean isError(QueryId queryId);
+
+  void close(QueryId queryId);
```
guozhangwang, line 59: nit: maybe rename to `stop(query)` to differentiate from `close()`?

ksqldb-engine/src/main/java/io/confluent/ksql/util/SharedKafkaStreamsRuntimeImpl.java
```diff
@@ -0,0 +1,225 @@
...
+  public void start(final QueryId queryId) {
+    if (metadata.containsKey(queryId.toString()) && !metadata.get(queryId.toString()).everStarted) {
+      if (!kafkaStreams.getTopologyByName(queryId.toString()).isPresent()) {
+        System.err.println(metadata.get(queryId.toString()).getTopology().describe());
+        kafkaStreams.addNamedTopology(metadata.get(queryId.toString()).getTopology());
+      } else {
+        throw new IllegalArgumentException("not done removing query: " + queryId);
+      }
+    } else {
+      throw new IllegalArgumentException("query not added to runtime");
```
guozhangwang, line 188: include the query id here as well?

ksqldb-engine/src/test/java/io/confluent/ksql/util/SharedKafkaStreamsRuntimeImplTest.java
```diff
@@ -0,0 +1,44 @@
...
+        sharedKafkaStreamsRuntimeImpl = new SharedKafkaStreamsRuntimeImpl(
+            kafkaStreamsBuilder,
+            5,
+            streamProps
+        );
+    }
+
+    @Test
+    public void shouldAddQuery() {
+//        sharedKafkaStreamsRuntime.addQuery();
```
guozhangwang, line 42: typo?

ksqldb-engine/src/main/java/io/confluent/ksql/util/SharedKafkaStreamsRuntime.java
```diff
@@ -0,0 +1,76 @@
...
+
+  KafkaStreams getKafkaStreams();
+
+  KafkaStreams.State state();
+
+  Collection<StreamsMetadata> allMetadata();
+
+  Set<StreamsTaskMetadata> getTaskMetadata();
+
+  void restart(QueryId queryId);
```
guozhangwang, line 55: Seems we do not have any callers for this function yet, should we add some unit tests?

ksqldb-engine/src/main/java/io/confluent/ksql/util/PersistentQueriesInSharedRuntimesImpl.java
```diff
@@ -0,0 +1,385 @@
...
+        ImmutableMap.copyOf(
+            Objects.requireNonNull(overriddenProperties, "overriddenProperties"));
+    this.sourceNames = Objects.requireNonNull(sourceNames, "sourceNames");
+    this.queryId = Objects.requireNonNull(queryId, "queryId");
+    this.processingLogger = requireNonNull(processingLogger, "processingLogger");
+    this.physicalPlan = requireNonNull(physicalPlan, "physicalPlan");
+    this.resultSchema = requireNonNull(schema, "schema");
+    this.materializationProviderBuilder =
+        requireNonNull(materializationProviderBuilder, "materializationProviderBuilder");
+    this.listener = requireNonNull(listener, "listen");
```
guozhangwang, line 133: typo: listener.

ksqldb-engine/src/main/java/io/confluent/ksql/util/SharedKafkaStreamsRuntimeImpl.java
```diff
@@ -0,0 +1,225 @@
...
+      // log in finally block to make sure that if there's ever an error in the classification
+      // we still get this in our logs.
+      final QueryError queryError =
+          new QueryError(
+              System.currentTimeMillis(),
+              Throwables.getStackTraceAsString(e),
+              errorType
+          );
+      for (PersistentQueriesInSharedRuntimesImpl query: metadata.values()) {
+        query.getListener().onError(metadata.get(query.getQueryId().toString()), queryError);
```
guozhangwang, line 107: Related to my other comment, this is the part I was concerning about: if it throws we just mark all runtimes / queries as failed --- even if some queries are on different runtimes. Is that okay?

ksqldb-engine/src/main/java/io/confluent/ksql/util/SharedKafkaStreamsRuntimeImpl.java
wcarlson5: I think our discussion concluded with something like restart is not used, But it would be good to have a way to get a fresh runtime if necessary.

ksqldb-engine/src/main/java/io/confluent/ksql/util/SharedKafkaStreamsRuntime.java
wcarlson5: I think just `add` would be a bit odd. how about `register`?

ksqldb-engine/src/main/java/io/confluent/ksql/util/SharedKafkaStreamsRuntimeImpl.java
wcarlson5: We decided that it was okay for a first round. It will require some more changes to make it possible to distinguish which `NamedTopology` the error comes from. But it shouldn't stop the query it will just let the metrics know there has been an error

ksqldb-engine/src/main/java/io/confluent/ksql/util/SharedKafkaStreamsRuntimeImpl.java
wcarlson5: There is a possible race where is the query was added and then replaced the `stop()` might not be able to remove the topology, once the `removeNamedTopology` returns a `future` we can remove this check. we swallow the possible error is stop to clean up the state and let the other queries continue processing so we might need to remove it again.

ksqldb-engine/src/test/java/io/confluent/ksql/util/SharedKafkaStreamsRuntimeImplTest.java
wcarlson5: pushed some more tests that clean this up

ksqldb-engine/src/main/java/io/confluent/ksql/util/SharedKafkaStreamsRuntime.java
wcarlson5: yes

ksqldb-engine/src/main/java/io/confluent/ksql/util/SharedKafkaStreamsRuntimeImpl.java
guozhangwang: Okay that makes sense. For the second part `we swallow the possible error is stop to clean up the state and let the other queries continue processing so we might need to remove it again.` though, not sure I fully follow that (maybe it's just due to typos..) could you re-explain to me?

ksqldb-engine/src/main/java/io/confluent/ksql/util/SharedKafkaStreamsRuntimeImpl.java
guozhangwang: If `restart` is used in the future, I think it is not necessary to get a fresh runtime still?

ksqldb-engine/src/main/java/io/confluent/ksql/util/SharedKafkaStreamsRuntimeImpl.java
wcarlson5: the `metadata.remove(queryId.toString())` only affects the ksql metadata I have stored in this runtime class.

ksqldb-engine/src/main/java/io/confluent/ksql/util/SharedKafkaStreamsRuntimeImpl.java
wcarlson5: wow I butchered that sentence.... I just remove the swallowing of the error so it shouldn't be an issue anymore

ksqldb-engine/src/main/java/io/confluent/ksql/util/SharedKafkaStreamsRuntimeImpl.java
wcarlson5: We can just remove it if it's not used if you would prefer

guozhangwang: Ah, that works too :)

ksqldb-engine/src/main/java/io/confluent/ksql/internal/KsqlEngineMetrics.java
ableegoldman: What other metric? cc @rodesai

wcarlson5: @ableegoldman these https://github.com/confluentinc/ksql/blob/81bfc1f74a4f05b0688a8f6d820808981e4c745b/ksqldb-engine/src/main/java/io/confluent/ksql/internal/KsqlEngineMetrics.java#L360

ksqldb-engine/src/main/java/io/confluent/ksql/internal/QueryStateMetricsReportingListener.java
```diff
@@ -104,35 +104,34 @@ public void onDeregister(final QueryMetadata query) {
...
           "query-status",
           groupPrefix + "ksql-queries",
           "The current status of the given query.",
-          Collections.singletonMap("status", queryApplicationId));
+          Collections.singletonMap("status", queryId));
 
       errorMetricName = metrics.metricName(
           "error-status",
           groupPrefix + "ksql-queries",
           "The current error status of the given query, if the state is in ERROR state",
```
ableegoldman, line 132: @rodesai what is this metric exactly? I don't think I see anything on the dashboard resembling it

ksqldb-engine/src/main/java/io/confluent/ksql/util/SharedKafkaStreamsRuntimeImpl.java
ableegoldman: I think it's still targeted for V1 to at least have the correct information about which query is affected. But not in this PR, we're _trying_ not to tackle every single thing at the same time 😜

ksqldb-engine/src/main/java/io/confluent/ksql/query/QueryExecutor.java
```diff
@@ -122,7 +137,9 @@
       final FunctionRegistry functionRegistry,
       final KafkaStreamsBuilder kafkaStreamsBuilder,
       final StreamsBuilder streamsBuilder,
-      final MaterializationProviderBuilderFactory materializationProviderBuilderFactory
+      final MaterializationProviderBuilderFactory materializationProviderBuilderFactory,
```
guozhangwang, line 140: It seems the only reason we are passing in this streamsBuilder in the line above is for testing (the only  non-testing caller always pass in a `new TopologyStreamsBuilder`). And now with the new call path this object is not always used and hence is a bit awkward to still maintain it as a class field.

How about not maintaining it at all, and pass in one at the `buildPersistentQuery` and let the caller `QueryRegistryImpl` to determine which StreamsBuilder to pass in based on the `ksqlConfig`. We can even just have two the two projected `build(Not)SharedPersistentQuery` as public and let the `QueryRegistryImpl` to determine which one to call.

ksqldb-engine/src/main/java/io/confluent/ksql/util/SandboxedPersistentQueriesInSharedRuntimesImpl.java
```diff
@@ -0,0 +1,52 @@
...
+ * specific language governing permissions and limitations under the License.
+ */
+
+package io.confluent.ksql.util;
+
+/**
+ * Sandboxed {@link PersistentQueryMetadata} that prevents to modify the state of the internal
+ * {@link org.apache.kafka.streams.KafkaStreams}.
+ */
+public final class SandboxedPersistentQueriesInSharedRuntimesImpl
```
guozhangwang, line 22: Please see my other comment: checking the sandbox call path here, I still do not know why we need to have an extended class here: since start/stop are not triggered anyways in the sandboxing, can't we just create the `PersistentQueriesInSharedRuntimesImpl` even for sandboxing?

ksqldb-engine/src/main/java/io/confluent/ksql/util/ValidationSharedKafkaStreamsRuntimeImpl.java
guozhangwang: I still feel that we can follow the same create-query -> close-runtime pattern here: since the goal for sandboxing are for 1) making sure the query is well-formed and hence would not fail compilation before writing to cmd topic, and 2) rbac stuff (not sure if this is really done at sandboxing, cc @rodesai @agavra could you clarify?) I think we can still do the sandboxing by creating a runtime and add that query to the runtime, and then close the runtime --- i.e. we only do a single-query shared runtime at sandboxing which would not influence the validations we'd want.

guozhangwang (commented):
Thanks @wcarlson5 ! Since now I'm done with the AK side pt.3 code ,I made another pass on the new commits.

ksqldb-engine/src/test/java/io/confluent/ksql/query/QueryExecutorTest.java
```diff
@@ -672,6 +681,64 @@ public void shouldAddMetricsInterceptorsToExistingStringList() {
...
+            SOURCES,
+            physicalPlan,
+            SUMMARY,
+            queryListener,
+            Collections::emptyList
+    ).start();
+  }
+
+  @Test
+  public void shouldMakePersistentQueriesWithDifferentSources() {
```
rodesai, line 714: should this assert that they choose the same runtime?

rodesai (commented):
Thanks @wcarlson5, took a first pass and left some feedback inline.

One improvement I think we should make is to rearrange this so that the KafkaStreams instance is selected when building the physical plan, rather than executing the physical plan. Doing the latter is brittle - it will be very easy for us to write bugs that cause queries to wind up on different runtimes across versions by mistake. It would be better to make the runtime for a query explicit by putting this in `QueryPlan` (which is then written to the command topic).

To do this you'll probably want to refactor what's here a bit by pulling runtime management out into its own class rather than relying on `QueryRegistryImpl` and `QueryExecutor`. Instead, we could have an interface like:
```
interface RuntimeInstanceManager {
    RuntimeId selectRuntimeForPhysicalPlan(final PhysicalPlan plan);
    KafkaStreams getRuntimeById(final RuntimeId id);
}
```
That you then call from `planQuery` to figure out the runtime for a query. Then, `QueryExecutor` can just get the runtime it needs to use from this object.

ksqldb-common/src/main/java/io/confluent/ksql/util/KsqlConfig.java
```diff
@@ -393,6 +393,14 @@
           + "if false, new lambda queries won't be processed but any existing lambda "
           + "queries are unaffected.";
 
+  public static final String KSQL_SHARED_RUNTIME_ENABLED = "ksql.SharedRuntime.enabled";
```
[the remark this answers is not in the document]
rodesai: Also, how do we ensure that older queries don't use shared runtimes?

wcarlson5 (commented):
@rodesai @guozhangwang I think I got to all your comments. Let me know if something was not addressed. Sorry for the huge diff so late in the PR process. It is mostly renaming things and moving them around :)

ksqldb-engine/src/main/java/io/confluent/ksql/query/QueryExecutor.java
guozhangwang: Seems not addressed yet?

guozhangwang (commented):
Hey @wcarlson5 is this PR ready for another look now? Seems the sandboxing logic has not been updated yet, or do you plan to do it in a later PR?

ksqldb-engine/src/main/java/io/confluent/ksql/query/QueryExecutor.java
wcarlson5: Ah I miss-understood your suggestion. I pretty much did the second half but did not actually remove the global var. However I just updated it so it should be ready for another pass. 

`StreamsBuilder` is also used in the transient query creation but you are right that only is testing is it inputed, otherwise it is just the default. As far as I can tell it was only used in one way too. 

Eventually I would like to have the named wrapper to take a normal topology as well. Then we only need one type and we can make it a global var again and it will be simple again. But that can wait.

ksqldb-engine/src/main/java/io/confluent/ksql/query/QueryRegistryImpl.java
```diff
@@ -111,6 +130,14 @@ private QueryRegistryImpl(final QueryRegistryImpl original) {
         .filter(Optional::isPresent)
         .map(Optional::get)
         .collect(Collectors.toList());
+    if (!original.sandbox) {
```
wcarlson5, line 133: @rodesai When fixing an issue with the parametrization of a test it surfaced an issue where the registry is creating a sandbox from a sandbox. This was causing an exception. This happens only during the replacement of a persistent query. It doesn't make sense to me why thing is happening. I will look into it more tomorrow but perhaps you have some idea? 

This fix does not work it just prevents the exception but no data is produced

wcarlson5: I adapted it to be able to copy a validation runtime but still no data is coming though

PR #7728, opened 2021-06-25
feat: update CONVERT_TZ to accept TIME types

jzaralim:
### Description 
^^^

The QTT tests won't pass until #7722 is merged, since it contains some changes needed to enable TIME for UDFs.

### Testing done 
QTT + unit tests

### Reviewer checklist


JimGalasyn (approved):

spena:
@jzaralim I'm not sure if we should allow converting timezones from time individually. Seems confusing if we don't have the date in it. Perhaps for this cases, users can just use a math UDF to sum 3 hours to their time.

I also tested Mysql, and it returns NULL when converting a time individually.

jzaralim:
@spena I'll close this then. I looked at some other systems and converting timezones for TIME isn't done anywhere.

Issue #7720, opened 2021-06-23
Cleanup abandoned state stores on startup


AlanConfluent:
**Describe the bug**
When queries are terminated, we attempt to call `kafkastreams.close(timeout)` and if we hit the timeout, we skip cleaning up the streams state via `kafkastreams.cleanup()`.  This was added in #7717 since calling cleanup anyway resulted in an exception. If the query has been terminated, then on startup, when we know the query is no longer running, the state should be cleaned up.

**To Reproduce**
Requires significantly stressing the server so that it misses the the timeout in `kafkastreams.close(timeout)`.  This has happened when running tests on CI.

**Expected behavior**
State must be kept so that on startup, we know which state stores we can clean up.  Might want to use a similar mechanism as was used for https://github.com/confluentinc/ksql/issues/4009, which was a local file of things to check up on implemented here: https://github.com/confluentinc/ksql/pull/6714

**Actual behaviour**
No cleanups.

nateab:
Fixed by #7729

PR #7729, opened 2021-06-25
[KCI-612] Clean up a terminated query's state stores

lct45:
### Description 
In a recent disk out of space incident, a few terminated queries had state stores that didn't get cleaned up. This took up a lot of disk space and ultimately had to be cleaned up manually. It seems like because the query was terminated and quickly after the pod was rolled, the query wasn't restarted but the stores didn't get cleaned up before we rolled the pod.

Now we'll check state stores when we initialize the app and remove and that don't match a currently running query.

### Testing done 
Unit + integration tests included

### Reviewer checklist

### DIFF
--- a/ksqldb-engine/src/main/java/io/confluent/ksql/engine/QueryCleanupService.java
+++ b/ksqldb-engine/src/main/java/io/confluent/ksql/engine/QueryCleanupService.java
@@ -22,7 +22,6 @@ import io.confluent.ksql.schema.registry.SchemaRegistryUtil;
 import io.confluent.ksql.services.ServiceContext;
 
 import java.io.File;
-import java.io.IOException;
 import java.util.Objects;
 import java.util.Set;
 import java.util.concurrent.BlockingQueue;
@@ -40,8 +39,7 @@ import org.slf4j.LoggerFactory;
  * real/sandboxed engines.</p>
  */
 @SuppressWarnings("UnstableApiUsage")
-public
-class QueryCleanupService extends AbstractExecutionThreadService {
+public class QueryCleanupService extends AbstractExecutionThreadService {
 
   private static final Logger LOG = LoggerFactory.getLogger(QueryCleanupService.class);
   private static final Runnable SHUTDOWN_SENTINEL = () -> { };
@@ -116,19 +114,18 @@ class QueryCleanupService extends AbstractExecutionThreadService {
     public void run() {
       try {
         FileUtils.deleteDirectory(new File(stateDir + appId));
-        LOG.warn(
-            "Deleted local state store for non-existing query {}. "
-                + "This is not expected and was likely due to a "
-                + "race condition when the query was dropped before.",
+        LOG.warn("Deleted local state store for non-existing query {}. "
+            + "This is not expected and was likely due to a "
+            + "race condition when the query was dropped before.",
             appId);
-      } catch (IOException e) {
+      } catch (Exception e) {
         LOG.error("Error cleaning up state directory {}\n. {}", appId, e);
       }
       tryRun(
           () -> SchemaRegistryUtil.cleanupInternalTopicSchemas(
-              appId,
-              serviceContext.getSchemaRegistryClient(),
-              isTransient),
+            appId,
+            serviceContext.getSchemaRegistryClient(),
+            isTransient),
           "internal topic schemas"
       );
 
@@ -147,6 +144,7 @@ class QueryCleanupService extends AbstractExecutionThreadService {
         LOG.warn("Failed to cleanup {} for {}", resource, appId, e);
       }
     }
+
     public void setStateDir(final String newStateDir) {
       stateDir = newStateDir;
     }
--- a/ksqldb-engine/src/main/java/io/confluent/ksql/engine/QueryCleanupService.java
+++ b/ksqldb-engine/src/main/java/io/confluent/ksql/engine/QueryCleanupService.java
@@ -17,8 +17,12 @@ package io.confluent.ksql.engine;
 
 import com.google.common.collect.ImmutableSet;
 import com.google.common.util.concurrent.AbstractExecutionThreadService;
+import com.spun.util.io.FileUtils;
 import io.confluent.ksql.schema.registry.SchemaRegistryUtil;
 import io.confluent.ksql.services.ServiceContext;
+
+import java.io.File;
+import java.io.IOException;
 import java.util.Objects;
 import java.util.Set;
 import java.util.concurrent.BlockingQueue;
@@ -36,6 +40,7 @@ import org.slf4j.LoggerFactory;
  * real/sandboxed engines.</p>
  */
 @SuppressWarnings("UnstableApiUsage")
+public
 class QueryCleanupService extends AbstractExecutionThreadService {
 
   private static final Logger LOG = LoggerFactory.getLogger(QueryCleanupService.class);
@@ -43,7 +48,7 @@ class QueryCleanupService extends AbstractExecutionThreadService {
 
   private final BlockingQueue<Runnable> cleanupTasks;
 
-  QueryCleanupService() {
+  public QueryCleanupService() {
     cleanupTasks = new LinkedBlockingDeque<>();
   }
 
@@ -85,19 +90,22 @@ class QueryCleanupService extends AbstractExecutionThreadService {
     cleanupTasks.add(task);
   }
 
-  static class QueryCleanupTask implements Runnable {
+  public static class QueryCleanupTask implements Runnable {
     private final String appId;
     private final boolean isTransient;
     private final ServiceContext serviceContext;
+    private String stateDir;
 
-    QueryCleanupTask(
+    public QueryCleanupTask(
         final ServiceContext serviceContext,
         final String appId,
-        final boolean isTransient
+        final boolean isTransient,
+        final String stateDir
     ) {
       this.serviceContext = Objects.requireNonNull(serviceContext, "serviceContext");
       this.appId = Objects.requireNonNull(appId, "appId");
       this.isTransient = isTransient;
+      this.stateDir = stateDir;
     }
 
     public String getAppId() {
@@ -106,6 +114,16 @@ class QueryCleanupService extends AbstractExecutionThreadService {
 
     @Override
     public void run() {
+      try {
+        FileUtils.deleteDirectory(new File(stateDir + appId));
+        LOG.warn(
+            "Deleted local state store for non-existing query {}. "
+                + "This is not expected and was likely due to a "
+                + "race condition when the query was dropped before.",
+            appId);
+      } catch (IOException e) {
+        LOG.error("Error cleaning up state directory {}\n. {}", appId, e);
+      }
       tryRun(
           () -> SchemaRegistryUtil.cleanupInternalTopicSchemas(
               appId,
@@ -128,6 +146,9 @@ class QueryCleanupService extends AbstractExecutionThreadService {
       } catch (final Exception e) {
         LOG.warn("Failed to cleanup {} for {}", resource, appId, e);
       }
+    }
+    public void setStateDir(final String newStateDir) {
+      stateDir = newStateDir;
     }
   }
 
--- a/ksqldb-engine/src/main/java/io/confluent/ksql/engine/QueryCleanupService.java
+++ b/ksqldb-engine/src/main/java/io/confluent/ksql/engine/QueryCleanupService.java
@@ -22,6 +22,8 @@ import io.confluent.ksql.schema.registry.SchemaRegistryUtil;
 import io.confluent.ksql.services.ServiceContext;
 
 import java.io.File;
+import java.nio.file.Path;
+import java.nio.file.Paths;
 import java.util.Objects;
 import java.util.Set;
 import java.util.concurrent.BlockingQueue;
@@ -113,7 +115,8 @@ public class QueryCleanupService extends AbstractExecutionThreadService {
     @Override
     public void run() {
       try {
-        FileUtils.deleteDirectory(new File(stateDir + appId));
+        final Path pathName = Paths.get(stateDir + "/" + appId);
+        FileUtils.deleteDirectory(new File(String.valueOf(pathName.normalize())));
         LOG.warn("Deleted local state store for non-existing query {}. "
             + "This is not expected and was likely due to a "
             + "race condition when the query was dropped before.",
--- a/ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/KsqlRestApplication.java
+++ b/ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/KsqlRestApplication.java
@@ -88,6 +88,7 @@ import io.confluent.ksql.rest.util.ClusterTerminator;
 import io.confluent.ksql.rest.util.ConcurrencyLimiter;
 import io.confluent.ksql.rest.util.KsqlInternalTopicUtils;
 import io.confluent.ksql.rest.util.KsqlUncaughtExceptionHandler;
+import io.confluent.ksql.rest.util.PersistentQueryCleanup;
 import io.confluent.ksql.rest.util.RocksDBConfigSetterHandler;
 import io.confluent.ksql.schema.registry.KsqlSchemaRegistryClientFactory;
 import io.confluent.ksql.security.KsqlAuthorizationValidator;
@@ -104,7 +105,6 @@ import io.confluent.ksql.util.AppInfo;
 import io.confluent.ksql.util.KsqlConfig;
 import io.confluent.ksql.util.KsqlException;
 import io.confluent.ksql.util.KsqlServerException;
-import io.confluent.ksql.util.PersistentQueryMetadata;
 import io.confluent.ksql.util.ReservedInternalTopics;
 import io.confluent.ksql.util.RetryUtil;
 import io.confluent.ksql.util.WelcomeMsgUtils;
@@ -118,19 +118,12 @@ import io.vertx.ext.dropwizard.DropwizardMetricsOptions;
 import io.vertx.ext.dropwizard.Match;
 import java.io.Console;
 import java.io.File;
-import java.io.IOException;
 import java.io.OutputStreamWriter;
 import java.io.PrintWriter;
 import java.net.MalformedURLException;
 import java.net.URI;
 import java.net.URL;
 import java.nio.charset.StandardCharsets;
-import java.nio.file.FileVisitResult;
-import java.nio.file.Files;
-import java.nio.file.Path;
-import java.nio.file.Paths;
-import java.nio.file.SimpleFileVisitor;
-import java.nio.file.attribute.BasicFileAttributes;
 import java.time.Duration;
 import java.util.Collections;
 import java.util.LinkedList;
@@ -138,7 +131,6 @@ import java.util.List;
 import java.util.Map;
 import java.util.Optional;
 import java.util.Properties;
-import java.util.Set;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.ExecutionException;
 import java.util.concurrent.Executors;
@@ -482,14 +474,15 @@ public final class KsqlRestApplication implements Executable {
         processingLogContext.getConfig(),
         ksqlConfigNoPort
     );
-    commandRunner.processPriorCommands(
+    commandRunner.processPriorCommands(new PersistentQueryCleanup(
         configWithApplicationServer
-            .getKsqlStreamConfigProps()
-            .getOrDefault(
-                StreamsConfig.STATE_DIR_CONFIG,
-                StreamsConfig.configDef().defaultValues().get(StreamsConfig.STATE_DIR_CONFIG))
+        .getKsqlStreamConfigProps()
+        .getOrDefault(
+          StreamsConfig.STATE_DIR_CONFIG,
+          StreamsConfig.configDef().defaultValues().get(StreamsConfig.STATE_DIR_CONFIG))
         .toString(),
-        serviceContext);
+        serviceContext)
+    );
 
     commandRunner.start();
     maybeCreateProcessingLogStream(
--- a/ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/KsqlRestApplication.java
+++ b/ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/KsqlRestApplication.java
@@ -88,7 +88,7 @@ import io.confluent.ksql.rest.util.ClusterTerminator;
 import io.confluent.ksql.rest.util.ConcurrencyLimiter;
 import io.confluent.ksql.rest.util.KsqlInternalTopicUtils;
 import io.confluent.ksql.rest.util.KsqlUncaughtExceptionHandler;
-import io.confluent.ksql.rest.util.PersistentQueryCleanup;
+import io.confluent.ksql.rest.util.PersistentQueryCleanupImpl;
 import io.confluent.ksql.rest.util.RocksDBConfigSetterHandler;
 import io.confluent.ksql.schema.registry.KsqlSchemaRegistryClientFactory;
 import io.confluent.ksql.security.KsqlAuthorizationValidator;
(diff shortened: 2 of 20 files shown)


ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/KsqlRestApplication.java
```diff
@@ -453,6 +462,40 @@ private void cleanupOldState() {
...
+    final String stateDir =
+        configWithApplicationServer
+            .getKsqlStreamConfigProps()
+            .get(StreamsConfig.STATE_DIR_CONFIG)
+            .toString();
+
+    final Set<String> stateStoreNames =
+        ksqlEngine.getPersistentQueries()
+            .stream()
+            .map(PersistentQueryMetadata::getQueryId)
```
lct45, line 475: I think we can use application ID here and it should match the state store directories. Will test locally once my mvn builds 🤪

ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/KsqlRestApplication.java
```diff
@@ -453,6 +462,40 @@ private void cleanupOldState() {
...
+            .map(QueryId::toString)
+            .collect(Collectors.toSet());
+    try {
+      Files.list(Paths.get(stateDir))
+          .map(Path::toFile)
+          .forEach(
+              f -> {
+                if (stateStoreNames.stream().noneMatch((name) -> f.getName().endsWith(name))) {
+                  try {
+                    Files.walk(f.toPath())
```
guozhangwang, line 485: nit: I personally like the `Files.walkFileTree` since it allows us to print a more detailed error message, e.g. which file delete encounters the error, with the visitor pattern.

An example of its usage: https://github.com/apache/kafka/blob/trunk/clients/src/main/java/org/apache/kafka/common/utils/Utils.java#L812

guozhangwang: Plus, since we would not expect to be finding any existing state store dirs that do not match existing persistent stores, we'd also log a warning if we indeed successfully delete something here.. e.g. "WARN: Deleted local state store for non-existing query {}, this is not expected and probably due to a race condition when query {} was dropped before".

ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/KsqlRestApplication.java
```diff
@@ -466,6 +509,8 @@ private void initialize(final KsqlConfig configWithApplicationServer) {
         ksqlConfigNoPort
     );
     commandRunner.processPriorCommands();
+    cleanupOldStateDirectories(configWithApplicationServer);
```
guozhangwang, line 512: If we indeed have large amount of dangling state store dirs here there's a small risk that the restoring of existing queries -- i.e. starting the execution of them -- may already failed. Hence I'm wondering if we can refactor the timing of triggering this function a bit, i.e. after we've consumed through all the commands from the cmd topic, but before we start executing the restoration of queries?

guozhangwang (commented):
I've made a quick pass on the non-testing code.

agavra:
does this fix https://github.com/confluentinc/ksql/issues/7720? cc @AlanConfluent

guozhangwang:
Also it makes me thinking: since we only tries to do the cleanup upon (re-)starting the app, we would still accumulate leaked state stores as we continue to run until we exhaust 500GB --- in the future, it may be 125GB --- in which case we would error out and trying to replace the thread only, and hence would fall into the loop still. I.e. unless the pod is bounced either due to upgrade/scaling or pro-actively by ourselves, we would still have this issue. Is that right?

ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/KsqlRestApplication.java
lct45: That makes sense to me, probably right here: https://github.com/confluentinc/ksql/blob/master/ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/computation/CommandRunner.java#L286. WDYT? I can make that shift

ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/KsqlRestApplication.java
```diff
@@ -453,6 +463,56 @@ private void cleanupOldState() {
     localCommands.ifPresent(lc -> lc.processLocalCommandFiles(serviceContext));
   }
 
+  public void cleanupOldStateDirectories(final KsqlConfig configWithApplicationServer) {
+    final String stateDir = configWithApplicationServer.getKsqlStreamConfigProps().getOrDefault(
+            StreamsConfig.STATE_DIR_CONFIG,
+            StreamsConfig.configDef().defaultValues().get(StreamsConfig.STATE_DIR_CONFIG)).toString();
```
lct45, line 469: This default is a little confusing to me... We confirm that the state directory is available [here](https://github.com/confluentinc/ksql/blob/master/ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/KsqlServerMain.java#L112), in ksql server main, but we don't set the config if we end up using the default. This means that if we're using a state store from the default, we can't use the config to access the store name anywhere... 

I assume if we don't set the config there (if it's not already set ofc), it's so users can set the config later ..? But it does make it a little tricky to remember to use the default instead of the config. cc @swist

ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/KsqlRestApplication.java
```diff
@@ -453,6 +463,56 @@ private void cleanupOldState() {
...
+
+    final Set<String> stateStoreNames =
+        ksqlEngine.getPersistentQueries()
+            .stream()
+            .map(PersistentQueryMetadata::getQueryApplicationId)
+            .collect(Collectors.toSet());
+    try {
+      Files.walkFileTree(Paths.get(stateDir), new SimpleFileVisitor<Path>() {
```
lct45, line 477: @guozhangwang updated this, lmk if it looks good to you

ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/KsqlRestApplication.java
guozhangwang: Hmm.. I agree this is indeed a bit weird, maybe overlooked rather than intentional?

ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/KsqlRestApplication.java
```diff
@@ -453,6 +463,56 @@ private void cleanupOldState() {
     localCommands.ifPresent(lc -> lc.processLocalCommandFiles(serviceContext));
   }
 
+  public void cleanupOldStateDirectories(final KsqlConfig configWithApplicationServer) {
+    final String stateDir = configWithApplicationServer.getKsqlStreamConfigProps().getOrDefault(
+            StreamsConfig.STATE_DIR_CONFIG,
+            StreamsConfig.configDef().defaultValues().get(StreamsConfig.STATE_DIR_CONFIG)).toString();
+
+    final Set<String> stateStoreNames =
```
guozhangwang, line 471: I think it is not stateStoreNames, but queryNames?

ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/KsqlRestApplication.java
```diff
@@ -453,6 +463,56 @@ private void cleanupOldState() {
...
+          return FileVisitResult.CONTINUE;
+        }
+
+        @Override
+        public FileVisitResult visitFile(Path path, BasicFileAttributes attrs) {
+          if (!stateStoreNames.contains(path.getFileName())) {
+            try {
+              Files.delete(path);
+              log.warn("Deleted local state store for non-existing query {}. This is not expected and was likely due to a " +
+                      "race condition when the query was dropped before.", path.getFileName());
```
guozhangwang, line 490: I think we need to be forward-looking a bit since as we binpack queries into KS, the local state dir names may not be the query ids but named topology ids (cc @ableegoldman @wcarlson5 could you clarify)?

ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/KsqlRestApplication.java
guozhangwang: Left a few more comments.

ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/KsqlRestApplication.java
```diff
@@ -453,6 +463,56 @@ private void cleanupOldState() {
...
+      Files.walkFileTree(Paths.get(stateDir), new SimpleFileVisitor<Path>() {
+        @Override
+        public FileVisitResult visitFileFailed(Path path, IOException exc) {
+          log.error("Error cleaning up obsolete state directories \n", exc);
+          return FileVisitResult.CONTINUE;
+        }
+
+        @Override
+        public FileVisitResult visitFile(Path path, BasicFileAttributes attrs) {
+          if (!stateStoreNames.contains(path.getFileName())) {
```
guozhangwang, line 486: Hmm.. are we deleting at the per-query granularity (i.e. if we check the top-level query is not in the persistent query metadata, we would just delete all its sub-folders/files) or per-state-store granularity here? I think the `stateStoreNames` here is really storing query ids right?

ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/KsqlRestApplication.java
guozhangwang: Yeah that seems a better ordering to me.

ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/KsqlRestApplication.java
swist: Renamed to `queryApplication Ids`

vvcephei:
Thanks for this, @lct45 . Just checking: does this resolve https://github.com/confluentinc/ksql/issues/7720 ?

ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/KsqlRestApplication.java
guozhangwang: Is this a valid comment? Please correct me if I'm wrong.

ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/KsqlRestApplication.java
```diff
@@ -461,6 +470,69 @@ private void cleanupOldState() {
...
+              return FileVisitResult.CONTINUE;
+            }
+
+            @Override
+            public FileVisitResult visitFile(final Path path, final BasicFileAttributes attrs) {
+              if (!stateStoreNames.contains(path.getFileName())) {
+                try {
+                  Files.delete(path);
+
+                  log.warn(
```
guozhangwang, line 503: Is `path.getFileName()` gonna be query names? Or would it be possible to be at the state store or even deeper level? Ditto below for `postVisitDirectory`.

lct45:
I think it would resolve https://github.com/confluentinc/ksql/issues/7720 @vvcephei @agavra if I'm reading that correctly. It sounds like the issue @AlanConfluent reported is the same one @guozhangwang reported in https://github.com/confluentinc/ksql/pull/7729

ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/KsqlRestApplication.java
lct45: It should be both, right? I guess semantically we can call it queryNames, although it's really queryApplicationId's since those have the formatting that matches the state store names. But this list should match the directories within the top level state store folder (aside from queries that didn't get cleaned up)

ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/KsqlRestApplication.java
lct45: I don't know how this file visting system works like the first one did, but everything within the first level of the state store should match query names. Anything within those directories that _don't_ match current query names should be deleted.

So, if we hit `visitFile` for the first layer within the state store it should match the query application id, which is the list we've pulled above. Anything deeper I don't know -> makes me think we need to rethink this algo a bit. We want to delete recursively those directories in the first layer without checking to see if the files match

rodesai (commented):
Sorry for the late feedback 😬 - this kept flying under my radar.  I think we can simplify the logic here, and reuse some of the existing interfaces for cleaning up leaked state that @AlanConfluent implemented in https://github.com/confluentinc/ksql/pull/6714

At a high level, this patch needs to do the following:
1. build up a list of IDs for queries that were terminated which we may have missed cleaning up.
2. clean up the resources associated with the list of IDs we built up in (1)

For (1) I think we can use the local state dirs like you've done here. However we just need to list the contents of the state dir and compare it to the IDs in `ksqlEngine.getPersistentQueries()` - we don't need to walk the whole tree.

To do the actual clean up I think we should reuse `QueryCleanupService.QueryCleanupTask`. This means we add the state directory cleanup to that class's run method.  This will also have us cleanup leaked stores for transient queries.

lct45:
@rodesai made some of the changes you suggested but I'm not totally sure I got everything you were suggesting so if you could glance over it that would be great. @guozhangwang I moved the obsolete state directory list into the command runner but had to pass in new parameters to get everything to work. LMK if this looks good and I'll update the tests that I broke

ksqldb-engine/src/main/java/io/confluent/ksql/engine/QueryCleanupService.java
```diff
@@ -106,6 +114,16 @@ public String getAppId() {
 
     @Override
     public void run() {
+      try {
+        FileUtils.deleteDirectory(new File(stateDir + appId));
```
rodesai, line 118: is there any way to get this path out of streams instead of computing it here?

ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/computation/CommandRunner.java
```diff
@@ -284,6 +294,19 @@ public void processPriorCommands() {
           .getKsqlEngine()
           .getPersistentQueries();
 
+      final Set<String> stateStoreNames =
```
rodesai, line 297: nit: this feels a little out of place at the top level of this method, and also leaks some internal details about the underlying stream processing runtime into this class. Could we break it into it's own interface like this that we pass into this class and call here?

```
class PersistentQueryCleanup {
    void cleanupLeakedQueries(List<PersistentQueryMetadata> activeQueryIds);
}
```

rodesai (commented):
This is looking good. Just a couple minor points inline.

ksqldb-engine/src/main/java/io/confluent/ksql/engine/QueryCleanupService.java
lct45: I don't think so, and I'm not sure it would make sense. We get the list of directories to cleanup as strings from the top-level state directory. To get the corresponding path, if it exists, we'd have to take the string of the directory, map it back to the metadata or streams instance, and then get the path name.

We could compute the path name in the `command runner` and pass in the full path here as appId, but that would make the next few checks somewhat messy since we wouldn't really have an appId in that case. That would also still require the `new File`. WDYT @rodesai ?

ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/computation/CommandRunner.java
lct45: This interface would get initialized from `ksqlServerRestApplication`, right? So we'd get the `stateDir` and `serviceContext` from there? What's the benefit of doing this as an interface instead of a class, and if we do an interface is the idea that it's implemented in `ksqlServerRestApplication` or `commandRunner`?

ksqldb-engine/src/main/java/io/confluent/ksql/engine/QueryCleanupService.java
rodesai: > We could compute the path name in the command runner and pass in the full path here as appId, but that would make the next few checks somewhat messy since we wouldn't really have an appId in that case. That would also still require the new File. WDYT @rodesai ?

Yeah I agree what's already implemented is preferable. No worries I think it's ok as is - it would be an incompatible change for streams to change the way the state dir is computed anyway.

ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/computation/CommandRunner.java
rodesai: We'd implement the interface from `KsqlServerRestApplication`, which is the benefit. Ideally we should avoid writing code in this class that assumes the underlying runtime is KafkaStreams.

lct45: I'm not sure I follow - what's the benefit of doing `PersistentQueryCleanup` as an interface in `KsqlServerRestApplication` instead of a stand-alone class?

ksqldb-engine/src/main/java/io/confluent/ksql/engine/QueryCleanupService.java
```diff
@@ -106,6 +114,16 @@ public String getAppId() {
...
+        LOG.warn(
+            "Deleted local state store for non-existing query {}. "
+                + "This is not expected and was likely due to a "
+                + "race condition when the query was dropped before.",
+            appId);
+      } catch (IOException e) {
+        LOG.error("Error cleaning up state directory {}\n. {}", appId, e);
+      }
```
lct45, line 126: After running through testing I actually think this is kind of messy. We don't really _want_ to try anything else in this `run()` call, right? With the tests, it spits out a bunch of errors because it can't clean up the schema registry etc. Repurposing this feels a little weird unless we really do want those checks for any leftover persistent queries. Thoughts @rodesai ?

rodesai: Not sure I follow what the problem is. Which checks are you referring to?

ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/computation/CommandRunner.java
rodesai: I think it's a good goal to try to keep the kafka-streams specific code abstracted away from the rest of the server/engine. It keeps things decoupled which makes the code easier to understand and reason about. Putting an interface between the command runner and the streams code that does clean up makes it easier to potentially try out different runtimes in the future. We'll probably never do this (might be a fun hack day project 😄) but it's a good principle to try to structure the code this way.

ksqldb-engine/src/main/java/io/confluent/ksql/engine/QueryCleanupService.java
lct45: Below this we check if we can cleanup internal topic schemas and delete internal topics. This is what the logs end up looking like:
[2021-07-15 08:08:07,100] WARN Could not clean up the schema registry for query: fakeStateStore (io.confluent.ksql.schema.registry.SchemaRegistryUtil:72)
java.lang.NullPointerException
	at io.confluent.ksql.schema.registry.SchemaRegistryUtil.getSubjectNames(SchemaRegistryUtil.java:70)
	at io.confluent.ksql.schema.registry.SchemaRegistryUtil.getInternalSubjectNames(SchemaRegistryUtil.java:194)
	at io.confluent.ksql.schema.registry.SchemaRegistryUtil.cleanupInternalTopicSchemas(SchemaRegistryUtil.java:53)
	at io.confluent.ksql.engine.QueryCleanupService$QueryCleanupTask.lambda$run$0(QueryCleanupService.java:127)
	at io.confluent.ksql.engine.QueryCleanupService$QueryCleanupTask.tryRun(QueryCleanupService.java:144)
	at io.confluent.ksql.engine.QueryCleanupService$QueryCleanupTask.run(QueryCleanupService.java:126)
	at io.confluent.ksql.engine.QueryCleanupService.run(QueryCleanupService.java:64)
	at com.google.common.util.concurrent.AbstractExecutionThreadService$1$2.run(AbstractExecutionThreadService.java:66)
	at com.google.common.util.concurrent.Callables$4.run(Callables.java:117)
[2021-07-15 08:08:07,105] WARN Failed to cleanup internal topics for fakeStateStore (io.confluent.ksql.engine.QueryCleanupService:146)
java.lang.NullPointerException
	at io.confluent.ksql.engine.QueryCleanupService$QueryCleanupTask.lambda$run$1(QueryCleanupService.java:134)
	at io.confluent.ksql.engine.QueryCleanupService$QueryCleanupTask.tryRun(QueryCleanupService.java:144)
	at io.confluent.ksql.engine.QueryCleanupService$QueryCleanupTask.run(QueryCleanupService.java:134)
	at io.confluent.ksql.engine.QueryCleanupService.run(QueryCleanupService.java:64)
	at com.google.common.util.concurrent.AbstractExecutionThreadService$1$2.run(AbstractExecutionThreadService.java:66)
	at com.google.common.util.concurrent.Callables$4.run(Callables.java:117)
[2021-07-15 08:08:07,108] WARN Failed to cleanup internal consumer groups for fakeStateStore (io.confluent.ksql.engine.QueryCleanupService:146)

ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/computation/CommandRunner.java
lct45: I made a pass at doing an interface - I can't tell if it's what you want. I get that we want to keep streams code abstracted, but it's odd to me that we'd want to implement the `PersistentQueryCleanup` inside of the rest app if we're trying to keep it separate, it makes more sense to me to create a standalone class for `PersistenteQueryCleanup` that the rest app instantiates and then passes to the command runner, who then calls a method in the standalone class, rather than doing a callback to a method inside of the rest app (which is what called the command runner in the first place).

The other option for interface would be to have `KsqlRestApplication` as a whole implement `PersistentQueryCleanup` but this seemed like it would make testing harder since there wouldn't be a simple constructor for a `PersistentQueryCleanup` object. LMK your thoughts

rodesai: > but it's odd to me that we'd want to implement the PersistentQueryCleanup inside of the rest app if we're trying to keep it separate, it makes more sense to me to create a standalone class for PersistenteQueryCleanup that the rest app instantiates and then passes to the command runner

Either way is fine by me - putting it in a standalone class is probably cleaner. It would be good to have an interface the standalone class implements to keep the implementation completely separate.

ksqldb-engine/src/main/java/io/confluent/ksql/engine/QueryCleanupService.java
rodesai: The cleanup task should do it's best to clean up everything that could possibly be left behind locally and in kafka/sr for the query. Are you worried about the case where the schema/group/topic doesn't exist? The cleanup code should be able to handle the case where something it's supposed to clean up doesn't exist by just logging-and-continuing. If the resource doesn't exist anymore there's nothing to clean up - so the state is what we want it to be.

In this case I'd guess that the test setup has some problem which causes us to throw an NPE where we would never actually throw an NPE when actually running ksql. Do we know what value is getting set to null?

rodesai: In this case I think it's the `context` you pass to `PersistentQueryCleanup` from `PersistentQueryCleanupTest` - you'd need to setup that `context` to return a mock schema registry client and admin client.

lct45: Ahhhhhhh okay that makes more sense, this cleanup order should be good then

rodesai (approved):
LGTM!

guozhangwang:
Sorry for getting late on the final pass of this, I also like the new approach better -- doing that in the background task than tying up the starting app thread. LGTM!

PR #7738, opened 2021-06-28
feat: pull query bandwidth based throttling

cprasad1:
### Description 
Limits the amount of bandwidth used by Pull Queries using a Sliding Window Log. The limit (in MB per hour) can be set using the `ksql.query.pull.max.hourly.bandwidth` config. Once the limit is hit, pull queries will fail immediately.

### Testing done 
- Manual Testing
- Unit testing
- Integration testing

### Reviewer checklist

### DIFF
--- a/ksqldb-common/src/main/java/io/confluent/ksql/util/KsqlConfig.java
+++ b/ksqldb-common/src/main/java/io/confluent/ksql/util/KsqlConfig.java
@@ -251,6 +251,13 @@ public class KsqlConfig extends AbstractConfig {
       "The maximum number of concurrent requests allowed for pull "
       + "queries on this host. Once the limit is hit, queries will fail immediately";
 
+  public static final String KSQL_QUERY_PULL_MAX_HOURLY_BANDWIDTH_CONFIG
+      = "ksql.query.pull.max.hourly.bandwidth";
+  public static final Integer KSQL_QUERY_PULL_MAX_HOURLY_BANDWIDTH_DEFAULT = Integer.MAX_VALUE;
+  public static final String KSQL_QUERY_PULL_MAX_HOURLY_BANDWIDTH_DOC
+      = "The maximum amount of pull query bandwidth allowed over a period of one hour."
+      + "Once the limit is hit, queries will fail immediately";
+
   public static final String KSQL_QUERY_PULL_THREAD_POOL_SIZE_CONFIG
       = "ksql.query.pull.thread.pool.size";
   public static final Integer KSQL_QUERY_PULL_THREAD_POOL_SIZE_DEFAULT = 100;
@@ -867,6 +874,13 @@ public class KsqlConfig extends AbstractConfig {
             KSQL_QUERY_PULL_MAX_CONCURRENT_REQUESTS_DEFAULT,
             Importance.LOW,
             KSQL_QUERY_PULL_MAX_CONCURRENT_REQUESTS_DOC
+        )
+        .define(
+            KSQL_QUERY_PULL_MAX_HOURLY_BANDWIDTH_CONFIG,
+            Type.INT,
+            KSQL_QUERY_PULL_MAX_HOURLY_BANDWIDTH_DEFAULT,
+            Importance.HIGH,
+            KSQL_QUERY_PULL_MAX_HOURLY_BANDWIDTH_DOC
         )
         .define(
             KSQL_QUERY_PULL_THREAD_POOL_SIZE_CONFIG,
--- a/ksqldb-common/src/main/java/io/confluent/ksql/util/KsqlConfig.java
+++ b/ksqldb-common/src/main/java/io/confluent/ksql/util/KsqlConfig.java
@@ -251,12 +251,13 @@ public class KsqlConfig extends AbstractConfig {
       "The maximum number of concurrent requests allowed for pull "
       + "queries on this host. Once the limit is hit, queries will fail immediately";
 
-  public static final String KSQL_QUERY_PULL_MAX_HOURLY_BANDWIDTH_CONFIG
-      = "ksql.query.pull.max.hourly.bandwidth";
-  public static final Integer KSQL_QUERY_PULL_MAX_HOURLY_BANDWIDTH_DEFAULT = Integer.MAX_VALUE;
-  public static final String KSQL_QUERY_PULL_MAX_HOURLY_BANDWIDTH_DOC
-      = "The maximum amount of pull query bandwidth allowed over a period of one hour."
-      + "Once the limit is hit, queries will fail immediately";
+  public static final String KSQL_QUERY_PULL_MAX_HOURLY_BANDWIDTH_MEGABYTES_CONFIG
+      = "ksql.query.pull.max.hourly.bandwidth.megabytes";
+  public static final Integer KSQL_QUERY_PULL_MAX_HOURLY_BANDWIDTH_MEGABYTES_DEFAULT
+      = Integer.MAX_VALUE;
+  public static final String KSQL_QUERY_PULL_MAX_HOURLY_BANDWIDTH_MEGABYTES_DOC
+      = "The maximum amount of pull query bandwidth in megabytes allowed over"
+      + " a period of one hour. Once the limit is hit, queries will fail immediately";
 
   public static final String KSQL_QUERY_PULL_THREAD_POOL_SIZE_CONFIG
       = "ksql.query.pull.thread.pool.size";
@@ -876,11 +877,11 @@ public class KsqlConfig extends AbstractConfig {
             KSQL_QUERY_PULL_MAX_CONCURRENT_REQUESTS_DOC
         )
         .define(
-            KSQL_QUERY_PULL_MAX_HOURLY_BANDWIDTH_CONFIG,
+            KSQL_QUERY_PULL_MAX_HOURLY_BANDWIDTH_MEGABYTES_CONFIG,
             Type.INT,
-            KSQL_QUERY_PULL_MAX_HOURLY_BANDWIDTH_DEFAULT,
+            KSQL_QUERY_PULL_MAX_HOURLY_BANDWIDTH_MEGABYTES_DEFAULT,
             Importance.HIGH,
-            KSQL_QUERY_PULL_MAX_HOURLY_BANDWIDTH_DOC
+            KSQL_QUERY_PULL_MAX_HOURLY_BANDWIDTH_MEGABYTES_DOC
         )
         .define(
             KSQL_QUERY_PULL_THREAD_POOL_SIZE_CONFIG,
--- a/ksqldb-rest-app/src/main/java/io/confluent/ksql/api/impl/QueryEndpoint.java
+++ b/ksqldb-rest-app/src/main/java/io/confluent/ksql/api/impl/QueryEndpoint.java
@@ -70,7 +70,6 @@ import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.atomic.AtomicReference;
 import java.util.function.Consumer;
 import java.util.stream.Collectors;
-import org.apache.kafka.common.utils.Time;
 
 // CHECKSTYLE_RULES.OFF: ClassDataAbstractionCoupling
 public class QueryEndpoint {
@@ -217,7 +216,6 @@ public class QueryEndpoint {
             PullQueryResult::getPlanType).orElse(PullPhysicalPlanType.UNKNOWN);
         final RoutingNodeType routingNodeType = Optional.ofNullable(r).map(
             PullQueryResult::getRoutingNodeType).orElse(RoutingNodeType.UNKNOWN);
-        pullBandRateLimiter.add(startTimeNanos / 1000000000, responseBytes);
         metrics.recordResponseSize(responseBytes, sourceType, planType, routingNodeType);
         metrics.recordLatency(startTimeNanos, sourceType, planType, routingNodeType);
         metrics.recordRowsReturned(
@@ -226,6 +224,7 @@ public class QueryEndpoint {
         metrics.recordRowsProcessed(
             Optional.ofNullable(r).map(PullQueryResult::getTotalRowsProcessed).orElse(0L),
             sourceType, planType, routingNodeType);
+        pullBandRateLimiter.add(responseBytes);
       });
     });
 
@@ -242,7 +241,7 @@ public class QueryEndpoint {
 
     PullQueryExecutionUtil.checkRateLimit(rateLimiter);
     final Decrementer decrementer = pullConcurrencyLimiter.increment();
-    pullBandRateLimiter.allow(Time.SYSTEM.milliseconds());
+    pullBandRateLimiter.allow();
 
     try {
       final PullQueryResult result = ksqlEngine.executePullQuery(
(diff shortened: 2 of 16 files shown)


ksqldb-rest-app/src/main/java/io/confluent/ksql/api/impl/QueryEndpoint.java
```diff
@@ -210,6 +217,7 @@ private QueryPublisher createPullQueryPublisher(
             PullQueryResult::getPlanType).orElse(PullPhysicalPlanType.UNKNOWN);
         final RoutingNodeType routingNodeType = Optional.ofNullable(r).map(
             PullQueryResult::getRoutingNodeType).orElse(RoutingNodeType.UNKNOWN);
+        pullBandRateLimiter.add(startTimeNanos / 1000000000, responseBytes);
```
AlanConfluent, line 220: Declare the literal as a constant or use `TimeUnit.NANOS.toMillis()`

ksqldb-rest-app/src/main/java/io/confluent/ksql/api/server/SlidingWindowRateLimiter.java
```diff
@@ -0,0 +1,66 @@
...
+
+import static io.confluent.ksql.util.KsqlPreconditions.checkArgument;
+
+import io.confluent.ksql.util.KsqlException;
+import io.confluent.ksql.util.Pair;
+import java.util.LinkedList;
+import java.util.Queue;
+
+public class SlidingWindowRateLimiter {
+  private static long NUM_BYTES_IN_ONE_MEGABYTE = 1048576;
```
AlanConfluent, line 26: It's a bit easier to read if you write this as `1 * 1024 * 1024` or something like that.  The compiler will obviously turn it into the above, but it makes it easier to change the constant.

ksqldb-rest-app/src/main/java/io/confluent/ksql/api/server/SlidingWindowRateLimiter.java
```diff
@@ -0,0 +1,66 @@
...
+  private final Queue<Pair<Long, Long>> responseSizesLog;
+  private final long throttleLimit;                            // throttle limit in Bytes
+  private final long slidingWindowSize;                        // window size in miliseconds
```
AlanConfluent, line 29: "milliseconds"

ksqldb-rest-app/src/main/java/io/confluent/ksql/api/server/SlidingWindowRateLimiter.java
```diff
@@ -0,0 +1,66 @@
...
+  private long numBytesInWindow;                               // bandwidth in the last window
+
+  public SlidingWindowRateLimiter(final int requestLimitInMB, final long slidingWindowSize) {
```
AlanConfluent, line 32: Maybe rename `slidingWindowSize` to `slidingWindowSizeMs` to make clear the unit you're expecting and change the field name too.

ksqldb-rest-app/src/main/java/io/confluent/ksql/api/server/SlidingWindowRateLimiter.java
```diff
@@ -0,0 +1,66 @@
...
+    checkArgument(slidingWindowSize >= 0,
+            "Pull Query throttle window size can't be negative");
+
+    this.throttleLimit = (long) requestLimitInMB * NUM_BYTES_IN_ONE_MEGABYTE;
+    this.slidingWindowSize = slidingWindowSize;
+    this.responseSizesLog = new LinkedList<>();
+    this.numBytesInWindow = 0;
+  }
+
+  public synchronized void allow(final long timestamp) throws KsqlException {
```
AlanConfluent, line 44: I know taking in a timestamp makes it a little easier to test, but it's better practice to inject a Clock in the constructor and just mock that in the test.  That way the common method interface is simpler and someone cannot pass in a bad value.

ksqldb-rest-app/src/main/java/io/confluent/ksql/api/server/SlidingWindowRateLimiter.java
```diff
@@ -0,0 +1,66 @@
...
+import static io.confluent.ksql.util.KsqlPreconditions.checkArgument;
+
+import io.confluent.ksql.util.KsqlException;
+import io.confluent.ksql.util.Pair;
+import java.util.LinkedList;
+import java.util.Queue;
+
+public class SlidingWindowRateLimiter {
+  private static long NUM_BYTES_IN_ONE_MEGABYTE = 1048576;
+  private final Queue<Pair<Long, Long>> responseSizesLog;
```
AlanConfluent, line 27: Document the key and value since they're just longs.

ksqldb-rest-app/src/main/java/io/confluent/ksql/api/server/SlidingWindowRateLimiter.java
```diff
@@ -0,0 +1,66 @@
...
+    while (!responseSizesLog.isEmpty()
+            && timestamp - responseSizesLog.peek().left >= slidingWindowSize) {
+      this.numBytesInWindow -= responseSizesLog.poll().right;
+    }
+    if (this.numBytesInWindow > throttleLimit) {
+      throw new KsqlException("Host is at bandwidth rate limit for pull queries.");
+    }
+  }
+
+  public synchronized void add(final long timestamp, final long responseSizeInBytes) {
```
AlanConfluent, line 57: We should probably add a bit of documentation to the class and methods.  Many rate limiters require you to ask for access before it's granted whereas this method always records access (post-facto) but asks that you check first via allow if previous calls put you in debt.  This is due to not knowing the cost of the request upfront.  These are just a few things to explain about this version of a rate limiter.

ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/KsqlRestApplication.java
```diff
@@ -745,7 +751,9 @@ static KsqlRestApplication buildApplication(
     final ConcurrencyLimiter pullQueryConcurrencyLimiter = new ConcurrencyLimiter(
         ksqlConfig.getInt(KsqlConfig.KSQL_QUERY_PULL_MAX_CONCURRENT_REQUESTS_CONFIG),
         "pull queries");
-
+    final SlidingWindowRateLimiter pullBandRateLimiter = new SlidingWindowRateLimiter(
+            ksqlConfig.getInt(KsqlConfig.KSQL_QUERY_PULL_MAX_HOURLY_BANDWIDTH_CONFIG),
+            3600 * 1000);
```
AlanConfluent, line 756: `3600 * 1000` might be good to either put in a constant here or maybe even in another config if we think we might adjust how this is enforced.

ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/resources/streaming/StreamedQueryResource.java
```diff
@@ -346,6 +356,7 @@ private EndpointResponse handlePullQuery(
             PullQueryResult::getPlanType).orElse(PullPhysicalPlanType.UNKNOWN);
         final RoutingNodeType routingNodeType = Optional.ofNullable(r).map(
             PullQueryResult::getRoutingNodeType).orElse(RoutingNodeType.UNKNOWN);
+        pullBandRateLimiter.add(startTimeNanos / 1000000000, responseBytes);
```
AlanConfluent, line 359: Constant, as above.

ksqldb-rest-app/src/main/java/io/confluent/ksql/api/impl/QueryEndpoint.java
AlanConfluent: I was originally slightly worried about piggybacking on the metrics because I didn't want to throw an exception and break our metrics reporting, but the way we've implemented this, it's actually guaranteed to add the response, so I think this works well.

ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/resources/streaming/PullQueryPublisher.java
```diff
@@ -99,6 +106,7 @@ public synchronized void subscribe(final Subscriber<Collection<StreamedRow>> sub
 
     PullQueryExecutionUtil.checkRateLimit(rateLimiter);
     final Decrementer decrementer = concurrencyLimiter.increment();
+    pullBandRateLimiter.allow(Time.SYSTEM.milliseconds());
```
AlanConfluent, line 109: We're still trying to figure out how to get the bytes here, right?

If we just retain the qps and concurrency rate limiter across all endpoints, and we want to fall back on that here, then that might be sufficient, but if we wanted to lift the other rate limiting on the other endpoints in favor of bandwidth, then we couldn't just disable it because it would disable it here too.

ksqldb-rest-app/src/test/java/io/confluent/ksql/api/server/SlidingWindowRateLimiterTest.java
```diff
@@ -0,0 +1,84 @@
...
+            fail(TEST_SHOULD_NOT_FAIL);
+        }
+    }
+
+    @Test
+    public void uniformResponsesOverLimit() {
+        Throwable exception = assertThrows(KsqlException.class, () -> {
+            for (int i = 0; i < 11; i += 1) {
+                limiter.add(i * 400, 100000);
+                limiter.allow(i * 400 + 1);
```
AlanConfluent, line 50: It might be useful to take note of the first timestamp that fails and assert it so we know it goes over the limit at the right time.

ksqldb-rest-app/src/test/java/io/confluent/ksql/api/integration/PullBandwidthThrottleIntegrationTest.java
```diff
@@ -0,0 +1,221 @@
...
+                    EmbeddedSingleNodeKafkaCluster.newBuilder()
+                            .withoutPlainListeners()
+                            .withSaslSslListeners()
+            ).build();
+
+    private static final TestKsqlRestApp REST_APP = TestKsqlRestApp
+            .builder(TEST_HARNESS::kafkaBootstrapServers)
+            .withProperty("security.protocol", "SASL_SSL")
+            .withProperty("sasl.mechanism", "PLAIN")
+            .withProperty("sasl.jaas.config", SecureKafkaHelper.buildJaasConfig(NORMAL_USER))
```
AlanConfluent, line 84: Just FYI, some of these jaas configs aren't strictly necessary for this test, though it doesn't hurt.

ksqldb-common/src/main/java/io/confluent/ksql/util/KsqlConfig.java
```diff
@@ -251,6 +251,13 @@
       "The maximum number of concurrent requests allowed for pull "
       + "queries on this host. Once the limit is hit, queries will fail immediately";
 
+  public static final String KSQL_QUERY_PULL_MAX_HOURLY_BANDWIDTH_CONFIG
+      = "ksql.query.pull.max.hourly.bandwidth";
+  public static final Integer KSQL_QUERY_PULL_MAX_HOURLY_BANDWIDTH_DEFAULT = Integer.MAX_VALUE;
+  public static final String KSQL_QUERY_PULL_MAX_HOURLY_BANDWIDTH_DOC
+      = "The maximum amount of pull query bandwidth allowed over a period of one hour."
```
AlanConfluent, line 258: You should specify that this is in MB in the comments.  For other disk limits, btw, it seems that we just do things in bytes.  Might want to do that for consistency.  If you do make that change, you probably want to change the config type to a long.  Also, it might make sense to tack on to the config name the unit, e.g. "ksql.query.pull.max.hourly.bandwidth.bytes" or "ksql.query.pull.max.hourly.bytes"

AlanConfluent (approved):

ksqldb-rest-app/src/main/java/io/confluent/ksql/api/server/SlidingWindowRateLimiter.java
```diff
@@ -0,0 +1,116 @@
...
+  }
+
+  @VisibleForTesting
+  protected synchronized void add(final long timestamp, final long responseSizeInBytes) {
+    checkArgument(timestamp >= 0,
+            "Timestamp can't be negative.");
+    checkArgument(responseSizeInBytes >= 0,
+            "Response size can't be negative.");
+
+    responseSizesLog.add(new Pair<>(timestamp, responseSizeInBytes));
```
guozhangwang, line 113: Just wearing my paranoid window processing interviewer hat here :) Just in case we may have a time shift from `Time.SYSTEM.milliseconds()`, maybe do a quick sorted insert from the head than blinding adding at the head? In practice most of it would just end at the head so perf-wise it should be fine.

ksqldb-rest-app/src/main/java/io/confluent/ksql/api/impl/QueryEndpoint.java
```diff
@@ -218,6 +224,7 @@ private QueryPublisher createPullQueryPublisher(
         metrics.recordRowsProcessed(
             Optional.ofNullable(r).map(PullQueryResult::getTotalRowsProcessed).orElse(0L),
             sourceType, planType, routingNodeType);
+        pullBandRateLimiter.add(responseBytes);
```
guozhangwang, line 227: We would need to call `system time` to record metrics above, maybe we can just get one system time once to save the number of currentSystimeMillis?

guozhangwang: Ditto for other `add` calls where we would potentially get `systemTime` multiple times.

guozhangwang (commented):
Hey @cprasad1 sorry I'm late on this. I made a quick pass on the non-testing part and just left some quick comments. Please feel free to ignore as the PR is already merged anyways.

ksqldb-rest-app/src/main/java/io/confluent/ksql/api/server/SlidingWindowRateLimiter.java
cprasad1: Since the purpose of the RateLimiter was to bound the bandwidth usage approximately over the span of an hour (on that time-scale, I don't see a significant time shift happening), I think the current algorithm will work just fine. Do you think it is worth making this optimization?

ksqldb-rest-app/src/main/java/io/confluent/ksql/api/impl/QueryEndpoint.java
cprasad1: Keeping the logic of getting the `systemTime` internal to the `pullBandRateLimiter` seemed like a cleaner design. That way the common method interface is simpler and someone cannot pass in a bad value. It does come at the expense of potentially multiple redundant function calls. 

It is also an attempt at decoupling the metrics reporting logic and bandwidth limiting logic. We are already piggy-backing on the metrics reporting framework for limiting the bandwidth, and I did not want further coupling between the two.

PR #7739, opened 2021-06-29
fix: enable KAFKA-10847 bugfix when GRACE PERIOD is used on Joins

spena:
### Description 
Now that https://github.com/confluentinc/ksql/pull/7695 is merged, this PR enables the stream-stream left/outer joins fix implemented in https://issues.apache.org/jira/browse/KAFKA-10847.

To enabled it, I removed the `ENABLE_KSTREAMS_OUTER_JOIN_SPURIOUS_RESULTS_FIX` compatibility flag (which was set to False) from `KsqlConfig.java` so that all QTT tests and joins commands use the bugfix. The flag is enabled by default, so there is no need to enable it in `KsqlConfig.java`.

I also call the new `JoinWindows.ofTimeDifferenceAndGrace()` API when the `GRACE PERIOD` clause is used in Joins. The new API has the bugfix enabled internally.

BREAKING CHANGE: If the user sets GRACE PERIOD explicitly in stream-stream left/outer joins, the grace period defines when the left/outer non-joined results are emitted. Contrary to the old syntax (without GRACE PERIOD) that emits left/outer non-joined results eagerly, which may cause "spurious" result records. Using the GRACE PERIOD syntax enables the bugfix implemented in KAFKA-10847.

### Testing done 
Updated QTT tests

### Reviewer checklist

### DIFF
--- a/docs/operate-and-deploy/installation/upgrading.md
+++ b/docs/operate-and-deploy/installation/upgrading.md
@@ -137,6 +137,24 @@ This will stop all processing and delete any internal topics in Kafka.
 
 ## Upgrade notes
 
+### Upgrading from ksqlDB 0.14.0 to 0.20.0
+
+In-place upgrades are supported from ksqlDB 0.14.0 to 0.20.0.
+See the [changelog](https://github.com/confluentinc/ksql/blob/master/CHANGELOG.md)
+for potential breaking changes that may affect the behavior or required syntax
+for new queries.
+
+#### Join statements
+
+The GRACE PERIOD clause is now supported in stream-stream joins since 0.20.0. However, when set
+explicitly, the grace period now defines when the left/outer non-joined results are emitted.
+Contrary to the old syntax (without GRACE PERIOD) that emits left/outer non-joined results eagerly,
+which may cause "spurious" result records.
+
+The use of GRACE PERIOD is highly recommended now that it is supported. It helps to reduce high
+disk usage when using small grace period values (default 24 hours if not set) and also provides
+better semantics for left/outer joins.
+
 ### Upgrading from ksqlDB 0.10.0 to 0.14.0
 
 In-place upgrades are supported from ksqlDB 0.10.0 to 0.14.0.
--- a/ksqldb-common/src/main/java/io/confluent/ksql/util/KsqlConfig.java
+++ b/ksqldb-common/src/main/java/io/confluent/ksql/util/KsqlConfig.java
@@ -16,7 +16,6 @@
 package io.confluent.ksql.util;
 
 import static io.confluent.ksql.configdef.ConfigValidators.zeroOrPositive;
-import static org.apache.kafka.streams.StreamsConfig.InternalConfig.ENABLE_KSTREAMS_OUTER_JOIN_SPURIOUS_RESULTS_FIX;
 
 import com.google.common.base.Splitter;
 import com.google.common.collect.ImmutableList;
@@ -547,12 +546,7 @@ public class KsqlConfig extends AbstractConfig {
           new CompatibilityBreakingStreamsConfig(
               StreamsConfig.TOPOLOGY_OPTIMIZATION_CONFIG,
               StreamsConfig.OPTIMIZE,
-              StreamsConfig.OPTIMIZE),
-          // Disable KAFKA-10847 for now until grace period is made configurable
-          new CompatibilityBreakingStreamsConfig(
-              ENABLE_KSTREAMS_OUTER_JOIN_SPURIOUS_RESULTS_FIX,
-              false,
-              false)
+              StreamsConfig.OPTIMIZE)
   );
 
   private static final class CompatibilityBreakingStreamsConfig {
@@ -563,7 +557,7 @@ public class KsqlConfig extends AbstractConfig {
     CompatibilityBreakingStreamsConfig(final String name, final Object defaultValueLegacy,
         final Object defaultValueCurrent) {
       this.name = Objects.requireNonNull(name);
-      if (!StreamsConfig.configDef().names().contains(name) && !isInternal(name)) {
+      if (!StreamsConfig.configDef().names().contains(name)) {
         throw new IllegalArgumentException(
             String.format("%s is not a valid streams config", name));
       }
@@ -573,10 +567,6 @@ public class KsqlConfig extends AbstractConfig {
 
     String getName() {
       return this.name;
-    }
-
-    private static boolean isInternal(final String name) {
-      return name.equals(ENABLE_KSTREAMS_OUTER_JOIN_SPURIOUS_RESULTS_FIX);
     }
   }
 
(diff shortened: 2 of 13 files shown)


ksqldb-common/src/main/java/io/confluent/ksql/util/KsqlConfig.java
```diff
@@ -16,7 +16,6 @@
 package io.confluent.ksql.util;
 
 import static io.confluent.ksql.configdef.ConfigValidators.zeroOrPositive;
-import static org.apache.kafka.streams.StreamsConfig.InternalConfig.ENABLE_KSTREAMS_OUTER_JOIN_SPURIOUS_RESULTS_FIX;
```
guozhangwang, line 19: qq: Since this is introduced in KS only for ksqlDB, could we remove the config in KS as well? If yes I will file a quick PR.

guozhangwang (approved):
LGTM!!

ksqldb-common/src/main/java/io/confluent/ksql/util/KsqlConfig.java
spena: Not sure. @mjsax should we remove the config from KS now that the fix is automatically enabled when using ofTimeAndGracePeriod()?

docs/operate-and-deploy/installation/upgrading.md
```diff
@@ -137,6 +137,24 @@ This will stop all processing and delete any internal topics in Kafka.
...
+
+The GRACE PERIOD clause is now supported in stream-stream joins since 0.20.0. However, when set
+explicitly, the grace period now defines when the left/outer non-joined results are emitted.
+Contrary to the old syntax (without GRACE PERIOD) that emits left/outer non-joined results eagerly,
+which may cause "spurious" result records.
+
+The use of GRACE PERIOD is highly recommended now that it is supported. It helps to reduce high
+disk usage when using small grace period values (default 24 hours if not set) and also provides
+better semantics for left/outer joins.
+
```
spena, line 157: @JimGalasyn @mjsax I added a note for upgrading

guozhangwang:
Corresponding AK PR for removing the internal flag: https://github.com/apache/kafka/pull/10941

docs/operate-and-deploy/installation/upgrading.md
```diff
@@ -137,6 +137,24 @@ This will stop all processing and delete any internal topics in Kafka.
 
 ## Upgrade notes
 
+### Upgrading from ksqlDB 0.14.0 to 0.20.0
```
mjsax, line 140: why 0.14.0 ?

ksqldb-common/src/main/java/io/confluent/ksql/util/KsqlConfig.java
mjsax: Don't have a strong opinion. But gives that we piggy-back opt-in/opt-out into the new semantics via API/language, it seems ok to remove both.

ksqldb-functional-tests/src/test/resources/query-validation-tests/joins.json
```diff
@@ -895,9 +895,7 @@
         {"topic": "right_topic", "key": 3, "value": {"L2": "d"}, "timestamp": 60000}
       ],
       "outputs": [
-        {"topic": "LEFT_JOIN", "key": 0, "value": {"L1": "A", "L2": null}, "timestamp": 0},
         {"topic": "LEFT_JOIN", "key": 0, "value": {"L1": "A", "L2": "a"}, "timestamp": 60000},
-        {"topic": "LEFT_JOIN", "key": 1, "value": {"L1": "B", "L2": null}, "timestamp": 330000},
         {"topic": "LEFT_JOIN", "key": 2, "value": {"L1": "C", "L2": null}, "timestamp": 90000},
```
mjsax, line 899: Why do we still have this result record?

ksqldb-functional-tests/src/test/resources/query-validation-tests/joins.json
```diff
@@ -895,9 +895,7 @@
         {"topic": "right_topic", "key": 3, "value": {"L2": "d"}, "timestamp": 60000}
       ],
       "outputs": [
-        {"topic": "LEFT_JOIN", "key": 0, "value": {"L1": "A", "L2": null}, "timestamp": 0},
         {"topic": "LEFT_JOIN", "key": 0, "value": {"L1": "A", "L2": "a"}, "timestamp": 60000},
-        {"topic": "LEFT_JOIN", "key": 1, "value": {"L1": "B", "L2": null}, "timestamp": 330000},
```
mjsax, line 900: Can we add one more input record, to bump stream-time such that this left-join result will be eventually emitted?

ksqldb-functional-tests/src/test/resources/query-validation-tests/joins.json
```diff
@@ -921,9 +919,7 @@
         {"topic": "left_topic", "key": 3, "value": {"L1": "D"}, "timestamp": 60000}
       ],
       "outputs": [
-        {"topic": "OUTER_JOIN", "key": 0, "value": {"T_ID": 0, "TT_ID": null, "L1": "A", "L2": null}, "timestamp": 0},
         {"topic": "OUTER_JOIN", "key": 0, "value": {"T_ID": 0, "TT_ID": 0, "L1": "A", "L2": "a"}, "timestamp": 60000},
-        {"topic": "OUTER_JOIN", "key": 1, "value": {"T_ID": null, "TT_ID": 1, "L1": null, "L2": "b"}, "timestamp": 330000},
```
mjsax, line 926: As above. Also, can we add a "right outer" join result, too?

docs/operate-and-deploy/installation/upgrading.md
spena: Just followed the pattern from the below upgrading paths. I don't know if just show 0.19 -> 0.20 and use the latest path show below.

ksqldb-functional-tests/src/test/resources/query-validation-tests/joins.json
spena: This is part of the issue about grace periods we talked before, where an expired record is still added to the window. The left/outer join, though, has the logic to emit expired records, but after the expired record is still added to the normal join window later.

ksqldb-functional-tests/src/test/resources/query-validation-tests/joins.json
spena: Added

ksqldb-functional-tests/src/test/resources/query-validation-tests/joins.json
spena: Added

ksqldb-functional-tests/src/test/resources/query-validation-tests/joins.json
mjsax: Ah. Good to know. -- Guess we need to fix this in a follow up... But the bug is there since day one, so not a big deal I guess.

mjsax (approved):

PR #7745, opened 2021-06-29
feat: record KSQL processing ratio per cluster

lct45:
### Description 
To facilitate additional user observability, we want to capture a compute utilization metric. As part of that, we want to calculate the overall % of time a node spent processing versus blocked. 

### Testing done 
Unit tests for math correctness, benchmarks for accuracy

### Reviewer checklist


ksqldb-engine/src/main/java/io/confluent/ksql/internal/UtilizationMetricsListener.java
```diff
@@ -0,0 +1,192 @@
...
+            final Map<String, Double> threadMetrics = allMetrics.stream()
+                    .filter(m -> m.metricName().group().equals(STREAM_METRICS) &&
+                            m.metricName().tags().getOrDefault(THREAD_ID, "").equals(threadName) &&
+                            metrics.contains(m.metricName().name()))
+                    .collect(Collectors.toMap(k -> k.metricName().name(), v -> (Double) v.metricValue()));
+
+            final List<Metric> startTimeList = allMetrics.stream()
+                    .filter(m -> m.metricName().group().equals(STREAM_METRICS) &&
+                            m.metricName().tags().getOrDefault(THREAD_ID, "").equals(threadName) &&
+                            m.metricName().name().equals("thread-start-time"))
```
guozhangwang, line 142: Do we have this metric in KS really?

ksqldb-engine/src/main/java/io/confluent/ksql/internal/UtilizationMetricsListener.java
```diff
@@ -0,0 +1,192 @@
...
+            final Long threadStartTime = startTimeList.size() != 0 ? (Long) startTimeList.get(0).metricValue() : 0L;
+
+            double blockedTime = 0.0;
+            if (threadStartTime > windowStart) {
+                blockedTime += threadStartTime - windowStart;
+            }
+            final double newFlushTime = threadMetrics.getOrDefault("flush-time-total", windowSize);
+            double newConsumerIOTime = consumerMetrics.getOrDefault("io-waittime-total", windowSize)
+                + consumerMetrics.getOrDefault("iotime-total", windowSize)
+                + restoreConsumerMetrics.getOrDefault("io-waittime-total", windowSize)
```
guozhangwang, line 153: Okay, I think I miss sth. here since these consumer metrics are not there in AK, did we add them ourselves? If yes could you point me to these metrics?

ksqldb-engine/src/main/java/io/confluent/ksql/internal/UtilizationMetricsListener.java
```diff
@@ -0,0 +1,192 @@
...
+                .collect(Collectors.toMap(k -> k.metricName().name(), v -> (Double) v.metricValue()));
+
+            final Map<String, Double> restoreConsumerMetrics = allMetrics.stream()
+                .filter(m -> m.metricName().group().equals(CONSUMER_METRICS)
+                    && metrics.contains(m.metricName().name())
+                    && m.metricName().tags().getOrDefault("client-id", "").contains("restore-consumer")
+                    && m.metricName().tags().getOrDefault("client-id", "").contains(threadName))
+                .collect(Collectors.toMap(k -> k.metricName().name(), v -> (Double) v.metricValue()));
+
+            final Map<String, Double> producerMetrics = allMetrics.stream()
```
guozhangwang, line 127: I think we should consider admin client's blocking portion as well?

ksqldb-engine/src/main/java/io/confluent/ksql/internal/UtilizationMetricsListener.java
```diff
@@ -0,0 +1,192 @@
...
+        } catch (Exception e) {
+            logger.error("oops: ", e);
+            throw e;
+        }
+    }
+
+    private void initializeMetrics() {
+        metrics.add("io-waittime-total");
+        metrics.add("iotime-total");
+        metrics.add("bufferpool-wait-time-total");
```
guozhangwang, line 185: Besides this one in producer, all other metrics seem unfamiliar to me. Could you point to me where do you find them, plus `thread-start-time`?

ksqldb-engine/src/main/java/io/confluent/ksql/internal/UtilizationMetricsListener.java
rodesai: It's a bit hard to tell how the metric names are assembled, but these are the metrics collected here:
https://github.com/confluentinc/kafka/blob/4fea624446b3e3fc889ca8f5938de2497ed5e091/clients/src/main/java/org/apache/kafka/common/network/Selector.java#L467
https://github.com/confluentinc/kafka/blob/4fea624446b3e3fc889ca8f5938de2497ed5e091/clients/src/main/java/org/apache/kafka/common/network/Selector.java#L492

They measure the total time spent in `select` waiting for data, and reading/writing data from/to the network, respectively.

We went with these lower-level metrics because the original metrics we tried to use, which wrapped all of `poll` and `send` were too coarse-grained. They included significant time spent calling interceptors and doing compression (for `send`), which we don't want to count as blocked, because we're using CPU to run the interceptors/compress.

ksqldb-engine/src/main/java/io/confluent/ksql/internal/UtilizationMetricsListener.java
rodesai: What does the streams thread use the admin client for during regular processing?

ksqldb-engine/src/main/java/io/confluent/ksql/internal/UtilizationMetricsListener.java
guozhangwang: I know these metrics, but the names are not exactly the same: e.g. there are `io-wait-time-ns-avg` / `io-time-ns-avg` but there are no `io-waittime-total` / `iotime-total`

ksqldb-engine/src/main/java/io/confluent/ksql/internal/UtilizationMetricsListener.java
guozhangwang: During rebalances, we use admin clients to get end offsets, and topic metadata, which are all blocking calls.

ksqldb-engine/src/main/java/io/confluent/ksql/internal/UtilizationMetricsListener.java
rodesai: It gets created by `Meter` (https://github.com/confluentinc/kafka/blob/4fea624446b3e3fc889ca8f5938de2497ed5e091/clients/src/main/java/org/apache/kafka/common/metrics/stats/Meter.java). The name is computed here: https://github.com/confluentinc/kafka/blob/4fea624446b3e3fc889ca8f5938de2497ed5e091/clients/src/main/java/org/apache/kafka/common/network/Selector.java#L1279

ksqldb-engine/src/main/java/io/confluent/ksql/internal/UtilizationMetricsListener.java
rodesai: Ah makes sense - I'd think we should be able to use the same network client metrics from the admin client here.

lct45: To make sure I'm following - gather the same metric names but also from the admin client? So that would be a different metric group, right?

ksqldb-engine/src/main/java/io/confluent/ksql/internal/UtilizationMetricsListener.java
lct45: I added that: https://github.com/apache/kafka/commit/b96f99c595694c954bec59d2ae1fdd9f66d249be but haven't opened the KIP to add to streams

We haven't productionized any of the streams changes but we probably want to soon since the processing ratio is somewhat useless without the thread start time. I think we should change the default time for thread start to be the last poll time to give us a slightly more accurate reading. WDYT @rodesai ?

ksqldb-engine/src/main/java/io/confluent/ksql/internal/UtilizationMetricsListener.java
```diff
@@ -0,0 +1,193 @@
...
+                    .filter(m -> m.metricName().name().equals("thread-start-time") &&
+                            m.metricName().tags().getOrDefault(THREAD_ID, "").equals(threadName))
+                    .collect(Collectors.toList());
+            final Long threadStartTime = startTimeList.size() != 0 ? (Long) startTimeList.get(0).metricValue() : 0L;
+
+            double blockedTime = 0.0;
+            if (threadStartTime > windowStart) {
+                blockedTime += threadStartTime - windowStart;
+            }
+            final double newFlushTime = threadMetrics.getOrDefault("flush-time-total", 0.0);
```
guozhangwang, line 151: I have not seen the change for adding this metric, but just assuming this is measuring the time we call `store.flush` here.

ksqldb-engine/src/main/java/io/confluent/ksql/internal/UtilizationMetricsListener.java
```diff
@@ -0,0 +1,193 @@
...
+    public double processingRatio() {
+        long sampleTime = time.milliseconds();
+
+        final long windowSize = sampleTime - lastSampleTime;
+        final long windowStart = lastSampleTime;
+        double blockedTime;
+        blockedTime = sampleTime - lastSampleTime;
+        for (KafkaStreams stream : kafkaStreams) {
+            for (ThreadMetadata thread : stream.metadataForLocalThreads()) {
+                final double threadTimes = getProcessingRatio(thread.threadName(), stream, windowStart, windowSize);
```
guozhangwang, line 99: nit: `threadTimes` here is a bit confusing. Should it just be `threadBlockedTime`?

Also I think it's cleaner to initialize `blockedTime` as `windowSize`, and then `blockedTime = Math.min(threadBlockedTime, blockedTime);` below.

Furthermore I'm wondering why we want to just use the min among all threads within instances to indicate the node's processing ratio? What about if some instances have very low processing ratio but others on the same pod have reasonably high processing ratio?

ksqldb-engine/src/main/java/io/confluent/ksql/internal/UtilizationMetricsListener.java
```diff
@@ -0,0 +1,193 @@
...
+
+            final List<Metric> startTimeList = allMetrics.stream()
+                    .filter(m -> m.metricName().name().equals("thread-start-time") &&
+                            m.metricName().tags().getOrDefault(THREAD_ID, "").equals(threadName))
+                    .collect(Collectors.toList());
+            final Long threadStartTime = startTimeList.size() != 0 ? (Long) startTimeList.get(0).metricValue() : 0L;
+
+            double blockedTime = 0.0;
+            if (threadStartTime > windowStart) {
+                blockedTime += threadStartTime - windowStart;
```
guozhangwang, line 149: What about when a thread is shutdown (and probably replaced) within a window?

guozhangwang (commented):
This is a meta comment: within a stream thread we have the following blocking procedures:

1. consumer.poll: block on selector IO.
2. consumer.#otherAPIs#: e.g. commit, position; mostly blocking calls on the req/resp with brokers.
3. producer.send: block on buffer full (selector IO are on producer's sender thread).
4. produecer.#otherAPIs#: e.g. txn-related; mostly blocking calls on the req/resp with brokers.
5. admin.#APIs#: mostly are blocking calls similar to 2/4 above.
6. store.#put/get#: these may still block depending on the underlying store, e.g. rocksDB may block put/get if write buffer is already full.
7. store.flush: blocking call.

My question is around how do we define "processing ratio", and do we plan to exclude all above when computing the ratio? Right now we area only counting some of them but not others. 

E.g., at the KS thread-level metric, we already have a processing-ratio which exclude some of 1/2/3/4 but not 5/6/7 above. If we think that is not what we want, could we just update that metric at the streams layer rather than doing fine grained counting at the ksql layer?

PR #7761, opened 2021-07-05
feat: add `ARRAY_CONCAT` UDF

patrickstuedi:
### Description 

Adding `array_concat` UDF function. 

DESCRIBE function array_concat:
```
Name        : ARRAY_CONCAT
Overview    : Concatenates two arrays, creating an array that contains all the elementsin the first array followed by all the elements in the second array.
Type        : SCALAR
Jar         : internal
Variations  : 

        Variation   : ARRAY_CONCAT(left ARRAY<T>, right ARRAY<T>)
        Returns     : ARRAY<T>
        left        : First array of values
        right       : Second array of values
```

### Testing done 

Tested with different array values, including duplicates, null value array members and null value input parameters

### DIFF
--- a/docs/developer-guide/ksqldb-reference/scalar-functions.md
+++ b/docs/developer-guide/ksqldb-reference/scalar-functions.md
@@ -426,6 +426,24 @@ ARRAY_UNION(ARRAY[1, 2, 3, 1, 2], [4, 1])  => [1, 2, 3, 4]
 ARRAY_UNION(ARRAY['apple', 'apple', NULL, 'cherry'], ARRAY['cherry'])  => ['apple', NULL, 'cherry']
 ```
 
+### `ARRAY_CONCAT`
+
+Since: 0.20.0
+
+```sql
+ARRAY_CONCAT(array1, array2)
+```
+
+Returns an array representing the concatenation of both input arrays.
+
+Returns NULL if both input arrays are NULL. If only one argument is NULL, the result is the other argument.
+
+Examples:
+```sql 
+ARRAY_CONCAT(ARRAY[1, 2, 3, 1, 2], [4, 1])  => [1, 2, 3, 1, 2, 4, 1]
+ARRAY_CONCAT(ARRAY['apple', 'apple', NULL, 'cherry'], ARRAY['cherry'])  => ['apple', 'apple', NULL, 'cherry', 'cherry']
+```
+
 ### `AS_MAP`
 
 Since: 0.6.0
--- a/ksqldb-engine/src/main/java/io/confluent/ksql/function/udf/array/ArrayConcat.java
+++ b/ksqldb-engine/src/main/java/io/confluent/ksql/function/udf/array/ArrayConcat.java
@@ -1,0 +1,52 @@
+/*
+ * Copyright 2020 Confluent Inc.
+ *
+ * Licensed under the Confluent Community License; you may not use this file except in compliance
+ * with the License. You may obtain a copy of the License at
+ *
+ * http://www.confluent.io/confluent-community-license
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed under the License
+ * is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and limitations under the
+ * License.
+ */
+
+package io.confluent.ksql.function.udf.array;
+
+import io.confluent.ksql.function.FunctionCategory;
+import io.confluent.ksql.function.udf.Udf;
+import io.confluent.ksql.function.udf.UdfDescription;
+import io.confluent.ksql.function.udf.UdfParameter;
+import java.util.ArrayList;
+import java.util.List;
+
+
+@UdfDescription(
+    name = "array_concat",
+    category = FunctionCategory.ARRAY,
+    description = "Concatenates two arrays, creating an array that contains all the elements"
+        + " in the first array followed by all the elements in the second array."
+        + " Returns NULL if both input arrays are NULL."
+        + " If only one argument is NULL, the result is the other argument."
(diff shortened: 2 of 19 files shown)


JimGalasyn (approved):
LGTM!

ksqldb-engine/src/test/java/io/confluent/ksql/function/udf/array/ArrayConcatTest.java
```diff
@@ -0,0 +1,91 @@
...
+public class ArrayConcatTest {
+
+  private final ArrayConcat udf = new ArrayConcat();
+
+  @Test
+  public void shouldConcatArraysOfLikeType() {
+    final List<String> input1 = Arrays.asList("foo", " ", "bar");
+    final List<String> input2 = Arrays.asList("baz");
+    final List<String> result = udf.concat(input1, input2);
+    assertThat(result.size(), is(input1.size()+input2.size()));
```
vvcephei, line 37:
```suggestion
    assertThat(result, is(Arrays.asList("foo", " ", "bar", "baz"));
```

ksqldb-engine/src/main/java/io/confluent/ksql/function/udf/array/ArrayConcat.java
```diff
@@ -0,0 +1,49 @@
...
+        + "The two arrays must be of the same type.")
+public class ArrayConcat {
+  @Udf
+  public <T> List<T> concat(
+      @UdfParameter(description = "First array of values") final List<T> left,
+      @UdfParameter(description = "Second array of values") final List<T> right) {
+    if (left == null && right == null) {
+      return null;
+    }
+    final List<T> result = Lists.newArrayList();
```
vvcephei, line 40: Let's initialize this arrayList to be size "left.size + right.size": `new ArrayList(left.size() + right.size())`

vvcephei (commented):
Thanks for this! Just a couple of thoughts.

ksqldb-engine/src/main/java/io/confluent/ksql/function/udf/array/ArrayConcat.java
```diff
@@ -0,0 +1,49 @@
...
+        + "The two arrays must be of the same type.")
+public class ArrayConcat {
+  @Udf
+  public <T> List<T> concat(
+      @UdfParameter(description = "First array of values") final List<T> left,
+      @UdfParameter(description = "Second array of values") final List<T> right) {
+    if (left == null && right == null) {
+      return null;
+    }
+    final List<T> result = new ArrayList(left.size() + right.size());
```
agavra, line 40: if `left == null || right == null` wouldn't we get an NPE here?

ksqldb-functional-tests/src/test/resources/query-validation-tests/array-set-functions.json
```diff
@@ -171,6 +171,44 @@
...
+      ]
+    },
+    {
+      "name": "array_concat with primitive type",
+      "statements": [
+        "CREATE STREAM INPUT (id STRING KEY, arr1 ARRAY<INT>, arr2 ARRAY<INT>) WITH (kafka_topic='test_topic', value_format='JSON');",
+        "CREATE STREAM OUTPUT AS SELECT id, array_concat(arr1, arr2) as result FROM INPUT;"
+      ],
+      "inputs": [
+        {"topic": "test_topic", "key": "r1", "value": {"arr1": [0,0,1,0,-1], "arr2": [1,-2,0]}},
```
agavra, line 197: let's also add the null cases here (left is null, right is null, both are null)

agavra (commented):
took a quick scan of the PR - thanks @patrickstuedi!

ksqldb-engine/src/main/java/io/confluent/ksql/function/udf/array/ArrayConcat.java
vvcephei: I thought so too, but there are unit tests for those cases. Not sure what to make of that.

docs/developer-guide/ksqldb-reference/scalar-functions.md
```diff
@@ -426,6 +426,24 @@ ARRAY_UNION(ARRAY[1, 2, 3, 1, 2], [4, 1])  => [1, 2, 3, 4]
...
+
+Since: 0.20.0
+
+```sql
+ARRAY_CONCAT(array1, array2)
+```
+
+Returns an array representing the concatenation of both input arrays.
+
+Returns NULL if both input arrays are NULL.
```
vvcephei, line 439:
```suggestion
Returns NULL if both input arrays are NULL. If only one argument is NULL, the result is the other argument.
```

... Or any other way you prefer to say it, but we should specify the behavior if just one arg is null.

ksqldb-engine/src/main/java/io/confluent/ksql/function/udf/array/ArrayConcat.java
patrickstuedi: thanks for pointing out. missed that when changing the initialization to the exact number of elements. No idea what's going on with the tests, they should fail here. Will double check.

patrickstuedi: Tests did catch this, sorry for missing to run them on that last change. Update fixes the issue in ArrayConcat.

vvcephei:
I just checked on the test failures: https://jenkins.confluent.io/job/Confluentinc%20Contributors/job/ksql/job/PR-7761/7/#showFailuresLink

It looks like the same set of tests started failing on master yesterday: https://jenkins.confluent.io/job/confluentinc/job/ksql/job/master/7992/#showFailuresLink

vvcephei (approved):
Thanks @patrickstuedi !

PR #7765, opened 2021-07-06
chore: remove flaky pull query functional test

lct45:
### Description 
Right now, the port check in `PullQueryRoutingFunctionalTest` errors out inconsistently because in a heavy load system, we'll sometimes use the standby port instead of the active port if the active port isn't in a good state or wasn't able to send health checks due to load. This check is valuable on some systems, but because the CI environment is less reliable, we're not getting consistent results there. Since this is giving us an inconsistent signal, it makes sense to remove this check for now. 

An alternative is to add a retry after the initial check, but that doesn't seem foolproof to me either. We could retry indefinitely but that wouldn't give us any signal for the test.

Recent master failures:
https://jenkins.confluent.io/job/confluentinc/job/ksql/job/master/7968/
https://jenkins.confluent.io/job/confluentinc/job/ksql/job/master/7966/

### Testing done 

### Reviewer checklist

### DIFF
--- a/ksqldb-rest-app/src/test/java/io/confluent/ksql/rest/integration/PullQueryRoutingFunctionalTest.java
+++ b/ksqldb-rest-app/src/test/java/io/confluent/ksql/rest/integration/PullQueryRoutingFunctionalTest.java
@@ -292,34 +292,6 @@ public class PullQueryRoutingFunctionalTest {
   }
 
   @Test
-  public void shouldQueryActiveWhenActiveAliveQueryIssuedToStandby() throws Exception {
-    // Given:
-    ClusterFormation clusterFormation = findClusterFormation(TEST_APP_0, TEST_APP_1, TEST_APP_2);
-    waitForClusterToBeDiscovered(clusterFormation.standBy.getApp(), 3, USER_CREDS);
-    waitForRemoteServerToChangeStatus(clusterFormation.router.getApp(),
-        clusterFormation.router.getHost(), HighAvailabilityTestUtil.lagsReported(3), USER_CREDS);
-
-    waitForRemoteServerToChangeStatus(
-        clusterFormation.standBy.getApp(),
-        clusterFormation.active.getHost(),
-        HighAvailabilityTestUtil::remoteServerIsUp,
-        USER_CREDS);
-
-    // When:
-    List<StreamedRow> rows_0 =
-        makePullQueryRequest(clusterFormation.standBy.getApp(), sql, null, USER_CREDS);
-
-    // Then:
-    assertThat(rows_0, hasSize(HEADER + 1));
-    KsqlHostInfoEntity host = rows_0.get(1).getSourceHost().get();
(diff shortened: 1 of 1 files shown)


ksqldb-rest-app/src/test/java/io/confluent/ksql/rest/integration/PullQueryRoutingFunctionalTest.java
```diff
@@ -313,7 +313,6 @@ public void shouldQueryActiveWhenActiveAliveQueryIssuedToStandby() throws Except
     assertThat(rows_0, hasSize(HEADER + 1));
     KsqlHostInfoEntity host = rows_0.get(1).getSourceHost().get();
     assertThat(host.getHost(), is(clusterFormation.active.getHost().getHost()));
-    assertThat(host.getPort(), is(clusterFormation.active.getHost().getPort()));
```
agavra, line 316: removing this seems to dismiss the point of this test - which is explicitly to check that the active is being queried - right? the fact that the `getHost` lines up is a coincidence because both standby and active are on he same node.

agavra: perhaps the alternative is to see if an attempt to query the active was made at all? either way, I think we should just disable this test and create a ticket, KQE will triage that work.

lct45: You and @vvcephei were on the same page - disabling makes more sense in terms of what the test is trying to check. Do we usually remove the full test or just remove the `@Test` label? Either way I'll make a ticket and link this PR to it.

vvcephei (changes requested):
Hey, thanks for this, @lct45 !

I had one concern.

ksqldb-rest-app/src/test/java/io/confluent/ksql/rest/integration/PullQueryRoutingFunctionalTest.java
vvcephei: Oh, sorry, I sat on my review for too long. I think it'd be better to just delete the test. Disabling it only results in dead code. If the follow-on ticket links to this PR, then the assignee will always be able to resurrect the text of the test if they desire.

lct45:
Deleted and created this: https://confluentinc.atlassian.net/browse/KQE-286
lmk what you think @agavra @vvcephei

ksqldb-rest-app/src/test/java/io/confluent/ksql/rest/integration/PullQueryRoutingFunctionalTest.java
agavra: I'm happy to delete it - but just because you asked usually we add `@Ignore` to disable a test

agavra (approved):

Issue #5868, opened 2020-07-23
Transient queries cannot be terminated by QueryID


wqwangq:
Some temp queries are running in the ksqldb. I cannot use "TERMINATE" to stop them. Please help on this. Thanks!

KSQL cli is 0.9.0 and KSQLDB is 5.5.

```
Query ID                                                            | Query Type | Status    | Sink Name | Sink Kafka Topic | Query String                                                               
----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
 _confluent-ksql-default_transient_4914033483836191816_1595404775294 | PUSH       | RUNNING:1 |           |                  | select * from WORKFLOW_INPUT_DOMAIN_SALESFORCE_ORDERS_STREAM emit changes; 
 _confluent-ksql-default_transient_2968417800293416691_1595412094814 | PUSH       | RUNNING:1 |           |                  | select * from WORKFLOW_INPUT_DOMAIN_SALESFORCE_ORDERS_STREAM emit changes; 
----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
```

In the ksqldb cli window. I input "terminate _confluent-ksql-default_transient_4914033483836191816_1595404775294;", it showed me like below:

```
**### line 1:21: mismatched input '-' expecting ';'
Statement: terminate _confluent-ksql-default_transient_4914033483836191816_1595404775294;
Caused by: line 1:21: mismatched input '-' expecting ';'
Caused by: org.antlr.v4.runtime.InputMismatchException**
```

agavra:
Hello @wqwangq thanks for reporting this issue, the first thing to note is that you need to use backticks (`` ` ``) to escape the `-`s: 
```
terminate `_confluent-ksql-default_transient_4914033483836191816_1595404775294`;
```

But the bigger problem here is that these queries are transient queries, meaning there might be some instance that is depending on them. I don't think we have a way to terminate transient queries as we don't track them. This could be a worthwhile improvement. I'll rename the ticket

GuyIEX:
Ran into the same problem using 0.10.1, but am not able to resolve it using backticks.

```
ksql> show queries;

 Query ID                                                                 | Query Type | Status    | Sink Name                      | Sink Kafka Topic               | Query String                    
-------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
[...]

PR #7766, opened 2021-07-06
fix: Make transient query IDs more user friendly 

nateab:
### Description 
Transient query IDs now use underscores "_" instead of minus signs "-", which no longer require users to use back-ticks in a query.  An example of user confusion is found in #5868. After discussing with Almog, we think this change should be backwards compatible, since for transient queries they get destroyed whenever we stop the ksqldb server, and so we never need to refer to a past transient query (unlike with persistent queries).

### Testing done 
Relevant unit tests updated, manual testing done as well.

When checking queries using `show queries;`, the transient query ID is more appropriately formatted as `transient_SOURCENAME_RANDOMLONG_TIMESTAMP`


agavra:
Just a note that we're getting an expected failure:
```
Expected: is "_confluent_ksql_default_transient_932097300573686369_1606940079718"
     but: was "_confluent_ksql_default_transient_123497300573686369_1606940012345"
	at io.confluent.ksql.engine.OrphanedTransientQueryCleanerTest.shouldCleanup_allApplicationIds(OrphanedTransientQueryCleanerTest.java:86)
```

ksqldb-common/src/main/java/io/confluent/ksql/util/QueryApplicationId.java
```diff
@@ -50,6 +50,6 @@ public static String build(
   }
 
   private static String addTimeSuffix(final String original) {
-    return String.format("%s_%d", original, System.currentTimeMillis());
+    return String.format("%s_%d", original, System.currentTimeMillis()).replace("-","_");
```
agavra, line 53: looking at this code I don't think this is the right way to make this change for a few reasons:

1. we're not respecting configurations that are possibly passed in by the user (e.g. `KSQL_TRANSIENT_QUERY_NAME_PREFIX_CONFIG`) - we shouldn't just be replacing chars in their passed in config
2. it doesn't really make sense for the `addTimeSuffix` method to be doing this

Instead, I think we should look for a way that solves the underlying problem instead of adding shim to fix a bad queryID after it's created. Namely, the underlying problem is that the default generation for transient query ids uses the `_` character. How can we fix that? (hint: maybe the queryID and the applicationID don't necessarily need to be the same - which they are not the same for persistent queries)

agavra (dismissed):

JimGalasyn (approved):

docs/operate-and-deploy/migrations-tool.md
```diff
@@ -49,8 +49,8 @@ types of ksqlDB statements:
 * `DROP TYPE`
 * `SET <property>`
 * `UNSET <property>`
-* `DEFINE <variable>`
-* `UNDEFINE <variable>`
+* `DEFINE <variable>` - available if both `ksql-migrations` and the server are version 0.18 or newer
```
agavra, line 52: it looks like you need to rebase or something - this is from another PR

PR #7770, opened 2021-07-07
test: add tests validating use of ROWOFFSET/ROWPARTITION user columns

Sullivan-Patrick:
### Description
Adds testing to ensure `ROWOFFSET` and `ROWPARTITION` can freely be used as user columns before [KLIP-50](https://github.com/confluentinc/ksql/blob/master/design-proposals/klip-50-partition-and-offset-in-ksqldb.mdl) is integrated.

### Testing done
Individual tests for `ROWOFFSET` and `ROWPARTITION`

### Reviewer checklist

- [ ]  Ensure docs are updated if necessary. (eg. if a user visible feature is being added or changed).
- [ ]  Ensure relevant issues are linked (description should include text like "Fixes #")


vcrfxia (commented):
Thanks @Sullivan-Patrick . The new tests you've added validate that the names `ROWPARTITION` and `ROWOFFSET` may be used as output column aliases, but we also need tests validating that they can be used in source schemas as well. 

Also, I think it's fine to just have a single test covering the use of both aliases (minor optimization in order to reduce the number of new tests being added, since we don't expect the behavior for the two aliases to be different). 

It might also be interesting to add a test where the new columns have the same data types as the pseudocolumns that will be added as part of KLIP-50.

I've also updated the PR title to reflect that this is a test-only change.

ksqldb-functional-tests/src/test/resources/query-validation-tests/row_offset_and_partition.json
```diff
@@ -0,0 +1,43 @@
...
+        {"topic":  "OUTPUT", "key": 1, "value": {"ROWPARTITION": "a", "ROWOFFSET":  "b"}}
+      ],
+      "post": {
+        "sources": [
+          {"name": "OUTPUT", "type": "stream", "schema": "ID INT KEY, ROWPARTITION STRING, ROWOFFSET STRING"}
+        ]
+      }
+    },
+    {
+      "name": "test ROWOFFSET and ROWPARTITION user columns with data types the pseudocolumns will later represent",
```
vcrfxia, line 25: We don't want to have to change the test name once KLIP-50 is implemented (we'd have to remember to rename the historic plan files if so).
```suggestion
      "name": "test ROWOFFSET and ROWPARTITION user columns with pseudocolumn data types",
```

vcrfxia (commented):
Hey @Sullivan-Patrick thanks for the updates. It looks like we're still missing tests for source columns named ROWPARTITION/ROWOFFSET, though. Both tests are currently testing the use of those names as output column aliases only.

ksqldb-functional-tests/src/test/resources/query-validation-tests/row_offset_and_partition.json
Sullivan-Patrick: Woops, I totally missed the source schema thing the first time around. Hopefully the latest commits address everything.

vcrfxia (approved):
Thanks @Sullivan-Patrick -- LGTM!

PR #3323, opened 2019-09-10
feat: Perform Schema Registry permissions checks

spena:
### Description 
This change makes KSQL perform Schema access permissions checks on the statement to execute in order to provide better authorization error messages when the User and/or the KSQL service principal do not have access to the Schema.

First, when SR is configured with Authorization, the following invalid error was displayed on KSQL when it cannot access the schema which was not helping users to recognize it as an access denied aerror:
```
io.confluent.kafka.schemaregistry.client.rest.exceptions.RestClientException: Unrecognized token 'User': was expecting ('true', 'false' or 'null')
 at [Source: (sun.net.www.protocol.http.HttpURLConnection$HttpInputStream); line: 1, column: 6]; error code: 50005
```

This PR changes the message to like:
```
ksql> create stream s10 with (kafka_topic='s10', value_format='avro');
Authorization denied to Read on Schema Registry subject(s): [s10-value]
```

It also perform the check for the KSQL service principal, and the error will look like:
```
ksql> create stream s10 with (kafka_topic='s10', value_format='avro');
The KSQL server is not permitted to execute the command
Caused by: Authorization denied to Read on Schema Registry subject(s):
	[s10-value]
```

SR checks are done in `KsqlAuthorizationValidatorImpl` (where Topic permissions checks are done). This class checks access to the schema of some statements that use AVRO topics only. 

How to review it?
1. Commit #1 added a `KsqlSchemaAuthorizationException` which will be used throw the desired error message.
2. Commit #2 calls the `KsqlSchemaAuthorizationException` on places where SR is denying access + when doing permissions checks on `KsqlAuthorizationValidatorImpl`

### Testing done 
- Unit tests
- Verified manually on CLI

### Reviewer checklist
- [ ] Ensure docs are updated if necessary. (eg. if a user visible feature is being added or changed).
- [ ] Ensure relevant issues are linked (description should include text like "Fixes #<issue number>")

PR #7773, opened 2021-07-08
feat: make SchemaRegistry permission validations on KSQL requests

spena:
### Description 
KSQL has already validations for Kafka permissions when a new query, DDL or DML statement is executed by a user. This PR adds another validation for Schema Registry permissions.

The idea behind this PR is to let users know exactly what's causing a query or DDL operation to fail and also produce a more generic authorization error message similar to the one we have with Kafka. Currently, some statements show the SR authorization error returned by SR, but queries, for instance, never shown anything. Queries run and do not show any error and users always wonder why there's no data streamed (reason was a SR denied operation).

NOTE
This PR is based on an very old PR (https://github.com/confluentinc/ksql/pull/3323). It was easier to push a new PR with the updates. Also, the comments in that PR suggested we change the SR API to produce the authorization error we want, but that requires more work in SR and KSQL. Also, SR may change error messages in the future. The idea of this PR is to have our own authorization message error, for both SR and Kafka, and no depend from other services when we implement better authorization alternatives (i.e. check against RBAC instead of SR and Kafka).

To help the reviewer understand the workflow of validations. 

1. Start on the `KsqlAuthorizationValidatorImpl`. This validator is called on every KSQL request to check a user has permissions to access a topic (and now a SR subject). It is also called by the DistributedExecutor before writing a command to the command topic to check if the KSQL server will have permissions to access the topic (and now a SR subject) later when the command is processed. If the user or the KSQL server do not have permissions, then it throws an authorization exception. No all perms checks are done in this class. Other perms are checked when calling the SR client directly, which should return the right authorization exception.

2. The new authorization exception for SR is called `KsqlSchemaAuthorizationException` that produces an error message with the denied operation. This is the message the users will see in the CLI. The message is similar to the `KsqlTopicAuthorizationException`.

3. The `KsqlAuthorizationValidatorImpl` calls the internal backed validation `KsqlBackendAccessValidator`, which now has a SR validation. The SR validation only supports READ operations. WRITE operations are not possible yet (need to figure out how). We might need to make changes in SR to have a new API to check permissions, but this may require a design doc for SR to add new REST API (will follow-up later). Other option is to check the permission directly in RBAC, but requires changes in the security plugins (will follow-up  in another PR).

4. The `KsqlCacheAccessValidator` is also updated to cache SR permissions checks. If you ask why a cache?, this cache was implemented in the past to make pull queries perform better.

5. The rest of the changes are places where the SR client is called to fetch or register a schema. Those places just make sure to return the `KsqlSchemaAuthorizationException` in case the returned SR error is 403 or 404.

### Testing done 
Added some unit tests and verified in an environment with RBAC + SR.

### Reviewer checklist

### DIFF
--- a/ksqldb-engine/src/main/java/io/confluent/ksql/security/KsqlAuthorizationValidatorFactory.java
+++ b/ksqldb-engine/src/main/java/io/confluent/ksql/security/KsqlAuthorizationValidatorFactory.java
@@ -36,43 +36,72 @@ public final class KsqlAuthorizationValidatorFactory {
 
   public static Optional<KsqlAuthorizationValidator> create(
       final KsqlConfig ksqlConfig,
+      final ServiceContext serviceContext,
+      final Optional<KsqlAuthorizationProvider> externalAuthorizationProvider
+  ) {
+    final Optional<KsqlAccessValidator> accessValidator = getAccessValidator(
+        ksqlConfig,
+        serviceContext,
+        externalAuthorizationProvider
+    );
+
+    return accessValidator.map(v ->
+        new KsqlAuthorizationValidatorImpl(cacheIfEnabled(ksqlConfig, v)));
+  }
+
+  private static Optional<KsqlAccessValidator> getAccessValidator(
+      final KsqlConfig ksqlConfig,
+      final ServiceContext serviceContext,
+      final Optional<KsqlAuthorizationProvider> externalAuthorizationProvider
+  ) {
+    if (externalAuthorizationProvider.isPresent()) {
+      return Optional.of(new KsqlProvidedAccessValidator(externalAuthorizationProvider.get()));
+    } else if (isTopicAccessValidatorEnabled(ksqlConfig, serviceContext)) {
+      return Optional.of(new KsqlBackendAccessValidator());
+    }
+
+    return Optional.empty();
+  }
+
+  private static KsqlAccessValidator cacheIfEnabled(
+      final KsqlConfig ksqlConfig,
+      final KsqlAccessValidator accessValidator
+  ) {
+    return isCacheEnabled(ksqlConfig)
+        ? new KsqlCacheAccessValidator(ksqlConfig, accessValidator)
+        : accessValidator;
+  }
+
+  private static boolean isCacheEnabled(final KsqlConfig ksqlConfig) {
+    // The cache expiry time is used to decided whether to enable the cache or not
+    return ksqlConfig.getLong(KsqlConfig.KSQL_AUTH_CACHE_EXPIRY_TIME_SECS) > 0;
+  }
+
+  private static boolean isTopicAccessValidatorEnabled(
+      final KsqlConfig ksqlConfig,
       final ServiceContext serviceContext
   ) {
     final String enabled = ksqlConfig.getString(KsqlConfig.KSQL_ENABLE_TOPIC_ACCESS_VALIDATOR);
     if (enabled.equals(KsqlConfig.KSQL_ACCESS_VALIDATOR_ON)) {
-      LOG.info("Forcing topic access validator");
-      return Optional.of(createAuthorizationValidator(ksqlConfig));
+      return true;
     } else if (enabled.equals(KsqlConfig.KSQL_ACCESS_VALIDATOR_OFF)) {
-      return Optional.empty();
+      return false;
     }
 
+    // If KSQL_ACCESS_VALIDATOR_AUTO, then check if Kafka has an authorizer enabled
     final Admin adminClient = serviceContext.getAdminClient();
-
     if (isKafkaAuthorizerEnabled(adminClient)) {
       if (KafkaClusterUtil.isAuthorizedOperationsSupported(adminClient)) {
         LOG.info("KSQL topic authorization checks enabled.");
-        return Optional.of(createAuthorizationValidator(ksqlConfig));
+        return true;
       }
 
       LOG.warn("The Kafka broker has an authorization service enabled, but the Kafka "
           + "version does not support authorizedOperations(). "
           + "KSQL topic authorization checks will not be enabled.");
     }
-    return Optional.empty();
-  }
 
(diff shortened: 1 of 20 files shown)


jzaralim (approved):
lgtm, just one question

ksqldb-engine/src/main/java/io/confluent/ksql/schema/ksql/inference/SchemaRegistryTopicSchemaSupplier.java
```diff
@@ -196,8 +197,7 @@ private static SchemaResult notFound(final String topicName, final boolean isKey
             + "\t-> Use the REST API to list available subjects"
             + "\t" + DocumentationLinks.SR_REST_GETSUBJECTS_DOC_URL
             + System.lineSeparator()
-            + "- You do not have permissions to access the Schema Registry.Subject: " + subject
-            + System.lineSeparator()
+            + "- You do not have permissions to access the Schema Registry."
```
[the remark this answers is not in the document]
spena: Not sure why this was added this way. I will look at it in a follow-up PR.

ksqldb-engine/src/main/java/io/confluent/ksql/security/KsqlProvidedAccessValidator.java
```diff
@@ -0,0 +1,61 @@
...
+import static java.util.Objects.requireNonNull;
+
+import com.google.common.collect.ImmutableList;
+import org.apache.kafka.common.acl.AclOperation;
+
+/**
+ * An implementation of {@link KsqlAccessValidator} that provides authorization checks
+ * from a external authorization provider.
+ */
+public class KsqlProvidedAccessValidator implements KsqlAccessValidator {
```
guozhangwang, line 27: nit: why not just name it `KsqlAccessValidatorImpl`? `KsqlProvidedAccessValidator` sounds a bit weird.

ksqldb-engine/src/main/java/io/confluent/ksql/security/KsqlAuthorizationValidatorImpl.java
```diff
@@ -69,10 +75,9 @@ private void validateQuery(
       final MetaStore metaStore,
       final Query query
   ) {
-    final SourceTopicsExtractor extractor = new SourceTopicsExtractor(metaStore);
-    extractor.process(query, null);
-    for (String kafkaTopic : extractor.getSourceTopics()) {
-      accessValidator.checkAccess(securityContext, kafkaTopic, AclOperation.READ);
+    for (KsqlTopic ksqlTopic : extractQueryTopics(query, metaStore)) {
```
guozhangwang, line 78: This is not a comment but a question for my own understanding: it seems `KsqlAuthorizationValidator` is wrapping a `KsqlAccessValidator` here, can we just merge these two interfaces? I guess that's boiled down to:

* What are the conceptual differences of these two?
* Would `KsqlAccessValidator` ever need to be leveraged directly without the `KsqlAuthorizationValidator`?

guozhangwang (commented):
I did not find the class `KsqlSchemaAuthorizationException`, is it missing in this PR?

ksqldb-engine/src/main/java/io/confluent/ksql/security/KsqlProvidedAccessValidator.java
spena: There are 3 classes that extend from this:
```
KsqlProvidedAccessValidator
KsqlCacheAccessValidator
KsqlBackendAccessValidator
```

I used `Provided...` because it is a validator  provided by the user on the `ksql-server.properties`. I wasn't sure if use this os `KsqlExternalAccessValidator`. Any ideas?

ksqldb-engine/src/main/java/io/confluent/ksql/security/KsqlAuthorizationValidatorImpl.java
spena: `KsqlAuthorizationValidator` validates the user can execute a specified statement. This class gets information about that, such as topics and schemas; and then validates each resource based on the type of statement (query, create, insert, ...).

Initially, the validation was part of this `KsqlAuthorizationValidator`, but I split it when I introduced an authorization cache, which lives in `KsqlCacheAccessValidator`. The cache wraps another validation if the cache is enabled. The current validation, besides caching, is the `KsqlBackendAccessValidator`, which just checks with the backend service (i.e. Kafka) to check for ACLs. The new one is `KsqlProvidedAccessValidator`, which checks with an external service, like RBAC. SR permissions will be checked by the KSQL confluent security plugins which can either use RBAC or call the SR `/permissions` endpoint to check for ACLs (both part of the confluent security plugins).

spena:
@guozhangwang The `KsqlSchemaAuthorizationException` is part of the KSQL code already. I split this PR into another that added that. Sorry I didn't update the PR description.

ksqldb-engine/src/main/java/io/confluent/ksql/security/KsqlAuthorizationValidatorFactory.java
```diff
@@ -35,44 +35,73 @@ private KsqlAuthorizationValidatorFactory() {
...
+  }
+
+  private static Optional<KsqlAccessValidator> getAccessValidator(
+      final KsqlConfig ksqlConfig,
+      final ServiceContext serviceContext,
+      final Optional<KsqlAuthorizationProvider> externalAuthorizationProvider
+  ) {
+    if (externalAuthorizationProvider.isPresent()) {
+      return Optional.of(new KsqlProvidedAccessValidator(externalAuthorizationProvider.get()));
+    } else if (isTopicAccessValidatorEnabled(ksqlConfig, serviceContext)) {
```
jzaralim, line 59: Should we also have a config for enabling SR auth?

jzaralim (approved):

PR #7778, opened 2021-07-09
feat: add BYTES type to ksqlDB

jzaralim:
### Description 
Adds SqlTypes.BYTES and BytesLiteral to ksqlDB. The main KLIP, #7764, is not merged yet but this PR is almost 100% wiring with no logic, so I'm filing this now to prep for future BYTES work.

### Testing done 
Updated unit tests

### Reviewer checklist

### DIFF
--- a/ksqldb-execution/src/main/java/io/confluent/ksql/execution/expression/tree/BytesLiteral.java
+++ b/ksqldb-execution/src/main/java/io/confluent/ksql/execution/expression/tree/BytesLiteral.java
@@ -1,0 +1,75 @@
+/*
+ * Copyright 2021 Confluent Inc.
+ *
+ * Licensed under the Confluent Community License (the "License"); you may not use
+ * this file except in compliance with the License.  You may obtain a copy of the
+ * License at
+ *
+ * http://www.confluent.io/confluent-community-license
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OF ANY KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations under the License.
+ */
+
+package io.confluent.ksql.execution.expression.tree;
+
+import static java.util.Objects.requireNonNull;
+
+import com.google.common.collect.ImmutableList;
+import com.google.errorprone.annotations.Immutable;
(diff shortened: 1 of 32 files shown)


ksqldb-execution/src/main/java/io/confluent/ksql/execution/expression/tree/BytesLiteral.java
```diff
@@ -0,0 +1,75 @@
...
+import io.confluent.ksql.parser.NodeLocation;
+import java.nio.ByteBuffer;
+import java.util.Objects;
+import java.util.Optional;
+import org.apache.commons.lang3.ArrayUtils;
+
+@Immutable
+public class BytesLiteral extends Literal {
+
+  private final ImmutableList<Byte> value;
```
jzaralim, line 31: Doing this because `ByteBuffer` and `byte[]` are both mutable.

ksqldb-execution/src/main/java/io/confluent/ksql/execution/expression/tree/BytesLiteral.java
```diff
@@ -0,0 +1,75 @@
...
+
+  public BytesLiteral(final ByteBuffer value) {
+    this(Optional.empty(), value);
+  }
+
+  public BytesLiteral(final Optional<NodeLocation> location, final ByteBuffer value) {
+    super(location);
+    final byte[] bytes = new byte[requireNonNull(value, "value").capacity()];
+    value.get(bytes);
+    this.value = ImmutableList.copyOf(ArrayUtils.toObject(bytes));
```
spena, line 41: What if you keep the `final byte[] bytes` at the class variable instead of the `ImmutableList<Byte>`? The `value.get(bytes)` will copy everything from ByteBuffer -> bytes, so you have an exact immutable copy.

Later, when you call `getValue()`, you can return `ByteBuffer.wrap(bytes).asReadOnlyBuffer()` to avoid the returned byte buffer can be modified. 

And, when you call `getByteArray()`, then clone the array 
```
return bytes == null ? null : bytes.clone();
```

jzaralim: Yeah that works!

spena (approved):
LGTM

PR #7782, opened 2021-07-12
test: fix flaky test where FileWatcher might be called multiple times for update

agavra:
### Description 

This is the counterpart to #6696 except for also file updated, not just created.

### Testing done 

Ran the test

### Reviewer checklist


vvcephei (approved):
Thanks, @agavra !

PR #7784, opened 2021-07-12
test: increase time out for migrations test

agavra:
### Description 

There were some times when the `shouldApplyMigrationsAndDisplayInfo` was failing due to a materialized view not being available within 30s. I've seen this happen before on really slow Jenkins nodes, so we're just increasing the timeout and hoping it doesn't happen again. The integration test is worth keeping.

### Testing done 

Ran the test

### Reviewer checklist


vvcephei (approved):
Thanks, @agavra !

It's worth trying this to see if it helps. We have also recently changed some default Client/Broker settings that might be making this stuff slower. Do you think we should try to introduce a blanket set of defaults to all integration tests to make them run more snappily?

(cc @guozhangwang )

agavra:
> Do you think we should try to introduce a blanket set of defaults to all integration tests to make them run more snappily?

@vvcephei I'm supportive of it - these are things like smaller buffering, faster commit intervals? Is there anything else we should consider?

PR #7785, opened 2021-07-12
fix: create more user friendly transient query ids

nateab:
### Description 
Transient query IDs now use underscores "_" instead of minus signs "-", which no longer require users to use back-ticks in a query.  An example of user confusion is found in #5868. After discussing with Almog, we think this change should be backwards compatible, since for transient queries they get destroyed whenever we stop the ksqldb server, and so we never need to refer to a past transient query (unlike with persistent queries).

Cleaner version of pr #7766 

### Testing done 
Relevant unit tests updated, manual testing done as well.

When checking queries using `show queries;`, the transient query ID is more appropriately formatted as `transient_SOURCENAME_RANDOMLONG_TIMESTAMP`


ksqldb-engine/src/main/java/io/confluent/ksql/engine/QueryIdUtil.java
```diff
@@ -100,7 +100,9 @@ static QueryId buildId(
     }
 
     if (!outputNode.getSinkName().isPresent()) {
-      return new QueryId(String.valueOf(Math.abs(ThreadLocalRandom.current().nextLong())));
+      final String prefix =
+          "transient_" + outputNode.getSource().getLeftmostSourceNode().getAlias().text() + "_";
```
agavra, line 104: there doesn't seem to be any tests that run through this part of the code - can you add tests for that (at least unit tests for `QueryIdUtil` that mock out `outputNode`, but ideally also some end-to-end tests)

nateab: I added a unit test in `QueryIdUtil ` that checks we've added the source name as desired. However, I have been struggling to find some end-to-end tests that I can follow to help me create one here, are there any you would recommend?

agavra: You can use `KsqlEngineTest` if you want - otherwise I'm actually OK waiting for when we do #5868 because we'll have to have more testing that covers this for that anyway.

agavra (approved):
LGTM! Thanks @nateab - let's wait for one more review to sanity check this is backwards compatible. @vcrfxia mind taking a look?

ksqldb-engine/src/main/java/io/confluent/ksql/engine/QueryIdUtil.java
AlanConfluent: is `getLeftmostSourceNode` guaranteed to always be present?  I assume it is, being that it seems it's either filled out in the case of a non-join or the left-most in the case of a join.  Do we do a similar naming convention with persistent queries?

AlanConfluent (approved):

rodesai (approved):
LGTM

spena (approved):

PR #7795, opened 2021-07-15
chore: improve error message for subscript misuse

agavra:
### Description 

I accidentally used `a_struct['foo']` instead of `a_struct->foo` and was entirely confused by the error message (I thought `TIMESTAMP` as not supported in structs). This gives a better error message (see unit tests)

### Testing done 

Unit testing

### Reviewer checklist


jzaralim (approved):
looks great, thanks for fixing!

PR #7801, opened 2021-07-16
chore: upgrade to VertX 3.9.8 and Netty 4.1.65.Final to avoid CVE

agavra:
### Description 

We temporarily used Netty 4.1.60 despite known CVEs because VertX did not support later versions of netty. This PR addresses that issue.

### Testing done 

Tested end-to-end with basic auth (which was the symptom of the incompatibility the first time around)

### Reviewer checklist


spena (approved):
LGTM

PR #7804, opened 2021-07-16
feat: implement bytes comparisons

jzaralim:
### Description 
Comparisons for BYTES - they are compared lexicographically using the `ByteByffer.compareTo` function.

### Testing done 
QTT and unit tests.

### Reviewer checklist

### DIFF
--- a/ksqldb-execution/src/main/java/io/confluent/ksql/execution/interpreter/ComparisonInterpreter.java
+++ b/ksqldb-execution/src/main/java/io/confluent/ksql/execution/interpreter/ComparisonInterpreter.java
@@ -97,7 +97,9 @@ public final class ComparisonInterpreter {
     };
   }
 
+  // CHECKSTYLE_RULES.OFF: CyclomaticComplexity
   private static Optional<ComparisonFunction> doCompareTo(final Term left, final Term right) {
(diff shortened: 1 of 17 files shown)


ksqldb-execution/src/main/java/io/confluent/ksql/execution/interpreter/ComparisonInterpreter.java
```diff
@@ -118,6 +120,12 @@ private static ComparisonNullCheckFunction getNullCheckFunction(
       return Optional.of((o1, o2) -> castLeft.cast(o1).compareTo(castRight.cast(o2)));
     } else if (leftType == SqlBaseType.STRING) {
       return Optional.of((o1, o2) -> o1.toString().compareTo(o2.toString()));
+    } else if (either(leftType, rightType, SqlBaseType.BYTES)) {
```
spena, line 123: Does this mean you can compare BYTES with other types?

jzaralim: It will eventually throw an error if either of them aren't BYTES. Updated to check that condition earlier.

spena (approved):
LGTM

PR #7806, opened 2021-07-16
chore: add nonfunctioning pseudocolumn versioning to source steps

Sullivan-Patrick:
### Description 
Adding psuedocolumn field to `SourceStep`s as part of [KLIP-50](https://github.com/confluentinc/ksql/blob/master/design-proposals/klip-50-partition-and-offset-in-ksqldb.md). With this PR, the pseudocolumns don't get written into the command topic, therefore are "nonfunctioning".  In future PRs, these will map to pseudocolumns corresponding to the pseudocolumn version number of a specific query.

### Testing done 
No testing on account of difficulties surrounding writing unit tests for this. Currently it seems like our current tests should handle this- a written versionNumber makes the entire server degrade and is fairly obvious- but regardless, I'm open to ideas if anyone thinks of a way to test this granularly.

### Reviewer checklist

### DIFF
--- a/ksqldb-common/src/main/java/io/confluent/ksql/schema/ksql/SystemColumns.java
+++ b/ksqldb-common/src/main/java/io/confluent/ksql/schema/ksql/SystemColumns.java
@@ -32,6 +32,8 @@ public final class SystemColumns {
   public static final ColumnName WINDOWEND_NAME = ColumnName.of("WINDOWEND");
   public static final SqlType WINDOWBOUND_TYPE = SqlTypes.BIGINT;
 
+  public static final int CURRENT_PSEUDOCOLUMN_VERSION_NUMBER = -2;
+
   private static final Set<ColumnName> PSEUDO_COLUMN_NAMES = ImmutableSet.of(
       ROWTIME_NAME
   );
--- a/ksqldb-common/src/main/java/io/confluent/ksql/schema/ksql/SystemColumns.java
+++ b/ksqldb-common/src/main/java/io/confluent/ksql/schema/ksql/SystemColumns.java
@@ -32,7 +32,7 @@ public final class SystemColumns {
   public static final ColumnName WINDOWEND_NAME = ColumnName.of("WINDOWEND");
   public static final SqlType WINDOWBOUND_TYPE = SqlTypes.BIGINT;
 
-  public static final int CURRENT_PSEUDOCOLUMN_VERSION_NUMBER = -2;
+  public static final int CURRENT_PSEUDOCOLUMN_VERSION_NUMBER = -1;
 
   private static final Set<ColumnName> PSEUDO_COLUMN_NAMES = ImmutableSet.of(
       ROWTIME_NAME
--- a/ksqldb-common/src/main/java/io/confluent/ksql/schema/ksql/SystemColumns.java
+++ b/ksqldb-common/src/main/java/io/confluent/ksql/schema/ksql/SystemColumns.java
@@ -32,7 +32,7 @@ public final class SystemColumns {
   public static final ColumnName WINDOWEND_NAME = ColumnName.of("WINDOWEND");
   public static final SqlType WINDOWBOUND_TYPE = SqlTypes.BIGINT;
 
-  public static final int CURRENT_PSEUDOCOLUMN_VERSION_NUMBER = -1;
+  public static final int CURRENT_PSEUDOCOLUMN_VERSION_NUMBER = 5;
 
   private static final Set<ColumnName> PSEUDO_COLUMN_NAMES = ImmutableSet.of(
       ROWTIME_NAME
--- a/ksqldb-common/src/main/java/io/confluent/ksql/schema/ksql/SystemColumns.java
+++ b/ksqldb-common/src/main/java/io/confluent/ksql/schema/ksql/SystemColumns.java
@@ -32,6 +32,7 @@ public final class SystemColumns {
   public static final ColumnName WINDOWEND_NAME = ColumnName.of("WINDOWEND");
   public static final SqlType WINDOWBOUND_TYPE = SqlTypes.BIGINT;
 
+  public static final int LEGACY_PSEUDOCOLUMN_VERSION_NUMBER = 0;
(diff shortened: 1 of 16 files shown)


ksqldb-engine/src/main/java/io/confluent/ksql/execution/json/PlanJsonMapper.java
```diff
@@ -41,7 +41,7 @@
           new KsqlTypesDeserializationModule()
       )
       .enable(DeserializationFeature.FAIL_ON_UNKNOWN_PROPERTIES)
-      .enable(DeserializationFeature.FAIL_ON_NULL_FOR_PRIMITIVES)
+      .disable(DeserializationFeature.FAIL_ON_NULL_FOR_PRIMITIVES)
```
Sullivan-Patrick, line 44: Anyone who has input on if this is a reasonable thing to do, please chime in. It's necessary to make this work with an int as versionNumbers, but if there is a good reason to, we can use Optionals instead.

vcrfxia (commented):
Thanks @Sullivan-Patrick -- comments inline. I've also updated the PR title from `feat` to `chore` since this PR is a non-functional change. (Another good litmus test when deciding which commit type to use is to ask whether it's useful for ksqlDB users for this PR to show up in the changelog or not. If the answer is no, then `feat` is not the right commit type for the PR.)

ksqldb-execution/src/main/java/io/confluent/ksql/execution/plan/SourceStep.java
```diff
@@ -33,21 +34,26 @@
   final Optional<TimestampColumn> timestampColumn;
   final LogicalSchema sourceSchema;
 
+  int versionNumber;
```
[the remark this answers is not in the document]
Sullivan-Patrick: My excuse is that both of these issues are remnants of playing around with Jackson! Originally this had an `JsonIgnore` above it and was given a default value of 0. Thanks for the catch.

ksqldb-execution/src/main/java/io/confluent/ksql/execution/plan/SourceStep.java
```diff
@@ -32,20 +33,23 @@
   final Formats formats;
   final Optional<TimestampColumn> timestampColumn;
   final LogicalSchema sourceSchema;
+  final int versionNumber;
```
vcrfxia, line 36: Can we rename this to `pseudocolumnVersion`? The current name is misleading since it makes it sound like we're versioning the step itself, but really we're just versioning the pseudocolumns.

ksqldb-engine/src/main/java/io/confluent/ksql/execution/json/PlanJsonMapper.java
vcrfxia: If we go the route of disabling this mapper property, let's just remove the line (in order to have the property disabled by default) rather than explicitly disabling, in order to maintain consistency with the other mappers that also disable the property.

vcrfxia (commented):
Thanks @Sullivan-Patrick -- almost there! Would be good to get @rodesai 's thoughts on removing `FAIL_ON_NULL_FOR_PRIMITIVES` from PlanJsonMapper versus using Optional for the new pseudocolumn version field since I'm not clear on the original motivation for enabling `FAIL_ON_NULL_FOR_PRIMITIVES`.

ksqldb-execution/src/main/java/io/confluent/ksql/execution/plan/SourceStep.java
Sullivan-Patrick: Ok, I'll wait on his feedback before merging. I've made the other changes locally.

ksqldb-common/src/main/java/io/confluent/ksql/schema/ksql/SystemColumns.java
```diff
@@ -32,6 +32,8 @@
   public static final ColumnName WINDOWEND_NAME = ColumnName.of("WINDOWEND");
   public static final SqlType WINDOWBOUND_TYPE = SqlTypes.BIGINT;
 
+  public static final int CURRENT_PSEUDOCOLUMN_VERSION_NUMBER = 5;
```
Sullivan-Patrick, line 35: arbitrary number since nothing is written

ksqldb-execution/src/main/java/io/confluent/ksql/execution/plan/StreamSource.java
```diff
@@ -41,14 +42,16 @@ public StreamSource(
...
+      @JsonProperty("pseudoColumnVersion") final OptionalInt pseudoColumnVersion
   ) {
     super(
         props,
         topicName,
         formats,
         timestampColumn,
-        sourceSchema
+        sourceSchema,
+        pseudoColumnVersion.orElse(0)
```
vcrfxia, line 54: Let's pull the value 0 out into a constant, e.g., `LEGACY_PSEUDOCOLUMN_VERSION_NUMBER`.

ksqldb-execution/src/test/java/io/confluent/ksql/execution/plan/StreamSourceTest.java
```diff
@@ -46,26 +47,29 @@
   @Test
   public void shouldImplementEquals() {
     new EqualsTester()
-        .addEqualityGroup(
+        .addEqualityGroup (
+            new StreamSource(
+                properties1, "topic1", formats1, Optional.of(timestamp1), schema1, OptionalInt.of(0)),
             new StreamSource(
-                properties1, "topic1", formats1, Optional.of(timestamp1), schema1),
+                properties1, "topic1", formats1, Optional.of(timestamp1), schema1, OptionalInt.of(0)))
```
vcrfxia, line 54: `OptionalInt.of(0)` and `OptionalInt.empty()` are expected to compare the same, right? If so, it'd be nice to add a case for this into this unit test (and similarly for the other classes too).

vcrfxia (commented):
LGTM besides the two minor comments inline!

vcrfxia (approved):
LGTM! Will merge once the build passes.

PR #7808, opened 2021-07-18
build(hotfix): temporarily disable spotbug for certain modules

mjsax:
Spotbug was bumped upstream from 4.2.0 to 4.3.0 recently, breaking ksqlDB build.
This PR disabled spotbug temporarily to let the build pass again.
We will re-enable spotbug on a per-module basis with follow up PRs.

PR #7809, opened 2021-07-18
build: update `api-client` module for spotbug 4.3.0 and re-enable plugin

mjsax:
Follow up to #7808

### DIFF
--- a/ksqldb-api-client/pom.xml
+++ b/ksqldb-api-client/pom.xml
@@ -95,6 +95,16 @@
             <version>${reactive-streams.version}</version>
         </dependency>
 
+        <dependency>
+            <groupId>com.google.guava</groupId>
+            <artifactId>guava</artifactId>
+        </dependency>
+
+        <dependency>
+            <groupId>com.github.spotbugs</groupId>
+            <artifactId>spotbugs-annotations</artifactId>
+        </dependency>
+
         <!-- Required for running tests -->
 
         <dependency>
@@ -192,13 +202,6 @@
                     <nodeprecated>true</nodeprecated>
                     <nohelp>true</nohelp>
                     <notree>true</notree>
-                </configuration>
-            </plugin>
-            <plugin>
-                <groupId>com.github.spotbugs</groupId>
-                <artifactId>spotbugs-maven-plugin</artifactId>
-                <configuration>
-                    <skip>true</skip>
                 </configuration>
             </plugin>
         </plugins>
--- a/ksqldb-api-client/src/main/java/io/confluent/ksql/api/client/BatchedQueryResult.java
+++ b/ksqldb-api-client/src/main/java/io/confluent/ksql/api/client/BatchedQueryResult.java
@@ -15,6 +15,7 @@
 
 package io.confluent.ksql.api.client;
 
+import edu.umd.cs.findbugs.annotations.SuppressFBWarnings;
 import java.util.List;
 import java.util.concurrent.CompletableFuture;
 
@@ -39,6 +40,7 @@ public abstract class BatchedQueryResult extends CompletableFuture<List<Row>> {
    *
    * @return a future containing the query ID (or null in the case of pull queries)
    */
+  @SuppressFBWarnings(value = "EI_EXPOSE_REP")
   public abstract CompletableFuture<String> queryID();
 
 }
--- a/ksqldb-api-client/src/main/java/io/confluent/ksql/api/client/impl/BatchedQueryResultImpl.java
+++ b/ksqldb-api-client/src/main/java/io/confluent/ksql/api/client/impl/BatchedQueryResultImpl.java
@@ -15,6 +15,7 @@
 
 package io.confluent.ksql.api.client.impl;
 
+import edu.umd.cs.findbugs.annotations.SuppressFBWarnings;
 import io.confluent.ksql.api.client.BatchedQueryResult;
 import java.util.concurrent.CompletableFuture;
 
@@ -32,6 +33,7 @@ public class BatchedQueryResultImpl extends BatchedQueryResult {
   }
 
   @Override
+  @SuppressFBWarnings(value = "EI_EXPOSE_REP")
   public CompletableFuture<String> queryID() {
     return queryId;
   }
--- a/ksqldb-api-client/src/main/java/io/confluent/ksql/api/client/impl/ConnectorDescriptionImpl.java
+++ b/ksqldb-api-client/src/main/java/io/confluent/ksql/api/client/impl/ConnectorDescriptionImpl.java
@@ -15,6 +15,8 @@
 
 package io.confluent.ksql.api.client.impl;
 
+import com.google.common.collect.ImmutableList;
+import edu.umd.cs.findbugs.annotations.SuppressFBWarnings;
 import io.confluent.ksql.api.client.ConnectorDescription;
 import io.confluent.ksql.api.client.ConnectorType;
 import java.util.List;
@@ -24,8 +26,8 @@ public class ConnectorDescriptionImpl implements ConnectorDescription {
 
   private final String name;
   private final String className;
-  private final List<String> sources;
-  private final List<String> topics;
+  private final ImmutableList<String> sources;
+  private final ImmutableList<String> topics;
   private final ConnectorType type;
   private final String state;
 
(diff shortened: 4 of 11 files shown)


ksqldb-api-client/pom.xml
```diff
@@ -95,6 +95,16 @@
             <version>${reactive-streams.version}</version>
         </dependency>
 
+        <dependency>
+            <groupId>com.google.guava</groupId>
+            <artifactId>guava</artifactId>
```
mjsax, line 100: Need to include `guava` as we are using `ImmutableList` etc now

ksqldb-api-client/pom.xml
```diff
@@ -95,6 +95,16 @@
...
+        </dependency>
+
+        <dependency>
+            <groupId>com.github.spotbugs</groupId>
+            <artifactId>spotbugs-annotations</artifactId>
```
mjsax, line 105: Need to include `spotbugs-annotations` for `@SuppressFBWarnings`

ksqldb-api-client/src/main/java/io/confluent/ksql/api/client/BatchedQueryResult.java
```diff
@@ -39,6 +40,7 @@
    *
    * @return a future containing the query ID (or null in the case of pull queries)
    */
+  @SuppressFBWarnings(value = "EI_EXPOSE_REP")
   public abstract CompletableFuture<String> queryID();
```
mjsax, line 44: Seems `CompletableFuture` is modifiable so we need to suppress I guess.

ksqldb-api-client/src/main/java/io/confluent/ksql/api/client/impl/BatchedQueryResultImpl.java
```diff
@@ -32,6 +33,7 @@
   }
 
   @Override
+  @SuppressFBWarnings(value = "EI_EXPOSE_REP")
   public CompletableFuture<String> queryID() {
```
mjsax, line 37: As above

ksqldb-api-client/src/main/java/io/confluent/ksql/api/client/impl/RowImpl.java
```diff
@@ -27,35 +30,37 @@
...
   @Override
+  @SuppressFBWarnings(value = "EI_EXPOSE_REP", justification = "columnTypes is ImmutableList")
   public List<ColumnType> columnTypes() {
     return columnTypes;
   }
 
   @Override
   public KsqlArray values() {
-    return values;
+    return values.copy();
```
mjsax, line 63: Make deep-copy instead of suppress spotbug warning.

ksqldb-api-client/src/main/java/io/confluent/ksql/api/client/impl/StreamInsertsSubscriber.java
```diff
@@ -35,6 +36,7 @@
   private int outstandingTokens;
   private boolean drainHandlerSet;
 
+  @SuppressFBWarnings(value = "EI_EXPOSE_REP2")
   public StreamInsertsSubscriber(final Context context, final HttpClientRequest httpRequest) {
```
mjsax, line 40: Getting mutable fields as input... Need to suppress.

ksqldb-cli/pom.xml
```diff
@@ -154,6 +154,13 @@
                     </execution>
                 </executions>
             </plugin>
+            <plugin>
```
mjsax, line 157: Those are the original changes from #7808 (included for not to make the build pass)

ksqldb-api-client/pom.xml
```diff
@@ -95,6 +95,16 @@
...
+        </dependency>
+
```
mjsax, line 107: The diff does not show that we "remove" the "skip config" for spotbugs, because it chancels out with the change from #7808 that are included in this PR. After #7808 is merged and this PR is rebased, the change should show up on Github diff.

ksqldb-api-client/src/main/java/io/confluent/ksql/api/client/impl/ConnectorDescriptionImpl.java
```diff
@@ -40,18 +42,20 @@ public ConnectorDescriptionImpl(
...
-    this.topics = Objects.requireNonNull(topics, "topics");
+    this.sources = ImmutableList.copyOf(Objects.requireNonNull(sources, "sources"));
+    this.topics = ImmutableList.copyOf(Objects.requireNonNull(topics, "topics"));
     this.type = Objects.requireNonNull(type, "type");
     this.state = Objects.requireNonNull(state, "state");
   }
 
   @Override
+  @SuppressFBWarnings(value = "EI_EXPOSE_REP", justification = "sources is ImmutableList")
   public List<String> sources() {
```
mjsax, line 53: Wondering if we should change the return type to `ImmutableList<String>` ?

Some classes specify the interfaces (`List`, `Map`, `Set`) while others use `ImmutableXxx`. Seems better to use `ImmutableXxx` IMHO but wanted to clarify first. If we don't use `ImmutableXxx` callers might need to make more (unnecessary) deep-copies?

If we want to stick with the interface, we should follow this pattern consistently though and should update the classes that specify `ImmutableXxx` as return type.

jzaralim: Not sure if I'm understanding this correctly, but right now none of the classes in the java-api-client module are using ImmutableXXX outside of tests. The interface is pretty consistent now, so if we just change all the constructors to make copies of `List/Set/Map` and getters to return copies, would that resolve the warning?

ksqldb-api-client/src/main/java/io/confluent/ksql/api/client/impl/ConnectorDescriptionImpl.java
```diff
@@ -40,18 +42,20 @@ public ConnectorDescriptionImpl(
...
     return sources;
   }
 
   @Override
+  @SuppressFBWarnings(value = "EI_EXPOSE_REP", justification = "topics is ImmutableList")
```
jzaralim, line 58: Can we return a deep copy instead of suppressing? 

`List.copyOf(topics)`

ksqldb-api-client/src/main/java/io/confluent/ksql/api/client/impl/ConnectorDescriptionImpl.java
mjsax: It a general question across the whole code base. I have seen both cases. It's nota about this module in particular.

Most of the time, we use the native Java interfaces `List` etc, but sometime we use `ImmutableList`.

ksqldb-api-client/src/main/java/io/confluent/ksql/api/client/impl/ConnectorDescriptionImpl.java
mjsax: We do return an `ImmutableList` do what do we gain by making a deep-copy? The issue with `ImmutableList` is, that spotbug does not detect it as immutable, and thus we need to add the suppress.

I agree that it's annoying that we need to annotation, and we could also return a deep-copy, but it seems unnecessary?

ksqldb-api-client/src/main/java/io/confluent/ksql/api/client/impl/ConnectorDescriptionImpl.java
mjsax: Cf https://github.com/confluentinc/ksql/pull/7817/files#r671942538

ksqldb-api-client/src/main/java/io/confluent/ksql/api/client/impl/ConnectorDescriptionImpl.java
jzaralim: Sounds good. If https://github.com/spotbugs/spotbugs/issues/1601 ever gets resolved, we should remove all the annotations so that we can differentiate the actual mutable methods from ones the ones that use Guava classes.

ksqldb-api-client/src/main/java/io/confluent/ksql/api/client/impl/ConnectorDescriptionImpl.java
jzaralim: I would vote on using ImmutableList across the rest of the code base, but this specific module is user facing, so I think we should keep this API as is.

jzaralim (approved):

PR #7810, opened 2021-07-18
build: update `common` module for spotbug 4.3.0 and re-enable plugin

mjsax:
Follow up to #7808

### DIFF
--- a/ksqldb-common/src/main/java/io/confluent/ksql/GenericKey.java
+++ b/ksqldb-common/src/main/java/io/confluent/ksql/GenericKey.java
@@ -16,6 +16,7 @@
 package io.confluent.ksql;
 
 import com.google.errorprone.annotations.Immutable;
+import edu.umd.cs.findbugs.annotations.SuppressFBWarnings;
 import io.confluent.ksql.schema.ksql.LogicalSchema;
 import io.confluent.ksql.testing.EffectivelyImmutable;
 import java.util.ArrayList;
@@ -59,6 +60,7 @@ public final class GenericKey {
     return values.get(index);
   }
 
+  @SuppressFBWarnings(value = "EI_EXPOSE_REP", justification = "values is unmodifiableList()")
   public List<?> values() {
     return values;
   }
--- a/ksqldb-common/src/main/java/io/confluent/ksql/function/AggregateFunctionInitArguments.java
+++ b/ksqldb-common/src/main/java/io/confluent/ksql/function/AggregateFunctionInitArguments.java
@@ -15,7 +15,10 @@
 
 package io.confluent.ksql.function;
 
+import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableMap;
+import edu.umd.cs.findbugs.annotations.SuppressFBWarnings;
+import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collections;
 import java.util.List;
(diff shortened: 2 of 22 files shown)


ksqldb-api-client/pom.xml
```diff
@@ -194,6 +194,13 @@
                     <notree>true</notree>
                 </configuration>
             </plugin>
+            <plugin>
```
mjsax, line 197: Those are the original changes from #7808 (included for not to make the build pass).

ksqldb-common/src/main/java/io/confluent/ksql/GenericKey.java
```diff
@@ -59,6 +60,7 @@ public Object get(final int index) {
     return values.get(index);
   }
 
+  @SuppressFBWarnings(value = "EI_EXPOSE_REP", justification = "values is unmodifiableList()")
   public List<?> values() {
```
mjsax, line 64: Same question as https://github.com/confluentinc/ksql/pull/7809/files#r671807853

Should we update the return type to `ImmutableList` ?

ksqldb-common/src/main/java/io/confluent/ksql/logging/processing/ProcessingLoggingJsonMapper.java
```diff
@@ -41,6 +41,6 @@
       .setNodeFactory(JsonNodeFactory.withExactBigDecimals(true));
 
   public ObjectMapper get() {
-    return mapper;
+    return mapper.copy();
```
mjsax, line 44: Make deep-copy instead of suppress spotbug warning.

mjsax:
The diff does not show `ksqldb-common/pom.xml` even if it "removes" the "skip config" for spotbugs, because it chancels out with the change from #7808 that are included in this PR. After #7808 is merged and this PR is rebased, the change should show up on Github diff.

ksqldb-common/src/main/java/io/confluent/ksql/function/AggregateFunctionInitArguments.java
```diff
@@ -34,8 +37,8 @@
       new AggregateFunctionInitArguments();
 
   private final int udafIndex;
-  private final List<Object> initArgs;
-  private final Map<String, ?> config;
+  private final List<Object> initArgs; // cannot use ImmutableList as we need to handle `null`
```
mjsax, line 40: Cf `UdafAggregateFunctionFactoryTest#shouldHandleNullLiteralParams()`.

ksqldb-common/src/main/java/io/confluent/ksql/function/KsqlScalarFunction.java
```diff
@@ -93,6 +94,7 @@ public static KsqlScalarFunction createLegacyBuiltIn(
     );
   }
 
+  @SuppressFBWarnings(value = "EI_EXPOSE_REP")
   public Class<? extends Kudf> getKudfClass() {
     return kudfClass;
```
jzaralim, line 99: `Class.forName(kudfClass.getName(), true, kudfClass.getClassLoader());` might work

jzaralim (approved):
I think we should update return types to `ImmutableXXX`, but this looks good if we decide not to.

ksqldb-common/src/main/java/io/confluent/ksql/function/KsqlScalarFunction.java
mjsax: Maybe, but for this case we need to add exception handling? I think we should discuss with the team (similar issue for `URL` and `URI`...)

PR #7821, opened 2021-07-19
feat: don't start queries when corruption is detected during startup

stevenpyzhang:
### Description 
If ksqlDB detects corruption during a restore it shouldn't start queries, because the queries may be terminated by a command following the offset at which we detect corruption. Instead, we should only apply DDLs and all queries should be reported as "failing/error" from the API.

I ended up mapping the Kafka Streams CREATE state to the KSQL Query state ERROR.

### Testing done 
Unit test

Also, started up a server with backup enables, and created 3 queries. Stopped the server and modified the backup file so that the last query created was different from what's in the command topic. Started up server again and verified that the corruption log was present and the `show queries` reported the 2 pre-corruption queries in ERROR state.

```
ksql> show queries;

 Query ID     | Query Type | Status  | Sink Name | Sink Kafka Topic | Query String                                                                                                                                  
----------------------------------------------------------------------------------------------------------------------------------------------------------------------------
 CSAS_TEST_1  | PERSISTENT | ERROR:1 | TEST      | TEST             | CREATE STREAM TEST WITH (KAFKA_TOPIC='TEST', PARTITIONS=1, REPLICAS=1) AS SELECT * FROM KSQL_PROCESSING_LOG KSQL_PROCESSING_LOG EMIT CHANGES; 
 CSAS_TEST2_3 | PERSISTENT | ERROR:1 | TEST2     | TEST2            | CREATE STREAM TEST2 WITH (KAFKA_TOPIC='TEST2', PARTITIONS=1, REPLICAS=1) AS SELECT * FROM TEST TEST EMIT CHANGES;                             
----------------------------------------------------------------------------------------------------------------------------------------------------------------------------
For detailed information on a Query run: EXPLAIN <Query ID>;
WARNING: The server has detected corruption in the command topic due to modifications performed on it. 
DDL statements will not be processed any further.
Queries may not run properly while the server is in a corrupted state.
If a backup of the command topic is available, restore the command topic using the backup file.
A server restart is required to restore full functionality.
```

```
ksql> show queries extended;
ID                   : CSAS_TEST2_3
Query Type           : PERSISTENT
SQL                  : CREATE STREAM TEST2 WITH (KAFKA_TOPIC='TEST2', PARTITIONS=1, REPLICAS=1) AS SELECT *
FROM TEST TEST
EMIT CHANGES;
Host Query Status    : {stevenz-mbp13.lan:8088=ERROR}
...
...
Error Date           : 2021-07-21 11:10:38,522 (PDT)
Error Details        : Query not started due to corruption in the command topic.
Error Type           : USER

WARNING: The server has detected corruption in the command topic due to modifications performed on it. 
DDL statements will not be processed any further.
Queries may not run properly while the server is in a corrupted state.
If a backup of the command topic is available, restore the command topic using the backup file.
A server restart is required to restore full functionality.
```
### Reviewer checklist


wcarlson5 (approved):
LGTM

ksqldb-common/src/main/java/io/confluent/ksql/util/KsqlConstants.java
```diff
@@ -63,7 +64,8 @@ private KsqlConstants() {
   }
 
   public static KsqlQueryStatus fromStreamsState(final KafkaStreams.State state) {
-    return state == KafkaStreams.State.ERROR ? KsqlQueryStatus.ERROR : KsqlQueryStatus.RUNNING;
+    return state == State.ERROR || state == State.CREATED
```
guozhangwang, line 67: very nit: maybe add a comment explaining why we map CREATED as ERROR as well, for future readers who are not aware of this PR.

guozhangwang (commented):
I have a small concern about mapping CREATED to ERROR: the transition of CREATED to REBALANCING is timing dependent, i.e. it would transit after it reached the broker (group coordinator) for the first time since starting up. In practice it is usually very fast, maybe in a few ms, but still before that the CREATED state is actually valid.

For list-queries / query-status purposes, if the broker was temporarily unavailable after the KS runtime is up and running, then the state may stuck in CREATED for a while, causing them to report false positives.

Some background context about this on JIRA: https://issues.apache.org/jira/browse/KAFKA-6520?src=confmacro

ksqldb-common/src/main/java/io/confluent/ksql/util/KsqlConstants.java
stevenpyzhang: not updating this mapping anymore

stevenpyzhang:
@guozhangwang @wcarlson5 I updated the PR so we're not updating the Streams state to Query state mapping. Instead, I just set a corruptionDetection variable in QueryMetadataImpl and the `QueryMetadata.getState()` returns KafkaStreams ERROR state if the corruption variable is set.

guozhangwang (approved):
LGTM!

wcarlson5 (approved):
Good idea

PR #7822, opened 2021-07-19
build: fix deprecation warnings from KIP-744

ableegoldman:
In [KIP-744](https://cwiki.apache.org/confluence/display/KAFKA/KIP-744%3A+Migrate+TaskMetadata+and+ThreadMetadata+to+an+interface+with+internal+implementation) we deprecated a number of classes and methods pertaining to the streams/thread/task metadata, and replaced those classes with interfaces under a different package. We just need to migrate from these deprecated APIs to the new ones.


ableegoldman:
cc @mjsax

ableegoldman:
Note that there may be a lurking bug in these new APIs behind a suddenly-flaky integration test in Streams. We should consider holding off on actually merging this until we've had a chance to do a cursory investigation into that test. See [KAFKA-13010](https://issues.apache.org/jira/browse/KAFKA-13010?focusedCommentId=17383445&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-17383445)

ksqldb-rest-model/src/main/java/io/confluent/ksql/rest/entity/StreamsTaskMetadata.java
```diff
@@ -59,7 +59,7 @@ public static StreamsTaskMetadata fromStreamsTaskMetadata(final TaskMetadata tas
       );
     }
     return new StreamsTaskMetadata(
-        taskMetadata.taskId(),
+        taskMetadata.taskId().toString(),
```
mjsax, line 62: Should we really pass in a `String` or update `StreamsTaskMetadata` to accept a `TaskId`?

mjsax (commented):
Leave it up to your judgement if it's safe to merge or not. One question though.

ksqldb-rest-model/src/main/java/io/confluent/ksql/rest/entity/StreamsTaskMetadata.java
wcarlson5: Yes this is just a wrapper class and should probably reflect the TaskMetadata. Though I am worried about what changing it will to the the ksqlHC

ableegoldman: I don't know enough about how this class is used to say how changing it from `String` to `TaskId` would do. On a similar note, I noticed (or rather, IntelliJ. noticed) that a number of the methods on this class are completely unused in the codebase. I guess that makes sense if they're public APIs used by the healthcheck, and I don't really know how (or if) the healthcheck does testing, but maybe we should have some test coverage within ksql too?

ableegoldman:
Ok I figured out what was causing that test failure and it's unrelated, so we should be good to merge this

ksqldb-rest-model/src/main/java/io/confluent/ksql/rest/entity/StreamsTaskMetadata.java
wcarlson5: The HC I don't think test this directly but it would cause some tests to fail so we would not worry about the HC not working unexpectedly. I think it might be better to just file a ticket to update the StreamsTaskMetadata and its uses in the HC and leave it as a String for now

PR #7829, opened 2021-07-20
chore: improve error messages for user columns with reserved names

Sullivan-Patrick:
### Description 
Previously, when trying to name a user column a reserved name (I'll use `ROWTIME` for this example), the error message was:
`'ROWTIME' is a reserved column name.`

Contrasted with trying to alias a reserved column name:
`ROWTIME is a reserved system column name. You cannot use it as an alias for a column.`

There are a lot of differences here which I'd like to iron out. The main idea is that these should be consistent, but the details I've chosen are:

- Apostrophes around the reserved column name should stay
- Drop the `system`. `Reserved column name` is enough.
- Keep the `You cannot use it as an {alias/name} for a column`

### Testing done 
_Changed relevant test files to expect new strings_

### Reviewer checklist

### DIFF
--- a/ksqldb-engine/src/main/java/io/confluent/ksql/ddl/commands/CreateSourceFactory.java
+++ b/ksqldb-engine/src/main/java/io/confluent/ksql/ddl/commands/CreateSourceFactory.java
@@ -227,7 +227,8 @@ public final class CreateSourceFactory {
 
     tableElements.forEach(e -> {
       if (SystemColumns.isSystemColumn(e.getName())) {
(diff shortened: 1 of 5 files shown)


vcrfxia (commented):
Thanks @Sullivan-Patrick -- LGTM once the build passes (looks like it might've just hit some flaky tests). I've also updated the PR title from `fix` to `chore` since this isn't really a bug fix, just an improved error message (users probably wouldn't be interested in seeing it in the changelog for releases).

vcrfxia (approved):

PR #7834, opened 2021-07-20
fix: Replace Functional test with Unit test

wcarlson5:
The functional test is basically solving the halting problem. So is this in away but its should be much quicker

### Reviewer checklist


lct45 (approved):
thanks @wcarlson5 ! lgtm

PR #7838, opened 2021-07-20
chore: avoid interpreting requests as file uploads

colinhicks:
The Vert.x BodyHandler's `create()` convenience constructor configures the handler to field file uploads by default. This means some requests, including those that don't specify a Content-Type header can be interpreted as file uploads.

This change disables the file upload support, since ksqlDB doesn't support such requests.


vcrfxia (approved):
Thanks @colinhicks !

PR #7843, opened 2021-07-21
feat: Add implementation of MetricsReporter for Confluent's telemetry

cadonna:
### Description 
This PR contains an implemenation of the MetricsReporter interface to report metrics to Confluent's 
telemetry framework. 

### Testing done 
unit tests

### Reviewer checklist

### DIFF
--- a/ksqldb-engine/src/main/java/io/confluent/ksql/internal/TelemetryMetricsReporter.java
+++ b/ksqldb-engine/src/main/java/io/confluent/ksql/internal/TelemetryMetricsReporter.java
@@ -1,0 +1,123 @@
+package io.confluent.ksql.internal;
+
+import com.google.common.collect.ImmutableMap;
+import io.confluent.telemetry.MetricKey;
+import io.confluent.telemetry.MetricsUtils;
+import io.confluent.telemetry.metrics.SinglePointMetric;
+import io.confluent.telemetry.provider.KsqlProvider;
+import io.confluent.telemetry.provider.ProviderRegistry;
+import io.confluent.telemetry.reporter.TelemetryReporter;
+import org.apache.kafka.common.metrics.KafkaMetricsContext;
+import org.apache.kafka.common.metrics.MetricsContext;
+import org.apache.kafka.common.utils.AppInfoParser;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.time.Instant;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+
+public class TelemetryMetricsReporter implements MetricsReporter {
+
+    public static final String RESOURCE_LABEL_PREFIX = "resource.";
+    public static final String RESOURCE_LABEL_TYPE = RESOURCE_LABEL_PREFIX + "type";
+    public static final String RESOURCE_LABEL_VERSION = RESOURCE_LABEL_PREFIX + "version";
+    public static final String RESOURCE_LABEL_COMMIT_ID = RESOURCE_LABEL_PREFIX + "commit.id";
+    public static final String RESOURCE_CLUSTER_ID = RESOURCE_LABEL_PREFIX + "cluster.id";
+
+    private final Logger logger = LoggerFactory.getLogger(TelemetryMetricsReporter.class);
+    private final TelemetryReporter reporter;
+    private final Map<String, String> metricTags = new HashMap<>();
+    // ToDo: How is the cluster ID passed to this class
+    private final String clusterId = "";
+    private final String metricGroup;
+
+
+    public TelemetryMetricsReporter(
+            final String metricGroup,
+            final String providerNamespace,
+            final String resourceType,
+            final Map<String, ?> config
+    ) {
+        this(metricGroup, providerNamespace, resourceType, config, new TelemetryReporter());
+    }
+
+    protected TelemetryMetricsReporter(
+            final String metricGroup,
+            final String providerNamespace,
+            final String resourceType,
+            final Map<String, ?> config,
+            final TelemetryReporter reporter
+    ) {
+        this.metricGroup = metricGroup;
+        this.reporter = reporter;
+        configureTelemetryReporter(providerNamespace, resourceType, config, reporter);
+    }
+
+    @Override
+    public void report(final List<DataPoint> dataPoints) {
+        for (final DataPoint dataPoint : dataPoints) {
+            emitMetric(
+                    dataPoint.getName(),
+                    dataPoint.getTime(),
+                    dataPoint.getValue(),
+                    ImmutableMap.<String, String>builder().putAll(metricTags).putAll(dataPoint.getTags()).build()
+            );
+        }
+    }
+
+    void emitMetric(
+            final String metricName,
+            final Instant timestamp,
+            final Object value,
+            final Map<String, String> metricTags
+    ) {
+        // Todo: Is it correct to use the KsqlProvider provided by ce-kafka?
+        final String fullMetricName = MetricsUtils.fullMetricName(
+                KsqlProvider.DOMAIN, metricGroup, metricName);
+        final MetricKey key = new MetricKey(fullMetricName, metricTags);
+
+        reporter.emitter().emitMetric(SinglePointMetric.gauge(key, (Double) value, timestamp));
+    }
+
+    private void configureTelemetryReporter(
+            final String providerNamespace,
+            final String resourceType,
+            final Map<String, ?> config,
+            final TelemetryReporter reporter
+    ) {
+        // Todo: Is it correct to use the KsqlProvider provided by ce-kafka?
+        ProviderRegistry.registerProvider(
+                providerNamespace,
+                KsqlProvider.class.getCanonicalName()
+        );
+
+        reporter.configure(config);
+        MetricsContext ctx = createMetricContext(providerNamespace, resourceType);
+        try {
+            reporter.contextChange(ctx);
+        } catch (final Throwable e) {
+            logger.error("Failed to properly start up telemetry reporter with the following error: ", e);
+            throw e;
+        }
+    }
+
+    private MetricsContext createMetricContext(
+            final String providerNameSpace,
+            final String resourcetype
+    ) {
+        Map<String, Object> rawConfig = new HashMap<>();
+        rawConfig.put(RESOURCE_LABEL_VERSION, AppInfoParser.getVersion());
+        rawConfig.put(RESOURCE_LABEL_TYPE, resourcetype);
+        rawConfig.put(RESOURCE_LABEL_COMMIT_ID, AppInfoParser.getCommitId());
+        rawConfig.put(RESOURCE_CLUSTER_ID, clusterId);
+
+        return new KafkaMetricsContext(providerNameSpace, rawConfig);
+    }
+
+    @Override
+    public void close() {
+        reporter.close();
+    }
+}
--- a/ksqldb-engine/src/main/java/io/confluent/ksql/internal/TelemetryMetricsReporter.java
+++ b/ksqldb-engine/src/main/java/io/confluent/ksql/internal/TelemetryMetricsReporter.java
@@ -1,3 +1,18 @@
+/*
+ * Copyright 2021 Confluent Inc.
+ *
+ * Licensed under the Confluent Community License (the "License"); you may not use
+ * this file except in compliance with the License.  You may obtain a copy of the
+ * License at
+ *
+ * http://www.confluent.io/confluent-community-license
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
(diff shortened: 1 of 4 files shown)


ksqldb-engine/src/main/java/io/confluent/ksql/internal/TelemetryMetricsReporter.java
```diff
@@ -0,0 +1,123 @@
...
+    public static final String RESOURCE_LABEL_TYPE = RESOURCE_LABEL_PREFIX + "type";
+    public static final String RESOURCE_LABEL_VERSION = RESOURCE_LABEL_PREFIX + "version";
+    public static final String RESOURCE_LABEL_COMMIT_ID = RESOURCE_LABEL_PREFIX + "commit.id";
+    public static final String RESOURCE_CLUSTER_ID = RESOURCE_LABEL_PREFIX + "cluster.id";
+
+    private final Logger logger = LoggerFactory.getLogger(TelemetryMetricsReporter.class);
+    private final TelemetryReporter reporter;
+    private final Map<String, String> metricTags = new HashMap<>();
+    // ToDo: How is the cluster ID passed to this class
+    private final String clusterId = "";
```
cadonna, line 33: Where do we get tags like cluster ID from?

ksqldb-engine/src/main/java/io/confluent/ksql/internal/TelemetryMetricsReporter.java
```diff
@@ -0,0 +1,123 @@
...
+    }
+
+    void emitMetric(
+            final String metricName,
+            final Instant timestamp,
+            final Object value,
+            final Map<String, String> metricTags
+    ) {
+        // Todo: Is it correct to use the KsqlProvider provided by ce-kafka?
+        final String fullMetricName = MetricsUtils.fullMetricName(
```
lct45, line 77: Yeah that should be correct - that's where the telemetry code is stored, this will just change the metric name to be their naming style

ksqldb-engine/src/main/java/io/confluent/ksql/internal/TelemetryMetricsReporter.java
```diff
@@ -0,0 +1,123 @@
...
+                KsqlProvider.DOMAIN, metricGroup, metricName);
+        final MetricKey key = new MetricKey(fullMetricName, metricTags);
+
+        reporter.emitter().emitMetric(SinglePointMetric.gauge(key, (Double) value, timestamp));
```
lct45, line 81: Just a nit / question - do we definitely want all of the metrics reported to telemetry to be doubles? I think the underlying disk metrics come as `BigInteger` and I'm not sure what @rodesai is planning for the streams side metrics

ksqldb-engine/src/main/java/io/confluent/ksql/internal/TelemetryMetricsReporter.java
lct45: @rodesai will know the structure better than I do but he and I talked a bit about this yesterday, cluster ID and node ID should come from the spec or scheduler call. In the HC we got them from a tag list in the config, they might be there for ksql as well, but I'd assume there's an easier way to pass it in than parsing a tag map 😀

ksqldb-engine/src/main/java/io/confluent/ksql/internal/TelemetryMetricsReporter.java
cadonna: Got it!

ksqldb-engine/src/main/java/io/confluent/ksql/internal/TelemetryMetricsReporter.java
cadonna: No, I just did consider all cases yet. I think this also something we need to discuss.

cadonna: `BigInteger` seem to be not supported by telemetry. A gauge can be a `long`, a `double`, or a distribution.

ksqldb-engine/src/main/java/io/confluent/ksql/internal/TelemetryMetricsReporter.java
```diff
@@ -0,0 +1,141 @@
...
+  public static final String RESOURCE_LABEL_TYPE = RESOURCE_LABEL_PREFIX + "type";
+  public static final String RESOURCE_LABEL_VERSION = RESOURCE_LABEL_PREFIX + "version";
+  public static final String RESOURCE_LABEL_COMMIT_ID = RESOURCE_LABEL_PREFIX + "commit.id";
+  public static final String RESOURCE_CLUSTER_ID = RESOURCE_LABEL_PREFIX + "cluster.id";
+
+  private final Logger logger = LoggerFactory.getLogger(TelemetryMetricsReporter.class);
+  private final TelemetryReporter reporter;
+  private final Map<String, String> metricTags = new HashMap<>();
+  // ToDo: How is the cluster ID passed to this class
+  private final String clusterId = "";
```
lct45, line 50: It doesn't look like this is getting configured anywhere, were you able to find where it's passed around?

ksqldb-engine/src/main/java/io/confluent/ksql/internal/TelemetryMetricsReporter.java
```diff
@@ -0,0 +1,141 @@
...
+
+  @Override
+  public void report(final List<DataPoint> dataPoints) {
+    for (final DataPoint dataPoint : dataPoints) {
+      emitMetric(
+              dataPoint.getName(),
+              dataPoint.getTime(),
+              dataPoint.getValue(),
+              ImmutableMap.<String, String>builder()
+                      .putAll(metricTags).putAll(dataPoint.getTags()).build()
```
lct45, line 83: If I'm following correctly, `metricTags` is always an empty map. Is it supposed to contain the basic tags in `createMetricContext`? Seems to me like it should at least contain the cluster / node ID since the metrics won't pass that in themselves through `dataPoint.getTags()`

ksqldb-engine/src/main/java/io/confluent/ksql/internal/TelemetryMetricsReporter.java
```diff
@@ -0,0 +1,141 @@
...
+    }
+  }
+
+  void emitMetric(
+          final String metricName,
+          final Instant timestamp,
+          final Object value,
+          final Map<String, String> metricTags
+  ) {
+    // Todo: Is it correct to use the KsqlProvider provided by ce-kafka?
```
lct45, line 94: nit: I think you left some of these comments around, not sure if it was on purpose

ksqldb-engine/src/main/java/io/confluent/ksql/internal/TelemetryMetricsReporter.java
```diff
@@ -0,0 +1,141 @@
...
+    final String fullMetricName = MetricsUtils.fullMetricName(
+            KsqlProvider.DOMAIN, metricGroup, metricName);
+    final MetricKey key = new MetricKey(fullMetricName, metricTags);
+
+    reporter.emitter().emitMetric(SinglePointMetric.gauge(key, (Double) value, timestamp));
```
lct45, line 99: I saw your comment from before about either doing doubles or longs. My preference would be to emit longs because they look cleaner to me but it seems like we usually do doubles in metrics, so maybe we want the precision you get with doubles. Thoughts? I'll transfer the `BigInteger` in disk util to whatever we decide

ksqldb-engine/src/main/java/io/confluent/ksql/internal/TelemetryMetricsReporter.java
```diff
@@ -0,0 +1,141 @@
...
+            KsqlProvider.class.getCanonicalName()
+    );
+
+    reporter.configure(config);
+    final MetricsContext ctx = createMetricContext(providerNamespace, resourceType);
+    try {
+      reporter.contextChange(ctx);
+    } catch (final Throwable e) {
+      logger.error("Failed to properly start up telemetry reporter with the following error: ", e);
+      throw e;
```
lct45, line 120: Do we definitely want to throw here? Just wondering if this would bubble up and cause the server to not start up - if so maybe we don't want to actually throw if the only side effect is not having metrics

ksqldb-engine/src/main/java/io/confluent/ksql/internal/TelemetryMetricsReporter.java
```diff
@@ -0,0 +1,141 @@
...
+
+  private MetricsContext createMetricContext(
+          final String providerNameSpace,
+          final String resourcetype
+  ) {
+    final Map<String, Object> rawConfig = new HashMap<>();
+    rawConfig.put(RESOURCE_LABEL_VERSION, AppInfoParser.getVersion());
+    rawConfig.put(RESOURCE_LABEL_TYPE, resourcetype);
+    rawConfig.put(RESOURCE_LABEL_COMMIT_ID, AppInfoParser.getCommitId());
+    rawConfig.put(RESOURCE_CLUSTER_ID, clusterId);
```
lct45, line 132: Just to double check - right now this will always be a "", right? Also @rodesai talked about wanting node ID but I'm not sure how to get that either, we'd want both cluster ID _and_ node ID, right? Alternatively, node ID can be through the metrics if we have access to it there

lct45 (commented):
Overall looks pretty good, will wait for @rodesai to take a peek as well

ksqldb-engine/src/main/java/io/confluent/ksql/internal/TelemetryMetricsReporter.java
guozhangwang: You mean by `ce-metrics` right?

ksqldb-engine/src/main/java/io/confluent/ksql/internal/TelemetryMetricsReporter.java
guozhangwang: Some metrics may be ratios, and hence would need double/float type I think?

ksqldb-engine/src/main/java/io/confluent/ksql/internal/TelemetryMetricsReporter.java
rodesai: I think it's ok to throw. I'd rather have us fail to start up than run in some undefined state where we're not emitting metrics. Anyway it's the type of thing that will almost always either work for all clusters running a given release or fail for all clusters.

ksqldb-engine/src/main/java/io/confluent/ksql/internal/TelemetryMetricsReporter.java
rodesai: we should had a node ID tag, but I'm not sure how to have it set correctly - let's sync with the telemetry folks tomorrow

rodesai (commented):
overall this looks good - we just need to check in with the observability team on how to have the node tag set

ksqldb-engine/src/main/java/io/confluent/ksql/internal/TelemetryMetricsReporter.java
rodesai: We will need to support a custom property that gets read by this class from the config map.

ksqldb-engine/src/main/java/io/confluent/ksql/internal/TelemetryMetricsReporter.java
```diff
@@ -0,0 +1,141 @@
...
+    reporter.emitter().emitMetric(SinglePointMetric.gauge(key, (Double) value, timestamp));
+  }
+
+  private void configureTelemetryReporter(
+          final String providerNamespace,
+          final String resourceType,
+          final Map<String, ?> config,
+          final TelemetryReporter reporter
+  ) {
+    // Todo: Is it correct to use the KsqlProvider provided by ce-kafka?
```
guozhangwang, line 108: I may have asked this before, but I think this is from ce-metrics not ce-kafka right? Anyways I think that's correct: the only other related provider I'm aware is the KsqlHealthCheckProvider from cc-ksql-healthcheck which seems definitely not what we want.

ksqldb-engine/src/main/java/io/confluent/ksql/internal/TelemetryMetricsReporter.java
guozhangwang: I'm also curious --- for node ID, how do we get its value? Would that be related to the pod name or some pre-specified value we would read from elsewhere?

ksqldb-engine/src/main/java/io/confluent/ksql/internal/TelemetryMetricsReporter.java
guozhangwang: +1. I think it should contain the common per-reporter tags, while so far it's always an empty map.

PR #7857, opened 2021-07-22
test: Exposes raw streams statuses in "show query" response for stabler tests

AlanConfluent:
### Description 
Currently, the RQTT tests for scalable push queries poll the query itself until it doesn't error.  Rather that do this to infer that the underlying push query is running, we want to issue the query `show queries;` directly and use that to know.  The only issue is that we don't expose raw kafka streams states, but rather a user friendly enum.  This adds the raw states to the response, which is ignored by the CLI, but can be used in tests as this one. 

### Testing done 
Ran updated tests.

### Reviewer checklist

### DIFF
--- a/ksqldb-functional-tests/src/test/java/io/confluent/ksql/test/rest/RestTestExecutor.java
+++ b/ksqldb-functional-tests/src/test/java/io/confluent/ksql/test/rest/RestTestExecutor.java
@@ -39,6 +39,8 @@ import io.confluent.ksql.rest.entity.KsqlEntity;
 import io.confluent.ksql.rest.entity.KsqlEntityList;
 import io.confluent.ksql.rest.entity.KsqlErrorMessage;
 import io.confluent.ksql.rest.entity.KsqlStatementErrorMessage;
+import io.confluent.ksql.rest.entity.Queries;
+import io.confluent.ksql.rest.entity.RunningQuery;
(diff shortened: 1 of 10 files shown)


agavra (approved):
I'm in a nit mood today, sorry! ✅

ksqldb-functional-tests/src/test/java/io/confluent/ksql/test/rest/RestTestExecutor.java
```diff
@@ -608,41 +613,32 @@ private void waitForWarmStateStore(
...
+      final RestResponse<KsqlEntityList> resp = restClient.makeKsqlRequest("SHOW QUERIES;");
       if (resp.isErroneous()) {
         final KsqlErrorMessage errorMessage = resp.getErrorMessage();
-        LOG.info("Server responded with an error code to a scalable push query. "
-            + "This could be because the persistent query hasn't started yet"
+        LOG.info("Server responded with an error code to show queries."
             + System.lineSeparator() + errorMessage);
         threadYield();
       } else {
-        resp.getResponse().close();
```
nateab, line 643: why don't we still close the response?

nateab: Nvm, after chatting with Almog clarified that this is because with the push query we keep getting results until we close, whereas with the `show queries` it is one and done

PR #7863, opened 2021-07-23
test: fix flaky version checker test by using simpler mock

agavra:
### Description 

This does two things:
1. fixes a flaky test by using a mock wire instead of spinning up an entire mocked proxy server 
2. removes a dependency that has historically caused us some grief as it depends on specific netty versions

Rejoice!

### Testing done 

Test would fail 1/10 runs locally before this change, didn't fail with 1k runs afterwards.

### Reviewer checklist

### DIFF
--- a/ksqldb-version-metrics-client/pom.xml
+++ b/ksqldb-version-metrics-client/pom.xml
@@ -56,35 +56,15 @@
         </dependency>
 
         <dependency>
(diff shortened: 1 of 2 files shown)


lct45 (approved):
Overall LGTM from what I can tell 😉 thanks @agavra

PR #7868, opened 2021-07-23
feat: add storage utilization metrics

lct45:
### Description 
As part of the observability metrics we're adding in Q3, we'd like to see storage metrics. This adds storage metrics by node, task, and query. 

### Testing done 
Tested locally that the metrics picked up, added unit tests, did benchmarking

### Reviewer checklist


ksqldb-engine/src/main/java/io/confluent/ksql/internal/UtilizationMetricsListener.java
```diff
@@ -15,52 +15,137 @@
...
+
+  public void nodeDiskUsage(
+      final List<MetricsReporter.DataPoint> dataPoints,
+      final Instant sampleTime) {
+    for (File f : streamsDirectories) {
+      final double freeSpace = f.getFreeSpace();
+      final long totalSpace = f.getTotalSpace();
+      final double percFree = percentage(freeSpace, (double) totalSpace);
+      dataPoints.add(new MetricsReporter.DataPoint(sampleTime,"storage-usage", freeSpace));
+      dataPoints.add(new MetricsReporter.DataPoint(sampleTime,"storage-usage-perc", percFree));
```
lct45, line 105: I found that looking at the raw free space number didn't give me a sense of how full my disk was, but maybe other people automatically know how much space there is total? For task level we don't have total file size so maybe we don't want it here either for consistency

ksqldb-engine/src/main/java/io/confluent/ksql/internal/UtilizationMetricsListener.java
```diff
@@ -15,52 +15,138 @@
...
   }
+
+  public void nodeDiskUsage(
+      final List<MetricsReporter.DataPoint> dataPoints,
+      final Instant sampleTime) {
+    for (File f : streamsDirectories) {
+      final long totalSpace = f.getTotalSpace();
+      final long usedSpace = totalSpace - f.getFreeSpace();
+      final double percFree = percentage((double) usedSpace, (double) totalSpace);
+      dataPoints.add(new MetricsReporter.DataPoint(sampleTime,"storage-usage", (double) usedSpace));
```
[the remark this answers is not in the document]
guozhangwang: Also: seems we are only keep three data points here since the next file's usage would simply overwrite the previous ones, is that intentional??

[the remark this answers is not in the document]
lct45: Hmm yeah good point @guozhangwang. Looking back at this I'm really reporting storage usage by query which I do later. I switched it to aggregate all of the files and return the final result as the node reporting. I saw your comment back on the one-pager about if getting the query storage usage here with `f.getFreeSpace` will be more accurate than aggregating `sst-file-size`, I'll leave log lines in and see if I can get a read on that during benchmarks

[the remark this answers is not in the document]
guozhangwang: Makes sense, thanks.

ksqldb-engine/src/main/java/io/confluent/ksql/internal/UtilizationMetricsListener.java
```diff
@@ -15,52 +15,258 @@
...
+      metricRegistry.addMetric(
+        metricRegistry.metricName(
+          "query-storage-usage",
+          "ksql-utilization-metrics",
+          ImmutableMap.of("query-id", queryId)),
+        (Gauge<Object>) (config, now) -> queryStorageMetrics.get(queryId)
+      );
+      metrics.put(queryId, new ArrayList<>());
+    }
+    // We've seen metric for this query before
```
guozhangwang, line 135: nit: for this query's task

ksqldb-engine/src/main/java/io/confluent/ksql/internal/UtilizationMetricsListener.java
```diff
@@ -15,52 +15,258 @@
...
-public class UtilizationMetricsListener implements Runnable, QueryEventListener {
+public class UtilizationMetricsListener implements MetricsReporter {
+  private final ConcurrentHashMap<String, Long> queryStorageMetrics;
+  private final Map<String, List<String>> metrics;
+  private final Metrics metricRegistry;
+  private final File baseDir;
+  private final Collection<TaskStorageMetric> registeredMetrics;
+
+  public UtilizationMetricsListener(final KsqlConfig config, final Metrics metricRegistry) {
+    this.queryStorageMetrics = new ConcurrentHashMap<>();
```
guozhangwang, line 51: Could you elaborate a bit more about the differences of these three collections: the `metrics` map, the `queryStorageMetrics` map, and the `registeredMetrics` set. From what I can see it seems:

* `queryStorageMetrics` is used for the per-query metrics, but it was not populated and updated anywhere.
* `registeredMetrics` is used for per-query-task metrics
* `metrics` is also used for storing per-query metrics?

The relationships of these and how they are leveraged / updated are all not very clear to me.

ksqldb-engine/src/main/java/io/confluent/ksql/internal/UtilizationMetricsListener.java
```diff
@@ -15,52 +15,258 @@
...
+          ImmutableMap.of("query-id", queryId)),
+        (Gauge<Object>) (config, now) -> queryStorageMetrics.get(queryId)
+      );
+      metrics.put(queryId, new ArrayList<>());
+    }
+    // We've seen metric for this query before
+    if (metrics.get(queryId).contains(taskId)) {
+      // we have this task metric already, just need to update it
+      resetMetric(metric);
+      for (TaskStorageMetric storageMetric : registeredMetrics) {
```
guozhangwang, line 139: Why not put this for loop into the `resetMetric` as well??

guozhangwang (commented):
Made a second pass; but it seems this PR is not complete and ready for review yet? @lct45 please feel free to ping me when it's done.

ksqldb-engine/src/main/java/io/confluent/ksql/internal/UtilizationMetricsListener.java
```diff
@@ -15,52 +15,256 @@
...
+      metricRegistry.addMetric(
+        metricRegistry.metricName(
+          "query-storage-usage",
+          "ksql-utilization-metrics",
+          ImmutableMap.of("query-id", queryId)),
+        (Gauge<BigInteger>) (config, now) -> computeQueryMetric(queryId)
+      );
+      metricsSeen.put(queryId, new ArrayList<>());
+    }
+    final TaskStorageMetric newMetric = new TaskStorageMetric(
```
rodesai, line 133: I think you can simplify this bit with a few changes:
- make `metricsSeen` a map from query id to a map from task id to `TaskStorageMetric` (e.g. `Map<String, Map<String, TaskStorageMetric>`)
- there's no need for `resetAndUpdateMetric` - you just need to call `add` on `TaskStorageMetric` and the value will get replaced in the map. Also technically that case should never actually happen - consider adding a warning log

Then this bit can just be:
```
final TaskStorageMetric taskMetric;
if (!metricsSeen.get(queryId).contains(taskId)) {
    taskMetric = new TaskStorageMetric(...);
    metricsSeen.get(queryId).add(taskId, new taskMetric);
    metricRegistry.addMetric(...)
} else {
    taskmetric = metricsSeen.get(queryId).get(taskId);
}
taskMetric.add(metric);
```

This should simplify removal too since you don't have to iterate the whole list.

ksqldb-engine/src/main/java/io/confluent/ksql/internal/UtilizationMetricsListener.java
```diff
@@ -15,52 +15,256 @@
...
+    if (!metric.metricName().name().equals("total-sst-files-size")) {
+      return;
+    }
+    final String taskId = metric.metricName().tags().getOrDefault("task-id", "");
+    final String queryIdTag = metric.metricName().tags().getOrDefault("thread-id", "");
+    Pattern pattern = Pattern.compile("(?<=query_)(.*?)(?=-)");
+    Matcher matcher = pattern.matcher(queryIdTag);
+    final String queryId = matcher.find() ? matcher.group(1) : "";
+    // if we haven't seen a task for this query yet
+    if (!metricsSeen.containsKey(queryId)) {
```
rodesai, line 123: everything from this line down to the end of this method needs to be in a synchronized block. I wouldn't synchronize this whole method though because most of the time it's called it doesn't need to do anything (since most metrics will fail the initial check).

ksqldb-engine/src/main/java/io/confluent/ksql/internal/UtilizationMetricsListener.java
```diff
@@ -15,52 +15,256 @@
...
+
+  public static class TaskStorageMetric {
+    final MetricName metricName;
+    private final Map<MetricName, KafkaMetric> metrics = new ConcurrentHashMap<>();
+
+    TaskStorageMetric(final MetricName metricName) {
+      this.metricName = metricName;
+    }
+
+    private void add(final KafkaMetric metric) {
```
rodesai, line 236: add, remove, and getValue need to be synchronized - streams threads may be creating/removing state stores while telemetry is reading the value

ksqldb-engine/src/main/java/io/confluent/ksql/internal/UtilizationMetricsListener.java
```diff
@@ -15,52 +15,256 @@
...
+    final List<TaskStorageMetric> taskMetrics = metricsSeen.get(queryId);
+    taskMetrics.forEach(r -> r.remove(metric));
+    for (TaskStorageMetric storageMetric : taskMetrics) {
+      if (storageMetric.getTaskId().equals(taskId)) {
+        storageMetric.add(metric);
+      }
+    }
+  }
+
+  final BigInteger computeQueryMetric(final String queryId) {
```
rodesai, line 220: this needs to be synchronized with metric addition and removal.

ksqldb-engine/src/main/java/io/confluent/ksql/internal/UtilizationMetricsListener.java
```diff
@@ -15,52 +15,256 @@
...
+      dir = map.get(StreamsConfig.STATE_DIR_CONFIG).toString();
+    } else {
+      dir =  StreamsConfig.configDef().defaultValues().get(StreamsConfig.STATE_DIR_CONFIG).toString();
+    }
+    configureShared(new File(dir), this.metricRegistry);
+  }
+
+  private static void configureShared(final File baseDir, final Metrics metricRegistry) {
+    synchronized (lock) {
+      if (registeredNodeMetrics != null) {
```
rodesai, line 85: Rather than keeping the registered metrics in a collection, why not just use a single boolean that you set to true once the node-level metrics are registered? You actually don't even need a lock in that case - it can just be an atomic boolean:

```
static AtomicBoolean  registeredNodeMetrics = new AtomicBoolean(false);

private static void configureShared(final File baseDir, final Metrics metricsRegistry) {
    if (registeredNodeMetrics.getAndSet(true)) {
        // already registered;
        return;
    }
    // add the node level metrics
}
```

ksqldb-engine/src/main/java/io/confluent/ksql/internal/UtilizationMetricsListener.java
rodesai: never mind - noticed this is a `ConcurrentHashMap`

ksqldb-engine/src/main/java/io/confluent/ksql/internal/UtilizationMetricsListener.java
rodesai: Also, to be extra careful I would call `metricValue` outside of the lock. `metricValue` is going to call into Streams, which could in theory (though very unlikely) try to take a lock that another stream thread holds while waiting to get our lock (because that thread is trying to register a metric).

ksqldb-engine/src/main/java/io/confluent/ksql/internal/UtilizationMetricsListener.java
lct45: What case should never happen, calling add on an existing `TaskStorageMetric`? The rest of this makes sense and makes it cleaner

ksqldb-engine/src/main/java/io/confluent/ksql/internal/UtilizationMetricsListener.java
lct45: Just to clarify, is this still true since the metric map is a concurrent hash map?

ksqldb-engine/src/main/java/io/confluent/ksql/internal/UtilizationMetricsListener.java
lct45: Synchronized on `metricsSeen`, not `metricsRegistry`, right?

ksqldb-engine/src/main/java/io/confluent/ksql/internal/StorageUtilizationMetrics.java
```diff
@@ -0,0 +1,254 @@
...
+  private static void configureShared(final File baseDir, final Metrics metricRegistry) {
+    if (registeredNodeMetrics.getAndSet(true)) {
+      return;
+    }
+    final MetricName nodeAvailable =
+        metricRegistry.metricName("node-storage-available", METRIC_GROUP);
+    final MetricName nodeTotal =
+        metricRegistry.metricName("node-storage-total", METRIC_GROUP);
+    final MetricName nodeUsed =
+        metricRegistry.metricName("node-storage-used", METRIC_GROUP);
```
lct45, line 84: @rodesai I can't remember what we landed on for node metrics - IIRC calculated storage used by doing (f.getTotalSpace - f.getFreeSpace) isn't actually accurate for what we want. Given that, do we just want to report f.getFreeSpace? Or f.getFreeSpace and f.getTotalSpace?

lct45: Additionally for naming these + the other storage metrics, do we want to add a unit suffix? eg `_bytes`

rodesai: > IIRC calculated storage used by doing (f.getTotalSpace - f.getFreeSpace) isn't actually accurate for what we want. Given that, do we just want to report f.getFreeSpace? Or f.getFreeSpace and f.getTotalSpace?

I thought we determined it was probably accurate, and the disparity we saw was very small (18MB) and probably explained by the workload continuing to run between the time you printed the metrics and the time you looked at `df`. So we decided to get a PR out and we can test on a cloud instance rather than on our macbooks.

And yeah a `_bytes` suffix makes sense - good call

ksqldb-engine/src/main/java/io/confluent/ksql/internal/StorageUtilizationMetrics.java
```diff
@@ -0,0 +1,254 @@
...
+    final Pattern pattern = Pattern.compile("(?<=query_|transient_)(.*?)(?=-)");
+    final Matcher matcher = pattern.matcher(queryIdTag);
+    return matcher.find() ? matcher.group(1) : "";
+  }
+
+  static void reset() {
+    registeredNodeMetrics.getAndSet(false);
+  }
+
+  public static class TaskStorageMetric {
```
rodesai, line 216: make this private

ksqldb-engine/src/main/java/io/confluent/ksql/internal/StorageUtilizationMetrics.java
```diff
@@ -0,0 +1,254 @@
...
+  @Override
+  public void metricChange(final KafkaMetric metric) {
+    if (!metric.metricName().name().equals("total-sst-files-size")) {
+      return;
+    }
+    final String taskId = metric.metricName().tags().getOrDefault("task-id", "");
+    final String queryId = getQueryId(metric);
+
+    // if we haven't seen a task for this query yet
+    synchronized (metricsSeen) {
```
rodesai, line 109: can you move this into another method like `private synchronized void handleNewSstFilesSizeMetric(final KafkaMetric metric, final String taskId, final String queryId) {...`?

ditto for the other side of this in `metricRemoval`

ksqldb-engine/src/main/java/io/confluent/ksql/internal/StorageUtilizationMetrics.java
```diff
@@ -0,0 +1,254 @@
...
+  @Override
+  public void metricRemoval(final KafkaMetric metric) {
+    final MetricName metricName = metric.metricName();
+    if (!metricName.name().equals("total-sst-files-size")) {
+      return;
+    }
+
+    final String queryId = getQueryId(metric);
+    final String taskId = metric.metricName().tags().getOrDefault("task-id", "");
+    final TaskStorageMetric taskMetric = metricsSeen.get(queryId).get(taskId);
```
rodesai, line 154: move everything below this into another method like:

`private synchronized void handleRemovedSstFileSizeMetric(...`

ksqldb-engine/src/main/java/io/confluent/ksql/internal/StorageUtilizationMetrics.java
```diff
@@ -0,0 +1,254 @@
...
+  }
+
+  @Override
+  public void contextChange(final MetricsContext metricsContext) {
+
+  }
+
+  private BigInteger computeQueryMetric(final String queryId) {
+    BigInteger queryMetricSum = BigInteger.ZERO;
+    for (Map.Entry<String, TaskStorageMetric> entry : metricsSeen.get(queryId).entrySet()) {
```
rodesai, line 199: we need to synchronize the bit here that iterates over the map. However we don't want to synchronize the part that gets the metric (to eliminate the risk of a deadlock). So you can write this like:

```
private BigInteger computeQueryMetric(final String queryId) {
    BigInteger queryMetricSum = BigInteger.ZERO;
    for (final Supplier<BigInteger> gauge : getGaugesForQuery(queryId)) {
        queryMetricSum = queryMetricSum.add(gauge.get());
    }
    return queryMetricSum;
}

private synchronized Collection<Supplier<BigInteger>> getGaugesForQuery(final String queryId) {
    return metricsSeen.get(queryId).values().stream()
       .map(v -> () -> v.getValue())
}
```

ksqldb-engine/src/main/java/io/confluent/ksql/internal/StorageUtilizationMetrics.java
lct45: And we have to do this because the hashmap is a concurrent hashmap, right? So when we access it we need to do it in a synchronized fashion? Is there something different about using a synchronized function rather than a synchronized block or is the synchronized function easier to use

ksqldb-engine/src/main/java/io/confluent/ksql/internal/StorageUtilizationMetrics.java
```diff
@@ -0,0 +1,280 @@
...
+    return matcher.find() ? matcher.group(1) : "";
+  }
+
+  static void reset() {
+    registeredNodeMetrics.getAndSet(false);
+  }
+
+  private static class TaskStorageMetric {
+    final MetricName metricName;
+    private final Map<MetricName, KafkaMetric> metrics = new ConcurrentHashMap<>();
```
guozhangwang, line 244: Hmm, are we having more than one metric here? Since we are filtering on `metric.metricName().name().equals("total-sst-files-size")` at the first place, we would end up with only that metric name right?

ksqldb-engine/src/main/java/io/confluent/ksql/internal/StorageUtilizationMetrics.java
```diff
@@ -0,0 +1,280 @@
...
+  }
+
+  private String getQueryId(final KafkaMetric metric) {
+    final String queryIdTag = metric.metricName().tags().getOrDefault("thread-id", "");
+    final Pattern pattern = Pattern.compile("(?<=query_|transient_)(.*?)(?=-)");
+    final Matcher matcher = pattern.matcher(queryIdTag);
+    return matcher.find() ? matcher.group(1) : "";
+  }
+
+  static void reset() {
```
guozhangwang, line 238: nit: add a comment indicating it is only for testing?

ksqldb-engine/src/main/java/io/confluent/ksql/internal/StorageUtilizationMetrics.java
```diff
@@ -0,0 +1,280 @@
...
+  }
+
+  private synchronized void handleRemovedSstFileSizeMetric(
+      final TaskStorageMetric taskMetric,
+      final KafkaMetric metric,
+      final String queryId,
+      final String taskId
+  ) {
+    // remove storage metric for this task
+    taskMetric.remove(metric);
```
guozhangwang, line 201: See my other comment: if we would only ever have one metric under that task, then we can simplify this a bit.

ksqldb-engine/src/main/java/io/confluent/ksql/internal/StorageUtilizationMetrics.java
```diff
@@ -0,0 +1,280 @@
...
+import org.apache.kafka.common.MetricName;
+import org.apache.kafka.common.config.ConfigException;
+import org.apache.kafka.common.metrics.Gauge;
+import org.apache.kafka.common.metrics.KafkaMetric;
+import org.apache.kafka.common.metrics.Metrics;
+import org.apache.kafka.common.metrics.MetricsContext;
+import org.apache.kafka.common.metrics.MetricsReporter;
+import org.apache.kafka.streams.StreamsConfig;
+
+public class StorageUtilizationMetrics implements MetricsReporter {
```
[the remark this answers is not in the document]
lct45: @rodesai may have something more specific here but IIUC `io.confluent.ksql.internal.MetricsReporter` is used [here](https://github.com/confluentinc/ksql/pull/7955/files#diff-4c1031f82c6992cfa0176cf66f3857a1902b9ace4d2afa5fd7fd55e0b58fc15e) to report a ksql metric object `dataPoint` and `org.apache.kafka.common.metrics.MetricsReporter` is used for `KafkaMetric`s

ksqldb-engine/src/main/java/io/confluent/ksql/internal/StorageUtilizationMetrics.java
```diff
@@ -0,0 +1,280 @@
...
+    return metricsSeen.get(queryId).values().stream()
+      .map(v -> (Supplier<BigInteger>) v::getValue)
+      .collect(Collectors.toList());
+  }
+
+  private String getQueryId(final KafkaMetric metric) {
+    final String queryIdTag = metric.metricName().tags().getOrDefault("thread-id", "");
+    final Pattern pattern = Pattern.compile("(?<=query_|transient_)(.*?)(?=-)");
+    final Matcher matcher = pattern.matcher(queryIdTag);
+    return matcher.find() ? matcher.group(1) : "";
```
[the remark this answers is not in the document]
lct45: We should always find a match for queries, although my regex skills aren't incredible so it depends on if anyone else sees any issues on L 233 😉.  But I think you're right - we should treat it as a bug. I'm hesitant to crash the app if the queryId isn't showing up for a metric, on the other hand, if we only log an error we likely won't know that we're missing metrics... We also don't want to continue reporting any of these metrics if we don't have a queryID imo, so maybe throwing is the right way to go. LMK what you think

ksqldb-engine/src/main/java/io/confluent/ksql/internal/StorageUtilizationMetrics.java
lct45: From what I understand, `total-sst-files-size` is broken down by store, so we may have multiple stores per task. The `storeName` is included in the tag map so the metric names would be different if you had multiple stores under one task

ksqldb-engine/src/main/java/io/confluent/ksql/internal/StorageUtilizationMetricsReporter.java
```diff
@@ -0,0 +1,288 @@
...
+
+    metricRegistry.addMetric(
+        nodeAvailable,
+        (Gauge<Long>) (config, now) -> baseDir.getFreeSpace()
+    );
+    metricRegistry.addMetric(
+        nodeTotal,
+        (Gauge<Long>) (config, now) -> baseDir.getTotalSpace()
+    );
+    metricRegistry.addMetric(
```
rodesai, line 100: can you add a `storage_utilization` metric that returns ((baseDir.getTotalSpace - baseDir.getFreeSpace) / baseDir.getTotalSpace)

lct45: There should be no risk of division by 0 here, right? Since our `baseDir` should always exist

lct45: And do you want this all the way up to the metrics api + exposed to users?

rodesai: > There should be no risk of division by 0 here, right? Since our baseDir should always exist

yeah that should never happen

> And do you want this all the way up to the metrics api + exposed to users?

yeah, the max value of this metric (over all the nodes) is what we want to show users in cloud

rodesai (approved):
LGTM

ksqldb-engine/src/main/java/io/confluent/ksql/internal/StorageUtilizationMetrics.java
guozhangwang: Ah yes, you're right :)

guozhangwang:
LGTM.

rodesai (approved):
LGTM

ksqldb-engine/src/main/java/io/confluent/ksql/internal/StorageUtilizationMetricsReporter.java
```diff
@@ -0,0 +1,286 @@
...
+import org.apache.kafka.common.MetricName;
+import org.apache.kafka.common.config.ConfigException;
+import org.apache.kafka.common.metrics.Gauge;
+import org.apache.kafka.common.metrics.KafkaMetric;
+import org.apache.kafka.common.metrics.Metrics;
+import org.apache.kafka.common.metrics.MetricsContext;
+import org.apache.kafka.common.metrics.MetricsReporter;
+import org.apache.kafka.streams.StreamsConfig;
+
+public class StorageUtilizationMetricsReporter implements MetricsReporter {
```
sirianni, line 46: I don't understand why you are implementing a new `MetricsReporter` here.

A `MetricsReporter` is for _reporting_ metrics via different mechanisms (JMX, etc.).  For defining new metrics, you typically just create a new `KafkaMetric` and call [Metrics.addMetric()](https://github.com/confluentinc/ce-kafka/blob/master/clients/src/main/java/org/apache/kafka/common/metrics/Metrics.java#L490).

PR #7869, opened 2021-07-23
chore: infrastructure for rowoffset rowpartition

Sullivan-Patrick:
### Description 
As part of [KLIP-50](https://github.com/confluentinc/ksql/blob/master/design-proposals/klip-50-partition-and-offset-in-ksqldb.md), this PR focuses on adding infrastructure necessary for the mapping of a query's pseudocolumns to a pseudocolumn version, as well as piping pseudoColumnVersion to functions which will read an existing query and call `withPseudoAndKeyColsInValue`. It also contains the logic needed to populate `ROWOFFSET/ROWPARTITION` pseudocolumns.

### Testing done 
_No testing at the moment as no functionality was changed or added- existing coverage should cover everything at this moment._

### Reviewer checklist


ksqldb-common/src/main/java/io/confluent/ksql/schema/ksql/LogicalSchema.java
```diff
@@ -123,10 +127,23 @@ public boolean valueContainsAny(final Set<ColumnName> names) {
...
    * WINDOWEND} columns will added to the value schema to represent the window bounds.
+   *
+   * @param pseudoColumnVersion indicates the pseudocolumn version number of the given query.
+   * Queries passed to this method will receive the corresponding pseudocolumns in their schema.
+   *
    * @return the new schema.
    */
+  public LogicalSchema withPseudoAndKeyColsInValue(
+      final boolean windowed, final int pseudoColumnVersion) {
+    return rebuild(//this will use current pseudocolumn version when feature is fully released
```
vcrfxia, line 138: Rather than leaving comments like this, we should set the current pseudocolumn version to be the same as legacy version for now and update this to use the current pseudocolumn version instead of the legacy one.

ksqldb-common/src/main/java/io/confluent/ksql/schema/ksql/LogicalSchema.java
```diff
@@ -123,10 +127,23 @@ public boolean valueContainsAny(final Set<ColumnName> names) {
...
+        true, windowed, SystemColumns.LEGACY_PSEUDOCOLUMN_VERSION_NUMBER);
+  }
+
+  //This will eventually use the current pseudocolumn version whenever it is not explicitly
+  //provided (as we WILL explicitly provide it whenever we are reading old queries)
   public LogicalSchema withPseudoAndKeyColsInValue(final boolean windowed) {
-    return rebuild(true, windowed);
+    return rebuild(
+        true, windowed, SystemColumns.LEGACY_PSEUDOCOLUMN_VERSION_NUMBER);
```
vcrfxia, line 146: Same as above.

ksqldb-common/src/main/java/io/confluent/ksql/schema/ksql/LogicalSchema.java
```diff
@@ -135,12 +152,12 @@ public LogicalSchema withPseudoAndKeyColsInValue(final boolean windowed) {
    * @return the new schema with the columns removed.
    */
   public LogicalSchema withoutPseudoAndKeyColsInValue() {
-    return rebuild(false, false);
+    return rebuild(false, false, -1);
```
vcrfxia, line 155: What does -1 mean? Should this be the current pseudocolumn version instead?

ksqldb-common/src/main/java/io/confluent/ksql/schema/ksql/LogicalSchema.java
```diff
@@ -263,6 +281,11 @@ private LogicalSchema rebuild(
     if (withPseudoAndKeyColsInValue) {
       builder.add(Column.of(ROWTIME_NAME, ROWTIME_TYPE, VALUE, valueIndex++));
 
+      if (pseudoColumnVersion > 0) {
+        builder.add(Column.of(ROWOFFSET_NAME, ROWOFFSET_TYPE, VALUE, valueIndex++));
```
vcrfxia, line 285: Understanding this code requires knowledge that pseudocolumn version 0 means rowtime is the only pseudocolumn, whereas pseudocolumn version 1 means rowpartition and rowoffset should be added too. This is "magic" knowledge that should be avoided.

How about we add new constants such as `ROWTIME_PSEUDOCOLUMN_VERSION = 0` and `ROWPARTITION_ROWOFFSET_PSEUDOCOLUMN_VERSION = 1` so that this logic can instead become
```
if (pseudoColumnVersion >= ROWTIME_PSEUDOCOLUMN_VERSION) {
  // add in rowtime
}
if (pseudoColumnVersion >= ROWPARTITION_ROWOFFSET_PSEUDOCOLUMN_VERSION) {
  // add in rowpartition and rowoffset
}
```

ksqldb-common/src/main/java/io/confluent/ksql/schema/ksql/SystemColumns.java
```diff
@@ -15,40 +15,68 @@
...
 
   public static final int LEGACY_PSEUDOCOLUMN_VERSION_NUMBER = 0;
-  public static final int CURRENT_PSEUDOCOLUMN_VERSION_NUMBER = 5;
+  public static final int CURRENT_PSEUDOCOLUMN_VERSION_NUMBER = 0;
+
+  public static Set<ColumnName> getPseudoColumnsFromVersion(final int version) {
+    return versionedPseudoColumns.get(version);
+  }
+
+  private static final Set<ColumnName> VERSION_ONE_NAMES = ImmutableSet.of(
```
vcrfxia, line 50: See my comment above for a suggestion on how to restructure this.

ksqldb-common/src/main/java/io/confluent/ksql/schema/ksql/SystemColumns.java
```diff
@@ -65,14 +93,26 @@ public static boolean isPseudoColumn(final ColumnName columnName) {
...
   public static boolean isSystemColumn(final ColumnName columnName) {
     return systemColumnNames().contains(columnName);
   }
 
   public static Set<ColumnName> systemColumnNames() {
-    return SYSTEM_COLUMN_NAMES;
+    return SYSTEM_COLUMN_NAMES_CURRENT;
+  }
+
+  //cache version numbers we have used, otherwise create as needed
```
vcrfxia, line 107: I'm not convinced this is optimization (the cache) is necessary. As long as this method is not on the critical path (i.e., we don't call it once per message processed), then we don't need to over-optimize at the expense of additional code complexity. 

(That said, this type of caching is a neat trick! File it away for the future.)

ksqldb-common/src/main/java/io/confluent/ksql/schema/ksql/SystemColumns.java
```diff
@@ -65,14 +93,26 @@ public static boolean isPseudoColumn(final ColumnName columnName) {
...
+  public static Set<ColumnName> systemColumnNames(final int versionNumber) {
+    if (!systemColumnNamesByVersion.containsKey(versionNumber)) {
```
vcrfxia, line 109: You can use `putIfAbsent()` to simplify this.

ksqldb-streams/src/main/java/io/confluent/ksql/execution/streams/SourceBuilder.java
```diff
@@ -508,8 +514,19 @@ public GenericRow transform(final K key, final GenericRow row) {
           final long timestamp = processorContext.timestamp();
           final Collection<?> keyColumns = keyGenerator.apply(key);
 
-          row.ensureAdditionalCapacity(1 + keyColumns.size());
+          final int additionalColumns = SystemColumns
```
vcrfxia, line 517:
```suggestion
          final int numPseudoColumns = SystemColumns
```
I think the name `additionalColumns` is confusing because the key columns are "additional" in some sense too.

ksqldb-streams/src/main/java/io/confluent/ksql/execution/streams/SourceBuilder.java
```diff
@@ -508,8 +514,19 @@ public GenericRow transform(final K key, final GenericRow row) {
...
+              .getPseudoColumnsFromVersion(pseudoColumnVersion).size();
+
+          row.ensureAdditionalCapacity(additionalColumns + keyColumns.size());
           row.append(timestamp);
+
+          if (pseudoColumnVersion > 0) {
```
vcrfxia, line 523: See my comment above on how to make this more readable (by avoiding magic numbers).

ksqldb-streams/src/main/java/io/confluent/ksql/execution/streams/StepSchemaResolver.java
```diff
@@ -353,12 +353,14 @@ private LogicalSchema sameSchema(final LogicalSchema schema, final ExecutionStep
     return schema;
   }
 
+  //todo: pipe pseudocolumnVersion here
```
vcrfxia, line 356: What's the comment here for? Is this going to be part of this PR?

ksqldb-common/src/main/java/io/confluent/ksql/schema/ksql/SystemColumns.java
```diff
@@ -15,40 +15,68 @@
...
   public static final ColumnName WINDOWSTART_NAME = ColumnName.of("WINDOWSTART");
   public static final ColumnName WINDOWEND_NAME = ColumnName.of("WINDOWEND");
   public static final SqlType WINDOWBOUND_TYPE = SqlTypes.BIGINT;
 
   public static final int LEGACY_PSEUDOCOLUMN_VERSION_NUMBER = 0;
-  public static final int CURRENT_PSEUDOCOLUMN_VERSION_NUMBER = 5;
+  public static final int CURRENT_PSEUDOCOLUMN_VERSION_NUMBER = 0;
+
+  public static Set<ColumnName> getPseudoColumnsFromVersion(final int version) {
+    return versionedPseudoColumns.get(version);
```
vcrfxia, line 47: We need to handle the case where an invalid version number is passed. (Throw an informative exception.)

ksqldb-common/src/main/java/io/confluent/ksql/schema/ksql/LogicalSchema.java
```diff
@@ -123,10 +127,23 @@ public boolean valueContainsAny(final Set<ColumnName> names) {
    *
    * @param windowed indicates that the source is windowed; meaning {@code WINDOWSTART} and {@code
    * WINDOWEND} columns will added to the value schema to represent the window bounds.
+   *
+   * @param pseudoColumnVersion indicates the pseudocolumn version number of the given query.
```
vcrfxia, line 131: nit: the first part of this docstring is not very informative since it simply repeats the name of the parameter. How about something like "version number that determines the set of pseudocolumns to be added to the schema" or even simply "determines the set of pseudocolumns to be added to the schema".

vcrfxia (commented):
Thanks @Sullivan-Patrick -- suggestions inline!

ksqldb-common/src/main/java/io/confluent/ksql/schema/ksql/LogicalSchema.java
Sullivan-Patrick: I miswrote this comment, this will actually use the passed `pseudoColumnVersion` parameter when the feature is released. How should I handle this given I can't just change the constant?

ksqldb-common/src/main/java/io/confluent/ksql/schema/ksql/LogicalSchema.java
Sullivan-Patrick: The passed `pseudoColumnVersion` doesn't actually see use if the first boolean is false, as if we are building without pseudocolumns, the version number is irrelevant. It seems overloading the method would be better here? Though there may be a fair bit of repeat code with the overload.

ksqldb-common/src/main/java/io/confluent/ksql/schema/ksql/LogicalSchema.java
Sullivan-Patrick: This would conflict with your above suggestion to make the ROWPARTITION/ROWOFFSET constant 0 for now, and both conditionals would always trigger. I could make a note to do this when the feature ships?

ksqldb-streams/src/main/java/io/confluent/ksql/execution/streams/StepSchemaResolver.java
Sullivan-Patrick: Just a remnant I didn't catch, removed.

ksqldb-common/src/main/java/io/confluent/ksql/schema/ksql/SystemColumns.java
```diff
@@ -15,39 +15,78 @@
...
-  public static final int CURRENT_PSEUDOCOLUMN_VERSION_NUMBER = 5;
 
-  private static final Set<ColumnName> PSEUDO_COLUMN_NAMES = ImmutableSet.of(
-      ROWTIME_NAME
-  );
+  public static final int ROWTIME_PSEUDOCOLUMN_VERSION = 0;
+  public static final int ROWPARTITION_ROWOFFSET_PSEUDOCOLUMN_VERSION = 1;
+
+  public static final int LEGACY_PSEUDOCOLUMN_VERSION_NUMBER = ROWTIME_PSEUDOCOLUMN_VERSION;
+  public static final int CURRENT_PSEUDOCOLUMN_VERSION_NUMBER =
```
vcrfxia, line 49: Setting this constant to 1 rather than 0 is a functional change, right? I thought the goal was to limit this PR to non-functional changes only.

ksqldb-common/src/main/java/io/confluent/ksql/schema/ksql/SystemColumns.java
```diff
@@ -15,39 +15,78 @@
...
+      ROWPARTITION_ROWOFFSET_PSEUDOCOLUMN_VERSION;
+
+  public static Set<ColumnName> getPseudoColumnsFromVersion(final int pseudoColumnVersion) {
+    if (!VERSIONED_PSEUDO_COLUMNS.containsKey(pseudoColumnVersion)) {
+      throw new KsqlException("Provided pseudoColumnVersion has no corresponding columns defined");
+    }
+    return VERSIONED_PSEUDO_COLUMNS.get(pseudoColumnVersion);
+  }
+
+  private static final Set<ColumnName> VERSION_ZERO_NAMES;
```
vcrfxia, line 59: We've added a lot of pseudocolumn-specific code at this point this is making this file (SystemColumns.java) fairly difficult to read. I think we should create a new util file specifically for pseudocolumns, e.g., `PseudoColumns.java`.

ksqldb-common/src/main/java/io/confluent/ksql/schema/ksql/SystemColumns.java
```diff
@@ -15,39 +15,78 @@
...
+  static {
+    VERSION_ZERO_NAMES = new HashSet<>();
+    VERSION_ZERO_NAMES.add(ROWTIME_NAME);
+
+    VERSION_ONE_NAMES = new HashSet<>(VERSION_ZERO_NAMES);
+    VERSION_ONE_NAMES.add(ROWPARTITION_NAME);
+    VERSION_ONE_NAMES.add(ROWOFFSET_NAME);
+  }
+
+  private static final Map<Integer, Set<ColumnName>> VERSIONED_PSEUDO_COLUMNS = ImmutableMap.of(
```
vcrfxia, line 71: nit: rename to `PSEUDO_COLUMN_NAMES_BY_VERSION` to be consistent with `SYSTEM_COLUMN_NAMES_BY_VERSION` below.

ksqldb-common/src/main/java/io/confluent/ksql/schema/ksql/SystemColumns.java
```diff
@@ -65,14 +104,25 @@ public static boolean isPseudoColumn(final ColumnName columnName) {
...
+    return getPseudoColumnsFromVersion(CURRENT_PSEUDOCOLUMN_VERSION_NUMBER);
   }
 
   public static boolean isSystemColumn(final ColumnName columnName) {
     return systemColumnNames().contains(columnName);
   }
 
   public static Set<ColumnName> systemColumnNames() {
-    return SYSTEM_COLUMN_NAMES;
+    return SYSTEM_COLUMN_NAMES_CURRENT;
```
vcrfxia, line 115: This method should call the overload. We can also remove `SYSTEM_COLUMN_NAMES_CURRENT` if we do this.

ksqldb-execution/src/main/java/io/confluent/ksql/execution/ddl/commands/CreateSourceCommand.java
```diff
@@ -88,9 +88,6 @@ public Boolean isOrReplace() {
   }
 
   private static void validate(final LogicalSchema schema, final boolean windowed) {
-    if (schema.valueContainsAny(SystemColumns.systemColumnNames())) {
```
vcrfxia, line 91: It's a bit confusing that this change is lumped into this PR. I thought the intention was to limit this PR to non-functional changes (adding new code that will be called in a subsequent PR), but this is changing existing production code. It'd be nice to move this into a separate PR so it's easier to reason about test coverage.

ksqldb-metastore/src/main/java/io/confluent/ksql/metastore/model/StructuredDataSource.java
```diff
@@ -74,10 +73,6 @@
     this.ksqlTopic = requireNonNull(ksqlTopic, "ksqlTopic");
     this.casTarget = casTarget;
 
-    if (schema.valueContainsAny(SystemColumns.systemColumnNames())) {
```
vcrfxia, line 77: Same as above -- it'd be nice to have this change in a separate PR in order to limit this one to non-functional changes only.

ksqldb-common/src/main/java/io/confluent/ksql/schema/ksql/LogicalSchema.java
```diff
@@ -123,10 +130,18 @@ public boolean valueContainsAny(final Set<ColumnName> names) {
...
    * @return the new schema.
    */
+  public LogicalSchema withPseudoAndKeyColsInValue(
+      final boolean windowed, final int pseudoColumnVersion) {
+    return rebuildWithPseudoAndKeyColsInValue(windowed, pseudoColumnVersion);
+  }
+
   public LogicalSchema withPseudoAndKeyColsInValue(final boolean windowed) {
-    return rebuild(true, windowed);
+    return rebuildWithPseudoAndKeyColsInValue(windowed, CURRENT_PSEUDOCOLUMN_VERSION_NUMBER);
```
vcrfxia, line 144: nit: let's have this method call the overload, rather than having both call `rebuildWithPseudoAndKeyColsInValue()`. IMO code is easier to read this way since it makes the relationship between the overloads more clear.

ksqldb-common/src/main/java/io/confluent/ksql/schema/ksql/LogicalSchema.java
```diff
@@ -260,19 +279,55 @@ private LogicalSchema rebuild(
       builder.add(Column.of(c.name(), c.type(), VALUE, valueIndex++));
     }
 
-    if (withPseudoAndKeyColsInValue) {
+    if (pseudoColumnVersion >= ROWTIME_PSEUDOCOLUMN_VERSION) {
       builder.add(Column.of(ROWTIME_NAME, ROWTIME_TYPE, VALUE, valueIndex++));
+    }
+
+    if (pseudoColumnVersion >= ROWPARTITION_ROWOFFSET_PSEUDOCOLUMN_VERSION) {
+      builder.add(Column.of(ROWOFFSET_NAME, ROWOFFSET_TYPE, VALUE, valueIndex++));
```
vcrfxia, line 287: This code adds ROWOFFSET then ROWPARTITION, while the code in SystemColumns.java returns pseudocolumn names with ROWPARTITION first instead. This is almost certainly going to lead to wacky bugs.

ksqldb-common/src/main/java/io/confluent/ksql/schema/ksql/LogicalSchema.java
```diff
@@ -260,19 +279,55 @@ private LogicalSchema rebuild(
...
+      builder.add(Column.of(ROWPARTITION_NAME, ROWPARTITION_TYPE, VALUE, valueIndex++));
+    }
+
+    for (final Column c : key) {
```
vcrfxia, line 291: Let's not duplicate this code between `rebuildWithPseudoAndKeyColsInValue()` and `rebuildWithoutPseudoAndKeyColsInValue()`. We can have them each call another helper method instead.

vcrfxia (commented):
Thanks @Sullivan-Patrick ! Comments inline.

ksqldb-common/src/main/java/io/confluent/ksql/schema/ksql/SystemColumns.java
Sullivan-Patrick: Correct, this is a mistake

ksqldb-common/src/main/java/io/confluent/ksql/schema/ksql/SystemColumns.java
Sullivan-Patrick: Sounds good, the only concern I have is that this will add interdependencies for functions like `isSystemColumn()` and `systemColumnNames()`, as Pseudo Columns are a subset of System Columns. I could make a set of package-private functions in `PseudoColumns.java` that allows SystemColumns to access this internal data as necessary. A problem with this though is I think that this internal will pretty much always be necessary, and the resulting code will be more complex  as `SystemColumns` will end up being more complex than it was before and we'll have another class, `PseudoColumns`, which is almost as complex as `SystemColumns` is now.

I'll try to prototype this change before our sync to see if my concerns are well-founded or not.

vcrfxia: Another option is to group all the pseudo-column specific code into a static class within SystemColumns.java, rather than creating a separate file. I think either is fine. The interdependencies in the case of having a separate file would be package-private rather than publicly accessible.

ksqldb-execution/src/main/java/io/confluent/ksql/execution/ddl/commands/CreateSourceCommand.java
Sullivan-Patrick: Since systemColumnNames was being called with it's current version and the current version was 1, this was causing issues and I wrongly anticipated we'd need to remove validation at this stage to allow QTTs to work. Fixed and removed these changes.

ksqldb-common/src/main/java/io/confluent/ksql/schema/ksql/LogicalSchema.java
Sullivan-Patrick: Ok, I'm standardizing (including naming) to always make this offset before partition.

ksqldb-common/src/main/java/io/confluent/ksql/schema/ksql/SystemColumns.java
```diff
@@ -64,15 +74,55 @@ public static boolean isPseudoColumn(final ColumnName columnName) {
...
+  }
+
+  private static Set<ColumnName> buildColumns(final int pseudoColumnVersion) {
+    return ImmutableSet.<ColumnName>builder()
+        .addAll(pseudoColumnNames(pseudoColumnVersion))
+        .addAll(WINDOW_BOUNDS_COLUMN_NAMES)
+        .build();
+  }
+
+  static class PseudoColumns {
```
vcrfxia, line 107: Make this class private and add an accessor method to expose looking up specific entries in the `PSEUDO_COLUMN_NAMES_BY_VERSION` map? Essentially, move the logic that's currently in `pseudoColumnNames()` into this class in order for it to be more contained. If this class owns the map, it'd be nice for it to own how the map is accessed too.

ksqldb-common/src/main/java/io/confluent/ksql/schema/ksql/SystemColumns.java
```diff
@@ -15,39 +15,49 @@
...
 import io.confluent.ksql.schema.ksql.types.SqlTypes;
+import io.confluent.ksql.util.KsqlException;
+import java.util.HashSet;
+import java.util.Map;
 import java.util.Set;
 
 public final class SystemColumns {
 
   public static final ColumnName ROWKEY_NAME = ColumnName.of("ROWKEY");
-
```
vcrfxia, line 27: nit: unnecessary whitespace change (I think it's slightly more readable with the extra line break since ROWKEY and ROWTIME are distinct system columns).

vcrfxia (commented):
Thanks @Sullivan-Patrick -- this is looking really great! A few last nits and then let's get this merged :)

ksqldb-streams/src/test/java/io/confluent/ksql/execution/streams/SourceBuilderTest.java
```diff
@@ -527,6 +531,20 @@ public void shouldAddRowTimeAndRowKeyColumnsToNonWindowedStream() {
...
+    final ValueTransformerWithKey<GenericKey, GenericRow, GenericRow> transformer =
+        getTransformerFromStreamSource(streamSource);
+
+    // When:
+    final GenericRow withTimestamp = transformer.transform(KEY, row);
+
+    // Then:
+    assertThat(withTimestamp, equalTo(GenericRow.genericRow("baz", 123, A_ROWTIME, A_ROWPARTITION, A_ROWOFFSET, A_KEY)));
+  }
+
```
Sullivan-Patrick, line 547: Should I add new tests like this? Basically every test in the file is checking the value columns of rows, so it's a debate between changing `shouldAddRowTimeAndRowKeyColumnsToNonWindowedStream()` to this and making the logic more discrete. I lean towards the former as I don't see a huge advantage to the latter outside of testing for the newer feature explicitly (but it shouldn't be horribly difficult to spot any bugs related directly to these fields failing anyway)

ksqldb-streams/src/test/java/io/confluent/ksql/execution/streams/SourceBuilderTest.java
```diff
@@ -825,11 +857,14 @@ private void givenWindowedSourceStream() {
         windowInfo,
         TIMESTAMP_COLUMN,
         SOURCE_SCHEMA,
-        OptionalInt.of(SystemColumns.CURRENT_PSEUDOCOLUMN_VERSION_NUMBER)
+        OptionalInt.of(pseudoColumnVersion)
     );
   }
+  private void givenWindowedSourceStream() {
+    givenWindowedSourceStream(SystemColumns.CURRENT_PSEUDOCOLUMN_VERSION_NUMBER);
+  }
```
Sullivan-Patrick, line 865: This overloading strategy might be overkill, too many extra lines/complexity for not that much extra functionality? I could just make the unit tests always use `ROWPARTITION_ROWOFFSET_PSEUDOCOLUMN_VERSION` until the feature is live, then swap to using current if we want active test coverage before the version becomes current. Or I could just manually test now by temporarily changing the current pseudocolumn version, then shipping the tests using the current version, which for the infrastructure PR would remain non-functional.

ksqldb-common/src/main/java/io/confluent/ksql/schema/ksql/SystemColumns.java
```diff
@@ -80,12 +82,13 @@ public static boolean isPseudoColumn(final ColumnName columnName) {
...
-    if (!PseudoColumns.PSEUDO_COLUMN_NAMES_BY_VERSION.containsKey(pseudoColumnVersion)) {
-      throw new KsqlException("Provided pseudoColumnVersion has no corresponding columns defined");
-    }
-    return PseudoColumns.PSEUDO_COLUMN_NAMES_BY_VERSION.get(pseudoColumnVersion);
+    return PseudoColumns.getPseudoColumnNamesByVersion(pseudoColumnVersion);
   }
 
+  @SuppressFBWarnings(
+      value = "MS_EXPOSE_REP",
+      justification = "PSEUDO_COLUMN_NAMES is ImmutableSet"
```
vcrfxia, line 90: Is the returned set of pseudocolumn names really an ImmutableSet? Looks like it's just a HashSet, unless I'm misreading the code.

ksqldb-streams/src/test/java/io/confluent/ksql/execution/streams/SourceBuilderTest.java
vcrfxia: We should have two versions of this test: one for `ROWTIME_PSEUDOCOLUMN_VERSION` and one for `ROWPARTITION_ROWOFFSET_PSEUDOCOLUMN_VERSION`, since the expected behavior is different and both are important. In other words, I think it's good that you added these two new tests. Be aware that once `CURRENT_PSEUDOCOLUMN_VERSION_NUMBER` is bumped to 1, the existing versions of these tests are going to fail. At that time they'll be redundant with the new test that explicitly tests `ROWPARTITION_ROWOFFSET_PSEUDOCOLUMN_VERSION`, so if you want you can update the existing version to explicitly use `ROWTIME_PSEUDOCOLUMN_VERSION` now, in order to save yourself some trouble later. (You won't be saving too much trouble later though since the rest of the test that use `CURRENT_PSEUDOCOLUMN_VERSION_NUMBER` will also need to be updated.)

We should also add new versions of the tests `shouldAddRowTimeAndTimeWindowedRowKeyColumnsToStream()` and `shouldAddRowTimeAndTimeWindowedRowKeyColumnsToTable()`.

ksqldb-streams/src/test/java/io/confluent/ksql/execution/streams/SourceBuilderTest.java
vcrfxia: I think the overloads are fine. It's not necessary to add overloads that are unused (such as the windowed versions) but if you follow my suggestion above to add new versions of windowed tests, then this comment becomes moot since all the new overloads will be used at that point.

vcrfxia (commented):
Thanks @Sullivan-Patrick . Did you get a chance to look into whether adding new unit tests to LogicalSchemaTest makes sense as well?

ksqldb-common/src/main/java/io/confluent/ksql/schema/ksql/SystemColumns.java
Sullivan-Patrick: It is, this is Matthias' code that I sort of copy-pasted where I thought it should go during a merge conflict. Should I just remove this as I am exposing a mutable set?

vcrfxia: You can try removing it but I think findbugs is going to fail (that's the source of the warning that's being suppressed). We should fix the issue so that findbugs does not fail without the suppression.

ksqldb-streams/src/test/java/io/confluent/ksql/execution/streams/SourceBuilderTest.java
Sullivan-Patrick: Sounds good. I think it makes sense to me to leave the tests which will become outdated all in a position that they'll fail when `CURRENT_PSEUDOCOLUMN_VERSION_NUMBER` is changed.

ksqldb-common/src/main/java/io/confluent/ksql/schema/ksql/LogicalSchema.java
```diff
@@ -156,13 +156,22 @@ public LogicalSchema withPseudoAndKeyColsInValue(final boolean windowed) {
     return withPseudoAndKeyColsInValue(windowed, CURRENT_PSEUDOCOLUMN_VERSION_NUMBER);
   }
 
+  /**
+   * Remove pseudo and key columns, according to the pseudocolumn version from the value schema.
```
vcrfxia, line 160:
```suggestion
   * Remove pseudo and key columns from the value schema, according to the pseudocolumn version from the value schema.
```

ksqldb-common/src/main/java/io/confluent/ksql/schema/ksql/LogicalSchema.java
```diff
@@ -156,13 +156,22 @@ public LogicalSchema withPseudoAndKeyColsInValue(final boolean windowed) {
...
+   * @param pseudoColumnVersion the version of pseudocolumns to evaluate against
+   * @return the new schema with the columns removed
+   */
+  public LogicalSchema withoutPseudoAndKeyColsInValue(final int pseudoColumnVersion) {
```
vcrfxia, line 164: Does this new method need to be public?

ksqldb-common/src/main/java/io/confluent/ksql/schema/ksql/LogicalSchema.java
```diff
@@ -322,14 +331,15 @@ private LogicalSchema rebuildWithoutPseudoAndKeyColsInValue() {
...
+  private int addNonPseudoAndKeyColsToValueSchema(
+      final ImmutableList.Builder<Column> builder, final int pseudoColumnVersion) {
     final Map<Namespace, List<Column>> byNamespace = byNamespace();
 
     final List<Column> value = byNamespace.get(VALUE);
 
     int addedColumns = 0;
     for (final Column c : value) {
-      if (SystemColumns.isSystemColumn(c.name())) {
+      if (SystemColumns.isSystemColumn(c.name(), pseudoColumnVersion)) {
```
vcrfxia, line 342: Nice catch!

ksqldb-common/src/main/java/io/confluent/ksql/schema/ksql/SystemColumns.java
```diff
@@ -77,8 +77,11 @@ public static boolean isWindowBound(final ColumnName columnName) {
     return WINDOW_BOUNDS_COLUMN_NAMES;
   }
 
+  public static boolean isPseudoColumn(final ColumnName columnName, final int pseudoColumnVersion) {
```
vcrfxia, line 80: Why is this needed? If it's just for testing purposes, we should mark it as such and it should also not be public.

ksqldb-common/src/test/java/io/confluent/ksql/schema/ksql/LogicalSchemaTest.java
```diff
@@ -239,6 +257,26 @@ public void shouldPreferValueOverMetaColumns() {
     );
   }
 
+  @Test
+  public void shouldPreferValueOverRowPartitionAndOffsetColumns() {
```
vcrfxia, line 261: nit: let's just merge this into the existing `shouldPreferValueOverMetaColumns()` test rather than adding a new one. This test doesn't depend on pseudocolumn version so the two tests are doing the same thing.

ksqldb-common/src/main/java/io/confluent/ksql/schema/ksql/LogicalSchema.java
Sullivan-Patrick: I'm using it in `LogicalSchemaTest` at least, and I assumed it'd likely need to be used down the line in some of the places it's 0-arg overload is used (as I thought we discussed in our last sync). What's your thought process on why it wouldn't be public?

ksqldb-common/src/main/java/io/confluent/ksql/schema/ksql/LogicalSchema.java
Sullivan-Patrick: The unit tests did the heavy lifting, but thanks :)

ksqldb-common/src/main/java/io/confluent/ksql/schema/ksql/SystemColumns.java
Sullivan-Patrick: As mentioned above, I haven't fully convinced myself of where it'll end up being called, though given it's 1-arg overload is called quite a few places, I expect we'll be using it in some of those, especially after CREATE OR REPLACE is added. I'm, like before, assuming you have understanding I don't have on this, if that understanding is complex we can go over it in our next sync.

ksqldb-common/src/main/java/io/confluent/ksql/schema/ksql/SystemColumns.java
```diff
@@ -129,12 +129,17 @@ public static boolean isSystemColumn(final ColumnName columnName) {
...
+      VERSION_ZERO_NAMES = versionZeroBuilder
+          .add(ROWTIME_NAME)
+          .build();;
+
+      ImmutableSet.Builder<ColumnName> versionOneBuilder = ImmutableSet.builder();
+      VERSION_ONE_NAMES = versionOneBuilder
+          .addAll(VERSION_ZERO_NAMES)
+          .add(ROWPARTITION_NAME)
+          .add(ROWOFFSET_NAME)
+          .build();
```
Sullivan-Patrick, line 142: I like this way of making the sets immutable vs making a `HashSet` then using the `ImmutableSet(Iterable)` constructor. This is scalable as we can always just create a versionTwoBuilder then add all from versionOne. Let me know if this works though

ksqldb-common/src/main/java/io/confluent/ksql/schema/ksql/SystemColumns.java
```diff
@@ -129,12 +129,17 @@ public static boolean isSystemColumn(final ColumnName columnName) {
...
+      VERSION_ZERO_NAMES = versionZeroBuilder
+          .add(ROWTIME_NAME)
+          .build();
+
+      ImmutableSet.Builder<ColumnName> versionOneBuilder = ImmutableSet.builder();
+      VERSION_ONE_NAMES = versionOneBuilder
+          .addAll(VERSION_ZERO_NAMES)
+          .add(ROWPARTITION_NAME)
+          .add(ROWOFFSET_NAME)
+          .build();
```
Sullivan-Patrick, line 142: I like this way of making the sets immutable vs making a HashSet then using the ImmutableSet(Iterable) constructor. This is scalable as we can always just create a versionTwoBuilder then add all from versionOne. Let me know if this works though

ksqldb-common/src/main/java/io/confluent/ksql/schema/ksql/LogicalSchema.java
vcrfxia: `LogicalSchemaTest` is in the same package so package-private is good enough for that use case. In general, it's easier to make things public later when it's determined that they need to be public than the other way around.

ksqldb-common/src/main/java/io/confluent/ksql/schema/ksql/SystemColumns.java
vcrfxia: As above, if the method doesn't need to be public (or if it doesn't even need to exist), then we should leave it out for now and introduce it later when it's actually needed, in order to avoid bloating the code.

ksqldb-common/src/main/java/io/confluent/ksql/schema/ksql/LogicalSchema.java
Sullivan-Patrick: That makes sense, thanks!

ksqldb-common/src/main/java/io/confluent/ksql/schema/ksql/LogicalSchema.java
```diff
@@ -236,48 +271,97 @@ public String toString(final FormatOptions formatOptions) {
...
+
+      builder.add(Column.of(c.name(), c.type(), VALUE, addedColumns++));
     }
+    return addedColumns;
+  }
 
-    return new LogicalSchema(builder.build());
+  /**
+   * Adds key columns to key schema of a provided builder
+   * @param builder the builder to be passed in. This builder will have key columns added to it's
```
vcrfxia, line 358:
```suggestion
   * @param builder the builder to be passed in. This builder will have key columns added to its
```

vcrfxia (commented):
Thanks @Sullivan-Patrick ! I think this is ready to be merged as soon as we get a passing build and remove the findbugs suppression we discussed earlier (see inline). Very exciting!

ksqldb-common/src/main/java/io/confluent/ksql/schema/ksql/LogicalSchema.java
Sullivan-Patrick: I've always been SURE it was the contraction and possessive, so I'm legitimately very happy you pointed this out. Maybe it's not a big improvement to the codebase, but it sure is to my English :)

vcrfxia (approved):
LGTM!

ksqldb-common/src/main/java/io/confluent/ksql/schema/ksql/LogicalSchema.java
```diff
@@ -236,48 +269,116 @@ public String toString(final FormatOptions formatOptions) {
...
+
+  /**
+   * Adds columns, except for key and pseudocolumns, to a provided builder's value schema
+   * and returns the number of added columns
+   * @param builder the builder to be passed in. Columns that don't qualify as key or pseudocolumns
+   *                will be added
+   * @return the number of columns added. This also functions as a "column index", to pass to
+   *                builder.add()
+   * */
+  private static ImmutableList<Column> nonPseudoAndKeyCols(
```
vcrfxia, line 346: Does this method need to be static? Rather than passing in a LogicalSchema to a static method, it'd be nice to have an instance method instead.

ksqldb-common/src/main/java/io/confluent/ksql/schema/ksql/LogicalSchema.java
```diff
@@ -236,48 +269,116 @@ public String toString(final FormatOptions formatOptions) {
...
+    return new LogicalSchema(builder.build());
+  }
+
+  /**
+   * Adds columns, except for key and pseudocolumns, to a provided builder's value schema
+   * and returns the number of added columns
+   * @param builder the builder to be passed in. Columns that don't qualify as key or pseudocolumns
+   *                will be added
+   * @return the number of columns added. This also functions as a "column index", to pass to
+   *                builder.add()
```
vcrfxia, line 344: Update this javadoc.

ksqldb-common/src/main/java/io/confluent/ksql/schema/ksql/LogicalSchema.java
vcrfxia: Let's also update the method name to clarify that the columns being returned are value columns, since that would not be my assumption otherwise as a reader.
```suggestion
  private static ImmutableList<Column> nonPseudoAndKeyColsAsValueCols(
```

ksqldb-common/src/main/java/io/confluent/ksql/schema/ksql/LogicalSchema.java
```diff
@@ -236,48 +269,116 @@ public String toString(final FormatOptions formatOptions) {
...
       }
+
+      builder.add(Column.of(c.name(), c.type(), VALUE, addedColumns++));
     }
+    return builder.build();
+  }
 
-    return new LogicalSchema(builder.build());
+  /**
+   * Adds key columns to key schema of a provided builder
```
vcrfxia, line 371: Update this javadoc.

vcrfxia (commented):
Thanks @Sullivan-Patrick . I'm excited to get this merged today! A few minor tweaks/suggestions inline.

ksqldb-common/src/main/java/io/confluent/ksql/schema/ksql/LogicalSchema.java
```diff
@@ -236,48 +269,115 @@ public String toString(final FormatOptions formatOptions) {
...
       }
+
+      builder.add(Column.of(c.name(), c.type(), VALUE, addedColumns++));
     }
+    return builder.build();
+  }
 
-    return new LogicalSchema(builder.build());
+  /**
+   * Adds key columns related to a namespace to an immutable list and returns the list
```
vcrfxia, line 370: What does "related to a namespace" mean?

ksqldb-common/src/main/java/io/confluent/ksql/schema/ksql/LogicalSchema.java
```diff
@@ -236,48 +269,115 @@ public String toString(final FormatOptions formatOptions) {
...
+   * Rebuilds schema without pseudocolumns or key columns
+   * @return the LogicalSchema created, with the corresponding pseudo and key columns excluded
+   */
+  private LogicalSchema rebuildWithoutPseudoAndKeyColsInValue(final int pseudoColumnVersion) {
+    final ImmutableList.Builder<Column> builder = ImmutableList.builder();
+
+    final List<Column> keyColumns = keyColumns(byNamespace());
+
+    final List<Column> nonPseudoAndKeyCols = nonPseudoAndKeyColsAsValueCols(
+        byNamespace(), pseudoColumnVersion);
```
vcrfxia, line 330: Let's just compute `byNamespace()` once rather than twice, else we're not getting the benefit of the more complicated method signatures for `keyColumns()` and `nonPseudoAndKeyColsAsValueCols()`.

vcrfxia (approved):
Thanks @Sullivan-Patrick -- two minor comments inline but I'm going to go ahead and merge this before they're addressed in order to unblock your PRs that are stacked on top of this one. Can you address them as part of https://github.com/confluentinc/ksql/pull/7990 instead? Thanks!

PR #7873, opened 2021-07-25
fix: make KsqlAvroSerializerTest work with Java 16

vvcephei:
Apparently some small things around casting behavior
and error messages have changed. This change makes the
test more generic so that it works regardless of the JVM.

### Testing done 
Ran the tests on my machine. If they also pass in Jenkins, it should be fine to merge.

### Reviewer checklist


agavra (approved):
LGTM, but for my own education - why are we running this with Jdk16? I'm pretty sure lots of things break at that language level

vvcephei:
Thanks, @agavra . It's not a matter of the compilation language level. I just happen to have the Java 16 runtime on my computer. The exception is fine, but the string that it returns in the exception varies across JVMs. Since the exception message is not in the JVM spec, it seems like we shouldn't write tests that rely on exact phrasing.

PR #7874, opened 2021-07-25
Refactor pull query execution

vvcephei:
### Description 

* Consolidate ad-hoc query analysis 
* Clarify the type of query in various "execute query" methods in the codebase
* Misc refactors to prepare for stream pull queries

### Testing done 

* Existing tests should be sufficient for a refactor

### Reviewer checklist

### DIFF
--- a/ksqldb-engine/src/main/java/io/confluent/ksql/KsqlExecutionContext.java
+++ b/ksqldb-engine/src/main/java/io/confluent/ksql/KsqlExecutionContext.java
@@ -15,6 +15,7 @@
 
 package io.confluent.ksql;
 
+import io.confluent.ksql.analyzer.ImmutableAnalysis;
 import io.confluent.ksql.engine.KsqlEngine;
 import io.confluent.ksql.engine.KsqlPlan;
 import io.confluent.ksql.execution.streams.RoutingOptions;
@@ -146,7 +147,7 @@ public interface KsqlExecutionContext {
    * Executes a query using the supplied service context.
    * @return the query metadata
    */
-  TransientQueryMetadata executeQuery(
+  TransientQueryMetadata executeTransientQuery(
       ServiceContext serviceContext,
       ConfiguredStatement<Query> statement,
       boolean excludeTombstones
@@ -163,7 +164,8 @@ public interface KsqlExecutionContext {
    *                         call PullQueryResult.start to start the query.
    * @return the rows that are the result of the query evaluation.
    */
-  PullQueryResult executePullQuery(
+  PullQueryResult executeTablePullQuery(
+      ImmutableAnalysis analysis,
       ServiceContext serviceContext,
       ConfiguredStatement<Query> statement,
       HARouting routing,
@@ -185,6 +187,7 @@ public interface KsqlExecutionContext {
    * @return A ScalablePushQueryMetadata object
    */
   ScalablePushQueryMetadata executeScalablePushQuery(
+      ImmutableAnalysis analysis,
       ServiceContext serviceContext,
       ConfiguredStatement<Query> statement,
       PushRouting pushRouting,
--- a/ksqldb-engine/src/main/java/io/confluent/ksql/analyzer/ImmutableAnalysis.java
+++ b/ksqldb-engine/src/main/java/io/confluent/ksql/analyzer/ImmutableAnalysis.java
@@ -21,6 +21,7 @@ import io.confluent.ksql.analyzer.Analysis.JoinInfo;
 import io.confluent.ksql.execution.expression.tree.ColumnReferenceExp;
 import io.confluent.ksql.execution.expression.tree.Expression;
 import io.confluent.ksql.execution.expression.tree.FunctionCall;
+import io.confluent.ksql.metastore.model.DataSource;
 import io.confluent.ksql.name.ColumnName;
 import io.confluent.ksql.parser.properties.with.CreateSourceAsProperties;
 import io.confluent.ksql.parser.tree.GroupBy;
@@ -74,4 +75,6 @@ public interface ImmutableAnalysis {
   AliasedDataSource getFrom();
 
   boolean getOrReplace();
+
+  DataSource getDataSource();
 }
--- a/ksqldb-engine/src/main/java/io/confluent/ksql/analyzer/ImmutableAnalysis.java
+++ b/ksqldb-engine/src/main/java/io/confluent/ksql/analyzer/ImmutableAnalysis.java
@@ -21,7 +21,6 @@ import io.confluent.ksql.analyzer.Analysis.JoinInfo;
 import io.confluent.ksql.execution.expression.tree.ColumnReferenceExp;
 import io.confluent.ksql.execution.expression.tree.Expression;
 import io.confluent.ksql.execution.expression.tree.FunctionCall;
-import io.confluent.ksql.metastore.model.DataSource;
 import io.confluent.ksql.name.ColumnName;
 import io.confluent.ksql.parser.properties.with.CreateSourceAsProperties;
 import io.confluent.ksql.parser.tree.GroupBy;
--- a/ksqldb-engine/src/main/java/io/confluent/ksql/analyzer/ImmutableAnalysis.java
+++ b/ksqldb-engine/src/main/java/io/confluent/ksql/analyzer/ImmutableAnalysis.java
@@ -75,6 +75,4 @@ public interface ImmutableAnalysis {
   AliasedDataSource getFrom();
 
   boolean getOrReplace();
-
-  DataSource getDataSource();
 }
--- a/ksqldb-engine/src/main/java/io/confluent/ksql/engine/EngineExecutor.java
+++ b/ksqldb-engine/src/main/java/io/confluent/ksql/engine/EngineExecutor.java
@@ -21,10 +21,7 @@ import com.google.common.base.Throwables;
 import com.google.common.collect.Iterables;
 import io.confluent.ksql.KsqlExecutionContext.ExecuteResult;
 import io.confluent.ksql.analyzer.ImmutableAnalysis;
-import io.confluent.ksql.analyzer.QueryAnalyzer;
-import io.confluent.ksql.analyzer.RewrittenAnalysis;
 import io.confluent.ksql.config.SessionConfig;
-import io.confluent.ksql.engine.QueryExecutionUtil.ColumnReferenceRewriter;
 import io.confluent.ksql.execution.ddl.commands.DdlCommand;
 import io.confluent.ksql.execution.plan.ExecutionStep;
 import io.confluent.ksql.execution.streams.RoutingOptions;
@@ -165,7 +162,8 @@ final class EngineExecutor {
    * @param pullQueryMetrics JMX metrics
    * @return the rows that are the result of evaluating the pull query
    */
-  PullQueryResult executePullQuery(
+  PullQueryResult executeTablePullQuery(
+      final ImmutableAnalysis analysis,
       final ConfiguredStatement<Query> statement,
       final HARouting routing,
       final RoutingOptions routingOptions,
@@ -183,12 +181,6 @@ final class EngineExecutor {
     RoutingNodeType routingNodeType = null;
 
     try {
-      final QueryAnalyzer queryAnalyzer = new QueryAnalyzer(engineContext.getMetaStore(),
-          NO_OUTPUT_TOPIC_PREFIX);
-      final ImmutableAnalysis analysis = new RewrittenAnalysis(
-          queryAnalyzer.analyze(statement.getStatement(), Optional.empty()),
-          new ColumnReferenceRewriter()::process
-      );
       // Do not set sessionConfig.getConfig to true! The copying is inefficient and slows down pull
       // query performance significantly.  Instead use QueryPlannerOptions which check overrides
       // deliberately.
@@ -263,6 +255,7 @@ final class EngineExecutor {
   }
 
   ScalablePushQueryMetadata executeScalablePushQuery(
+      final ImmutableAnalysis analysis,
       final ConfiguredStatement<Query> statement,
       final PushRouting pushRouting,
       final PushRoutingOptions pushRoutingOptions,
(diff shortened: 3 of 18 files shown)


ksqldb-engine/src/main/java/io/confluent/ksql/analyzer/ImmutableAnalysis.java
```diff
@@ -74,4 +75,6 @@
   AliasedDataSource getFrom();
 
   boolean getOrReplace();
+
+  DataSource getDataSource();
```
agavra, line 79:
```suggestion
  DataSource getFromDataSource();
```
in the spirit of bike-shedding, I feel like we shouldn't introduce this extra method since it makes the API a little confusing (when using autocomplete, I'll have to try to understand why we have `getFrom` and `getDataSource` and read the code to work out the difference - spoiler, there isn't one). But if we insist in saving the extra `.getDataSource()` call in the code base, let's try to make it clear that they're really the same method.

ksqldb-engine/src/main/java/io/confluent/ksql/engine/EngineExecutor.java
```diff
@@ -165,7 +162,8 @@ ExecuteResult execute(final KsqlPlan plan) {
    * @param pullQueryMetrics JMX metrics
    * @return the rows that are the result of evaluating the pull query
    */
-  PullQueryResult executePullQuery(
+  PullQueryResult executeTablePullQuery(
+      final ImmutableAnalysis analysis,
```
agavra, line 166: leaving notes to make sure I understand the refactor. this is the "meat", right? we no longer do the analysis at each part of the query but rather ahead-of-time, which allows us to "understand" whether it's a table pull/stream pull before choosing the corresponding method?

ksqldb-rest-app/src/main/java/io/confluent/ksql/api/impl/QueryEndpoint.java
```diff
@@ -133,19 +132,36 @@ public QueryPublisher createQueryPublisher(
...
+      return createScalablePushQueryPublisher(
+          analysis,
+          context,
+          serviceContext,
+          statement,
+          workerExecutor,
+          requestProperties
+      );
     } else {
       return createPushQueryPublisher(context, serviceContext, statement, workerExecutor);
```
agavra, line 159: do non-scalable push queries not benefit from having the analysis also done up front? I was thinking why not do the analysis outside of the if/else tree?

ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/resources/streaming/PullQueryPublisher.java
```diff
@@ -109,7 +110,11 @@ public synchronized void subscribe(final Subscriber<Collection<StreamedRow>> sub
 
     PullQueryResult result = null;
     try {
-      result = ksqlEngine.executePullQuery(
+      final ImmutableAnalysis analysis = ksqlEngine
```
agavra, line 113: why not pass the analysis into the constructor instead of doing it again here? it looks like you are creating it before creating the publisher in `QueryEndpoint`. Does it get lost before getting here? (this might be a different code path that I'm missing)

ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/resources/streaming/PushQueryPublisher.java
```diff
@@ -126,12 +127,14 @@ public synchronized void subscribe(final Flow.Subscriber<Collection<StreamedRow>
               query.getSessionConfig().getConfig(false),
               query.getSessionConfig().getOverrides());
 
+      final ImmutableAnalysis analysis = ksqlEngine
+          .analyzeQueryWithNoOutput(query.getStatement(), query.getStatementText());
```
agavra, line 131: ditto from above, why not pass the analysis in?

ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/resources/streaming/StreamedQueryResource.java
```diff
@@ -460,18 +488,17 @@ private EndpointResponse handleScalablePushQuery(
...
-
     final QueryStreamWriter queryStreamWriter = new QueryStreamWriter(
         query,
         disconnectCheckInterval.toMillis(),
         OBJECT_MAPPER,
         connectionClosedFuture
     );
 
-    log.info("Streaming query '{}'", statement.getStatementText());
+    QueryLogger.info("Streaming query", statement.getStatementText());
```
agavra, line 501: putting some thoughts down here: we probably want to have another PR that goes through and standardizes what we do for query logging across the transient query types (and also logs the messages meaningfully, e.g. there's no indication that this is a scalable push query from the query log). doesn't belong in this PR though, can you create a ticket to track this (tag `fix-it-week` and maybe also `p1`)

agavra (approved):
LGTM after rebasing with a green master

ksqldb-engine/src/main/java/io/confluent/ksql/KsqlExecutionContext.java
```diff
@@ -138,7 +139,7 @@
    * Executes a query using the supplied service context.
    * @return the query metadata
    */
-  TransientQueryMetadata executeQuery(
+  TransientQueryMetadata executeTransientQuery(
```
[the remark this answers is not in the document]
agavra: /cc @AlanConfluent but I don't think we're planning on entirely deprecating the traditional push queries in favor of scalable (e.g. you'll never be able to do a join/aggregate with a scalable push query IMO)

ksqldb-engine/src/main/java/io/confluent/ksql/engine/EngineExecutor.java
vvcephei: Yes!

ksqldb-rest-app/src/main/java/io/confluent/ksql/api/impl/QueryEndpoint.java
vvcephei: IIRC, the analysis currently needs to be a little different for transient and persistent queries, so I just didn't mess with it for now.

ksqldb-engine/src/main/java/io/confluent/ksql/KsqlExecutionContext.java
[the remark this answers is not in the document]
vvcephei: Thanks, all. Regardless of our actual plans, just a little feedback on the feedback: the reason I wound up making these changes in this PR was that I had a really hard time tracking down what all the different "query" methods were actually doing. I think "execute old push query" is better than just "execute query", but it still requires you to know the context: that we plan to deprecate this query path, for example. Even then, there are any number of code paths we may plan to deprecate in the future, so "old" doesn't tell you which old query path this is for.

ksqldb-engine/src/main/java/io/confluent/ksql/engine/KsqlEngine.java
```diff
@@ -369,6 +383,20 @@ public static boolean isExecutableStatement(final Statement statement) {
         || statement instanceof Query;
   }
 
+  public ImmutableAnalysis analyzeQueryWithNoOutput(final Query query, final String queryText) {
```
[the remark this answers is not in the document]
vvcephei: I meant that the query is of a type that has no output topic. I guess I could burn an extra five characters and clarify that in the method name ;)

ksqldb-engine/src/main/java/io/confluent/ksql/KsqlExecutionContext.java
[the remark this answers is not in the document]
agavra: +1 ^ thanks @vvcephei

ksqldb-engine/src/main/java/io/confluent/ksql/analyzer/ImmutableAnalysis.java
vvcephei: Ah, I didn't know about `getFrom()`. I also just noticed that this method is unused in this PR, so I should have saved it for the next one. Sorry about that.

I just tried it out on the follow-on branch, and it looks like (Law of Demeter violation notwithstanding) I can just use `.getFrom().getDataSource()` instead.

vvcephei: Hmm, now that I'm removing it, I notice that I'd written exactly that code in the implementation, so now I don't know _what_ I was thinking...

ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/resources/streaming/PullQueryPublisher.java
vvcephei: I don't think there's anywhere before this point that we do compute the analysis. That said, we can certainly push it up more and further consolidate calls to the analyzer. I'll give it a shot.

ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/resources/streaming/StreamedQueryResource.java
vvcephei: Sure, I agree we should audit the codebase and make sure we always use the logger for queries. I thought we already had, so I was surprised to find this here.

ksqldb-engine/src/main/java/io/confluent/ksql/engine/OrphanedTransientQueryCleaner.java
```diff
@@ -37,7 +36,6 @@
   private final QueryCleanupService cleanupService;
   private final KsqlConfig ksqlConfig;
 
-  @SuppressFBWarnings(value = "EI_EXPOSE_REP")
```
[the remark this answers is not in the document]
vvcephei: Oh, this was probably some kind of rebase mistake.

ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/resources/streaming/StreamedQueryResource.java
```diff
@@ -334,65 +308,117 @@ private EndpointResponse handleStatement(
...
-      });
-    });
-
-    final ConfiguredStatement<Query> configured = ConfiguredStatement
-        .of(statement, SessionConfig.of(ksqlConfig, configOverrides));
+      final Map<String, Object> configProperties,
+      final Context context,
+      final SlidingWindowRateLimiter pullBandRateLimiter) {
+
+    if (statement.getStatement().isPullQuery()) {
```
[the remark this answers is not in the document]
vvcephei: That would be nice, but differentiating those actually requires the analysis. I did consider creating an AnalyzedStatement similar to ConfiguredStatement to just keep everything together, but it was too big of a change for right now.

ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/resources/streaming/StreamedQueryResource.java
```diff
@@ -334,65 +308,117 @@ private EndpointResponse handleStatement(
...
+          securityContext.getServiceContext(),
+          statement,
+          configProperties,
+          request.getRequestProperties(),
+          connectionClosedFuture,
+          context
+      );
+    } else {
+      // log validated statements for query anonymization
+      QueryLogger.info("Transient query created", statement.getStatementText());
```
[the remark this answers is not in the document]
vvcephei: Hmm, I'll take a look.

ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/resources/streaming/PushQueryPublisher.java
vvcephei: Oh, so I implemented this advice, but now we actually are computing the analysis twice for _transient_ queries. The reason is that this "push query publisher" has the same constructors whether you're doing a transient or scalable push queries. I tried to make it so we could pass the analysis into the transient query execution, but it turns into a real nightmare without completely disentangling it from persistent queries.

vvcephei: Ok, I went ahead and partially walked this back, so for pull queries, we still do it at the top level for pull queries, but for push queries, we do the analysis as soon as we can distinguish between scalable vs. transient push queries.

ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/resources/streaming/StreamedQueryResource.java
vvcephei: Filed https://github.com/confluentinc/ksql/issues/7979

ksqldb-engine/src/main/java/io/confluent/ksql/analyzer/ImmutableAnalysis.java
agavra: 😆

ksqldb-rest-app/src/main/java/io/confluent/ksql/api/impl/QueryEndpoint.java
vvcephei: After digging back through the code, I can confirm that was the reason, and I made another attempt to unify the paths, but changing transient queries has too much splash damage.

ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/resources/streaming/StreamedQueryResource.java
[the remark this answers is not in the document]
vvcephei: Ok, I've fixed the other log line to say "scalable"

vvcephei:
Thanks for the reviews, @agavra and @guozhangwang ! I've addressed all your comments and made another pass over the diff myself. I think this is ready to merge if/when the tests pass.

PR #7880, opened 2021-07-26
Improves RQTT by using offsets to measure readiness

AlanConfluent:
### Description 
Rather than constantly comparing pull query outputs against expected results as a basis for checking correctness, this ensures that persistent queries have consumed input by looking at offsets and ensuring the consumer groups have reached the end.

This has a couple advantages:
- Partial results can't be matched against expected values, prematurely calling the output correct.  This can otherwise lead to flaky tests.
- With the current code, when output doesn't match, it retries over and over, which leads to bad error messages that usually take the form of timeouts rather than assertion errors.

### Testing done 
Ran tests.

### Reviewer checklist

### DIFF
--- a/ksqldb-functional-tests/src/test/java/io/confluent/ksql/test/rest/RestQueryTranslationTest.java
+++ b/ksqldb-functional-tests/src/test/java/io/confluent/ksql/test/rest/RestQueryTranslationTest.java
@@ -88,7 +88,8 @@ public class RestQueryTranslationTest {
           StreamsConfig.EXACTLY_ONCE_V2 // To stabilize tests
       )
       // Setting to anything lower will cause the tests to fail because we won't correctly commit
-      // transaction marker offsets.
+      // transaction marker offsets. This was previously set to 0 to presumably avoid flakiness,
+      // so we should keep an eye out for this.
       .withProperty(KsqlConfig.KSQL_STREAMS_PREFIX + StreamsConfig.COMMIT_INTERVAL_MS_CONFIG, 2000)
       .withProperty(KsqlConfig.KSQL_STREAMS_PREFIX + StreamsConfig.NUM_STREAM_THREADS_CONFIG, 1)
       .withProperty(KsqlConfig.SCHEMA_REGISTRY_URL_PROPERTY, "set")
--- a/ksqldb-functional-tests/src/test/java/io/confluent/ksql/test/rest/RestQueryTranslationTest.java
+++ b/ksqldb-functional-tests/src/test/java/io/confluent/ksql/test/rest/RestQueryTranslationTest.java
@@ -83,6 +83,13 @@ public class RestQueryTranslationTest {
           KsqlConfig.KSQL_STREAMS_PREFIX + StreamsConfig.STATE_DIR_CONFIG,
           TestUtils.tempDirectory().toAbsolutePath().toString()
       )
+      .withProperty(
+          KsqlConfig.KSQL_STREAMS_PREFIX + StreamsConfig.PROCESSING_GUARANTEE_CONFIG,
+          StreamsConfig.EXACTLY_ONCE_V2 // To stabilize tests
+      )
+      // Setting to anything lower will cause the tests to fail because we won't correctly commit
+      // transaction marker offsets.
+      .withProperty(KsqlConfig.KSQL_STREAMS_PREFIX + StreamsConfig.COMMIT_INTERVAL_MS_CONFIG, 2000)
       .withProperty(KsqlConfig.KSQL_STREAMS_PREFIX + StreamsConfig.NUM_STREAM_THREADS_CONFIG, 1)
       .withProperty(KsqlConfig.SCHEMA_REGISTRY_URL_PROPERTY, "set")
       .withProperty(KsqlConfig.KSQL_QUERY_PULL_TABLE_SCAN_ENABLED, true)
--- a/ksqldb-functional-tests/src/test/java/io/confluent/ksql/test/rest/RestQueryTranslationTest.java
+++ b/ksqldb-functional-tests/src/test/java/io/confluent/ksql/test/rest/RestQueryTranslationTest.java
@@ -83,10 +83,6 @@ public class RestQueryTranslationTest {
           KsqlConfig.KSQL_STREAMS_PREFIX + StreamsConfig.STATE_DIR_CONFIG,
           TestUtils.tempDirectory().toAbsolutePath().toString()
       )
-      .withProperty(
-          KsqlConfig.KSQL_STREAMS_PREFIX + StreamsConfig.PROCESSING_GUARANTEE_CONFIG,
-          StreamsConfig.EXACTLY_ONCE_V2 // To stabilize tests
-      )
       .withProperty(KsqlConfig.KSQL_STREAMS_PREFIX + StreamsConfig.NUM_STREAM_THREADS_CONFIG, 1)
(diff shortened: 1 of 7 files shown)


ksqldb-functional-tests/src/test/resources/rest-query-validation-tests/pull-queries-against-materialized-aggregates.json
```diff
@@ -841,7 +841,7 @@
         {"admin": {"@type": "currentStatus"}},
         {"query": [
            {"header":{"schema":"`ID` STRING KEY, `WINDOWSTART` BIGINT KEY, `WINDOWEND` BIGINT KEY, `COUNT` BIGINT"}},
-          {"row":{"columns":["10", 1582501512000, 1582501513000, 1]}}
+          {"row":{"columns":["10", 1582501512000, 1582501513000, 2]}}
```
cprasad1, line 844: 1) How was this test passing all this time? 
2) Do we treat records with the same key+timestamp combination as a single record or multiple records while materializing aggregates?

cprasad1 (approved):
Thanks @AlanConfluent! LGTM. Just had a couple of questions. Feel free to merge this once CI is green.

ksqldb-functional-tests/src/test/resources/rest-query-validation-tests/pull-queries-against-materialized-aggregates.json
AlanConfluent: I think it passed before because it would consistently match against partial results.  It may also have been flaky.

For a windowed table, the table key is effectively the key+timestamp, so in this case, we get count 2 because the inputs share both a key and the same timestamp.  If we changed the timestamp of one of them to be outside the window, we would have two outputs, each with count 1.

AlanConfluent:
> Why are we explicitly avoiding StreamsConfig.EXACTLY_ONCE_V2 here?

I got rid of it because the new method wasn't working with this config set...  When set, offsets for transaction markers weren't being read by streams and some tests would never complete.  This turns out to have been related to the value of `streams.commit.interval.ms` being 0 by default.  I've now set it to the server default and it appears to work fine, so I've added EOS back.

ksqldb-functional-tests/src/test/java/io/confluent/ksql/test/rest/RestQueryTranslationTest.java
```diff
@@ -87,6 +87,9 @@
           KsqlConfig.KSQL_STREAMS_PREFIX + StreamsConfig.PROCESSING_GUARANTEE_CONFIG,
           StreamsConfig.EXACTLY_ONCE_V2 // To stabilize tests
       )
+      // Setting to anything lower will cause the tests to fail because we won't correctly commit
+      // transaction marker offsets.
+      .withProperty(KsqlConfig.KSQL_STREAMS_PREFIX + StreamsConfig.COMMIT_INTERVAL_MS_CONFIG, 2000)
```
vvcephei, line 92: Thanks. You were concerned that this might introduce flakiness, and I think you're right. But we can also cross that bridge later.

Namely, if the actual results are sensitive to the relative timing of processing and committing, then when the test agent is slower or faster, we may get different results.

ksqldb-test-util/src/main/java/io/confluent/ksql/test/util/KafkaEmbedded.java
```diff
@@ -228,6 +239,7 @@ void waitForTopicsToBePresent(final String... topicNames) {
 
       final Supplier<Collection<String>> remaining = () -> {
         final Set<String> names = getTopicNames(adminClient);
+        System.out.println("NAMES ARE " + names);
```
vvcephei, line 242: Looks like this is vestigial.

ksqldb-functional-tests/src/test/resources/rest-query-validation-tests/pull-queries-against-materialized-aggregates.json
vvcephei: Hmm, I agree it was probably passing on an intermediate result. But how in the world did we write this expectation to begin with? It seems like it was obviously wrong. Oh, well, I agree it looks like it's right now.

ksqldb-functional-tests/src/test/resources/rest-query-validation-tests/push-queries.json
```diff
@@ -380,7 +380,7 @@
         "CREATE STREAM S1 (K INT KEY, ID bigint) WITH (kafka_topic='left_topic', value_format='JSON', WINDOW_TYPE='Tumbling', WINDOW_SIZE='2 SECONDS');",
         "CREATE STREAM S2 (K INT KEY, ID bigint) WITH (kafka_topic='right_topic', value_format='JSON', WINDOW_TYPE='Tumbling', WINDOW_SIZE='2 SECOND');",
         "CREATE STREAM OUTPUT as SELECT * FROM S1 JOIN S2 WITHIN 1 MINUTE ON S1.K = S2.K;",
-        "SELECT * FROM OUTPUT EMIT CHANGES LIMIT 2;"
+        "SELECT * FROM OUTPUT EMIT CHANGES LIMIT 3;"
```
vvcephei, line 383: It is so interesting that setting a longer commit interval gives us a new result in the middle...

vvcephei (approved):
Thanks, @AlanConfluent !

I really like this approach. I don't know if I'm too happy about the mystery of the commit interval, but it feels like this is an improvement over the status quo anyway.

ksqldb-functional-tests/src/test/resources/rest-query-validation-tests/push-queries.json
AlanConfluent: Heh, yep!

guozhangwang: I think this is likely to be due to the same issue of https://issues.apache.org/jira/browse/KAFKA-4609

Basically, the join operator is composed of a materialized node first, then a join node, on each side. The materialized node's state store could be cached, and the cache is being flushed independently. When the flush happens, they could be triggering the join twice on each side respectively and generate duplicates.

In the past with commit interval == 0, we effectively disabled any caching effect so we did not see this.

ksqldb-functional-tests/src/test/java/io/confluent/ksql/test/rest/RestTestExecutor.java
```diff
@@ -552,62 +561,139 @@ private static Object coerceExpectedKey(
...
-      final List<RqttResponse> actualResponses = ImmutableList
-          .of(RqttResponse.query(resp.getResponse()));
+      for (String queryApplicationId : queryApplicationIds) {
+        final List<TopicPartition> topicPartitions
+            = topicPartitionsByAppId.get(queryApplicationId);
+
+        Map<TopicPartition, Long> currentOffsets =
+            kafkaCluster.getConsumerGroupOffset(queryApplicationId);
+        Map<TopicPartition, Long> endOffsets = kafkaCluster.getEndOffsets(topicPartitions,
+            // Since we're doing At Least Once, we can do read uncommitted.
```
guozhangwang, line 649: This comment seems inconsistent with the actual logic?

guozhangwang (commented):
Sorry I'm late to this.

guozhangwang (commented):
@vvcephei @AlanConfluent I've found an potential issue why we would stop committing over the txn marker, and filed a PR.

Issue #7707, opened 2021-06-21
Issue adding fields to stream using JSON_SR


mikebin:
**Describe the bug**
Issue using `CREATE OR REPLACE STREAM` to add new field(s) with `JSON_SR`. 

**To Reproduce**
Version: ksqlDB 6.2.0/0.17.0

```sql
-- Initial stream with 1 field
create stream s (val string) with (kafka_topic='s', value_format='json_sr', partitions=1);

-- Modify stream to add field
create or replace stream s (val string, newfield string) with (kafka_topic='s', value_format='json_sr', partitions=1);

-- Insert will fail attempting to register new schema version
insert into s (val, newfield) values ('a', '1');
```

**Expected behavior**
New schema version will be registered containing `newfield`. 

**Actual behaviour**
```Failed to insert values into 'S'. Could not serialize value: [ 'a' | '1' ]. Error serializing message to topic: s. Failed to access Avro data from topic s : Schema being registered is incompatible with an earlier schema for subject "s-value"; error code: 409```

Note the error message above says `Avro`, even though this is JSON Schema. That should also be corrected.

Log from schema registry:

```
[2021-06-21 09:45:30,307] WARN Found incompatible change: Difference{jsonPath='#/properties/NEWFIELD', type=PROPERTY_ADDED_TO_OPEN_CONTENT_MODEL} (io.confluent.kafka.schemaregistry.json.JsonSchema:322)
```

**Additional context**

- The same steps above work fine for `AVRO` and `PROTOBUF`. This issue only occurs with `JSON_SR`
- Problem might be that ksqlDB needs to set `additionalProperties: false`.

**Comment by colinhicks** [comment]
Workaround: Manually register the initial schema with `additionalProperties: false`. This must be done before creating the stream.

```
curl -X POST -H "Content-Type: application/vnd.schemaregistry.v1+json" \
--data '{"schemaType":"JSON","schema":"{\"type\":\"object\",\"properties\":{\"VAL\":{\"connect.index\":0,\"oneOf\":[{\"type\":\"null\"},{\"type\":\"string\"}]}}, \"additionalProperties\":false}"}' \
http://localhost:8081/subjects/s-value/versions
```

```
ksql> create stream s (val string) with (kafka_topic='s', value_format='json_sr', partitions=1);

 Message
----------------
 Stream created
----------------
ksql> create or replace stream s (val string, newfield string) with (kafka_topic='s', value_format='json_sr', partitions=1);

 Message
----------------
[...]

Issue #6059, opened 2020-08-20
KsqlTester does not allow for DDL evolution


agavra:
The following test does not work because after we register the schema for A the first time around, for some reason the testing tool does not let us evolve it:
```sql
----------------------------------------------------------------------------------------------------
--@test: ddl evolution
----------------------------------------------------------------------------------------------------

CREATE STREAM a (id INT KEY, col1 INT) WITH (kafka_topic='a', value_format='JSON');
CREATE STREAM b AS SELECT * FROM a;

INSERT INTO a (id, col1) VALUES (1, 1);
CREATE OR REPLACE STREAM a (id INT KEY, col1 INT, col2 INT) WITH (kafka_topic='a', value_format='JSON');

INSERT INTO a (id, col1, col2) VALUES (1, 1, 1);
```

It fails with this rather cryptic error message:
```
io.confluent.ksql.test.KsqlTestException: Test failure for statement `INSERT INTO a (id, col1, col2) VALUES (1, 1, 1);` (Line: 64, Col: 1):
	Field count mismatch. topic: a, expected: 1, got: 2
	file:///Users/almog.gavra/dev/confluent/ksql/ksqldb-functional-tests/target/test-classes/sql-tests/query-upgrades/projections.sql:64

	at io.confluent.ksql.test.driver.KsqlTesterTest.handleExpectedException(KsqlTesterTest.java:324)
	at io.confluent.ksql.test.driver.KsqlTesterTest.test(KsqlTesterTest.java:165)
	at com.intellij.junit4.JUnit4IdeaTestRunner.startRunnerWithArgs(JUnit4IdeaTestRunner.java:69)
	at com.intellij.rt.junit.IdeaTestRunner$Repeater.startRunnerWithArgs(IdeaTestRunner.java:33)
	at com.intellij.rt.junit.JUnitStarter.prepareStreamsAndStart(JUnitStarter.java:220)
	at com.intellij.rt.junit.JUnitStarter.main(JUnitStarter.java:53)
Caused by: org.apache.kafka.common.errors.SerializationException: Field count mismatch. topic: a, expected: 1, got: 2
```

agavra:
I think the reason for this is because we get the `keySerde` and `valueSerde` for the input topics the first time we register it with the pipeline:
```java
        .map(ds -> new TopicInfo(ds.getKafkaTopicName(), keySerde(ds), valueSerde(ds)))
```

[...]

PR #7882, opened 2021-07-26
fix: ensure queries using JSON_SR can be upgraded

colinhicks:
### Description 
Addresses #7707. With the JsonSchema format, upgradeability is determined by the `additionalProperties` flag on the schema. Somewhat counterintuitively, this value must be `false` in order to support backward compatibility when new properties are added to an existing schema.

Prior to https://github.com/confluentinc/schema-registry/pull/1968, there was no way to configure the ConnectSchema-to-JsonSchema conversion such as to set the `additionalProperties` value. Because the default value of the property is `true`, the effect has been that all JsonSchema schemas created by ksqlDB are non-evolvable, meaning query upgrades are not possible when JSON_SR is used. See #7707 for the symptoms.

This PR configures the JsonSchema converter so that object schemas are registered with the `additionalProperties` flag set to false. In cases (presumably very unlikely) where users wish to prevent the schema from being evolved (and thus the query from being upgraded), the schema can be later updated to toggle the `additionalProperties` flag via the Schema Registry REST API or Confluent UI.

### Testing done 
Added unit test.

### Reviewer checklist


ksqldb-serde/src/test/java/io/confluent/ksql/serde/json/KsqlJsonSerializerTest.java
```diff
@@ -656,6 +656,27 @@ public void shouldNotIncludeBadValueInExceptionAsThatWouldBeASecurityIssue() {
...
+  @SuppressWarnings("unchecked")
+  @Test
+  public void jsonSchemaShouldSetAdditionalPropertiesToFalse() throws Exception {
+    // Given:
+    final Serializer serializer = givenSerializerForSchema(ORDER_SCHEMA, Struct.class);
+
+    // When:
+    serializer.serialize(SOME_TOPIC, new Struct(ORDER_SCHEMA));
+
+    if (useSchemas) {
```
agavra, line 668: nit: can we skip the entire test if `!useSchemas`?

ksqldb-serde/src/main/java/io/confluent/ksql/serde/json/KsqlJsonSerdeFactory.java
```diff
@@ -127,6 +128,7 @@ private static Converter getSchemaConverter(
         ksqlConfig.getString(KsqlConfig.SCHEMA_REGISTRY_URL_PROPERTY)
     );
     config.put(JsonConverterConfig.DECIMAL_FORMAT_CONFIG, DecimalFormat.NUMERIC.name());
+    config.put(JsonSchemaDataConfig.OBJECT_ADDITIONAL_PROPERTIES_CONFIG, false);
```
agavra, line 131: I haven't dug into the code, but what happens if we're using a schema that already has it explicitly set to `true`? will our schema override that? (e.g. does this make it impossible to `INSERT INTO` a source that uses JSON_SR with the value for true?) 

Also what does this mean for backwards compatibility with existing streams/tables that were created with the config being false. Can you test `CREATE STREAM AS SELECT ...` from a `JSON_SR` stream before upgrading, upgrade, and then see if we can insert and read from the stream?

agavra (commented):
thanks @colinhicks - this PR LGTM, but I'd like to see a little bit more testing around upgrading before/after this change

ksqldb-serde/src/main/java/io/confluent/ksql/serde/json/KsqlJsonSerdeFactory.java
colinhicks: Thanks for taking a look, @agavra!

Given `additionalProperties=true`, the schema can never be evolved, even when the updated schema sets `additionalProperties=false`.

> Also what does this mean for backwards compatibility with existing streams/tables that were created with the config being false. Can you test CREATE STREAM AS SELECT ... from a JSON_SR stream before upgrading, upgrade, and then see if we can insert and read from the stream?

With a stream created before this change, upgrade then insert will never work. Unfortunately ksqlDB cannot feasibly fix forward existing queries with schemas created before this change because of the way JsonSchema works. I found @rayokota's post useful for understanding the behavior: https://yokota.blog/2021/03/29/understanding-json-schema-compatibility/.

Reading from the upgraded stream would work because no schema is written on CREATE OR UPDATE, and, if I understand correctly, the existing reader schema is loaded based on the version in the record header.

agavra: I don't think my intention came across clearly - I meant does this affect existing queries when upgrading ksqlDB versions, without issuing a `CREATE OR REPLACE`?

colinhicks: Ah sorry for misinterpreting. My understanding is the change would not affect existing queries. We only register schemas on CREATE and INSERT, and this is the only point where the new property comes into play.

colinhicks:
Test failures are known flaky tests:
```
io.confluent.ksql.cli.CliTest.testPersistentSelectStar
io.confluent.ksql.cli.CliTest.shouldRunScriptOnRunScript
io.confluent.ksql.cli.CliTest.shouldRunScriptOnRunInteractively
io.confluent.ksql.cli.CliTest.shouldHandlePullQuery
```

colinhicks:
I was looking into the feasibility of adding a YATT-based test in order to assert that we can insert values after updating an existing JSON_SR-based stream with a new column. It looks like this won't be possible without first addressing https://github.com/confluentinc/ksql/issues/6059.

ksqldb-serde/src/main/java/io/confluent/ksql/serde/json/KsqlJsonSerdeFactory.java
agavra: We register schemas every time we serialize something (if the schema differs) - so it might still affect existing queries if we are serializing into a topic. What I'm worried about is that the converter is not configured in the physical plan, so we'd suddenly start using this serializer for old queries as well.

It might be better to hop on a quick zoom to see if one of us is misunderstanding the other 😆

colinhicks: Thanks for checking me on this.

I had missed the implications of upgrading ksqlDB itself upon existing JSON_SR-based queries. The change would indeed cause the converter to attempt to serialize an incompatible schema. Unfortunately, while the effect is essentially to add `additionalProperties: false` to the schema, this is not a backwards-compatible operation against an existing schema that does not set this property.

One recourse is to configure the converter based on a new property in KsqlConfig, documenting opting in to the `additionalProperties=false` as the best practice.

maxbol:
Any movement on this?

vvcephei:

PR #7883, opened 2021-07-26
docs: add changelog for 0.20

cprasad1:
### Description 
Adds changelog for 0.20 release

### Testing done 

### Reviewer checklist


CHANGELOG.md
```diff
@@ -1,5 +1,45 @@
...
+* add FROM_DAYS and update UNIX_DATE function ([#7742](https://github.com/confluentinc/ksql/pull/7742)) ([3c68710](https://github.com/confluentinc/ksql/commit/3c6871089e290be60787dda260ff8f8be0422483))
+* add PARSE_DATE and FORMAT_DATE functions ([#7733](https://github.com/confluentinc/ksql/pull/7733)) ([5a64ed7](https://github.com/confluentinc/ksql/commit/5a64ed7e762a06b9a2969def7c493b84451c0ac1))
+* add PARSE_TIME and FORMAT_TIME functions ([#7722](https://github.com/confluentinc/ksql/pull/7722)) ([9a381a8](https://github.com/confluentinc/ksql/commit/9a381a8ec8efc3e49632068d65de6e8b235d0527))
+* add TIMEADD and TIMESUB functions ([#7727](https://github.com/confluentinc/ksql/pull/7727)) ([75806a0](https://github.com/confluentinc/ksql/commit/75806a0233d838b3980a9b367cab42d7ab3b62d3))
+* call KStreams API grace() when GRACE PERIOD is used in joins ([#7678](https://github.com/confluentinc/ksql/pull/7678)) ([9bf9abf](https://github.com/confluentinc/ksql/commit/9bf9abf0d8093d2b53859abd5fe54a0d280a3cc7))
```
guozhangwang, line 13: I feel we can just merge this into the second item in bug fixes below.

CHANGELOG.md
```diff
@@ -1,5 +1,45 @@
...
+* enable schema inference for timestamp/time/date ([#7737](https://github.com/confluentinc/ksql/pull/7737)) ([35b1cad](https://github.com/confluentinc/ksql/commit/35b1cadc08faea68ca0b76e6c08ac664a5122c0b))
+* enable time unit functions for interpreter ([#7709](https://github.com/confluentinc/ksql/pull/7709)) ([a26a297](https://github.com/confluentinc/ksql/commit/a26a297af55ad46c3f46efae78f470a33d0df251))
+* fixed nondeterminism in UdfIndex ([#7719](https://github.com/confluentinc/ksql/pull/7719)) ([cd1a988](https://github.com/confluentinc/ksql/commit/cd1a9880aa39e0fb9b4e84d9f713235c8569b104))
+* Fixes slow streams shut down bug which made tests flaky ([#7717](https://github.com/confluentinc/ksql/pull/7717)) ([9b9b196](https://github.com/confluentinc/ksql/commit/9b9b1964a4123862af5c577229e26800feb933f5))
+* longer IT timeouts ([#7670](https://github.com/confluentinc/ksql/pull/7670)) ([47dd216](https://github.com/confluentinc/ksql/commit/47dd2169fd826bbffb2758be6489e7810b9e8553))
```
guozhangwang, line 33: nit: longer integration timeouts to avoid timing-dependent flakiness.

CHANGELOG.md
```diff
@@ -1,5 +1,45 @@
...
+* remove time/date component when casting timestamp to date/time ([#7724](https://github.com/confluentinc/ksql/pull/7724)) ([87cd3c7](https://github.com/confluentinc/ksql/commit/87cd3c727c7a27b717d89f51459a133cfa2d7cad))
+* return an error message on http2 /query-stream endpoint ([#7750](https://github.com/confluentinc/ksql/pull/7750)) ([3a4348b](https://github.com/confluentinc/ksql/commit/3a4348b9f394c9ac0dd3a3eeb64bd2e6b4663f69))
+* Fixes race condition exposing uninitialized query ([#7627](https://github.com/confluentinc/ksql/pull/7627)) ([98b1e3c](https://github.com/confluentinc/ksql/commit/98b1e3c220a6891321eecca0330adc8229eb089b))
+
+### BREAKING CHANGES
+
+* Existing queries that relied on vague implicit casting will not be started after an upgrade, and new queries that rely on vague implicit casting will be rejected. For example, foo(INT, INT) will not be able to resolve against two underlying function signatures of foo(BIGINT, BIGINT) and foo(DOUBLE, DOUBLE). Calling a function whose only parameter is variadic with an explicit null will also result in the call being rejected as vague.
```
agavra, line 41: @Sullivan-Patrick we should consider calling out that this was never _really_ supported because it would be in-deterministic across server restarts before this change

Sullivan-Patrick:
```suggestion
* Existing queries that relied on vague implicit casting will not be started after an upgrade, and new queries that rely on vague implicit casting will be rejected. For example, foo(INT, INT) will not be able to resolve against two underlying function signatures of foo(BIGINT, BIGINT) and foo(DOUBLE, DOUBLE). Calling a function whose only parameter is variadic with an explicit null will also result in the call being rejected as vague.

It's worth noting that queries which relied on this vague implicit casting were never truly supported, as they would have been nondeterministic. UDFs most likely to be adversely effected are ones which have multiple numerical overloads with repeated logic. For example, if you had two UDFs `foo(INT, INT)` and `foo(BIGINT, BIGINT)`, both which relied on logic which returns null if both input parameters are null, then the prior behavior would have been to nondeterministically route to one of the two functions, at which null would be returned either way (due to the duplicate logic). This change will break existing queries that relied on this nondeterministic routing.

```

This may be a bit wordy, but I figure I'd include a complete example here and we can remove as much as we want from this.

cprasad1: thanks @Sullivan-Patrick ! committing the suggestion

CHANGELOG.md
```diff
@@ -1,5 +1,48 @@
...
+* enable date/time for the query interpreter ([#7740](https://github.com/confluentinc/ksql/pull/7740)) ([79d14fb](https://github.com/confluentinc/ksql/commit/79d14fb60d7d784b3e099bcd59f102012c6a0d63))
+* implement casting for DATE/TIME ([#7708](https://github.com/confluentinc/ksql/pull/7708)) ([18cc030](https://github.com/confluentinc/ksql/commit/18cc030f453adcaa525d79e6bb773e35827a2ba7))
+* implement comparisons for TIME/DATE ([#7734](https://github.com/confluentinc/ksql/pull/7734)) ([78b9ae8](https://github.com/confluentinc/ksql/commit/78b9ae8666f0a74460aa9b1bed01d437930be2ce))
+* implement serde for DATE ([#7718](https://github.com/confluentinc/ksql/pull/7718)) ([a94f1f2](https://github.com/confluentinc/ksql/commit/a94f1f234607fb594fe4b617c83bba1b97a812d2))
+* pull query bandwidth based throttling ([#7738](https://github.com/confluentinc/ksql/pull/7738)) ([8f01ad9](https://github.com/confluentinc/ksql/commit/8f01ad986c6d9e61be3da9b011941fef45ab2ec7))
+* serde for TIME type ([#7664](https://github.com/confluentinc/ksql/pull/7664)) ([7537d87](https://github.com/confluentinc/ksql/commit/7537d87a77f37afc557d061fd2b816dfcdbfb12a))
```
jzaralim, line 20: Anything DATE/TIME related besides the new functions can be rolled into `add the DATE and TIME sql types`

jzaralim (commented):
sry didnt know how to add suggestions across multiple lines, but committing all these suggestions should combine them all

CHANGELOG.md
```diff
@@ -1,5 +1,41 @@
...
+* enable KAFKA-10847 bugfix when GRACE PERIOD is used on Joins ([#7739](https://github.com/confluentinc/ksql/pull/7739)) ([c1f5da6](https://github.com/confluentinc/ksql/commit/c1f5da627541b0500a4c4c6ebc439bcce899b426))
+([#7678](https://github.com/confluentinc/ksql/pull/7678)) ([9bf9abf](https://github.com/confluentinc/ksql/commit/9bf9abf0d8093d2b53859abd5fe54a0d280a3cc7))
+* enable schema inference for timestamp/time/date ([#7737](https://github.com/confluentinc/ksql/pull/7737)) ([35b1cad](https://github.com/confluentinc/ksql/commit/35b1cadc08faea68ca0b76e6c08ac664a5122c0b))
+* enable time unit functions for interpreter ([#7709](https://github.com/confluentinc/ksql/pull/7709)) ([a26a297](https://github.com/confluentinc/ksql/commit/a26a297af55ad46c3f46efae78f470a33d0df251))
+* fixed nondeterminism in UdfIndex ([#7719](https://github.com/confluentinc/ksql/pull/7719)) ([cd1a988](https://github.com/confluentinc/ksql/commit/cd1a9880aa39e0fb9b4e84d9f713235c8569b104))
+* Fixes slow streams shut down bug which made tests flaky ([#7717](https://github.com/confluentinc/ksql/pull/7717)) ([9b9b196](https://github.com/confluentinc/ksql/commit/9b9b1964a4123862af5c577229e26800feb933f5))
```
agavra, line 25:
```suggestion
```
nit: test only change doesn't need to be here

CHANGELOG.md
```diff
@@ -1,5 +1,41 @@
...
+* longer integration timeouts to avoid timing-dependent flakiness ([#7670](https://github.com/confluentinc/ksql/pull/7670)) ([47dd216](https://github.com/confluentinc/ksql/commit/47dd2169fd826bbffb2758be6489e7810b9e8553))
```
agavra, line 26:
```suggestion
```
nit: test only change doesn't need to be here

agavra (approved):
thanks @cprasad1

PR #7886, opened 2021-07-27
chore: improve error message for pull query failure

### LINKED ISSUE #7772: Improve the error message/logs for failed pull queries (All nodes are dead)
I was looking at some logs and for a cluster that has no lag and only a single host (which was alive at the time) I was seeing the following exception:
```
Unable to execute pull query. All nodes are dead or exceed max allowed lag.
```
The error message doesn't help me understand what happened, which partition failed or if there's some other issue. We should print more debugging information.

agavra:
### Description 

fixes #7772 by adding an improved error log message. to accomplish this, this PR does not return only the nodes that need to be queried for each partition when we `locate` in the `HARouting` class; instead, it will return the list of all hosts it _tried_ to consider and any reason why they should not be selected, if they were not selected.

### Testing done 

Updated existing unit tests

### Reviewer checklist


ksqldb-engine/src/main/java/io/confluent/ksql/physical/pull/HARouting.java
```diff
@@ -106,22 +107,39 @@ public void close() {
...
+            KsqlPartitionLocation::getPartition,
+            loc -> loc.getNodes().stream().map(KsqlNode::getHost).collect(Collectors.toList())));
+
+    if (!emptyPartitions.isEmpty()) {
+      final MaterializationException materializationException = new MaterializationException(
+          "Unable to execute pull query. "
+              + emptyPartitions.entrySet()
+              .stream()
+              .map(kv -> String.format(
+                  "Partition %s failed to find valid host. Hosts scanned: %s",
```
guozhangwang, line 129: nit: should we add an extra space after `Hosts scanned: %s ` to separate multiple partitions?

ksqldb-engine/src/main/java/io/confluent/ksql/physical/pull/HARouting.java
```diff
@@ -106,22 +107,39 @@ public void close() {
...
             routingFilterFactory
     );
 
-    final boolean anyPartitionsEmpty = locations.stream()
-        .anyMatch(location -> location.getNodes().isEmpty());
-    if (anyPartitionsEmpty) {
-      LOG.debug("Unable to execute pull query: {}. All nodes are dead or exceed max allowed lag.",
-                statement.getStatementText());
-      throw new MaterializationException("Unable to execute pull query. "
-          + "All nodes are dead or exceed max allowed lag.");
```
guozhangwang, line 122: Just making sure: the message `or exceed max allowed lag` is actually incorrect and hence we exclude it in the PR, right?

ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/LivenessFilter.java
```diff
@@ -40,14 +40,22 @@ public LivenessFilter(final Optional<HeartbeatAgent> heartbeatAgent) {
...
 
     if (!heartbeatAgent.isPresent()) {
-      return true;
+      return Host.include(host);
     }
 
     final Map<KsqlHostInfo, HostStatus> allHostsStatus = heartbeatAgent.get().getHostsStatus();
     final HostStatus status = allHostsStatus.get(host);
-    return status == null ? true : allHostsStatus.get(host).isHostAlive();
+    if (status == null) {
```
guozhangwang, line 51: For my own education: why we treat `null` as okay alive?

ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/LivenessFilter.java
```diff
@@ -40,14 +40,22 @@ public LivenessFilter(final Optional<HeartbeatAgent> heartbeatAgent) {
...
+      return Host.include(host);
+    }
+
+    if (status.isHostAlive()) {
+      return Host.include(host);
+    } else {
+      return Host.exclude(host, "Host is not alive as of " + status.getLastStatusUpdateMs());
```
guozhangwang, line 58: nit: as of time

ksqldb-streams/src/main/java/io/confluent/ksql/execution/streams/RoutingFilter.java
```diff
@@ -31,7 +35,57 @@
...
+
+  @Immutable
+  class Host {
+    private final KsqlHostInfo sourceInfo;
+    private final boolean selected;
+    private final String reason;
+
+    private Host(final KsqlHostInfo sourceInfo, final boolean selected, final String reason) {
+      this.sourceInfo = Objects.requireNonNull(sourceInfo, "sourceInfo");
+      this. selected = selected;
```
guozhangwang, line 48: typo? :)

ksqldb-streams/src/main/java/io/confluent/ksql/execution/streams/RoutingFilter.java
```diff
@@ -31,7 +35,57 @@
...
    * @return If the host should be included for consideration.
    */
-  boolean filter(KsqlHostInfo hostInfo);
+  Host filter(KsqlHostInfo hostInfo);
+
+  @Immutable
+  class Host {
+    private final KsqlHostInfo sourceInfo;
+    private final boolean selected;
+    private final String reason;
```
guozhangwang, line 44: nit: reasonNotSelected? since if it is selected it's always ""

ksqldb-streams/src/main/java/io/confluent/ksql/execution/streams/materialization/ks/KsLocator.java
```diff
@@ -444,6 +463,26 @@ public String toString() {
...
+        return true;
+      }
+      if (o == null || getClass() != o.getClass()) {
+        return false;
+      }
+
+      final PartitionLocation that = (PartitionLocation) o;
+      return partition == that.partition
+          && Objects.equals(keys, that.keys)
+          && Objects.equals(nodes, that.nodes);
```
guozhangwang, line 479: Since nodes are stored in a list, do we guarantee the ordering is always the same (otherwise they would be treated as not equal)?

ksqldb-engine/src/main/java/io/confluent/ksql/physical/pull/HARouting.java
agavra: it could've been correct, but it also may not have been. the new error message will say whether the hosts were dead, whether the lag was exceeded, or something else entirely happened (no hosts were found at all for that partition)

ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/LivenessFilter.java
agavra: I think there was a discussion around this early on and we decided that this can happen right when a node starts up and it was better to assume alive than assume dead. I can try to dig up the history on that discussion, I believe it was on Vicky's original PR

ksqldb-engine/src/main/java/io/confluent/ksql/physical/pull/HARouting.java
agavra: I added the space the to `joining` collector so that the separator will be `", "` instead

ksqldb-streams/src/main/java/io/confluent/ksql/execution/streams/RoutingFilter.java
agavra: hehe wonder how that happened

ksqldb-streams/src/main/java/io/confluent/ksql/execution/streams/materialization/ks/KsLocator.java
agavra: good catch, we do not guarantee ordering - this was used only for testing so it wasn't caught. I'll change it to a set comparison

guozhangwang (approved):
LGTM!

PR #7888, opened 2021-07-27
WIP

mjsax:
### Description 

### Testing done 

### Reviewer checklist


ksqldb-package/pom.xml
```diff
@@ -154,6 +154,50 @@
...
...                            <argument>-x ,confluent-log4j-${confluent-log4j.version}.jar,connect-api-${kafka.version}.jar,connect-json-${kafka.version}.jar,connect-runtime-${kafka.version}.jar,connect-transforms-${kafka.version}.jar,kafka-avro-serializer-${avro.version}.jar,kafka-clients-${kafka.version}.jar,kafka-clients-${kafka.version}-test.jar,kafka-connect-avro-converter-${kafka.version}.jar,kafka-connect-avro-data-${kafka.version}.jar,kafka-connect-json-schema-converter-${kafka.version}.jar,kafka-connect-protobuf-converter-${kafka.version}.jar,kafka-json-schema-provider-${kafka.version}.jar,kafka-json-schema-serializer-${kafka.version}.jar,kafka-json-serializer-${kafka.version}.jar,kafka-log4j-appender-${kafka.version}.jar,kafka-protobuf-provider-${kafka.version}.jar,kafka-protobuf-serializer-${kafka.version}.jar,kafka-protobuf-types-${kafka.version}.jar,kafka-schema-registry-client-${kafka.version}.jar,kafka-schema-serializer-${kafka.version}.jar,kafka-streams-${kafka.version}.jar,kafka-streams-test-utils-${kafka.version}.jar,kafka-tools-${kafka.version}.jar,licenses-${io.confluent.license-file-generator.version}.jar,confluent-licensing-new-${ce.kafka.version}.jar</argument>
```
mjsax, line 176: Please double check if this list of excludes makes sense?

One issue atm is, that not all are excluded though because `${kafka.version}` is a "normal" version, while some dependencies are includes via nano-versions -- but I am not sure how I could put the picked nano-version in here... The hope was, that on an actual release, we won't pull in any nano-versions, but concrete versions and thus the issue would go away (but I am not sure).

ksqldb-package/src/assembly/package.xml
```diff
@@ -35,8 +35,6 @@
                 <include>README*</include>
                 <include>LICENSE*</include>
                 <include>NOTICE*</include>
```
mjsax, line 37: Btw: There is no `NOTICE` file... Should we have one?

PR #7889, opened 2021-07-27
feat: added missing tostring method

tolgadur:
Jira: [KCI-778](https://confluentinc.atlassian.net/browse/KCI-778)

### Description 
Anonymized queries were printed out wrongly due to a missing `toString` method in `QueryLogger`. This PR adds this missing method. 

### Testing done 
- Made sure tests still run and manually checked if logged queries are printed out correctly now.,
 
### Reviewer checklist


vvcephei (commented):
Thanks, @tolgadur ! 

It seems like the intent here was also to log the message and the guid. Did you want to include those in the output as well?

swist (approved):

PR #7904, opened 2021-07-27
feat: allow expressions on left table columns in FK-joins

mjsax:
### Description 
Currently, FK-join require to use a plain column reference for the left join expression. We want to lift this restriction and allow for actual expression in the left join expression.

### Testing done 
Unit tests and new QTT tests.

### Reviewer checklist


ksqldb-engine/src/main/java/io/confluent/ksql/planner/LogicalPlanner.java
```diff
@@ -755,25 +757,34 @@ private static void verifyStreamTableJoin(final JoinInfo joinInfo, final PlanNod
...
-    final SqlType fkColumnType =
-        leftNode.getSchema().findColumn(fkColumnReference.getColumnName()).get().type();
+    final VisitParentExpressionVisitor<Optional<Expression>, Context<Void>> unqualifiedRewritter =
+        new VisitParentExpressionVisitor<Optional<Expression>, Context<Void>>(Optional.empty()) {
+          @Override
+          public Optional<Expression> visitQualifiedColumnReference(
+              final QualifiedColumnReferenceExp node,
+              final Context<Void> ctx
+          ) {
+            return Optional.of(new UnqualifiedColumnReferenceExp(node.getColumnName()));
```
mjsax, line 773: The `ExpressionEvaluator` cannot handle fully qualified column references, thus we rewrite the given expression accordingly.

ksqldb-engine/src/main/java/io/confluent/ksql/planner/LogicalPlanner.java
```diff
@@ -786,7 +797,24 @@ private static void verifyStreamTableJoin(final JoinInfo joinInfo, final PlanNod
...
+
+    final VisitParentExpressionVisitor<Optional<Expression>, Context<Void>> aliasRewritter =
+        new VisitParentExpressionVisitor<Optional<Expression>, Context<Void>>(Optional.empty()) {
+          @Override
+          public Optional<Expression> visitQualifiedColumnReference(
+              final QualifiedColumnReferenceExp node,
+              final Context<Void> ctx
+          ) {
+            return Optional.of(new UnqualifiedColumnReferenceExp(
+                ColumnNames.generatedJoinColumnAlias(node.getQualifier(), node.getColumnName())
```
mjsax, line 809: We need to rewrite the given expression again, to prepare it for physical code generation later.

ksqldb-execution/src/test/java/io/confluent/ksql/execution/plan/ForeignKeyTableTableJoinTest.java
```diff
@@ -54,26 +55,26 @@
   public void shouldImplementEquals() {
     new EqualsTester()
         .addEqualityGroup(
-            new ForeignKeyTableTableJoin<>(props1, INNER, JOIN_COLUMN_NAME, formats1, left1, right1),
-            new ForeignKeyTableTableJoin<>(props1, INNER, JOIN_COLUMN_NAME, formats1, left1, right1)
+            new ForeignKeyTableTableJoin<>(props1, INNER, JOIN_COLUMN_NAME, formats1, left1, right1, null),
```
mjsax, line 58: Open TODO...

ksqldb-streams/src/test/java/io/confluent/ksql/execution/streams/ForeignKeyJoinParamsFactoryTest.java
```diff
@@ -48,7 +51,7 @@ public void shouldBuildCorrectKeyedSchema() {
 
     // When:
     final ForeignKeyJoinParams<String> joinParams =
-        ForeignKeyJoinParamsFactory.create(leftJoinColumnName, LEFT_SCHEMA, RIGHT_SCHEMA);
+        ForeignKeyJoinParamsFactory.create(Optional.of(leftJoinColumnName), Optional.empty(), LEFT_SCHEMA, RIGHT_SCHEMA, null);
```
mjsax, line 54: TODO...

ksqldb-streams/src/test/java/io/confluent/ksql/execution/streams/ForeignKeyTableTableJoinBuilderTest.java
```diff
@@ -136,26 +136,26 @@ public void init() {
     );
   }
 
-  @Test
```
mjsax, line 139: TODO..

ksqldb-engine/src/main/java/io/confluent/ksql/planner/LogicalPlanner.java
```diff
@@ -872,21 +883,33 @@ private static boolean isInnerNode(final PlanNode node) {
...
         leftSource.getDataSource().getSchema().key().stream()
             .map(c -> new QualifiedColumnReferenceExp(alias, c.name()))
             .collect(Collectors.toList());
 
-    final ColumnName foreignKeyColumnName = columnRef.maybeQualifier().isPresent()
-        ? ColumnNames.generatedJoinColumnAlias(
-            columnRef.maybeQualifier().get(),
-            columnRef.getColumnName())
-        : columnRef.getColumnName();
+    final VisitParentExpressionVisitor<Optional<Expression>, Context<Void>> aliasRewritter =
```
mjsax, line 895: Similar to the code we replace, we need to rewrite the expression with the alias prefix...

mjsax:
Needed to update the generated schemas... I _think_ it's safe. Do you agree @vcrfxia ?

ksqldb-engine/src/main/java/io/confluent/ksql/planner/LogicalPlanner.java
```diff
@@ -872,21 +883,33 @@ private static boolean isInnerNode(final PlanNode node) {
...
+        new VisitParentExpressionVisitor<Optional<Expression>, Context<Void>>(Optional.empty()) {
+          @Override
+          public Optional<Expression> visitQualifiedColumnReference(
+              final QualifiedColumnReferenceExp node,
+              final Context<Void> ctx
+          ) {
+            return Optional.of(new UnqualifiedColumnReferenceExp(
+                ColumnNames.generatedJoinColumnAlias(node.getQualifier(), node.getColumnName())
```
vcrfxia, line 903: I'm not following the logic with qualifiers here. The old code only sometimes generated the join column alias (if a qualifier is present) but this new code seems to always generate the join column alias (before rewriting the alias to be unqualified). Is this a change in behavior?

ksqldb-execution/src/main/java/io/confluent/ksql/execution/plan/ForeignKeyTableTableJoin.java
```diff
@@ -42,24 +45,37 @@ public ForeignKeyTableTableJoin(
...
+      final ExecutionStep<KTableHolder<KRightT>> rightSource,
+      @JsonProperty(value = "leftJoinExpression")
+      final Optional<Expression> leftJoinExpression
   ) {
     this.properties = requireNonNull(props, "props");
     this.joinType = requireNonNull(joinType, "joinType");
     if (joinType == JoinType.OUTER) {
       throw new IllegalArgumentException("OUTER join not supported.");
     }
-    this.leftJoinColumnName = requireNonNull(leftJoinColumnName, "leftJoinColumnName");
```
vcrfxia, line 59: We can still require non-null for `Optional` types. It'd be nice to add this back, and also add it for `leftJoinExpression` below.

ksqldb-streams/src/main/java/io/confluent/ksql/execution/streams/ExecutionStepFactory.java
```diff
@@ -334,10 +334,11 @@ public static WindowedTableSource tableSourceWindowed(
       foreignKeyTableTableJoin(
           final QueryContext.Stacker stacker,
           final JoinType joinType,
-          final ColumnName leftJoinColumnName,
+          final Optional<ColumnName> leftJoinColumnName,
           final Formats formats,
           final ExecutionStep<KTableHolder<KLeftT>> left,
-          final ExecutionStep<KTableHolder<KRightT>> right
+          final ExecutionStep<KTableHolder<KRightT>> right,
+          final Optional<Expression> expression
```
vcrfxia, line 341: nit: rename to `leftJoinExpression` to be more informative?

ksqldb-streams/src/main/java/io/confluent/ksql/execution/streams/ForeignKeyTableTableJoinBuilder.java
```diff
@@ -40,8 +45,32 @@ private ForeignKeyTableTableJoinBuilder() {
...
+      );
+
+      expressionEvaluator = Optional.of(
+          codeGenRunner.buildCodeGenFromParseTree(
+              leftJoinExpression.get(),
+              "Left Join Expression"
+          )
+      );
+    } else {
+      expressionEvaluator = Optional.empty();
```
vcrfxia, line 69: Instead of having two separate cases here, can we combine them by always passing an expression evaluator in order to simplify the code? The expression evaluator for the single column case would be simple, just extracting the appropriate column.

ksqldb-streams/src/main/java/io/confluent/ksql/execution/streams/KsqlKeyExtractor.java
```diff
@@ -19,26 +19,60 @@
...
+  private final Optional<Supplier<String>> errorMessage;
+
   KsqlKeyExtractor(final int leftJoinColumnIndex) {
     checkArgument(
         leftJoinColumnIndex >= 0,
         "leftJoinColumnIndex negative: " + leftJoinColumnIndex
     );
 
     this.leftJoinColumnIndex = leftJoinColumnIndex;
+    expressionEvaluator = Optional.empty();
```
vcrfxia, line 44: As above, it'd be nice if we could simplify this code by always using an expression evaluator.

ksqldb-streams/src/main/java/io/confluent/ksql/execution/streams/ForeignKeyJoinParamsFactory.java
```diff
@@ -27,18 +29,28 @@ private ForeignKeyJoinParamsFactory() {
...
       throw new IllegalStateException("rightSchema must have single column key");
     }
-    return new ForeignKeyJoinParams<>(
-        createKeyExtractor(leftSchema, leftJoinColumnName),
-        new KsqlValueJoiner(leftSchema.value().size(), rightSchema.value().size(), 0),
-        createSchema(leftSchema, rightSchema)
-    );
+    if (leftJoinColumnName.isPresent()) {
+      return new ForeignKeyJoinParams<>(
+          createKeyExtractor(leftSchema, leftJoinColumnName.get()),
```
vcrfxia, line 43: nit: I think it's slightly more readable to create the foreign key extractor in an if-statement and then have a single return statement for building the ForeignKeyJoinParams. Also less code duplication.

vcrfxia (commented):
Thanks @mjsax -- LGTM with some questions/comments inline. It'd also be nice to add a test for mismatched types involving expressions (ensure that the error message is sensible).

ksqldb-engine/src/main/java/io/confluent/ksql/planner/LogicalPlanner.java
mjsax: I think it not a behavior change, because we use the visiter pattern and this method is only executed for qualified column references (its name is `visitQualifiedColumnReference`). Thus unqualified column references should be untouched, as before.

ksqldb-streams/src/main/java/io/confluent/ksql/execution/streams/ForeignKeyTableTableJoinBuilder.java
mjsax: Good idea!

ksqldb-streams/src/test/java/io/confluent/ksql/execution/streams/ForeignKeyTableTableJoinBuilderTest.java
```diff
@@ -145,10 +155,11 @@ public void shouldDoLeftJoinOnNonKey() {
     // When:
     final KTableHolder<Struct> result = join.build(planBuilder, planInfo);
 
+    final ExpressionEvaluator e = mock(ExpressionEvaluator.class);
     // Then:
     verify(leftKTable).leftJoin(
         same(rightKTable),
-        eq(new KsqlKeyExtractor<>(1)),
+        any(KsqlKeyExtractor.class),
```
mjsax, line 162: @vcrfxia After the refactoring to always use `ExpressionEvaluator`, it pretty hard to verify that `KsqlKeyExtractor` is created correctly now... I just simplified this test to make it pass for now. But this change kinda make the test useless... Thoughts?

vcrfxia: If we really want to test this, we can use an ArgumentCaptor to capture the KsqlKeyExtractor, add a getter for the expression evaluator (visible for testing only), get the expression from the expression evaluator, and verify that the expression is a column reference expression for the appropriate column. I don't think this is too much effort if we wrap it in a helper method.

As you pointed out in the first version of this PR, test coverage for this (and some of the other unit tests) is lacking at the moment since we only test single columns for the left join expression, and not any non-column expressions. If we want to extend test coverage to include non-column expressions, I think we'll have to do something similar anyway. (I'm happy deferring this to a follow-up PR if you prefer since I know you've got some time constraints.)

ksqldb-streams/src/main/java/io/confluent/ksql/execution/streams/KsqlKeyExtractor.java
```diff
@@ -15,30 +15,43 @@
...
+  private final Supplier<String> errorMessage;
 
-  KsqlKeyExtractor(final int leftJoinColumnIndex) {
-    checkArgument(
-        leftJoinColumnIndex >= 0,
-        "leftJoinColumnIndex negative: " + leftJoinColumnIndex
+  KsqlKeyExtractor(final ExpressionEvaluator expressionEvaluator,
+                   final ProcessingLogger processingLogger) {
+    this.expressionEvaluator = requireNonNull(expressionEvaluator);
+    this.processingLogger = processingLogger;
```
vcrfxia, line 37: Any reason we don't require this to be non-null?

vcrfxia (approved):
Thanks @mjsax ! The new QTT coverage is great. Love the refactor too 🎉

vcrfxia:
> Needed to update the generated schemas... I think it's safe. Do you agree @vcrfxia ?

Assuming we're talking about the execution step/plan schemas, I agree this is safe. The updates reflect the changes to the ForeignKeyTableTableJoin execution step.

ksqldb-streams/src/main/java/io/confluent/ksql/execution/streams/KsqlKeyExtractor.java
mjsax: We only "forward" it to `expressionEvaluator.evaluate()` but don't use it ourselves, so I though we don't need this (make testing simpler :))

ksqldb-engine/src/main/java/io/confluent/ksql/planner/LogicalPlanner.java
guozhangwang: I had a sync with @mjsax offline about this specifically, and our thoughts are that since the codeGenRunner is used for physical plan generation as well at which stage there's no qualified column refs since we would just prefix the names as `table/stream__column`, is that right @vcrfxia ??

If yes, just as a general note here I'm feeling that we should use a separate class than codeGenRunner for traversals in the logical/physical plan generation respectively, than reusing it, since for logical planning all we really need are return type / param types etc, while nevertheless today we always run the same full-fledged traversal multiple times at different stages. Is my understanding reasonable?

guozhangwang (approved):
I do not have any major comments, just a question about the tech debt we have for using the same codeGenRunner for logical / physical planning.

ksqldb-streams/src/test/java/io/confluent/ksql/execution/streams/ForeignKeyTableTableJoinBuilderTest.java
mjsax: Oh nice! #gold

ksqldb-engine/src/main/java/io/confluent/ksql/planner/plan/JoinNode.java
```diff
@@ -643,10 +645,10 @@ static JoinKey syntheticColumn() {
     }
 
     static JoinKey foreignKeyColumn(
-        final ColumnName foreignKeyColumn,
+        final Expression foreignKeyExpression,
```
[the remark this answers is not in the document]
mjsax: Let's got with `foreignKey()`

ksqldb-engine/src/main/java/io/confluent/ksql/planner/plan/JoinNode.java
```diff
@@ -806,17 +808,34 @@ public JoinKey rewriteWith(
...
   private static final class ForeignJoinKey implements JoinKey {
-    private final ColumnName foreignKeyColumn;
+    private final Optional<ColumnName> foreignKeyColumn;
+    private final Optional<Expression> foreignKeyExpression;
     private final ImmutableList<? extends ColumnReferenceExp> leftSourceKeyColumns;
 
     static JoinKey of(final ColumnName foreignKeyColumn,
                       final Collection<QualifiedColumnReferenceExp> leftSourceKeyColumns) {
-      return new ForeignJoinKey(foreignKeyColumn, leftSourceKeyColumns);
+      return new ForeignJoinKey(
```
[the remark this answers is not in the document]
mjsax: Good catch!

ksqldb-functional-tests/src/test/resources/historical_plans/fk-join_-_Should_support_inner_join_with_left_key-column_expression/7.1.0_1627438328984/plan.json
```diff
@@ -0,0 +1,249 @@
...
+        "valueFormat" : {
+          "format" : "JSON"
+        },
+        "keyFeatures" : [ "UNWRAP_SINGLES" ]
+      },
+      "orReplace" : false
+    }
+  }, {
+    "@type" : "ksqlPlanV1",
+    "statementText" : "CREATE TABLE OUTPUT AS SELECT\n  LEFT_TABLE.L_ID L_ID,\n  LEFT_TABLE.L_ID_2_FOREIGN_KEY L_ID_2_FOREIGN_KEY,\n  RIGHT_TABLE.R_ID R_ID,\n  LEFT_TABLE.NAME NAME,\n  RIGHT_TABLE.F1 F1\nFROM LEFT_TABLE LEFT_TABLE\nINNER JOIN RIGHT_TABLE RIGHT_TABLE ON ((LEFT_TABLE.L_ID_2_FOREIGN_KEY = RIGHT_TABLE.R_ID))\nEMIT CHANGES",
```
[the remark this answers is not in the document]
mjsax: Those are historic plans that are auto-generated...

mjsax:
Merging this PR as-is, and do a follow up PR to address the open comments.

mjsax:
Follow up PR: https://github.com/confluentinc/ksql/pull/7916

PR #7918, opened 2021-07-29
feat: add bytes functionality to REPLACE

Sullivan-Patrick:
### Description 
Adding bytes to REPLACE udf

### Testing done 
QTTs and Unit testing covering the usage of bytes within the REPLACE function

### Reviewer checklist

### DIFF
--- a/ksqldb-engine/src/main/java/io/confluent/ksql/function/udf/string/Replace.java
+++ b/ksqldb-engine/src/main/java/io/confluent/ksql/function/udf/string/Replace.java
@@ -19,14 +19,17 @@ import io.confluent.ksql.function.FunctionCategory;
 import io.confluent.ksql.function.udf.Udf;
 import io.confluent.ksql.function.udf.UdfDescription;
 import io.confluent.ksql.function.udf.UdfParameter;
+import io.confluent.ksql.util.BytesUtils;
 import io.confluent.ksql.util.KsqlConstants;
+import java.nio.ByteBuffer;
+import java.nio.charset.StandardCharsets;
 
 @SuppressWarnings("unused") // Invoked via reflection.
 @UdfDescription(
     name = "replace",
     category = FunctionCategory.STRING,
     author = KsqlConstants.CONFLUENT_AUTHOR,
-    description = "Replaces all occurances of a substring in a string with a new substring."
+    description = "Replaces all occurences of a substring in a string with a new substring."
 )
 public class Replace {
   @Udf(description = "Returns a new string with all occurences of oldStr in str with newStr")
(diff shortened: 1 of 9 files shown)


Sullivan-Patrick:
A couple things I'd like feedback on:

1. I'd like to edit the docs to reflect functionality with bytes. However, I'm not sure how to add this to the description "Replace all instances of a substring in a string with a new string."

- I could just add "also works with BYTES" or something, but that doesn't sound great
- I could change the description to something like "Replace all instances of a substring `or byte subarray` in a string `or byte array` with a new string `or byte subarray`." This feels a bit like using "she or he" instead of "they": grammatically correct maybe, but sounds clunky.
- I could just not change it. A user could probably understand by virtue of the function accepting BYTES that they can act on the `byte[]` the same way.

2. Implementing replace in a subarray seems a little nontrivial. As far as I know, there isn't any simple answer like an Apache Commons function that does this. If there is, please link me and I will just use that. Based off the assumption there is not, here are my possible solutions:

- Convert the underlying byte[] of the input buffers to strings, use Java’s String.replace() , convert back
- Use Apple’s foundationDB code here https://apple.github.io/foundationdb/javadoc/com/apple/foundationdb/tuple/ByteArrayUtil.html. Since it’s a whole DB I’d have to manually copy-paste the code. I don’t know what the implications for fair use are here? They’re under the Apache license
- Use my own suspicious code. Not implementing KMP like the other two. As far as I can tell my solution is O(N) but has a lot of overhead
Any input is appreciated here- regarding which option to take, or if there is a better 4th option (maybe some library I missed)

ksqldb-engine/src/main/java/io/confluent/ksql/function/udf/string/Replace.java
```diff
@@ -45,4 +48,46 @@ public String replace(
...
+
+    if (source == null || toReplace == null || replaceWith == null) {
+      return null;
+    }
+
+    //if replaceWith.length == toReplace.length, DON'T create a new array
+    //else, create an arraylist. Create another arraylist2 and a value matchingBytesNum.
+    //populate arraylist2 as long as the next byte matches the next byte in our toReplace[matchingBytesNum]
+    //if matchingBytesNum reaches the length of toReplace, we have a matching subbytes and append
+    //to our 1st arraylist the replaceWith bytes. If they diverge, just dump the original arraylist
```
Sullivan-Patrick, line 72: This was how I was going to solve the problem with a custom solution if we decide to go down that route (please view my comment in the discussion section to understand my options). I'll leave this here. AFAICT the solution is O(N), but it's probably slower and more difficult to read than others in the end.

PR #7921, opened 2021-07-29
fix: increase default command response timeout

nateab:
### Description 
The CI machine appears to be overloaded and is being slow, so it seems to be timing out when reading from the command topic. This results in several flaky tests such as `shouldRunScriptOnRunScript`,`shouldRunScriptOnRunInteractively`, `shouldHandlePullQuery` which all fail due to `Timeout while waiting for command topic consumer to process command topic`.

### Testing done 
Ran locally several hundred times but unable to easily reproduce bug as the MacBook is too fast. 

### Reviewer checklist

### DIFF
--- a/ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/KsqlRestConfig.java
+++ b/ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/KsqlRestConfig.java
@@ -540,7 +540,7 @@ public class KsqlRestConfig extends AbstractConfig {
         ).define(
             DISTRIBUTED_COMMAND_RESPONSE_TIMEOUT_MS_CONFIG,
             Type.LONG,
-            5000L,
+            15000L,
             Importance.LOW,
             DISTRIBUTED_COMMAND_RESPONSE_TIMEOUT_MS_DOC
         ).define(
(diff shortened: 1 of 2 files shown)


ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/KsqlRestConfig.java
```diff
@@ -198,10 +198,10 @@
           + "JSON; this helps proactively determine if the connection has been terminated in "
           + "order to avoid keeping the created streams job alive longer than necessary";
 
-  static final String DISTRIBUTED_COMMAND_RESPONSE_TIMEOUT_MS_CONFIG =
+  public static final String DISTRIBUTED_COMMAND_RESPONSE_TIMEOUT_MS_CONFIG =
       KSQL_CONFIG_PREFIX + "server.command.response.timeout.ms";
 
-  private static final String DISTRIBUTED_COMMAND_RESPONSE_TIMEOUT_MS_DOC =
+  protected static final String DISTRIBUTED_COMMAND_RESPONSE_TIMEOUT_MS_DOC =
```
agavra, line 204: any reason for this change?

agavra (approved):
Thanks!

ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/KsqlRestConfig.java
nateab: I saw that a lot of the other configs that were public had protected docs like `ACCESS_CONTROL_ALLOW_HEADERS,` so I just followed that

agavra: 👍

AlanConfluent (approved):

PR #7924, opened 2021-07-30
chore: remove \r\n chars from Base64 encoded strings

spena:
### Description 
When encoding bytes with to_bytes(..., 'base64'), I noticed that `\r\n` characters were added to the string. This PR sure not to add those \r\n characters.

### Testing done 

### Reviewer checklist

### DIFF
--- a/ksqldb-common/src/main/java/io/confluent/ksql/util/BytesUtils.java
+++ b/ksqldb-common/src/main/java/io/confluent/ksql/util/BytesUtils.java
@@ -26,10 +26,7 @@ import java.util.Map;
 import java.util.function.Function;
 
 public final class BytesUtils {
-  private static final Base64.Encoder BASE64_ENCODER = Base64.getMimeEncoder();
-  private static final Base64.Decoder BASE64_DECODER = Base64.getMimeDecoder();
-
-  enum Encoding {
+  public enum Encoding {
     HEX,
     UTF8,
     ASCII,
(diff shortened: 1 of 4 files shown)


ksqldb-common/src/main/java/io/confluent/ksql/util/BytesUtils.java
```diff
@@ -26,10 +26,7 @@
 import java.util.function.Function;
 
 public final class BytesUtils {
-  private static final Base64.Encoder BASE64_ENCODER = Base64.getMimeEncoder();
-  private static final Base64.Decoder BASE64_DECODER = Base64.getMimeDecoder();
```
colinhicks, line 30: Do any usages of these utils rely on or assume the MIME-based encoding. IIUC, the line breaks are used in those schemes in order to make the bytes more easily human readable

spena: We don't rely on the MIME-based encoding. As you say, this encoding is useful for easily human readable, but it didn't apply to the CLI.

Using MIME
```
+-----------------------------------------------------------------------------------------------------------------------------+-----------------------------------------------------------------------------------------------------------------------------+
|ID                                                                                                                           |KSQL_COL_0                                                                                                                   |
+-----------------------------------------------------------------------------------------------------------------------------+-----------------------------------------------------------------------------------------------------------------------------+
|1                                                                                                                            |VGhpcyBpcyBhIHN0cmluZyBsaW5lIHRoYXQgd2lsbCBiZSBjb252ZXJ0ZWQgdG8gYmFzZTY0IGFu
ZCB0aGVuIHRvIGJ5dGVzLiBJIHdvbiAkMTAwMCBkb2xsYXJ|
|                                                                                                                             |zLiBJZiBJIG11bHRpcGx5IDEwMDAq
MTAsIHRoZW4gSSBnZXQgJDEwLDAwMC4gU2VlIGFib3ZlIF4uIE5vIG0=                                      |
```

No MIME
```
+-----------------------------------------------------------------------------------------------------------------------------+-----------------------------------------------------------------------------------------------------------------------------+
|ID                                                                                                                           |KSQL_COL_0                                                                                                                   |
+-----------------------------------------------------------------------------------------------------------------------------+-----------------------------------------------------------------------------------------------------------------------------+
|1                                                                                                                            |54686973206973206120737472696E67206C696E6520746861742077696C6C20626520636F6E76657274656420746F2062617365363420616E64207468656|
|                                                                                                                             |E20746F2062797465732E204920776F6E20243130303020646F6C6C6172732E2049662049206D756C7469706C7920313030302A31302C207468656E204920|
|                                                                                                                             |676574202431302C3030302E205365652061626F7665205E2E204E6F206D
```

The CLI already formats the lines to fit the into the column box. Maybe this would be useful for how users would like to see it in the topic, but I don't think users would like the \r\n embedded in the base64 strings. If they use KSQL, then KSQL formats it correctly. If they use a different system, then I think they expect only base64 characters.

colinhicks (approved):
LGTM and seems safe since the `BytesUtils` class is new and was never released, but I defer to the eng team

jzaralim (approved):

PR #7932, opened 2021-08-02
fix: dont use default package for version.properties

agavra:
### Description 

We deployed our ksqlDB cluster alongside a `dogstatd` client which led to exactly https://github.com/DataDog/java-dogstatsd-client/issues/146. This PR fixes it by moving our `version.properties` file to a non-default resource package so that the loading doesn't rely on the classpath ordering (see: https://stackoverflow.com/a/1497843/2258040)

### Testing done 

Ran the `AppInfo.main` to make sure that the file is loaded properly

### Reviewer checklist


ksqldb-common/pom.xml
```diff
@@ -157,6 +157,11 @@
                 <filtering>true</filtering>
             </resource>
 
+            <resource>
```
rodesai, line 160: did you need to add this to ensure the properties file is included in the jar?

rodesai (approved):
LGTM! really nice find!

ksqldb-common/pom.xml
agavra: without this it wouldn't resolve the `${project.version}` when I built the module

PR #7936, opened 2021-08-03
refactor: support BYTES on QTT for Protobuf

spena:
### Description 
Allow QTT run tests with Protobuf bytes. The tests in JSON require bytes to be encoded using base64 encodind and also define the protobuf schema as part of the QTT test. QTT will identify there's a schema with the bytes type and attempt to decode the base64 string to a ByteBuffer. This ByteBuffer is written to the input topics.

Later, when the test is executed, the QTT gets the ByteBuffer from the output topics, and then is converted to base64 in order to compare it with the base64 string used in the QTT output topic.

### Testing done 

### Reviewer checklist

### DIFF
--- a/ksqldb-functional-tests/src/test/resources/query-validation-tests/bytes.json
+++ b/ksqldb-functional-tests/src/test/resources/query-validation-tests/bytes.json
@@ -28,6 +28,31 @@
       ]
     },
     {
+      "name": "PROTOBUF in/out",
+      "statements": [
+        "CREATE STREAM TEST WITH (kafka_topic='test', value_format='PROTOBUF');",
+        "CREATE STREAM TEST2 AS SELECT * FROM TEST;"
+      ],
+      "topics": [
+        {
+          "name": "test",
+          "valueFormat": "PROTOBUF",
+          "valueSchema": "syntax = \"proto3\"; message ConfluentDefault1 {bytes b = 1;}"
+        }
+      ],
+      "inputs": [
+        {"topic": "test", "value": {"b": "dmFyaWF0aW9ucw=="}}
+      ],
+      "outputs": [
+        {"topic": "TEST2", "value": {"B": "dmFyaWF0aW9ucw=="}}
+      ],
+      "post": {
+        "sources": [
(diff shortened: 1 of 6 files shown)


jzaralim (approved):

ksqldb-functional-tests/src/test/resources/query-validation-tests/bytes.json
```diff
@@ -27,6 +27,31 @@
         {"topic": "TEST2", "value": {"B": "dmFyaWF0aW9ucw=="}}
       ]
     },
+    {
```
jzaralim, line 30: This AVRO test works if you also add the Avro schema translator to RecordNode:

```
    {
      "name": "AVRO in/out",
      "statements": [
        "CREATE STREAM TEST (ID STRING KEY, b BYTES) WITH (kafka_topic='test', value_format='AVRO');",
        "CREATE STREAM TEST2 AS SELECT * FROM TEST;"
      ],
      "topics": [
        {
          "name" : "test",
          "valueSchema" : {
            "type" : "record",
            "name" : "KsqlDataSourceSchema",
            "namespace" : "io.confluent.ksql.avro_schemas",
            "fields" : [ {
              "name" : "ID",
              "type" : [ "null", "string" ],
              "default" : null
            }, {
              "name" : "B",
              "type" : [ "null", "bytes" ],
              "default" : null
            }],
            "connect.name" : "io.confluent.ksql.avro_schemas.KsqlDataSourceSchema"
          },
          "valueFormat" : "AVRO"
        }
      ],
      "inputs": [
        {"topic": "test", "value": {"B": "dmFyaWF0aW9ucw=="}}
      ],
      "outputs": [
        {"topic": "TEST2", "value": {"B": "dmFyaWF0aW9ucw=="}}
      ]
    },
```

spena: Thanks. I'll put this in another PR. When added the AVRO translator with an AVRO schema, the `bytes.json` worked, but I got other issues with the `avro.json` 'cause I required that the translator uses WRAP_SINGLES, which later caused other issues in other cases. So I ended up just submitting the code for Protobuf.

PR #7939, opened 2021-08-03
fix: migrate integration tests to use the working directory for state

vvcephei:
Previously, we would default to `/tmp`, but are seeing occasional test failures when the `/tmp` directory gets cleared during the tests by the OS.

### Testing done 
I ran the tests on my machine and made sure that no `junit` files are created in `/tmp`, and that I do see them in the working directory. I also verified that the files are deleted when tests finish, and also if I terminate the test process.

### Reviewer checklist

### DIFF
--- a/ksqldb-engine/src/main/java/io/confluent/ksql/engine/QueryCleanupService.java
+++ b/ksqldb-engine/src/main/java/io/confluent/ksql/engine/QueryCleanupService.java
@@ -115,11 +115,14 @@ public class QueryCleanupService extends AbstractExecutionThreadService {
     public void run() {
       try {
         final Path pathName = Paths.get(stateDir + "/" + appId);
-        FileUtils.deleteDirectory(new File(String.valueOf(pathName.normalize())));
-        LOG.warn("Deleted local state store for non-existing query {}. "
-            + "This is not expected and was likely due to a "
-            + "race condition when the query was dropped before.",
-            appId);
+        final File directory = new File(String.valueOf(pathName.normalize()));
+        if (directory.exists()) {
+          FileUtils.deleteDirectory(directory);
+          LOG.warn("Deleted local state store for non-existing query {}. "
+                  + "This is not expected and was likely due to a "
+                  + "race condition when the query was dropped before.",
+              appId);
+        }
       } catch (Exception e) {
         LOG.error("Error cleaning up state directory {}\n. {}", appId, e);
       }
--- a/ksqldb-test-util/src/main/java/io/confluent/ksql/test/util/KsqlTestFolder.java
+++ b/ksqldb-test-util/src/main/java/io/confluent/ksql/test/util/KsqlTestFolder.java
@@ -1,0 +1,66 @@
+/*
+ * Copyright 2021 Confluent Inc.
+ *
+ * Licensed under the Confluent Community License (the "License"; you may not use
+ * this file except in compliance with the License. You may obtain a copy of the
+ * License at
+ *
+ * http://www.confluent.io/confluent-community-license
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OF ANY KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations under the License.
+ */
+
+package io.confluent.ksql.test.util;
+
+import java.io.File;
+import java.lang.ref.WeakReference;
+import org.junit.rules.TemporaryFolder;
+
+/**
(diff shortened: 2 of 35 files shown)


ksqldb-engine/src/main/java/io/confluent/ksql/engine/QueryCleanupService.java
```diff
@@ -115,11 +115,14 @@ public String getAppId() {
     public void run() {
       try {
         final Path pathName = Paths.get(stateDir + "/" + appId);
-        FileUtils.deleteDirectory(new File(String.valueOf(pathName.normalize())));
-        LOG.warn("Deleted local state store for non-existing query {}. "
-            + "This is not expected and was likely due to a "
-            + "race condition when the query was dropped before.",
-            appId);
+        final File directory = new File(String.valueOf(pathName.normalize()));
+        if (directory.exists()) {
```
vvcephei, line 119: While staring at the test output, I noticed that the director sometimes doesn't exist when we enter this block. This check will add a small amount of extra race-condition protection, but we could still have two threads traversing the directory tree at the same time.

ksqldb-test-util/src/main/java/io/confluent/ksql/test/util/KsqlTestFolder.java
```diff
@@ -0,0 +1,67 @@
...
+  /**
+   * Create a temporary folder for testing. Rather than the system default
+   * temp directory, use the current working directory to avoid directories
+   * being deleted during tests.
+   */
+  public static TemporaryFolder temporaryFolder() {
+    final TemporaryFolder temporaryFolder = TemporaryFolder
+        .builder()
+        .assureDeletion()
+        .parentFolder(new File(System.getProperty("user.dir")))
```
vvcephei, line 42: By the way, this unintuitively named system property resolves to the current working directory.

vvcephei:
There was only this unrelated test failure, which is being fixed in another PR:
https://jenkins.confluent.io/job/Confluentinc%20Contributors/job/ksql/job/PR-7939/2/testReport/junit/io.confluent.ksql.test.planned/PlannedTestsUpToDateTest/shouldHaveLatestPlans_bytes___PROTOBUF_in_out_/

ksqldb-test-util/src/main/java/io/confluent/ksql/test/util/KsqlTestFolder.java
```diff
@@ -0,0 +1,66 @@
...
+   * temp directory, use the current working directory to avoid directories
+   * being deleted during tests.
+   */
+  public static TemporaryFolder temporaryFolder() {
+    final TemporaryFolder temporaryFolder = TemporaryFolder
+        .builder()
+        .parentFolder(new File(System.getProperty("user.dir")))
+        .build();
+
+    // Just in case the JVM exits without a proper test failure.
```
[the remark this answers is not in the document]
AlanConfluent: https://stackoverflow.com/questions/20829228/shutdown-hook-vs-finalizer-method

I guess finalize is less reliable.

AlanConfluent (approved):

ksqldb-test-util/src/main/java/io/confluent/ksql/test/util/KsqlTestFolder.java
vvcephei: It's just part of the JVM: https://docs.oracle.com/javase/tutorial/essential/environment/sysprop.html

vvcephei: Ah, I see what you mean. No; it'd be nice if they had an enum or something, but it's just a string.

ksqldb-test-util/src/main/java/io/confluent/ksql/test/util/KsqlTestFolder.java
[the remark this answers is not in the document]
vvcephei: Thanks! I actually hadn't thought of the finalizer, since this is more related to what happens when the JVM exits abruptly. For situations where the object would just get GCed normally, we'd expect JUnit to already have called `after()`.

But the finalizer might be a good way to try and remove the shutdown hook after the test finishes. I'll give that a shot before merging.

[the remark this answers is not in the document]
vvcephei: Oh, nevermind. That would also require wrapping the TemporaryFolder with a new object, which I already tried (to intercept `after()` and unregister the hook). It can be done, but it's a pain.

I think I'll just merge this as-is for now.

PR #7945, opened 2021-08-04
feat: add CREATE SOURCE TABLE syntax and metadata info

spena:
### Description 
This is part of https://github.com/confluentinc/ksql/pull/7474 which adds support of source tables and streams.

This PR just adds new syntax to create source tables using `CREATE SOURCE TABLE` statements. It also marks those tables internally for future reference, such as `CreateTable`, `CreateTableCommand` and `KsqlTable`. 

The `KsqlTable` is the object saved in the metastore. This object contains two new variables, `readOnly` and `materialized`. The `readOnly` just specifies the table is read-only; and `materialized` specifies this table is materialized (or should be materialized). I split the SOURCE keyword into these two variables because they mean different things. Also, we might support source streams in the future which they are read-only objects, but not materialized.

### Testing done 

### Reviewer checklist

### DIFF
--- a/ksqldb-engine/src/main/java/io/confluent/ksql/ddl/commands/CreateSourceFactory.java
+++ b/ksqldb-engine/src/main/java/io/confluent/ksql/ddl/commands/CreateSourceFactory.java
@@ -144,7 +144,7 @@ public final class CreateSourceFactory {
         Formats.from(outputNode.getKsqlTopic()),
         outputNode.getKsqlTopic().getKeyFormat().getWindowInfo(),
         Optional.of(outputNode.getOrReplace()),
-        Optional.of(CreateTable.Type.NORMAL.name())
+        Optional.of(false)
     );
   }
 
@@ -192,7 +192,7 @@ public final class CreateSourceFactory {
         buildFormats(statement.getName(), schema, props, ksqlConfig),
         getWindowInfo(props),
         Optional.of(statement.isOrReplace()),
-        Optional.of(statement.getType().name())
+        Optional.of(statement.isSource())
     );
   }
 
--- a/ksqldb-engine/src/main/java/io/confluent/ksql/ddl/commands/CreateSourceFactory.java
+++ b/ksqldb-engine/src/main/java/io/confluent/ksql/ddl/commands/CreateSourceFactory.java
@@ -134,6 +134,7 @@ public final class CreateSourceFactory {
     );
   }
 
+  // This method is called by CREATE_AS statements
   public CreateTableCommand createTableCommand(final KsqlStructuredDataOutputNode outputNode) {
     return new CreateTableCommand(
         outputNode.getSinkName().get(),
@@ -142,10 +143,12 @@ public final class CreateSourceFactory {
         outputNode.getKsqlTopic().getKafkaTopicName(),
         Formats.from(outputNode.getKsqlTopic()),
         outputNode.getKsqlTopic().getKeyFormat().getWindowInfo(),
-        Optional.of(outputNode.getOrReplace())
+        Optional.of(outputNode.getOrReplace()),
+        Optional.of(CreateTable.Type.NORMAL.name())
     );
   }
 
+  // This method is called by simple CREATE statements
   public CreateTableCommand createTableCommand(
       final CreateTable statement,
       final KsqlConfig ksqlConfig
@@ -188,7 +191,8 @@ public final class CreateSourceFactory {
         topicName,
         buildFormats(statement.getName(), schema, props, ksqlConfig),
         getWindowInfo(props),
-        Optional.of(statement.isOrReplace())
+        Optional.of(statement.isOrReplace()),
+        Optional.of(statement.getType().name())
     );
   }
 
--- a/ksqldb-engine/src/test/java/io/confluent/ksql/engine/rewrite/StatementRewriterTest.java
+++ b/ksqldb-engine/src/test/java/io/confluent/ksql/engine/rewrite/StatementRewriterTest.java
@@ -660,7 +660,7 @@ public class StatementRewriterTest {
         false,
         false,
         sourceProperties,
-        CreateTable.Type.NORMAL
+        false
     );
     when(mockRewriter.apply(tableElement1, context)).thenReturn(rewrittenTableElement1);
     when(mockRewriter.apply(tableElement2, context)).thenReturn(rewrittenTableElement2);
@@ -679,7 +679,7 @@ public class StatementRewriterTest {
                 false,
                 false,
                 sourceProperties,
-                CreateTable.Type.NORMAL
+                false
             )
         )
     );
--- a/ksqldb-engine/src/test/java/io/confluent/ksql/engine/rewrite/StatementRewriterTest.java
+++ b/ksqldb-engine/src/test/java/io/confluent/ksql/engine/rewrite/StatementRewriterTest.java
@@ -660,7 +660,7 @@ public class StatementRewriterTest {
         false,
         false,
         sourceProperties,
-        CreateTable.Type.SOURCE
+        CreateTable.Type.NORMAL
     );
     when(mockRewriter.apply(tableElement1, context)).thenReturn(rewrittenTableElement1);
     when(mockRewriter.apply(tableElement2, context)).thenReturn(rewrittenTableElement2);
@@ -679,7 +679,7 @@ public class StatementRewriterTest {
                 false,
                 false,
                 sourceProperties,
-                CreateTable.Type.SOURCE
+                CreateTable.Type.NORMAL
             )
         )
     );
--- a/ksqldb-engine/src/test/java/io/confluent/ksql/engine/rewrite/StatementRewriterTest.java
+++ b/ksqldb-engine/src/test/java/io/confluent/ksql/engine/rewrite/StatementRewriterTest.java
@@ -659,7 +659,8 @@ public class StatementRewriterTest {
         TableElements.of(tableElement1, tableElement2),
         false,
         false,
-        sourceProperties
+        sourceProperties,
+        CreateTable.Type.SOURCE
     );
     when(mockRewriter.apply(tableElement1, context)).thenReturn(rewrittenTableElement1);
     when(mockRewriter.apply(tableElement2, context)).thenReturn(rewrittenTableElement2);
@@ -677,7 +678,8 @@ public class StatementRewriterTest {
                 TableElements.of(rewrittenTableElement1, rewrittenTableElement2),
                 false,
                 false,
-                sourceProperties
+                sourceProperties,
+                CreateTable.Type.SOURCE
             )
         )
     );
--- a/ksqldb-execution/src/main/java/io/confluent/ksql/execution/ddl/commands/CreateTableCommand.java
+++ b/ksqldb-execution/src/main/java/io/confluent/ksql/execution/ddl/commands/CreateTableCommand.java
@@ -15,6 +15,7 @@
 
 package io.confluent.ksql.execution.ddl.commands;
 
+import com.fasterxml.jackson.annotation.JsonIgnore;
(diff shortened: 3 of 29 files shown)


ksqldb-parser/src/main/java/io/confluent/ksql/parser/tree/CreateTable.java
```diff
@@ -27,15 +27,22 @@
 
 @Immutable
 public class CreateTable extends CreateSource implements ExecutableDdlStatement {
+  public enum Type {
+    SOURCE,
+    NORMAL
+  }
```
spena, line 33: I had to create the `Type` inside the `CreateTable.Type` enum because that gives better meaning to the table type we want to use (source or normal). I could have added to the parent class, `CreateSource.Type`, but that sounds like source is a table or stream type. I'm still not sure whether I put the type in the right place, so feedback is welcome.

ksqldb-execution/src/main/java/io/confluent/ksql/execution/ddl/commands/CreateTableCommand.java
```diff
@@ -36,7 +37,8 @@ public CreateTableCommand(
       @JsonProperty(value = "topicName", required = true) final String topicName,
       @JsonProperty(value = "formats", required = true) final Formats formats,
       @JsonProperty(value = "windowInfo") final Optional<WindowInfo> windowInfo,
-      @JsonProperty(value = "orReplace", defaultValue = "false") final Optional<Boolean> orReplace
+      @JsonProperty(value = "orReplace", defaultValue = "false") final Optional<Boolean> orReplace,
+      @JsonProperty(value = "type", defaultValue = "NORMAL") final Optional<String> type
```
spena, line 41: I used a `String` instead of the `CreateTable.Type` because of dependencies. I would have to add ksql-parser deps to the ksql-execution, which I'm not sure about it. So I just converted the type to a string from the `CreateSourceFactory` which calls this class. I then converted back to `CreateTable.Type` on the `DdlCommandExec`.

ksqldb-parser/src/main/java/io/confluent/ksql/parser/tree/CreateTable.java
jzaralim: Some ideas:
* Put it under `CreateSource` with a different name, like `ownershipType`
* Add the enum to `KsqlConstants` in `ksqldb-common` - `CreateTableCommand` won't have to use strings anymore

ksqldb-rest-app/src/test/java/io/confluent/ksql/rest/server/TemporaryEngine.java
```diff
@@ -161,7 +161,9 @@ protected void after() {
                 SCHEMA,
                 Optional.empty(),
                 false,
-                topic
+                topic,
+                false,
+                false
```
mjsax, line 166: What is `TemporaryEngine` and would we need to add support for source tables?

ksqldb-rest-app/src/test/java/io/confluent/ksql/rest/server/execution/InsertValuesExecutorTest.java
```diff
@@ -1121,7 +1121,9 @@ private void givenDataSourceWithSchema(
           schema,
           Optional.empty(),
           false,
-          topic
+          topic,
+          false,
+          false
```
mjsax, line 1126: Does this PR add a test for rejected `INSERT INTO` for source tables? Should it? Or do we have some other work item for it?

ksqldb-rest-app/src/test/java/io/confluent/ksql/rest/server/resources/KsqlResourceTest.java
```diff
@@ -2640,7 +2640,9 @@ private void givenSource(
             schema,
             Optional.empty(),
             false,
-            ksqlTopic
+            ksqlTopic,
+            false,
+            false
```
mjsax, line 2645: As above?

mjsax (commented):
Couple of question (mainly for my own education...)

ksqldb-engine/src/test/java/io/confluent/ksql/engine/rewrite/StatementRewriterTest.java
```diff
@@ -659,7 +659,8 @@ public void shouldRewriteCreateTable() {
         TableElements.of(tableElement1, tableElement2),
         false,
         false,
-        sourceProperties
+        sourceProperties,
+        CreateTable.Type.SOURCE
```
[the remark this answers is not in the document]
spena: I'll revert this change (back to NORMAL). I don't think it is testing anything related to the type.

ksqldb-execution/src/main/java/io/confluent/ksql/execution/ddl/commands/CreateTableCommand.java
spena: Will use a boolean value instead of a Enum. I had another idea with the types to support types like `external`, `temporary`, etc in the future, so for compatibility I wanted to do it now, but I don't think that's really necessary. If they're supported, then users could mix the types, so we'll end up adding more parameters. I'll switch with the boolean instead.

ksqldb-metastore/src/test/java/io/confluent/ksql/util/MetaStoreFixture.java
```diff
@@ -116,7 +116,9 @@ public static MutableMetaStore getNewMetaStore(
         test2Schema,
         Optional.empty(),
         false,
-        ksqlTopic2
+        ksqlTopic2,
+        false,
+        false
```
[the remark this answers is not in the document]
spena: Seems for testing only. Just a factory class to create a metastore with some data.

ksqldb-parser/src/main/java/io/confluent/ksql/parser/AstBuilder.java
```diff
@@ -1438,7 +1443,8 @@ public Node visitAssertTable(final AssertTableContext context) {
           TableElements.of(elements),
           false,
           false,
-          CreateSourceProperties.from(properties)
+          CreateSourceProperties.from(properties),
+          CreateTable.Type.NORMAL
```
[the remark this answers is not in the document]
spena: This is part of the new testing framework we've been talking about and nobody has time to work on it. I think we should change the assert methods later to support source tables so we leave less work for the person who'll work on the testing framework.

ksqldb-parser/src/main/java/io/confluent/ksql/parser/SqlFormatter.java
```diff
@@ -259,7 +259,11 @@ protected Void visitCreateStream(final CreateStream node, final Integer indent)
 
     @Override
     protected Void visitCreateTable(final CreateTable node, final Integer indent) {
-      formatCreate(node, "TABLE");
+      formatCreate(node,
+          node.getType() == CreateTable.Type.SOURCE
```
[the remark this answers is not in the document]
spena: It's used by the `QueryLogger` and other injector classes that re-formats the string statement when changing something in the statement (i.e. injecting a schema fetched from Schema Registry). I don't think we should re-format string statements, but this is an old class used by some legacy code.

ksqldb-parser/src/test/java/io/confluent/ksql/parser/KsqlParserTest.java
```diff
@@ -180,7 +180,9 @@ public void init() {
         ORDERS_SCHEMA,
         Optional.empty(),
         false,
-        ksqlTopicItems
+        ksqlTopicItems,
+        false,
+        false
```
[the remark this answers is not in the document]
spena: This object is not even used anywhere in the test. I'll remove all the unused code. I'll add a test to parse CREATE SOURCE TABLE too.

ksqldb-rest-app/src/test/java/io/confluent/ksql/rest/server/execution/InsertValuesExecutorTest.java
spena: This PR does not add it. Will be in a follow-up.

ksqldb-rest-app/src/test/java/io/confluent/ksql/rest/server/TemporaryEngine.java
spena: We'll add it in a follow-up PR. This `TemporaryEngine` is used by the `ListSourceExecutorTest`, which tests the DESCRIBE and SHOW commands. There's a JIRA to make sure DESCRIBE correctly handles source streams & tables.

ksqldb-parser/src/main/java/io/confluent/ksql/parser/AstBuilder.java
```diff
@@ -281,13 +281,18 @@ public Node visitCreateTable(final SqlBaseParser.CreateTableContext context) {
 
       final Map<String, Literal> properties = processTableProperties(context.tableProperties());
 
+      final CreateTable.Type type = context.SOURCE() != null
```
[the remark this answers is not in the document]
mjsax: If it's not `null`, what does it return?

ksqldb-parser/src/main/java/io/confluent/ksql/parser/SqlFormatter.java
```diff
@@ -259,7 +259,11 @@ protected Void visitCreateStream(final CreateStream node, final Integer indent)
 
     @Override
     protected Void visitCreateTable(final CreateTable node, final Integer indent) {
-      formatCreate(node, "TABLE");
+      formatCreate(node,
+          node.isSource()
+              ? "SOURCE TABLE"
```
jzaralim, line 264: We can keep this method as is, and update formatCreate to add the "SOURCE" keyword if the node is a source. Then we won't have to do the same thing for streams.

jzaralim (approved):
lgtm, with one comment

ksqldb-parser/src/main/java/io/confluent/ksql/parser/SqlFormatter.java
spena: Done

ksqldb-engine/src/main/java/io/confluent/ksql/ddl/commands/CreateSourceFactory.java
```diff
@@ -142,10 +143,12 @@ public CreateTableCommand createTableCommand(final KsqlStructuredDataOutputNode
         outputNode.getKsqlTopic().getKafkaTopicName(),
         Formats.from(outputNode.getKsqlTopic()),
         outputNode.getKsqlTopic().getKeyFormat().getWindowInfo(),
-        Optional.of(outputNode.getOrReplace())
+        Optional.of(outputNode.getOrReplace()),
+        Optional.of(false)
```
guozhangwang, line 147: nit: how about define another `isDerived` here, rather than using `false` to indicate `is not source`?

guozhangwang (commented):
Sorry for the late reviews. All LGTM, I'm just wondering if we want to keep the `isSource` as a boolean or keep it as a enum (of course for not it would only have two possible values, source, or derived).

ksqldb-engine/src/main/java/io/confluent/ksql/ddl/commands/CreateSourceFactory.java
spena: The parameter name is tricky. I initially wanted to use a Enum for future purposes like External, Temp, Source, etc types, but I found that those other types can be mixed with source too, like external source, temp source, or something. So I ended up using a boolean. IsDerived makes sense too. I think any of isDerived or isSource could work.

PR #7950, opened 2021-08-05
feat: fail query if rebalance or error detected

nateab:
### Description 
Listen for errors or rebalances in the persistent query and if one occurs, fail the scalable push query since there is a chance we can miss something or get redundant rows 
### Testing done 
Unit tests added 

### Reviewer checklist

### DIFF
--- a/ksqldb-engine/src/main/java/io/confluent/ksql/physical/scalablepush/ProcessingQueue.java
+++ b/ksqldb-engine/src/main/java/io/confluent/ksql/physical/scalablepush/ProcessingQueue.java
@@ -37,6 +37,8 @@ public class ProcessingQueue {
   private final int queueSizeLimit;
   private boolean closed = false;
   private boolean droppedRows = false;
+  private boolean hasError = false;
+  private boolean hasStateChange = false;
   private Runnable newRowCallback = () -> { };
 
   public ProcessingQueue(final QueryId queryId) {
@@ -104,6 +106,14 @@ public class ProcessingQueue {
    */
   public synchronized boolean hasDroppedRows() {
     return droppedRows;
+  }
+
+  public synchronized boolean hasError() {
+    return hasError;
+  }
+
+  public synchronized boolean hasStateChange() {
+    return hasStateChange;
   }
 
   public QueryId getQueryId() {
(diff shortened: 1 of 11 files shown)


ableegoldman:
Why are we failing the query if a rebalance is detected? It's a natural part of a consumer group. Plus it can be quite frequent, so these queries would be _very_ short lived if the first rebalance took them down (or the first rebalance after the real first rebalance that it goes through during startup).

Not to mention we manually trigger rebalances from within Streams for all sorts of things 😈

nateab:
I think what Alan said is that for scalable push queries, if the persistent query errors or rebalances, we can no longer guarantee the correctness of the scalable push queries so we should throw an error. I think specifically that we cannot determine which offset to read from or something like that. I may have also butchered the explanation, so I will let @AlanConfluent confirm/correct me haha.

ableegoldman:
Fair enough -- I definitely don't know enough about the scalable push query details to understand why that is, but I believe it. 

Even so, that might really throw a wrench in the usefulness of scalable push queries -- but I also definitely don't know enough about how they are used to really say what is/isn't useful about them 🤷‍♀️

AlanConfluent:
@ableegoldman 

> Why are we failing the query if a rebalance is detected? It's a natural part of a consumer group. Plus it can be quite frequent, so these queries would be very short lived if the first rebalance took them down (or the first rebalance after the real first rebalance that it goes through during startup).

At the moment, if a scalable push query starts off processing data with all of the nodes handling persistent query tasks and no rebalancing occurs, we'll give results exactly once.  If we rebalance, there's the possibility that we'll miss something or get redundant rows.  For this reason, we wanted to play it safe and start off by just issuing an error and giving the best experience.

I was chatting with John a bit about this and actually, it does seem like we can potentially do some followup changes that allow for rebalances and that have at least once semantics.  Which route to go depends in part on how often rebalances happen in practice.  If they do happen often enough, as it seems that they do, it makes more sense to try to avoid failing the request and keep ALO.  This is especially true since at the moment, there's no way to pick up where you leave off after a failure, so EOS isn't really helpful in the current form.

Let me discuss it offline and check back in with this PR as to whether it's helpful to have this as a checkpoint.

PR #7955, opened 2021-08-05
feat: add persistent query saturation metric

rodesai:
Add a persistent query saturation metric, and report it over JMX. Saturation
is computed by sampling total-blocked-time for each stream thread. We sample
the total blocked time every minute, and compute saturation by looking back
5 minutes and computing how long the thread was blocked in that interval.
Saturation is (5 minutes - blocked time)/(5 minutes).

This patch also adds a reporter that exposes reported data points over JMX.
The reporter adds a new Metric for every new data point name/tag combination
it sees, and implements the metric by reading the latest data point for the
name/tag, with a threshold on staleness.

### DIFF
--- a/ksqldb-engine/src/main/java/io/confluent/ksql/internal/JmxDataPointsReporter.java
+++ b/ksqldb-engine/src/main/java/io/confluent/ksql/internal/JmxDataPointsReporter.java
@@ -1,0 +1,98 @@
+/*
+ * Copyright 2021 Confluent Inc.
+ *
+ * Licensed under the Confluent Community License (the "License"); you may not use
+ * this file except in compliance with the License.  You may obtain a copy of the
+ * License at
+ *
+ * http://www.confluent.io/confluent-community-license
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OF ANY KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations under the License.
+ */
+
+package io.confluent.ksql.internal;
+
+import java.time.Duration;
+import java.time.Instant;
+import java.util.List;
+import java.util.Map;
+import java.util.Objects;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.atomic.AtomicReference;
+import org.apache.kafka.common.MetricName;
+import org.apache.kafka.common.metrics.Gauge;
+import org.apache.kafka.common.metrics.MetricConfig;
+import org.apache.kafka.common.metrics.Metrics;
+
+public class JmxDataPointsReporter implements MetricsReporter {
+  private final Metrics metrics;
+  private final String group;
+  private final Map<MetricName, DataPointBasedGauge> gauges = new ConcurrentHashMap<>();
+  private final Duration staleThreshold;
+
+  public JmxDataPointsReporter(
+      final Metrics metrics,
+      final String group,
+      final Duration staleThreshold
+  ) {
+    this.metrics = Objects.requireNonNull(metrics, "metrics");
+    this.group = Objects.requireNonNull(group, "group");
+    this.staleThreshold = Objects.requireNonNull(staleThreshold, "staleThreshold");
+  }
+
+  @Override
+  public void report(final List<DataPoint> dataPoints) {
+    dataPoints.forEach(this::report);
+  }
+
+  private void report(final DataPoint dataPoint) {
+    final MetricName metricName
+        = metrics.metricName(dataPoint.getName(), group, dataPoint.getTags());
+    if (gauges.containsKey(metricName)) {
+      gauges.get(metricName).dataPointRef.set(dataPoint);
+    } else {
+      gauges.put(metricName, new DataPointBasedGauge(dataPoint, staleThreshold));
+      metrics.addMetric(metricName, gauges.get(metricName));
+    }
+  }
+
+  @Override
+  public void cleanup(final String name, final Map<String, String> tags) {
+    final MetricName metricName = metrics.metricName(name, group, tags);
+    metrics.removeMetric(metricName);
+    gauges.remove(metricName);
+  }
+
+  @Override
+  public void close() {
+  }
+
+  @Override
+  public void configure(final Map<String, ?> map) {
+  }
+
+  private static final class DataPointBasedGauge implements Gauge<Object> {
+    private final AtomicReference<DataPoint> dataPointRef;
+    private final Duration staleThreshold;
+
+    private DataPointBasedGauge(
+        final DataPoint initial,
+        final Duration staleThreshold
+    ) {
+      this.dataPointRef = new AtomicReference<>(initial);
+      this.staleThreshold = staleThreshold;
+    }
+
+    @Override
+    public Object value(final MetricConfig metricConfig, final long now) {
+      final DataPoint dataPoint = dataPointRef.get();
+      if (dataPoint.getTime().isAfter(Instant.ofEpochMilli(now).minus(staleThreshold))) {
+        return dataPoint.getValue();
+      }
+      return null;
+    }
+  }
+}
--- a/ksqldb-engine/src/main/java/io/confluent/ksql/utilization/PersistentQuerySaturationMetrics.java
+++ b/ksqldb-engine/src/main/java/io/confluent/ksql/utilization/PersistentQuerySaturationMetrics.java
@@ -233,7 +233,7 @@ public class PersistentQuerySaturationMetrics implements Runnable {
             threadName,
             k -> {
               LOGGER.debug("Adding saturation for new thread: {}", k);
-              return new ThreadSaturation(startTime, window, sampleMargin);
+              return new ThreadSaturation(threadName, startTime, window, sampleMargin);
             }
         );
         LOGGER.debug("Record thread {} sample {} {}", threadName, totalBlocked, startTime);
@@ -265,16 +265,19 @@ public class PersistentQuerySaturationMetrics implements Runnable {
   }
 
   private static final class ThreadSaturation {
+    private final String threadName;
     private final List<BlockedTimeSample> samples = new LinkedList<>();
     private final Instant startTime;
     private final Duration window;
     private final Duration sampleMargin;
 
     private ThreadSaturation(
+        final String threadName,
         final long startTime,
         final Duration window,
         final Duration sampleMargin
     ) {
+      this.threadName = Objects.requireNonNull(threadName, "threadName");
       this.startTime = Instant.ofEpochMilli(startTime);
       this.window = Objects.requireNonNull(window, "window");
       this.sampleMargin = Objects.requireNonNull(sampleMargin, "sampleMargin");
@@ -288,18 +291,23 @@ public class PersistentQuerySaturationMetrics implements Runnable {
       final Instant windowStart = now.minus(window);
       final Instant earliest = now.minus(window.plus(sampleMargin));
       final Instant latest = now.minus(window.minus(sampleMargin));
+      LOGGER.debug("{}: record and measure with now {},  window {} ({} : {})", threadName, now, windowStart, earliest, latest);
       samples.add(current);
       samples.removeIf(s -> s.timestamp.isBefore(earliest));
       if (!inRange(samples.get(0).timestamp, earliest, latest) && !startTime.isAfter(windowStart)) {
         return Optional.empty();
       }
+      BlockedTimeSample startSample = samples.get(0);
+      LOGGER.debug("{}: start sample {}", threadName, startSample);
       double blocked =
-          Math.max(current.totalBlockedTime - samples.get(0).totalBlockedTime, 0);
+          Math.max(current.totalBlockedTime - startSample.totalBlockedTime, 0);
       Instant observedStart = samples.get(0).timestamp;
       if (startTime.isAfter(windowStart)) {
+        LOGGER.debug("{}: start time {} is after window start", threadName, startTime);
         blocked += Duration.between(windowStart, startTime).toNanos();
         observedStart = windowStart;
       }
+
       final Duration duration = Duration.between(observedStart, current.timestamp);
       final double durationNs = duration.toNanos();
       return Optional.of((durationNs - Math.min(blocked, durationNs)) / durationNs);
@@ -313,6 +321,14 @@ public class PersistentQuerySaturationMetrics implements Runnable {
     private BlockedTimeSample(final Instant timestamp, final double totalBlockedTime) {
       this.timestamp = timestamp;
       this.totalBlockedTime = totalBlockedTime;
+    }
+
+    @Override
+    public String toString() {
+      return "BlockedTimeSample{" +
+          "timestamp=" + timestamp +
+          ", totalBlockedTime=" + totalBlockedTime +
+          '}';
     }
   }
 }
--- a/ksqldb-engine/src/main/java/io/confluent/ksql/utilization/PersistentQuerySaturationMetrics.java
+++ b/ksqldb-engine/src/main/java/io/confluent/ksql/utilization/PersistentQuerySaturationMetrics.java
@@ -315,13 +315,20 @@ public class PersistentQuerySaturationMetrics implements Runnable {
       final Instant windowStart = now.minus(window);
       final Instant earliest = now.minus(window.plus(sampleMargin));
       final Instant latest = now.minus(window.minus(sampleMargin));
-      LOGGER.debug("{}: record and measure with now {},  window {} ({} : {})", threadName, now, windowStart, earliest, latest);
+      LOGGER.debug(
+          "{}: record and measure with now {},  window {} ({} : {})",
+          threadName,
+          now,
+          windowStart,
+          earliest,
+          latest
+      );
       samples.add(current);
       samples.removeIf(s -> s.timestamp.isBefore(earliest));
       if (!inRange(samples.get(0).timestamp, earliest, latest) && !startTime.isAfter(windowStart)) {
         return Optional.empty();
       }
-      BlockedTimeSample startSample = samples.get(0);
+      final BlockedTimeSample startSample = samples.get(0);
       LOGGER.debug("{}: start sample {}", threadName, startSample);
       double blocked =
           Math.max(current.totalBlockedTime - startSample.totalBlockedTime, 0);
@@ -349,10 +356,10 @@ public class PersistentQuerySaturationMetrics implements Runnable {
 
     @Override
     public String toString() {
-      return "BlockedTimeSample{" +
-          "timestamp=" + timestamp +
-          ", totalBlockedTime=" + totalBlockedTime +
-          '}';
+      return "BlockedTimeSample{"
+          + "timestamp=" + timestamp
+          + ", totalBlockedTime=" + totalBlockedTime
+          + '}';
     }
   }
 }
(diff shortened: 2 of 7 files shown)


ksqldb-engine/src/main/java/io/confluent/ksql/internal/JmxDataPointsReporter.java
```diff
@@ -0,0 +1,98 @@
...
+    this.staleThreshold = Objects.requireNonNull(staleThreshold, "staleThreshold");
+  }
+
+  @Override
+  public void report(final List<DataPoint> dataPoints) {
+    dataPoints.forEach(this::report);
+  }
+
+  private void report(final DataPoint dataPoint) {
+    final MetricName metricName
```
wcarlson5, line 52: would either the group of the datapoint contain the QueryID?

ksqldb-engine/src/main/java/io/confluent/ksql/utilization/PersistentQuerySaturationMetrics.java
```diff
@@ -0,0 +1,318 @@
...
+    try {
+      final Collection<PersistentQueryMetadata> queries = engine.getPersistentQueries();
+      final Optional<Double> saturation = queries.stream()
+          .map(q -> measure(now, q))
+          .max(PersistentQuerySaturationMetrics::compareSaturation)
+          .orElse(Optional.of(0.0));
+      saturation.ifPresent(s -> report(now, s));
+
+      final Set<String> appIds = queries.stream()
+          .map(PersistentQueryMetadata::getQueryApplicationId)
```
wcarlson5, line 103: Can you use QueryID instead? That will work better with the shared runtimes.

ksqldb-engine/src/main/java/io/confluent/ksql/utilization/PersistentQuerySaturationMetrics.java
```diff
@@ -0,0 +1,318 @@
...
+    } else if (!b.isPresent()) {
+      return -1;
+    } else {
+      return Double.compare(a.get(), b.get());
+    }
+  }
+
+  private Optional<Double> measure(final Instant now, final PersistentQueryMetadata queryMetadata) {
+    final KafkaStreamsSaturation ksSaturation = perKafkaStreamsStats.computeIfAbsent(
+        queryMetadata.getQueryApplicationId(),
```
wcarlson5, line 130: ditto here, or filter for duplicates if you are intending to have it be per runtime not per query

wcarlson5 (commented):
Hey not sure if this was ready for review but I have a couple of comments that might affect this if the shared runtimes gets merged first

ksqldb-engine/src/main/java/io/confluent/ksql/utilization/PersistentQuerySaturationMetrics.java
```diff
@@ -0,0 +1,334 @@
...
+        LOGGER.debug("Record thread {} sample {} {}", threadName, totalBlocked, startTime);
+        final BlockedTimeSample blockedTimeSample = new BlockedTimeSample(now, totalBlocked);
+        final Optional<Double> measured = threadSaturation.measure(now, blockedTimeSample);
+        LOGGER.debug(
+            "Measured value for thread {}: {}",
+            threadName,
+            (measured.map(Object::toString).orElse(""))
+        );
+        measured.ifPresent(m -> reportThreadSaturation(now, m, threadName, reporter));
+        saturation = compareSaturation(saturation, measured) > 0 ? saturation : measured;
```
lct45, line 248: To see if I'm following - this basically picks the larger of saturation or measured, right? So as we go through the loop to calculate thread metrics, we'll track the largest thread saturation and then ultimately report that when we exit the for loop. How would saturation ever not exist? We set it to `Optional.of(0.0)` to begin with so if the first measured doesn't exist we'd return -1 and reset saturation to `Optional.of(0.0)`, right? So is the check for if the first parameter exits just for when we compare saturation in `run()`?

ksqldb-engine/src/main/java/io/confluent/ksql/utilization/PersistentQuerySaturationMetrics.java
```diff
@@ -0,0 +1,334 @@
...
+          : Sets.difference(new HashSet<>(perThreadSaturation.keySet()), byThread.keySet())) {
+        perThreadSaturation.remove(threadName);
+        reporter.cleanup(QUERY_THREAD_SATURATION, ImmutableMap.of(THREAD_ID, threadName));
+      }
+      return saturation;
+    }
+
+    private void cleanup(final MetricsReporter reporter) {
+      for (final String threadName : perThreadSaturation.keySet()) {
+        reporter.cleanup(QUERY_THREAD_SATURATION, ImmutableMap.of(THREAD_ID, threadName));
```
lct45, line 261: Are the extra thread metrics not already cleaned up at the end of `KafkaStreamsSaturation#measure`?

ksqldb-engine/src/main/java/io/confluent/ksql/utilization/PersistentQuerySaturationMetrics.java
```diff
@@ -0,0 +1,334 @@
...
+
+  @Override
+  public void run() {
+    final Instant now = time.get();
+
+    try {
+      final Collection<PersistentQueryMetadata> queries = engine.getPersistentQueries();
+      final Optional<Double> saturation = queries.stream()
+          .map(q -> measure(now, q))
+          .max(PersistentQuerySaturationMetrics::compareSaturation)
```
lct45, line 98: What two metrics are we comparing here?

ksqldb-engine/src/main/java/io/confluent/ksql/internal/JmxDataPointsReporter.java
rodesai: yep - that's correct for metrics that are specific to a given query

ksqldb-engine/src/main/java/io/confluent/ksql/utilization/PersistentQuerySaturationMetrics.java
rodesai: On the contrary I think it makes sense to track saturation at the runtime/app level. At the lowest level it's a property of a stream thread. We then take the max saturation over all stream threads to compute runtime-level saturation. And max over all runtimes to compute node-level saturation.

ksqldb-engine/src/main/java/io/confluent/ksql/utilization/PersistentQuerySaturationMetrics.java
```diff
@@ -0,0 +1,334 @@
...
+        final String name,
+        final MetricsReporter reporter
+    ) {
+      LOGGER.info("Reporting thread saturation {} for {}", saturation, name);
+      reporter.report(ImmutableList.of(
+          new DataPoint(
+              now,
+              QUERY_THREAD_SATURATION,
+              saturation,
+              ImmutableMap.of(THREAD_ID, name)
```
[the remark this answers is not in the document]
rodesai: we're reporting query-level below

ksqldb-engine/src/main/java/io/confluent/ksql/utilization/PersistentQuerySaturationMetrics.java
rodesai: It may not exist if we don't have good data to compute saturation. This agent runs every minute and samples the total blocked time for each stream thread. It then computes the blocked time for the current window by looking back five minutes and taking the difference between the current value and that sample. However, it's possible that there isn't a good (as defined as a sample that's within sampleMargin) sample from five minutes ago (e.g. if the thread couldn't run for some reason). If that happens, we don't know the current saturation value so we return Optional.empty().

ksqldb-engine/src/main/java/io/confluent/ksql/utilization/PersistentQuerySaturationMetrics.java
wcarlson5: That makes sense. It looks like the metrics are still being created for each query, can you filter out duplicates? Otherwise I think we get an illegal argument exception

ksqldb-engine/src/main/java/io/confluent/ksql/utilization/PersistentQuerySaturationMetrics.java
rodesai: we need to clean up the thread metrics in the case a thread restarts, or in the case an app is closed. This method is called in the latter case.

ksqldb-engine/src/main/java/io/confluent/ksql/utilization/PersistentQuerySaturationMetrics.java
```diff
@@ -0,0 +1,334 @@
...
+    }
+
+    private Optional<Double> measure(final Instant now, final BlockedTimeSample current) {
+      final Instant windowStart = now.minus(window);
+      final Instant earliest = now.minus(window.plus(sampleMargin));
+      final Instant latest = now.minus(window.minus(sampleMargin));
+      LOGGER.debug("{}: record and measure with now {},  window {} ({} : {})", threadName, now, windowStart, earliest, latest);
+      samples.add(current);
+      samples.removeIf(s -> s.timestamp.isBefore(earliest));
+      if (!inRange(samples.get(0).timestamp, earliest, latest) && !startTime.isAfter(windowStart)) {
```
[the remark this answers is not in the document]
rodesai: > Why is it okay for the metric to be out of range if the thread was started after windowStart?
If the thread was started after `windowStart` we want to consider the time from `windowStart` to the thread start time as blocked.

>  I guess since we're no longer able to guarantee that the sample time happens every 5 minutes this helps us get a good set of metrics for our window?

Yep.

>  If so, why is sampleMargin a variable instead of a fixed time we decide on?

I wanted to leave the option open to have it be configurable. But it shouldn't ever change once the server is started.

ksqldb-engine/src/main/java/io/confluent/ksql/utilization/PersistentQuerySaturationMetrics.java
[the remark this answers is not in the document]
lct45: Ahh myb, i read thread-id as task-id

ksqldb-engine/src/main/java/io/confluent/ksql/utilization/PersistentQuerySaturationMetrics.java
```diff
@@ -0,0 +1,334 @@
...
+import org.apache.kafka.common.Metric;
+import org.apache.kafka.streams.KafkaStreams;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class PersistentQuerySaturationMetrics implements Runnable {
+  private static final Logger LOGGER
+      = LoggerFactory.getLogger(PersistentQuerySaturationMetrics.class);
+
+  private static final String QUERY_SATURATION = "query-saturation";
```
lct45, line 50: nit: it looks like for all the other ksql metrics we use `_` and not `-`, imo it would be good to be consistent and use `_` instead here

rodesai: where do you see that? `KsqlEngineMetrics` for example uses `-` as the delimiter. It gets converted to `_` by the telemetry reporter though, so it should all come out the same on the other end anyway.

lct45: Ahh I was looking on the cc-spec-ksql allow-list and on the metrics pipeline but I guess that's after it's cleaned so it shouldn't matter

ksqldb-engine/src/main/java/io/confluent/ksql/internal/JmxDataPointsReporter.java
```diff
@@ -0,0 +1,98 @@
...
+
+  @Override
+  public void configure(final Map<String, ?> map) {
+  }
+
+  private static final class DataPointBasedGauge implements Gauge<Object> {
+    private final AtomicReference<DataPoint> dataPointRef;
+    private final Duration staleThreshold;
+
+    private DataPointBasedGauge(
```
guozhangwang, line 81: My understanding is that when the underlying stream thread within the KS runtime comes and goes as we replace them, we would keep each as a separate data point and automatically "mute" those data points that are too stale, meaning the thread may be already gone. So when we apply the saturation equation we may temporarily be counting on some threads that are already gone but that would resume back to normal. Is that right?

ksqldb-engine/src/main/java/io/confluent/ksql/utilization/PersistentQuerySaturationMetrics.java
```diff
@@ -0,0 +1,334 @@
...
+        final MetricsReporter reporter
+    ) {
+      final Map<String, Map<String, Metric>> byThread = metricsByThread(kafkaStreams);
+      Optional<Double> saturation = Optional.of(0.0);
+      for (final Map.Entry<String, Map<String, Metric>> entry : byThread.entrySet()) {
+        final String threadName = entry.getKey();
+        final Map<String, Metric> metricsForThread = entry.getValue();
+        if (!metricsForThread.containsKey(STREAMS_TOTAL_BLOCKED_TIME)
+            || !metricsForThread.containsKey(STREAMS_THREAD_START_TIME)) {
+          LOGGER.info("Missing required metrics for thread: {}", threadName);
```
guozhangwang, line 225: If the thread is replaced after `metricsByThread` we would not see the metrics since they are GC'ed and would end here, is that right?

ksqldb-engine/src/main/java/io/confluent/ksql/utilization/PersistentQuerySaturationMetrics.java
```diff
@@ -0,0 +1,334 @@
...
+        final Optional<Double> measured = threadSaturation.measure(now, blockedTimeSample);
+        LOGGER.debug(
+            "Measured value for thread {}: {}",
+            threadName,
+            (measured.map(Object::toString).orElse(""))
+        );
+        measured.ifPresent(m -> reportThreadSaturation(now, m, threadName, reporter));
+        saturation = compareSaturation(saturation, measured) > 0 ? saturation : measured;
+      }
+      saturation.ifPresent(s -> reportQuerySaturation(now, s, queryId, reporter));
```
guozhangwang, line 250: It may be a duplicate comment with @wcarlson5 but with the shared runtime the mapping from queries -> threads is no longer one to many, but many to many, so we may end up double maxing the thread saturation for some queries? Since now it is just taking the max that maybe fine, but if we ever change it to a different equation we may end up duplicating the measurements.

ksqldb-engine/src/main/java/io/confluent/ksql/internal/JmxDataPointsReporter.java
```diff
@@ -0,0 +1,98 @@
...
+      this.staleThreshold = staleThreshold;
+    }
+
+    @Override
+    public Object value(final MetricConfig metricConfig, final long now) {
+      final DataPoint dataPoint = dataPointRef.get();
+      if (dataPoint.getTime().isAfter(Instant.ofEpochMilli(now).minus(staleThreshold))) {
+        return dataPoint.getValue();
+      }
+      return null;
```
guozhangwang, line 95: How would we interpret `null` on the other end? Let's say for some reason we do not get newly reported value for a while and the old value is stale, would this end up showing as a `unknown` value on the user-facing UI / our own monitoring, or something else?

ksqldb-engine/src/main/java/io/confluent/ksql/utilization/PersistentQuerySaturationMetrics.java
guozhangwang: Also with the new shared runtime, using `max` over all the threads that host at least one task of the query may be largely exaggerating the saturation metrics: e.g. let's say a single thread hosting some task from query A and some task from query B, and when query B's workload becomes a beast to drag down that thread, query A's saturation metric would also report as "full". This would be bad.. 

In the new shared runtime, there's no good way to tell how much of the saturation of a thread was contributed by one of its hosted query than the other, so a second least approach may be to just consider averaging over the threads.. thoughts @lct45 @wcarlson5

ksqldb-engine/src/main/java/io/confluent/ksql/internal/JmxDataPointsReporter.java
rodesai: Yep - once the saturation metric agent sees that a thread is missing, it will clean up the underlying `KafkaMetric` by calling `cleanup` on this class.

ksqldb-engine/src/main/java/io/confluent/ksql/utilization/PersistentQuerySaturationMetrics.java
rodesai: We really should never hit this condition. If a thread is replaced then it won't be returned by `metricsByThread` at all.

ksqldb-engine/src/main/java/io/confluent/ksql/internal/JmxDataPointsReporter.java
rodesai: The reporter should just skip reporting the value, so we'd see missing data in the metrics store (which is what we want)

ksqldb-engine/src/main/java/io/confluent/ksql/utilization/PersistentQuerySaturationMetrics.java
rodesai: > let's say a single thread hosting some task from query A and some task from query B, and when query B's workload becomes a beast to drag down that thread, query A's saturation metric would also report as "full". This would be bad..

I agree that's what would happen, but I think that's actually what we want. The saturation metric should tell you whether a given query is bottlenecked or not - not what % of the capacity that one query is using. The latter is a more challenging problem (I have an idea here around measuring the average record processing time and multiplying by the observed throughput into the query's source partitions) that we're not aiming to solve with this patch. I can see why this might be confusing though - we could leave the query-level metric internal for now.

guozhangwang:
LGTM!

wcarlson5 (approved):
LGTM

ksqldb-engine/src/main/java/io/confluent/ksql/utilization/PersistentQuerySaturationMetrics.java
guozhangwang: Cool, let's table that for now and merge as-is. We can see how it works well or not in production for monitoring purposes and decide whether we want to do something differently. I was concerned that we may exaggerating the saturation situation and hence put too much monitoring noise to ourselves, where a query that was only partially saturated on one of the hosting threads but not on others would be shown as 100% saturated. But that could be totally over-thinking too :)

guozhangwang:
LGTM!

ksqldb-engine/src/main/java/io/confluent/ksql/utilization/PersistentQuerySaturationMetrics.java
```diff
@@ -0,0 +1,365 @@
...
+      Instant observedStart = samples.get(0).timestamp;
+      if (startTime.isAfter(windowStart)) {
+        LOGGER.debug("{}: start time {} is after window start", threadName, startTime);
+        blocked += Duration.between(windowStart, startTime).toNanos();
+        observedStart = windowStart;
+      }
+
+      final Duration duration = Duration.between(observedStart, current.timestamp);
+      final double durationNs = duration.toNanos();
+      return Optional.of((durationNs - Math.min(blocked, durationNs)) / durationNs);
```
pbadani, line 344: @rodesai - Why did we take the difference between the time interval and blocked time to compute the saturation. Shouldn't this be just `Math.min(blocked, durationNs) / duration`

PR #7983, opened 2021-08-10
feat: Propose KLIP 53: Pull Queries on Streams

vvcephei:
Just proposing the included KLIP.

### DIFF
--- a/design-proposals/klip-53-pull-queries-on-streams.md
+++ b/design-proposals/klip-53-pull-queries-on-streams.md
@@ -1,0 +1,87 @@
+# KLIP 53 - Pull Queries on Streams
+
+**Author**: John Roesler (@vvcephei) | 
+**Release Target**: 0.22 | 
+**Status**: _In Discussion_ | 
+**Discussion**: _link to the design discussion PR_
+
+**tl;dr:** People should be able to issue pull (ad-hoc, self-terminating) queries on streams,
+           just as they can with tables.
+           
+## Motivation and background
+
+ksqlDB allows the creation of both pull queries (which are issued ad-hoc and run to completion like
+a typical database query) and push queries (which are long-running subscriptions that continuously).
+ksqlDB also has a dual table/stream data model. You can issue push queries on tables and streams, 
+and you can issue pull queries on tables, but you cannot currently issue pull queries on tables.
+
+Previously, we believed it was unnecessary to implement pull queries on streams because for a
+stream, the "current state" is equivalent to the entire history of records. In other words, a
+pull query on a stream would return the same thing as a push query on a stream, if it is configured
+to start at the beginning of the stream and if it's cancelled once it reaches the end.
+
+In practice, though, it is not clear when you have reached the end of a stream, so you cannot
+easily determine when to cancel the query. Additionally, it is cumbersome to have to configure
+push queries to start from the beginning of the stream each time you want to issue one of these
+pull-type queries.
+
+## What is in scope
+
+* Queries should be supported on STREAM objects without “EMIT CHANGES” at the end
+* Pull queries should start at the beginning of the stream
+* Pull queries should terminate at the (current) end of the stream
+* Pull queries on streams will support the same range of query operations as they
+  do on tables. In other words, you will be able to filter the stream using a WHERE clause.
+
+## What is not in scope
+
+* Pull queries on streams will not currently support grouping/aggregations.
+  This would be useful but is left as future work.
+  Note that aggregations are also not currently supported on table pull queries.
+
+## Value/Return
+
+After this feature, you will be able to more easily debug data flows, for example
+by inspecting whether a certain record is present in a stream. You will also be able
+to interpret a table's backing topic as a stream and look for specific events in its
+history. The ability to scan over a topic and apply complex filtering logic will also
+enable new use cases, such as looking for certain kinds of events.
+
+## Public APIS
+
+No new syntax is added. The only change is that queries on stream objects will no
+longer be required to end in `EMIT CHANGES`.
+
+## Design
+
+In order to support the full query expressiveness of ksqlDB, the current design
+is to simply treat pull queries on streams as syntactic sugar. Internally, we will:
+1. Create a typical push query
+2. Configure the query to start from "earliest" (the beginning of the stream)
+3. Find out the end offsets of the stream's partitions
+4. Start the push query
+5. Monitor the query's progress.
+   When it reaches or passes all the offsets in step 3, terminate the query.
+
+## Test plan
+
+We will update the existing tests that validate an error if a pull query is attempted
+on a stream. The logic will be updated to expect a valid response. We will also add new
+unit and integration tests to ensure the query produces the desired output.
+
+## Documentation Updates
+
+* We will add a new block to the quickstart.
+* We will update the docs to remove references
+  to the former restriction and document the new ability.
+* We should add a new example or amend some current examples
+  to show more complex versions than the quickstart does.
+
+## Compatibility Implications
+
+As this is only removing a restriction, no compatibility problems are anticipated.
+
+## Security Implications
+
+This is only adding a more convenient way to issue queries on data that is already
+queriable, therefore the existing access control mechanisms are sufficient.


design-proposals/klip-53-pull-queries-on-streams.md
```diff
@@ -0,0 +1,87 @@
...
+
+**tl;dr:** People should be able to issue pull (ad-hoc, self-terminating) queries on streams,
+           just as they can with tables.
+           
+## Motivation and background
+
+ksqlDB allows the creation of both pull queries (which are issued ad-hoc and run to completion like
+a typical database query) and push queries (which are long-running subscriptions that continuously).
+ksqlDB also has a dual table/stream data model. You can issue push queries on tables and streams, 
+and you can issue pull queries on tables, but you cannot currently issue pull queries on tables.
```
mjsax, line 16: > but you cannot currently issue pull queries on tables.

-> on streams?

design-proposals/klip-53-pull-queries-on-streams.md
```diff
@@ -0,0 +1,87 @@
...
+**Status**: _In Discussion_ | 
+**Discussion**: _link to the design discussion PR_
+
+**tl;dr:** People should be able to issue pull (ad-hoc, self-terminating) queries on streams,
+           just as they can with tables.
+           
+## Motivation and background
+
+ksqlDB allows the creation of both pull queries (which are issued ad-hoc and run to completion like
+a typical database query) and push queries (which are long-running subscriptions that continuously).
```
mjsax, line 14: `that continuously [run until expliclity terminated by the user].`

design-proposals/klip-53-pull-queries-on-streams.md
```diff
@@ -0,0 +1,87 @@
...
+In practice, though, it is not clear when you have reached the end of a stream, so you cannot
+easily determine when to cancel the query. Additionally, it is cumbersome to have to configure
+push queries to start from the beginning of the stream each time you want to issue one of these
+pull-type queries.
+
+## What is in scope
+
+* Queries should be supported on STREAM objects without “EMIT CHANGES” at the end
+* Pull queries should start at the beginning of the stream
+* Pull queries should terminate at the (current) end of the stream
```
mjsax, line 32: `(current)` seems fuzzy and should be explained/defined more precise.

Ie, the stream end-offsets at query start time defines the termination point of the query. Newly appended data after the query was started, is not included in the query.

design-proposals/klip-53-pull-queries-on-streams.md
```diff
@@ -0,0 +1,87 @@
...
+* Pull queries on streams will support the same range of query operations as they
+  do on tables. In other words, you will be able to filter the stream using a WHERE clause.
+
+## What is not in scope
+
+* Pull queries on streams will not currently support grouping/aggregations.
+  This would be useful but is left as future work.
+  Note that aggregations are also not currently supported on table pull queries.
```
mjsax, line 40: What about joins ?

mjsax (commented):
Very excited to see progress on this front! Overall LGTM. A few nits.

design-proposals/klip-53-pull-queries-on-streams.md
```diff
@@ -0,0 +1,87 @@
...
+
+## Test plan
+
+We will update the existing tests that validate an error if a pull query is attempted
+on a stream. The logic will be updated to expect a valid response. We will also add new
+unit and integration tests to ensure the query produces the desired output.
+
+## Documentation Updates
+
+* We will add a new block to the quickstart.
```
agavra, line 74: I don't think adding it to the quickstart is necessary - a single table pull query (which @patrickstuedi is adding) is probably sufficient to "get the point". There is value in keeping the quickstart as "quick" as possible, though happy to get @derekjn's thoughts here.

agavra (approved):
LGTM!

design-proposals/klip-53-pull-queries-on-streams.md
vvcephei: Oops. I somehow mixed up the most important word in the sentence.

design-proposals/klip-53-pull-queries-on-streams.md
vvcephei: Sounds good. I elaborated in the algorithm, but I agree it's worth specifying.

design-proposals/klip-53-pull-queries-on-streams.md
vvcephei: Joins are also not supported in pull queries. It's good to mention. Thanks!

design-proposals/klip-53-pull-queries-on-streams.md
vvcephei: Oy. Sorry about that.

design-proposals/klip-53-pull-queries-on-streams.md
vvcephei: Sounds good.

design-proposals/klip-53-pull-queries-on-streams.md
```diff
@@ -27,17 +28,26 @@ pull-type queries.
...
+* Pull queries on streams will not currently support grouping/aggregations or joins.
   This would be useful but is left as future work.
-  Note that aggregations are also not currently supported on table pull queries.
+  Note that neither aggregations nor joins are currently supported on table pull queries.
+* If old records have already been dropped from the stream due to retention
+  settings, they will not be included in the results.
+* The scan over the stream may or may not pick up some new records that arrive during the query
+  execution. The only guarantee is that the query will include all records that are already in
+  the stream at the start of the query. If desired, we can make the end point stricter in the
+  future.
```
mjsax, line 50: > may or may not
Why are we fuzzy here?

> If desired, we can make the end point stricter in the future.
Why not be strict from the beginning on?

design-proposals/klip-53-pull-queries-on-streams.md
```diff
@@ -27,17 +28,26 @@ pull-type queries.
...
 
 ## What is not in scope
 
-* Pull queries on streams will not currently support grouping/aggregations.
+* Pull queries on streams will not currently support grouping/aggregations or joins.
   This would be useful but is left as future work.
-  Note that aggregations are also not currently supported on table pull queries.
+  Note that neither aggregations nor joins are currently supported on table pull queries.
+* If old records have already been dropped from the stream due to retention
+  settings, they will not be included in the results.
```
mjsax, line 46: Think we can omit this bullet point. Seems to be obvious?

design-proposals/klip-53-pull-queries-on-streams.md
vvcephei: It's because of the way the code is currently set up. All we are doing is starting up a transient query and then terminating it when it passes the endpoint, but since there are multiple partitions involved, there are actually multiple endpoints, which won't all get passed at the same time.

It would be cool to add a feature to Streams where you tell it to process up to a particular offset in each partition, so that it would be able to strictly stop processing each task at the right point, but that would turn this into a much bigger project. This way, we can release the feature sooner and then circle back later and tighten up the endpoints if we want to.

Also, it doesn't seem like strictly not including any new records that arrive after the start of the query is really needed by any known use cases. Overwhelmingly, people just want to be able to have a convenient way to inspect what is in their streams. Having strict start and end points would enable new use cases, but the use cases we have today are better served by getting the feature released asap.

design-proposals/klip-53-pull-queries-on-streams.md
vvcephei: Thanks :). I also felt like it might be a little too obvious, but I went ahead and specified it in case people don't immediately put two and two together regarding the interaction of retention and their stream queries.

design-proposals/klip-53-pull-queries-on-streams.md
mjsax: Fair enough.

I agree that we might want to extend Kafka Streams. It reminds me on the old https://cwiki.apache.org/confluence/display/KAFKA/KIP-95%3A+Incremental+Batch+Processing+for+Kafka+Streams -- the KIP tried to include the "end offset discovery" what made very complex. If we would descope the KIP and let the user provide the end-offsets, it would be much simpler.

I guess for the current use case of "scan the full topic" it might be ok to be fuzzy on the end-point... If we want to allow to do "time range scans" over a stream, if "give me all record from 10am to 1pm yesterday" we will need stricter bounds...

mjsax (approved):

PR #7987, opened 2021-08-11
feat: add CumulativeSum total bytes metric

stevenpyzhang:
### Description 
New metric in consumer collector to get bytes consumed across all consumers

### Testing done 
Unit tests
Jconsole, verified the Engine metric and this new metric were showing equivalent values

### Reviewer checklist

### DIFF
--- a/ksqldb-common/src/main/java/io/confluent/ksql/metrics/ConsumerCollector.java
+++ b/ksqldb-common/src/main/java/io/confluent/ksql/metrics/ConsumerCollector.java
@@ -46,14 +46,19 @@ public class ConsumerCollector implements MetricCollector, ConsumerInterceptor<O
   public static final String CONSUMER_TOTAL_BYTES = "consumer-total-bytes";
   public static final String CONSUMER_ALL_TOTAL_BYTES_SUM = "consumer-all-total-bytes-sum";
   public static final String CONSUMER_COLLECTOR_METRICS_GROUP_NAME = "consumer-metrics";
+  private static final Sensor totalBytesSum;
 
   private final Map<String, TopicSensors<ConsumerRecord<Object, Object>>> topicSensors =
       new HashMap<>();
-  private Sensor totalBytesSum;
   private Metrics metrics;
   private String id;
   private String groupId;
   private Time time;
+
+  static
+  {
+    totalBytesSum = configureTotalBytesSum(MetricCollectors.getMetrics());
+  }
 
   public void configure(final Map<String, ?> map) {
     String id = (String) map.get(ConsumerConfig.GROUP_ID_CONFIG);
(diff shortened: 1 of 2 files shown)


ksqldb-engine/src/main/java/io/confluent/ksql/internal/KsqlEngineMetrics.java
```diff
@@ -171,8 +177,10 @@ private void recordMessagesConsumed(final double value) {
     this.messagesIn.record(value);
   }
 
-  private void recordTotalBytesConsumed(final double value) {
+  private void recordTotalBytesConsumedAndPerSecond(final double value) {
```
lct45, line 180: nit: the title feels kinda clunky, I know what you mean but it took me a second to understand. Maybe something like `recordTotalAndPerSecondBtyesConsumed`?

lct45 (approved):
lgtm

ksqldb-common/src/main/java/io/confluent/ksql/metrics/ConsumerCollector.java
```diff
@@ -196,4 +206,20 @@ public double aggregateStat(final String name, final boolean isError) {
...
+    sensor.add(
+        metrics.metricName(
+            CONSUMER_ALL_TOTAL_BYTES_SUM,
+            CONSUMER_COLLECTOR_METRICS_GROUP_NAME,
+            description),
+        new CumulativeSum());
+    return sensor;
+  }
+
+  protected Metrics getMetrics() {
```
guozhangwang, line 222: This function seems not used? Only MetricsCollectors.metrics() is needed elsewhere.

ksqldb-common/src/main/java/io/confluent/ksql/metrics/ConsumerCollector.java
```diff
@@ -105,6 +113,8 @@ private void record(
     topicSensors.computeIfAbsent(getCounterKey(topic), k ->
         new TopicSensors<>(topic, buildSensors(k))
     ).increment(record, isError);
+    totalBytesSum.record(record == null ? 0.0
```
guozhangwang, line 116: Curious why not just record as a long instead of a double?

ksqldb-common/src/main/java/io/confluent/ksql/metrics/ConsumerCollector.java
stevenpyzhang: I'll remove, this was used in a previous iteration of the PR

ksqldb-common/src/main/java/io/confluent/ksql/metrics/ConsumerCollector.java
stevenpyzhang: The KsqlEngineMetrics, totalBytesIn metric and the existing consumer total bytes metric were all Double so I just used that type also for this new metric

guozhangwang (approved):

PR #7988, opened 2021-08-11
feat: Allow scalable push queries to handle rebalances

AlanConfluent:
### Description 
Currently, we don't handle rebalances explicitly in SPQ.  If a new node joins the cluster or is removed from the cluster, we don't do anything.  In this PR, it attempts to start a new request to the node or expects the node to drop out of the cluster and tries to keep the connections ongoing, ensuring now data is missed.

### Testing done 
Ran unit tests.

### Reviewer checklist

### DIFF
--- a/ksqldb-engine/src/main/java/io/confluent/ksql/physical/scalablepush/PushRouting.java
+++ b/ksqldb-engine/src/main/java/io/confluent/ksql/physical/scalablepush/PushRouting.java
@@ -46,6 +46,7 @@ import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Map.Entry;
+import java.util.Optional;
 import java.util.Set;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.ConcurrentHashMap;
@@ -53,6 +54,7 @@ import java.util.concurrent.ExecutionException;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicReference;
 import java.util.function.Consumer;
+import java.util.function.Function;
 import java.util.stream.Collectors;
 import org.reactivestreams.Subscription;
 import org.slf4j.Logger;
@@ -64,18 +66,20 @@ public class PushRouting implements AutoCloseable {
 
   private static final long CLUSTER_CHECK_INTERVAL = 2000;
 
-  private LoadingCache<ScalablePushRegistry, Set<KsqlNode>> registryToNodes
-      = CacheBuilder.newBuilder()
-      .maximumSize(40)
-      .expireAfterWrite(2000, TimeUnit.MILLISECONDS)
-      .build(new CacheLoader<ScalablePushRegistry, Set<KsqlNode>>() {
-        @Override
-        public Set<KsqlNode> load(ScalablePushRegistry scalablePushRegistry) {
-          return loadCurrentHosts(scalablePushRegistry);
-        }
-      });
+  private final Function<ScalablePushRegistry, Set<KsqlNode>> registryToNodes;
+  private final long clusterCheckInterval;
 
   public PushRouting() {
+    this(createLoadingCache(), CLUSTER_CHECK_INTERVAL);
+  }
+
+  @VisibleForTesting
+  public PushRouting(
+      final Function<ScalablePushRegistry, Set<KsqlNode>> registryToNodes,
+      final long clusterCheckInterval
+      ) {
+    this.registryToNodes = registryToNodes;
+    this.clusterCheckInterval = clusterCheckInterval;
   }
 
   @Override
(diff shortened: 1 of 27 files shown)


ksqldb-engine/src/main/java/io/confluent/ksql/physical/scalablepush/PushRouting.java
```diff
@@ -238,13 +290,86 @@ public void close() {
...
+    }
+    final Set<KsqlNode> updatedHosts = registryToNodes.apply(
+        pushPhysicalPlan.getScalablePushRegistry());
+    final Set<KsqlNode> hosts = pushConnectionsHandle.getAllHosts();
+    final Set<KsqlNode> newHosts = Sets.difference(updatedHosts, hosts).stream()
+        .filter(node -> !node.isLocal())
+        .collect(Collectors.toSet());
+    final Set<KsqlNode> removedHosts = Sets.difference(hosts, updatedHosts);
+    if (newHosts.size() > 0) {
+      LOG.info("Dynamically adding new hosts {}", newHosts);
```
nateab, line 327: I tried following all long with the logic of finding new hosts but can't quite find exactly where you do it, I was expecting a similar call to `getRemoteHosts()`. Or is it just tracked by `pushConnectionsHandle`?

ksqldb-engine/src/main/java/io/confluent/ksql/physical/scalablepush/PushRouting.java
```diff
@@ -51,7 +64,22 @@
 
   private static final Logger LOG = LoggerFactory.getLogger(HARouting.class);
 
+  private static final long CLUSTER_CHECK_INTERVAL = 2000;
```
nateab, line 67: in milliseconds?

ksqldb-engine/src/main/java/io/confluent/ksql/physical/scalablepush/PushRouting.java
```diff
@@ -238,13 +290,86 @@ public void close() {
...
+  private static Set<KsqlNode> loadCurrentHosts(final ScalablePushRegistry scalablePushRegistry) {
+    return new HashSet<>(scalablePushRegistry
+        .getLocator()
+        .locate());
+  }
+
+  private static Function<ScalablePushRegistry, Set<KsqlNode>> createLoadingCache() {
+    final LoadingCache<ScalablePushRegistry, Set<KsqlNode>> cache = CacheBuilder.newBuilder()
+        .maximumSize(40)
+        .expireAfterWrite(2000, TimeUnit.MILLISECONDS)
```
agavra, line 353: is this meant to be `CLUSTER_CHECK_INTERVAL`?

ksqldb-engine/src/main/java/io/confluent/ksql/physical/scalablepush/PushRouting.java
```diff
@@ -238,13 +290,86 @@ public void close() {
...
+          public Set<KsqlNode> load(final ScalablePushRegistry scalablePushRegistry) {
+            return loadCurrentHosts(scalablePushRegistry);
+          }
+        });
+    return cache::getUnchecked;
+  }
+
   public enum RoutingResultStatus {
-    SUCCESS
+    IN_PROGRESS,
```
agavra, line 364: nit: javadocs for each of these could help :)

agavra (commented):
I skimmed the PR, it makes sense to me - though PushQueryRouting is getting a little too large for comfort

nateab:
> I skimmed the PR, it makes sense to me - though PushQueryRouting is getting a little too large for comfort

For my understanding, when does something get too large for comfort? When it's logic is trying to encompass too much and is better served by breaking it up more modularly?

agavra:
> when does something get too large for comfort

@nateab My barometer is usually when I have to pull up the IDE to effectively review the code, it's getting too large. There isn't a science to this, unfortunately, and everyone has different opinions.

nateab (approved):
LGTM! just left some comments

ksqldb-engine/src/main/java/io/confluent/ksql/physical/scalablepush/PushRouting.java
```diff
@@ -98,18 +132,34 @@ public void close() {
...
+          node, new RoutingResult(RoutingResultStatus.IN_PROGRESS, () -> { }));
+      final CompletableFuture<Void> callback = new CompletableFuture<>();
+      callback.handle((v, t) -> {
+        if (t == null) {
+          pushConnectionsHandle.get(node)
+              .ifPresent(result -> {
+                result.close();
+                result.updateStatus(RoutingResultStatus.COMPLETE);
+              });
+          LOG.info("Host {} completed request.", node);
```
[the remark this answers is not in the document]
AlanConfluent: Yes, I added pushPhysicalPlan.getQueryId() to most of the logging calls to make it a bit easier to understand, though we don't log every request because it could result in a lot of logging.  They probably happen less often than pull queries, but we want them to be that lightweight.

I still think there's some value having semi exceptional or uncommon things in the logs to at least know something even if we don't know how to tie it to an exact query.  This usually should only happen if a node is shut down which doesn't happen often.

ksqldb-engine/src/main/java/io/confluent/ksql/physical/scalablepush/PushRouting.java
AlanConfluent: Yeah, it probably should be.  We want them to expire at least that often.  I'll actually create a constant for this.

ksqldb-engine/src/main/java/io/confluent/ksql/physical/scalablepush/PushRouting.java
AlanConfluent: Done, added a bunch of javadoc.

PR #7991, opened 2021-08-11
refactor: pull query metrics

vvcephei:
### Description 

Refactor pull query metrics to remove the need for confusing "UNKNOWN" enum values on
the various query types.

### Testing done 

Internal refactor. Existing tests should be sufficient.

### Reviewer checklist

### DIFF
--- a/ksqldb-engine/src/main/java/io/confluent/ksql/engine/EngineExecutor.java
+++ b/ksqldb-engine/src/main/java/io/confluent/ksql/engine/EngineExecutor.java
@@ -39,8 +39,6 @@ import io.confluent.ksql.parser.tree.Statement;
 import io.confluent.ksql.physical.PhysicalPlan;
 import io.confluent.ksql.physical.pull.HARouting;
 import io.confluent.ksql.physical.pull.PullPhysicalPlan;
-import io.confluent.ksql.physical.pull.PullPhysicalPlan.PullPhysicalPlanType;
-import io.confluent.ksql.physical.pull.PullPhysicalPlan.PullSourceType;
 import io.confluent.ksql.physical.pull.PullPhysicalPlan.RoutingNodeType;
 import io.confluent.ksql.physical.pull.PullPhysicalPlanBuilder;
 import io.confluent.ksql.physical.pull.PullQueryQueuePopulator;
@@ -176,9 +174,10 @@ final class EngineExecutor {
       throw new IllegalArgumentException("Executor can only handle pull queries");
     }
     final SessionConfig sessionConfig = statement.getSessionConfig();
-    PullSourceType sourceType = null;
-    PullPhysicalPlanType planType = null;
-    RoutingNodeType routingNodeType = null;
+    final RoutingNodeType routingNodeType = routingOptions.getIsSkipForwardRequest()
+        ? RoutingNodeType.REMOTE_NODE : RoutingNodeType.SOURCE_NODE;
+
+    PullPhysicalPlan plan = null;
 
     try {
       // Do not set sessionConfig.getConfig to true! The copying is inefficient and slows down pull
@@ -187,16 +186,16 @@ final class EngineExecutor {
       final KsqlConfig ksqlConfig = sessionConfig.getConfig(false);
       final LogicalPlanNode logicalPlan = buildAndValidateLogicalPlan(
           statement, analysis, ksqlConfig, queryPlannerOptions, false);
-      final PullPhysicalPlan physicalPlan = buildPullPhysicalPlan(
+
+      plan = buildPullPhysicalPlan(
           logicalPlan,
           analysis
       );
-      sourceType = physicalPlan.getSourceType();
-      planType = physicalPlan.getPlanType();
+      final PullPhysicalPlan physicalPlan = plan;
+
       // If we ever change how many hops a request can do, we'll need to update this for correct
       // metrics.
-      routingNodeType = routingOptions.getIsSkipForwardRequest()
-          ? RoutingNodeType.REMOTE_NODE : RoutingNodeType.SOURCE_NODE;
+
       final PullQueryQueue pullQueryQueue = new PullQueryQueue();
       final PullQueryQueuePopulator populator = () -> routing.handlePullQuery(
           serviceContext,
@@ -210,13 +209,16 @@ final class EngineExecutor {
       }
       return result;
     } catch (final Exception e) {
-      final PullSourceType finalSourceType = sourceType;
-      final PullPhysicalPlanType finalPlanType = planType;
-      final RoutingNodeType finalRoutingNodeType = routingNodeType;
-      pullQueryMetrics.ifPresent(metrics -> metrics.recordErrorRate(1,
-          Optional.ofNullable(finalSourceType).orElse(PullSourceType.UNKNOWN),
-          Optional.ofNullable(finalPlanType).orElse(PullPhysicalPlanType.UNKNOWN),
-          Optional.ofNullable(finalRoutingNodeType).orElse(RoutingNodeType.UNKNOWN)));
+      if (plan == null) {
+        pullQueryMetrics.ifPresent(m -> m.recordErrorRateForNoResult(1));
+      } else {
+        final PullPhysicalPlan physicalPlan = plan;
+        pullQueryMetrics.ifPresent(metrics -> metrics.recordErrorRate(1,
+            physicalPlan.getSourceType(),
+            physicalPlan.getPlanType(),
+            routingNodeType
+        ));
+      }
 
       final String stmtLower = statement.getStatementText().toLowerCase(Locale.ROOT);
       final String messageLower = e.getMessage().toLowerCase(Locale.ROOT);
--- a/ksqldb-engine/src/main/java/io/confluent/ksql/internal/PullQueryExecutorMetrics.java
+++ b/ksqldb-engine/src/main/java/io/confluent/ksql/internal/PullQueryExecutorMetrics.java
@@ -741,9 +741,9 @@ public class PullQueryExecutorMetrics implements Closeable {
         final PullPhysicalPlanType planType,
         final RoutingNodeType routingNodeType
     ) {
-      this.sourceType = sourceType;
-      this.planType = planType;
-      this.routingNodeType = routingNodeType;
+      this.sourceType = Objects.requireNonNull(sourceType, "sourceType");
+      this.planType = Objects.requireNonNull(planType, "planType");
+      this.routingNodeType = Objects.requireNonNull(routingNodeType, "routingNodeType");
     }
 
     @Override
--- a/ksqldb-engine/src/main/java/io/confluent/ksql/internal/PullQueryExecutorMetrics.java
+++ b/ksqldb-engine/src/main/java/io/confluent/ksql/internal/PullQueryExecutorMetrics.java
@@ -16,6 +16,7 @@
 package io.confluent.ksql.internal;
 
 import com.google.common.collect.ImmutableMap;
+import com.google.common.collect.ImmutableMap.Builder;
 import edu.umd.cs.findbugs.annotations.SuppressFBWarnings;
 import io.confluent.ksql.metrics.MetricCollectors;
 import io.confluent.ksql.physical.pull.PullPhysicalPlan.PullPhysicalPlanType;
@@ -129,12 +130,21 @@ public class PullQueryExecutorMetrics implements Closeable {
       final PullPhysicalPlanType planType,
       final RoutingNodeType routingNodeType
   ) {
+    final MetricsKey key = new MetricsKey(sourceType, planType, routingNodeType);
+    innerRecordLatency(startTimeNanos, key);
+  }
+
+  public void recordLatencyForError(final long startTimeNanos) {
+    final MetricsKey key = new MetricsKey();
+    innerRecordLatency(startTimeNanos, key);
+  }
+
+  private void innerRecordLatency(final long startTimeNanos, final MetricsKey key) {
     // Record latency at microsecond scale
     final long nowNanos = time.nanoseconds();
     final double latency = TimeUnit.NANOSECONDS.toMicros(nowNanos - startTimeNanos);
     this.latencySensor.record(latency);
     this.requestRateSensor.record(1);
-    final MetricsKey key = new MetricsKey(sourceType, planType, routingNodeType);
     if (latencySensorMap.containsKey(key)) {
       latencySensorMap.get(key).record(latency);
     } else {
@@ -157,6 +167,16 @@ public class PullQueryExecutorMetrics implements Closeable {
     }
   }
 
+  public void recordErrorRateForNoResult(final double value) {
+    this.errorRateSensor.record(value);
+    final MetricsKey key = new MetricsKey();
+    if (errorRateSensorMap.containsKey(key)) {
+      errorRateSensorMap.get(key).record(value);
+    } else {
+      throw new IllegalStateException("Metrics not configured correctly, missing " + key);
+    }
+  }
+
   public void recordRequestSize(final double value) {
     this.requestSizeSensor.record(value);
   }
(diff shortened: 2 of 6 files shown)


ksqldb-engine/src/main/java/io/confluent/ksql/engine/EngineExecutor.java
```diff
@@ -176,9 +176,10 @@ PullQueryResult executeTablePullQuery(
       throw new IllegalArgumentException("Executor can only handle pull queries");
     }
     final SessionConfig sessionConfig = statement.getSessionConfig();
-    PullSourceType sourceType = null;
-    PullPhysicalPlanType planType = null;
-    RoutingNodeType routingNodeType = null;
+    final RoutingNodeType routingNodeType = routingOptions.getIsSkipForwardRequest()
+        ? RoutingNodeType.REMOTE_NODE : RoutingNodeType.SOURCE_NODE;
```
vvcephei, line 178: pulled this up because it cannot throw an exception and doesn't depend on the plan.

ksqldb-engine/src/main/java/io/confluent/ksql/engine/EngineExecutor.java
```diff
@@ -187,16 +188,16 @@ PullQueryResult executeTablePullQuery(
...
       final LogicalPlanNode logicalPlan = buildAndValidateLogicalPlan(
           statement, analysis, ksqlConfig, queryPlannerOptions, false);
-      final PullPhysicalPlan physicalPlan = buildPullPhysicalPlan(
+
+      plan = buildPullPhysicalPlan(
           logicalPlan,
           analysis
       );
-      sourceType = physicalPlan.getSourceType();
-      planType = physicalPlan.getPlanType();
```
vvcephei, line 195: These are either both or neither present, so I just set the whole `plan` instead.

ksqldb-engine/src/main/java/io/confluent/ksql/engine/EngineExecutor.java
```diff
@@ -210,13 +209,16 @@ PullQueryResult executeTablePullQuery(
       }
       return result;
     } catch (final Exception e) {
-      final PullSourceType finalSourceType = sourceType;
-      final PullPhysicalPlanType finalPlanType = planType;
-      final RoutingNodeType finalRoutingNodeType = routingNodeType;
-      pullQueryMetrics.ifPresent(metrics -> metrics.recordErrorRate(1,
-          Optional.ofNullable(finalSourceType).orElse(PullSourceType.UNKNOWN),
-          Optional.ofNullable(finalPlanType).orElse(PullPhysicalPlanType.UNKNOWN),
-          Optional.ofNullable(finalRoutingNodeType).orElse(RoutingNodeType.UNKNOWN)));
```
vvcephei, line 219: The main purpose of this PR was to get rid of the XYType `UNKNOWN` values because they only exist for statements like these. The thing to note is that they are either all known or none known, so we don't need a per-field null-check and a per-type "unknown" element.

As a newcomer to the code base, I found myself really confused as to how the could possibly not know the type for a query. The answer is if the planning fails for the query, then there is no query and therefore no known type. However, if planning fails for the query, we also don't know anything else about it, so it's more straightforward to reason about it that way than on a per-field basis.

ksqldb-engine/src/main/java/io/confluent/ksql/engine/EngineExecutor.java
```diff
@@ -210,13 +209,16 @@ PullQueryResult executeTablePullQuery(
...
+      if (plan == null) {
+        pullQueryMetrics.ifPresent(m -> m.recordErrorRateForNoResult(1));
+      } else {
+        final PullPhysicalPlan physicalPlan = plan;
```
guozhangwang, line 215: Hmm why we need to re-assign to a different variable?

ksqldb-engine/src/main/java/io/confluent/ksql/engine/EngineExecutor.java
```diff
@@ -187,16 +186,16 @@ PullQueryResult executeTablePullQuery(
...
           statement, analysis, ksqlConfig, queryPlannerOptions, false);
-      final PullPhysicalPlan physicalPlan = buildPullPhysicalPlan(
+
+      plan = buildPullPhysicalPlan(
           logicalPlan,
           analysis
       );
-      sourceType = physicalPlan.getSourceType();
-      planType = physicalPlan.getPlanType();
+      final PullPhysicalPlan physicalPlan = plan;
```
guozhangwang, line 194: Same here, maybe we can just rename the variable at line 180 as `PullPhysicalPlan physicalPlan`?

ksqldb-engine/src/main/java/io/confluent/ksql/internal/PullQueryExecutorMetrics.java
```diff
@@ -129,12 +130,21 @@ public void recordLatency(
...
       final RoutingNodeType routingNodeType
   ) {
+    final MetricsKey key = new MetricsKey(sourceType, planType, routingNodeType);
+    innerRecordLatency(startTimeNanos, key);
+  }
+
+  public void recordLatencyForError(final long startTimeNanos) {
+    final MetricsKey key = new MetricsKey();
+    innerRecordLatency(startTimeNanos, key);
+  }
```
vvcephei, line 140: Added these methods to make it clear when we know all the metric dimensions or none of them.

ksqldb-engine/src/main/java/io/confluent/ksql/internal/PullQueryExecutorMetrics.java
```diff
@@ -202,6 +232,15 @@ public void recordRowsReturned(
     }
   }
 
+  public void recordZeroRowsReturnedForError() {
```
vvcephei, line 235: I also hard-coded the "zero", since that's what we would always record for errors.

ksqldb-engine/src/main/java/io/confluent/ksql/internal/PullQueryExecutorMetrics.java
```diff
@@ -699,5 +773,31 @@ public String toString() {
...
+      return getName(routingNodeType);
+    }
+
+    private String getName(final Enum<?> o) {
+      if (o == null) {
+        return "unknown";
+      } else {
+        return o.name().toLowerCase();
+      }
+    }
```
vvcephei, line 801: Consolidated the logic for getting the metric-friendly version of the enum values. This inlines the logic for unknown values, which was previously split between the enum classes, which each had an "UNKNOWN" value, and the metric logic, which would take the lowercase version of the word as the metric dimension.

ksqldb-engine/src/main/java/io/confluent/ksql/physical/pull/PullPhysicalPlan.java
```diff
@@ -172,8 +172,7 @@ public QueryId getQueryId() {
   public enum PullPhysicalPlanType {
     // Could be one or more keys
     KEY_LOOKUP,
-    TABLE_SCAN,
-    UNKNOWN
```
vvcephei, line 176: In any case where we actually have a plan, we always know the kind of plan it is.

ksqldb-engine/src/main/java/io/confluent/ksql/physical/pull/PullPhysicalPlan.java
```diff
@@ -182,8 +181,7 @@ public QueryId getQueryId() {
    */
   public enum PullSourceType {
     NON_WINDOWED,
-    WINDOWED,
-    UNKNOWN
```
vvcephei, line 186: Similarly, if we actually have a plan, we know whether or not it's a windowed source.

ksqldb-engine/src/main/java/io/confluent/ksql/physical/pull/PullPhysicalPlan.java
```diff
@@ -192,7 +190,6 @@ public QueryId getQueryId() {
    */
   public enum RoutingNodeType {
     SOURCE_NODE,
-    REMOTE_NODE,
-    UNKNOWN
```
vvcephei, line 196: And finally, ksql is never uncertain whether it's answering a query for a user or another ksql node.

ksqldb-engine/src/main/java/io/confluent/ksql/internal/PullQueryExecutorMetrics.java
```diff
@@ -620,38 +668,54 @@ private void addSensor(
...
+        PULL_QUERY_METRIC_GROUP + "-"
+            + PULL_REQUESTS + "-"
+            + sensorBaseName + "-"
+            + variantName);
+
+    final ImmutableMap<String, String> tags = ImmutableMap.<String, String>builder()
+        .putAll(customMetricsTags)
+        .put(KsqlConstants.KSQL_QUERY_SOURCE_TAG, metricsKey.sourceTypeName())
+        .put(KsqlConstants.KSQL_QUERY_PLAN_TYPE_TAG, metricsKey.planTypeName())
+        .put(KsqlConstants.KSQL_QUERY_ROUTING_TYPE_TAG, metricsKey.routingNodeTypeName())
```
guozhangwang, line 705: `.toLowerCase()` missing?

ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/resources/streaming/PullQueryPublisher.java
```diff
@@ -141,27 +144,30 @@ public synchronized void subscribe(final Subscriber<Collection<StreamedRow>> sub
...
 
   private void recordMetrics(
-      final PullQueryExecutorMetrics metrics, final Optional<PullQueryResult> result) {
-    final PullSourceType sourceType = result.map(
-        PullQueryResult::getSourceType).orElse(PullSourceType.UNKNOWN);
-    final PullPhysicalPlanType planType = result.map(
-        PullQueryResult::getPlanType).orElse(PullPhysicalPlanType.UNKNOWN);
-    final RoutingNodeType routingNodeType = result.map(
-        PullQueryResult::getRoutingNodeType).orElse(RoutingNodeType.UNKNOWN);
+      final PullQueryExecutorMetrics metrics, final PullQueryResult result) {
```
vvcephei, line 154: I noticed that we're not recording the response size here... Is that intentional? (cc @cprasad1  )

ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/resources/streaming/PullQueryPublisher.java
```diff
@@ -124,12 +124,15 @@ public synchronized void subscribe(final Subscriber<Collection<StreamedRow>> sub
           true
       );
 
+      if (pullQueryMetrics.isPresent()) {
```
guozhangwang, line 127: This seems duplicated with line 134-136?

ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/resources/streaming/PullQueryPublisher.java
guozhangwang, line 154: Mis-alignment.

ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/resources/streaming/StreamedQueryResource.java
```diff
@@ -333,20 +333,34 @@ private EndpointResponse handleQuery(final KsqlSecurityContext securityContext,
...
-                metrics.recordLatency(startTimeNanos, sourceType, planType, routingNodeType);
-                metrics.recordRowsReturned(
-                    Optional.ofNullable(r).map(PullQueryResult::getTotalRowsReturned).orElse(0L),
-                    sourceType, planType, routingNodeType);
-                metrics.recordRowsProcessed(
-                    Optional.ofNullable(r).map(PullQueryResult::getTotalRowsProcessed).orElse(0L),
-                    sourceType, planType, routingNodeType);
+                if (r == null) {
+                  metrics.recordResponseSizeForError(responseBytes);
+                  metrics.recordLatencyForError(startTimeNanos);
```
guozhangwang, line 338: Can use `recordErrorMetrics`?

guozhangwang (commented):
LGTM only minor comments, pleas feel free to merge afterwards.

guozhangwang (approved):

ksqldb-engine/src/main/java/io/confluent/ksql/engine/EngineExecutor.java
vvcephei: Yeah, it's a pain. It's because you can only reference final variables from inside a callback. The reason we have a non-final variable involved is that we want to reference the plan in the `catch` block _if_ the plan happens to have gotten built.

I started to refactor the method so this wouldn't be necessary, but it's not trivial, and believe it or not, I showed some restraint.

ksqldb-engine/src/main/java/io/confluent/ksql/engine/EngineExecutor.java
vvcephei: Yep, it's a bit mysterious, but the only reason we need this final reference is for the callback below. The Java authors thought it would make user code safer if they only let you reference final variables.

ksqldb-engine/src/main/java/io/confluent/ksql/internal/PullQueryExecutorMetrics.java
vvcephei: It is now encapsulated inside the xYName methods.

ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/resources/streaming/PullQueryPublisher.java
vvcephei: Oops! This was a partial refactor I overlooked. Good catch!

ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/resources/streaming/StreamedQueryResource.java
vvcephei: Good eye! That's a method in PullQueryPublisher, so we can't use it here. I did think about pulling these common sets of metric-recording operations into the PullQueryMetrics class, though.

ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/resources/streaming/PullQueryPublisher.java
AlanConfluent: Yes, it is.   This is the web socket interface and we couldn't easily figure out how to report response size, unfortunately.  So we don't have complete metrics for this interface, though it's not used outsize of the UI.

AlanConfluent (approved):
I'm not sure I feel as enthusiastically about getting rid of the structured reference to unknown, but I guess it's been replaced by the use of `MetricsKey()` which is easy to search for.  So long as we know where it's possible to get unknown and can do an easy code search, I'm fine with it.

ksqldb-engine/src/main/java/io/confluent/ksql/internal/PullQueryExecutorMetrics.java
vvcephei: Thanks. Just to make sure we're on the same page, I'm just preserving and making explicit the existing logic. I did think it would make more sense not to report these metrics in error cases, but I just left it alone.

If you think it's fine (and maybe better) not to report these metrics at all for errors, we can clean things up even more.

ksqldb-engine/src/main/java/io/confluent/ksql/internal/PullQueryExecutorMetrics.java
```diff
@@ -699,5 +773,31 @@ public String toString() {
...
+      return getName(planType);
+    }
+
+    public String routingNodeTypeName() {
+      return getName(routingNodeType);
+    }
+
+    private String getName(final Enum<?> o) {
+      if (o == null) {
+        return "unknown";
```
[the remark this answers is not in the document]
vvcephei: That's a good idea. I specifically avoided adding nullable references to the enum values as a way to get away from "unknown" because, I agree, that's a strictly worse way to do the same thing. I really wanted to capture that it's not that a specific attribute of the query/result can be unknown, but rather that we want to register metrics some times in error cases when we just don't know what kind of query was run, for example because we couldn't parse it.

ksqldb-engine/src/main/java/io/confluent/ksql/internal/PullQueryExecutorMetrics.java
```diff
@@ -662,6 +726,16 @@ private void addSensor(
...
+     * API call that didn't have a result (because it had an error instead)
+     */
+    MetricsKey() {
+      this.sourceType = null;
+      this.planType = null;
+      this.routingNodeType = null;
+    }
+
     MetricsKey(
         final PullSourceType sourceType,
```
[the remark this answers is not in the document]
vvcephei: Right, and yes, this is a good idea.

PR #7992, opened 2021-08-11
fix: Fixes pull query latency distribution metrics

AlanConfluent:
### Description 
Fixes the distribution metrics since they were not setup correctly.

Previously, the max value was set to be 1000 microseconds, which is much less than the single digit milliseconds we generally get.  I also changed it to be a "linear" bucketizer so as to have finer grained smaller metrics, even though the max value is ten seconds.

### Testing done 
Manual and unit testing.

### Reviewer checklist

### DIFF
--- a/ksqldb-engine/src/main/java/io/confluent/ksql/internal/PullQueryExecutorMetrics.java
+++ b/ksqldb-engine/src/main/java/io/confluent/ksql/internal/PullQueryExecutorMetrics.java
@@ -50,6 +50,8 @@ public class PullQueryExecutorMetrics implements Closeable {
 
   private static final String PULL_QUERY_METRIC_GROUP = "pull-query";
   private static final String PULL_REQUESTS = "pull-query-requests";
+  private static final long MAX_LATENCY_BUCKET_VALUE_MICROS = TimeUnit.SECONDS.toMicros(10);
+  private static final int NUM_LATENCY_BUCKETS = 1000;
 
   private final List<Sensor> sensors;
   private final Sensor localRequestsSensor;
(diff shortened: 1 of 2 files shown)


ksqldb-engine/src/main/java/io/confluent/ksql/internal/PullQueryExecutorMetrics.java
```diff
@@ -50,6 +50,8 @@
 
   private static final String PULL_QUERY_METRIC_GROUP = "pull-query";
   private static final String PULL_REQUESTS = "pull-query-requests";
+  private static final long MAX_LATENCY_BUCKET_VALUE_MICROS = TimeUnit.SECONDS.toMicros(10);
```
agavra, line 53: are we ever expecting to be in sub-millisecond latency buckets? why not just record at the ms level? That would give us better granularities around the part of the spectrum we care about

agavra (commented):
Great catch, though I'm wondering if starting at the microsecond granularity really makes sense for us.

vvcephei (approved):
Thanks for the fix, @AlanConfluent !

I'd be fine with measuring in millis, as @agavra suggested, but I'm also fine with micros. Offhand, I think most database latency metrics I've seen are measured in micros, I think precisely because you'd hope to have response times in the single-digit millis, at least for some queries.

ksqldb-engine/src/main/java/io/confluent/ksql/internal/PullQueryExecutorMetrics.java
guozhangwang: I think I agree: if we think any latency smaller than 1ms is considered efficient enough and no need to worry then it seems just millis are sufficient (in which case any latency under 1ms would just be considered 1ms). 

On the other hand that also means we consider the pull query throughput would not exceed 1000/s in common cases. For materialized view / cache use cases I think that is fine, for analytics it should be even infrequent.

AlanConfluent: I was a bit misleading above when I said that we generally get tens of milliseconds.  Locally for key lookups, I get a p50 of anywhere from 900 to 1100 micros and the p90 is 2000 micros.  For table scans, I get tens of milliseconds for a 10,000 entry table.  With the linear bucketizer, I'm hoping we still get pretty decent granularity here anyway.  If I switched to ms, then key lookups would just be 1 or 2 most of the time, which is maybe ok for our purposes, but seems like we can get a little more detail.

I think that to preserve granularity for key lookups, I'll leave it in micros, if that works for everyone.

Issue #6105, opened 2020-08-26
Pull Queries: Range query support in pull-queries


agavra:
Similar to #5850, but optimized for the scenario where the predicate is a range query on the key field.

koolay:
Has it  already been removed?

agavra:
Update - table scans are now a proxy for this: you can issue a range query but it will result in a full table scan. The scope of this ticket is now to leverage the key index to make it more performant.

Taymindis:
is this enhancement ready ? @patrickstuedi @agavra

PR #7993, opened 2021-08-12
Optimize key-range queries in pull queries (#6105)

patrickstuedi:
### Description 
Use range interface in state store for range pull queries, instead of doing a table scan

### Testing done 
no testing yet

### Reviewer checklist

### DIFF
--- a/ksqldb-engine/src/main/java/io/confluent/ksql/physical/pull/PullPhysicalPlan.java
+++ b/ksqldb-engine/src/main/java/io/confluent/ksql/physical/pull/PullPhysicalPlan.java
@@ -23,11 +23,11 @@ import io.confluent.ksql.execution.streams.materialization.Materialization;
 import io.confluent.ksql.physical.common.operators.AbstractPhysicalOperator;
 import io.confluent.ksql.physical.pull.operators.DataSourceOperator;
 import io.confluent.ksql.planner.plan.KeyConstraint;
-import io.confluent.ksql.planner.plan.KeyConstraint.ConstraintOperator;
 import io.confluent.ksql.planner.plan.LookupConstraint;
 import io.confluent.ksql.query.PullQueryQueue;
 import io.confluent.ksql.query.QueryId;
 import io.confluent.ksql.schema.ksql.LogicalSchema;
+import java.util.ArrayList;
 import java.util.Collections;
 import java.util.List;
 import java.util.Objects;
@@ -123,26 +123,17 @@ public class PullPhysicalPlan {
   }
 
   public List<KsqlKey> getKeys() {
-    if (requiresRequestsToAllPartitions()) {
-      return Collections.emptyList();
+    final List<KsqlKey> list = new ArrayList<>();
+    for (LookupConstraint c : lookupConstraints) {
+      if (c instanceof KeyConstraint) {
+        final KeyConstraint kc = (KeyConstraint) c;
+        list.add(kc.getKsqlKey());
+      } else {
+        //we shouldn't see any NonKeyContraints here
+        return Collections.emptyList();
+      }
     }
-    return lookupConstraints.stream()
-        .filter(lookupConstraint -> lookupConstraint instanceof KeyConstraint)
-        .map(KeyConstraint.class::cast)
-        .filter(keyConstraint -> keyConstraint.getConstraintOperator() == ConstraintOperator.EQUAL)
-        .map(KeyConstraint::getKsqlKey)
-        .collect(ImmutableList.toImmutableList());
-  }
-
-  private boolean requiresRequestsToAllPartitions() {
-    return lookupConstraints.stream()
-        .anyMatch(lookupConstraint -> {
-          if (lookupConstraint instanceof KeyConstraint) {
-            final KeyConstraint keyConstraint = (KeyConstraint) lookupConstraint;
-            return keyConstraint.getConstraintOperator() != ConstraintOperator.EQUAL;
-          }
-          return true;
-        });
+    return ImmutableList.copyOf(list);
   }
 
   public LogicalSchema getOutputSchema() {
@@ -172,6 +163,7 @@ public class PullPhysicalPlan {
   public enum PullPhysicalPlanType {
     // Could be one or more keys
     KEY_LOOKUP,
+    RANGE_SCAN,
     TABLE_SCAN
   }
 
--- a/ksqldb-engine/src/main/java/io/confluent/ksql/physical/pull/PullPhysicalPlanBuilder.java
+++ b/ksqldb-engine/src/main/java/io/confluent/ksql/physical/pull/PullPhysicalPlanBuilder.java
@@ -36,7 +36,10 @@ import io.confluent.ksql.physical.pull.operators.KeyedWindowedTableLookupOperato
 import io.confluent.ksql.physical.pull.operators.TableScanOperator;
 import io.confluent.ksql.physical.pull.operators.WindowedTableScanOperator;
 import io.confluent.ksql.planner.LogicalPlanNode;
+import io.confluent.ksql.planner.QueryPlannerOptions;
 import io.confluent.ksql.planner.plan.DataSourceNode;
+import io.confluent.ksql.planner.plan.KeyConstraint;
+import io.confluent.ksql.planner.plan.KeyConstraint.ConstraintOperator;
 import io.confluent.ksql.planner.plan.KsqlBareOutputNode;
 import io.confluent.ksql.planner.plan.LookupConstraint;
 import io.confluent.ksql.planner.plan.NonKeyConstraint;
@@ -66,6 +69,7 @@ public class PullPhysicalPlanBuilder {
   private final PersistentQueryMetadata persistentQueryMetadata;
   private final QueryId queryId;
   private final Materialization mat;
+  private final QueryPlannerOptions queryPlannerOptions;
 
   private List<LookupConstraint> lookupConstraints;
   private PullPhysicalPlanType pullPhysicalPlanType;
@@ -75,7 +79,8 @@ public class PullPhysicalPlanBuilder {
   public PullPhysicalPlanBuilder(
       final ProcessingLogContext processingLogContext,
       final PersistentQueryMetadata persistentQueryMetadata,
-      final ImmutableAnalysis analysis
+      final ImmutableAnalysis analysis,
+      final QueryPlannerOptions queryPlannerOptions
   ) {
     this.processingLogContext = Objects.requireNonNull(
         processingLogContext, "processingLogContext");
@@ -86,6 +91,7 @@ public class PullPhysicalPlanBuilder {
     mat = this.persistentQueryMetadata
         .getMaterialization(queryId, contextStacker)
         .orElseThrow(() -> notMaterializedException(getSourceName(analysis)));
+    this.queryPlannerOptions = queryPlannerOptions;
   }
 
   /**
@@ -180,27 +186,53 @@ public class PullPhysicalPlanBuilder {
     return new SelectOperator(logicalNode, logger);
   }
 
+  private PullPhysicalPlanType getPlanType() {
+    if (!seenSelectOperator) {
+      lookupConstraints = Collections.emptyList();
+      return PullPhysicalPlanType.TABLE_SCAN;
+    } else if (lookupConstraints.stream().anyMatch(lc -> lc instanceof NonKeyConstraint)) {
+      lookupConstraints = Collections.emptyList();
+      return PullPhysicalPlanType.TABLE_SCAN;
+    } else if (lookupConstraints.stream().allMatch(lc -> ((KeyConstraint) lc).getOperator()
+        == ConstraintOperator.EQUAL)) {
+      return PullPhysicalPlanType.KEY_LOOKUP;
+    } else if (lookupConstraints.size() == 1
+        && lookupConstraints.stream().allMatch(lc -> ((KeyConstraint) lc).getOperator()
+        != ConstraintOperator.EQUAL)) {
+      return PullPhysicalPlanType.RANGE_SCAN;
+    } else {
+      lookupConstraints = Collections.emptyList();
+      return PullPhysicalPlanType.TABLE_SCAN;
+    }
+  }
+
   private AbstractPhysicalOperator translateDataSourceNode(
       final DataSourceNode logicalNode
   ) {
-    boolean isTableScan = false;
-    if (!seenSelectOperator) {
-      lookupConstraints = Collections.emptyList();
-      isTableScan = true;
-    } else if (lookupConstraints.stream().anyMatch(lc -> lc instanceof NonKeyConstraint)) {
-      isTableScan = true;
+    pullPhysicalPlanType = getPlanType();
+    if (pullPhysicalPlanType == PullPhysicalPlanType.RANGE_SCAN
+        && (!queryPlannerOptions.getRangeScansEnabled() || logicalNode.isWindowed())) {
+      pullPhysicalPlanType = PullPhysicalPlanType.TABLE_SCAN;
     }
+    if (pullPhysicalPlanType == PullPhysicalPlanType.TABLE_SCAN) {
+      if (queryPlannerOptions.getTableScansEnabled()) {
+        lookupConstraints = Collections.emptyList();
+      } else {
+        throw new KsqlException("Query requires table scan to be enabled. Table scans can be"
+          + " enabled by setting ksql.query.pull.table.scan.enabled=true");
+      }
+    }
+
     pullSourceType = logicalNode.isWindowed()
         ? PullSourceType.WINDOWED : PullSourceType.NON_WINDOWED;
-    if (isTableScan) {
-      pullPhysicalPlanType = PullPhysicalPlanType.TABLE_SCAN;
+    if (pullPhysicalPlanType == PullPhysicalPlanType.TABLE_SCAN) {
       if (!logicalNode.isWindowed()) {
         return new TableScanOperator(mat, logicalNode);
       } else {
         return new WindowedTableScanOperator(mat, logicalNode);
       }
     }
-    pullPhysicalPlanType = PullPhysicalPlanType.KEY_LOOKUP;
+
     if (!logicalNode.isWindowed()) {
       return new KeyedTableLookupOperator(mat, logicalNode);
     } else {
--- a/ksqldb-engine/src/main/java/io/confluent/ksql/physical/pull/operators/KeyedTableLookupOperator.java
+++ b/ksqldb-engine/src/main/java/io/confluent/ksql/physical/pull/operators/KeyedTableLookupOperator.java
@@ -25,6 +25,8 @@ import io.confluent.ksql.execution.streams.materialization.Row;
 import io.confluent.ksql.physical.common.operators.AbstractPhysicalOperator;
 import io.confluent.ksql.physical.common.operators.UnaryPhysicalOperator;
 import io.confluent.ksql.planner.plan.DataSourceNode;
+import io.confluent.ksql.planner.plan.KeyConstraint;
+import io.confluent.ksql.planner.plan.KeyConstraint.ConstraintOperator;
 import io.confluent.ksql.planner.plan.PlanNode;
 import java.util.Iterator;
 import java.util.List;
@@ -43,10 +45,10 @@ public class KeyedTableLookupOperator
 
   private ImmutableList<KsqlPartitionLocation> partitionLocations;
   private Iterator<Row> resultIterator;
-  private Iterator<GenericKey> keyIterator;
+  private Iterator<KsqlKey> keyIterator;
   private Iterator<KsqlPartitionLocation> partitionLocationIterator;
   private KsqlPartitionLocation nextLocation;
-  private GenericKey nextKey;
+  private KsqlKey nextKey;
   private long returnedRows = 0;
 
   public KeyedTableLookupOperator(
@@ -65,13 +67,10 @@ public class KeyedTableLookupOperator
       if (!nextLocation.getKeys().isPresent()) {
         throw new IllegalStateException("Table lookup queries should be done with keys");
       }
-      keyIterator = nextLocation.getKeys().get().stream().map(KsqlKey::getKey).iterator();
+      keyIterator = nextLocation.getKeys().get().stream().iterator();
       if (keyIterator.hasNext()) {
         nextKey = keyIterator.next();
-        resultIterator = mat.nonWindowed()
-            .get(nextKey, nextLocation.getPartition())
-            .map(ImmutableList::of)
-            .orElse(ImmutableList.of()).iterator();
+        resultIterator = getMatIterator(nextKey);
       }
     }
   }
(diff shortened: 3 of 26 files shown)


patrickstuedi:
Not ready for review

ksqldb-streams/src/main/java/io/confluent/ksql/execution/streams/materialization/ks/KsMaterializedTable.java
```diff
@@ -72,4 +74,23 @@
...
     }
   }
+
+  @Override
+  public Iterator<Row> get(final int partition, final GenericKey from, final GenericKey to) {
+    try {
+      final ReadOnlyKeyValueStore<GenericKey, ValueAndTimestamp<GenericRow>> store = stateStore
+          .store(QueryableStoreTypes.timestampedKeyValueStore(), partition);
+
+      System.out.println("### KsMaterializationTable, range, from " + from + ", to " + to);
```
vvcephei, line 84: TODO: printlns obviously need to be removed before merging.

ksqldb-engine/src/main/java/io/confluent/ksql/physical/pull/PullPhysicalPlan.java
```diff
@@ -138,8 +137,7 @@ private boolean requiresRequestsToAllPartitions() {
     return lookupConstraints.stream()
         .anyMatch(lookupConstraint -> {
           if (lookupConstraint instanceof KeyConstraint) {
-            final KeyConstraint keyConstraint = (KeyConstraint) lookupConstraint;
-            return keyConstraint.getConstraintOperator() != ConstraintOperator.EQUAL;
```
vvcephei, line 142: I didn't follow the reason for this change. It seems like range scans still do require requests to all partitions, right?

vvcephei (commented):
Thanks, @patrickstuedi !

I has a few questions for you.

ksqldb-engine/src/main/java/io/confluent/ksql/physical/pull/PullPhysicalPlan.java
patrickstuedi: Good catch, thx!

ksqldb-engine/src/main/java/io/confluent/ksql/planner/plan/KeyConstraint.java
```diff
@@ -86,10 +79,13 @@ public boolean isRangeOperator() {
 
     private final GenericKey key;
     private final Optional<WindowBounds> windowBounds;
+    private final ConstraintOperator operator;
 
-    public KeyConstraintKey(final GenericKey key, final Optional<WindowBounds> windowBounds) {
+    public KeyConstraintKey(final GenericKey key, final Optional<WindowBounds> windowBounds,
+                            final ConstraintOperator operator) {
```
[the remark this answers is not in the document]
patrickstuedi: I also can't see any reason why we should be have a class KeyConstraintKey that replicates KeyConstraint, deleted the former and made sure KeyConstraint implements KsqlKey.

ksqldb-engine/src/main/java/io/confluent/ksql/physical/pull/PullPhysicalPlan.java
```diff
@@ -123,26 +123,16 @@ public Materialization getMaterialization() {
...
-  private boolean requiresRequestsToAllPartitions() {
-    return lookupConstraints.stream()
-        .anyMatch(lookupConstraint -> {
-          if (lookupConstraint instanceof KeyConstraint) {
-            final KeyConstraint keyConstraint = (KeyConstraint) lookupConstraint;
-            return keyConstraint.getConstraintOperator() != ConstraintOperator.EQUAL;
-          }
-          return true;
-        });
+    return ImmutableList.copyOf(list);
```
AlanConfluent, line 135: It's possible if you have a query like `(KEY < 'abc' AND KEY > 'a') OR (KEY > 'def' AND KEY < 'e') OR (KEY <= `z` AND KEY >= 'y'`), you'll send up doing three scans which might be more expensive than doing a single table scans.  This is where a query optimizer could eventually be smart.  I think it's fine to do this at the moment, but it might make sense to try a benchmark at some point for some representative queries to see what's faster.

Also, at the moment, we do some deduping for the request for queries like `(KEY = 'a' OR KEY = 'a')` so that it will return a single row and won't issue the same query twice (even though that would be a nonsensical query).  It's possible to issue a query like `KEY < 'abc' OR KEY < 'a'`, which could also result in duplicates, though I don't think we'd dedupe the requests since they're a bit different.  The options are to:
- Recognize ranges and optimize the above query to be just `KEY < 'a'` -- this is something that would be nice to do, though it probably shouldn't be part of this PR.
- Dedupe results, not just queries.  This is a bit hard because we stream back results at the moment and are somewhat stateless.  This would require keeping a set of keys we've returned -- not a bad idea, though it means we could run out of memory or use a lot of it with a sufficiently large table size.  Probably a no-go for our use cases.
- Allow duplicates. I think the SQL purists would probably be angry here since this would break the model.
- Do a single table scan if you see more than one range query.  This works without issue, but might not be as performant as a few small double-bounded range queries.

I'm for the last solution since it's the simplest.  Curious what you think.

AlanConfluent (commented):
Looking good.  Just had a handful of comments.

ksqldb-engine/src/main/java/io/confluent/ksql/physical/pull/PullPhysicalPlan.java
```diff
@@ -123,26 +123,16 @@ public Materialization getMaterialization() {
   }
 
   public List<KsqlKey> getKeys() {
-    if (requiresRequestsToAllPartitions()) {
-      return Collections.emptyList();
+    final List<KsqlKey> list = new ArrayList<>();
+    for (LookupConstraint c : lookupConstraints) {
+      if (c instanceof KeyConstraint) {
+        final KeyConstraint kc = (KeyConstraint) c;
```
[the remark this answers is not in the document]
patrickstuedi: Currently for KeyContraints with a range operation we return the keys, because the keys are used by KeyedTableLookupOperator. Does this make sense?

ksqldb-engine/src/main/java/io/confluent/ksql/physical/pull/operators/KeyedTableLookupOperator.java
```diff
@@ -90,19 +90,45 @@ public Object next() {
...
+      throw new IllegalStateException(String.format("Keyed lookup queries should be done with "
+        + "key constraints: %s", ksqlKey.toString()));
+    }
+    final KeyConstraint keyConstraintKey = (KeyConstraint) ksqlKey;
+    if (keyConstraintKey.getOperator() == ConstraintOperator.EQUAL) {
+      return mat.nonWindowed()
+        .get(ksqlKey.getKey(), nextLocation.getPartition())
+        .map(ImmutableList::of)
+        .orElse(ImmutableList.of()).iterator();
+    } else if (keyConstraintKey.getOperator() == ConstraintOperator.GREATER_THAN
```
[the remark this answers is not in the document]
patrickstuedi: Yup, exactly, there is currently no other option. In the future we might want to consider change the range interface of the state stores to be more expressive so that we don't need extra filtering to trim the borders.

ksqldb-engine/src/main/java/io/confluent/ksql/planner/plan/QueryFilterNode.java
```diff
@@ -447,10 +441,44 @@ public Void visitComparisonExpression(
...
     }
 
+    public LookupConstraint getLookupConstraint(final Optional<WindowBounds> windowBounds) {
+      if (seenKeys.isEmpty()) {
+        return new NonKeyConstraint();
+      }
+
+      if (type == Type.EQUAL) {
+        return new KeyConstraint(ConstraintOperator.EQUAL,
+          GenericKey.fromList(Arrays.asList(keyContents)), windowBounds);
```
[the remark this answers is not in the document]
patrickstuedi: Agree that code is not nice, just not sure how to cleanly (without having the same type checks twice) factor out that statement without adding overhead to that one case where we return NonKeyConstraint.

ksqldb-engine/src/main/java/io/confluent/ksql/planner/plan/QueryFilterNode.java
```diff
@@ -447,10 +441,44 @@ public Void visitComparisonExpression(
...
+      } else if (isString) {
```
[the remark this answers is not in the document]
patrickstuedi: Yea, for non string types we currently fallback to NonKeyContraint (which leads to a tablescan) because we can't guarantee the order returned by the state store is going to be correct for non-string types.

ksqldb-engine/src/main/java/io/confluent/ksql/planner/plan/QueryFilterNode.java
```diff
@@ -340,16 +337,10 @@ public Void visitComparisonExpression(
...
 
         if (col.namespace() == Namespace.KEY) {
-          if (node.getType() != Type.EQUAL) {
-            setTableScanOrElseThrow(() ->
-                invalidWhereClauseException("Bound on key columns '"
-                        + getSource().getSchema().key() + "' must currently be '='", isWindowed));
-          }
           if (seenKeys.get(col.index())) {
             setTableScanOrElseThrow(() -> invalidWhereClauseException(
                 "An equality condition on the key column cannot be combined with other comparisons"
```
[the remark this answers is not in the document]
patrickstuedi: Agree, changed the message.

Currently we don't optimize disjunctions or conjunctions of range expresssions, means we fall back to table scans for those. Main reason is that the best place to make multi-range expresssions work efficiently is to optimize (compact) at the logical plan level, either when generating the logical plan or as an optimization just before we generate the physical plan. Probably best to have a separate PR for this.

ksqldb-engine/src/main/java/io/confluent/ksql/physical/pull/operators/KeyedTableLookupOperator.java
```diff
@@ -90,19 +90,45 @@ public Object next() {
...
+        || keyConstraintKey.getOperator() == ConstraintOperator.GREATER_THAN_OR_EQUAL) {
+      final GenericKey fromKey = keyConstraintKey.getKey();
+      final GenericKey toKey = null;
```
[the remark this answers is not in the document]
patrickstuedi: Agree, see my comment above.

ksqldb-engine/src/main/java/io/confluent/ksql/physical/pull/PullPhysicalPlan.java
patrickstuedi: Yea, the worst would be to have multiple range scans for such multi-range queries, the best would be to optimize those queries and compact the expressions. As for now the plan is to defer the optimization/compaction to a separate PR and fallback to table scan for multi-range queries so that at least we are not any worse than without this PR.

ksqldb-engine/src/test/java/io/confluent/ksql/physical/pull/operators/KeyedTableLookupOperatorTest.java
```diff
@@ -142,4 +151,51 @@ public void shouldLookupRowsForMultipleKeys() {
...
+    multipleKeysPartitionLocations.add(new KsLocator.PartitionLocation(
+      Optional.of(ImmutableSet.of(KEY1, KEY2)), 1, ImmutableList.of(node1)));
+    multipleKeysPartitionLocations.add(new KsLocator.PartitionLocation(
+      Optional.of(ImmutableSet.of(KEY3, KEY4)), 3, ImmutableList.of(node3)));
+
+    final KeyedTableLookupOperator lookupOperator = new KeyedTableLookupOperator(materialization, logicalNode);
+    when(materialization.nonWindowed()).thenReturn(nonWindowedTable);
+    when(materialization.nonWindowed().get(1,null, GKEY2)).thenReturn(Arrays.asList(ROW1).iterator());
+    when(materialization.nonWindowed().get(3,null, GKEY3)).thenReturn(Arrays.asList(ROW3).iterator());
+
```
patrickstuedi, line 191: @AlanConfluent This test is supposed to test a range query "<= KEY3 " in a setting with multiple partitions. Originally I didn't have line 179 and  189 which I thought would get me the desired behavior, but somehow I ended up adding those lines. Now I'm not sure if the test is correctly testing the case I described or if there is a bug in the code how keys/partitions are used for range queries. What do you think?

ksqldb-engine/src/main/java/io/confluent/ksql/physical/pull/PullPhysicalPlan.java
AlanConfluent: Sounds good to me.

I think the code at the moment doesn't yet do this, if I'm not mistaken.  We'd probably want to do a check at the logical layer in QueryFilterNode.  I made a suggestion about that below.

ksqldb-engine/src/main/java/io/confluent/ksql/planner/plan/QueryFilterNode.java
```diff
@@ -340,16 +340,10 @@ public Void visitComparisonExpression(
...
-            setTableScanOrElseThrow(() ->
-                invalidWhereClauseException("Bound on key columns '"
-                        + getSource().getSchema().key() + "' must currently be '='", isWindowed));
-          }
           if (seenKeys.get(col.index())) {
             setTableScanOrElseThrow(() -> invalidWhereClauseException(
-                "An equality condition on the key column cannot be combined with other comparisons"
-                    + " such as an IN predicate",
-                isWindowed));
+                "A comparison condition on the key column cannot be combined with other"
```
AlanConfluent, line 345: Regarding doing a table scan if there's more than one range query, just a suggestion about that check: 

In this validator, you probably want to keep track of `requiresRangeScan`, and then where the validator is called in `validateWhereClauseAndCheckTableScan`, you can count the number of them across all disjuncts and if you see more than one, return a table scan for the whole thing rather than doing multiple range scans.

ksqldb-engine/src/test/java/io/confluent/ksql/physical/pull/operators/KeyedTableLookupOperatorTest.java
patrickstuedi: @agavra regarding your above comment about a mult-partition functional tests, is this what you had in mind?

ksqldb-engine/src/main/java/io/confluent/ksql/planner/plan/QueryFilterNode.java
```diff
@@ -447,10 +444,49 @@ public Void visitComparisonExpression(
         final Object key = resolveKey(other, col.get(), metaStore, ksqlConfig, node);
         keyContents[col.get().index()] = key;
         seenKeys.set(col.get().index());
+        operatorType = node.getType();
+        rangeQuerySupported = isSupportedType(other);
```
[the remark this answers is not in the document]
patrickstuedi: Great!

ksqldb-engine/src/main/java/io/confluent/ksql/planner/plan/QueryFilterNode.java
```diff
@@ -447,10 +444,49 @@ public Void visitComparisonExpression(
         final Object key = resolveKey(other, col.get(), metaStore, ksqlConfig, node);
         keyContents[col.get().index()] = key;
         seenKeys.set(col.get().index());
+        operatorType = node.getType();
```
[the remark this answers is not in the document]
vvcephei: Interesting. Thanks for pointing this out, @AlanConfluent . You're saying that currently, if we have a multi-column key, we can't do table-scan queries today? It seems like we ought to have a test for that kind of thing. It also seems like that restriction isn't strictly necessary.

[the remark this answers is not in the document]
vvcephei: FWIW, I think multi-column keys are already excluded from the optimization, since @patrickstuedi restricted the key domain to just plain String keys. It's worth checking that multi-column keys do indeed fall back to table scans.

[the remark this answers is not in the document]
patrickstuedi: @vvcephei @AlanConfluent yes we currently fall back to table scan for multi-range expressions, there are a set of tests in the rest validation tests "pull-queries-with-scan.json"

https://github.com/confluentinc/ksql/pull/7993/files#diff-ace0366edb6fa80100210b650c4b01793f397c85d352a9683d59851a70637b3b

ksqldb-engine/src/main/java/io/confluent/ksql/planner/plan/QueryFilterNode.java
patrickstuedi: For queries with multiple range expressions we currently fall back to table scans. That code is validateWhereClause and was there even before this PR

ksqldb-engine/src/main/java/io/confluent/ksql/physical/pull/PullPhysicalPlanBuilder.java
```diff
@@ -180,27 +186,52 @@ private SelectOperator translateFilterNode(final QueryFilterNode logicalNode) {
...
+    pullPhysicalPlanType = getPlanType();
+    if (pullPhysicalPlanType == PullPhysicalPlanType.RANGE_SCAN
+        && !queryPlannerOptions.getRangeScansEnabled()) {
+      pullPhysicalPlanType = PullPhysicalPlanType.TABLE_SCAN;
+    }
+    if (pullPhysicalPlanType == PullPhysicalPlanType.TABLE_SCAN) {
+      if (queryPlannerOptions.getTableScansEnabled()) {
+        lookupConstraints = Collections.emptyList();
+      } else {
+        throw new KsqlException("Query requires table scan to enabled");
```
agavra, line 221: nit: let's make the error message actionable with exactly what they should do to enable it

ksqldb-engine/src/main/java/io/confluent/ksql/planner/plan/QueryFilterNode.java
```diff
@@ -123,6 +126,7 @@ public QueryFilterNode(
     extractKeysAndSystemCols();
 
     // Extraction of lookup constraints
+    this.typeManager = new ExpressionTypeManager(source.getSchema(), metaStore);
```
agavra, line 129: because (smartly so) you're using the column type instead of the literal type (`col.get().type()`) this is no longer used -right?

ksqldb-functional-tests/src/test/resources/rest-query-validation-tests/pull-queries-with-range-scan.json
```diff
@@ -0,0 +1,354 @@
...
+        "CREATE TABLE INPUT (ID STRING PRIMARY KEY, V0 STRING, V1 STRING) WITH (kafka_topic='test_topic', value_format='JSON');",
+        "CREATE TABLE MATVIEW AS SELECT ID, V0, V1 FROM INPUT;",
+        "SELECT ID, V0, V1 FROM MATVIEW WHERE ID <= '10';"
+      ],
+      "properties": {
+        "ksql.query.pull.table.scan.enabled":  false,
+        "ksql.query.pull.range.scan.enabled":  false
+      },
+      "expectedError": {
+        "message": "Query requires table scan to enabled"
```
agavra, line 18: could we be smart here and suggest enabling range scans in the case when the comparison would otherwise support range scans? this can be a follow-up ticket

ksqldb-functional-tests/src/test/resources/rest-query-validation-tests/pull-queries-with-range-scan.json
```diff
@@ -0,0 +1,354 @@
...
+        "CREATE TABLE INPUT (ID INTEGER PRIMARY KEY, V0 STRING, V1 STRING) WITH (kafka_topic='test_topic', value_format='JSON');",
+        "CREATE TABLE MATVIEW AS SELECT ID, V0, V1 FROM INPUT;",
+        "SELECT ID, V0, V1 FROM MATVIEW WHERE ID >= 12;"
+      ],
+      "properties": {
+        "ksql.query.pull.table.scan.enabled":  false,
+        "ksql.query.pull.range.scan.enabled":  true
+      },
+      "expectedError": {
+        "message": "Query requires table scan to enabled"
```
agavra, line 147: I'm not 100% sure about this suggestion, but in the same spirit as above we might want to specifically call out that range scans only work on string keys at the moment in this error message.

ksqldb-streams/src/main/java/io/confluent/ksql/execution/streams/materialization/ks/KsLocator.java
```diff
@@ -92,20 +93,30 @@
     this.applicationId = requireNonNull(applicationId, "applicationId");
   }
 
+  public List<KsqlPartitionLocation> locate(final List<KsqlKey> keys,
```
agavra, line 96: why are we introducing this method? I feel like `isRangeScan` should always be passed in to maintain correctness. If it's just for testing, let's just update the tests. And speaking of... this change is missing test coverage in `KsLocatorTest` 😉

agavra (approved):
This LGTM! You should wait for @AlanConfluent or @vvcephei's +1 as well as they have a bit more context into the code than I do.

ksqldb-functional-tests/src/test/resources/rest-query-validation-tests/pull-queries-with-range-scan.json
patrickstuedi: not sure how do this in a single validation test. for those few tests where we force range queries to fall back to table scans we can't determine if they could in theory use the range optimization, that's why we have those tests paired with a test that enables range scan.

ksqldb-functional-tests/src/test/resources/rest-query-validation-tests/pull-queries-with-range-scan.json
patrickstuedi: There might be a way to make this work. Problem is, non-string range scans are not an error and throw no exception but just fallback to table scans, so it's currently hard to catch. Will think about it in a follow-up.

ksqldb-streams/src/main/java/io/confluent/ksql/execution/streams/materialization/ks/KsLocator.java
patrickstuedi: Fair point.

ksqldb-engine/src/main/java/io/confluent/ksql/planner/plan/QueryFilterNode.java
AlanConfluent: I'm not referring to more than one range in a single disjunct (e.g. `KEY < 'a' AND KEY > 'b'`), I'm referring to multiple disjuncts: `KEY < 'a' OR KEY > 'b'`.  The latter would do multiple scans and require deduping or can be done in a single table scan.  I'm hoping to do a table scan in the latter case too.

ksqldb-engine/src/main/java/io/confluent/ksql/planner/plan/QueryFilterNode.java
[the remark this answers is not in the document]
AlanConfluent: > You're saying that currently, if we have a multi-column key, we can't do table-scan queries today? It seems like we ought to have a test for that kind of thing. It also seems like that restriction isn't strictly necessary.

I didn't meant to imply that.  Table scan should work for this today. The validator above will check to see if any operation isn't an "=" and potentially make it a scan if it is.  Currently, only if they're all "=" will it skip a table scan and do a key lookup.  To handle range queries, this all still holds true, but we can additionally carve out the case of a single key column with range operator from the cases that were being handled by table scan, and now handle that by range query.

ksqldb-functional-tests/src/test/resources/rest-query-validation-tests/pull-queries-with-range-scan.json
```diff
@@ -0,0 +1,354 @@
...
+      "expectedError": {
+        "message": "A comparison condition on the key column cannot be combined with other comparisons such as an IN predicate"
+      }
+    },
+    {
+      "name": "range scan query must fall back to table scan for multiple range expressions if table scan is enabled",
+      "statements": [
+        "CREATE TABLE INPUT (ID STRING PRIMARY KEY, V0 STRING, V1 STRING) WITH (kafka_topic='test_topic', value_format='JSON');",
+        "CREATE TABLE MATVIEW AS SELECT ID, V0, V1 FROM INPUT;",
+        "SELECT ID, V0, V1 FROM MATVIEW WHERE ID >= '11' AND ID <= '13';"
```
AlanConfluent, line 199: This is one great case to test, one that was mentioned.

I was also trying to ask for testing of `ID >= '11' OR ID <= '13'`.  We'd also like this to be a table scan.

ksqldb-functional-tests/src/test/resources/rest-query-validation-tests/pull-queries-with-range-scan.json
```diff
@@ -0,0 +1,354 @@
+{
+  "comments": [
```
AlanConfluent, line 2: We should also test some of the multi-column key cases. One way to do this is:

```
CREATE TABLE AGGREGATE AS SELECT ID1, ID2, COUNT(1) AS COUNT FROM INPUT GROUP BY ID1, ID2;
SELECT * FROM AGGREGATE WHERE ID1='11' AND ID2=10;
```

ksqldb-engine/src/main/java/io/confluent/ksql/planner/plan/QueryFilterNode.java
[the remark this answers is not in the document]
AlanConfluent: I believe the current code, if you have multiple key columns, will not behave correctly for a query like this:
- `KEY1 < 10 AND KEY2 = 5`

operatorType will be set to the latest operator seen in the AST traversal, which is equals.  It will, based on that, assume they're all equality, which was a simplifying assumption we could make before because we only allowed equality, but now isn't quite accurate.  I think the current code would just do an equality lookup rather than range operator on one column.  If we see this intermixing of operators, we should just fall back on table scan.

To keep track of that, I think we need to keep track of all of the operators.  Only if they're all "=" can the lookup operator be equal.  I think intermixing is a table scan and if it's a single column key, then range operators can result in lookup operators that are range.  My comments above were meant to say that even if we have all "<" for a multi column key, we wouldn't want to do a range scan operation since it wouldn't make sense.

> At the moment multi key columns only work for equality, so other operators should throw an error in that case.

I should have said fall back on table scan.

ksqldb-engine/src/main/java/io/confluent/ksql/physical/pull/PullPhysicalPlan.java
AlanConfluent: I think the current code handles `KEY < 'a' AND KEY > 'b'`, allowing this to be a table scan, but currently, `KEY < 'a' OR KEY > 'b'` would result is two separate range queries and some of the issues I was referring to requiring deduping.

patrickstuedi: Queries with multiple disjuncts will either be key lookups if all disjuncts are Equality, or table scans. Based on how we determine the plan in PullPhysicalPlanBuilder. Validation tests should cover this.

AlanConfluent:
I think I commented on a version that is the same as your latest, but not after your rebase by mistake.

This looks good.  I don't think you handled windowed tables yet.  If you're bold, you can try sticking it in this PR, but it would also be fine to do it in a followup.

AlanConfluent (approved):

ksqldb-engine/src/main/java/io/confluent/ksql/planner/plan/QueryFilterNode.java
patrickstuedi: Multiple range expressions, both across disjuncts or within a single disjuncts fall back to table scan. Added tests to cover those cases.

ksqldb-functional-tests/src/test/resources/rest-query-validation-tests/pull-queries-with-range-scan.json
patrickstuedi: Thanks for pointing out the multi-column cases. Got all those cases covered in the validation tests now.

PR #7994, opened 2021-08-12
test: fix drop statement tests in QTT

jzaralim:
### Description 

Fixed because it is needed for #7474.

In the drop_sources.json QTT test, all of the drop statements dropped streams that shared a topic with another stream, for example.

```
"CREATE STREAM input2 (K STRING KEY, data VARCHAR) WITH (kafka_topic='input', value_format='DELIMITED');",
"DROP STREAM IF EXISTS input2;",
"CREATE STREAM input (K STRING KEY, data VARCHAR) WITH (kafka_topic='input', value_format='DELIMITED');",
"CREATE STREAM output AS SELECT * FROM input;"
```

This is because the QTT would have thrown an error if the dropped source had a topic that was only referenced once.

This PR catches and ignores the error so that we could run DROP tests without forcing streams/tables to share topics:

```
"CREATE SOURCE STREAM abc (K STRING KEY, data VARCHAR) WITH (kafka_topic='abc', value_format='DELIMITED');",
"CREATE STREAM input (K STRING KEY, data VARCHAR) WITH (kafka_topic='input', value_format='DELIMITED');",
"CREATE STREAM output AS SELECT * FROM input;",
"DROP STREAM abc DELETE TOPIC;"
```

The DROP statements still doesn't completely work though - the following test runs, but the test code thinks that the topic `abc` was deleted even though it wasn't.

```
"CREATE STREAM abc (K STRING KEY, data VARCHAR) WITH (kafka_topic='abc', value_format='DELIMITED');",
"CREATE STREAM input (K STRING KEY, data VARCHAR) WITH (kafka_topic='input', value_format='DELIMITED');",
"CREATE STREAM output AS SELECT * FROM input;",
"DROP STREAM abc;"
```

However, if we just want to use these tests to check the behavior of `DROP` vs `DROP.... DELETE TOPIC;` for `SOURCE` sources without actually checking the topics, then this is good enough.

### Reviewer checklist

### DIFF
--- a/ksqldb-functional-tests/src/main/java/io/confluent/ksql/test/tools/TestExecutor.java
+++ b/ksqldb-functional-tests/src/main/java/io/confluent/ksql/test/tools/TestExecutor.java
@@ -217,6 +217,7 @@ public class TestExecutor implements Closeable {
             try {
               topicInfoCache.get(name);
             } catch (Exception e) {
+              e.printStackTrace(System.out);
             }
           });
 
--- a/ksqldb-functional-tests/src/main/java/io/confluent/ksql/test/tools/TestExecutor.java
+++ b/ksqldb-functional-tests/src/main/java/io/confluent/ksql/test/tools/TestExecutor.java
@@ -213,7 +213,12 @@ public class TestExecutor implements Closeable {
 
       kafka.getAllTopics().stream()
           .map(Topic::getName)
-          .forEach(topicInfoCache::get);
+          .forEach(name -> {
+            try {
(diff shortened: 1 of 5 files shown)


mjsax:
@jzaralim It seems #7474 was merged, so this fix was actually not needed for it?

Not sure if I understand the examples in the PR description? Which statements actually fail, and what it the error? It seems your PR only swallows some exceptions, but I am wondering what the actual root cause of those exceptions is, and if we should rather fix those, to avoid that an exception is thrown incorrectly in the first place?

PR #7995, opened 2021-08-12
fix: Fixes pull query latency distribution metrics

AlanConfluent:
### Description 
Fixes the distribution metrics since they were not setup correctly.

Previously, the max value was set to be 1000 microseconds, which is much less than the single digit milliseconds we generally get. I also changed it to be a "linear" bucketizer so as to have finer grained smaller metrics, even though the max value is ten seconds.

### Testing done 
Manual and unit testing.

### Reviewer checklist


ksqldb-engine/src/main/java/io/confluent/ksql/internal/PullQueryExecutorMetrics.java
```diff
@@ -475,10 +477,9 @@ private void addRequestMetricsToSensor(
     );
 
     sensor.add(new Percentiles(
-        100,
-        0,
-        1000,
-        BucketSizing.CONSTANT,
+        4 * NUM_LATENCY_BUCKETS,
```
nateab, line 480: Where does the 4 come from?

AlanConfluent: Conversion of buckets to bytes, which is what the api asks for...  I wish they would just change the API.  I'll add a comment.

PR #7996, opened 2021-08-12
feat: fail scalable push query if error is detected in subscribed persistent query

nateab:
### Description 
Listen for errors in the persistent query that a scalable push query is subscribed to and if one occurs, fail the scalable push query. 

### Testing done 
Unit tests added, as well as manual testing to ensure that we see `Persistent query has error` on the CLI.

### Reviewer checklist


ksqldb-engine/src/main/java/io/confluent/ksql/util/PersistentQueryMetadataImpl.java
```diff
@@ -105,7 +107,24 @@ public PersistentQueryMetadataImpl(
         maxQueryErrorsQueueSize,
         retryBackoffInitialMs,
         retryBackoffMaxMs,
-        listener
+        new QueryMetadata.Listener() {
```
AlanConfluent, line 110: Let just define a QueryMetadata.Listener below as a named class to make this a little cleaner.

ksqldb-engine/src/main/java/io/confluent/ksql/util/PersistentQueryMetadataImpl.java
```diff
@@ -105,7 +107,24 @@ public PersistentQueryMetadataImpl(
...
+          @Override
+          public void onError(final QueryMetadata queryMetadata, final QueryError error) {
+            listener.onError(queryMetadata, error);
+            scalablePushRegistry.get().onError();
```
AlanConfluent, line 114: Do `scalablePushRegistry.ifPresent`.  Otherwise, if SPQs are disabled, this will throw an exception.

ksqldb-engine/src/main/java/io/confluent/ksql/util/QueryMetadataImpl.java
```diff
@@ -196,14 +196,14 @@ public void initialize() {
...
           );
-      listener.onError(this, queryError);
       queryErrors.add(queryError);
       LOG.error(
           "Unhandled exception caught in streams thread {}. ({})",
           Thread.currentThread().getName(),
           errorType,
           e
       );
+      listener.onError(this, queryError);
```
AlanConfluent, line 206: Curious, why was it required to move this call?

ksqldb-engine/src/test/java/io/confluent/ksql/physical/scalablepush/ScalablePushRegistryTest.java
```diff
@@ -52,6 +53,8 @@
   @Mock
   private ProcessingQueue processingQueue;
   @Mock
+  private ConcurrentHashMap<QueryId, ProcessingQueue> processingQueues;
```
AlanConfluent, line 56: Unused?

ksqldb-engine/src/test/java/io/confluent/ksql/physical/scalablepush/PushPhysicalPlanTest.java
```diff
@@ -124,6 +124,39 @@ public void shouldStopOnDroppedRows() throws InterruptedException {
...
+      runnableCaptor.getValue().run();
+
+      context.owner().cancelTimer(timerId);
+    });
+
+    while (subscriber.getError() == null) {
+      Thread.sleep(100);
+    }
+
+    assertThat(subscriber.getError().getMessage(), containsString("Persistent query has error"));
```
AlanConfluent, line 154: Nice test

AlanConfluent (commented):
Looking good.  Just a couple comments

ksqldb-engine/src/main/java/io/confluent/ksql/util/QueryMetadataImpl.java
nateab: I was failing some tests in `SecureIntegrationTest` that called `assertQueryFailsWithUserError`. It seemed that the `queryError` was not being added and so the tests failed, but when i moved the listener to the end they passed

ksqldb-engine/src/main/java/io/confluent/ksql/util/PersistentQueryMetadataImpl.java
```diff
@@ -226,4 +228,32 @@ public synchronized void stop() {
   public Optional<ScalablePushRegistry> getScalablePushRegistry() {
     return scalablePushRegistry;
   }
+
+  private static final class ScalablePushQueryListener implements Listener {
```
AlanConfluent, line 232: This technically also calls the other listener, so maybe give it a more generic name like PushQueryListenerWrapper?

ksqldb-engine/src/main/java/io/confluent/ksql/query/TransientQueryQueue.java
```diff
@@ -138,7 +138,8 @@ public void acceptRow(final List<?> key, final GenericRow value) {
   public boolean acceptRowNonBlocking(final List<?> key, final GenericRow value) {
     try {
       if (!callback.shouldQueue()) {
-        return false;
+        callback.onQueued();
```
AlanConfluent, line 141: We talked about this offline.  Seems like onQueued doesn't need to be called.

ksqldb-engine/src/main/java/io/confluent/ksql/util/QueryMetadataImpl.java
AlanConfluent: We looked into this offline and this line can be reverted.

AlanConfluent (approved):

PR #8003, opened 2021-08-16
chore: pin commons-compress to avoid transitive CVE

agavra:
### Description 
chore, 1.20 of commons compress had a known vulnerability - it's pulled in transitively via avro, but we're pinning in here instead because upgrading avro historically has caused some issues.

### Testing done 
ran the test suite and also spun up a ksql cluster and created/tested some topics with AVRO encoding.

### Reviewer checklist


nateab (approved):
LGTM!

PR #8028, opened 2021-08-19
chore: revert GRACE PERIOD syntax from stream-stream joins

spena:
### Description 
Reverts the GRACE PERIOD syntax from stream-stream joins. There's an issue with the new semantics (see https://github.com/apache/kafka/pull/11235) that's need to be fixed first.

### Testing done 

### Reviewer checklist


guozhangwang (approved):

ableegoldman:
@spena we should have reverted this in the 7.0 branch, not `master` -- we still want to fix forward but just need to cut it from the 7.0 release, right? Maybe you can cherrypick this commit to revert it in 7.0, then apply another commit to revert the reversion in master? Same goes for any other PRs

PR #8039, opened 2021-08-23
feat: perform SchemaRegistry permissions on C*AS sink subjects

spena:
### Description 
Added SchemaRegistry permission checks to the sink subjects used by C*AS commands.

### Testing done 

### Reviewer checklist

### DIFF
--- a/ksqldb-engine/src/main/java/io/confluent/ksql/security/KsqlAuthorizationValidatorImpl.java
+++ b/ksqldb-engine/src/main/java/io/confluent/ksql/security/KsqlAuthorizationValidatorImpl.java
@@ -19,18 +19,25 @@ import io.confluent.ksql.execution.ddl.commands.KsqlTopic;
 import io.confluent.ksql.metastore.MetaStore;
 import io.confluent.ksql.metastore.model.DataSource;
 import io.confluent.ksql.name.SourceName;
+import io.confluent.ksql.parser.properties.with.CreateSourceAsProperties;
 import io.confluent.ksql.parser.tree.CreateAsSelect;
 import io.confluent.ksql.parser.tree.CreateSource;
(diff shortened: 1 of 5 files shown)


ksqldb-engine/src/main/java/io/confluent/ksql/security/KsqlAuthorizationValidatorImpl.java
```diff
@@ -150,12 +166,51 @@ private String getSourceTopicName(final MetaStore metaStore, final SourceName st
...
+      final Format valueFormat = FormatFactory.fromName(
+          properties.getValueFormat().orElse(defaultTopicValueFormat));
+
+      sinkTopicName = properties.getKafkaTopic().get();
+
+      sinkKeyFormat = KeyFormat.of(
+          FormatInfo.of(keyFormat.name()),
+          keyFormat.supportsFeature(SerdeFeature.SCHEMA_INFERENCE)
+              ? SerdeFeatures.of(SerdeFeature.SCHEMA_INFERENCE)
+              : SerdeFeatures.of(),
```
spena, line 201: `keyFormat` has a `supportedFeatures()` method that returns a set of features. I could call `SerdeFeatures.from(keyFormat.supportedFeatures())` here, but `SerdeFeatures.from` throws an exception because of incompatible features returned. Because I just need SCHEMA_INFERENCE, then I just added this condition here.

jzaralim: IIUC, the only way to hit the incompatible features exception is if UNWRAP_SINGLES and WRAP_SINGLES are both included. Why is that happening?

jzaralim (approved):
lgtm, just one question

PR #8043, opened 2021-08-23
feat: create and write a query plan for source tables

spena:
### Description 
As part of the source table materialization feature, this PR creates a query plan for a source table command.
The query plan is not executed in this PR yet (a follow-up PR will include this). 

### Testing done 
I will add unit tests in `KsqlEngineTest` in the PR that adds the CST query execution.

### Reviewer checklist

### DIFF
--- a/ksqldb-engine/src/main/java/io/confluent/ksql/engine/EngineExecutor.java
+++ b/ksqldb-engine/src/main/java/io/confluent/ksql/engine/EngineExecutor.java
@@ -19,23 +19,41 @@ import static io.confluent.ksql.metastore.model.DataSource.DataSourceType;
 
 import com.google.common.base.Throwables;
 import com.google.common.collect.Iterables;
+import edu.umd.cs.findbugs.annotations.SuppressFBWarnings;
 import io.confluent.ksql.KsqlExecutionContext.ExecuteResult;
 import io.confluent.ksql.analyzer.ImmutableAnalysis;
 import io.confluent.ksql.config.SessionConfig;
+import io.confluent.ksql.execution.ddl.commands.CreateTableCommand;
 import io.confluent.ksql.execution.ddl.commands.DdlCommand;
+import io.confluent.ksql.execution.ddl.commands.KsqlTopic;
+import io.confluent.ksql.execution.expression.tree.UnqualifiedColumnReferenceExp;
 import io.confluent.ksql.execution.plan.ExecutionStep;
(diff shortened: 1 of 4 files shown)


ksqldb-engine/src/main/java/io/confluent/ksql/engine/QueryEngine.java
```diff
@@ -72,7 +71,7 @@ static OutputNode buildQueryLogicalPlan(
   PhysicalPlan buildPhysicalPlan(
       final LogicalPlanNode logicalPlanNode,
       final SessionConfig config,
-      final MutableMetaStore metaStore,
+      final MetaStore metaStore,
```
spena, line 74: A **mutable** metastore is not required when building the physical plan.

ksqldb-engine/src/main/java/io/confluent/ksql/engine/EngineExecutor.java
```diff
@@ -347,28 +369,128 @@ TransientQueryMetadata executeTransientQuery(
     );
   }
 
+  @SuppressFBWarnings(value = "NP_NULL_PARAM_DEREF_NONVIRTUAL")
```
spena, line 372: I had to add this because calling the `new QueryPlan(...)` with null values was failing.

ksqldb-engine/src/main/java/io/confluent/ksql/engine/EngineExecutor.java
```diff
@@ -347,28 +369,128 @@ TransientQueryMetadata executeTransientQuery(
...
+        Optional.empty(),
+        tempMetastore
+    );
+
+    final KsqlBareOutputNode outputNode =
+        (KsqlBareOutputNode) plans.logicalPlan.getNode().get();
+
+    final QueryPlan queryPlan = new QueryPlan(
+        getSourceNames(outputNode),
+        null,
```
spena, line 447: The `QueryPlan` requires a `SourceName` for the sink topic. However, CST plans do not have sink topics. Passing an empty source name also fails because source names validate they're not empty. I couldn't find another way to avoid passing a null.

jzaralim (approved):

ksqldb-engine/src/main/java/io/confluent/ksql/engine/EngineExecutor.java
jzaralim: Sure, approved

PR #8062, opened 2021-08-26
fix: Removes config check to insert SPQ Processor

AlanConfluent:
### Description 
Replaces the check for `ksql.query.push.scalable.enabled` with `ksql.query.push.scalable.registry.installed` to insert the streams Processor for Scalable Push Queries. The idea behind this is that we want the user to be able to enable SPQs using the config `ksql.query.push.scalable.enabled` per request or in the CLI and control the registry installation separately.

Also, ups the ProcessingQueue limit to 1000 to allow for larger bursts.

### Testing done 
Verified it works as intended.

### Reviewer checklist


AlanConfluent:
> I'm +1 to the idea of making this enabling at the per-query/request level.
> 
> Just clarifying, at the per-request level this config is still turned off by default or not?

It is off by default, but will be possible to enable per query, in which case it will still only use SPQs if they are appropriate.

PR #8064, opened 2021-08-26
feat: introduce stream pull queries for HTTP/1

vvcephei:
Partially implement stream pull queries as proposed in:
[KLIP-53](https://github.com/confluentinc/ksql/blob/master/design-proposals/klip-53-pull-queries-on-streams.md)

This PR only adds the new query type to the HTTP/1 endpoint.
The others will be covered in later PRs.

This PR intentionally does not record metrics for stream pull queries. That requires some refactoring and
will be covered in later PRs.

### Reviewer checklist


ksqldb-engine/src/main/java/io/confluent/ksql/engine/KsqlEngine.java
```diff
@@ -293,6 +316,186 @@ public TransientQueryMetadata executeTransientQuery(
...
+    } catch (final ExecutionException e) {
+      log.error("Error executing Admin#listOffsets(" + topicDescription.name() + ")", e);
+      throw new KsqlServerException("Internal Server Error");
+    } catch (final TimeoutException e) {
+      log.error("Admin#listOffsets(" + topicDescription.name() + ") timed out", e);
+      throw new KsqlServerException("Backend timed out");
+    }
+  }
+
+  private boolean passedEndOffsets(final Admin admin,
```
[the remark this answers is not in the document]
vvcephei: Eh, I tried this refactor out, and I think it makes the code slightly more complicated to follow, since we then have to book-keep which partitions we're done with. Since the perf improvement is marginal, I'll just keep this the same.

ksqldb-engine/src/main/java/io/confluent/ksql/engine/KsqlEngine.java
```diff
@@ -293,6 +316,186 @@ public TransientQueryMetadata executeTransientQuery(
...
+            .put(ConsumerConfig.AUTO_OFFSET_RESET_CONFIG, "earliest")
+            .put(StreamsConfig.NUM_STREAM_THREADS_CONFIG, 1)
+            .put(StreamsConfig.COMMIT_INTERVAL_MS_CONFIG, 100)
+            .build()
+    );
+    try {
+
+      final TransientQueryMetadata transientQueryMetadata = EngineExecutor
+          .create(primaryContext, serviceContext, statement.getSessionConfig())
+          .executeTransientQuery(statement, excludeTombstones);
```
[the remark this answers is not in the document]
vvcephei: Hmm, it looks like this is actually config driven
```
  public static final String
      KSQL_TRANSIENT_QUERY_NAME_PREFIX_CONFIG = "ksql.transient.prefix";
  public static final String
      KSQL_TRANSIENT_QUERY_NAME_PREFIX_DEFAULT = "transient_";
```

Do you think it's worth introducing a new config+default to distinguish "transient pull queries" from "transient (push) queries"? Offhand, it seems like it could be nice, but I'm not sure exactly why. Do we need to reason directly about consumer id / app ids often? I.e., how much information do we need to encode directly in the app id vs. being able to look up given the app id?

ksqldb-engine/src/main/java/io/confluent/ksql/engine/KsqlEngine.java
```diff
@@ -293,6 +316,186 @@ public TransientQueryMetadata executeTransientQuery(
...
+        try {
+          Thread.sleep(50L);
+        } catch (final InterruptedException e) {
+          log.error("Interrupted waiting for stream pull query to complete", e);
+          throw new KsqlServerException("Interrupted");
+        }
+      }
+    }
+    transientQueryMetadata.getKafkaStreams().close(Duration.ZERO);
+    QueryLogger.info("Finished pull query results", statement.getStatementText());
```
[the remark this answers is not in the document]
vvcephei: Yep, it gets registered in the StreamedQueryResource when we do this:
```
      // stop the response stream before sending the successful response.
      queryStreamWriter.close();
```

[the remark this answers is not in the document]
vvcephei: Now that I'm thinking about it again, IIRC, the proactive close here is actually just to make the query termination faster. Otherwise, I think it had to wait for some component's polling interval to close the query asynchronously. But since we're waiting on the query synchronously in this case, it's beneficial to start the process of closing KS immediately. Admittedly a bit hacky, but then again this whole implementation strategy is more or less a hack.

vvcephei:
Added a bunch more positive and negative tests for stream pull queries. Also moved them to their own file.

ksqldb-api-client/src/test/java/io/confluent/ksql/api/client/integration/ClientIntegrationTest.java
```diff
@@ -541,7 +541,7 @@ public void shouldHandleErrorResponseFromExecuteQuery() {
     // Then
     assertThat(e.getCause(), instanceOf(KsqlClientException.class));
     assertThat(e.getCause().getMessage(), containsString("Received 400 response from server"));
-    assertThat(e.getCause().getMessage(), containsString("Missing WHERE clause"));
+    assertThat(e.getCause().getMessage(), containsString("A_FAKE_TABLE_NAME does not exist."));
```
vvcephei, line 544: This test only cares that it gets an error. It was producing that error by running the query that we are now supporting. I replaced it with a query that will never work.

ksqldb-functional-tests/src/main/java/io/confluent/ksql/test/tools/TestJsonMapper.java
```diff
@@ -40,6 +41,7 @@
       .registerModule(new StructSerializationModule())
       .registerModule(new KsqlTypesSerializationModule())
       .registerModule(new KsqlTypesDeserializationModule())
+      .enable(Feature.ALLOW_COMMENTS)
```
vvcephei, line 44: To let us add comments in the *QTT test files.

ksqldb-functional-tests/src/test/resources/rest-query-validation-tests/pull-queries-against-materialized-aggregates.json
```diff
@@ -1211,19 +1211,6 @@
         "status": 400
       }
     },
-    {
-      "name": "on stream",
```
vvcephei, line 1215: relocated to new test file.

ksqldb-functional-tests/src/test/resources/rest-query-validation-tests/pull-queries-against-materialized-aggregates.json
```diff
@@ -2894,24 +2881,6 @@
         ]}
       ]
     },
-    {
-      "name": "should give good error message when querying a stream with multiple input queries",
```
vvcephei, line 2898: Testing for this error is no longer valid.

ksqldb-functional-tests/src/test/resources/rest-query-validation-tests/pull-queries-against-streams.json
```diff
@@ -0,0 +1,137 @@
...
+    },
+    {
+      "name": "error on ORDER BY",
+      "statements": [
+        "CREATE STREAM S1 (MYKEY INT KEY, MYVALUE INT) WITH (kafka_topic='test_topic', value_format='JSON');",
+        "SELECT * FROM S1 ORDER BY MYKEY DESC;"
+      ],
+      "expectedError": {
+        "type": "io.confluent.ksql.rest.entity.KsqlStatementErrorMessage",
+        "message": "line 1:24: mismatched input 'BY' expecting {';', 'EMIT', 'WHERE', 'WINDOW', 'GROUP', 'HAVING', 'LIMIT', 'PARTITION'}\nStatement: SELECT * FROM S1 ORDER BY MYKEY DESC;\nCaused by: line 1:24: mismatched input 'BY' expecting {';', 'EMIT', 'WHERE',\n\t'WINDOW', 'GROUP', 'HAVING', 'LIMIT', 'PARTITION'}\nCaused by: org.antlr.v4.runtime.InputMismatchException",
```
vvcephei, line 51: This error message is pretty nasty.

ksqldb-functional-tests/src/test/resources/rest-query-validation-tests/pull-queries-against-streams.json
```diff
@@ -0,0 +1,137 @@
+  {
+  "comments": [
+    "Tests covering Pull queries of streams"
```
vvcephei, line 3: @agavra , I added all the positive and negative cases I could think of here. Any I missed?

ksqldb-engine/src/main/java/io/confluent/ksql/engine/KsqlEngine.java
```diff
@@ -293,6 +319,195 @@ public TransientQueryMetadata executeTransientQuery(
...
+          // special case where we expect no work at all on the partition, so we don't even
+          // need to check the committed offset (if we did, we'd potentially wait forever,
+          // since Streams won't commit anything for an empty topic).
+          .filter(e -> e.getValue().offset() > 0L)
+          .collect(toMap(Entry::getKey, e -> e.getValue().offset()));
+    } catch (final InterruptedException e) {
+      log.error("Admin#listOffsets(" + topicDescription.name() + ") interrupted", e);
+      throw new KsqlServerException("Interrupted");
+    } catch (final ExecutionException e) {
+      log.error("Error executing Admin#listOffsets(" + topicDescription.name() + ")", e);
```
guozhangwang, line 482: Usually `ExecutionException` would wrap the actual root cause, so better including that in the constructor as well, like: `KsqlServerException("Internal Server Error", e.getCause())`

vvcephei: Thanks, @guozhangwang . I'm leaning toward logging this information, but not including it in the exception.

I was concerned about exposing too much information in HTTP responses: https://github.com/confluentinc/ksql/pull/8064#discussion_r697691629

ksqldb-engine/src/main/java/io/confluent/ksql/engine/KsqlEngine.java
```diff
@@ -293,6 +319,195 @@ public TransientQueryMetadata executeTransientQuery(
...
+          .executeTransientQuery(statement, excludeTombstones);
+
+      QueryLogger.info(
+          "Streaming stream pull query results '{}'",
+          statement.getStatementText()
+      );
+
+      return transientQueryMetadata;
+
+    } catch (final KsqlStatementException e) {
```
AlanConfluent, line 373: For what it's worth, I think KsqlStatementException is taken as a 4XX for at least `/query-stream`.  Making things consistent is probably part of a larger fix, but just wanted to point that out here since all exceptions are wrapped in it.

ksqldb-engine/src/main/java/io/confluent/ksql/engine/KsqlEngine.java
```diff
@@ -293,6 +316,186 @@ public TransientQueryMetadata executeTransientQuery(
...
+    final IsolationLevel isolationLevel =
+        StreamsConfig.AT_LEAST_ONCE.equals(processingGuarantee)
+            ? IsolationLevel.READ_UNCOMMITTED
+            : IsolationLevel.READ_COMMITTED;
+
+    final Map<TopicPartition, Long> endOffsets = getEndOffsets(
+        admin,
+        topicDescription,
+        isolationLevel
+    );
```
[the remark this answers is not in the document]
AlanConfluent: Nice.  Not sure if there's some generic way to share this code or it's a little different from the test code.  Might be worth factoring out.

ksqldb-engine/src/main/java/io/confluent/ksql/engine/KsqlEngine.java
```diff
@@ -293,6 +319,195 @@ public TransientQueryMetadata executeTransientQuery(
...
+        .getConfig(true)
+        .getKsqlStreamConfigProps()
+        .get(StreamsConfig.PROCESSING_GUARANTEE_CONFIG);
+
+    final IsolationLevel isolationLevel =
+        StreamsConfig.AT_LEAST_ONCE.equals(processingGuarantee)
+            ? IsolationLevel.READ_UNCOMMITTED
+            : IsolationLevel.READ_COMMITTED;
+
+    final Map<TopicPartition, Long> endOffsets = getEndOffsets(
```
AlanConfluent, line 405: This is a snapshot in time and the end offsets can change while the query runs.  We're effectively saying that we're going to give you the results up to the point when the query was run, which seems pretty reasonable since we're making no promise about reaching any particular point, but really just giving the illusion of a point in time query.

ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/resources/streaming/StreamedQueryResource.java
```diff
@@ -396,10 +399,16 @@ private EndpointResponse handleQuery(final KsqlSecurityContext securityContext,
           );
         }
         case KSTREAM: {
-          throw new KsqlStatementException(
-              "Pull queries are not supported on streams."
-                  + PullQueryValidator.PULL_QUERY_SYNTAX_HELP,
-              statement.getStatementText()
+          final SessionConfig sessionConfig = SessionConfig.of(ksqlConfig, configProperties);
```
AlanConfluent, line 402: Is there any reason not to allow this for tables which are not materialized? I guess there's the risk that they may not know the difference and assume this will be as performant as a state store lookup, but it seems like this should more or less work right out of the box for those as well.

ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/resources/streaming/StreamedQueryResource.java
```diff
@@ -540,6 +549,44 @@ private EndpointResponse handleScalablePushQuery(
...
+        connectionClosedFuture
+    )) {
+      ksqlEngine.waitForStreamPullQuery(
+          serviceContext,
+          analysis,
+          configured,
+          transientQueryMetadata
+      );
+      // stop the response stream before sending the successful response.
+      queryStreamWriter.close();
```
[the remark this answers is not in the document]
AlanConfluent: It should always get called at the end of `QueryStreamWriter.write`.  We don't explicitly call this for push queries.  Not a huge deal, but curious why they're different.

ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/resources/streaming/StreamedQueryResource.java
```diff
@@ -540,6 +549,43 @@ private EndpointResponse handleScalablePushQuery(
...
+      final ConfiguredStatement<Query> configured,
+      final CompletableFuture<Void> connectionClosedFuture) {
+
+    final TransientQueryMetadata transientQueryMetadata = ksqlEngine
+        .createStreamPullQuery(
+            serviceContext,
+            configured,
+            false
+        );
+    localCommands.ifPresent(lc -> lc.write(transientQueryMetadata));
```
AlanConfluent, line 564: Nice, I think this is important.  This ensures that any transient queries that fail to get cleaned up will with a roll, and this is no different, so good to have it.

ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/resources/streaming/StreamedQueryResource.java
```diff
@@ -540,6 +549,43 @@ private EndpointResponse handleScalablePushQuery(
...
+          serviceContext,
+          analysis,
+          configured,
+          transientQueryMetadata
+      );
+      // stop the response stream before sending the successful response.
+      queryStreamWriter.close();
+      return EndpointResponse.ok(queryStreamWriter);
+    } catch (final KsqlServerException e) {
+      throw new KsqlApiException(e.getMessage(), HttpResponseStatus.INTERNAL_SERVER_ERROR.code());
```
AlanConfluent, line 585: I'm confused why this gets wrapped in a KsqlApiException.  I assume it's so that we can return a 500 error rather than a 400, though `KsqlApiException` seems to be intended for 400 errors.  We're so inconsistent in all of our broader code about which exceptions we throw and how they map to error codes, I cannot recommend what to do here.  We should prioritize making all query related endpoints consistent @agavra .

AlanConfluent:
Very cool @vvcephei.  I like the intermeshing of the conventional pull and push worlds.  I think "doing what the user intends" is a great thing and this should make things easier to use. We should prioritize `EXPLAIN` because this will have entirely different performance characteristics than a conventional pull query.

ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/resources/streaming/StreamedQueryResource.java
vvcephei: That's an interesting idea. I think this would have to be a separate feature, though, since the semantics are different. For non-materialized tables, we'd have to read the whole topic (like we're doing here), but collect it in a data structure instead of emitting results as we go, and then compute the query and emit the results at the end.

ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/resources/streaming/StreamedQueryResource.java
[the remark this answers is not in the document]
vvcephei: I'll take another look. IIRC, this is a difference between the handling of push and pull queries. Since we decided to "hack" this feature in by reusing most of the code paths for transient queries, there's no "at the end" like there is for pull queries.

If we want it to be a clean implementation, I'd have to create a whole new stack of query-handling logic.

ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/resources/streaming/StreamedQueryResource.java
vvcephei: Interesting. I thought it was just intended to be able to specify the error code, from the constructor. Is there another exception type for 5xx ?

ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/resources/streaming/StreamedQueryResource.java
vvcephei: Good to hear! I just copied it without understanding why.

ksqldb-engine/src/main/java/io/confluent/ksql/engine/KsqlEngine.java
vvcephei: Yep!

ksqldb-engine/src/main/java/io/confluent/ksql/engine/KsqlEngine.java
vvcephei: I see. In retrospect, it probably makes more sense just to rethrow KsqlExceptions up to the API layer so it can return either a 4xx or a 5xx.

ksqldb-engine/src/main/java/io/confluent/ksql/engine/KsqlEngine.java
```diff
@@ -293,6 +316,186 @@ public TransientQueryMetadata executeTransientQuery(
...
+      throw new KsqlStatementException(
+          "Pull queries on streams are disabled. To create a push query on the stream,"
+              + " add EMIT CHANGES to the end. To enable pull queries on streams, set"
+              + " the " + KsqlConfig.KSQL_QUERY_STREAM_PULL_QUERY_ENABLED + " config to 'true'.",
+          statementOrig.getStatementText()
+      );
+    }
+
+    // stream pull query overrides: start from earliest, use one  thread,
+    // and use a tight commit interval for responsiveness.
```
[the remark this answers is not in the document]
agavra: sgtm

ksqldb-engine/src/main/java/io/confluent/ksql/engine/KsqlEngine.java
```diff
@@ -293,6 +316,186 @@ public TransientQueryMetadata executeTransientQuery(
...
+        admin,
+        topicDescription,
+        isolationLevel
+    );
+
+    // wait for the query to pass the endOffsets
+    while (!passedEndOffsets(admin, transientQueryMetadata, endOffsets)) {
+      if (transientQueryMetadata.getState() == State.ERROR) {
+        log.error("Stream pull query execution has failed");
+        throw new KsqlServerException("Server error");
```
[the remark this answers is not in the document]
agavra: I suppose I don't have much experience in this realm, so I'll defer to you :) my intuition tells me that a database is different from a typical backend server in that the errors are surfaced in a backend rather than a frontend but let's not block on this discussion

ksqldb-engine/src/main/java/io/confluent/ksql/engine/KsqlEngine.java
guozhangwang: Ack.

ksqldb-engine/src/main/java/io/confluent/ksql/engine/KsqlEngine.java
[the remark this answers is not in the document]
agavra: are these exposed when you `LIST QUERIES`? can we terminate them using `TERMINATE` (c.f. #7947) - I think we may have broken the 1:1 mapping from queryID to appID but that's the line of thought I had in mind

ksqldb-engine/src/main/java/io/confluent/ksql/engine/KsqlEngine.java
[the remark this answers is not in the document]
agavra: 👍

ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/resources/streaming/StreamedQueryResource.java
vvcephei: Ah, upon further investigation, the exception handling does need to be handled more broadly. Apparently, there's insufficient test coverage, and it actually looks like the HTTP/1 API turns all exceptions into 400s regardless of type.

I'll schedule a follow-on PR to address that.

ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/resources/streaming/StreamedQueryResource.java
AlanConfluent: Hmm, that's a good point.  If we outputted a series of "A": 1 and then "A":2 as it changed, that would violate the semantics of what we would expect from a state store lookup. I agree that would take more work so that can be another feature.  That said, it is slightly confusing what's queryable with pull queries and what's not.

ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/resources/streaming/StreamedQueryResource.java
AlanConfluent: Yep, we should prioritize that for next release.

AlanConfluent (approved):

ksqldb-engine/src/test/java/io/confluent/ksql/engine/KsqlEngineTest.java
```diff
@@ -1926,6 +1926,47 @@ public void shouldIgnoreLegacyDeleteTopicPartOfDropCommand() {
     verifyNoMoreInteractions(schemaRegistryClient);
   }
 
+  @Test
```
AlanConfluent, line 1929: It seems like we should test some of the logic for waiting to get past the appropriate offsets.

ksqldb-engine/src/main/java/io/confluent/ksql/engine/KsqlEngine.java
```diff
@@ -293,6 +316,186 @@ public TransientQueryMetadata executeTransientQuery(
...
+
+    } catch (final KsqlStatementException e) {
+      throw e;
+    } catch (final KsqlException e) {
+      // add the statement text to the KsqlException
+      throw new KsqlStatementException(e.getMessage(), statement.getStatementText(), e.getCause());
+    }
+  }
+
+  public void waitForStreamPullQuery(
```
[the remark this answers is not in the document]
AlanConfluent: Much of the logic in KsqlEngine is somewhat orchestration rather than core logic.  Maybe we should factor out the offset fetching and checking to another class under the pull query logic?

ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/resources/streaming/StreamedQueryResource.java
```diff
@@ -540,6 +549,43 @@ private EndpointResponse handleScalablePushQuery(
...
+    // Be sure to close the stream in the event of any error. It's safe to close it twice,
+    // which will happen if we call close inside the block, and then again when we exit the
+    // try-with-resources
+    try (QueryStreamWriter queryStreamWriter = new QueryStreamWriter(
+        transientQueryMetadata,
+        disconnectCheckInterval.toMillis(),
+        OBJECT_MAPPER,
+        connectionClosedFuture
+    )) {
+      ksqlEngine.waitForStreamPullQuery(
```
AlanConfluent, line 575: One critical thing that occurs to me is that this will block until the query is complete, but before it's actually begun writing the output to the output stream.  I think this relies on buffering all of the rows in transientQueryMetadata.rowqueue.  If it were to fill up, I think it might deadlock.

Also, we want to be able to be able to stream back results and not hold it all in memory, otherwise we'd be fairly limited.  We probably need the logic that determines when the query is complete to be a separate thread in the background that eventually closes that queue when it's complete.  That's similar to how we handle this with pull queries at the moment.

ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/resources/streaming/StreamedQueryResource.java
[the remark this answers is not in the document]
AlanConfluent: I think I understand why this is happening.  It's because QueryStreamWriter.write isn't called at all until after this returns, but you're wanting to complete it here.  I think this shouldn't be required.

ksqldb-engine/src/main/java/io/confluent/ksql/engine/KsqlEngine.java
```diff
@@ -293,6 +319,195 @@ public TransientQueryMetadata executeTransientQuery(
...
+    );
+
+    // wait for the query to pass the endOffsets
+    while (!passedEndOffsets(admin, transientQueryMetadata, endOffsets)) {
+      if (transientQueryMetadata.getState() == State.ERROR) {
+        log.error("Stream pull query execution has failed");
+        throw new KsqlServerException("Server error");
+      } else {
+        try {
+          Thread.sleep(50L);
```
AlanConfluent, line 418: One thing that occurs to me is that for the async endpoints, you're not going to be able to call this directly on the request thread.  For the reasons I mentioned elsewhere, I think you're going to have to do this polling in a background thread anyway to match the current invariants, so that seems like a reasonable solution.

AlanConfluent (changes requested):
Something occurred to me right after I approved this that I think we actually have to address.  Please take a look at my thread comments.

ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/resources/streaming/StreamedQueryResource.java
vvcephei: This is an excellent catch. Thank you.

After digging in again, I realized that the root of this evil was the way that I tried to wrap the waiting logic up into the engine and re-use the QueryStreamWriter from push queries. The "right" way to do it is to create a new xQueryStreamWriter (I called it `StreamPullQueryStreamWriter` for maximum confusion). Then, we can check for completion during writer's poll loop the way the pull queries do.

ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/resources/streaming/StreamedQueryResource.java
vvcephei: Yep, it's a good point. The feature request for stream pull queries mentioned "filling out the missing quadrant", but what you're pointing out is that there's still a missing quadrant: pull queries against non-materialized tables.

ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/resources/streaming/StreamedQueryResource.java
```diff
@@ -540,6 +548,64 @@ private EndpointResponse handleScalablePushQuery(
...
+        disconnectCheckInterval.toMillis(),
+        OBJECT_MAPPER,
+        connectionClosedFuture,
+        new StreamPullQueryCompletionCheck(ksqlEngine, streamPullQueryMetadata)
+    );
+
+    return EndpointResponse.ok(queryStreamWriter);
+  }
+
+  private static final class StreamPullQueryCompletionCheck implements Supplier<Boolean> {
```
vvcephei, line 577: Added some extra logic here so that once we're "complete", we continue to report "true" without any more API calls. And also so that we won't check more than every 100ms.

ksqldb-rest-client/src/main/java/io/confluent/ksql/rest/client/KsqlTarget.java
```diff
@@ -451,8 +451,10 @@ private KsqlRequest createKsqlRequest(
         if (begin != i) { // Ignore random newlines - the server can send these
           final Buffer sliced = buff.slice(begin, i);
           final Buffer tidied = StreamPublisher.toJsonMsg(sliced, true);
-          final StreamedRow row = deserialize(tidied, StreamedRow.class);
-          rows.add(row);
+          if (tidied.length() > 0) {
+            final StreamedRow row = deserialize(tidied, StreamedRow.class);
+            rows.add(row);
+          }
```
vvcephei, line 457: I happened to find a bug during the implementation. `toJsonMsg` in the prior line may trim everything out and return an empty string, which we can't deserialize, so we shouldn't try.

ksqldb-rest-client/src/main/java/io/confluent/ksql/rest/client/StreamPublisher.java
```diff
@@ -40,7 +40,7 @@ public static Buffer toJsonMsg(final Buffer responseLine, final boolean stripArr
         end -= 1;
       }
     }
-    if (responseLine.getByte(end) == (byte) ',') {
+    if (end > 0 && responseLine.getByte(end) == (byte) ',') {
```
vvcephei, line 43: Another small bug. The prior couple of modifications may strip off the square brackets. If those are the only thing in the buffer, we're left with an empty string, and we were getting an out-of-bounds exception here. Simple enough to avoid checking if an empty buffer ends in a comma.

vvcephei (commented):
Hey @AlanConfluent , I made another pass on the fix for the problem you pointed out above.

Offline, we discussed sticking to the original version of this PR and just moving the bounds check to a separate thread(pool). I actually had done that in my POC, so I have a little bit of an idea of how it plays out.

Although this approach moves some of the logic out of the KsqlEngine and into the endpoint, I think doing so will actually make it easier to implement for the other two endpoints.

This revision removes the duplication that was in my last commit by converging the stream pull query writer and the push query writer.

I might have to backpedal in the next PR, but I _think_ this is the approach I prefer for now. WDYT?

ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/resources/streaming/StreamedQueryResource.java
AlanConfluent: I suspect this will be something that we will want to factor out to the other endpoints in some manner if we continue to use roughly this approach with the others.

That said, this does a blocking call to kafka, so you probably will have to wrap this in a runnable to put on the blocking worker pool, since `/query-stream` doesn't allow blocking calls.  Not hard, but just wanting to point that out.

AlanConfluent (approved):
Thanks for making those changes.  It's looking good!

Nit: I think KsqlEngine is getting pretty large.  Might want to consider pulling some of the core logic (that's not orchestration) like calls to kafka for offsets and checking for passing the end into another class.  Can be in a followup.

vvcephei:
Thanks @AlanConfluent ! I agree with both of those feedbacks. I'll take care of them in the subsequent PRs.

Issue #8055, opened 2021-08-25
Windowed table joins fail on triggering record


Sullivan-Patrick:
**Describe the bug**
When trying to join two windowed tables, a query seems to start with no issue. However, upon trying to insert a triggering record, the query terminates with the following message:

`A serializer (io.confluent.ksql.logging.processing.LoggingSerializer) is not compatible to the actual key type (key type: org.apache.kafka.streams.kstream.Windowed). Change the default Serdes in StreamConfig or provide correct Serdes via method parameters.`

**To Reproduce**
On master (SHA f0b2751f90d7265e082049b7a748cee26fe56366), this can be re-created by running the following queries:

`CREATE STREAM x (i int) WITH (kafka_topic='x', value_format='JSON', partitions=1);`
`CREATE STREAM y (j int) WITH (kafka_topic='y', value_format='JSON', partitions=1);`
`CREATE TABLE tbl AS SELECT i, AVG(i) AS average FROM x WINDOW TUMBLING (SIZE 7 DAYS) GROUP BY i  EMIT CHANGES;`
`CREATE TABLE tbl2 AS SELECT j, AVG(j) AS average FROM y WINDOW TUMBLING (SIZE 7 DAYS) GROUP BY j EMIT CHANGES;`
`SELECT i, j FROM tbl JOIN tbl2 ON i = j EMIT CHANGES`
`INSERT INTO x (i) VALUES (3);`
`INSERT INTO y (j) VALUES (3);`

**Expected behavior**
The query is not started, or the join works.

**Actual behaviour**
The query is started, but upon a triggering record being entered, the above exception is thrown.

Sullivan-Patrick:
For reference: in order to support joins on windowed tables going forward, it'll be necessary to bump the execution step version for `WindowedTableSource`, to support a new topology which materializes partition and offset. I actually made these changes, then walked them back since we don't currently support said joins.

https://github.com/confluentinc/ksql/pull/7990 shows the changes being made (I don't actually change `WindowedTableSource` to materialize the partition/offset information here, for that you can copy the changes I made to unwindowed tables)
https://github.com/confluentinc/ksql/pull/8111 shows the changes being reversed.

PR #8072, opened 2021-08-30
chore: Materialize only partition and offset information in V2 table source steps

Sullivan-Patrick:
### Description 
As a result of https://github.com/confluentinc/ksql/pull/7990, we now materialize all key and pseudo columns in the value schema in order to allow for `ROWPARTITION` and `ROWOFFSET` in joins involving tables. This is wasteful as we are able to access timestamp and key columns after the materialization step, this PR adds just the partition and offset information without also storing the information we don't need.

Before:
```
   Sub-topology: 0
      --> KTABLE-SOURCE-0000000002
      --> KTABLE-MAPVALUES-0000000003
      <-- KSTREAM-SOURCE-0000000001
    Processor: KTABLE-MAPVALUES-0000000003 (stores: [KsqlTopic-Reduce])
      --> KTABLE-TRANSFORMVALUES-0000000004
      <-- KTABLE-SOURCE-0000000002
    Processor: KTABLE-TRANSFORMVALUES-0000000004 (stores: [])
      --> Project
      <-- KTABLE-MAPVALUES-0000000003
    Processor: Project (stores: [])
      --> KTABLE-TOSTREAM-0000000006
      <-- KTABLE-TRANSFORMVALUES-0000000004
    Processor: KTABLE-TOSTREAM-0000000006 (stores: [])
      --> KSTREAM-SINK-0000000007
      <-- Project
    Sink: KSTREAM-SINK-0000000007 (topic: TT)
      <-- KTABLE-TOSTREAM-0000000006
```

After:
```
   Sub-topology: 0
      --> KTABLE-SOURCE-0000000002
      --> KTABLE-TRANSFORMVALUES-0000000003
      <-- KSTREAM-SOURCE-0000000001
    Processor: KTABLE-TRANSFORMVALUES-0000000003 (stores: [])
      --> KTABLE-MAPVALUES-0000000004
      <-- KTABLE-SOURCE-0000000002
    Processor: KTABLE-MAPVALUES-0000000004 (stores: [KsqlTopic-Reduce])
      --> KTABLE-TRANSFORMVALUES-0000000005
      <-- KTABLE-TRANSFORMVALUES-0000000003
    Processor: KTABLE-TRANSFORMVALUES-0000000005 (stores: [])
      --> Project
      <-- KTABLE-MAPVALUES-0000000004
    Processor: Project (stores: [])
      --> KTABLE-TOSTREAM-0000000007
      <-- KTABLE-TRANSFORMVALUES-0000000005
    Processor: KTABLE-TOSTREAM-0000000007 (stores: [])
      --> KSTREAM-SINK-0000000008
      <-- Project
    Sink: KSTREAM-SINK-0000000008 (topic: TT)
      <-- KTABLE-TOSTREAM-0000000007
```

### Testing done 
Changed unit tests to expect new behavior

### Reviewer checklist

### DIFF
--- a/ksqldb-common/src/main/java/io/confluent/ksql/schema/ksql/LogicalSchema.java
+++ b/ksqldb-common/src/main/java/io/confluent/ksql/schema/ksql/LogicalSchema.java
@@ -157,6 +157,7 @@ public final class LogicalSchema {
 
   /**
    * Remove pseudo and key columns from the value schema, according to the pseudocolumn version
+   *
    * @param pseudoColumnVersion the version of pseudocolumns to evaluate against
    * @return the new schema with the columns removed
    */
@@ -198,7 +199,6 @@ public final class LogicalSchema {
    * {@code ROWOFFSET}.
    *
    * @param pseudoColumnVersion the version of pseudocolumns to evaluate against
-   *
    * @return the new schema with the columns removed
    */
   public LogicalSchema withPseudoColumnsToMaterialize(final int pseudoColumnVersion) {
@@ -371,9 +371,9 @@ public final class LogicalSchema {
     builder.addAll(keyColumns);
     builder.addAll(nonPseudoAndKeyCols);
 
-    //put all pseudocolumns which are not able to be accessed via processorContext after a join,
-    //and therefore require materialization, in a conditional block corresponding to their
-    //pseudocolumn version number
+    //put pseudocolumns which are not able to be accessed when getting value from an upstream
+    //state store, and therefore require materialization, in a conditional block corresponding to
+    //its pseudocolumn version number
     if (pseudoColumnVersion >= ROWPARTITION_ROWOFFSET_PSEUDOCOLUMN_VERSION) {
       builder.add(Column.of(ROWPARTITION_NAME, ROWPARTITION_TYPE, VALUE, valueIndex++));
       builder.add(Column.of(ROWOFFSET_NAME, ROWOFFSET_TYPE, VALUE, valueIndex++));
--- a/ksqldb-common/src/main/java/io/confluent/ksql/schema/ksql/LogicalSchema.java
+++ b/ksqldb-common/src/main/java/io/confluent/ksql/schema/ksql/LogicalSchema.java
@@ -199,7 +199,7 @@ public final class LogicalSchema {
    * {@code ROWOFFSET}.
    *
    * @param pseudoColumnVersion the version of pseudocolumns to evaluate against
-   * @return the new schema with the columns removed
+   * @return the new schema with the columns added
    */
   public LogicalSchema withPseudoColumnsToMaterialize(final int pseudoColumnVersion) {
     return rebuildWithPseudoColumnsToMaterialize(pseudoColumnVersion);
@@ -371,9 +371,8 @@ public final class LogicalSchema {
     builder.addAll(keyColumns);
     builder.addAll(nonPseudoAndKeyCols);
 
-    //put pseudocolumns which are not able to be accessed when getting value from an upstream
-    //state store, and therefore require materialization, in a conditional block corresponding to
-    //its pseudocolumn version number
+    // add pseudocolumns which are not guaranteed to be accessible from downstream
+    // processor contexts to the schema for materialization in a state store
     if (pseudoColumnVersion >= ROWPARTITION_ROWOFFSET_PSEUDOCOLUMN_VERSION) {
       builder.add(Column.of(ROWPARTITION_NAME, ROWPARTITION_TYPE, VALUE, valueIndex++));
       builder.add(Column.of(ROWOFFSET_NAME, ROWOFFSET_TYPE, VALUE, valueIndex++));
--- a/ksqldb-common/src/main/java/io/confluent/ksql/schema/ksql/LogicalSchema.java
+++ b/ksqldb-common/src/main/java/io/confluent/ksql/schema/ksql/LogicalSchema.java
@@ -192,6 +192,20 @@ public final class LogicalSchema {
   }
 
   /**
+   * Adds just the pseudocolumns that must be materialized as part of a join involving a table.
+   * As of pseudocolumn version {@code SystemColumns.ROWPARTITION_ROWOFFSET_PSEUDOCOLUMN_VERSION},
+   * there are only two pseudocolumns that fall under this definition: {@code ROWPARTITION} and
+   * {@code ROWOFFSET}.
+   *
+   * @param pseudoColumnVersion the version of pseudocolumns to evaluate against
+   *
+   * @return the new schema with the columns removed
+   */
+  public LogicalSchema withPseudoColumnsToMaterialize(final int pseudoColumnVersion) {
+    return rebuildWithPseudoColumnsToMaterialize(pseudoColumnVersion);
+  }
+
+  /**
    * @param columnName the column name to check
    * @return {@code true} if the column matches the name of any key column.
    */
@@ -332,6 +346,38 @@ public final class LogicalSchema {
 
     builder.addAll(keyColumns);
     builder.addAll(nonPseudoAndKeyCols);
+
+    return new LogicalSchema(builder.build());
+  }
+
+
+  /**
+   * Rebuilds schema with only certain pseudocolumns to materialize in an intermittent store
+   * @param pseudoColumnVersion indicates which set of pseudocolumns should be used
+   * @return the LogicalSchema created, with the corresponding pseudocolumns
+   */
+  private LogicalSchema rebuildWithPseudoColumnsToMaterialize(final int pseudoColumnVersion) {
+    final Map<Namespace, List<Column>> byNamespace = byNamespace();
+
+    final ImmutableList.Builder<Column> builder = ImmutableList.builder();
+
+    final List<Column> keyColumns = keyColumns(byNamespace);
+
+    final List<Column> nonPseudoAndKeyCols = nonPseudoAndKeyColsAsValueCols(
+        byNamespace, pseudoColumnVersion);
+
+    int valueIndex = nonPseudoAndKeyCols.size();
+
+    builder.addAll(keyColumns);
+    builder.addAll(nonPseudoAndKeyCols);
+
+    //put all pseudocolumns which are not able to be accessed via processorContext after a join,
+    //and therefore require materialization, in a conditional block corresponding to their
+    //pseudocolumn version number
+    if (pseudoColumnVersion >= ROWPARTITION_ROWOFFSET_PSEUDOCOLUMN_VERSION) {
+      builder.add(Column.of(ROWPARTITION_NAME, ROWPARTITION_TYPE, VALUE, valueIndex++));
+      builder.add(Column.of(ROWOFFSET_NAME, ROWOFFSET_TYPE, VALUE, valueIndex++));
+    }
 
     return new LogicalSchema(builder.build());
   }
--- a/ksqldb-common/src/main/java/io/confluent/ksql/schema/ksql/SystemColumns.java
+++ b/ksqldb-common/src/main/java/io/confluent/ksql/schema/ksql/SystemColumns.java
@@ -114,6 +114,10 @@ public final class SystemColumns {
     return systemColumnNames(CURRENT_PSEUDOCOLUMN_VERSION_NUMBER);
   }
 
+  public static Boolean mustBeMaterializedForTableJoins(ColumnName columnName) {
+    return PseudoColumns.MUST_BE_MATERIALIZED_FOR_TABLE_JOINS.contains(columnName);
+  }
+
   private static Set<ColumnName> buildColumns(final int pseudoColumnVersion) {
     return ImmutableSet.<ColumnName>builder()
         .addAll(pseudoColumnNames(pseudoColumnVersion))
@@ -144,6 +148,12 @@ public final class SystemColumns {
         ImmutableMap.of(
             0, VERSION_ZERO_NAMES,
             1, VERSION_ONE_NAMES
+        );
+
+    private static final Set<ColumnName> MUST_BE_MATERIALIZED_FOR_TABLE_JOINS =
+        ImmutableSet.of(
+            ROWPARTITION_NAME,
+            ROWOFFSET_NAME
         );
 
     private static Set<ColumnName> getPseudoColumnNamesByVersion(final int pseudoColumnVersion) {
--- a/ksqldb-common/src/main/java/io/confluent/ksql/schema/ksql/SystemColumns.java
+++ b/ksqldb-common/src/main/java/io/confluent/ksql/schema/ksql/SystemColumns.java
@@ -114,7 +114,7 @@ public final class SystemColumns {
     return systemColumnNames(CURRENT_PSEUDOCOLUMN_VERSION_NUMBER);
   }
 
-  public static boolean mustBeMaterializedForTableJoins(ColumnName columnName) {
+  public static boolean mustBeMaterializedForTableJoins(final ColumnName columnName) {
     return PseudoColumns.MUST_BE_MATERIALIZED_FOR_TABLE_JOINS.contains(columnName);
   }
 
--- a/ksqldb-common/src/main/java/io/confluent/ksql/schema/ksql/SystemColumns.java
+++ b/ksqldb-common/src/main/java/io/confluent/ksql/schema/ksql/SystemColumns.java
@@ -114,7 +114,7 @@ public final class SystemColumns {
     return systemColumnNames(CURRENT_PSEUDOCOLUMN_VERSION_NUMBER);
   }
 
-  public static Boolean mustBeMaterializedForTableJoins(ColumnName columnName) {
+  public static boolean mustBeMaterializedForTableJoins(ColumnName columnName) {
     return PseudoColumns.MUST_BE_MATERIALIZED_FOR_TABLE_JOINS.contains(columnName);
   }
 
--- a/ksqldb-streams/src/main/java/io/confluent/ksql/execution/streams/SourceBuilder.java
+++ b/ksqldb-streams/src/main/java/io/confluent/ksql/execution/streams/SourceBuilder.java
@@ -22,6 +22,7 @@ import static io.confluent.ksql.execution.streams.SourceBuilderUtils.getKeySerde
 import static io.confluent.ksql.execution.streams.SourceBuilderUtils.getRegisterCallback;
 import static io.confluent.ksql.execution.streams.SourceBuilderUtils.getValueSerde;
 import static io.confluent.ksql.execution.streams.SourceBuilderUtils.getWindowedKeySerde;
+import static java.util.Objects.requireNonNull;
 
 import io.confluent.ksql.GenericKey;
 import io.confluent.ksql.GenericRow;
@@ -32,7 +33,9 @@ import io.confluent.ksql.execution.plan.PlanInfo;
 import io.confluent.ksql.execution.plan.SourceStep;
 import io.confluent.ksql.execution.plan.WindowedTableSource;
 import io.confluent.ksql.execution.runtime.RuntimeBuildContext;
+import io.confluent.ksql.schema.ksql.LogicalSchema;
 import io.confluent.ksql.schema.ksql.PhysicalSchema;
+import io.confluent.ksql.schema.ksql.SystemColumns;
 import io.confluent.ksql.serde.FormatInfo;
 import io.confluent.ksql.serde.KeyFormat;
 import io.confluent.ksql.serde.SerdeFeatures;
@@ -45,7 +48,10 @@ import org.apache.kafka.common.utils.Bytes;
 import org.apache.kafka.streams.kstream.Consumed;
 import org.apache.kafka.streams.kstream.KTable;
 import org.apache.kafka.streams.kstream.Materialized;
+import org.apache.kafka.streams.kstream.ValueTransformerWithKey;
+import org.apache.kafka.streams.kstream.ValueTransformerWithKeySupplier;
 import org.apache.kafka.streams.kstream.Windowed;
+import org.apache.kafka.streams.processor.ProcessorContext;
 import org.apache.kafka.streams.state.KeyValueStore;
 
 final class SourceBuilder extends SourceBuilderBase {
@@ -81,7 +87,9 @@ final class SourceBuilder extends SourceBuilderBase {
     final boolean forceMaterialization = !planInfo.isRepartitionedInPlan(streamSource);
 
     final KTable<K, GenericRow> transformed = source.transformValues(
-        new AddKeyAndPseudoColumns<>(keyGenerator, streamSource.getPseudoColumnVersion()));
+        new AddPseudoColumnsToMaterialize<>(streamSource.getPseudoColumnVersion()));
+
+    final KTable<K, GenericRow> toPotentiallyMaterialize;
 
     if (forceMaterialization) {
       // add this identity mapValues call to prevent the source-changelog
@@ -89,17 +97,19 @@ final class SourceBuilder extends SourceBuilderBase {
       // be enabled because we cannot require symmetric serialization between
       // producer and KSQL (see https://issues.apache.org/jira/browse/KAFKA-10179
       // and https://github.com/confluentinc/ksql/issues/5673 for more details)
-      return transformed.mapValues(row -> row, materialized);
-
+      toPotentiallyMaterialize = transformed.mapValues(row -> row, materialized);
     } else {
       // if we know this table source is repartitioned later in the topology,
       // we do not need to force a materialization at this source step since the
       // re-partitioned topic will be used for any subsequent state stores, in lieu
       // of the original source topic, thus avoiding the issues above.
       // See https://github.com/confluentinc/ksql/issues/6650
-      return transformed.mapValues(row -> row);
+      toPotentiallyMaterialize = transformed.mapValues(row -> row);
     }
 
+    return toPotentiallyMaterialize.transformValues(new AddRemainingPseudoAndKeyCols<>(
+        keyGenerator,
+        streamSource.getPseudoColumnVersion()));
   }
 
   @Override
@@ -138,7 +148,7 @@ final class SourceBuilder extends SourceBuilderBase {
       final String stateStoreName
   ) {
 
-    final PhysicalSchema physicalSchema = getPhysicalSchemaWithKeyAndPseudoCols(source);
+    final PhysicalSchema physicalSchema = getPhysicalSchemaWithPseudoColumnsToMaterialize(source);
 
     final QueryContext queryContext = addMaterializedContext(source);
 
@@ -221,6 +231,28 @@ final class SourceBuilder extends SourceBuilderBase {
     );
   }
 
+  private static PhysicalSchema getPhysicalSchemaWithPseudoColumnsToMaterialize(
+      final SourceStep<?> streamSource) {
+
+    FormatInfo f = streamSource.getFormats().getKeyFormat();
+    SerdeFeatures s = streamSource.getFormats().getKeyFeatures();
+    KeyFormat k = streamSource instanceof WindowedTableSource
+        ? KeyFormat.windowed(f, s, ((WindowedTableSource) streamSource).getWindowInfo())
+        : KeyFormat.nonWindowed(f, s);
+
+    Formats format = of(k, streamSource.getFormats().getValueFormat());
+
+    LogicalSchema withPseudosToMaterialize
+        = streamSource.getSourceSchema().withPseudoColumnsToMaterialize(
+        streamSource.getPseudoColumnVersion());
+
+    return PhysicalSchema.from(
+        withPseudosToMaterialize,
+        format.getKeyFeatures(),
+        format.getValueFeatures()
+    );
+  }
+
   //todo: put this logic into TableSource and WindowedTableSource
   private static Formats of(final KeyFormat keyFormat, final FormatInfo valueFormat) {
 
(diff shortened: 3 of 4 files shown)


ksqldb-streams/src/main/java/io/confluent/ksql/execution/streams/SourceBuilder.java
```diff
@@ -221,6 +231,28 @@ private static PhysicalSchema getPhysicalSchemaWithKeyAndPseudoCols(
...
   }
 
+  private static PhysicalSchema getPhysicalSchemaWithPseudoColumnsToMaterialize(
+      final SourceStep<?> streamSource) {
+
+    FormatInfo f = streamSource.getFormats().getKeyFormat();
+    SerdeFeatures s = streamSource.getFormats().getKeyFeatures();
+    KeyFormat k = streamSource instanceof WindowedTableSource
+        ? KeyFormat.windowed(f, s, ((WindowedTableSource) streamSource).getWindowInfo())
+        : KeyFormat.nonWindowed(f, s);
```
Sullivan-Patrick, line 241: This should, in it's current state, never be called on `WindowedTableSource`s. While it's probably a bug that it isn't, [considering joins on windowed table sources don't even work](https://github.com/confluentinc/ksql/issues/8055), I think I should remove this for now to reduce code complexity. WDYT @vcrfxia?

ksqldb-streams/src/test/java/io/confluent/ksql/execution/streams/SourceBuilderTest.java
```diff
@@ -580,15 +584,20 @@ public void shouldBuildTableWithCorrectStoreName() {
   }
 
   @SuppressWarnings("unchecked")
-  private <K> ValueTransformerWithKey<K, GenericRow, GenericRow> getTransformerFromTableSource(
+  private <K> List<ValueTransformerWithKey<K, GenericRow, GenericRow>> getTransformersFromTableSource(
```
Sullivan-Patrick, line 587: Not sure if this is the ideal way of applying multiple transformers? It's extensible, but I doubt we'll be using more transformers anytime soon. I don't think it reads _too_ badly, but it's definitely worse than before.

ksqldb-common/src/main/java/io/confluent/ksql/schema/ksql/LogicalSchema.java
```diff
@@ -336,6 +350,38 @@ private LogicalSchema rebuildWithoutPseudoAndKeyColsInValue(final int pseudoColu
...
+        byNamespace, pseudoColumnVersion);
+
+    int valueIndex = nonPseudoAndKeyCols.size();
+
+    builder.addAll(keyColumns);
+    builder.addAll(nonPseudoAndKeyCols);
+
+    //put pseudocolumns which are not able to be accessed when getting value from an upstream
+    //state store, and therefore require materialization, in a conditional block corresponding to
+    //its pseudocolumn version number
```
Sullivan-Patrick, line 376: `its`

ksqldb-streams/src/main/java/io/confluent/ksql/execution/streams/SourceBuilder.java
```diff
@@ -221,6 +233,28 @@ private static PhysicalSchema getPhysicalSchemaWithKeyAndPseudoCols(
...
   }
 
+  private static PhysicalSchema getPhysicalSchemaWithPseudoColumnsToMaterialize(
+      final SourceStep<?> streamSource) {
+
+    final FormatInfo f = streamSource.getFormats().getKeyFormat();
+    final SerdeFeatures s = streamSource.getFormats().getKeyFeatures();
+    final KeyFormat k = streamSource instanceof WindowedTableSource
+        ? KeyFormat.windowed(f, s, ((WindowedTableSource) streamSource).getWindowInfo())
+        : KeyFormat.nonWindowed(f, s);
```
Sullivan-Patrick, line 243: This should, in it's current state, never be called on WindowedTableSources. While it's probably a bug that it isn't, considering joins on windowed table sources don't even work, I think I should remove this for now to reduce code complexity. WDYT @vcrfxia?

ksqldb-streams/src/main/java/io/confluent/ksql/execution/streams/SourceBuilder.java
```diff
@@ -81,25 +89,29 @@ public static SourceBuilder instance() {
...
 
     if (forceMaterialization) {
       // add this identity mapValues call to prevent the source-changelog
       // optimization in kafka streams - we don't want this optimization to
       // be enabled because we cannot require symmetric serialization between
       // producer and KSQL (see https://issues.apache.org/jira/browse/KAFKA-10179
       // and https://github.com/confluentinc/ksql/issues/5673 for more details)
-      return transformed.mapValues(row -> row, materialized);
-
+      toPotentiallyMaterialize = transformed.mapValues(row -> row, materialized);
```
vcrfxia, line 102: We don't need a separate mapValues() step for materialization anymore -- we can just materialize the result of the transformValues() above directly.

ksqldb-streams/src/main/java/io/confluent/ksql/execution/streams/SourceBuilder.java
```diff
@@ -81,25 +89,29 @@ public static SourceBuilder instance() {
     final boolean forceMaterialization = !planInfo.isRepartitionedInPlan(streamSource);
 
     final KTable<K, GenericRow> transformed = source.transformValues(
-        new AddKeyAndPseudoColumns<>(keyGenerator, streamSource.getPseudoColumnVersion()));
+        new AddPseudoColumnsToMaterialize<>(streamSource.getPseudoColumnVersion()));
+
+    final KTable<K, GenericRow> toPotentiallyMaterialize;
```
vcrfxia, line 94:
```suggestion
    final KTable<K, GenericRow> maybeMaterialized;
```

ksqldb-streams/src/main/java/io/confluent/ksql/execution/streams/SourceBuilder.java
```diff
@@ -81,25 +89,29 @@ public static SourceBuilder instance() {
...
       // if we know this table source is repartitioned later in the topology,
       // we do not need to force a materialization at this source step since the
       // re-partitioned topic will be used for any subsequent state stores, in lieu
       // of the original source topic, thus avoiding the issues above.
       // See https://github.com/confluentinc/ksql/issues/6650
-      return transformed.mapValues(row -> row);
+      toPotentiallyMaterialize = transformed.mapValues(row -> row);
     }
 
+    return toPotentiallyMaterialize.transformValues(new AddRemainingPseudoAndKeyCols<>(
```
vcrfxia, line 112: The way that this is setup right now, when `forceMaterialization` is false we'll still have two transformValues() steps rather than one even though a single transformValues() would suffice. It's more performant to have a single one though it's also nice that the Streams topology doesn't vary based on whether we happen to be materializing or not. I could see us going either way on this decision... let's leave it as is for now and potentially revisit before releasing this feature.

ksqldb-streams/src/main/java/io/confluent/ksql/execution/streams/SourceBuilder.java
```diff
@@ -221,6 +233,28 @@ private static PhysicalSchema getPhysicalSchemaWithKeyAndPseudoCols(
     );
   }
 
+  private static PhysicalSchema getPhysicalSchemaWithPseudoColumnsToMaterialize(
```
vcrfxia, line 236: This method is basically identical to `getPhysicalSchemaWithKeyAndPseudoCols()`. Can we de-dup?

ksqldb-streams/src/main/java/io/confluent/ksql/execution/streams/SourceBuilder.java
```diff
@@ -221,6 +233,28 @@ private static PhysicalSchema getPhysicalSchemaWithKeyAndPseudoCols(
...
+      final SourceStep<?> streamSource) {
+
+    final FormatInfo f = streamSource.getFormats().getKeyFormat();
```
vcrfxia, line 239: Update these variable names to be more informative.

ksqldb-streams/src/main/java/io/confluent/ksql/execution/streams/SourceBuilder.java
vcrfxia: If this block of code is not meant to be called on windowed table sources, then I'd have the top of the method throw an exception if a windowed table source is encountered and then simplify the body of the code with the assumption that the method is not called with windowed table sources.

ksqldb-streams/src/main/java/io/confluent/ksql/execution/streams/SourceBuilder.java
```diff
@@ -232,4 +266,116 @@ private static Formats of(final KeyFormat keyFormat, final FormatInfo valueForma
...
+        @Override
+        public GenericRow transform(final K key, final GenericRow row) {
+          if (row == null) {
+            return row;
+          }
+
+          final int numPseudoColumns = SystemColumns
+              .pseudoColumnNames(pseudoColumnVersion).size();
+
+          row.ensureAdditionalCapacity(numPseudoColumns - 1);
```
vcrfxia, line 297: Why -1? This does not look like it is extensible if additional pseudocolumns are added later. It would be better for the logic for which pseudocolumns need to be materialized to be centralized into a single place, such as where the pseudocolumns are defined.

ksqldb-common/src/main/java/io/confluent/ksql/schema/ksql/LogicalSchema.java
```diff
@@ -191,6 +192,19 @@ public LogicalSchema withKeyColsOnly() {
...
   }
 
+  /**
+   * Adds just the pseudocolumns that must be materialized as part of a join involving a table.
+   * As of pseudocolumn version {@code SystemColumns.ROWPARTITION_ROWOFFSET_PSEUDOCOLUMN_VERSION},
+   * there are only two pseudocolumns that fall under this definition: {@code ROWPARTITION} and
+   * {@code ROWOFFSET}.
+   *
+   * @param pseudoColumnVersion the version of pseudocolumns to evaluate against
+   * @return the new schema with the columns removed
```
vcrfxia, line 202: Why are columns being removed? I thought they were being added?

ksqldb-common/src/main/java/io/confluent/ksql/schema/ksql/LogicalSchema.java
```diff
@@ -336,6 +350,38 @@ private LogicalSchema rebuildWithoutPseudoAndKeyColsInValue(final int pseudoColu
...
+
+    final List<Column> nonPseudoAndKeyCols = nonPseudoAndKeyColsAsValueCols(
+        byNamespace, pseudoColumnVersion);
+
+    int valueIndex = nonPseudoAndKeyCols.size();
+
+    builder.addAll(keyColumns);
+    builder.addAll(nonPseudoAndKeyCols);
+
+    //put pseudocolumns which are not able to be accessed when getting value from an upstream
```
vcrfxia, line 374:
```suggestion
    // add pseudocolumns which are not guaranteed to be accessible from downstream processor contexts to the schema for materialization in a state store
```

ksqldb-common/src/main/java/io/confluent/ksql/schema/ksql/LogicalSchema.java
```diff
@@ -336,6 +350,38 @@ private LogicalSchema rebuildWithoutPseudoAndKeyColsInValue(final int pseudoColu
...
+    //state store, and therefore require materialization, in a conditional block corresponding to
```
vcrfxia, line 375: There is no need to describe "in a conditional block corresponding to its pseudocolumn version number" in the comment since the reader can see that from the code itself. In other words, comments are valuable if they make reading the code easier, but too many comments (especially redundant ones) can be harmful if they make reading the code harder :)

ksqldb-streams/src/main/java/io/confluent/ksql/execution/streams/SourceBuilder.java
```diff
@@ -232,4 +266,116 @@ private static Formats of(final KeyFormat keyFormat, final FormatInfo valueForma
...
+
+          //remove pseudocolumns we previously materialized so we can add them back in correct order
+
+          //ensure extra capacity equal to number of pseudoColumns which we haven't materialized
+          final int pseudoColumnsToAdd = SystemColumns.pseudoColumnNames(
+              SystemColumns.ROWTIME_PSEUDOCOLUMN_VERSION).size();
+
+          row.ensureAdditionalCapacity(pseudoColumnsToAdd);
+
+          //calculate number of user columns and
```
vcrfxia, line 354: Incomplete comment?

ksqldb-streams/src/main/java/io/confluent/ksql/execution/streams/SourceBuilder.java
```diff
@@ -232,4 +266,116 @@ private static Formats of(final KeyFormat keyFormat, final FormatInfo valueForma
...
+          final Set<ColumnName> columnNames = SystemColumns.pseudoColumnNames(pseudoColumnVersion);
+          final int totalPseudoColumns = columnNames.size();
+          final int pseudoColumnsToShift = totalPseudoColumns - pseudoColumnsToAdd;
+          final int numUserColumns = row.size() - pseudoColumnsToShift;
+
+          Object toShift = processorContext.timestamp();
```
vcrfxia, line 360: This pattern of shifting a single column works for now but does not extend nicely if we add more pseudocolumns in the future. I think it'd be cleaner to simply copy out the relevant pseudocolumns and then put them back in in the right places. That said, this can always be updated later so don't worry about it if you've got other things to work on.

Either way, can we add a comment to describe what this code is doing? Takes a while to decipher otherwise.

ksqldb-streams/src/test/java/io/confluent/ksql/execution/streams/SourceBuilderTest.java
vcrfxia: I think it's fine. This is just for these unit tests anyway.

ksqldb-streams/src/test/java/io/confluent/ksql/execution/streams/SourceBuilderTest.java
```diff
@@ -603,10 +611,22 @@ public void shouldBuildTableWithCorrectStoreName() {
     return transformer;
   }
 
+  private GenericRow applyAllTransformers(
+      final GenericKey key,
+      final List<ValueTransformerWithKey<GenericKey, GenericRow, GenericRow>> transformerList
+  ) {
+    GenericRow last = row;
```
vcrfxia, line 618: Let's pass `row` in as a param in order to make this method static.

ksqldb-streams/src/test/java/io/confluent/ksql/execution/streams/SourceBuilderTest.java
```diff
@@ -645,6 +665,7 @@ private void givenMultiColumnSourceTable() {
   }
 
   private <K> void givenConsumed(final Consumed<K, GenericRow> consumed, final Serde<K> keySerde) {
+    when(consumedWindowed.withValueSerde(any())).thenReturn(consumedWindowed);
```
vcrfxia, line 668: I don't understand -- how come we're passing in `consumed` as a parameter but mocking behavior on the class member `consumedWindowed` instead?

vcrfxia (commented):
Thanks @Sullivan-Patrick -- awesome to see that this PR is nicely contained now that the execution step versioning is complete! Comments inline. Can you also update the PR title to follow conventional commits, and clarify what the phrase "we are able to access timestamp and key columns after the materialization step" in the PR description means? I think that phrasing is rather unclear for people who don't have context for what this PR is trying to achieve.

Also, re: testing -- did you run these changes against QTT with the new pseudocolumn versioning, in addition to updating the unit tests? It'd be good to check as we're going that we aren't regressing on the new pseudocolumn behavior.

ksqldb-streams/src/main/java/io/confluent/ksql/execution/streams/SourceBuilder.java
Sullivan-Patrick: The method takes a `SourceStep`, so it would be equally wrong to give it a `StreamSource` or `WindowedStreamSource`. The only reason to explicitly disallow `WindowedTableSource` would be because we don't work with the stream steps in this file- I think throwing an exception on a `WindowedTableSource` might confuse someone down the line, as I think the method could eventually be used with it, and explicitly adding an exception might give someone the idea that there is a greater reason for not accepting `WindowedTableSource`s. Because of this, maybe I could just make the method accept a `TableSource` and cast `source` where it's called to a `TableSource{V2}`. Do you agree with my train of thought?

ksqldb-streams/src/main/java/io/confluent/ksql/execution/streams/SourceBuilder.java
Sullivan-Patrick: Sounds good, I'd be happy to discuss this as you see fit.

ksqldb-streams/src/main/java/io/confluent/ksql/execution/streams/SourceBuilder.java
Sullivan-Patrick: If we're forcing a materialization above (outside of the `forceMaterialization` block, in the `transformValues()` step itself I assume), then should I get rid of this if/else block entirely, and make the materialization in the `transformValues()` step dependent on `forceMaterialization?

ksqldb-common/src/main/java/io/confluent/ksql/schema/ksql/LogicalSchema.java
Sullivan-Patrick: I agree, this felt clunky when I wrote it but I wanted to be extra explicit that people don't add new pseudocolumns in the future that correspond with a different number into that block. Now that I'm thinking about it, I think that detail is already clear from how we've named `ROWPARTITION_ROWOFFSET_PSEUDOCOLUMN_VERSION`.

ksqldb-streams/src/main/java/io/confluent/ksql/execution/streams/SourceBuilder.java
Sullivan-Patrick: Committed, but can you explain why you think this is better?

ksqldb-streams/src/main/java/io/confluent/ksql/execution/streams/SourceBuilder.java
Sullivan-Patrick: I'll add a comment for now and a note to re-visit this.

ksqldb-streams/src/main/java/io/confluent/ksql/execution/streams/SourceBuilder.java
Sullivan-Patrick: I could add logic into `SystemColumns` which divides the set corresponding to a given `pseudoColumnVersion` into columns which may need to be materialized for access in downstream processor contexts vs those which are not. Is this what you had in mind?

ksqldb-streams/src/main/java/io/confluent/ksql/execution/streams/SourceBuilder.java
vcrfxia: You're still going to need the if-else block in order to decide which method variation to call on transformValues(), i.e., to decide whether to materialize or not, but you can pull logic into/out of the if-else block as needed to get this to work.

ksqldb-streams/src/main/java/io/confluent/ksql/execution/streams/SourceBuilder.java
vcrfxia: That's conceptually what I had in mind but I suspect there are probably cleaner ways to implement it. For example, you could have the SystemColumns class store of list of pseudocolumn objects/structs where each one has a name and a boolean that specifies whether the column should be materialized or not.

ksqldb-streams/src/main/java/io/confluent/ksql/execution/streams/SourceBuilder.java
vcrfxia: 1. This variable is assigned after potential materialization, not before. `maybeMaterialized` conveys this, while `toPotentiallyMaterialize` conveys the opposite.
2. It's shorter :)

ksqldb-streams/src/main/java/io/confluent/ksql/execution/streams/SourceBuilder.java
Sullivan-Patrick: So I remember why I didn't make a solution with a list in the first place: I can't just use `appendAll()` after building the list, as `GenericRow` doesn't expose a `remove()` function or anything like that- also, since trying to use `set()` with an in-capacity but not yet populated index throws an out of bounds exception, I'd need to iterate an add the first n pseudocolumns, where n is the number of pseudocolumns which do need to be materialized and were therefore already in the row, by using `set()`, then appending the last ones. I think my method ends up being more readable than this, at least with a comment. Thoughts?

ksqldb-streams/src/main/java/io/confluent/ksql/execution/streams/SourceBuilder.java
Sullivan-Patrick: Makes total sense in the context of your next comment. However, given how I changed the code to materialize on `transform`, I left it as `toPotentiallyMaterialize` (see next push)

ksqldb-streams/src/main/java/io/confluent/ksql/execution/streams/SourceBuilder.java
```diff
@@ -91,7 +91,7 @@ public static SourceBuilder instance() {
     final KTable<K, GenericRow> toPotentiallyMaterialize;
 
     if (forceMaterialization) {
-      // add this identity mapValues call to prevent the source-changelog
+      // materialize to prevent the source-changelog
```
Sullivan-Patrick, line 94: Not sure if this is necessarily accurate now, as we're just as much materializing to access offset/partition later down the line?

ksqldb-streams/src/main/java/io/confluent/ksql/execution/streams/SourceBuilder.java
vcrfxia: How about calling `append()` the requisite number times to just append nulls (as a form of initialization), and then calling `set()` in a single for-loop?

Sullivan-Patrick: This would work well. My question is if the readability would be better at this point. At the least, I think it'll be easier to comment, and extend. I'll implement it and we can decide which way is better.

ksqldb-streams/src/main/java/io/confluent/ksql/execution/streams/SourceBuilder.java
```diff
@@ -212,44 +211,31 @@ public static SourceBuilder instance() {
     );
   }
 
-  private static PhysicalSchema getPhysicalSchemaWithKeyAndPseudoCols(
+  private static PhysicalSchema getPhysicalSchemaWithRelevantPseudoAndKeyCols(
```
vcrfxia, line 214: I don't understand these changes. How come the two previous methods `getPhysicalSchemaWithKeyAndPseudoCols()` and `getPhysicalSchemaWithPseudoColumnsToMaterialize()` were collapsed into a single implementation? Does the functionality still work with these changes?

ksqldb-common/src/main/java/io/confluent/ksql/schema/ksql/SystemColumns.java
```diff
@@ -114,6 +114,10 @@ public static boolean isSystemColumn(final ColumnName columnName) {
     return systemColumnNames(CURRENT_PSEUDOCOLUMN_VERSION_NUMBER);
   }
 
+  public static Boolean mustBeMaterializedForTableJoins(ColumnName columnName) {
```
vcrfxia, line 117: Why is the return type `Boolean` rather than `boolean`?

ksqldb-common/src/main/java/io/confluent/ksql/schema/ksql/SystemColumns.java
```diff
@@ -146,6 +150,12 @@ public static boolean isSystemColumn(final ColumnName columnName) {
             1, VERSION_ONE_NAMES
         );
 
+    private static final Set<ColumnName> MUST_BE_MATERIALIZED_FOR_TABLE_JOINS =
```
vcrfxia, line 153: A slightly cleaner way to structure this would be to create an object representing a pseudocolumn. This object could contain the pseudocolumn name and a boolean to specify whether the column should be materialized or not. By tying the two together in this way, it is not possible for someone to come along and add a new pseudocolumn without thinking about whether that new column should be materialized or not.

That said, I think what you've implemented here is fine for now since adding pseudocolumns to the codebase is rare.

ksqldb-streams/src/main/java/io/confluent/ksql/execution/streams/SourceBuilder.java
vcrfxia: It looks like `toPotentiallyMaterialize` is still being set after the table might've already been materialized? I don't understand the change referred to in this comment:

> given how I changed the code to materialize on transform, I left it as toPotentiallyMaterialize

ksqldb-streams/src/main/java/io/confluent/ksql/execution/streams/SourceBuilder.java
vcrfxia: Good point. We can clarify the comment with something like "besides materializing necessary pseudocolumns, we also materialize to prevent ..."

ksqldb-streams/src/main/java/io/confluent/ksql/execution/streams/SourceBuilder.java
```diff
@@ -351,12 +339,16 @@ public GenericRow transform(final K key, final GenericRow row) {
 
           row.ensureAdditionalCapacity(pseudoColumnsToAdd);
```
vcrfxia, line 340: This logic should also be updated to use `mustBeMaterializedForTableJoins()`.

ksqldb-streams/src/main/java/io/confluent/ksql/execution/streams/SourceBuilder.java
Sullivan-Patrick: It's still functional because we always use the new functionality with `TableSource{V2}`s and the old functionality (adding pseudo and key cols) with `WindowedTableSource{V2}`.

ksqldb-common/src/main/java/io/confluent/ksql/schema/ksql/SystemColumns.java
Sullivan-Patrick: This is a really good thing to think about, I haven't really gone "up" a layer in abstraction (ie added my own)- but I've seen it a lot in the codebase, and while it's taken a while to get used to, I can definitely see how it helps maintainable and readability in the long term. I'd be interested in doing this, let me earmark it as something I can do later for time's sake though.

ksqldb-streams/src/main/java/io/confluent/ksql/execution/streams/SourceBuilder.java
Sullivan-Patrick: I was reading it as "we potentially materialize this" when declaring the variable, not when assigning it. Thinking about it, I agree that naming based off of when it's assigned makes more sense.

ksqldb-streams/src/main/java/io/confluent/ksql/execution/streams/SourceBuilder.java
vcrfxia: OK, I probably misread an earlier version of this PR. Can we rename this method to `getPhysicalSchemaWithPseudoColumnsToMaterialize()` in order to clarify the method's behavior? "Relevant pseudo and key columns" isn't very informative as a name.

ksqldb-streams/src/main/java/io/confluent/ksql/execution/streams/SourceBuilder.java
```diff
@@ -200,24 +213,33 @@ public static SourceBuilder instance() {
...
     final KeyFormat keyFormat = windowed
         ? KeyFormat.windowed(
-            formatInfo, serdeFeatures, ((WindowedTableSource) streamSource).getWindowInfo())
+        formatInfo, serdeFeatures, ((WindowedTableSource) streamSource).getWindowInfo())
         : KeyFormat.nonWindowed(formatInfo, serdeFeatures);
 
-    final Formats formats = of(keyFormat, streamSource.getFormats().getValueFormat());
+    final Formats format = of(keyFormat, streamSource.getFormats().getValueFormat());
+
+    //we currently don't materialize partition and offset to get them downstream for windowed
```
vcrfxia, line 231: Am I misremembering or did you say that push queries _do_ allow windowed table joins, in which case we do need to materialize partition and offset for windowed tables, else the new pseudocolumns won't work with windowed joins?

ksqldb-streams/src/main/java/io/confluent/ksql/execution/streams/SourceBuilder.java
```diff
@@ -232,4 +254,129 @@ private static Formats of(final KeyFormat keyFormat, final FormatInfo valueForma
...
+            row.append(null);
+          }
+
+          //calculate number of user columns, pseudo columns, and columns to shift for next steps
+          final Set<ColumnName> columnNames = SystemColumns.pseudoColumnNames(pseudoColumnVersion);
+          final int totalPseudoColumns = columnNames.size();
+          final int pseudoColumnsToShift = totalPseudoColumns - pseudoColumnsToAdd;
+          final int numUserColumns = row.size() - totalPseudoColumns;
+
+          //create a list of pseudocolumns, in the order which they should be added
```
vcrfxia, line 357: nit:
```suggestion
          // collect pseudocolumn values, in the order in which they should be added
```

vcrfxia (approved):
Thanks @Sullivan-Patrick -- LGTM with a question inline. Let's merge once the build is passing 🎉

ksqldb-streams/src/main/java/io/confluent/ksql/execution/streams/SourceBuilder.java
Sullivan-Patrick: push queries do _not_, see https://github.com/confluentinc/ksql/issues/8055

vcrfxia: In that case why do we need to bump the execution step version for WindowedSourceTable at all? Can we just use the original WindowedSourceTableV1 and the original SourceBuilderV1 for the windowed table case to save on new code?

Sullivan-Patrick: We could. At the time I wrote https://github.com/confluentinc/ksql/pull/7990 I didn't know about the constraint on windowed tables. I first thought about exactly this when when I brought it up in our sync and you mentioned how rare windowed table joins are. Since we have no need to change the topology, we can definitely still use the V1 step. The tradeoff is that if we did want to support these joins in the future, someone would need to rewrite the V2 step. This is opposed to if we complete the windowed step as part of KLIP-50, which will give windowed tables a parallel topology change with unwindowed tables and allow for joins, or anything else where we need to access pseudocolumns in a state store.

This was a pretty big oversight by me to not bring it up in our sync, but I'd be interested to hear what you think about the best way to do this.

vcrfxia: I think it's preferable to not bump the execution step version, as doing so adds significant code. If joins between windowed tables haven't worked until now, I think the fact that we haven't heard any reports about it means no one is using them and it doesn't make sense for us to invest in supporting them in the near future. This is not surprising to me since a windowed join as a push query offers limited usefulness given that it cannot be converted into a persistent query at this time.

vcrfxia: As discussed offline, let's remove the unnecessary execution step (WindowedSourceTableV2) in a follow-up PR. Merging this one for now. Thanks @Sullivan-Patrick !

PR #8074, opened 2021-08-30
feat: add shared runtime config to QueryPlan

wcarlson5:
Make it so the config value for shared runtimes are enabled is written into the cmd topic so it is specific to when each query was launched 

### Reviewer checklist

### DIFF
--- a/ksqldb-engine/src/main/java/io/confluent/ksql/engine/QueryPlan.java
+++ b/ksqldb-engine/src/main/java/io/confluent/ksql/engine/QueryPlan.java
@@ -33,14 +33,14 @@ public final class QueryPlan  {
   private final Optional<SourceName> sink;
   private final ExecutionStep<?> physicalPlan;
   private final QueryId queryId;
-  private final Optional<String> usesSharedRuntime;
+  private final Optional<String> runtimeId;
 
   public QueryPlan(
       @JsonProperty(value = "sources", required = true) final Set<SourceName> sources,
       @JsonProperty(value = "sink") final Optional<SourceName> sink,
       @JsonProperty(value = "physicalPlan", required = true) final ExecutionStep<?> physicalPlan,
       @JsonProperty(value = "queryId", required = true) final QueryId queryId,
-      @JsonProperty(value = "sharedRuntimes") final Optional<String> usesSharedRuntime
+      @JsonProperty(value = "runtimeId") final Optional<String> runtimeId
   ) {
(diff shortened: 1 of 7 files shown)


guozhangwang (commented):
Overall LGTM. Could we add a test covering the case e.g. even with the global config set to true, if a specific query has it false, or does not set it (since it's written from an older version) in the query plan, that specific query would not be assigned to the shared runtime still.

wcarlson5:
@guozhangwang good idea. I added a test that should verify that global config is only used when creating the query plan. I also made it so that for any query plans that get serialized with out it set will default to false. Eventually when we remove the FF we will remove it from the query plan as well

guozhangwang (approved):
Thanks for converting the field to a string as well. I do not have further comments. cc @rodesai .

ksqldb-engine/src/main/java/io/confluent/ksql/engine/QueryPlan.java
```diff
@@ -33,12 +33,14 @@
   private final Optional<SourceName> sink;
   private final ExecutionStep<?> physicalPlan;
   private final QueryId queryId;
+  private final Optional<String> usesSharedRuntime;
```
ableegoldman, line 36: super nit: name makes it sound like this is a boolean, can you describe what it's actually supposed to be used for/mean?

ableegoldman (approved):
Makes sense to me, but @rodesai should probably give it a quick final pass since I'm still just a ksql noob 🙂

ksqldb-engine/src/main/java/io/confluent/ksql/engine/QueryPlan.java
```diff
@@ -33,12 +33,14 @@
...
 
   public QueryPlan(
       @JsonProperty(value = "sources", required = true) final Set<SourceName> sources,
       @JsonProperty(value = "sink") final Optional<SourceName> sink,
       @JsonProperty(value = "physicalPlan", required = true) final ExecutionStep<?> physicalPlan,
-      @JsonProperty(value = "queryId", required = true) final QueryId queryId
+      @JsonProperty(value = "queryId", required = true) final QueryId queryId,
+      @JsonProperty(value = "sharedRuntimes") final Optional<String> usesSharedRuntime
```
rodesai, line 43: we should probably name this something different right? is the intent to store the runtime ID in there? maybe call it runtimeId then?

PR #8085, opened 2021-08-31
feat: add feature flag to disable source table materialization

spena:
### Description 
Adds a feature flag disable source table materialization. The config to disable is:
```
ksql.source.table.materialization.enabled=true
```

The  config can only be disabled from the `ksql-server.properties`. This config cannot be set on the CLI or other clients to avoid ksql clients users to enable it without notice of the ksql server admin (like Cloud users).

When the config is disabled, KSQL will not accept new CST commands. However, current CST commands found in the command topic will be executed but with no materialization and pull query support. I left the behavior this way to prevent crashing the KSQL on startup if this config is disabled and there were CST commands already running. CST tables will still be read-only, though.

### Testing done 
I didn't add tests because there's still a PR pending where to add them - https://github.com/confluentinc/ksql/pull/8079
I tested manually.

Note that the below message does not show what property is required to set. I left it this way as the feature flag should be an internal flag to disable the feature and users, especially cloud users, should not know what the property to enable it is.
```
ksql> create source table t1(id int primary key, name string) with (kafka_topic='t1', value_format='json', partitions=1);
Cannot execute command because source table materialization is disabled.

ksql> set 'ksql.source.table.materialization.enabled'='true';
Cannot override property 'ksql.source.table.materialization.enabled'
```

### Reviewer checklist

### DIFF
--- a/ksqldb-common/src/main/java/io/confluent/ksql/util/KsqlConfig.java
+++ b/ksqldb-common/src/main/java/io/confluent/ksql/util/KsqlConfig.java
@@ -411,10 +411,11 @@ public class KsqlConfig extends AbstractConfig {
       "ksql.source.table.materialization.enabled";
   private static final Boolean KSQL_SOURCE_TABLE_MATERIALIZATION_ENABLED_DEFAULT = true;
   private static final String KSQL_SOURCE_TABLE_MATERIALIZATION_ENABLED_DOC =
-      "Feature flag that enables table materialization on source tables. "
-          + "A source table is created with the CREATE SOURCE TABLE syntax. "
-          + "Default is true. If false, CREATE SOURCE TABLE will not be materialized, "
-          + "thus not able to run pull queries on them.";
+      "Feature flag that enables table materialization on source tables. Default is true. "
+          + "If false, CREATE SOURCE [TABLE|STREAM] statements will be rejected. "
+          + "Current CREATE SOURCE TABLE statements found in the KSQL command topic will "
+          + "not be materialized and pull queries won't be allowed on them. However, current "
+          + "CREATE SOURCE [TABLE|STREAM] statements will continue being read-only.";
 
   public static final String KSQL_SHARED_RUNTIME_ENABLED = "ksql.runtime.feature.shared.enabled";
   public static final Boolean KSQL_SHARED_RUNTIME_ENABLED_DEFAULT = false;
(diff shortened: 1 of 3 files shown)


config/ksql-server.properties
```diff
@@ -66,6 +66,7 @@ ksql.logging.processing.stream.auto.create=true
 
 # The set of Kafka brokers to bootstrap Kafka cluster information from:
 bootstrap.servers=localhost:9092
+#ksql.source.table.materialization.enabled=false
```
colinhicks, line 69: I think we should leave it out of the properties file. It could cause confusion, and we haven't been doing this generally for feature flags, right?.

ksqldb-common/src/main/java/io/confluent/ksql/util/KsqlConfig.java
```diff
@@ -407,6 +407,15 @@
       "Feature flag for ROWPARTITION and ROWOFFSET pseudocolumns. If enabled, new queries will be"
           + "built with ROWPARTITION and ROWOFFSET pseudocolumns. If off, they will not be.";
 
+  public static final String KSQL_SOURCE_TABLE_MATERIALIZATION_ENABLED =
+      "ksql.source.table.materialization.enabled";
+  private static final Boolean KSQL_SOURCE_TABLE_MATERIALIZATION_ENABLED_DEFAULT = true;
+  private static final String KSQL_SOURCE_TABLE_MATERIALIZATION_ENABLED_DOC =
+      "Feature flag that enables table materialization on source tables. "
+          + "A source table is created with the CREATE SOURCE TABLE syntax. "
+          + "Default is true. If false, CREATE SOURCE TABLE will not be materialized, "
```
colinhicks, line 416: if the flag is false, it's more accurate to say that CST statements would be rejected, right?

colinhicks:
From the description:
> The config to disable is:

```
ksql.source.table.materialization.enabled=true
```

It should be `ksql.source.table.materialization.enabled=false` right?

config/ksql-server.properties
spena: This shouldn't be here. I was testing and forgot to remove this part from the config. I'm probably going to mention this feature flag in the docs.

ksqldb-common/src/main/java/io/confluent/ksql/util/KsqlConfig.java
spena: Yes, I'll change this. I initially was going to disable materialization only, and forgot to edit this part.

jzaralim (approved):
The code looks good

colinhicks (approved):

PR #8092, opened 2021-09-01
test: add logging to shouldStreamInserts in case it times out again

agavra:
### Description 
This test very rarely flakes out, but when it does it just "times out". I'd like to know which part is timing out.

### Testing done 
N/A logging only

### Reviewer checklist


nateab (approved):
LGTM!

wcarlson5 (approved):

PR #8094, opened 2021-09-02
fix: Enables ALPN for internal requests when http2 and tls are in use

AlanConfluent:
### Description 
Vertx throws an error when http2 and tls are enabled, but the ALPN option isn't set on the options:
https://github.com/eclipse-vertx/vert.x/blob/master/src/main/java/io/vertx/core/http/impl/HttpClientImpl.java#L548

This sets the option when tls is used for http2, internal requests.

### Testing done 
Tested using http2 and tls.

### Reviewer checklist


nateab (approved):
LGTM!

agavra (approved):

PR #8095, opened 2021-09-02
feat: update storage utilization metrics to start when app is initialized

lct45:
### Description 
per https://confluentinc.atlassian.net/browse/KCI-938, make storage utilization per node metrics initialized when the app starts up. this means that even if the app has no queries, we'll still be getting metrics at the node level. previously, we'd only see node level metrics after we created a query

### Testing done 
kept existing unit tests

### Reviewer checklist


rodesai (approved):
LGTM!

guozhangwang (approved):

PR #8096, opened 2021-09-02
chore: do not reset offset on source tables

spena:
### Description 
This PR makes sure that users do not override the `auto.offset.reset` config when creating a new source table. I found this problem while testing. If I user sets `auto.offset.reste=latest` and then create a source table, the internal query will populate the state materialization from the latest changes of the topic. 

This PR always sets `auto.offset.reset` to `earliest`, so the table starts materialization from the beginning of the topic.

### Testing done 
Added unit tests and verified manually

### Reviewer checklist

### DIFF
--- a/ksqldb-engine/src/main/java/io/confluent/ksql/statement/Injectors.java
+++ b/ksqldb-engine/src/main/java/io/confluent/ksql/statement/Injectors.java
@@ -33,7 +33,8 @@ public enum Injectors implements BiFunction<KsqlExecutionContext, ServiceContext
       new DefaultSchemaInjector(
           new SchemaRegistryTopicSchemaSupplier(sc.getSchemaRegistryClient())),
       new TopicCreateInjector(ec, sc),
-      new SchemaRegisterInjector(ec, sc)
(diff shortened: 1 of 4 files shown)


spena:
@colinhicks Just to confirm that we should not allow users to override the `auto.offset.reset` to `latest` when creating new source tables. This PR addresses that.

jzaralim (approved):
The code looks fine to me, but I definitely would like to hear @colinhicks opinion on this.

mjsax:
Not sure if we should apply this restriction. This problem exists today, and might require a more general solution. For example:
```
set "auto.offset.reset"="latest";
CREATE TABLE t1 <schema> WITH (...);
CREATE TABLE t2 AS SELECT * FROM t1;
```
Same problem for the existing code to materialize a table.

I think it might be best to approach this problem end-to-end, and not piggy-back on KLIP-49 (it's also not part of the approved KLIP).

spena:
Closing this ticket without merging. A decision was discussed that allowing users to reset the offset is required for short-term. We are going to fix this offset issue in a long-term solution for other tables and queries.

PR #8097, opened 2021-09-02
chore: make CREATE OR REPLACE use execution step and pseudocolumn versions of original query

Sullivan-Patrick:
### Description 
As part of [KLIP-50](https://github.com/confluentinc/ksql/blob/master/design-proposals/klip-50-partition-and-offset-in-ksqldb.md), When using `CREATE OR REPLACE` to upgrade a query that has an old pseudocolumn version number, we want to use the old one to ensure backwards compatibility.

In the case of `TableSourceV1`s that are being upgraded from, we want to continue to use the V1 step as to not change the KStreams topology. For context on why the V2 source exists and how it changes the topology, see https://github.com/confluentinc/ksql/pull/7990, https://github.com/confluentinc/ksql/pull/8072, https://github.com/confluentinc/ksql/pull/8111.

This PR makes it so replaced queries use old `SourceStep.pseudoColumnVersion` fields, and queries with a `TableSourceV1` source continue to do so.

### Testing done 
Unit tests in `SchemaKSourceFactoryTest`, SQL testing tool tests. It's worth noting that the SQL testing tool isn't fully compatible with `ROWPARTITION` and `ROWOFFSET` currently. These fields are able to be selected as part of a query, but they do not work directly with `ASSERT`, and when checked indirectly (by selecting them in a new query and using `ASSERT` on the new stream's aliased columns), the values are always 0. Regardless, these tests to add some extra coverage, so I think it's good to have them.

### Reviewer checklist


ksqldb-engine/src/main/java/io/confluent/ksql/structured/SchemaKSourceFactory.java
```diff
@@ -174,15 +182,34 @@ private SchemaKSourceFactory() {
...
+          .stream()
+          .anyMatch(executionStep -> executionStep instanceof TableSourceV1);
+    }
+
+    if (useOldExecutionStepVersion && pseudoColumnVersionToUse != 0) {
+      throw new IllegalStateException("TableSourceV2 was released in conjunction with pseudocolumn"
+          + "version 1. Something has gone very wrong");
+    }
+
+    // When feature flag is removed, remove getBoolean() from condition
```
vcrfxia, line 202: nit: I don't think this comment is necessary. If `KsqlConfig.KSQL_ROWPARTITION_ROWOFFSET_ENABLED` is deleted, then this line will have to be updated (else compilation will fail) and it's pretty clear how this should be updated when that happens.

ksqldb-engine/src/main/java/io/confluent/ksql/engine/QueryEngine.java
```diff
@@ -72,7 +73,8 @@ PhysicalPlan buildPhysicalPlan(
       final LogicalPlanNode logicalPlanNode,
       final SessionConfig config,
       final MetaStore metaStore,
-      final QueryId queryId
+      final QueryId queryId,
+      final Optional<PlanInfo> oldPlan
```
vcrfxia, line 77: nit:
```suggestion
      final Optional<PlanInfo> oldPlanInfo
```

ksqldb-engine/src/main/java/io/confluent/ksql/physical/PhysicalPlanBuilder.java
```diff
@@ -54,15 +56,17 @@ public PhysicalPlanBuilder(
 
   public PhysicalPlan buildPhysicalPlan(
       final LogicalPlanNode logicalPlanNode,
-      final QueryId queryId
+      final QueryId queryId,
+      final Optional<PlanInfo> oldPlanToMaybeUse
```
vcrfxia, line 60:
```suggestion
      final Optional<PlanInfo> oldPlanInfo
```

ksqldb-execution/src/main/java/io/confluent/ksql/execution/plan/PlanInfo.java
```diff
@@ -70,6 +71,10 @@ public PlanInfo merge(final PlanInfo other) {
     return new PlanInfo(allSources.merge(other.allSources));
   }
 
+  public Set<ExecutionStep<?>> getSourceSet() {
```
vcrfxia, line 74: nit:
```suggestion
  public Set<ExecutionStep<?>> getSources() {
```

ksqldb-engine/src/test/java/io/confluent/ksql/structured/SchemaKSourceFactoryTest.java
```diff
@@ -239,12 +248,243 @@ public void shouldBuildCorrectFormatsForV2NonWindowedTable() {
...
+        contextStacker
+    );
+
+    // Then:
+    assertThat(((WindowedStreamSource) result.getSourceStep()).getPseudoColumnVersion(), equalTo(NEW_PSEUDOCOLUMN_VERSION));
+    assertValidSchema(result);
+  }
+
+  @Test
+  public void shouldCreateNonWindowedTableSourceV2WithNewPseudoColumnVersionIfNoOldQuery() {
```
vcrfxia, line 326: Looks like there's no test checking that TableSourceV2 with an existing query uses the existing pseudocolumn version? Unless it throws an error to use an old version with TableSourceV2, I think it makes sense to add the test even if it's unrealistic at this time (you can leave a comment to point out the fact that it's unrealistic, if that makes it feel less weird).

ksqldb-engine/src/test/java/io/confluent/ksql/structured/SchemaKSourceFactoryTest.java
```diff
@@ -239,12 +248,243 @@ public void shouldBuildCorrectFormatsForV2NonWindowedTable() {
...
+        contextStacker
+    );
+
+    // Then:
+    assertThat(((TableSourceV1) result.getSourceStep()).getPseudoColumnVersion(), equalTo(OLD_PSEUDOCOLUMN_VERSION));
+    assertValidSchema(result);
+  }
+
+  @Test
+  public void shouldReplaceOldWindowedTableSourceWithMatchingPseudoColumnVersion() {
```
vcrfxia, line 381:
```suggestion
  public void shouldReplaceWindowedTableSourceWithMatchingPseudoColumnVersion() {
```

ksqldb-engine/src/test/java/io/confluent/ksql/structured/SchemaKSourceFactoryTest.java
```diff
@@ -239,12 +248,243 @@ public void shouldBuildCorrectFormatsForV2NonWindowedTable() {
...
+  private void givenExistingTableSourceV1() {
+    givenExistingQueryWithOldPseudoColumnVersion();
+    when(buildContext.getPlanInfo()).thenReturn(Optional.of(planInfo));
+    when(planInfo.getSourceSet()).thenReturn(ImmutableSet.of(tableSourceV1));
+    when(tableSource.getPseudoColumnVersion()).thenReturn(OLD_PSEUDOCOLUMN_VERSION);
+  }
+
+  private void givenExistingQueryWithOldPseudoColumnVersion() {
+    when(buildContext.getPlanInfo()).thenReturn(Optional.of(planInfo));
+    when(planInfo.getSourceSet()).thenReturn(ImmutableSet.of(streamSource));
```
vcrfxia, line 476: Why is this method always using `streamSource` even from the tests that are testing tables, etc? I think we should make this wiring more airtight -- you can pass in the source step to mock as a param if you don't want to make copies of this method.

ksqldb-engine/src/test/java/io/confluent/ksql/structured/SchemaKSourceFactoryTest.java
```diff
@@ -64,6 +69,8 @@
       .build();
 
   private static final String TOPIC_NAME = "fred";
+  private static final int OLD_PSEUDOCOLUMN_VERSION = SystemColumns.ROWTIME_PSEUDOCOLUMN_VERSION;
```
vcrfxia, line 72: Some of the tests added in this file fail unless `OLD_PSEUDOCOLUMN_VERSION` is exactly `SystemColumns.LEGACY_PSEUDOCOLUMN_VERSION_NUMBER`. In other words, it doesn't work for this test file to use a generic "old" version and a generic "new version". As such, I think the tests should just use `LEGACY_PSEUDOCOLUMN_VERSION_NUMBER` directly, instead of defining an alias. Similarly, we can just use `SystemColumns.CURRENT_PSEUDOCOLUMN_VERSION_NUMBER` in place of defining a new variable `NEW_PSEUDOCOLUMN_VERSION` as well.

vcrfxia (commented):
Thanks @Sullivan-Patrick . LGTM overall assuming this has been tested. Did you get a chance to look at KsqlTesterTest? Also, we need to test this with CREATE SOURCE TABLE. I'm not familiar enough with the implementation of CREATE SOURCE TABLE to comment on whether this PR impacts it or not. Let's be sure to follow up on that before merging.

ksqldb-engine/src/test/java/io/confluent/ksql/structured/SchemaKSourceFactoryTest.java
Sullivan-Patrick: Why this but not the others with the same convention (ie `shouldReplaceOldWindowedStreamSourceWithMatchingPseudoColumnVersion`)? I don't feel strongly that the `Old` is necessary in any, though.

ksqldb-execution/src/main/java/io/confluent/ksql/execution/plan/PlanInfo.java
vcrfxia: `Sources` is a private class, so I don't think the name is ambiguous. If you prefer we can do `getSourceSteps()` instead, but I do think `getSourceSet()` is no good since the fact that it's a set is not relevant (and can be seen at a glance from the type itself).

ksqldb-engine/src/test/java/io/confluent/ksql/structured/SchemaKSourceFactoryTest.java
vcrfxia: I'm all for updating the others too! 😁

vcrfxia: I think "old" is unnecessary/confusing because:

* there's no such thing as an "old" windowed table source, i.e., there's only WindowedTableSourceV1 and no WindowedTableSourceV2
* if "old" is meant to mean "existing" rather than "of an old version", then it is redundant because if it's being replaced then it must be old
* table sources actually do have two versions (TableSourceV1 and TableSourceV2) so "old" does make sense in this context (though ideally we'd say TableSourceV1 and TableSourceV2 rather than OldTableSource and NewTableSource, in order to age better). Using "old" in places where it doesn't make sense detracts from places where it does make sense.

ksqldb-engine/src/main/java/io/confluent/ksql/planner/plan/PlanBuildContext.java
```diff
@@ -91,7 +78,8 @@ public PlanBuildContext withKsqlConfig(final KsqlConfig newConfig) {
     return of(
         newConfig,
         serviceContext,
-        functionRegistry
+        functionRegistry,
+        Optional.empty()
```
vcrfxia, line 82:
```suggestion
        planInfo
```

ksqldb-engine/src/test/java/io/confluent/ksql/structured/SchemaKSourceFactoryTest.java
```diff
@@ -252,7 +255,7 @@ public void shouldBuildCorrectFormatsForV2NonWindowedTable() {
   public void shouldReplaceOldNonWindowedStreamSourceWithMatchingPseudoColumnVersion() {
     // Given:
     givenNonWindowedStream();
-    givenExistingQueryWithOldPseudoColumnVersion();
+    givenExistingQueryWithOldPseudoColumnVersion(windowedStreamSource);
```
vcrfxia, line 258: Why are we passing in a windowedStreamSource for a test that tests non-windowed stream sources?

ksqldb-engine/src/test/java/io/confluent/ksql/structured/SchemaKSourceFactoryTest.java
```diff
@@ -263,7 +266,7 @@ public void shouldReplaceOldNonWindowedStreamSourceWithMatchingPseudoColumnVersi
     );
 
     // Then:
-    assertThat(((StreamSource) result.getSourceStep()).getPseudoColumnVersion(), equalTo(OLD_PSEUDOCOLUMN_VERSION));
+    assertThat(((StreamSource) result.getSourceStep()).getPseudoColumnVersion(), equalTo(SystemColumns.LEGACY_PSEUDOCOLUMN_VERSION_NUMBER));
```
vcrfxia, line 269: nit: you can use a static important to shorten appearances of `SystemColumns.LEGACY_PSEUDOCOLUMN_VERSION_NUMBER` to simply `LEGACY_PSEUDOCOLUMN_VERSION_NUMBER`, and similarly for `CURRENT_PSEUDOCOLUMN_VERSION_NUMBER`.

ksqldb-engine/src/test/java/io/confluent/ksql/structured/SchemaKSourceFactoryTest.java
```diff
@@ -336,7 +339,27 @@ public void shouldCreateNonWindowedTableSourceV2WithNewPseudoColumnVersionIfNoOl
...
 
     // Then:
-    assertThat(((TableSource) result.getSourceStep()).getPseudoColumnVersion(), equalTo(NEW_PSEUDOCOLUMN_VERSION));
+    assertThat(((TableSource) result.getSourceStep()).getPseudoColumnVersion(), equalTo(SystemColumns.CURRENT_PSEUDOCOLUMN_VERSION_NUMBER));
+    assertValidSchema(result);
+  }
+
+  //this state shouldn't exist in normal use, testing for completeness
+  @Test
+  public void shouldReplaceOldWTableSourceV2WithMatchingPseudoColumnVersion() {
```
vcrfxia, line 348: What is an "OldWTableSourceV2"? See other discussion about the meaning of "old" in these tests.

ksqldb-engine/src/test/java/io/confluent/ksql/structured/SchemaKSourceFactoryTest.java
```diff
@@ -336,7 +339,27 @@ public void shouldCreateNonWindowedTableSourceV2WithNewPseudoColumnVersionIfNoOl
     );
 
     // Then:
-    assertThat(((TableSource) result.getSourceStep()).getPseudoColumnVersion(), equalTo(NEW_PSEUDOCOLUMN_VERSION));
+    assertThat(((TableSource) result.getSourceStep()).getPseudoColumnVersion(), equalTo(SystemColumns.CURRENT_PSEUDOCOLUMN_VERSION_NUMBER));
+    assertValidSchema(result);
+  }
+
+  //this state shouldn't exist in normal use, testing for completeness
```
vcrfxia, line 346: Clarify what this comment means -- it's only true today because TableSourceV2 was introduced alongside pseudocolumn versioning, and there are only two pseudocolumn versions today. This test may become relevant in the future (else there's no point adding it).

ksqldb-functional-tests/src/test/resources/sql-tests/query-upgrades/rowpartition-rowoffset.sql
```diff
@@ -0,0 +1,59 @@
+----------------------------------------------------------------------------------------------------
+--@test: upgrade query across pseudocolumn versions
```
vcrfxia, line 2:
```suggestion
--@test: upgrade stream across pseudocolumn versions
```

ksqldb-functional-tests/src/test/resources/sql-tests/query-upgrades/rowpartition-rowoffset.sql
```diff
@@ -0,0 +1,59 @@
...
+----------------------------------------------------------------------------------------------------
+SET 'ksql.create.or.replace.enabled' = 'true';
```
vcrfxia, line 4: Why does this need to be set? Isn't this already the default?

ksqldb-functional-tests/src/test/resources/sql-tests/query-upgrades/rowpartition-rowoffset.sql
```diff
@@ -0,0 +1,59 @@
...
+CREATE OR REPLACE STREAM b AS SELECT id, col1, col2 FROM a WHERE col1 > 0 EMIT CHANGES;
+
+INSERT INTO a (id, col1, col2) VALUES (1, 0, 1);
+INSERT INTO a (id, col1, col2) VALUES (1, 1, 1);
+
+ASSERT VALUES b (id, col1, col2) VALUES (1, 1, 1);
+
+--partition and offset don't work with SQL testing tool at the moment,
+--this just exists as a sanity check to ensure we have enabled the feature.
+CREATE STREAM c AS SELECT ID, ROWPARTITION AS rp, ROWOFFSET AS ro FROM b EMIT CHANGES;
```
vcrfxia, line 25: This test doesn't seem to be testing what it claims to test? As written, the test verifies that CSAS is allowed on sources that have been CREATE OR REPLACE'd. It does not verify that CREATE OR REPLACE is allowed/disallowed for ROWPARTITION and ROWOFFSET. I think the more useful tests to add in this file are to check that the user gets a good error message if they try to CREATE OR REPLACE an old query and add ROWPARTITION or ROWOFFSET.

vcrfxia (commented):
Thanks @Sullivan-Patrick -- some comments inline. Thanks for looking into KsqlTesterTest behavior with regards to ROWOFFSET/ROWPARTITION. Can you file a Github issue for the lack of behavior, so we don't lose track? Also, did you get a chance to try CREATE OR REPLACE with CREATE SOURCE TABLE, to check that the changes here haven't broken existing behavior there?

ksqldb-execution/src/main/java/io/confluent/ksql/execution/plan/PlanInfo.java
Sullivan-Patrick: I think the private class justification is a good one, and `getSources()` is concise.

ksqldb-functional-tests/src/test/resources/sql-tests/query-upgrades/rowpartition-rowoffset.sql
Sullivan-Patrick: Good point- this was just  a sanity check that we are able to actually use ROWPARTITION//ROWOFFSET (ie the feature was turned on). That doesn't really seem great, as the test is testing itself a bit (the SET functionality), and nothing else. The test you mentioned is a lot better.

ksqldb-functional-tests/src/test/resources/sql-tests/query-upgrades/rowpartition-rowoffset.sql
```diff
@@ -0,0 +1,130 @@
+----------------------------------------------------------------------------------------------------
+--@test: upgrade stream across pseudocolumn versions
```
vcrfxia, line 2: I thought we discussed removing this test (and the one below) in https://github.com/confluentinc/ksql/pull/8097#discussion_r717135179?

ksqldb-functional-tests/src/test/resources/sql-tests/query-upgrades/rowpartition-rowoffset.sql
```diff
@@ -0,0 +1,130 @@
...
+INSERT INTO a (id, col1, col2) VALUES (1, 0, 1);
+ASSERT VALUES b (id, col1) VALUES (1, 0);
+
+SET 'ksql.rowpartition.rowoffset.enabled' = 'true';
+
+CREATE OR REPLACE TABLE b AS SELECT id, col1, ROWPARTITION AS ro FROM a WHERE col1 > 0 EMIT CHANGES;
+
+
+----------------------------------------------------------------------------------------------------
+--@test: should fail on CREATE OR REPLACE of an old query with ROWOFFSET in SELECT clause
```
vcrfxia, line 79: I don't think we need all four of these tests (each of ROWPARTITION/ROWOFFSET in each of SELECT/WHERE clauses). Let's just keep two of them: 
- rename `should fail on CREATE OR REPLACE of an old query with ROWPARTITION in SELECT clause` to `should fail on CREATE OR REPLACE of a legacy query with ROWPARTITION`
- rename `should fail on CREATE OR REPLACE of a legacy query with ROWOFFSET in WHERE clause` to `should fail on CREATE OR REPLACE of a legacy query with ROWOFFSET`

ksqldb-functional-tests/src/test/resources/sql-tests/query-upgrades/rowpartition-rowoffset.sql
```diff
@@ -0,0 +1,130 @@
...
+CREATE TABLE c AS SELECT ID, ROWPARTITION AS rp, ROWOFFSET AS ro FROM b EMIT CHANGES;
+
+INSERT INTO a (id, col1, col2) VALUES (1, 1, 1);
+
+ASSERT VALUES c (id, rp, ro) VALUES (1, 0, 0);
+
+----------------------------------------------------------------------------------------------------
+--@test: should fail on CREATE OR REPLACE of an old query with ROWPARTITION in SELECT clause
+--@expected.error: io.confluent.ksql.util.KsqlStatementException
+--@expected.message: Invalid Select: Cannot find the select field in the available fields. field: `ROWPARTITION`
```
vcrfxia, line 62: This error message is not great. Ideally we'd print a more targeted error to explain that new pseudocolumns are not supported in CREATE OR REPLACE on existing queries.

We can achieve this by updating `CodeGenRunner#addRequiredColumn()` where this error message is thrown to add a check for whether the column name matches a pseudocolumn name and add to the error message something like "If this is a CREATE OR REPLACE query, pseudocolumns added in newer versions of ksqlDB after the original query was issued are not available for use in CREATE OR REPLACE."

vcrfxia (commented):
Thanks @Sullivan-Patrick -- LGTM besides the comments inline.

ksqldb-functional-tests/src/test/resources/sql-tests/query-upgrades/rowpartition-rowoffset.sql
Sullivan-Patrick: I interpreted this as adding tests, not necessarily removing. I'll go ahead and do that.

ksqldb-execution/src/main/java/io/confluent/ksql/execution/codegen/CodeGenRunner.java
```diff
@@ -319,10 +320,16 @@ public Void visitLambdaExpression(
 
     private void addRequiredColumn(final ColumnName columnName) {
       final Column column = schema.findValueColumn(columnName)
-          .orElseThrow(() -> new KsqlException(
-              "Cannot find the select field in the available fields."
-                  + " field: " + columnName
-                  + ", schema: " + schema.value()));
+          .orElseThrow(() -> SystemColumns.isPseudoColumn(columnName, ksqlConfig)
+              ? new KsqlException(
+              "If this is a CREATE OR REPLACE query, pseudocolumns added in newer versions of"
```
vcrfxia, line 325: Sorry if my previous message wasn't clear -- this additional message should be added in addition to the original message. The original message is still important for users to know, otherwise they don't know what the actual failure was.

vcrfxia (commented):
Thanks @Sullivan-Patrick -- almost there! One comment inline.

ksqldb-functional-tests/src/test/resources/sql-tests/query-upgrades/rowpartition-rowoffset.sql
Sullivan-Patrick: I think you were clear enough, I'm being a bit sloppy right now as I'm scrambling to tie up all the loose ends before end of week.

ksqldb-execution/src/main/java/io/confluent/ksql/execution/codegen/CodeGenRunner.java
Sullivan-Patrick: I think you were clear enough, I'm being a bit sloppy right now as I'm scrambling to tie up all the loose ends before end of week.

ksqldb-execution/src/main/java/io/confluent/ksql/execution/codegen/CodeGenRunner.java
```diff
@@ -318,11 +319,18 @@ public Void visitLambdaExpression(
     }
 
     private void addRequiredColumn(final ColumnName columnName) {
+      final String informationAboutPseudoColumns =
```
vcrfxia, line 322: nit: let's only compute this if we need to the throw the error message. You can define a helper method `fieldNotFoundErrorMessage(columnName)` to return the appropriate error message string. Besides being more efficient, this will also help with readability.

vcrfxia (commented):
Thanks @Sullivan-Patrick -- LGTM besides the minor comment inline, and the failing build. Let's merge afterwards! 🎉

vcrfxia (approved):

PR #8106, opened 2021-09-07
chore: raise the limit on number of builds to keep

vvcephei:
We want to keep builds for three months to support investigations,
but have been hampered by the default value of maxBuildsToKeep (15).

I just raised it arbitrarily to 99. I also made the default retention time
of 90 days explicit in the Jenkinsfile.

### Reviewer checklist

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -11,5 +11,7 @@ dockerfile {
     dockerImageClean = false
     downStreamRepos = ["confluent-security-plugins", "confluent-cloud-plugins"]
     nanoVersion = true
+    maxBuildsToKeep = 999
+    maxDaysToKeep = 90
 }
 
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -11,7 +11,7 @@ dockerfile {
     dockerImageClean = false
     downStreamRepos = ["confluent-security-plugins", "confluent-cloud-plugins"]
     nanoVersion = true
-    maxBuildsToKeep = 999
+    maxBuildsToKeep = 99
     maxDaysToKeep = 90
 }
 


agavra (approved):

wcarlson5 (approved):

alapidas (approved):

Jenkinsfile
```diff
@@ -11,5 +11,7 @@ dockerfile {
     dockerImageClean = false
     downStreamRepos = ["confluent-security-plugins", "confluent-cloud-plugins"]
     nanoVersion = true
+    maxBuildsToKeep = 999
+    maxDaysToKeep = 90
```
xli1996, line 15: We might not need to set `maxDaysToKeep` since it is set default as 90 days. Consider Ksql logs is fairly big, "999" will used around 188GB, maybe "99" is more reasonable until we find a way to archive logs in other storage place.

vvcephei: Thanks, @xli1996 !

I'll change it to 99 and we can re-evaluate to see if that's long enough (without using too much disk).

Regarding the retention time, I wanted to go ahead and make that default explicit so there's no mystery about what our retention time is, and it won't magically change on us if the defaults change.

Jenkinsfile
```diff
@@ -11,5 +11,7 @@ dockerfile {
     dockerImageClean = false
     downStreamRepos = ["confluent-security-plugins", "confluent-cloud-plugins"]
     nanoVersion = true
+    maxBuildsToKeep = 999
```
vvcephei, line 14:
```suggestion
    maxBuildsToKeep = 99
```

xli1996 (approved):

PR #8108, opened 2021-09-07
chore: Move where TableSource's state store format is created

Sullivan-Patrick:
### Description 
As part of [klip-50]( https://github.com/confluentinc/ksql/blob/master/design-proposals/klip-50-partition-and-offset-in-ksqldb.md), it is necessary to materialize partition and offset information in `TableSource` to allow joins (see https://github.com/confluentinc/ksql/pull/7990 and https://github.com/confluentinc/ksql/pull/8072).

Previously, for the intermittent materialization step, we only needed the format information of the source topic, as we were only materializing user columns. Therefore, if `WRAP_SINGLES` was set to false in the value features of a source topic, it guaranteed that there was a single column for the materialization. Since now we are materializing an extra 2 (partition and offset), this would cause the step to fail in these situations. 

In order to merge the previous PRs, we simply copied logic from `InternalFormats.of()` into `SourceBuilder`. This PR cleans up the order in which this happens by creating the `Formats` before the `TableSource` is actually created.

### Testing done 
Changed relevant unit tests to accommodate for new field in `TableSource`, and make sure new logic is verified.

### Reviewer checklist


ksqldb-streams/src/main/java/io/confluent/ksql/execution/streams/SourceBuilder.java
```diff
@@ -200,36 +214,159 @@ public static SourceBuilder instance() {
...
     final SerdeFeatures serdeFeatures = streamSource.getFormats().getKeyFeatures();
+
     final KeyFormat keyFormat = windowed
         ? KeyFormat.windowed(
-            formatInfo, serdeFeatures, ((WindowedTableSource) streamSource).getWindowInfo())
+        formatInfo, serdeFeatures, ((WindowedTableSource) streamSource).getWindowInfo())
         : KeyFormat.nonWindowed(formatInfo, serdeFeatures);
 
-    final Formats formats = of(keyFormat, streamSource.getFormats().getValueFormat());
+    final Formats format = ((TableSource) streamSource).getStateStoreVersion();
```
Sullivan-Patrick, line 230: This makes windowed tests fail, but once we revert `WindowedTableSource` to be the V1 version, this won't be an issue.

ksqldb-execution/src/main/java/io/confluent/ksql/execution/plan/TableSource.java
```diff
@@ -45,16 +48,23 @@ public TableSource(
       @JsonProperty(value = "formats", required = true) final Formats formats,
       @JsonProperty("timestampColumn") final Optional<TimestampColumn> timestampColumn,
       @JsonProperty(value = "sourceSchema", required = true) final LogicalSchema sourceSchema,
-      @JsonProperty("pseudoColumnVersion") final OptionalInt pseudoColumnVersion
+      @JsonProperty("pseudoColumnVersion") final OptionalInt pseudoColumnVersion,
+      @JsonProperty("stateStoreVersion") final Formats stateStoreVersion
```
vcrfxia, line 52: This variable name does not make sense. We're adding formats, not a version?

ksqldb-execution/src/main/java/io/confluent/ksql/execution/plan/TableSource.java
```diff
@@ -31,12 +31,15 @@
...
+  final Formats stateStoreVersion;
+
   private static final ImmutableList<Property> MUST_MATCH = ImmutableList.of(
       new Property("class", Object::getClass),
       new Property("properties", ExecutionStep::getProperties),
       new Property("topicName", s -> ((TableSource) s).topicName),
       new Property("formats", s -> ((TableSource) s).formats),
-      new Property("timestampColumn", s -> ((TableSource) s).timestampColumn)
+      new Property("timestampColumn", s -> ((TableSource) s).timestampColumn),
+      new Property("stateStoreSchema", s -> ((TableSource) s).timestampColumn)
```
vcrfxia, line 42: This looks like a bug.

ksqldb-execution/src/main/java/io/confluent/ksql/execution/plan/TableSource.java
```diff
@@ -45,16 +48,23 @@ public TableSource(
...
   ) {
     super(
         props,
         topicName,
         formats,
         timestampColumn,
         sourceSchema,
-        pseudoColumnVersion.orElse(SystemColumns.LEGACY_PSEUDOCOLUMN_VERSION_NUMBER)
+        pseudoColumnVersion.orElseThrow(() -> new IllegalStateException(
+            "V2 TableSource should always have a pseudocolumn version serialized"))
```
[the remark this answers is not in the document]
vcrfxia: If it's required, then the type should not be `OptionalInt`. Let's change it to just be `int` instead. (This was an oversight from a previous PR -- nice catch.)

ksqldb-streams/src/main/java/io/confluent/ksql/execution/streams/ExecutionStepFactory.java
```diff
@@ -137,7 +137,8 @@ public static TableSource tableSource(
       final LogicalSchema sourceSchema,
       final String topicName,
       final Formats formats,
-      final Optional<TimestampColumn> timestampColumn
+      final Optional<TimestampColumn> timestampColumn,
+      final Formats stateStoreFormat
```
vcrfxia, line 141: nit: 
```suggestion
      final Formats stateStoreFormats
```

ksqldb-streams/src/test/java/io/confluent/ksql/execution/streams/SourceBuilderTest.java
```diff
@@ -645,26 +645,31 @@ private void givenWindowedSourceTable() {
   private void givenUnwindowedSourceTable() {
     when(buildContext.buildKeySerde(any(), any(), any())).thenReturn(keySerde);
     givenConsumed(consumed, keySerde);
+    Formats genericFormat = Formats.of(keyFormatInfo, valueFormatInfo, KEY_FEATURES, VALUE_FEATURES);
```
vcrfxia, line 648: How about just `formats` instead of `genericFormat`? (Unless the "generic" bit adds anything to the name, shorter variable names are preferred.)

Since this is always created in the same way, we can create it in the `setUp()` method instead of in multiple places throughout the tests. (See how `planBuilder` is being created as an example.)

ksqldb-engine/src/test/java/io/confluent/ksql/structured/SchemaKSourceFactoryTest.java
```diff
@@ -180,6 +181,7 @@ public void shouldBuildNonWindowedTable() {
     when(dataSource.getDataSourceType()).thenReturn(DataSourceType.KTABLE);
     when(keyFormat.isWindowed()).thenReturn(false);
     when(keyFormat.getWindowInfo()).thenReturn(Optional.empty());
+    when(keyFormat.getFormatInfo().getFormat()).thenReturn("asdf");
```
vcrfxia, line 184: nit: make this example more realistic, e.g.,
```suggestion
    when(keyFormat.getFormatInfo().getFormat()).thenReturn("JSON");
```

Also, it'd be nice to add unit test coverage for the code changes in this PR, i.e., to ensure that `InternalFormats.of()` is properly called by the relevant method. Because of the feature flag, the relevant code path isn't actually being exercised by these unit tests at the moment. You can add a new test that sets the feature flag on the `CONFIG` used in these tests, and then validate that the state store formats are as expected.

Also also, I didn't know that Mockito let you mock chained methods in this way! The internet seems split on whether this is a bad practice or not. This unit test is simple enough that I'm not too worried but could be good to keep in mind.

vcrfxia (commented):
Thanks @Sullivan-Patrick -- LGTM overall besides the comments inline.

Can you also clarify what "we need information about the internal state store format" in the PR description means? It's not clear from that line what's different about the pre-KLIP-50 and the post-KLIP-50 worlds such that we need to pass an additional field (representing state store formats) into the source builder in the new world but not the old world. Thanks!

ksqldb-engine/src/test/java/io/confluent/ksql/structured/SchemaKSourceFactoryTest.java
Sullivan-Patrick: From what I'm reading, since `InternalFormats.of()` is called in a static method (`SchemaKSourceFactory.buildTable()`), I can't use `verify()` to check that specifically without bringing in another framework like [PowerMock](https://www.javadoc.io/doc/org.powermock/powermock-api-mockito/1.6.2/org/powermock/api/mockito/PowerMockito.html#verifyStatic--).

As far as testing the formats, that shouldn't be a problem.

vcrfxia: I'm not suggesting verifying a mock to check that the method is called. I meant that you can check the value of `result.getSourceStep().getStateStoreFormats()` (or whatever you've renamed the last method to) directly, and compare it to what's expected.

ksqldb-engine/src/test/java/io/confluent/ksql/structured/SchemaKSourceFactoryTest.java
```diff
@@ -64,7 +64,6 @@
       .build();
 
   private static final String TOPIC_NAME = "fred";
-  private static final KsqlConfig CONFIG = new KsqlConfig(ImmutableMap.of());
```
Sullivan-Patrick, line 66: I'm not sure if there's a compelling reason to keep this now that I'm mocking some config behavior?

ksqldb-engine/src/test/java/io/confluent/ksql/structured/SchemaKSourceFactoryTest.java
```diff
@@ -198,10 +197,59 @@ public void shouldBuildNonWindowedTable() {
...
+    // Then:
+    assertThat(((TableSource) result.getSourceStep()).getStateStoreFormats(), is(
+        InternalFormats.of(
+            keyFormat,
+            valueFormatInfo
+        )
+    ));
+  }
+
+  private void setUpForNonWindowedTable() {
```
Sullivan-Patrick, line 243: It feels like there should be a naming convention for what I'm trying to do here but I didn't find it with a cursory Google search.

ksqldb-engine/src/test/java/io/confluent/ksql/structured/SchemaKSourceFactoryTest.java
```diff
@@ -176,12 +177,10 @@ public void shouldBuildWindowedTable() {
...
   @Test
-  public void shouldBuildNonWindowedTable() {
+  public void shouldBuildV1NonWindowedTable() {
     // Given:
-    when(dataSource.getDataSourceType()).thenReturn(DataSourceType.KTABLE);
-    when(keyFormat.isWindowed()).thenReturn(false);
-    when(keyFormat.getWindowInfo()).thenReturn(Optional.empty());
-    when(keyFormat.getFormatInfo().getFormat()).thenReturn("asdf");
+    setUpForNonWindowedTable();
+    when(ksqlConfig.getBoolean(KsqlConfig.KSQL_ROWPARTITION_ROWOFFSET_ENABLED)).thenReturn(false);
```
vcrfxia, line 183: This mock is not necessary. The default mockito return value for boolean methods is false anyway. We can make the tests more readable by only mocking the necessary methods (i.e., the ones where this config returns true).

ksqldb-engine/src/test/java/io/confluent/ksql/structured/SchemaKSourceFactoryTest.java
vcrfxia: The rest of the codebase uses `givenNonWindowedTable()`.

ksqldb-execution/src/main/java/io/confluent/ksql/execution/plan/TableSource.java
```diff
@@ -48,23 +46,22 @@ public TableSource(
       @JsonProperty(value = "formats", required = true) final Formats formats,
       @JsonProperty("timestampColumn") final Optional<TimestampColumn> timestampColumn,
       @JsonProperty(value = "sourceSchema", required = true) final LogicalSchema sourceSchema,
-      @JsonProperty("pseudoColumnVersion") final OptionalInt pseudoColumnVersion,
-      @JsonProperty("stateStoreVersion") final Formats stateStoreVersion
+      @JsonProperty("pseudoColumnVersion") final int pseudoColumnVersion,
+      @JsonProperty("stateStoreFormats") final Formats stateStoreFormats
```
vcrfxia, line 50: The changes in this file should be accompanied by changes in schema.json (the one associated with KsqlPlanSchemaTest.json), no?

ksqldb-execution/src/main/java/io/confluent/ksql/execution/plan/TableSource.java
```diff
@@ -48,23 +46,22 @@ public TableSource(
...
         topicName,
         formats,
         timestampColumn,
         sourceSchema,
-        pseudoColumnVersion.orElseThrow(() -> new IllegalStateException(
-            "V2 TableSource should always have a pseudocolumn version serialized"))
+        pseudoColumnVersion
     );
-    this.stateStoreVersion = stateStoreVersion;
+    this.stateStoreFormats = stateStoreFormats;
```
vcrfxia, line 60: Add a null check.

ksqldb-engine/src/test/java/io/confluent/ksql/structured/SchemaKSourceFactoryTest.java
vcrfxia: There is not. (Good call on removing this.)

ksqldb-engine/src/test/java/io/confluent/ksql/structured/SchemaKSourceFactoryTest.java
Sullivan-Patrick: I considered this, but the`given` methods I saw before assign a class-level variable with the table. Now looking at something like `PushQueryValidatorTest`, I see that's not always the case. I could do something like this? 

```
// Given/When:
givenNonWindowedTable()

// Then:
{assertions}
```

Or if there doesn't really need to be a strict parallel, I could just leave the semantics as is and change the variable name. WDYT?

vcrfxia: Sorry, I don't understand the distinction you're trying to point out. Why would it be inconsistent to rename this method `givenNonWindowedTable()`? The way you'd use it is:
```
// Given:
givenNonWindowedTable();

// When:
{statement}

// Then:
{assertions}
```

ksqldb-engine/src/test/java/io/confluent/ksql/structured/SchemaKSourceFactoryTest.java
```diff
@@ -136,7 +136,6 @@ public void shouldBuildWindowedStream() {
   public void shouldBuildNonWindowedStream() {
     // Given:
     when(dataSource.getDataSourceType()).thenReturn(DataSourceType.KSTREAM);
-    when(keyFormat.isWindowed()).thenReturn(false);
```
vcrfxia, line 137: It seems I wasn't clear with the scope of my previous comment: I think it's fine to leave this mock in here even though mockito would default to false anyway, but we should remove the feature flag default since feature flag defaults are assumed to be false unless otherwise specified, while there's no default assumption for windowed vs unwindowed for sources.

ksqldb-execution/src/main/java/io/confluent/ksql/execution/plan/TableSource.java
```diff
@@ -57,7 +57,7 @@ public TableSource(
         sourceSchema,
         pseudoColumnVersion
     );
-    this.stateStoreFormats = stateStoreFormats;
+    this.stateStoreFormats = Objects.requireNonNull(stateStoreFormats);
```
vcrfxia, line 60:
```suggestion
    this.stateStoreFormats = Objects.requireNonNull(stateStoreFormats, "stateStoreFormats");
```

ksqldb-execution/src/main/java/io/confluent/ksql/execution/plan/TableSource.java
```diff
@@ -21,22 +21,23 @@
...
 import io.confluent.ksql.util.KsqlException;
 import java.util.Objects;
 import java.util.Optional;
-import java.util.OptionalInt;
 import javax.annotation.Nonnull;
 
 @Immutable
 public final class TableSource extends SourceStep<KTableHolder<GenericKey>> {
 
+  final Formats stateStoreFormats;
```
vcrfxia, line 32: How come is this not private?

vcrfxia (commented):
Thanks @Sullivan-Patrick ! LGTM besides the comment inline.

ksqldb-execution/src/main/java/io/confluent/ksql/execution/plan/TableSource.java
Sullivan-Patrick: No good reason.

vcrfxia (approved):

PR #8110, opened 2021-09-07
Supplies proposal for KLIP-55, tablespaces

MichaelDrogalis:
Initial proposal sketching the grammar.

### DIFF
--- a/design-proposals/klip-55-tablespaces.md
+++ b/design-proposals/klip-55-tablespaces.md
@@ -1,3 +1,210 @@
 # KLIP 55 - Tablespaces
 
-Just a placeholder. :-)
+**Author**: @MichaelDrogalis | 
+**Release Target**: TBD | 
+**Status**: In Discussion | 
+**Discussion**: _to be added_
+
+Tablespaces help you use knowledge of the usage pattern of a set of ksqlDB objects to optimize performance.
+
+## Motivation and background
+
+Real-world SQL programs make use of many kinds of objects. Sometimes,
+a subset of those objects have a specific usage pattern. If you could tune for
+that usage pattern on a granular level, ksqlDB could treat those objects more efficiently.
+
+Tablespaces are a construct for defining groups of configuration that can be applied
+to one or more database objects. They exist in
+[Postgres](https://www.postgresql.org/docs/10/manage-ag-tablespaces.html),
+[MySQL](https://dev.mysql.com/doc/refman/8.0/en/general-tablespaces.html),
+[Oracle](https://docs.oracle.com/cd/A57673_01/DOC/server/doc/SCN73/ch4.htm), and so on.
+In the context on ksqlDB, a tablespace would allow you to override default server
+configuration for one or more objects.
+
+This is particularly important given the impact on how ksqlDB's runtime is being rearchitected.
+When multiple persistent queries begin to share a runtime, it will become more challenging to
+vary configuration parameters across them. Tablespaces give you a logical boundary to hint to
+ksqlDB where it must respect varying parameters.
+
+Tablespaces don't enforce namespacing or scoping: two objects in different tablespaces
+can be used with one another.
+
+## What is in scope
+
+- The grammar to support tablespace commands
+- The behavior of those commands
+- Backward compatibility
+
+## What is not in scope
+
+- How it's implemented
+
+## Value/Return
+
+- Improved program performance
+- Isolation of functional parameters to a subset of objects
+
+## Public APIs
+
+**Using the default tablespace**
+
+Creating an object works like normal—it's placed into a default tablespace so
+that you never need to use this feature unless you want to.
+
+```sql
+CREATE stream object WITH (
+    ...
+);
+```
+
+The default tablespace is `ksql_default`.
+
+**Creating a tablespace**
+
+You can supply one or more key/value parameters. These parameters are server-level
+configuration parameters that you want to group together.
+
+```sql
+CREATE TABLESPACE tablespace_name WITH (
+    property_name = expression [,
+    ...
+    ]
+);
+```
+
+By default, no objects are placed within a new tablespace.
+
+- TODO: what happens if `tablespace_name` already exists?
+
+**Adding a newly created object to a tablespace**
+
+Explicitly define what tablespace a stream or table should reside in.
+
+```sql
+CREATE [STREAM|TABLE] obj WITH (
+    'tablespace' = 'tablespace_name',
+    ...
+);
+```
+
+If `tablespace_name` doesn't exist, `obj`'s creation is rejected with a clear error message why.
+
+**Seeing what tablespace an object is in**
+
+```sql
+DESCRIBE obj EXTENDED;
+```
+
+Output fragment:
+
+```
+Name                 : KSQL_PROCESSING_LOG
+Type                 : STREAM
+Timestamp field      : Not set - using <ROWTIME>
+Tablespace           : tablespace_name
+```
+
+**Setting a default tablespace**
+
+You can override the default tablespace so you don't
+need to supply it to every proceeding object that gets created.
+
+```sql
+SET 'ksql.default.tablespace' = 'tablespace_name';
+```
+
+If `tablespace_name` doesn't exist, `obj`'s creation is rejected with a clear error message why.
+
+**Seeing the current default tablespace**
+
+```sql
+SHOW PROPERTIES;
+```
+
+TODO: can we get more granular than that?
+
+**Changing an existing object's tablespace**
+
+```sql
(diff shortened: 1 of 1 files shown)


design-proposals/klip-55-tablespaces.md
```diff
@@ -1,3 +1,210 @@
 # KLIP 55 - Tablespaces
```
blueedgenick, line 1: disappointed about the name :)   All joking aside, I think it's a good opportunity to emphasize the "same but different" nature of ksqlDB by making reference to a familiar construct while also suggesting it may be a little different than what you're used to

design-proposals/klip-55-tablespaces.md
```diff
@@ -1,3 +1,210 @@
...
 
-Just a placeholder. :-)
+**Author**: @MichaelDrogalis | 
+**Release Target**: TBD | 
+**Status**: In Discussion | 
+**Discussion**: _to be added_
+
+Tablespaces help you use knowledge of the usage pattern of a set of ksqlDB objects to optimize performance.
```
blueedgenick, line 8: they also provide a useful hook from which to hang various other things in future e.g. aggregated stats, security permissions

design-proposals/klip-55-tablespaces.md
```diff
@@ -1,3 +1,210 @@
...
+
+Real-world SQL programs make use of many kinds of objects. Sometimes,
+a subset of those objects have a specific usage pattern. If you could tune for
+that usage pattern on a granular level, ksqlDB could treat those objects more efficiently.
+
+Tablespaces are a construct for defining groups of configuration that can be applied
+to one or more database objects. They exist in
+[Postgres](https://www.postgresql.org/docs/10/manage-ag-tablespaces.html),
+[MySQL](https://dev.mysql.com/doc/refman/8.0/en/general-tablespaces.html),
+[Oracle](https://docs.oracle.com/cd/A57673_01/DOC/server/doc/SCN73/ch4.htm), and so on.
```
blueedgenick, line 20: i believe the closest thing in MSSQL is the (confusingly named) 'database'

design-proposals/klip-55-tablespaces.md
```diff
@@ -1,3 +1,210 @@
...
+In the context on ksqlDB, a tablespace would allow you to override default server
+configuration for one or more objects.
+
+This is particularly important given the impact on how ksqlDB's runtime is being rearchitected.
+When multiple persistent queries begin to share a runtime, it will become more challenging to
+vary configuration parameters across them. Tablespaces give you a logical boundary to hint to
+ksqlDB where it must respect varying parameters.
+
+Tablespaces don't enforce namespacing or scoping: two objects in different tablespaces
+can be used with one another.
```
blueedgenick, line 30: "and/or reference each other"

design-proposals/klip-55-tablespaces.md
```diff
@@ -1,3 +1,210 @@
...
+Creating an object works like normal—it's placed into a default tablespace so
+that you never need to use this feature unless you want to.
+
+```sql
+CREATE stream object WITH (
+    ...
+);
+```
+
+The default tablespace is `ksql_default`.
```
blueedgenick, line 60: not sure the additional 'ksql_' prefix adds much value here?

design-proposals/klip-55-tablespaces.md
```diff
@@ -1,3 +1,210 @@
...
+Tablespace           : tablespace_name
+```
+
+**Setting a default tablespace**
+
+You can override the default tablespace so you don't
+need to supply it to every proceeding object that gets created.
+
+```sql
+SET 'ksql.default.tablespace' = 'tablespace_name';
```
blueedgenick, line 113: i think a better analogy here than setting the value of a "session variable" like that is to take inspiration from the rdbms world: `USE space-name;`

design-proposals/klip-55-tablespaces.md
```diff
@@ -1,3 +1,210 @@
...
+```
+
+If `tablespace_name` doesn't exist, `obj`'s creation is rejected with a clear error message why.
+
+**Seeing the current default tablespace**
+
+```sql
+SHOW PROPERTIES;
```
blueedgenick, line 121: i find this is a bit confusing. I'd expect the prompt to change if using the CLI (`space-name:>` or similar) and something visual in the web UI, again for familiarity with the parallel concept in the rdbms world

design-proposals/klip-55-tablespaces.md
```diff
@@ -1,3 +1,210 @@
...
+```sql
+ALTER TABLESPACE tablespace_name SET 'key' = 'value';
+```
+
+TBD: should this support bulk updating parameters?
+
+**Viewing a tablespace's configuration**
+
+```sql
+DESCRIBE tablespace_name [EXTENDED];
```
blueedgenick, line 146: what would we consider to be in the `extended` description here?

design-proposals/klip-55-tablespaces.md
```diff
@@ -1,3 +1,210 @@
...
+`DROP` only succeeds if it has no objects in it. If it has objects, it lists
+out which ones need to be dropped first.
+
+```sql
+DROP TABLESPACE [IF EXISTS] tablespace_name;
+```
+
+**Seeing all objects in a tablespace**
+
+TBD.
```
blueedgenick, line 172: i would expect the extant `show streams|tables` command to show me only objects from the current space

design-proposals/klip-55-tablespaces.md
```diff
@@ -1,3 +1,210 @@
...
+## Compatibility Implications
+
+All existing objects get placed into the default tablespace.
+
+## Security Implications
+
+TBD.
+
+## Open questions
+
```
blueedgenick, line 208: i didn't see a discussion of what happens with non-persistent queries ? do they also execute in the context of the current space ?

vcrfxia:
Seems similar to https://github.com/confluentinc/ksql/pull/7485. Is this proposal meant to replace the earlier one?

design-proposals/klip-55-tablespaces.md
MichaelDrogalis: Yeah I feel the same. An ever-so-slight riff on the norm.

design-proposals/klip-55-tablespaces.md
MichaelDrogalis: Open to anything here, just copied what Postgres does. `default` seems just as good.

design-proposals/klip-55-tablespaces.md
MichaelDrogalis: Do you mean that `SHOW <tablespace>` name should do what the header indicates, or something else? +1 on the former, that's a great idea.

MichaelDrogalis:
@vcrfxia Different idea - schemas give you isolation of names (namespaces). Tablespaces give you isolation of configuration. They can be used together.

design-proposals/klip-55-tablespaces.md
blueedgenick: i don't think `SHOW space-name` is required for anything ? It doesn't fit the pattern of what a `SHOW` or `LIST` generally does, which is to enumerate all objects of a type, rather than describe a specific object or its contents. We don't, for example, intuitively expect `SHOW my-specific-stream-name` to be a thing we can ever do, so why would it work for streamspaces ?

blueedgenick: i would totally support `SHOW STREAMSPACES` to get a list of all of them though!   (you might have that covered already someplace else in the doc)

design-proposals/klip-55-tablespaces.md
blueedgenick: `STREAMSPACE` ftw

design-proposals/klip-55-tablespaces.md
blueedgenick: actually, thinking some more, we should perhaps think twice about my original comment here (that `show streams|tables` should only enumerate objects from the current space). While I personally like the idea, it does edge towards making the space be more of a 'container' than perhaps you were intending to advocate for in this particular KLIP ? We perhaps need a position on what the eventual containment and namespacing model of the system wants to be prior to 'accidentally' starting to back into it. Not sure it does much harm either, just floating that thought.  Different databases approach this somewhat differently so there isn't necessarily one prevalent model to follow. 
(I just happen to personally like the MSSQL flavor, which kind of smushes together the "physical storage properties" and "namespacing boundary" behaviors into a single construct. IMHO this is more intuitive but YMMV. Don't want this bias to leak over into your more surgical proposal here)

design-proposals/klip-55-tablespaces.md
spena: I've seen other Dbs use `default` too.

vvcephei:

PR #8113, opened 2021-09-08
chore: Inlines inner class PseudoColumns since the static order was incorrect

AlanConfluent:
### Description 
This PR inlines the inner class `PseudoColumns` into `SystemColumns`.  Since `SystemColumns` relied on static blocks defined below in `PseudoColumns`, they weren't being initialized in time and resulted in NPEs in tests (and maybe in non test code?).

The main solution was to either inline, or pull the (private) logic out to another class in the package, so I did this.  Also circular dependent classes  are also tricky, so avoided that: https://stackoverflow.com/questions/6416408/static-circular-dependency-in-java

I'm not sure how this works on CI except maybe that it uses a different version of java, but I still don't see how it could pass the same tests.

### Testing done 
Ran package tests

### Reviewer checklist

### DIFF
--- a/ksqldb-common/src/main/java/io/confluent/ksql/schema/ksql/SystemColumns.java
+++ b/ksqldb-common/src/main/java/io/confluent/ksql/schema/ksql/SystemColumns.java
@@ -56,11 +56,35 @@ public final class SystemColumns {
       WINDOWEND_NAME
   );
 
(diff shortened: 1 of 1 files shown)


nateab (approved):
LGTM! Thanks for the quick fix

vcrfxia (approved):
LGTM, though I updated the PR title so this doesn't show up in the changelog (not relevant for users).

Issue #7948, opened 2021-08-04
Inserting values into tables is allowed by the CLI, but not the Java API


colinhicks:
**Describe the bug**
Unlike the CLI, all `INSERT VALUES` statements are rejected by the Java API when the target is a table.

**To Reproduce**
Since the migrations tool uses the Java API, you can use it to repro.

Create a migration:
```sql
CREATE OR REPLACE TABLE `facility` (
  CODE STRING PRIMARY KEY,
  DESCRIPTION VARCHAR
) WITH (
  KAFKA_TOPIC='facility',
  VALUE_FORMAT='AVRO',
  PARTITIONS=1
);

INSERT INTO `facility` (CODE, DESCRIPTION) VALUES ('0','kernel messages');
```
Then apply:
`ksql-migrations apply ...`
=>
```
Received 400 response from server: Cannot insert into a table. Error code: 40001
```

**Expected behavior**
The behavior should be consistent between the CLI and Java API.

**Additional context**
It seems that the Java API adopted this constraint with the thinking discussed in https://github.com/confluentinc/ksql/issues/3954 in mind. There has been no corresponding check added to the CLI.

Also, the constraint on the Java API does not distinguish between inserts into simple tables, including those created by other SQL statements (like the example above) and tables that act as materialized views, which should be read-only according to the thinking described in the referenced issue and other planned language fundamentals work.

One option is to remove the current check in the Java API so that the behavior is consistent until more sophisticated checks can be added to all APIs/clients.

PR #8114, opened 2021-09-08
fix: allow insert into table using HTTP/2 endpoint and Java API

lihaosky:
## Description

Inserting into tables using CLI is permitted but using Java API is not. Allow it in Java API to make the behavior consistent.

Fixes https://github.com/confluentinc/ksql/issues/7948

## Changes
* Remove checks for table insertion in `ksqldb-rest-app/src/main/java/io/confluent/ksql/api/impl/InsertsStreamEndpoint.java`.
* Remove two unit tests in `ksqldb-api-client/src/test/java/io/confluent/ksql/api/client/ClientTest.java` as they use `testEndpoint` and doesn't actually test anything in production endpoint.
* Add 4 test in integration test. All other changes in Integration test comes from Intellij auto formatting
  * Insert and Pull from table -> Exception
  * Insert and Push query from table -> Success
  * Insert and Pull from CTAS table -> Nothing returns
  * Insert to topic and Pull from CTAS table -> Success

### DIFF
--- a/ksqldb-api-client/src/test/java/io/confluent/ksql/api/client/ClientTest.java
+++ b/ksqldb-api-client/src/test/java/io/confluent/ksql/api/client/ClientTest.java
@@ -18,6 +18,7 @@ package io.confluent.ksql.api.client;
 import static io.confluent.ksql.api.client.util.ClientTestUtil.awaitLatch;
 import static io.confluent.ksql.api.client.util.ClientTestUtil.subscribeAndWait;
 import static io.confluent.ksql.rest.Errors.ERROR_CODE_BAD_REQUEST;
+import static io.confluent.ksql.rest.Errors.ERROR_CODE_BAD_STATEMENT;
 import static io.confluent.ksql.test.util.AssertEventually.assertThatEventually;
 import static org.hamcrest.MatcherAssert.assertThat;
 import static org.hamcrest.Matchers.contains;
@@ -595,7 +596,7 @@ public class ClientTest extends BaseApiTest {
   @Test
   public void shouldHandleErrorResponseFromInsertInto() {
     // Given
-    KsqlApiException exception = new KsqlApiException("Cannot insert into a table", ERROR_CODE_BAD_REQUEST);
+    KsqlApiException exception = new KsqlApiException("Invalid target name", ERROR_CODE_BAD_STATEMENT);
     testEndpoints.setCreateInsertsSubscriberException(exception);
 
     // When
@@ -607,7 +608,7 @@ public class ClientTest extends BaseApiTest {
     // Then
     assertThat(e.getCause(), instanceOf(KsqlClientException.class));
     assertThat(e.getCause().getMessage(), containsString("Received 400 response from server"));
-    assertThat(e.getCause().getMessage(), containsString("Cannot insert into a table"));
+    assertThat(e.getCause().getMessage(), containsString("Invalid target name"));
   }
 
   @Test
@@ -671,7 +672,7 @@ public class ClientTest extends BaseApiTest {
   @Test
   public void shouldHandleErrorResponseFromStreamInserts() {
     // Given
-    KsqlApiException exception = new KsqlApiException("Cannot insert into a table", ERROR_CODE_BAD_REQUEST);
+    KsqlApiException exception = new KsqlApiException("Invalid target name", ERROR_CODE_BAD_STATEMENT);
     testEndpoints.setCreateInsertsSubscriberException(exception);
 
     // When
@@ -683,7 +684,7 @@ public class ClientTest extends BaseApiTest {
     // Then
     assertThat(e.getCause(), instanceOf(KsqlClientException.class));
     assertThat(e.getCause().getMessage(), containsString("Received 400 response from server"));
-    assertThat(e.getCause().getMessage(), containsString("Cannot insert into a table"));
+    assertThat(e.getCause().getMessage(), containsString("Invalid target name"));
   }
 
   @Test
--- a/ksqldb-engine/src/test/java/io/confluent/ksql/integration/IntegrationTestHarness.java
+++ b/ksqldb-engine/src/test/java/io/confluent/ksql/integration/IntegrationTestHarness.java
@@ -133,7 +133,7 @@ public final class IntegrationTestHarness extends ExternalResource {
   /**
    * Ensure topics with the given {@code topicNames} exist.
    *
-   * <p>Topics will be creates, if they do not already exist, with a single partition and replica.
+   * <p>Topics will be created, if they do not already exist, with a single partition and replica.
    *
    * @param topicNames the names of the topics to create.
    */
@@ -144,7 +144,7 @@ public final class IntegrationTestHarness extends ExternalResource {
   /**
    * Ensure topics with the given {@code topicNames} exist.
    *
-   * <p>Topics will be creates, if they do not already exist, with the specified
+   * <p>Topics will be created, if they do not already exist, with the specified
    * {@code partitionCount}.
    *
    * @param topicNames the names of the topics to create.
--- a/ksqldb-rest-app/src/main/java/io/confluent/ksql/api/impl/InsertsStreamEndpoint.java
+++ b/ksqldb-rest-app/src/main/java/io/confluent/ksql/api/impl/InsertsStreamEndpoint.java
@@ -25,7 +25,6 @@ import io.confluent.ksql.api.server.KsqlApiException;
 import io.confluent.ksql.engine.KsqlEngine;
 import io.confluent.ksql.metastore.MetaStore;
 import io.confluent.ksql.metastore.model.DataSource;
-import io.confluent.ksql.metastore.model.DataSource.DataSourceType;
 import io.confluent.ksql.name.SourceName;
 import io.confluent.ksql.services.ServiceContext;
 import io.confluent.ksql.util.Identifiers;
@@ -73,11 +72,9 @@ public class InsertsStreamEndpoint {
           "Invalid target name: " + e.getMessage(), ERROR_CODE_BAD_STATEMENT);
     }
 
+
     final DataSource dataSource = getDataSource(ksqlEngine.getMetaStore(),
         SourceName.of(target));
-    if (dataSource.getDataSourceType() == DataSourceType.KTABLE) {
-      throw new KsqlApiException("Cannot insert into a table", ERROR_CODE_BAD_STATEMENT);
-    }
     return InsertsSubscriber.createInsertsSubscriber(serviceContext, properties, dataSource,
         ksqlConfig, context, acksSubscriber, workerExecutor);
   }
@@ -89,12 +86,13 @@ public class InsertsStreamEndpoint {
     final DataSource dataSource = metaStore.getSource(sourceName);
     if (dataSource == null) {
       throw new KsqlApiException(
-          "Cannot insert values into an unknown stream: " + sourceName, ERROR_CODE_BAD_STATEMENT);
+          "Cannot insert values into an unknown stream/table: " + sourceName,
+            ERROR_CODE_BAD_STATEMENT);
     }
 
     if (dataSource.getKsqlTopic().getKeyFormat().isWindowed()) {
       throw new KsqlApiException(
-          "Cannot insert values into windowed stream", ERROR_CODE_BAD_STATEMENT);
+          "Cannot insert values into windowed stream/table", ERROR_CODE_BAD_STATEMENT);
     }
 
     if (reservedInternalTopics.isReadOnly(dataSource.getKafkaTopicName())) {
--- a/ksqldb-api-client/src/test/java/io/confluent/ksql/api/client/integration/ClientMutationIntegrationTest.java
+++ b/ksqldb-api-client/src/test/java/io/confluent/ksql/api/client/integration/ClientMutationIntegrationTest.java
@@ -121,6 +121,8 @@ public class ClientMutationIntegrationTest {
   private static final Format KEY_FORMAT = FormatFactory.JSON;
   private static final Format VALUE_FORMAT = FormatFactory.JSON;
 
+  private static final String TEST_TABLE = "TEST_TABLE";
+  private static final String NON_EXIST_TABLE = "NON_EXIST_TABLE";
   private static final String AGG_TABLE = "AGG_TABLE";
   private static final PhysicalSchema AGG_SCHEMA = PhysicalSchema.from(
       LogicalSchema.builder()
@@ -179,6 +181,9 @@ public class ClientMutationIntegrationTest {
     makeKsqlRequest("CREATE TABLE " + AGG_TABLE + " AS "
         + "SELECT K, LATEST_BY_OFFSET(LONG) AS LONG FROM " + TEST_STREAM + " GROUP BY K;"
     );
+    makeKsqlRequest(
+        "CREATE TABLE " + TEST_TABLE + " (K STRING PRIMARY KEY, LONG BIGINT) WITH (KAFKA_TOPIC='"
+            + TEST_TABLE + "', VALUE_FORMAT='json', PARTITIONS=1);");
 
     TEST_HARNESS.verifyAvailableUniqueRows(
         AGG_TABLE,
@@ -410,13 +415,42 @@ public class ClientMutationIntegrationTest {
     // When
     final Exception e = assertThrows(
         ExecutionException.class, // thrown from .get() when the future completes exceptionally
-        () -> client.insertInto(AGG_TABLE, insertRow).get()
+        () -> client.insertInto(NON_EXIST_TABLE, insertRow).get()
     );
 
     // Then
     assertThat(e.getCause(), instanceOf(KsqlClientException.class));
     assertThat(e.getCause().getMessage(), containsString("Received 400 response from server"));
-    assertThat(e.getCause().getMessage(), containsString("Cannot insert into a table"));
+    assertThat(e.getCause().getMessage(), containsString("Cannot insert values into an unknown stream/table"));
+  }
+
+  @Test
+  public void shouldInsertIntoTable() throws Exception {
+    // Given
+    final Map<String, Object> properties = new HashMap<>();
+    properties.put("auto.offset.reset", "earliest");
+
+    final KsqlObject insertRow = new KsqlObject()
+        .put("K", "my_key")
+        .put("LONG", 11L);
+
+    // When
+    final String query = "SELECT * from " + TEST_TABLE + " WHERE K='my_key' EMIT CHANGES LIMIT 1;";
+    StreamedQueryResult queryResult = client.streamQuery(query, properties).get();
+
+    final Row row = assertThatEventually(() -> {
+      // Potentially try inserting multiple times, in case the query wasn't started by the first time
+      try {
+        client.insertInto(TEST_TABLE, insertRow).get();
+      } catch (Exception e) {
+        throw new RuntimeException(e);
+      }
+      return queryResult.poll(Duration.ofMillis(10));
+    }, is(notNullValue()));
+
+    // Then: a newly inserted row arrives
+    assertThat(row.getString("K"), is("my_key"));
+    assertThat(row.getLong("LONG"), is(11L));
   }
 
   @Test
@@ -494,13 +528,14 @@ public class ClientMutationIntegrationTest {
     // When
     final Exception e = assertThrows(
         ExecutionException.class, // thrown from .get() when the future completes exceptionally
-        () -> client.streamInserts(AGG_TABLE, new InsertsPublisher()).get()
+        () -> client.streamInserts(NON_EXIST_TABLE, new InsertsPublisher()).get()
     );
 
     // Then
     assertThat(e.getCause(), instanceOf(KsqlClientException.class));
     assertThat(e.getCause().getMessage(), containsString("Received 400 response from server"));
-    assertThat(e.getCause().getMessage(), containsString("Cannot insert into a table"));
+    assertThat(e.getCause().getMessage(),
+        containsString("Cannot insert values into an unknown stream/table"));
   }
 
   @Test
--- a/ksqldb-rest-app/src/test/java/io/confluent/ksql/api/integration/ApiIntegrationTest.java
+++ b/ksqldb-rest-app/src/test/java/io/confluent/ksql/api/integration/ApiIntegrationTest.java
@@ -514,7 +514,8 @@ public class ApiIntegrationTest {
         .put("COMPLEX", COMPLEX_FIELD_VALUE);
 
     // Then: request fails because stream name is invalid
-    shouldRejectInsertRequest(target, row, "Cannot insert values into an unknown stream: " + target);
+    shouldRejectInsertRequest(target, row,
+        "Cannot insert values into an unknown stream/table: " + target);
   }
 
   @Test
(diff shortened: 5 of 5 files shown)


ksqldb-api-client/src/test/java/io/confluent/ksql/api/client/integration/ClientIntegrationTest.java
```diff
@@ -634,21 +652,25 @@ public void shouldInsertInto() throws Exception {
         .put("TIME", "00:00:01");
 
     // When
-    client.insertInto(EMPTY_TEST_STREAM.toLowerCase(), insertRow).get(); // Stream name is case-insensitive
+    client.insertInto(EMPTY_TEST_STREAM.toLowerCase(), insertRow)
+        .get(); // Stream name is case-insensitive
 
     // Then: should receive new row
     final String query = "SELECT * FROM " + EMPTY_TEST_STREAM + " EMIT CHANGES LIMIT 1;";
+    final StreamedQueryResult streamedQueryResult = client.streamQuery(PULL_QUERY_ON_TABLE).get();
```
mjsax, line 660: Seems this is unused?

ksqldb-api-client/src/test/java/io/confluent/ksql/api/client/integration/ClientIntegrationTest.java
```diff
@@ -657,22 +679,111 @@ public void shouldInsertInto() throws Exception {
   }
 
   @Test
-  public void shouldHandleErrorResponseFromInsertInto() {
+  public void shouldPullQueryTableAfterInsert() throws Exception {
```
mjsax, line 682: A regular table should _not_ be pull-queryable? Is the test name incorrect?

Overall, I am wondering what you try to do: insert into the table and verify that the data was written? For this case you could use a push query in the test.

ksqldb-api-client/src/test/java/io/confluent/ksql/api/client/integration/ClientIntegrationTest.java
```diff
@@ -657,22 +679,111 @@ public void shouldInsertInto() throws Exception {
...
+
+    final KsqlObject insertRow = new KsqlObject()
+        .put("K", "my_key")
+        .put("LONG", 11L);
+
+    // When
+    final String query = "SELECT * from " + TEST_TABLE + " WHERE K='my_key' EMIT CHANGES LIMIT 1;";
+    StreamedQueryResult queryResult = client.streamQuery(query, properties).get();
+
+    final Row row = assertThatEventually(() -> {
```
mjsax, line 718: Does `assertThatEventually` apply any timeout?

ksqldb-api-client/src/test/java/io/confluent/ksql/api/client/integration/ClientIntegrationTest.java
```diff
@@ -657,22 +679,111 @@ public void shouldInsertInto() throws Exception {
...
+      return queryResult.poll(Duration.ofMillis(10));
+    }, is(notNullValue()));
+
+    // Then: a newly inserted row arrives
+    assertThat(row.getString("K"), is("my_key"));
+    assertThat(row.getLong("LONG"), is(11L));
+  }
+
+  @Test
+  public void shouldPullQueryCTASTableAfterInsert() throws Exception {
```
mjsax, line 734: Not sure if this test buys us much?

ksqldb-api-client/src/test/java/io/confluent/ksql/api/client/integration/ClientIntegrationTest.java
```diff
@@ -657,22 +679,111 @@ public void shouldInsertInto() throws Exception {
...
+    final List<Row> rows = client.executeQuery(query).get();
+
+    // Verify inserted row is as expected
+    // Pull query read from state store, inserting into CTAS table won't
+    // insert to state store
+    assertThat(rows, hasSize(0));
+  }
+
+  @Test
+  public void shouldPullQueryCTASTableAfterInsertStream() throws Exception {
```
mjsax, line 755: `insert into stream` did always work, so it seems the test is actually not related to this PR?

ksqldb-engine/src/test/java/io/confluent/ksql/integration/IntegrationTestHarness.java
```diff
@@ -133,7 +133,7 @@ public boolean topicExists(final String topicName) {
   /**
    * Ensure topics with the given {@code topicNames} exist.
    *
-   * <p>Topics will be creates, if they do not already exist, with a single partition and replica.
+   * <p>Topics will be created, if they do not already exist, with a single partition and replica.
```
mjsax, line 136: Nice catch!

ksqldb-api-client/src/test/java/io/confluent/ksql/api/client/integration/ClientIntegrationTest.java
lihaosky: Good catch 👍

ksqldb-api-client/src/test/java/io/confluent/ksql/api/client/integration/ClientIntegrationTest.java
lihaosky: Yeah, test name should be `shouldNot...` 

I'm trying to test you can insert into table created using `with` but pull query it will throw exception

ksqldb-api-client/src/test/java/io/confluent/ksql/api/client/integration/ClientIntegrationTest.java
lihaosky: Trying to make sure direct insert to CTAS table will have no effect

ksqldb-api-client/src/test/java/io/confluent/ksql/api/client/integration/ClientIntegrationTest.java
lihaosky: Want to test inserting to stream which backs CTAS table and it will be inserted to table. I'm trying to cover all possible ways to insert to a table.

ksqldb-api-client/src/test/java/io/confluent/ksql/api/client/integration/ClientIntegrationTest.java
lihaosky: Yes. It has default timeout:

https://github.com/confluentinc/ksql/blob/dd36ca7728103f9fad6fd83d797f70b52ffc5bd0/ksqldb-test-util/src/main/java/io/confluent/ksql/test/util/AssertEventually.java#L101

mjsax:
Seems there is some spotbug issue:
```
[ERROR] Medium: Dead store to streamedQueryResult in io.confluent.ksql.api.client.integration.ClientIntegrationTest.shouldInsertInto() [io.confluent.ksql.api.client.integration.ClientIntegrationTest] At ClientIntegrationTest.java:[line 660] DLS_DEAD_LOCAL_STORE
```

ksqldb-api-client/src/test/java/io/confluent/ksql/api/client/integration/ClientIntegrationTest.java
mjsax: My question is really: this PR is about allowing INSERT INTO for tables -- however, INSERT INTO and pull-queries have nothing to do with each other. So what do we gain by issuing a pull query (that we know is not supported for this case anyway)? -- Overall, it seems that the second part of this test is unnecessary and the first part is covered by the next test that uses a push query to verify if the insert was successful. Thus, should we remove this test?

ksqldb-api-client/src/test/java/io/confluent/ksql/api/client/integration/ClientIntegrationTest.java
mjsax: What do you mean by "will not affected" exaclty?

ksqldb-api-client/src/test/java/io/confluent/ksql/api/client/integration/ClientIntegrationTest.java
mjsax: If this closes a test gap, great. I was just wondering if this case was already covered or not.

ksqldb-api-client/src/test/java/io/confluent/ksql/api/client/integration/ClientIntegrationTest.java
lihaosky: I mean you can insert to the table successfully but you won't see the result in pull request

lihaosky (commented):
> Seems there is some spotbug issue:
> 
> ```
> [ERROR] Medium: Dead store to streamedQueryResult in io.confluent.ksql.api.client.integration.ClientIntegrationTest.shouldInsertInto() [io.confluent.ksql.api.client.integration.ClientIntegrationTest] At ClientIntegrationTest.java:[line 660] DLS_DEAD_LOCAL_STORE
> ```
> 
> Might be fixed already by removing the unused variable.

Yeah. It's checkstyle failure and I fixed it

ksqldb-api-client/src/test/java/io/confluent/ksql/api/client/integration/ClientIntegrationTest.java
mjsax: > I mean you can insert to the table successfully but you won't see the result in pull request

Well, while this is true, to me it's rather a design flaw / bug than a feature. Thus, I am not sure if it makes sense to test this case?

mjsax (approved):
Overall LGTM. Call for second review @vcrfxia to sanity check, this part of the code is also new to me. -- In particular if we should/need to test "insert into" a table created via CTAS.

ksqldb-api-client/src/test/java/io/confluent/ksql/api/client/ClientTest.java
```diff
@@ -591,24 +591,6 @@ public void shouldInsertInto() throws Exception {
     assertThat(testEndpoints.getLastTarget(), is("test-stream"));
   }
 
-  @Test
-  public void shouldHandleErrorResponseFromInsertInto() {
```
vcrfxia, line 595: Why are these tests being removed? They are validating that an informative 400 response error message is returned to the user. We can update the exception message to be more realistic (in light of the changes in this PR) but I think we should keep the test coverage.

ksqldb-api-client/src/test/java/io/confluent/ksql/api/client/integration/ClientIntegrationTest.java
```diff
@@ -255,7 +259,12 @@ public static void setUpClass() throws Exception {
         AGG_SCHEMA
     );
 
-    final String testDir = Paths.get(TestUtils.tempDirectory().getAbsolutePath(), "client_integ_test").toString();
+    makeKsqlRequest(
```
vcrfxia, line 262: nit: can we move this up a few lines to be alongside the other `makeKsqlRequest()`?

ksqldb-api-client/src/test/java/io/confluent/ksql/api/client/integration/ClientIntegrationTest.java
```diff
@@ -657,22 +678,103 @@ public void shouldInsertInto() throws Exception {
   }
 
   @Test
-  public void shouldHandleErrorResponseFromInsertInto() {
+  public void shouldHandlePullQueryTableException() throws Exception {
```
vcrfxia, line 681: Why is this test being added? Testing the specific error (query against a table this is not query-able) doesn't make sense in this class since it's not specific to the client. Regarding client behavior for this error, this test file already contains `shouldHandleErrorResponseFromExecuteQuery()` and `shouldHandleErrorResponseFromStreamQuery()`.

ksqldb-api-client/src/test/java/io/confluent/ksql/api/client/integration/ClientIntegrationTest.java
```diff
@@ -657,22 +678,103 @@ public void shouldInsertInto() throws Exception {
...
+    );
+
+    // Then
+    assertThat(e.getCause(), instanceOf(KsqlClientException.class));
+    assertThat(e.getCause().getMessage(), containsString("Received 400 response from server"));
+    assertThat(e.getCause().getMessage(), containsString("table isn't queryable"));
+  }
+
+  @Test
+  public void shouldPushQueryTableAfterInsert() throws Exception {
```
vcrfxia, line 696: The existing test `shouldInsertInto()` already issues a push query after an `insertInto()` call. What additional coverage does this test buy us?

ksqldb-api-client/src/test/java/io/confluent/ksql/api/client/integration/ClientIntegrationTest.java
vcrfxia: +1 to Matthias: I don't think this test makes sense in this file. This test file is meant to test client-specific behavior. If the purpose of this test is to verify that INSERT VALUES does not update pull query state stores, then the test does not belong in this file. (In general, integration tests are (relatively) expensive so I'd err against adding redundant/duplicate tests.)

ksqldb-api-client/src/test/java/io/confluent/ksql/api/client/integration/ClientIntegrationTest.java
vcrfxia: Ditto to my comment above -- I'm not convinced this test belongs in this file. I definitely think we should have a test that verifies that INSERT VALUES is allowed for tables, but I don't think we need to test auxiliary behaviors.

ksqldb-api-client/src/test/java/io/confluent/ksql/api/client/integration/ClientIntegrationTest.java
```diff
@@ -657,22 +678,103 @@ public void shouldInsertInto() throws Exception {
   }
 
   @Test
-  public void shouldHandleErrorResponseFromInsertInto() {
```
vcrfxia, line 660: This is important test coverage (that error responses are properly exposed to the user). If we are removing the specific cause of the error used in the existing test, then let's replace the error with a different error thrown from the endpoint, rather than removing the test entirely.

ksqldb-api-client/src/test/java/io/confluent/ksql/api/client/integration/ClientIntegrationTest.java
```diff
@@ -858,25 +975,12 @@ public void shouldStreamInserts() throws Exception {
     assertThat(acksPublisher.isFailed(), is(false));
   }
 
-  @Test
-  public void shouldHandleErrorResponseFromStreamInserts() {
```
vcrfxia, line 862: As above: this is important test coverage (that error responses are properly exposed to the user). If we are removing the specific cause of the error used in the existing test, then let's replace the error with a different error thrown from the endpoint, rather than removing the test entirely.

vcrfxia (commented):
Thanks for the fix @lihaosky ! Some comments about test coverage inline.

I also updated the PR title to be more informative to users for when it's eventually included in the changelog for a release.

ksqldb-api-client/src/test/java/io/confluent/ksql/api/client/ClientTest.java
mjsax: But an "insert into" should be supported now? What kind of error do you have in mind that we should test?

vcrfxia: The purpose of this test is to check that _any_ error from the insert values endpoint is properly propagated to the user. If you look at the code for the endpoint, you'll see that there are plenty of other errors that could be triggered (besides attempting to insert into a table), e.g., trying to insert into a non-existent source.

vcrfxia: Deleting this test opens us up to potential regressions where the client does not properly propagate error messages from the server response to the user. We should not delete this coverage.

lihaosky: Hi @vcrfxia, do you mean we should also test handling error like these:
https://github.com/confluentinc/ksql/blob/master/ksqldb-rest-app/src/main/java/io/confluent/ksql/api/impl/InsertsStreamEndpoint.java#L61-L80

I agree we should cover those but that seems belong to another PR since these PR is about allowing insert to table. I can create another issue about above missing tests and fix those in a separate PR. What do you think?

ksqldb-api-client/src/test/java/io/confluent/ksql/api/client/integration/ClientIntegrationTest.java
lihaosky: As mentioned above, I can create separate PR to add more tests catching different exceptions

ksqldb-api-client/src/test/java/io/confluent/ksql/api/client/integration/ClientIntegrationTest.java
lihaosky: I'm trying to cover more error cases. This is client behavior triggering exception by query non-queryable table. `shouldHandleErrorResponseFromExecuteQuery` and `shouldHandleErrorResponseFromStreamQuery` seem cover other behavior like query non-exist table. I feel query non-queryable table is also one of the cases. If you feel this is not needed, I can remove this

ksqldb-api-client/src/test/java/io/confluent/ksql/api/client/integration/ClientIntegrationTest.java
lihaosky: `shouldInsertInto` insert into a stream `EMPTY_TEST_STREAM `, this one `insertInto` a table.

ksqldb-api-client/src/test/java/io/confluent/ksql/api/client/integration/ClientIntegrationTest.java
lihaosky: Mentioned above, I can add more case in different PR

ksqldb-api-client/src/test/java/io/confluent/ksql/api/client/integration/ClientIntegrationTest.java
lihaosky: Sure. I'll remove this

ksqldb-api-client/src/test/java/io/confluent/ksql/api/client/integration/ClientIntegrationTest.java
lihaosky: I can remove this as well

ksqldb-api-client/src/test/java/io/confluent/ksql/api/client/ClientTest.java
vcrfxia: I'm saying we should have one test to validate that _any_ error is properly propagated from the server to the user (via the client). We do not need to test every possible server error. The proper place for testing each possible server error would be in a unit/integration test for the server endpoint, not the client. The client tests should just contain a single test to validate that any error is properly propagated to the user.

vcrfxia: I think we should simply update the test highlighted in this comment (in this PR) to test a different error, rather than removing the test in this PR and opening a follow-up PR to add it back.

ksqldb-api-client/src/test/java/io/confluent/ksql/api/client/integration/ClientIntegrationTest.java
vcrfxia: As mentioned above, whether the table is queryable or not is not relevant to the client. For client test coverage, we should simply check that the client properly propagates errors from the server to the user. This is not the right place to try exercising all possible server-side errors. That belongs in the server unit/integration tests instead.

ksqldb-api-client/src/test/java/io/confluent/ksql/api/client/integration/ClientIntegrationTest.java
vcrfxia: If that's the purpose of this new test, can we rename the test to `shouldInsertIntoTable()` or similar, in order to better capture what the test is for?

ksqldb-api-client/src/test/java/io/confluent/ksql/api/client/integration/ClientIntegrationTest.java
vcrfxia: As above, I don't think we should be removing test coverage in this PR. Let me know if what I'm saying isn't making sense, I'm happy to sync offline.

ksqldb-api-client/src/test/java/io/confluent/ksql/api/client/integration/ClientIntegrationTest.java
vcrfxia: As above: https://github.com/confluentinc/ksql/pull/8114#discussion_r709545064

lihaosky:
Synced with @vcrfxia offline. Will
* Add insert exception handling back in unit test
* Remove unnecessary test in integration test

Thanks @mjsax and @vcrfxia for reviewing

ksqldb-api-client/src/test/java/io/confluent/ksql/api/client/integration/ClientIntegrationTest.java
```diff
@@ -666,13 +686,42 @@ public void shouldHandleErrorResponseFromInsertInto() {
...
         ExecutionException.class, // thrown from .get() when the future completes exceptionally
-        () -> client.insertInto(AGG_TABLE, insertRow).get()
+        () -> client.insertInto("nonexist_table", insertRow).get()
     );
 
     // Then
     assertThat(e.getCause(), instanceOf(KsqlClientException.class));
     assertThat(e.getCause().getMessage(), containsString("Received 400 response from server"));
-    assertThat(e.getCause().getMessage(), containsString("Cannot insert into a table"));
+    assertThat(e.getCause().getMessage(), containsString("Cannot insert values into an unknown stream"));
```
vcrfxia, line 695: Let's update the error message (in the actual server endpoint) so that it says "unknown stream/table" rather than "unknown stream".

vcrfxia (commented):
Thanks @lihaosky -- LGTM besides the minor comment inline.

ksqldb-rest-app/src/main/java/io/confluent/ksql/api/impl/InsertsStreamEndpoint.java
```diff
@@ -89,7 +86,8 @@ private DataSource getDataSource(
     final DataSource dataSource = metaStore.getSource(sourceName);
     if (dataSource == null) {
       throw new KsqlApiException(
-          "Cannot insert values into an unknown stream: " + sourceName, ERROR_CODE_BAD_STATEMENT);
+          "Cannot insert values into an unknown stream/table: " + sourceName,
+            ERROR_CODE_BAD_STATEMENT);
     }
 
     if (dataSource.getKsqlTopic().getKeyFormat().isWindowed()) {
```
vcrfxia, line 93: Can we also update the error message in this block to say `stream/table` rather than just `stream`?

vcrfxia (approved):
LGTM -- thanks @lihaosky !

PR #8115, opened 2021-09-09
feat: disable stream pull queries by default

vvcephei:
Since stream pull queries are only implemented for the HTTP/1 REST endpoint, they will be disabled by default.

The config is overridable on a per-query basis, though, in case people want to use the feature in a preview state.

### Reviewer checklist


vpapavas (approved):
Lacking background of why but LGTM :)

vvcephei:
Thanks, @vpapavas ! 

Jenkins failed on a unit test that I forgot: `io.confluent.ksql.engine.KsqlEngineTest#shouldCheckStreamPullQueryEnabledFlag`

I fixed the test, and I'll watch for the job to complete before merging.

PR #8120, opened 2021-09-10
chore: feature flag for ROWPARTITION and ROWOFFSET

Sullivan-Patrick:
### Description 
When set to true, this enables `ROWPARTITION` and `ROWOFFSET` pseudocolumns. This includes using the V2 table source step, to allow for joins (see https://github.com/confluentinc/ksql/pull/7990), as well as setting the `pseudoColumnVersion` field of all `SourceStep`s to 1, prompting them to use the new pseudocolumns.

With or without the flag enabled, on this commit serialization of the `pseudoColumnVersion` field will occur.

### Testing done 
Unit tests changed to allow for compilation with the `KsqlConfig` piped in, QTTs changed to expect new behaviors.

### Reviewer checklist


vcrfxia (commented):
Thanks @Sullivan-Patrick ! Great to see this getting in so we can get more test coverage on the feature in development. Comments inline.

ksqldb-execution/src/main/java/io/confluent/ksql/execution/ddl/commands/CreateSourceCommand.java
```diff
@@ -95,9 +94,6 @@ public Boolean getIsSource() {
   }
 
   private static void validate(final LogicalSchema schema, final boolean windowed) {
-    if (schema.valueContainsAny(SystemColumns.systemColumnNames())) {
```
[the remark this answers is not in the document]
Sullivan-Patrick: Okay, thanks for pointing this out. It makes sense that I'd want to decouple the changes, I'll keep that in for future work. Side note: `SingleColumn` is probably going to get a `KsqlConfig` piped into it after all, as I found with some more testing that it allowed users to alias columns with reserved names in push queries.

ksqldb-engine/src/main/java/io/confluent/ksql/engine/rewrite/AstSanitizer.java
```diff
@@ -66,19 +68,28 @@
   private AstSanitizer() {
   }
 
-
+  @VisibleForTesting
```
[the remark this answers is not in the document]
Sullivan-Patrick: I was going to ask about this but I forgot, thanks for the clarification.

ksqldb-engine/src/main/java/io/confluent/ksql/planner/plan/DataSourceNode.java
```diff
@@ -51,28 +52,32 @@
...
+      final boolean isWindowed,
+      final KsqlConfig ksqlConfig
   ) {
     super(id, dataSource.getDataSourceType(), Optional.of(alias));
-    this.schema = buildSchema(dataSource);
+    this.schema = buildSchema(dataSource, ksqlConfig);
     this.dataSource = requireNonNull(dataSource, "dataSource");
     this.schemaKStreamFactory = requireNonNull(schemaKStreamFactory, "schemaKStreamFactory");
     this.isWindowed = isWindowed;
+    this.ksqlConfig = ksqlConfig;
```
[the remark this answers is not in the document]
Sullivan-Patrick: Sounds good on fixing them all. I haven't thought about it too much prior, but I love when stuff fails fast and using `requireNonNull` everywhere is something I can get behind.

ksqldb-engine/src/main/java/io/confluent/ksql/planner/plan/QueryProjectNode.java
```diff
@@ -76,6 +76,7 @@
   private final boolean addAdditionalColumnsToIntermediateSchema;
   private final KsqlConfig ksqlConfig;
 
+  @SuppressFBWarnings("EI_EXPOSE_REP2")
```
Sullivan-Patrick, line 79: I suppressed this warning as I didn't see how adding this config added any mutability. Configs are effectively immutable, right?

ksqldb-common/src/main/java/io/confluent/ksql/schema/ksql/SystemColumns.java
```diff
@@ -115,11 +117,18 @@ public static boolean isPseudoColumn(
...
 
   public static Set<ColumnName> pseudoColumnNames(final int pseudoColumnVersion) {
-    if (!PSEUDO_COLUMN_NAMES_BY_VERSION.containsKey(pseudoColumnVersion)) {
+    final Set<ColumnName> names = pseudoColumns
+        .stream()
+        .filter(col -> col.version <= pseudoColumnVersion)
+        .map(col -> col.name)
+        .collect(Collectors.toSet());
+
+    if (names.isEmpty()) {
```
vcrfxia, line 126: This check is no longer the same as the previous check. Previously, we were checking whether the pseudocolumn version passed was valid or not. We should continue to do that by checking that `pseudoColumnVersion` is at least `LEGACY_PSEUDOCOLUMN_VERSION_NUMBER` and no greater than `CURRENT_PSEUDOCOLUMN_VERSION_NUMBER`.

ksqldb-common/src/main/java/io/confluent/ksql/schema/ksql/SystemColumns.java
```diff
@@ -104,7 +104,9 @@ public static boolean isWindowBound(final ColumnName columnName) {
 
   @VisibleForTesting
   static boolean isPseudoColumn(final ColumnName columnName, final int pseudoColumnVersion) {
-    return pseudoColumnNames(pseudoColumnVersion).contains(columnName);
+    return pseudoColumns
+        .stream()
+        .anyMatch(col -> col.name.equals(columnName) && col.version <= pseudoColumnVersion);
```
vcrfxia, line 109: nit: we can make this slightly more readable by turning the second part of the boolean expression into a call to `filter()` before `anyMatch()`.

ksqldb-common/src/main/java/io/confluent/ksql/schema/ksql/SystemColumns.java
```diff
@@ -134,28 +143,58 @@ public static boolean isSystemColumn(final ColumnName columnName, final KsqlConf
...
-  @SuppressFBWarnings(
-      value = "MS_EXPOSE_REP",
-      justification = "SYSTEM_COLUMN_NAMES is ImmutableSet"
-  )
-  private static Set<ColumnName> systemColumnNames(final int pseudoColumnVersion) {
-    return SYSTEM_COLUMN_NAMES_BY_VERSION.get(pseudoColumnVersion);
+  public static Set<ColumnName> systemColumnNames(final int pseudoColumnVersion) {
+    return Stream.concat(
+        WINDOW_BOUNDS_COLUMN_NAMES.stream(),
+        pseudoColumns
```
vcrfxia, line 149: Call `pseudoColumnNames(...)`, rather than duplicating the code here.

ksqldb-common/src/main/java/io/confluent/ksql/schema/ksql/SystemColumns.java
```diff
@@ -134,28 +143,58 @@ public static boolean isSystemColumn(final ColumnName columnName, final KsqlConf
...
+            .stream()
+            .filter(col -> col.version <= pseudoColumnVersion)
+            .map(col -> col.name)
+    )
+        .collect(Collectors.toSet());
   }
 
   public static boolean mustBeMaterializedForTableJoins(final ColumnName columnName) {
-    return MUST_BE_MATERIALIZED_FOR_TABLE_JOINS.contains(columnName);
+    return pseudoColumns
```
vcrfxia, line 158: I'm confused about the contract of this method (`mustBeMaterializedForTableJoins`). The previous implementation did not throw if called on a non-pseudocolumn, but this new implementation does. Was this change in behavior from an earlier version of the PR intentional? Let's add a javadoc to clarify.

ksqldb-common/src/main/java/io/confluent/ksql/schema/ksql/SystemColumns.java
```diff
@@ -134,28 +143,58 @@ public static boolean isSystemColumn(final ColumnName columnName, final KsqlConf
...
+    final int version;
+    final boolean mustBeMaterializedForTableJoins;
+
+    private PseudoColumn(
+        final ColumnName name,
+        final SqlType type,
+        final int version,
+        final boolean mustBeMaterializedForTableJoins
+    ) {
+      this.name = name;
```
vcrfxia, line 185: nit: null checks.

ksqldb-common/src/main/java/io/confluent/ksql/schema/ksql/SystemColumns.java
```diff
@@ -134,28 +143,58 @@ public static boolean isSystemColumn(final ColumnName columnName, final KsqlConf
...
+      this.type = type;
+      this.version = version;
+      this.mustBeMaterializedForTableJoins = mustBeMaterializedForTableJoins;
+    }
+
+    private static PseudoColumn of(
```
vcrfxia, line 191: Out of curiosity, what's the benefit of having a static creator method here?

ksqldb-common/src/main/java/io/confluent/ksql/schema/ksql/SystemColumns.java
```diff
@@ -57,35 +60,32 @@
...
+      ROWPARTITION_ROWOFFSET_PSEUDOCOLUMN_VERSION,
+      true
+  );
+
+  private static final PseudoColumn ROWOFFSET = PseudoColumn.of(
+      ROWOFFSET_NAME,
+      ROWOFFSET_TYPE,
+      ROWPARTITION_ROWOFFSET_PSEUDOCOLUMN_VERSION,
+      true
+  );
```
vcrfxia, line 82: nit: inline these into the creation of `pseudoColumn` below. They are not used elsewhere and the code is slightly more readable by having fewer static variables IMO.

ksqldb-common/src/test/java/io/confluent/ksql/schema/ksql/LogicalSchemaTest.java
```diff
@@ -44,15 +44,23 @@
...
 import org.junit.Test;
+import org.mockito.Mock;
+import org.mockito.junit.MockitoJUnit;
+import org.mockito.junit.MockitoRule;
 
 @SuppressWarnings({"UnstableApiUsage", "unchecked"})
 public class LogicalSchemaTest {
 
+  @Rule
+  public final MockitoRule rule = MockitoJUnit.rule();
```
vcrfxia, line 62: nit: use `@RunWith(MockitoJUnitRunner.class)` instead to be consistent with all the other test files using Mockito in this codebase.

ksqldb-engine/src/main/java/io/confluent/ksql/analyzer/Analyzer.java
```diff
@@ -594,21 +594,44 @@ private void analyzeHaving(final Expression node) {
...
 
+    private void checkForReservedToken(
+        final SingleColumn singleColumn,
+        final ColumnName reservedToken
+    ) {
+      final ColumnName alias = singleColumn.getAlias().orElseThrow(IllegalStateException::new);
+      final Expression expression = singleColumn.getExpression();
+
+      if (alias.text().equalsIgnoreCase(reservedToken.text())) {
+        if (!expressionMatchesAlias(expression, alias)) {
```
vcrfxia, line 616: Can you add comments to explain both the if and else-if branches of this block? The logic here is not easy for readers to grasp by looking at the code. I'd be confused if not for the fact that we talked about this offline.

ksqldb-engine/src/main/java/io/confluent/ksql/analyzer/SourceSchemas.java
```diff
@@ -42,7 +42,7 @@
       final Map<SourceName, LogicalSchema> sourceSchemas,
       final KsqlConfig ksqlConfig) {
     this.sourceSchemas = ImmutableMap.copyOf(requireNonNull(sourceSchemas, "sourceSchemas"));
-    this.ksqlConfig = ksqlConfig;
+    this.ksqlConfig = requireNonNull(ksqlConfig);
```
vcrfxia, line 45: Use the two-arg version of `requireNonNull` instead, and same for all the other places (in other classes) you've added the one-arg version as well.

ksqldb-execution/src/main/java/io/confluent/ksql/execution/ddl/commands/CreateSourceCommand.java
```diff
@@ -94,6 +95,9 @@ public Boolean getIsSource() {
   }
 
   private static void validate(final LogicalSchema schema, final boolean windowed) {
+    if (schema.valueContainsAny(SystemColumns.windowBoundsColumnNames())) {
```
vcrfxia, line 98: What is the purpose of this change (from `systemColumnNames()` to `windowBoundsColumnNames()`)?

ksqldb-streams/src/main/java/io/confluent/ksql/execution/streams/SourceBuilder.java
```diff
@@ -192,9 +194,15 @@ private static Formats of(final KeyFormat keyFormat, final FormatInfo valueForma
       implements ValueTransformerWithKeySupplier<K, GenericRow, GenericRow> {
 
     private final int pseudoColumnVersion;
+    private final int numPseudoColumnsToMaterialize;
 
     AddPseudoColumnsToMaterialize(final int pseudoColumnVersion) {
       this.pseudoColumnVersion = pseudoColumnVersion;
+      this.numPseudoColumnsToMaterialize =
```
vcrfxia, line 201: The other `ValueTransformerWithKeySupplier` in this file also needs an analogous change.

ksqldb-streams/src/main/java/io/confluent/ksql/execution/streams/SourceBuilder.java
```diff
@@ -112,7 +112,9 @@ public static SourceBuilder instance() {
       // we do not need to force a materialization at this source step since the
       // re-partitioned topic will be used for any subsequent state stores, in lieu
       // of the original source topic, thus avoiding the issues above.
-      // See https://github.com/confluentinc/ksql/issues/6650
+      // See https://github.com/confluentinc/ksql/issues/6650.
+      // This intermittent transformValues() call is unneccessary in this case,
```
vcrfxia, line 116:
```suggestion
      // This intermittent transformValues() call is unneccessary in this case,
      // i.e., it could be combined with the subsequent transformValues() call into a single transformValues() call,
```

ksqldb-functional-tests/src/test/resources/query-validation-tests/row-offset-and-partition.json
```diff
@@ -22,13 +19,11 @@
         "sources": [
           {"name": "OUTPUT", "type": "stream", "schema": "ID INT KEY, ROWPARTITION STRING, ROWOFFSET STRING"}
         ]
-      }
+      },
+      "comments": "Should be removed after rowpartition.rowoffset.enabled feature flag is removed"
```
vcrfxia, line 23: Almost but not quite -- when the feature flag is removed we'll actually want to remove the other versions of these tests (the newly added ones with `- ROWOFFSET/ROWPARTITION enabled` in the names).

Also, we typically put `comments` above `name` so that readers know to read the comments (typically containing info relevant for understanding the rest of the test) before actually reading the test.

vcrfxia (commented):
Thanks @Sullivan-Patrick -- SystemColumns.java is much cleaner now, with the new PseudoColumn abstraction!

ksqldb-common/src/main/java/io/confluent/ksql/schema/ksql/SystemColumns.java
Sullivan-Patrick: Yeah, I changed the contract as I think it aligns more with the general philosophy of the future methods which will be similar: the extra properties I'm adding to pseudocolumns in the next 2 PRs were best served by throwing exceptions if a non-system column name was provided.

ksqldb-common/src/main/java/io/confluent/ksql/schema/ksql/SystemColumns.java
Sullivan-Patrick: I was debating this, if you think it's good then I'm all for this change. I trust a reader to judge readability over a writer :)

ksqldb-common/src/main/java/io/confluent/ksql/schema/ksql/SystemColumns.java
Sullivan-Patrick: Just readability. If you think it hurts readability or doesn't contribute, I don't feel strongly about it either way.

ksqldb-execution/src/main/java/io/confluent/ksql/execution/ddl/commands/CreateSourceCommand.java
Sullivan-Patrick: This is one of the removed redundant validation steps we talked about, but the validation for window columns was necessary IIRC (also doesn't require piping a config, so no extra work).

ksqldb-functional-tests/src/test/resources/query-validation-tests/row-offset-and-partition.json
Sullivan-Patrick: At the risk of sounding dumb: I'm confused, don't we want to ensure that when the flag is enabled we are disallowing these reserved names to be used? I thought the whole reason we added these tests was so we could add them and only persist the historical plans. Since this is the last change, I'll push the branch without any changes for now until I understand why this is the case.

ksqldb-common/src/main/java/io/confluent/ksql/schema/ksql/SystemColumns.java
vcrfxia: I don't think it contributes but I'm also not opposed to leaving it, was just curious what your motivation was.

ksqldb-execution/src/main/java/io/confluent/ksql/execution/ddl/commands/CreateSourceCommand.java
vcrfxia: Why is the validation for window bounds column names different from that for other system column names? Are there other places where we check for all system column names except window bounds column names? That feels very strange.

ksqldb-functional-tests/src/test/resources/query-validation-tests/row-offset-and-partition.json
vcrfxia: > don't we want to ensure that when the flag is enabled we are disallowing these reserved names to be used?

We do, but when the flag is removed then there won't be anything special about ROWPARTITION and ROWOFFSET anymore; they'll just be regular pseudocolumns like all the others. So when that happens, it won't make sense to have a test called "- ROWPARTITION and ROWOFFSET enabled". We'll have to update the regular versions of the tests to account for ROWPARTITION and ROWOFFSET and so it's just a matter of deciding which test name we want. We should keep the original since the name is shorter and we won't have to introduce a new historic plan folder.

ksqldb-common/src/main/java/io/confluent/ksql/schema/ksql/SystemColumns.java
```diff
@@ -144,16 +139,24 @@ public static boolean isSystemColumn(final ColumnName columnName, final KsqlConf
   }
 
   public static Set<ColumnName> systemColumnNames(final int pseudoColumnVersion) {
+
+    validatePseudoColumnVersion(pseudoColumnVersion);
```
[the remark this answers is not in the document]
Sullivan-Patrick: Nice catch, thanks.

ksqldb-functional-tests/src/test/resources/query-validation-tests/row-offset-and-partition.json
Sullivan-Patrick: So we should change the functionality of the above test to be equivalent to  the `enabled` test after the flag is ungated? With a corresponding comment saying "should remove and change {other test} to reflect this behavior after feature flag is ungated" under the `-enabled` tests.

ksqldb-engine/src/main/java/io/confluent/ksql/engine/KsqlEngine.java
```diff
@@ -174,7 +174,7 @@ public KsqlEngine(
         1000,
         TimeUnit.MILLISECONDS
     );
-    this.ksqlConfig = ksqlConfig;
+    this.ksqlConfig = Objects.requireNonNull(ksqlConfig);
```
vcrfxia, line 177: Use the two-arg version of `requireNonNull()` rather than the one-arg version.

ksqldb-functional-tests/src/test/resources/query-validation-tests/row-offset-and-partition.json
vcrfxia: We don't need to leave a comment on this test since we'll be forced to change it once the flag is ungated -- otherwise the test will fail. We just need to leave a comment on the other test saying it should be removed once the flag is ungated, else we'll have duplicate tests.

ksqldb-engine/src/main/java/io/confluent/ksql/engine/KsqlEngine.java
Sullivan-Patrick: Ahh, that again. Sorry!

ksqldb-execution/src/main/java/io/confluent/ksql/execution/ddl/commands/CreateSourceCommand.java
Sullivan-Patrick: Now that I worked with the code more, I vaguely remember leaving this in just because I felt like the extra check was good. Even if that's a false memory, I tested and this check is indeed unnecessary. Now that I think about it more, I think it's better we fail columns aliased with window bound column names in the same place as we do other system columns (which I'm guessing is your conclusion too?).

ksqldb-engine/src/main/java/io/confluent/ksql/engine/KsqlEngine.java
vcrfxia: This conversation was marked resolved but I don't see the change made?

ksqldb-engine/src/main/java/io/confluent/ksql/structured/SchemaKSourceFactory.java
```diff
@@ -237,4 +243,10 @@ private static LogicalSchema resolveSchema(
         buildContext.getFunctionRegistry()
     ).resolve(step, dataSource.getSchema());
   }
+
+  private static int getPseudoColumnVersionFromConfig(final KsqlConfig ksqlConfig) {
```
vcrfxia, line 247: nit: use the public version in SystemColumns, rather than duplicating the code here.

ksqldb-engine/src/test/java/io/confluent/ksql/planner/plan/JoinNodeTest.java
```diff
@@ -638,10 +638,14 @@ public void shouldHaveFullyQualifiedJoinSchemaWithNonSyntheticKey() {
         .valueColumn(ColumnName.of(LEFT_ALIAS.text() + "_C0"), SqlTypes.BIGINT)
         .valueColumn(ColumnName.of(LEFT_ALIAS.text() + "_L1"), SqlTypes.STRING)
         .valueColumn(ColumnName.of(LEFT_ALIAS.text() + "_ROWTIME"), SqlTypes.BIGINT)
+        .valueColumn(ColumnName.of(LEFT_ALIAS.text() + "_ROWPARTITION"), SqlTypes.INTEGER)
```
vcrfxia, line 641: How come these unit tests are being updated? I thought the feature flag was disabled by default?

ksqldb-functional-tests/src/test/resources/query-validation-tests/partition-by.json
```diff
@@ -276,6 +282,52 @@
         ]
       }
     },
+    {
```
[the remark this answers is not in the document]
vcrfxia: This conversation was marked resolved but I don't see the changes?

ksqldb-rest-app/src/test/resources/ksql-plan-schema/schema.json
```diff
@@ -591,10 +591,13 @@
         },
         "sourceSchema" : {
           "type" : "string"
+        },
```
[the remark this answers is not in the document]
vcrfxia: Discussed offline -- it's because of the removal of the `@JsonIgnore` in SourceStep.java.

vcrfxia (commented):
Thanks @Sullivan-Patrick ! I think we're just about ready to merge. One question about some test changes inline.

ksqldb-execution/src/main/java/io/confluent/ksql/execution/ddl/commands/CreateSourceCommand.java
vcrfxia: If our existing test coverage for this (aliasing columns with a window bound name) is good and the removal of this check doesn't break tests, then yes I think it's preferable to keep all the system column checks in the same place. (To clarify, I don't know the state of existing test coverage for this scenario -- it'd be good to check and add tests for it if it's lacking.)

Sullivan-Patrick: We have coverage in `system-columns.json` and `CreateSourceFactoryTest`.

ksqldb-engine/src/test/java/io/confluent/ksql/planner/plan/JoinNodeTest.java
Sullivan-Patrick: This is because `LogicalSchema.withPseudoAndKeyColsInValue(boolean)`, which uses the `CURRENT_PSEUDOCOLUMN_VERSION_NUMBER` is being used by some unit tests . Since for all of those tests (except this one) the results are unaffected by which version we are using, I thought it'd be easier for whoever ends up removing this flag to just leave the current constructor as an overload instead of fully replacing it with a `(boolean, KsqlConfig`) constructor. In fact, in it's current state, `LogicalSchema.withPseudoAndKeyColsInValue(boolean)` should be annotated with `@VisibleForTesting`.

It's definitely not a major issue to mock the `KsqlConfig` into each unit test that uses it if you think that'd make things more explicit. Or I could do it just for this unit test. Or I could just annotate the constructor. I lean towards the second, just mocking this unit test (not entirely sure why I didn't before). WDYT?

vcrfxia: As long as we have an understanding of why this is happening, and that understanding lends confidence to the fact that this PR is non-functional changes only (all the new functionality is hidden behind the feature flag), then we're good. I do think it's preferable to mock the KsqlConfig in this unit test file, though, similar to what you've done for the other unit tests.

Sullivan-Patrick: SGTM, I'll do that.

ksqldb-engine/src/test/java/io/confluent/ksql/planner/plan/JoinNodeTest.java
```diff
@@ -231,6 +232,7 @@ public void setUp() {
 
   @Test
   public void shouldBuildSourceNode() {
+    when(ksqlConfig.getString(KsqlConfig.KSQL_DEFAULT_KEY_FORMAT_CONFIG)).thenReturn("abc");
```
vcrfxia, line 235: It's not great that this had to be added into individuals tests. I think it's preferable to simply disable the rowpartition/rowoffset feature flag in the `KsqlConfig()` constructor. Can you pick up this change in your next PR? I'm going to merge this one now to unblock subsequent work.

vcrfxia (approved):
Thanks @Sullivan-Patrick -- LGTM besides the minor follow-ups discussed (deferred to the next PR).

ksqldb-functional-tests/src/test/resources/query-validation-tests/partition-by.json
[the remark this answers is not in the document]
Sullivan-Patrick: I commented their corresponding methods instead (the ones that should be changed, not the ones that should be deleted). Fixed this in the `INSERT VALUES` pr

ksqldb-engine/src/main/java/io/confluent/ksql/engine/KsqlEngine.java
Sullivan-Patrick: Looks like I accidentally made the change in the pull query PR instead of this one...

Issue #8122, opened 2021-09-13
UDF EXTRACTJSONFIELD does not work in combination of TRANSFORM


stefanfrehse:
**Describe the bug**
Using the UDF `EXTRACTJSONFIELD` does not correctly parse JSON within a `TRANSFORM` lambda call while taking the path variable from the `TRANSFORM` argument. 

**To Reproduce**
Steps to reproduce the behavior, include:

* ksqlDb: 0.20.0 probably earlier as well
* Apply the following statements demonstrates the faulty behavior:
```sql
set 'auto.offset.reset'='earliest';
CREATE STREAM STREAM_A  (
    "key" VARCHAR KEY,
    "value" VARCHAR,
    "extractor" ARRAY<VARCHAR>
) WITH(KAFKA_TOPIC = 'test-topic-transform', VALUE_FORMAT = 'JSON', PARTITIONS = 1, REPLICAS = 3);

INSERT INTO STREAM_A ("key", "value", "extractor") VALUES( 'key_1', '{\"sensor_1\": 10, \"sensor_2\": 15}', ARRAY['.sensor_1', '.sensor_2']);
INSERT INTO STREAM_A ("key", "value", "extractor") VALUES( 'key_2', '{\"sensor_1\": 3, \"sensor_2\": 11}', ARRAY['.sensor_1', '.sensor_2']);

SELECT
    "key",
    "value",
    "extractor",
    TRANSFORM("extractor", (data) => EXTRACTJSONFIELD("value", ('$' + data))) AS "unpackedValue"
    FROM STREAM_A
EMIT CHANGES;
```
which results currently in `[10,10]`, and, `[3,3]`

**Expected behavior**
The result of the query should be: `[10,15]`, and, `[3,11]`.

spena:
Issue fixed here - https://github.com/confluentinc/ksql/pull/8123

PR #8123, opened 2021-09-13
fix: Allow multiple EXTRACTJSONFIELD invocations on different paths (#8122)

stefanfrehse:
### Description 
Allows the `extract` function to be invoked on the same instance multiple times with different `path` variables.

### Testing done 
A unit test has been added and the steps that reproduce the bug validated.

### Reviewer checklist

### DIFF
--- a/ksqldb-engine/src/main/java/io/confluent/ksql/function/udf/json/JsonExtractString.java
+++ b/ksqldb-engine/src/main/java/io/confluent/ksql/function/udf/json/JsonExtractString.java
@@ -37,7 +37,8 @@ public class JsonExtractString {
 
   private static final ObjectReader OBJECT_READER = UdfJsonMapper.INSTANCE.get().reader();
 
-  private List<String> tokens = null;
+  private String latestPath = null;
+  private List<String> latestTokens = null;
 
   @Udf
   public String extract(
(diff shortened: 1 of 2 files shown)


spena (approved):
It looks good.
+1

ksqldb-engine/src/main/java/io/confluent/ksql/function/udf/json/JsonExtractString.java
```diff
@@ -48,10 +46,10 @@ public String extract(
       return null;
     }
 
-    if (tokens == null) {
-      final JsonPathTokenizer tokenizer = new JsonPathTokenizer(path);
-      tokens = ImmutableList.copyOf(tokenizer);
-    }
+
+    final JsonPathTokenizer tokenizer = new JsonPathTokenizer(path);
+    final List<String> tokens = ImmutableList.copyOf(tokenizer);
```
spena, line 51: It should be good to keep a the `tokens` initialized once in case the `path` is equals to the latest path. What about adding a new class variable to keep the latest path seen, and check if the new path is equals to the latest path, if true, then re-use the current `tokens` class variable, if not, then initialize it with a new JsonPathTokenizer and List objects?

stefanfrehse: ok, thanks for the feedback. I will cover your points.

stefanfrehse: Done

spena:
@stefanfrehse I created a PR for testing your PR here - https://github.com/confluentinc/ksql/pull/8152. There are some limitations for allowing testing external contributions. I'll merge this PR once the tests pass. If they do not pass, could you check what needs to be fixed?

spena:
@stefanfrehse nvm, I see your job run, but failed. In the past, I've seen issue with contributors PRs not running. But this one just run, so you can take a look at the issue.

stefanfrehse:
@spena build fixed. Thanks to the static analysis, bug has been detected and is fixed now.

PR #8124, opened 2021-09-13
feat: Add stream pull queries to http2

vvcephei:
Adapting the stream pull query execution to the HTTP/2 endpoint, which wound up
being a significant refactor.

### Reviewer checklist


ksqldb-engine/src/main/java/io/confluent/ksql/query/QueryExecutor.java
```diff
@@ -510,17 +529,22 @@ private ProcessingLogger getUncaughtExceptionProcessingLogger(final QueryId quer
   private static TransientQueryQueue buildTransientQueryQueue(
       final Object buildResult,
       final OptionalInt limit,
-      final boolean excludeTombstones
+      final boolean excludeTombstones,
+      final Optional<ImmutableMap<TopicPartition, Long>> endOffsets
   ) {
+    // If we are supposed to stop at the endOffsets, but we didn't get any, then we should
```
vpapavas, line 535: In what scenario are we not getting any endOffsests even though we are supposed to?

ksqldb-engine/src/main/java/io/confluent/ksql/query/QueryExecutor.java
```diff
@@ -647,4 +671,107 @@ private static void updateListProperty(
...
+      final boolean alreadyDone =
+          topicPartition.isPresent() && donePartitions.contains(topicPartition.get());
+
+      // We ignore null values in streams as invalid data and drop the record.
+      // If we're already done with the current partition, we just drop the record.
+      if (record.value() != null && !alreadyDone) {
+        queue.acceptRow(null, record.value());
+      }
+
+      checkCompletion();
```
vpapavas, line 726: For every record added to the queue, we check the endOffsets across all partitions in `checkCompletion`. I am wondering if we can make it cheaper (might not be needed if this is already very cheap) by just checking for completion in the partition we are currently reading from, where the current record is read from

ksqldb-engine/src/main/java/io/confluent/ksql/query/QueryExecutor.java
```diff
@@ -647,4 +671,107 @@ private static void updateListProperty(
...
+      this.context = context;
+      if (endOffsets.isPresent()) {
+        // registering a punctuation here is the easiest way to check
+        // if we're complete even if we don't see all records.
+        // The in-line check in process() is faster,
+        // but some records may be filtered out before reaching us.
+        context.schedule(
+            Duration.ofMillis(100),
+            PunctuationType.WALL_CLOCK_TIME,
+            timestamp -> checkCompletion()
```
vpapavas, line 705: We check for completion both periodically and when reading records.  For my education, why is the periodic one needed as well?

ksqldb-engine/src/main/java/io/confluent/ksql/query/QueryExecutor.java
vvcephei: It's a bit weird, but if a partition is empty, then we just won't get an "end offset" for it at all. So, if _all_ the partitions are empty, then the whole map will be empty. In this case, it just means we're supposed to stop at the "end" of an empty topic. Aka, we should return zero results.

ksqldb-engine/src/main/java/io/confluent/ksql/query/QueryExecutor.java
vvcephei: That's a good observation, but it should actually be pretty cheap already, since this is only checking local variables. I'd be surprised if just checking the current record's partition is noticeably faster, but I'll go ahead and do it.

ksqldb-engine/src/main/java/io/confluent/ksql/query/QueryExecutor.java
vvcephei: You're right; I should add a comment :) 

The `process` call is only triggered when processing a record. If an upstream processor filters records out, then this processor (which is right at the end of the topology) might never get triggered and hence might never complete. I figured that out the hard way :/

ksqldb-rest-app/src/main/java/io/confluent/ksql/api/server/LoggingHandler.java
```diff
@@ -87,9 +88,10 @@ public void handle(final RoutingContext routingContext) {
       final String userAgent = Optional.ofNullable(
           routingContext.request().getHeader(HTTP_HEADER_USER_AGENT)).orElse("-");
       final String timestamp = Utils.formatRFC1123DateTime(clock.millis());
+      final SocketAddress socketAddress = routingContext.request().remoteAddress();
       final String message = String.format(
           "%s - %s [%s] \"%s %s %s\" %d %d \"-\" \"%s\" %d",
-          routingContext.request().remoteAddress().host(),
+          socketAddress == null ? "null" : socketAddress.host(),
```
vvcephei, line 94: Not sure why, but I got an NPE on this line a couple of times. This should prevent it.

vvcephei (commented):
Hey @vpapavas and @AlanConfluent ,

I'm ready for a real review pass now. I've left some breadcrumbs to help your review.

ksqldb-engine/src/main/java/io/confluent/ksql/engine/KsqlEngine.java
```diff
@@ -471,7 +422,7 @@ private TopicDescription getTopicDescription(final Admin admin, final String sou
     final ListOffsetsResult listOffsetsResult = admin.listOffsets(
         topicPartitions,
         new ListOffsetsOptions(
-            isolationLevel
+            IsolationLevel.READ_UNCOMMITTED
```
AlanConfluent, line 425: Why read uncommitted rather than the code from before?

ksqldb-engine/src/main/java/io/confluent/ksql/query/PullQueryQueue.java
```diff
@@ -82,6 +82,12 @@ public void setLimitHandler(final LimitHandler limitHandler) {
     this.limitHandler = limitHandler;
   }
 
+  @Override
+  public void setCompletionHandler(final CompletionHandler completionHandler) {
+    // not currently used in pull queries, although future refactoring might be able to
+    // take advantage of this mechanism.
```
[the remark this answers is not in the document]
AlanConfluent: I poked through the code and saw how we do this for pull queries at the moment, which is that we use the LimitHandler and just fire it, even if there's nothing in the queue.  It's effectively the same thing as this completion handler.  Can we make a followup task to consolidate these two?

ksqldb-engine/src/main/java/io/confluent/ksql/query/QueryBuilder.java
```diff
@@ -650,4 +671,131 @@ private static void updateListProperty(
...
+          context
+              .recordMetadata()
+              .map(m -> new TopicPartition(m.topic(), m.partition()));
+
+      final boolean alreadyDone =
+          topicPartition.isPresent() && donePartitions.contains(topicPartition.get());
+
+      // We ignore null values in streams as invalid data and drop the record.
+      // If we're already done with the current partition, we just drop the record.
+      if (record.value() != null && !alreadyDone) {
```
AlanConfluent, line 722: Do push queries not handle null data values?  For some reason I thought they did.

ksqldb-engine/src/main/java/io/confluent/ksql/query/QueryBuilder.java
```diff
@@ -650,4 +671,131 @@ private static void updateListProperty(
...
+        queue.acceptRow(null, record.value());
+      }
+
+      if (topicPartition.isPresent()) {
+        checkCompletion(topicPartition.get());
+      }
+    }
+
+    private void checkCompletion(final TopicPartition topicPartition) {
```
AlanConfluent, line 731: I found it just a bit confusing having 3 different versions of checkCompletion.  Maybe it makes sense to call this something like `checkCompletionPartition` the next `checkCompletionAllPartitions`?

If you had `checkCompletion()` call this method, you could inline the body of the third method here.  Is the idea that you want a consistent snapshot of currentPositions?  There is only one thread acting on everything, right?

ksqldb-engine/src/main/java/io/confluent/ksql/query/QueryBuilder.java
```diff
@@ -650,4 +671,131 @@ private static void updateListProperty(
...
+      // It might make sense to actually propose adding the concept of getting the
+      // "current position" in the ProcessorContext, but for now we just expose it
+      // by reflection. Obviously, this code is pretty brittle.
+      try {
+        if (context.getClass().equals(ProcessorContextImpl.class)) {
+          final Field streamTask;
+          streamTask = ProcessorContextImpl.class.getDeclaredField("streamTask");
+          streamTask.setAccessible(true);
+          final StreamTask task = (StreamTask) streamTask.get(context);
+          final Method committableOffsetsAndMetadata =
```
AlanConfluent, line 781: We should make an issue to expose this in streams so that this isn't required.

ksqldb-engine/src/main/java/io/confluent/ksql/query/TransientQueryQueue.java
```diff
@@ -49,39 +53,24 @@ public TransientQueryQueue(
...
-      @Override
-      public boolean shouldQueue() {
-        return parent.shouldQueue();
-      }
-
-      @Override
-      public void onQueued() {
-        parent.onQueued();
-        queuedCallback.run();
-      }
```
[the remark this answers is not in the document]
AlanConfluent: I agree.  I always found this parent, chaining to be a bit hard to understand.

ksqldb-engine/src/main/java/io/confluent/ksql/query/QueryBuilder.java
```diff
@@ -650,4 +671,131 @@ private static void updateListProperty(
     valueList.add(value);
     properties.put(key, valueList);
   }
+
+  private static class TransientQuerySinkProcessor implements
```
AlanConfluent, line 675: This is complex enough, we might want to pull it out into its own class.

ksqldb-engine/src/main/java/io/confluent/ksql/engine/KsqlEngine.java
```diff
@@ -364,15 +354,16 @@ public StreamPullQueryMetadata createStreamPullQuery(
             .putAll(statementOrig.getSessionConfig().getOverrides())
             .put(ConsumerConfig.AUTO_OFFSET_RESET_CONFIG, "earliest")
             .put(StreamsConfig.NUM_STREAM_THREADS_CONFIG, 1)
```
AlanConfluent, line 356: Is this important for threadsafety in your processor?  Might make sense to comment on that here if so.

ksqldb-engine/src/main/java/io/confluent/ksql/engine/KsqlEngine.java
vvcephei: I hard-coded the stream pull query app to be "at least once" (because it is pure side-effect, there is no point in EOS), which means that we will only need to get the end offsets in "read uncommitted" mode.

ksqldb-engine/src/main/java/io/confluent/ksql/query/QueryBuilder.java
vvcephei: Yeah, this code was here before, so I guess not.

ksqldb-engine/src/main/java/io/confluent/ksql/query/QueryBuilder.java
vvcephei: Roger that. I kind of phoned it in on naming these methods. I'll fix it.

It's handy to keep the third method so that we don't have to separately call `currentPositions` before invoking the single-partition version. Maybe you can take another look after I push my current batch of updates and see if you think it's ok.

ksqldb-engine/src/main/java/io/confluent/ksql/query/QueryBuilder.java
vvcephei: I wanted to sanity-check the overall algorithm before starting that process. Since it seems like you agree with the approach, I'll go ahead and propose it for AK now.

ksqldb-engine/src/main/java/io/confluent/ksql/query/QueryBuilder.java
vvcephei: Great idea.

ksqldb-engine/src/main/java/io/confluent/ksql/engine/KsqlEngine.java
vvcephei: No, I just think it'd be wasteful to spend more than one thread on this operation.

ksqldb-engine/src/main/java/io/confluent/ksql/engine/KsqlEngine.java
vvcephei: Ok, I added a comment to clarify this.

ksqldb-engine/src/main/java/io/confluent/ksql/engine/KsqlEngine.java
vvcephei: Turns out, I did have a comment explaining the full set of configs a few lines above, but I'd already forgotten to update it when I removed the commit interval, which isn't a good sign.

I moved the comments down to explain each config option independently, so this is hopefully clearer now.

ksqldb-api-client/src/test/java/io/confluent/ksql/api/client/integration/ClientIntegrationTest.java
```diff
@@ -675,119 +745,6 @@ public void shouldHandleErrorResponseFromInsertInto() {
     assertThat(e.getCause().getMessage(), containsString("Cannot insert into a table"));
   }
 
-  @Test
-  public void shouldStreamQueryWithProperties() throws Exception {
```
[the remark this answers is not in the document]
guozhangwang: SGTM.

ksqldb-engine/src/main/java/io/confluent/ksql/engine/KsqlEngine.java
```diff
@@ -364,15 +354,16 @@ public StreamPullQueryMetadata createStreamPullQuery(
...
+    final ImmutableMap<TopicPartition, Long> endOffsets =
+        getQueryInputEndOffsets(analysis, serviceContext.getAdminClient());
+
     final TransientQueryMetadata transientQueryMetadata = EngineExecutor
         .create(primaryContext, serviceContext, statement.getSessionConfig())
-        .executeTransientQuery(statement, excludeTombstones);
-
-    final ImmutableMap<TopicPartition, Long> endOffsets =
-        getQueryInputEndOffsets(analysis, statement, serviceContext.getAdminClient());
+        .executeTransientQuery(statement, excludeTombstones, Optional.of(endOffsets));
```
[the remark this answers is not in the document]
guozhangwang: This makes me wonder: are we categorizing `push` and `pull` queries both as part of `transient` queries now? In the past `push == transient`, and `pull` queries are another category. But now since we are executing `stream pull queries` as a transient query. Or are we consider the `stream pull queries` still as part of the `push == transient` queries but just expose them as a `pull` query to users?

ksqldb-engine/src/main/java/io/confluent/ksql/query/TransientQueryQueue.java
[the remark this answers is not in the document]
guozhangwang: +1

ksqldb-engine/src/main/java/io/confluent/ksql/query/QueryBuilder.java
guozhangwang: That is consistent with KS semantics, that null-values are considered invalid and always filtered out. In tables they are considered as tombstones of course, and we sort of letting the users be responsible of the difference between the two when converting in between.

ksqldb-engine/src/main/java/io/confluent/ksql/query/TransientQuerySinkProcessor.java
```diff
@@ -0,0 +1,170 @@
...
+    // It might make sense to actually propose adding the concept of getting the
+    // "current position" in the ProcessorContext, but for now we just expose it
+    // by reflection. Obviously, this code is pretty brittle.
+    try {
+      if (context.getClass().equals(ProcessorContextImpl.class)) {
+        final Field streamTask;
+        streamTask = ProcessorContextImpl.class.getDeclaredField("streamTask");
+        streamTask.setAccessible(true);
+        final StreamTask task = (StreamTask) streamTask.get(context);
+        final Method committableOffsetsAndMetadata =
```
guozhangwang, line 151: If the current position is referring to the source topic's position, could we get them via `ProcessorContext#recordMetadata()#offset()`? cc @AlanConfluent

guozhangwang (commented):
Made a pass as well, I really like the cleanup you did alongside this PR! 

One meta question I had is around why we need to split the phases of `stream pull queries` into `create` first, and then `wait`, instead as `execute` as in others. I think I just lack some critical context here :P Will sync with @vvcephei separately offline.

ksqldb-engine/src/main/java/io/confluent/ksql/query/TransientQuerySinkProcessor.java
vvcephei: I tried that first, only to sadly be reminded that the `endOffsets` returned by the admin client are not the "offsets of the last records in the partitions", but rather "the offset _after_ the last record in the partition" aka "the offset that the _next produced_ record will get".

The reason I cracked the safe to get at this method here is because it specifically computes that position for the purpose of committing offsets. I.e., we don't commit the offsets of the records we have processed, we commit the offsets of the _next records_ we need to process. It might be tempting to just use `recordMetadata#offset() + 1` for this, but the reality is that it's not so trivial, thanks for skips in the offset sequence due to compaction, "invisible" offsets for transaction markers, and stuff like that.

The method I'm exposing here specifically accounts for all of those effects:
https://github.com/apache/kafka/blob/trunk/streams/src/main/java/org/apache/kafka/streams/processor/internals/StreamTask.java#L435-L455

In addition to the above, we have another problem: upstream processors might filter out arbitrary records before they even reach this processor. Specifically, if you write a stream pull query that should only match one record, and there are more records after it in the topic, the query would never terminate, since the `process()` method would never see any offsets passed the one we're filtering for. That's why I also registered the punctuator, and the punctuator needs to be able to check "the current position" of the task as a whole, since it has no `recordMetadata` available to get the offset of.

vvcephei:
Thanks for taking a look, @guozhangwang ! I responded above to your question about the Streams state I'm peeking into. Please let me know if it still seems wrong to you.

About the create/wait separation: that separation is actually gone now. Specifically, this PR removes the "wait" method in favor of using the same mechanism to complete the query that we're currently using to implement the "limit" clause for push queries. That would have been a better way to do it from the start, but I didn't have the inspiration until now.

I considered renaming the "create" method to "execute" like the others, but I'm actually more inclined to rename them all to "create" for the simple reason that none of those methods actually execute anything. They all only create the query metadata and then some other component later calls `start()` to actually execute the query. But I decided to just leave it alone, since I'd like to continue iteratively refactoring this part of the code to simplify things.

guozhangwang (commented):
Looking at the function names in KsqlEngine, and I'm wondering if we could consider renaming some of them, since:

* `executeTransientQuery` only create the query, and then trigger start() in the publisher
* `executeTablePullQuery` may trigger start() if startImmediately is true (and in new APIs it is set to true indeed); if not it would trigger in the publisher
* `executeScalablePushQuery` is the same as executeTransientQuery

So maybe it’s better to consider renaming all of them as createXXX similar to `createStreamPullQuery` except `createAndMaybeExecuteTablePullQuery`? WDYT @vvcephei @AlanConfluent @vpapavas

AlanConfluent (approved):
Looks good

vvcephei:
Thanks for looking at that, @guozhangwang ! I agree with your proposal, but I'd like to tackle it in a later PR.

vvcephei:
Thanks for the review @AlanConfluent !

ksqldb-engine/src/main/java/io/confluent/ksql/query/TransientQuerySinkProcessor.java
guozhangwang: Thanks John, that makes sense!

ksqldb-engine/src/main/java/io/confluent/ksql/query/QueryBuilder.java
vvcephei: Here's a POC I put together for an AK feature, but I haven't taken the time to formally propose it: https://github.com/apache/kafka/pull/11336

PR #8126, opened 2021-09-14
feat: Add metrics and limiters for stream pull queries for HTTP1 endpoint

vpapavas:
### Description 
Add metrics and limiters (rate, bandwidth, concurrency) for stream pull queries. The limiters are counting both table and stream pull queries in the same limit. For example, if we allow 100 concurrent queries it must hold: table-pull + stream-pull <= 100

### Testing done 
Added new junit tests for metrics. Did manual tests for limiters. 

### Reviewer checklist

### DIFF
--- a/ksqldb-engine/src/main/java/io/confluent/ksql/physical/pull/PullPhysicalPlan.java
+++ b/ksqldb-engine/src/main/java/io/confluent/ksql/physical/pull/PullPhysicalPlan.java
@@ -164,7 +164,8 @@ public class PullPhysicalPlan {
     // Could be one or more keys
     KEY_LOOKUP,
     RANGE_SCAN,
-    TABLE_SCAN
+    TABLE_SCAN,
+    UNKNOWN
   }
 
   /**
@@ -173,7 +174,9 @@ public class PullPhysicalPlan {
    */
   public enum PullSourceType {
     NON_WINDOWED,
-    WINDOWED
+    WINDOWED,
+    NON_WINDOWED_STREAM,
+    WINDOWED_STREAM
   }
 
   /**
--- a/ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/resources/streaming/StreamedQueryResource.java
+++ b/ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/resources/streaming/StreamedQueryResource.java
@@ -385,8 +385,8 @@ public class StreamedQueryResource implements KsqlConfigurable {
           final AtomicReference<Optional<Decrementer>> optionalDecrementer =
               new AtomicReference<>(null);
           metricsCallbackHolder.setCallback(
-              pullMetricsInitializer.initializeStreamMetricsCallback(analysis, resultForMetrics,
-                                                             optionalDecrementer));
+              pullMetricsInitializer.initializeStreamMetricsCallback(
+                  analysis, resultForMetrics, optionalDecrementer));
 
           final SessionConfig sessionConfig = SessionConfig.of(ksqlConfig, configProperties);
           final ConfiguredStatement<Query> configured = ConfiguredStatement
--- a/ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/resources/streaming/StreamedQueryResource.java
+++ b/ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/resources/streaming/StreamedQueryResource.java
@@ -362,7 +362,6 @@ public class StreamedQueryResource implements KsqlConfigurable {
           final AtomicReference<PullQueryResult> resultForMetrics = new AtomicReference<>(null);
           metricsCallbackHolder.setCallback(
               pullMetricsInitializer.initializeTableMetricsCallback(resultForMetrics));
-//          metricsCallbackHolder.setCallback(initializeTablePullMetricsCallback(resultForMetrics));
 
           final SessionConfig sessionConfig = SessionConfig.of(ksqlConfig, configProperties);
           final ConfiguredStatement<Query> configured = ConfiguredStatement
@@ -388,8 +387,6 @@ public class StreamedQueryResource implements KsqlConfigurable {
           metricsCallbackHolder.setCallback(
               pullMetricsInitializer.initializeStreamMetricsCallback(analysis, resultForMetrics,
                                                              optionalDecrementer));
-//          metricsCallbackHolder.setCallback(initializeStreamPullMetricsCallback(analysis,
-//              resultForMetrics, optionalDecrementer));
 
           final SessionConfig sessionConfig = SessionConfig.of(ksqlConfig, configProperties);
           final ConfiguredStatement<Query> configured = ConfiguredStatement
@@ -561,7 +558,7 @@ public class StreamedQueryResource implements KsqlConfigurable {
     // Apply the same rate, bandwidth and concurrency limits as with table pull queries
     PullQueryExecutionUtil.checkRateLimit(rateLimiter);
     final Decrementer decrementer = concurrencyLimiter.increment();
-    pullBandRateLimiter.allow();
+    pullBandRateLimiter.allow(KsqlQueryType.PULL);
     optionalDecrementer.set(Optional.ofNullable(decrementer));
 
     final StreamPullQueryMetadata streamPullQueryMetadata = ksqlEngine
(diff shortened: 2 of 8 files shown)


ksqldb-engine/src/main/java/io/confluent/ksql/physical/pull/PullPhysicalPlan.java
```diff
@@ -164,7 +164,8 @@ public QueryId getQueryId() {
     // Could be one or more keys
     KEY_LOOKUP,
     RANGE_SCAN,
-    TABLE_SCAN
+    TABLE_SCAN,
+    UNKNOWN
```
vpapavas, line 168: We cannot know what the WHERE clause is for stream pull queries since it's not analyzed by the logical plan

ksqldb-rest-app/src/test/java/io/confluent/ksql/rest/integration/PullQueryMetricsFunctionalTest.java
```diff
@@ -0,0 +1,283 @@
...
+    @Override
+    public void close() {
+    }
+
+    @Override
+    public void configure(final Map<String, ?> configs) {
+    }
+  }
+
+  @Test
```
vpapavas, line 155: Unfortunately, I have to add all assertions in one test since I haven't managed to find a way for the metrics to get cleaned between tests. I tried even creating a new rest app for each test but even that didn't work.

ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/resources/streaming/StreamedQueryResource.java
```diff
@@ -571,7 +552,15 @@ private EndpointResponse handleStreamPullQuery(
...
-      final CompletableFuture<Void> connectionClosedFuture) {
+      final CompletableFuture<Void> connectionClosedFuture,
+      final AtomicReference<StreamPullQueryMetadata> resultForMetrics,
+      final AtomicReference<Optional<Decrementer>> optionalDecrementer) {
+
+    // Apply the same rate, bandwidth and concurrency limits as with table pull queries
+    PullQueryExecutionUtil.checkRateLimit(rateLimiter);
+    final Decrementer decrementer = concurrencyLimiter.increment();
+    pullBandRateLimiter.allow(KsqlQueryType.PULL);
+    optionalDecrementer.set(Optional.ofNullable(decrementer));
```
vvcephei, line 563: Looking at the ConcurrencyLimiter implementation, I don't believe `decrementer` actually could ever be null.

More generally, it doesn't look like we really need to have an `AtomicReference<Optional<Decrementer>>`. It seems like just an `AtomicReference<Decrementer>` would do just fine.

The reason is that the AtomicReferenced starts off initialized to null, which means the referencer needs to check for null anyway, so the Optional really isn't buying us anything here.

ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/resources/streaming/StreamedQueryResource.java
```diff
@@ -625,6 +615,129 @@ public Boolean get() {
...
+    private PullMetricsInitializer(
+        final Optional<PullQueryExecutorMetrics> pullQueryMetrics,
+        final SlidingWindowRateLimiter pullBandRateLimiter) {
+
+      this.pullQueryMetrics = Objects.requireNonNull(pullQueryMetrics, "pullQueryMetrics");
+      this.pullBandRateLimiter = Objects.requireNonNull(pullBandRateLimiter, "pullBandRateLimiter");
+    }
+
+    private MetricsCallback initializeTableMetricsCallback(
+        final AtomicReference<PullQueryResult> resultForMetrics) {
```
vvcephei, line 632: Thanks for factoring these out from the method above. I'm wondering if we can skip this extra class definition, though. It seems like just having a static method would do just fine and would save on both lines of code and on complexity for us to understand the code.

```suggestion
  private static MetricsCallback initializeTableMetricsCallback(
      final Optional<PullQueryExecutorMetrics> pullQueryMetrics,
      final SlidingWindowRateLimiter pullBandRateLimiter
      final AtomicReference<PullQueryResult> resultForMetrics) {
```

(for example... I'd propose to make the other method also just a static method by adding these two arguments)

ksqldb-rest-app/src/test/java/io/confluent/ksql/rest/integration/PullQueryMetricsFunctionalTest.java
vvcephei: I think this is fine for now. A test is better than no test!

We can always iterate on the metric testing strategy in the future.

ksqldb-rest-app/src/test/java/io/confluent/ksql/rest/integration/PullQueryMetricsFunctionalTest.java
```diff
@@ -0,0 +1,283 @@
...
+
+    RestIntegrationTestUtil.makeQueryRequest(
+        REST_APP,
+        "SELECT * from " + PAGE_VIEW_STREAM + " WHERE PAGEID='" + A_STREAM_KEY + "';",
+        Optional.empty());
+
+    for (MetricName metric: TestMetricsReporter.METRICS.keySet()) {
+      if (metric.name().startsWith("pull-query")) {
+        if ((Double)TestMetricsReporter.METRICS.get(metric).metricValue() > 0) {
+          System.out.println(metric + " , " + (Double)TestMetricsReporter.METRICS.get(metric).metricValue());
```
vvcephei, line 254: This is still a draft PR, but just to make sure it doesn't slip through... we probably don't want this println here.

ksqldb-rest-app/src/test/java/io/confluent/ksql/rest/server/resources/streaming/StreamedQueryResourceTest.java
```diff
@@ -471,6 +523,29 @@ public void shouldReachConcurrentLimit() {
...
+            securityContext,
+            new KsqlRequest(PULL_QUERY_STRING, Collections.emptyMap(), Collections.emptyMap(), null),
+            new CompletableFuture<>(),
+            Optional.empty(),
+            KsqlMediaType.LATEST_FORMAT,
+            new MetricsCallbackHolder(),
+            context);
+
+    // Then:
+    assertThat(response.getStatus(), is(500));
```
vvcephei, line 545: It looks like we have already been returning a 500 for limit violations, so this isn't a problem with your PR specifically, but just to point it out, 500 (server error) is the wrong response code. It should be a 4xx (bad request) type error. 429 is probably the best error, though it's not a perfect match for concurrency violations: https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/429

vvcephei (commented):
Thanks for this @vpapavas !

I had a minute, and it seems like this is getting close, so I went ahead and reviewed it even though it's in draft. Please disregard anything that's not applicable (because this is only a draft PR right now).

ksqldb-rest-app/src/test/java/io/confluent/ksql/rest/integration/PullQueryMetricsFunctionalTest.java
vpapavas: Ooops good catch, thank you!

ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/resources/streaming/StreamedQueryResource.java
```diff
@@ -571,7 +552,15 @@ private EndpointResponse handleStreamPullQuery(
...
       final ConfiguredStatement<Query> configured,
-      final CompletableFuture<Void> connectionClosedFuture) {
+      final CompletableFuture<Void> connectionClosedFuture,
+      final AtomicReference<StreamPullQueryMetadata> resultForMetrics,
+      final AtomicReference<Optional<Decrementer>> optionalDecrementer) {
+
+    // Apply the same rate, bandwidth and concurrency limits as with table pull queries
+    PullQueryExecutionUtil.checkRateLimit(rateLimiter);
+    final Decrementer decrementer = concurrencyLimiter.increment();
+    pullBandRateLimiter.allow(KsqlQueryType.PULL);
```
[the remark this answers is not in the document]
vpapavas: I followed the order of table pull queries but I guess for stream pull queries the bandwidth limit is more possible to be hit first. I will change the order

ksqldb-rest-app/src/test/java/io/confluent/ksql/rest/server/resources/streaming/StreamedQueryResourceTest.java
vpapavas: Hmm, I am not sure where the error codes get assigned in the code base. I will open a ticket to fix this as part of the general error code fixing initiative we will do

vvcephei (approved):
Thanks, @vpapavas !

Of course we have to figure out why that test is failing, but from a code review perspective, I'm +1.

vpapavas:
I adopted the extra checks you added in your example @vvcephei . I also ran the test 10 times locally and it didn't fail so I am going ahead and merging

PR #8132, opened 2021-09-15
fix: Ensures background timer completes for scalable push queries

AlanConfluent:
### Description 
The Vertx timer which is used to ensure connections with other nodes in the cluster correctly checks if the handle has been closed, but the condition for isClosed only considered if there was an error, not if it had been manually closed as when the connection is closed.

### Testing done 
Called it in unit tests and manually verified.

### Reviewer checklist


nateab (approved):
LGTM!

PR #8121, opened 2021-09-12
chore: disallow insert values for ROWPARTITION and ROWOFFSET

Sullivan-Patrick:
### Description 
This change is part of [KLIP-50](https://github.com/confluentinc/ksql/blob/master/design-proposals/klip-50-partition-and-offset-in-ksqldb.md). It disallows the use of `INSERT VALUES` into `ROWPARTITION` and `ROWOFFSET` pseudocolumns. The implementation allows for user columns with these names to continue to allow for inserts (this situation can occur iff a user created a stream or table which contained a user column named `ROWPARTITION`/`ROWOFFSET` before upgrading to a server version with the pseudocolumns enabled),

To make this easier to write and test, it's based on top of the still-unmerged branch associated with https://github.com/confluentinc/ksql/pull/8120. This change only concerns commits after 2ae058a, inclusive.

### Testing done 
New QTTs to verify behavior with and without feature enabled.

### Reviewer checklist
- [ ] Ensure docs are updated if necessary. (eg. if a user visible feature is being added or changed).
- [ ] Ensure relevant issues are linked (description should include text like "Fixes #<issue number>")

PR #8135, opened 2021-09-15
chore: disallow pull queries with pseudocolumns named ROWPARTITION or ROWOFFSET

Sullivan-Patrick:
### Description 
As a part of KLIP-50, for the reasons described [here](https://github.com/confluentinc/ksql/blob/master/design-proposals/klip-50-partition-and-offset-in-ksqldb.md#pull-queries), `ROWPARTITION` and `ROWOFFSET` will not be allowed as part of pull queries. (When reading the KLIP it is worth noting that we do, [in certain cases],(https://github.com/confluentinc/ksql/pull/7990) materialize partition and offset information now, but the points still remain).

In order to do this, a new `ExpressionVisitor` which collects column names was added. This traverses the expression tree of a given clause, then returns a set of all column names within that clause. This will be useful later when routing queries that contain the new pseudocolumns away from scalable push queries.

### Testing done 
Unit tests testing the new rules in `PullQueryValidator`, rQTTs that test the new functionality.

### Reviewer checklist


ksqldb-common/src/main/java/io/confluent/ksql/schema/ksql/SystemColumns.java
```diff
@@ -134,6 +140,26 @@ public static boolean isSystemColumn(final ColumnName columnName, final KsqlConf
     return isSystemColumn(columnName, getPseudoColumnVersionFromConfig(ksqlConfig));
   }
 
+  public static int getPseudoColumnVersionFromConfig(final KsqlConfig ksqlConfig) {
```
Sullivan-Patrick, line 143: Decided to make this public. WDYT? I don't see a huge issue in exposing it, especially given it's temporary, and I can clean up some instances of repeated logic from this.

ksqldb-common/src/main/java/io/confluent/ksql/schema/ksql/SystemColumns.java
```diff
@@ -134,6 +140,26 @@ public static boolean isSystemColumn(final ColumnName columnName, final KsqlConf
...
+  public static boolean disallowedForPullQueries(
+      final ColumnName columnName,
+      final int pseudoColumnVersion
+  ) {
+    // option 2: make a set that is the union of DISALLOWED and systemcols,
+    // return if the set contains columnName
+    return isSystemColumn(columnName, pseudoColumnVersion)
+        && DISALLOWED_FOR_PULL_QUERIES.contains(columnName);
+  }
+
```
Sullivan-Patrick: Moved these above private methods.

ksqldb-engine/src/main/java/io/confluent/ksql/analyzer/ColumnReferenceExtractor.java
```diff
@@ -0,0 +1,61 @@
...
+import io.confluent.ksql.execution.expression.tree.QualifiedColumnReferenceExp;
+import io.confluent.ksql.execution.expression.tree.UnqualifiedColumnReferenceExp;
+import io.confluent.ksql.execution.expression.tree.VisitParentExpressionVisitor;
+import io.confluent.ksql.name.ColumnName;
+import java.util.HashSet;
+import java.util.Optional;
+import java.util.Set;
+
+/**
+ * The purpose of this class is to extract ColumnNames from an {@link Expression}.
```
Sullivan-Patrick, line 15: I want to make sure I'm implementing at least every necessary method to handle pull queries and SPQs, but I'm not very familiar with this part of the codebase and what is/isn't compatible. I'd appreciate if you could take a look at `ExpressionVisitor` and let me know if there are any other methods I'd need to implement , assuming that'd be trivial for you @vcrfxia

ksqldb-engine/src/main/java/io/confluent/ksql/analyzer/PullQueryValidator.java
```diff
@@ -62,6 +66,14 @@
       Rule.of(
           analysis -> !analysis.getRefinementInfo().isPresent(),
           "Pull queries don't support EMIT clauses."
+      ),
+      Rule.of(
+          analysis -> !PullQueryValidator.disallowedColumnNameInSelectClause(analysis),
+          "Pull queries don't support ROWPARTITION or ROWOFFSET in select clauses"
```
Sullivan-Patrick, line 72: Because the string has no knowledge of the predicate's result, I can't really customize the message to include the exact column that is offending- or rather, I don't see a clean way of including it. I see a small extensibility and message clarity issue, probably worth fixing if there is a clean way, probably not if not. WDYT?

ksqldb-common/src/main/java/io/confluent/ksql/schema/ksql/SystemColumns.java
```diff
@@ -134,6 +140,26 @@ public static boolean isSystemColumn(final ColumnName columnName, final KsqlConf
...
+  public static boolean mustBeMaterializedForTableJoins(final ColumnName columnName) {
+    return MUST_BE_MATERIALIZED_FOR_TABLE_JOINS.contains(columnName);
+  }
+
+  public static boolean disallowedForPullQueries(
+      final ColumnName columnName,
+      final int pseudoColumnVersion
+  ) {
+    // option 2: make a set that is the union of DISALLOWED and systemcols,
```
Sullivan-Patrick, line 157: while it's a draft PR I left these in (also a todo later in the file). As commented I originally used the intersection (not union, sorry) of `DISALLOWED_FOR_PULL_QUERIES` and `getSystemColumnNames(pseudoColumnVersion)`, which ultimately felt less readable and is less performant if I create the set each time. I prefer the way written, but I figured I'd mention it here.

ksqldb-common/src/main/java/io/confluent/ksql/schema/ksql/SystemColumns.java
```diff
@@ -134,6 +140,24 @@ public static boolean isSystemColumn(final ColumnName columnName, final KsqlConf
...
+  }
+
+  public static boolean disallowedForPullQueries(
+      final ColumnName columnName,
+      final int pseudoColumnVersion
+  ) {
+    return isSystemColumn(columnName, pseudoColumnVersion)
+        && DISALLOWED_FOR_PULL_QUERIES.contains(columnName);
+  }
+
```
Sullivan-Patrick: Just moved above private methods.

ksqldb-engine/src/main/java/io/confluent/ksql/analyzer/ColumnReferenceExtractor.java
```diff
@@ -0,0 +1,61 @@
...
+ * This currently successfully traverses the expression tree only if
+ * non-leaf nodes are composed exclusively of ComparisonExpressions and LogicalBinaryExpressions,
+ * as this was the behavior necessary to disable pull queries and scalable push queries
+ * with ROWPARTITION and ROWOFFSET pseudocolumns (as these queries do not work with other clauses).
+ */
+public class ColumnReferenceExtractor extends
+    VisitParentExpressionVisitor<Optional<Expression>, Void> {
```
vcrfxia, line 22: Extend `TraversalExpressionVisitor` instead of `VisitParentExpressionVisitor`. Then you won't need to implement methods like `visitComparisonExpression()`, and the functionality of this visitor won't be limited to just comparisons and logical binary expressions. (It's also not true that those are the only two types of expressions we need for pull queries and scalable push queries -- for example, there could be other (arbitrary) types of expressions in the select statements).

vcrfxia: BTW, while looking at the code just now I saw that https://github.com/confluentinc/ksql/blob/master/ksqldb-execution/src/main/java/io/confluent/ksql/execution/util/ColumnExtractor.java already exists. I think you can just use that directly.

ksqldb-common/src/main/java/io/confluent/ksql/schema/ksql/SystemColumns.java
```diff
@@ -81,6 +81,12 @@
           ROWOFFSET_NAME
       );
 
+  private static final Set<ColumnName> DISALLOWED_FOR_PULL_QUERIES =
```
vcrfxia, line 84: This is now a fourth thing that anyone adding a new pseudocolumn needs to be reason about. Can we please make the changes described in https://github.com/confluentinc/ksql/pull/8121#discussion_r707600810?

Sullivan-Patrick: Sorry about that, haven't read that code review yet. I'll make a `PseudoColumn` class.

ksqldb-engine/src/test/java/io/confluent/ksql/analyzer/PullQueryValidatorTest.java
```diff
@@ -15,33 +15,42 @@
...
 import org.junit.runner.RunWith;
 import org.mockito.Mock;
+import org.mockito.MockedStatic;
 import org.mockito.junit.MockitoJUnitRunner;
 
 @RunWith(MockitoJUnitRunner.class)
 public class PullQueryValidatorTest {
 
   private static final Expression AN_EXPRESSION = mock(Expression.class);
+  private static final ColumnName A_BAD_COLUMN_NAME = ROWPARTITION_NAME;
```
vcrfxia, line 53: Let's remove this variable and use `ROWPARTITION_NAME` inline directly. I think the name `A_BAD_COLUMN_NAME` is actually more confusing since it's not clear what "bad" means.

vcrfxia (commented):
Thanks @Sullivan-Patrick . Can we also add integration test coverage for this PR, in addition to the unit tests?

ksqldb-engine/src/main/java/io/confluent/ksql/analyzer/PullQueryValidator.java
```diff
@@ -62,6 +71,14 @@
       Rule.of(
           analysis -> !analysis.getRefinementInfo().isPresent(),
           "Pull queries don't support EMIT clauses."
+      ),
+      Rule.of(
+          analysis -> !disallowedColumnNameInSelectClause(analysis),
+          "Pull queries don't support ROWPARTITION or ROWOFFSET in SELECT clauses."
```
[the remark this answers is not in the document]
Sullivan-Patrick: I'd be interested in hearing how you think this is best done, I don't see a super obvious/clean way of making the change to support the error message while evaluating the predicate. I'll make a note for our next sync.

ksqldb-engine/src/main/java/io/confluent/ksql/analyzer/PullQueryValidator.java
```diff
@@ -91,29 +89,47 @@ public void validate(final Analysis analysis) {
...
         .stream()
         .filter(col -> col instanceof SingleColumn) //filter out select *
         .map(SingleColumn.class::cast)
         .map(SingleColumn::getExpression)
         .map(ColumnExtractor::extractColumns)
         .flatMap(Collection::stream)
         .map(ColumnReferenceExp::getColumnName)
-        .anyMatch(name -> disallowedInPullQueries(name, analysis.getKsqlConfig()));
+        .filter(name -> disallowedInPullQueries(name, analysis.getKsqlConfig()))
+        .findFirst();
```
vcrfxia, line 102: Instead of returning just the first unsupported column, it'd be more helpful to the user to return all unsupported columns.

ksqldb-engine/src/main/java/io/confluent/ksql/analyzer/PullQueryValidator.java
```diff
@@ -126,23 +142,31 @@ private static boolean disallowedInPullQueries(
...
+
     private Rule(final Predicate<Analysis> condition, final String failureMsg) {
-      this.condition = Objects.requireNonNull(condition, "condition");
-      this.failureMsg = Objects.requireNonNull(failureMsg, "failureMsg");
+      this.potentialErrorMessage = (analysis) ->
+          !condition.test(analysis) ? Optional.of(failureMsg) : Optional.empty();
+    }
+
+    private Rule(final Function<Analysis, Optional<String>> function) {
+      this.potentialErrorMessage = function;
```
vcrfxia, line 161: null check.

vcrfxia (commented):
Looking good! Minor suggestions inline.

config/ksql-server.properties
```diff
@@ -80,3 +80,4 @@ compression.type=snappy
 # Uncomment and complete the following to enable KSQL's integration to the Confluent Schema Registry:
 # ksql.schema.registry.url=http://localhost:8081
 
+ksql.rowpartition.rowoffset.enabled = true
```
vcrfxia, line 83: This does not look like it's meant to be checked in.

ksqldb-engine/src/main/java/io/confluent/ksql/analyzer/PullQueryValidator.java
```diff
@@ -74,23 +92,84 @@ public void validate(final Analysis analysis) {
...
+      Function<Analysis, Optional<String>> potentialErrorMessageGenerator = (analysis) ->
+          !condition.test(analysis) ? Optional.of(failureMsg) : Optional.empty();
+
+      return new Rule(potentialErrorMessageGenerator);
     }
 
-    private Rule(final Predicate<Analysis> condition, final String failureMsg) {
-      this.condition = Objects.requireNonNull(condition, "condition");
-      this.failureMsg = Objects.requireNonNull(failureMsg, "failureMsg");
+    private static Rule of(final Function<Analysis, Optional<String>> function) {
```
vcrfxia, line 160:
```suggestion
    private static Rule of(final Function<Analysis, Optional<String>> potentialErrorMessageGenerator) {
```

ksqldb-engine/src/main/java/io/confluent/ksql/analyzer/PullQueryValidator.java
```diff
@@ -74,23 +92,84 @@ public void validate(final Analysis analysis) {
...
+      return new Rule(function);
+    }
+
+    private Rule(final Function<Analysis, Optional<String>> function) {
+      this.potentialErrorMessageGenerator = Objects.requireNonNull(function);
```
vcrfxia, line 165: Use two-arg version of `requireNonNull()`

vcrfxia (commented):
Just about ready to merge! A few minor things inline.

config/ksql-server.properties
Sullivan-Patrick: No idea how I missed this and a `FIXME` comment in a previous commit... I definitely got confused and forgot to review this PR before pushing. Sorry for the basic errors recently.

ksqldb-engine/src/main/java/io/confluent/ksql/analyzer/PullQueryValidator.java
Sullivan-Patrick: Out of curiosity, why don't we have this as a checkstyle rule? Are there exceptions to the rule?

vcrfxia: I'm not sure I'd call this a style thing, more a code convention. It's not the end of the world if we slip up, but I think it's nice to be consistent throughout the codebase.

ksqldb-engine/src/test/java/io/confluent/ksql/analyzer/PullQueryValidatorTest.java
```diff
@@ -51,13 +59,22 @@
...
+  private ColumnReferenceExp columnReferenceExp;
+  @Mock
+  private ColumnReferenceExp columnReferenceExp2;
 
   private QueryValidator validator;
 
   @Before
   public void setUp() {
     validator = new PullQueryValidator();
-    when(analysis.getRefinementInfo()).thenReturn(Optional.of(RefinementInfo.of(OutputRefinement.FINAL)));
```
vcrfxia, line 60: Why is this being changed?

ksqldb-functional-tests/src/test/resources/rest-query-validation-tests/pull-queries-against-cst-materialized.json
```diff
@@ -68,6 +68,21 @@
           {"row":{"columns":[2,"a2"]}}
         ]}
       ]
+    },
+    {
+      "name": "disallowed column in pull query select clause- ROWPARTITION",
```
vcrfxia, line 73: nit: can we have a consistent naming scheme for all the newly added RQTTs scattered across the different pull query RQTT files? For example, "should not allow pull query with ROWPARTITION in SELECT clause" for this one, and similarly for the others below.

vcrfxia (commented):
Thanks @Sullivan-Patrick -- LGTM besides the comments inline, and the failing build. Ready to merge once addressed!

ksqldb-engine/src/test/java/io/confluent/ksql/analyzer/PullQueryValidatorTest.java
Sullivan-Patrick: Otherwise the test fails with `Pull queries don't support EMIT clauses` before the condition can reach the check I added. Previously this wasn't a problem as `PullQueryValidatorTest` only checks for failing conditions, and `!analysis.getRefinementInfo().isPresent()` was the last check.

vcrfxia: OK. IIUC you're saying that the fact that this line was present was a bug since the presence of this line causes all queries to fail validation. The bug was previously hidden since all the other expected failures happened before this one, but that's no longer true. Removing it (as you've done) sounds good.

vcrfxia (approved):

ksqldb-engine/src/test/java/io/confluent/ksql/analyzer/PullQueryValidatorTest.java
Sullivan-Patrick: We seem to be on the same page- I'll keep it removed then.

ksqldb-common/src/main/java/io/confluent/ksql/schema/ksql/LogicalSchema.java
```diff
@@ -365,30 +366,35 @@ private LogicalSchema rebuildWithPseudoAndKeyColsInValue(
 
     int valueIndex = nonPseudoAndKeyCols.size();
 
-    final List<Column> pseudoColumns = new ArrayList<>();
+    final List<Pair<ColumnName, SqlType>> pseudoColumns = new ArrayList<>();
```
Sullivan-Patrick, line 369: I think the name isn't bad since they still represent pseudocolumns.

ksqldb-common/src/main/java/io/confluent/ksql/schema/ksql/LogicalSchema.java
```diff
@@ -365,30 +366,35 @@ private LogicalSchema rebuildWithPseudoAndKeyColsInValue(
...
-    for (Column column : toAddToBuilder) {
-      final Column columnWithIndex = Column.of(column.name(), column.type(), column.namespace(), valueIndex++);
-      builder.add(columnWithIndex);
+    final List<Column> toAddToBuilder = new ArrayList<>();
+
+    //if query is pull or scalable push, need to check if column is disallowed
+    if(isPullOrScalablePushQuery) {
+      for (Pair<ColumnName, SqlType> pair : pseudoColumns) {
+        if(!SystemColumns.isDisallowedInPullOrScalablePushQueries(pair.left, pseudoColumnVersion)) {
+          toAddToBuilder.add(Column.of(pair.left, pair.right, VALUE, valueIndex++));
```
vcrfxia, line 387: Rather than adding columns to `toAddToBuilder` and then immediately adding them to the builder, let's just add to the builder directly here. You can add an "else" branch to the `if(isPullOrScalablePushQuery)` that just adds all the pseudocolumns.

ksqldb-common/src/main/java/io/confluent/ksql/schema/ksql/LogicalSchema.java
```diff
@@ -343,7 +346,10 @@ public String toString(final FormatOptions formatOptions) {
    * @return the LogicalSchema created, with the corresponding pseudo and key columns included
    */
   private LogicalSchema rebuildWithPseudoAndKeyColsInValue(
-      final boolean windowedKey, final int pseudoColumnVersion) {
+      final boolean windowedKey,
+      final int pseudoColumnVersion,
+      final boolean isPullOrScalablePushQuery
```
vcrfxia, line 351: nit: 
```suggestion
      final boolean forPullOrScalablePushQuery
```
to match the name in `withPseudoAndKeyColsInValue()` above.

vcrfxia (approved):
Thanks @Sullivan-Patrick -- LGTM minus the minor comments inline and the failing build. Let's merge once the build is green!

PR #8137, opened 2021-09-15
refactor: rename QueryExecutor to QueryBuilder

spena:
### Description 
For so long, I've been getting confused about the real job of the `QueryExecutor`. The class does not executes anything, but only builds transient and persistent queries metadata. So why not refactoring to avoid more confusion?

### Testing done 

### Reviewer checklist


wcarlson5 (approved):
I strongly support this

PR #8141, opened 2021-09-16
add logical cluster id to observability metrics

lct45:
### Description 
We need to pass the `customTags` from the `ksqlConfig` to both the saturation and storage metrics to make sure we're picking up all the necessary tags. Right now, we pass the logical cluster as a tag in through this config and we need that to query through the metrics API

### Testing done 
will test on devel shortly

### Reviewer checklist

### DIFF
--- a/ksqldb-engine/src/main/java/io/confluent/ksql/internal/StorageUtilizationMetricsReporter.java
+++ b/ksqldb-engine/src/main/java/io/confluent/ksql/internal/StorageUtilizationMetricsReporter.java
@@ -52,6 +52,7 @@ public class StorageUtilizationMetricsReporter implements MetricsReporter {
   private final Map<String, Map<String, TaskStorageMetric>> metricsSeen;
   private final Metrics metricRegistry;
   static AtomicBoolean registeredNodeMetrics = new AtomicBoolean(false);
+  private static Map<String, String> customTags = new HashMap<>();
 
   public StorageUtilizationMetricsReporter() {
     this(MetricCollectors.getMetrics());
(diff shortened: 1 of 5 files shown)


rodesai (approved):
LGTM

ksqldb-engine/src/main/java/io/confluent/ksql/internal/StorageUtilizationMetricsReporter.java
```diff
@@ -162,14 +167,17 @@ private synchronized void handleNewSstFilesSizeMetric(
       final String taskId,
       final String queryId
   ) {
+    final Map<String, String> queryMetricTags = getQueryTags(queryId);
+    final Map<String, String> taskMetricTags = new HashMap<>(queryMetricTags);
```
rodesai, line 171: move to a function like:
```
final Map<String, String> taskMetricTags = getTaskMetricTags(queryMetricTgs, taskId);
```
Then the function can return an immutable map

ksqldb-engine/src/main/java/io/confluent/ksql/internal/StorageUtilizationMetricsReporter.java
```diff
@@ -245,6 +253,17 @@ private String getQueryId(final KafkaMetric metric) {
       throw new KsqlException("Missing query ID when reporting utilization metrics");
     }
   }
+  
+  private Map<String, String> getQueryTags(final String queryId) {
+    final Map<String, String> queryMetricTags = new HashMap<>(customTags);
+    queryMetricTags.put("query-id", queryId);
+    return queryMetricTags;
```
rodesai, line 260: return an immutable copy

ksqldb-engine/src/main/java/io/confluent/ksql/utilization/PersistentQuerySaturationMetrics.java
```diff
@@ -55,6 +54,7 @@
   private static final String STREAMS_THREAD_METRICS_GROUP = "stream-thread-metrics";
   private static final String THREAD_ID = "thread-id";
   private static final String QUERY_ID = "query-id";
+  private static Map<String, String> customTags = new HashMap<>();
```
rodesai, line 57: this can be final

lct45: I'm not sure it can, we set it to have the passed in tags in `setTags` and `configureShared`

PR #8142, opened 2021-09-16
refactor: clean up push query publisher

vvcephei:
Clean up the Websocket PushQueryPublisher to remove a bunch of
unnecessary fields and code paths.

### Reviewer checklist

### DIFF
--- a/ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/resources/streaming/PushQueryPublisher.java
+++ b/ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/resources/streaming/PushQueryPublisher.java
@@ -53,35 +53,33 @@ final class PushQueryPublisher implements Flow.Publisher<Collection<StreamedRow>
 
   private static final Logger log = LoggerFactory.getLogger(PushQueryPublisher.class);
 
-  private final KsqlEngine ksqlEngine;
-  private final ServiceContext serviceContext;
-  private final ConfiguredStatement<Query> query;
-  private final Optional<LocalCommands> localCommands;
   private final ListeningScheduledExecutorService exec;
-  private final PushRouting pushRouting;
-  private final boolean isScalablePush;
-  private final Context context;
-  private final SlidingWindowRateLimiter scalablePushBandRateLimiter;
+  private final PushQueryMetadata queryMetadata;
 
   private PushQueryPublisher(
+      final ListeningScheduledExecutorService exec,
+      final PushQueryMetadata queryMetadata
+  ) {
+    this.exec = exec;
+    this.queryMetadata = queryMetadata;
+  }
+
+  public static PushQueryPublisher createPushQueryPublisher(
       final KsqlEngine ksqlEngine,
       final ServiceContext serviceContext,
       final ListeningScheduledExecutorService exec,
       final ConfiguredStatement<Query> query,
       final Optional<LocalCommands> localCommands
   ) {
-    this.ksqlEngine = requireNonNull(ksqlEngine, "ksqlEngine");
-    this.serviceContext = requireNonNull(serviceContext, "serviceContext");
-    this.exec = requireNonNull(exec, "exec");
-    this.query = requireNonNull(query, "query");
-    this.localCommands = requireNonNull(localCommands, "localCommands");
-    this.pushRouting = null;
-    this.isScalablePush = false;
-    this.context = null;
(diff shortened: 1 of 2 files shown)


ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/resources/streaming/PushQueryPublisher.java
```diff
@@ -53,67 +53,33 @@
...
+    this.queryMetadata = queryMetadata;
   }
 
-  public static PushQueryPublisher createPublisher(
+  public static PushQueryPublisher createPushQueryPublisher(
       final KsqlEngine ksqlEngine,
       final ServiceContext serviceContext,
       final ListeningScheduledExecutorService exec,
       final ConfiguredStatement<Query> query,
       final Optional<LocalCommands> localCommands
```
vvcephei, line 72: High level description of this refactor:

We were previously capturing all these arguments as instance fields, and an overlapping set of arguments for the scalable push query publisher before, and then branching later on in `subscribe` to use those fields to create a PushQueryMetadata of the appropriate type.

But after that branch, the only thing we needed was the query metadata and the executor service. So this PR cuts out the middle man: we just go ahead and create that metadata up front in the static factor methods so that we neither need that big list of fields nor the branched logic later on.

ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/resources/streaming/PushQueryPublisher.java
```diff
@@ -123,45 +89,26 @@ public static PushQueryPublisher createScalablePublisher(
...
+        query.getSessionConfig().getConfig(false),
+        query.getSessionConfig().getOverrides());
 
-      final PushQueryConfigPlannerOptions plannerOptions = new PushQueryConfigPlannerOptions(
-              query.getSessionConfig().getConfig(false),
-              query.getSessionConfig().getOverrides());
+    scalablePushBandRateLimiter.allow(KsqlQueryType.PUSH);
 
-      scalablePushBandRateLimiter.allow(KsqlQueryType.PUSH);
+    final ScalablePushQueryMetadata pushQueryMetadata = ksqlEngine
```
AlanConfluent, line 102: My only comment is that we're actually kicking it off before the subscribe call.  I'm not sure what the implications of that are, if any.  Here is where the inspiration for this interface is defined: https://docs.oracle.com/javase/9/docs/api/java/util/concurrent/Flow.Publisher.html#subscribe-java.util.concurrent.Flow.Subscriber-

In theory, we're not meant to unleash all of the supply until subscribe is called.  After `subscribe` is called on the publisher, `onSubscribe()` is called on the subscriber and `request(1)` is made on the subscription.  Each time it requests 1 unit of data, it returns a `Collection<StreamedRow>` of all the current entries of the queue. 

 All of this is still true except subscribe no longer kicks off the process.  I'm fine with the refactoring, but I think it might be a small breach of the API.

AlanConfluent (approved):

ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/resources/streaming/PushQueryPublisher.java
vvcephei: Huh, I thought the query wouldn't actually get kicked off until we call `start()` (which still happens in the subscribe call). Regardless of the next question I'll ask, I'd really like clarity on when the query actually starts running.

Ok, next question: I get the theory of this whole publisher/subscriber pattern, but I couldn't quite wrap my mind around whether there's any point to it here, since we call `subscribe` _immediately_ on the returned object. Am I missing something here?

guozhangwang: I also think that the query is not "deployed" until the `start()` call, but please correct me if I have been wrong all these days..

For the next question, I think in our case there's little difference since `subscribe` is always triggered right after construction. In general that may not be the case, and hence it may be better deferring the construction of the actual resources until the first `subscribe` call is triggered?

vvcephei:
Since the review was a +1, and it looks like the questions are more on the informational side. I'll go ahead and merge so we can base the next set of PRs on `master`. I'll keep an eye on the thread for any responses and send follow-up patches as needed.

PR #8144, opened 2021-09-16
feat: add scalable push query metrics

nateab:
### Description 
Add some metrics on scalable push queries such as:
- Count of local/remote requests
- Count/rate of errors
- Min, max, and avg connection duration
- Request/response size
- Rows returned/processed  
- status code of the request - 2XX, 3XX, 4XX, 5XX

These are then further broken down by source and routing node type. 
Adds additional pull query metrics:

For source, we current support:

- Windowed
- Non Windowed

For routing node type, we currently support:

- source node (the initial node receiving the request)
- remote node (any node receiving a forwarded request)

### Testing done 
Unit tests added 

### Reviewer checklist


ksqldb-engine/src/main/java/io/confluent/ksql/internal/ScalablePushQueryExecutorMetrics.java
```diff
@@ -0,0 +1,728 @@
...
+  private final String ksqlServiceIdLegacyPrefix;
+  private final String ksqlServicePrefix;
+  private final Time time;
+
+  public ScalablePushQueryExecutorMetrics(
+      final String ksqlServiceId,
+      final Map<String, String> customMetricsTags,
+      final Time time
+  ) {
+    this.ksqlServiceIdLegacyPrefix = ReservedInternalTopics.KSQL_INTERNAL_TOPIC_PREFIX
```
AlanConfluent, line 84: We should be able to skip all of the legacy metrics since nothing is counting on them existing, so we can go right to the new method for metrics.

ksqldb-engine/src/main/java/io/confluent/ksql/internal/ScalablePushQueryExecutorMetrics.java
```diff
@@ -0,0 +1,728 @@
...
+
+    this.ksqlServicePrefix = ReservedInternalTopics.KSQL_INTERNAL_TOPIC_PREFIX;
+    final Map<String, String> metricsTags = new HashMap<>(customMetricsTags);
+    metricsTags.put(KsqlConstants.KSQL_SERVICE_ID_METRICS_TAG, ksqlServiceId);
+    this.customMetricsTags = ImmutableMap.copyOf(metricsTags);
+
+    this.time = Objects.requireNonNull(time, "time");
+    this.metrics = MetricCollectors.getMetrics();
+    this.sensors = new ArrayList<>();
+    this.localRequestsSensor = configureLocalRequestsSensor();
```
AlanConfluent, line 97: I don't think we need `localRequestsSensor` and `remoteRequestsSensor` if we're going to have the tag from RoutingNodeType.  These were legacy metrics for pull queries before we added these tags.

ksqldb-engine/src/main/java/io/confluent/ksql/internal/ScalablePushQueryExecutorMetrics.java
```diff
@@ -0,0 +1,728 @@
...
+  @SuppressFBWarnings(value = "EI_EXPOSE_REP", justification = "should be mutable")
+  public Metrics getMetrics() {
+    return metrics;
+  }
+
+  private Sensor configureLocalRequestsSensor() {
+    final Sensor sensor = metrics.sensor(
+        SCALABLE_PUSH_QUERY_METRIC_GROUP + "-" + SCALABLE_PUSH_REQUESTS + "-local");
+
+    // legacy
```
AlanConfluent, line 274: Can remove legacy metrics here and elsewhere in the file

ksqldb-engine/src/main/java/io/confluent/ksql/internal/ScalablePushQueryExecutorMetrics.java
```diff
@@ -0,0 +1,728 @@
...
+        sensor,
+        SCALABLE_PUSH_REQUESTS + "-error-total",
+        ksqlServiceIdLegacyPrefix + SCALABLE_PUSH_QUERY_METRIC_GROUP,
+        "Total number of erroneous scalable push query requests",
+        legacyCustomMetricsTags,
+        new CumulativeCount()
+    );
+
+    // new metrics with ksql service id in tags
+    addSensor(
```
AlanConfluent, line 348: If we're going to have a rate sensor for errors, we should also have it for requests as well.  I know I was talking about the use of these vs using totals and their timestamps and favoring the latter.  It's probably best to emit both rate and total metrics and we can leave it to the consumer to decide which is preferable to use.

ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/resources/streaming/PushQueryPublisher.java
```diff
@@ -151,11 +167,25 @@ public synchronized void subscribe(final Flow.Subscriber<Collection<StreamedRow>
...
-      queryMetadata = pushQueryMetadata;
+
+      ScalablePushQueryMetadata pushQueryMetadata = null;
+      try {
+        pushQueryMetadata = ksqlEngine
+            .executeScalablePushQuery(analysis, serviceContext, query, pushRouting, routingOptions,
+                plannerOptions, context, scalablePushQueryMetrics);
+
+        final ScalablePushQueryMetadata finalPushQueryMetadata = pushQueryMetadata;
+        pushQueryMetadata.onCompletionOrException((v, throwable) ->
```
AlanConfluent, line 178: We still keep track if this was an error, right?

ksqldb-engine/src/main/java/io/confluent/ksql/util/ScalablePushQueryMetadata.java
```diff
@@ -139,8 +159,33 @@ public ResultType getResultType() {
 
   public void onException(final Consumer<Throwable> consumer) {
     runningFuture.exceptionally(t -> {
+      scalablePushQueryMetrics.ifPresent(metrics ->
```
AlanConfluent, line 162: Ah, ok.  I see, I think we do the same thing in pull queries.  So if an error occurs while running, after the initial start, this is what counts it as an error, right?

ksqldb-rest-app/src/main/java/io/confluent/ksql/api/impl/QueryEndpoint.java
```diff
@@ -191,8 +196,39 @@ private QueryPublisher createScalablePushQueryPublisher(
       final ConfiguredStatement<Query> statement,
       final WorkerExecutor workerExecutor,
       final Map<String, Object> requestProperties,
-      final MetricsCallbackHolder metricsCallbackHolder
+      final MetricsCallbackHolder metricsCallbackHolder,
+      final Optional<ScalablePushQueryExecutorMetrics> scalablePushQueryMetrics
+
   ) {
+    // First thing, set the metrics callback so that it gets called, even if we hit an error
+    final AtomicReference<ScalablePushQueryMetadata> resultForMetrics =
```
AlanConfluent, line 204: I don't see where this reference is being set if it's an SPQ

ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/resources/streaming/StreamedQueryResource.java
```diff
@@ -433,10 +444,45 @@ private EndpointResponse handleQuery(final KsqlSecurityContext securityContext,
       final ImmutableAnalysis analysis = ksqlEngine
           .analyzeQueryWithNoOutputTopic(statement.getStatement(), statement.getStatementText());
 
-      QueryLogger.info("Scalable push query created", statement.getStatementText());
+      // First thing, set the metrics callback so that it gets called, even if we hit an error
+      final AtomicReference<ScalablePushQueryMetadata> resultForMetrics =
```
AlanConfluent, line 448: Same here and for the third endpoint.

ksqldb-engine/src/main/java/io/confluent/ksql/internal/ScalablePushQueryExecutorMetrics.java
```diff
@@ -0,0 +1,636 @@
...
+import org.apache.kafka.common.metrics.stats.Max;
+import org.apache.kafka.common.metrics.stats.Min;
+import org.apache.kafka.common.metrics.stats.Percentile;
+import org.apache.kafka.common.metrics.stats.Percentiles;
+import org.apache.kafka.common.metrics.stats.Percentiles.BucketSizing;
+import org.apache.kafka.common.metrics.stats.Rate;
+import org.apache.kafka.common.utils.Time;
+
+@SuppressWarnings("checkstyle:ClassDataAbstractionCoupling")
+public class ScalablePushQueryExecutorMetrics implements Closeable {
```
AlanConfluent, line 49: I think "Executor" is in the name of the pull query version for legacy reasons -- The original code was in PullQueryExecutor.  I think we can remove that from the name here.

AlanConfluent (approved):

PR #8161, opened 2021-09-22
Enable fetching of subsets of partitions in HARouting

patrickstuedi:
### Description 

Currently in HARouting we cannot detect cases where a node fails to fetch only a subset of the requested partitions. As a consequence, when encountering a fetch error from a node, we assume that none of the partitions could be fetched and subsequent retries against a standby will attempt to fetch the complete set of partitions of the failed node.

The task is make detection of fetch failures more fine grained by allowing the coordinator to detect exactly which partitions could not be fetched, and then retry only the failed partitions.

### Testing done 

Unit tests

### Reviewer checklist

Not ready to review yet


vpapavas (approved):
Thanks @patrickstuedi! LGTM once all the checks pass

ksqldb-engine/src/main/java/io/confluent/ksql/physical/pull/HARouting.java
```diff
@@ -282,45 +281,48 @@ static RoutingResult executeOrRouteQuery(
...
-        forwardTo(node, locations, statement, serviceContext, pullQueryQueue, rowFactory,
-            outputSchema);
-        return RoutingResult.SUCCESS;
-      } catch (StandbyFallbackException e) {
-        LOG.warn("Error forwarding query to node {}. Falling back to standby state which may "
-                + "return stale results", node.location(), e.getCause());
-        return RoutingResult.STANDBY_FALLBACK;
-      } catch (Exception e) {
-        throw new KsqlException(
+          forwardTo(node, ImmutableList.of(locations.get(i)), statement, serviceContext,
```
agavra, line 311: if I'm reading this correctly, we're no longer batching requests so this might multiply the amount of requests to other nodes by the number of remaining partitions - right?

patrickstuedi: Yea, good point and that's a concern, that's also why the PR was marked "not ready" :-).

Ideally we would want the add support to recover individual partitions (as opposed to all partitions of a node) fast, but also keep the request efficiency we have current (where requests for multiple nodes are batched per node). I can see 3 basic ways of implementing something close to this, as discussed below. The first approach is related to the PR update of today.

1) Don't batch but parallelize partition fetching (this update)
- All partition fetching funnels through the executorservice. Means that even though they are separate requests, they are basically back-to-back on the wire on the same connection per node. 
- The advantage of this mode is, it's simple, and it can be extended quite comfortably to a mode where we re-try failed partitions immediately without having to wait for other partitions (from the same node) to be fetched, something we currently do because we have one single batched request per node. 

2) Batch requests per node and return IDs of failed partitions
- Requires changing the RPC interface we currently have in SimpleKsqlClient
- Head-of-line blocking problem: we can't retry failed partitions before the entire batch completed

3) Batch requests per node but notify coordinator about failures asynchronously
- Efficient: We can retry failed partitions immediately without having to wait for the batch to complete
- Complex logic, needs async notification channel, and we need to keep state per partition

AlanConfluent: > All partition fetching funnels through the executorservice. Means that even though they are separate requests, they are basically back-to-back on the wire on the same connection per node.

Technically, if we used http2, they would be multiplexed on the same connection, which could save us some of that efficiency.

> Head-of-line blocking problem: we can't retry failed partitions before the entire batch completed

We always wait until the entire round of requests are done to fallback, which isn't super efficient, but it's simple.  So this wouldn't be any worse, but you could say it prevents us from being more sophisticated in the future.

One thing that also occurs to me is that a partition is not the finest grained unit of data.  Even for a single partition, we could fail mid way after having returned some rows and decide we want to start all over again.  Just pointing this out that we'll never be able (without much more work) to avoid repeats of data during failures.  Obviously, the finer grained the better, but also, especially with heartbeating, failures happen pretty rarely.

If we decided that we wanted to go this route, the least complex solution would be to avoid any kind of signaling about what partition succeeded and what didn't (since a whole request could always fail) and just have one request per partition over http2.

vvcephei: ### Performance concern

Thanks, all. I agree that firing off a request per partition likely has more overhead than batching all the partitions into one request per node. Also that switching to HTTP/2 in the future would diminish this overhead.

I think we'd like to know about the potential performance impact before merging this PR. If it's bad, we should have a plan to immediately switch to HTTP/2 or re-design this feature to multiplex the partition responses. And of course, we would also want to re-evaluate the performance with those fixes in place _also_ to make sure they actually fix the problem.

The punchline is, this is a legitimate performance concern, and we should perform a benchmark test to validate it.

### Failure/retry

Alan is correct. Any time we retry part of the query internally, we would repeat part of the response. This is also true today. There's not much we can do about it in the routing layer. The only way to avoid it is to never retry internally or to buffer the entire query result in the routing node before returning anything to the user.

vvcephei: Separate feedback: we shouldn't be blocking right away on `future.get`. Instead we should just repeatedly check async whether the futures have succeeded or failed and retry the failed ones right away.

vpapavas: A background question: Under what scenarios is it possible for a partition on a node to fail but other partitions on that same node don't fail? What kind of failures are these? And in these cases, do we want to route the query to a different node that hosts the failed partition or do we want to retry on that same node? If the node is not down and the failure is an ephemeral fluke, it is cheaper to retry on the same node instead of rerouting.

patrickstuedi:
I was looking into the jenkins failures (many of the RQTT functional tests were failing with this PR) and I realized that there are two problems related to parallelizing partition fetching (within a node):

- Core fetching routine is not thread safe: the main method executeOrRouteQuery cannot safely be called from multiple threads for different partitions because those calls share the same physicalplan and PullPhysicalPlan.execute is stateful. The underlying source operator is stateful as well. This problem is relevant to local partitions, concurrent fetching different remote partitions is no problem as those tasks will be serialized on the wire.

- Tests are sensitive to row order: several functional tests (e.g., pull-queries-with-range-scan) use the default number of partitions (which i think is 4) but the RQTT test suite requires output rows to be received in the order that's defined by the test case. Parallelizing partition fetching causes the order to be non-deterministic and therefore cause tests to fail.

patrickstuedi:
@vvcephei thanks for commenting on the benchmark. The numbers above are from a setup with 4 ksql servers, 1 active 3 standbys. Are you suggesting to run more actives or more standbys?

PR #8164, opened 2021-09-22
fix: Ensure that we always close /query writer

AlanConfluent:
### Description 
Before this change, when a connection closed, we didn't properly close `PullQueryStreamWriter` immediately which allowed the query to run to completion, which can cause lock out on large tables.

Also, splits the pull query threadpool into two segments, the coordinator and router.  Otherwise, given how the code is written, it's possible to hit a deadlock where all of the coordinator request take all of the threads, which are waiting on routing tasks they spawn, but which can never complete because they cannot be scheduled due to all of the waiting coordinator threads.

Unit tests and manual tests

### Testing done 

### Reviewer checklist


ksqldb-engine/src/main/java/io/confluent/ksql/physical/pull/HARouting.java
```diff
@@ -87,16 +88,26 @@ public HARouting(
   ) {
     this.routingFilterFactory =
         Objects.requireNonNull(routingFilterFactory, "routingFilterFactory");
-    this.executorService = Executors.newFixedThreadPool(
-        ksqlConfig.getInt(KsqlConfig.KSQL_QUERY_PULL_THREAD_POOL_SIZE_CONFIG),
-        new ThreadFactoryBuilder().setNameFormat("pull-query-executor-%d").build());
+    // Split the thread pool in two between coordinator and router
```
agavra, line 91: should they be equal pool sizes? It seems to me like the routing is more CPU bound (routing may route locally and execute query, correct?) and therefore it's OK if the coordinator has more threads

ksqldb-engine/src/main/java/io/confluent/ksql/physical/pull/HARouting.java
```diff
@@ -462,4 +473,6 @@ private static void validateSchema(
     SUCCESS,
     STANDBY_FALLBACK
   }
+
+  private static class Alan1 { }
```
agavra, line 477: 😆 I'm tempted to keep this as an easter egg

agavra (approved):

ksqldb-engine/src/main/java/io/confluent/ksql/physical/pull/HARouting.java
AlanConfluent: I was debating this.  The thing is that the coordinator lives at least as long as the routing nodes and there's a fanout for the number of nodes in the cluster.  For a single node cluster or just a single key lookup, every request will use one of each.  For a two node cluster and scatter gather, it'll be one coordinator and two router. I just optimized for the common case of a single node.  Alternatively, I can introduce another config and make them independent, but I didn't know if it was worth increasing the surface area there.

ksqldb-engine/src/main/java/io/confluent/ksql/physical/pull/HARouting.java
AlanConfluent: oops.  Removed!

ksqldb-engine/src/main/java/io/confluent/ksql/physical/pull/HARouting.java
vpapavas: Not in scope of this PR but I am trying to understand something here. We fire off the futures for each node we need to route a query to. However, we go over the nodes and futures in a for loop and block until a future completes, correct? This hinders us from returning results as soon as they are available as we must wait for the first future to complete until we can proceed to the next although the first might have the most keys or largest partitions and take the longest to complete. Would it be a better design to get notified when a future completes and handle it then instead of blocking in a for loop?

vpapavas: Regarding the thread pool size, the coordinator handles requests and per request we can expect to see no more than 2-3 routing threads since we don't expect to have larger cluster sizes, right? So, it seems to me the router pool would need more threads than the coordinator. On the other hand, they are blocking and the general case is a single node cluster so, I guess it's ok keeping it the same.

AlanConfluent: >Not in scope of this PR but I am trying to understand something here. We fire off the futures for each node we need to route a query to. However, we go over the nodes and futures in a for loop and block until a future completes, correct? 

Yes, that's true.  But we let the resulting rows trickle back and they are enqueued incrementally even though the coordinator is waiting. So we don't wait on one server or another to begin showing results.  We just wait on the longest for for the coordinator to consider the whole request done.

Try running it against a long table scan.

> Regarding the thread pool size, the coordinator handles requests and per request we can expect to see no more than 2-3 routing threads since we don't expect to have larger cluster sizes, right? So, it seems to me the router pool would need more threads than the coordinator. On the other hand, they are blocking and the general case is a single node cluster so, I guess it's ok keeping it the same.

On a single node system, or a single key lookup they are 1:1.  On other cluster sizes, there might be more router pool use, though they can also finish sooner whereas the coordinator waits until the whole thing is done, so it's a bit hard to reason about.

ksqldb-engine/src/main/java/io/confluent/ksql/physical/pull/operators/TableScanOperator.java
```diff
@@ -82,6 +87,10 @@ public Object next() {
           .get(nextLocation.getPartition());
     }
 
+    if (shouldCancelOperations.isDone()) {
```
agavra, line 90: shouldn't this be the first thing that is checked?

ksqldb-engine/src/main/java/io/confluent/ksql/physical/pull/operators/WindowedTableScanOperator.java
```diff
@@ -83,6 +88,10 @@ public Object next() {
           .get(nextLocation.getPartition(), Range.all(), Range.all());
     }
 
+    if (shouldCancelOperations.isDone()) {
```
agavra, line 91: ditto

agavra (approved):

ksqldb-engine/src/main/java/io/confluent/ksql/physical/pull/operators/TableScanOperator.java
AlanConfluent: Sure.  For some reason, it seemed fine to setup the iterator, but that's really not required as you point out.

PR #8166, opened 2021-09-23
fix: stream pull query internal overrides shouldn't clash with query configs

vvcephei:
Previously, if you `SET 'auto.offset.reset'='earliest'` or `'latest'`, the query would fail
with an exception. This patch fixes that and simply overrides the configs as desired.

### Reviewer checklist

### DIFF
--- a/ksqldb-engine/src/main/java/io/confluent/ksql/engine/KsqlEngine.java
+++ b/ksqldb-engine/src/main/java/io/confluent/ksql/engine/KsqlEngine.java
@@ -67,6 +67,7 @@ import io.confluent.ksql.util.TransientQueryMetadata;
 import io.vertx.core.Context;
 import java.io.Closeable;
 import java.util.Collections;
+import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Map.Entry;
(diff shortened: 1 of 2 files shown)


ksqldb-rest-app/src/test/java/io/confluent/ksql/rest/integration/RestApiTest.java
```diff
@@ -499,6 +497,42 @@ public void shouldExecutePushQueryThatReturnsStreamOverRestV1() {
     assertThat(messages.get(3), is("{\"finalMessage\":\"Limit Reached\"}]"));
   }
 
+  @Test
+  public void shouldExecutePullQueryThatReturnsStreamOverRestV1AndIgnoreReset() {
+    // When:
+    final KsqlRequest request =
+        new KsqlRequest(
+            "SELECT USERID, PAGEID, VIEWTIME from " + PAGE_VIEW_STREAM + ";",
+            ImmutableMap.of("auto.offset.reset", "lastest"), // should get ignored
```
vvcephei, line 506: Hey @vpapavas , I did also verify I got the same exception with exactly the config you reported (`earliest`). I figured I could make this test run double-duty as a regression test for the problem you reported and as a test that we really override the property, since setting the config to latest should _not_ return the rows below. So we are also verifying that the override in the code works properly.

ksqldb-engine/src/main/java/io/confluent/ksql/engine/KsqlEngine.java
```diff
@@ -336,18 +337,17 @@ public StreamPullQueryMetadata createStreamPullQuery(
...
+        new HashMap<>(statementOrig.getSessionConfig().getOverrides());
+    // Starting from earliest is semantically necessary.
+    overrides.put(ConsumerConfig.AUTO_OFFSET_RESET_CONFIG, "earliest");
+    // Using a single thread keeps these queries as lightweight as possible, since we are
+    // not counting them against the transient query limit.
+    overrides.put(StreamsConfig.NUM_STREAM_THREADS_CONFIG, 1);
+    // There's no point in EOS, since this query only produces side effects.
+    overrides.put(StreamsConfig.PROCESSING_GUARANTEE_CONFIG, StreamsConfig.AT_LEAST_ONCE);
+
+    final ConfiguredStatement<Query> statement = statementOrig.withConfigOverrides(overrides);
```
vvcephei, line 350: The bug was that I assumed the ImmutableMap builder would let me put in the desired overrides after the `putAll` (like a regular map) and then become immutable only after you call `build`, but it turns out it actually checks that you put each key only once.

I just went ahead and made a copy of the input map (so it will be mutable) and then did the overrides.

agavra (approved):
thanks @vvcephei!

vpapavas (approved):
Thank you @vvcephei! LGTM!

PR #8175, opened 2021-09-24
chore: fix for bug where QueryAnalyzers weren't being constructed with session configs

Sullivan-Patrick:
### Description 
To add a feature flag for [KLIP-50](https://github.com/confluentinc/ksql/blob/master/design-proposals/klip-50-partition-and-offset-in-ksqldb.md) in https://github.com/confluentinc/ksql/pull/8120, a `KsqlConfig` had to be available in every location where a decision had to be made about which psuedo column version to use. For certain queries that require session config overrides, these were being ignored, as only the engine config was being provided. This PR fixes that.

### Testing done 
None. I think this is acceptable because this will be removed when the flag is removed, and other tests should fail if this fix is undone. If there is smart test coverage to be done here, I'm open to adding it.

### Reviewer checklist

### DIFF
--- a/ksqldb-engine/src/main/java/io/confluent/ksql/engine/EngineExecutor.java
+++ b/ksqldb-engine/src/main/java/io/confluent/ksql/engine/EngineExecutor.java
@@ -601,12 +601,15 @@ final class EngineExecutor {
         query,
         sink,
         metaStore,
-        ksqlConfig.cloneWithPropertyOverwrite(statement.getSessionConfig().getOverrides())
+        ksqlConfig,
+        statement.getSessionConfig()
     );
+
     final LogicalPlanNode logicalPlan = new LogicalPlanNode(
         statement.getStatementText(),
         Optional.of(outputNode)
     );
+
     final QueryId queryId = QueryIdUtil.buildId(
         statement.getStatement(),
         engineContext,
--- a/ksqldb-engine/src/main/java/io/confluent/ksql/engine/EngineExecutor.java
+++ b/ksqldb-engine/src/main/java/io/confluent/ksql/engine/EngineExecutor.java
@@ -602,7 +602,7 @@ final class EngineExecutor {
         sink,
         metaStore,
         ksqlConfig,
-        statement.getSessionConfig()
+        statement.getSessionConfig().getConfig(true)
     );
 
     final LogicalPlanNode logicalPlan = new LogicalPlanNode(
--- a/ksqldb-engine/src/main/java/io/confluent/ksql/engine/EngineExecutor.java
+++ b/ksqldb-engine/src/main/java/io/confluent/ksql/engine/EngineExecutor.java
@@ -601,7 +601,7 @@ final class EngineExecutor {
         query,
         sink,
         metaStore,
-        ksqlConfig
+        ksqlConfig.cloneWithPropertyOverwrite(statement.getSessionConfig().getOverrides())
     );
     final LogicalPlanNode logicalPlan = new LogicalPlanNode(
         statement.getStatementText(),
--- a/ksqldb-engine/src/main/java/io/confluent/ksql/engine/QueryEngine.java
+++ b/ksqldb-engine/src/main/java/io/confluent/ksql/engine/QueryEngine.java
@@ -56,12 +56,13 @@ class QueryEngine {
       final Query query,
       final Optional<Sink> sink,
       final MetaStore metaStore,
-      final KsqlConfig config
+      final KsqlConfig config,
+      final SessionConfig sessionConfig
   ) {
     final String outputPrefix = config.getString(KsqlConfig.KSQL_OUTPUT_TOPIC_NAME_PREFIX_CONFIG);
 
     final QueryAnalyzer queryAnalyzer =
-        new QueryAnalyzer(metaStore, outputPrefix, config);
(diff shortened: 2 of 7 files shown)


ksqldb-engine/src/main/java/io/confluent/ksql/engine/EngineExecutor.java
```diff
@@ -601,7 +601,7 @@ private ExecutorPlans planQuery(
         query,
         sink,
         metaStore,
-        ksqlConfig
+        ksqlConfig.cloneWithPropertyOverwrite(statement.getSessionConfig().getOverrides())
```
vcrfxia, line 604: This instance of passing `ksqlConfig` around wasn't added as part of the feature flag PR that introduced this bug. Updating this line looks like it has implications beyond the scope of the feature flag?

ksqldb-engine/src/main/java/io/confluent/ksql/engine/QueryEngine.java
```diff
@@ -56,12 +56,13 @@ static OutputNode buildQueryLogicalPlan(
...
       final MetaStore metaStore,
-      final KsqlConfig config
+      final KsqlConfig config,
+      final SessionConfig sessionConfig
   ) {
     final String outputPrefix = config.getString(KsqlConfig.KSQL_OUTPUT_TOPIC_NAME_PREFIX_CONFIG);
 
     final QueryAnalyzer queryAnalyzer =
-        new QueryAnalyzer(metaStore, outputPrefix, config);
+        new QueryAnalyzer(metaStore, outputPrefix, config.cloneWithPropertyOverwrite(sessionConfig.getOverrides()));
```
vcrfxia, line 65: Is there a reason we can't just use `sessionConfig.getConfig(true)` directly?

Also, if this is the only thing we need to pipe the session config through for, we could also just pass `KsqlConfig configWithOverrides` directly.

Sullivan-Patrick: I think both of these points are good, pushing changes with the fixes now.

ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/resources/streaming/StreamedQueryResource.java
```diff
@@ -339,7 +339,8 @@ private EndpointResponse handleQuery(final KsqlSecurityContext securityContext,
 
     if (statement.getStatement().isPullQuery()) {
       final ImmutableAnalysis analysis = ksqlEngine
-          .analyzeQueryWithNoOutputTopic(statement.getStatement(), statement.getStatementText());
+          .analyzeQueryWithNoOutputTopic(
+              statement.getStatement(), statement.getStatementText(), request.getConfigOverrides());
```
vcrfxia, line 343: nit: use `configProperties` instead of `request.getConfigOverrides()` (and also below). The way that the methods in this class are set up is really confusing and would probably benefit from a larger refactor but that's outside the scope of this fix.

ksqldb-rest-app/src/main/java/io/confluent/ksql/api/impl/QueryEndpoint.java
```diff
@@ -138,7 +138,8 @@ public QueryPublisher createQueryPublisher(
 
     if (statement.getStatement().isPullQuery()) {
       final ImmutableAnalysis analysis = ksqlEngine
-          .analyzeQueryWithNoOutputTopic(statement.getStatement(), statement.getStatementText());
+          .analyzeQueryWithNoOutputTopic(
+              statement.getStatement(), statement.getStatementText(), properties);
```
vcrfxia, line 142: nit: let's be consistent. If we're going to use `properties`, let's also use it below (in this file). If we're going to use `statement.getSessionConfig().getOverrides()` below, then that's use that here too. (I know they're the same, but it's unintuitive as a reader that that's the case; consistency removes that extra question.)

vcrfxia (commented):
Thanks @Sullivan-Patrick -- LGTM besides the two nits inline. Hopefully this code is removed soon enough (when the feature flag is removed) that the code smells here won't be an issue but let's make some minor improvements in case optimism fails.

Also, I've updated the title of this PR from "fix" to "chore" since we don't want this change to appear in the changelog even though it's a bugfix -- it's a bugfix to an unreleased feature, so it's not of interest to users.

ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/resources/streaming/StreamedQueryResource.java
```diff
@@ -410,7 +411,11 @@ private EndpointResponse handleQuery(final KsqlSecurityContext securityContext,
             configProperties)) {
 
       final ImmutableAnalysis analysis = ksqlEngine
-          .analyzeQueryWithNoOutputTopic(statement.getStatement(), statement.getStatementText());
+          .analyzeQueryWithNoOutputTopic(
+              statement.getStatement(),
+              statement.getStatementText(),
+              request.getConfigOverrides()
```
vcrfxia, line 417:
```suggestion
              configProperties
```
Let's be consistent with above.

vcrfxia (approved):

PR #8176, opened 2021-09-24
fix: prevent duplicates from being written to cmd

wcarlson5:
move if not exists checks from executors to the plan

### Testing done 

### Reviewer checklist

### DIFF
--- a/ksqldb-engine/src/main/java/io/confluent/ksql/ddl/commands/CreateSourceFactory.java
+++ b/ksqldb-engine/src/main/java/io/confluent/ksql/ddl/commands/CreateSourceFactory.java
@@ -60,16 +60,14 @@ public final class CreateSourceFactory {
   private final SerdeFeaturessSupplier valueSerdeFeaturesSupplier;
   private final KeySerdeFactory keySerdeFactory;
   private final ValueSerdeFactory valueSerdeFactory;
-  private final MetaStore metaStore;
 
-  public CreateSourceFactory(final ServiceContext serviceContext, final MetaStore metaStore) {
+  public CreateSourceFactory(final ServiceContext serviceContext) {
     this(
         serviceContext,
         (s, f, e, k) -> SerdeFeaturesFactory.buildKeyFeatures(s, f),
         SerdeFeaturesFactory::buildValueFeatures,
         new GenericKeySerDe(),
-        new GenericRowSerDe(),
-        metaStore
+        new GenericRowSerDe()
     );
   }
 
@@ -79,9 +77,7 @@ public final class CreateSourceFactory {
       final SerdeFeaturessSupplier keySerdeFeaturesSupplier,
       final SerdeFeaturessSupplier valueSerdeFeaturesSupplier,
       final KeySerdeFactory keySerdeFactory,
-      final ValueSerdeFactory valueSerdeFactory,
-      final MetaStore metaStore
-  ) {
+      final ValueSerdeFactory valueSerdeFactory) {
     this.serviceContext = requireNonNull(serviceContext, "serviceContext");
     this.keySerdeFeaturesSupplier =
         requireNonNull(keySerdeFeaturesSupplier, "keySerdeFeaturesSupplier");
@@ -89,7 +85,6 @@ public final class CreateSourceFactory {
         requireNonNull(valueSerdeFeaturesSupplier, "valueSerdeFeaturesSupplier");
     this.keySerdeFactory = requireNonNull(keySerdeFactory, "keySerdeFactory");
     this.valueSerdeFactory = requireNonNull(valueSerdeFactory, "valueSerdeFactory");
-    this.metaStore = requireNonNull(metaStore);
   }
 
   // This method is called by CREATE_AS statements
@@ -117,14 +112,6 @@ public final class CreateSourceFactory {
     final LogicalSchema schema = buildSchema(statement.getElements(), ksqlConfig);
     final Optional<TimestampColumn> timestampColumn =
         buildTimestampColumn(ksqlConfig, props, schema);
-    final DataSource dataSource = metaStore.getSource(sourceName);
-
-    if (dataSource != null && !statement.isOrReplace() && !statement.isNotExists()) {
-      final String sourceType = dataSource.getDataSourceType().getKsqlType();
-      throw new KsqlException(
-          String.format("Cannot add stream '%s': A %s with the same name already exists",
-             sourceName.text(), sourceType.toLowerCase()));
-    }
 
     return new CreateStreamCommand(
         sourceName,
@@ -161,14 +148,7 @@ public final class CreateSourceFactory {
     final CreateSourceProperties props = statement.getProperties();
     final String topicName = ensureTopicExists(props, serviceContext);
     final LogicalSchema schema = buildSchema(statement.getElements(), ksqlConfig);
-    final DataSource dataSource = metaStore.getSource(sourceName);
 
-    if (dataSource != null && !statement.isOrReplace() && !statement.isNotExists()) {
-      final String sourceType = dataSource.getDataSourceType().getKsqlType();
-      throw new KsqlException(
-          String.format("Cannot add table '%s': A %s with the same name already exists",
-              sourceName.text(), sourceType.toLowerCase()));
-    }
     if (schema.key().isEmpty()) {
       final boolean usingSchemaInference = props.getValueSchemaId().isPresent();
 
--- a/ksqldb-engine/src/main/java/io/confluent/ksql/ddl/commands/CreateSourceFactory.java
+++ b/ksqldb-engine/src/main/java/io/confluent/ksql/ddl/commands/CreateSourceFactory.java
@@ -25,6 +25,8 @@ import io.confluent.ksql.execution.plan.Formats;
 import io.confluent.ksql.execution.streams.timestamp.TimestampExtractionPolicyFactory;
 import io.confluent.ksql.execution.timestamp.TimestampColumn;
 import io.confluent.ksql.logging.processing.NoopProcessingLogContext;
+import io.confluent.ksql.metastore.MetaStore;
+import io.confluent.ksql.metastore.model.DataSource;
 import io.confluent.ksql.name.ColumnName;
 import io.confluent.ksql.name.SourceName;
 import io.confluent.ksql.parser.properties.with.CreateSourceProperties;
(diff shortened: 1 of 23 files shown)


ksqldb-engine/src/main/java/io/confluent/ksql/ddl/commands/CreateSourceFactory.java
```diff
@@ -117,14 +110,6 @@ public CreateStreamCommand createStreamCommand(
     final LogicalSchema schema = buildSchema(statement.getElements(), ksqlConfig);
     final Optional<TimestampColumn> timestampColumn =
         buildTimestampColumn(ksqlConfig, props, schema);
-    final DataSource dataSource = metaStore.getSource(sourceName);
```
rodesai, line 120: why did we need to get rid of this check?

ksqldb-engine/src/main/java/io/confluent/ksql/engine/EngineExecutor.java
```diff
@@ -511,12 +507,28 @@ private boolean isSourceTableMaterializationEnabled() {
         .getBoolean(KsqlConfig.KSQL_SOURCE_TABLE_MATERIALIZATION_ENABLED);
   }
 
+  private SourceName getSourceName(final ConfiguredStatement<?> statement) {
```
rodesai, line 510: return `Optional<SourceName>` here instead of using `null` to signal no source name

ksqldb-engine/src/main/java/io/confluent/ksql/engine/EngineExecutor.java
```diff
@@ -511,12 +507,28 @@ private boolean isSourceTableMaterializationEnabled() {
...
+  }
+
   // Known to be non-empty
   @SuppressWarnings("OptionalGetWithoutIsPresent")
   KsqlPlan plan(final ConfiguredStatement<?> statement) {
     try {
       throwOnNonExecutableStatement(statement);
 
+      final SourceName sourceName = getSourceName(statement);
+      if (sourceName != null && engineContext.getMetaStore().getSource(sourceName) != null) {
```
rodesai, line 528: you need to check that this is an `IF NOT EXISTS` statement and only ignore the missing source in that case

ksqldb-engine/src/test/java/io/confluent/ksql/engine/KsqlEngineTest.java
```diff
@@ -1392,29 +1392,6 @@ public void shouldNotThrowWhenPreparingDuplicateTable() {
     // Then: no exception thrown
   }
 
-  @Test
```
wcarlson5, line 1395: These shouldn't be needed as in order to pass validation it needs to not fail the if not exists part

rodesai (commented):
LGTM other than the one comment inline. Also, can you add a case to `RecoveryTest`?

ksqldb-engine/src/main/java/io/confluent/ksql/engine/EngineExecutor.java
```diff
@@ -511,12 +506,61 @@ private boolean isSourceTableMaterializationEnabled() {
...
+        && ((CreateStreamAsSelect) statement.getStatement()).isNotExists()) {
+      return Optional.ofNullable(((CreateAsSelect) statement.getStatement()).getName());
+    }
+    return Optional.empty();
+  }
+
   // Known to be non-empty
+  // CHECKSTYLE_RULES.OFF: CyclomaticComplexity
   @SuppressWarnings("OptionalGetWithoutIsPresent")
   KsqlPlan plan(final ConfiguredStatement<?> statement) {
```
rodesai, line 529: can you add a `KsqlEngine` test that ensures we return an empty plan?

ksqldb-engine/src/main/java/io/confluent/ksql/engine/EngineExecutor.java
```diff
@@ -511,12 +506,61 @@ private boolean isSourceTableMaterializationEnabled() {
...
     try {
       throwOnNonExecutableStatement(statement);
 
+      final Optional<SourceName> sourceName = getSourceNameIfIsNotExists(statement);
+      if (sourceName.isPresent()
+          && engineContext.getMetaStore().getSource(sourceName.get()) != null) {
+        return new KsqlPlan() {
```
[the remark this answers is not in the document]
rodesai: Actually can we just have this return an `Optional` - I was worried there would be a bunch of references but this method is actually not referenced in many places.

ksqldb-engine/src/main/java/io/confluent/ksql/engine/EngineExecutor.java
```diff
@@ -511,12 +506,36 @@ private boolean isSourceTableMaterializationEnabled() {
...
+  Optional<KsqlPlan> plan(final ConfiguredStatement<?> statement) {
     try {
       throwOnNonExecutableStatement(statement);
 
+      final Optional<SourceName> sourceName = getSourceNameIfIsNotExists(statement);
+      if (sourceName.isPresent()
+          && engineContext.getMetaStore().getSource(sourceName.get()) != null) {
+        return Optional.empty();
+      }
+
```
spena, line 538: It would be more readable to have boolean method that just checks if the CREATE  statement can continue with the plan or not, like:
```
    if (!shouldGeneratePlanFor(statement)) { 
       return Optional.empty();
    }
```

This way we could handle any future statement that should not participate in the plan.

ksqldb-engine/src/main/java/io/confluent/ksql/engine/EngineExecutor.java
```diff
@@ -511,12 +506,36 @@ private boolean isSourceTableMaterializationEnabled() {
...
+    } else if (statement.getStatement() instanceof CreateTable
+        && ((CreateTable) statement.getStatement()).isNotExists()) {
+      return Optional.ofNullable(((CreateTable) statement.getStatement()).getName());
+    } else if (statement.getStatement() instanceof CreateTableAsSelect
+        && ((CreateTableAsSelect) statement.getStatement()).isNotExists()) {
+      return Optional.ofNullable(((CreateAsSelect) statement.getStatement()).getName());
+    } else if (statement.getStatement() instanceof CreateStreamAsSelect
+        && ((CreateStreamAsSelect) statement.getStatement()).isNotExists()) {
+      return Optional.ofNullable(((CreateAsSelect) statement.getStatement()).getName());
+    }
```
spena, line 522: We can reduce this by having the `isNotExists` in the `CreateSource` and `CreateAsSelect` classes instead.

ksqldb-engine/src/main/java/io/confluent/ksql/engine/KsqlEngine.java
```diff
@@ -273,17 +273,20 @@ public KsqlPlan plan(
   @Override
   public ExecuteResult execute(final ServiceContext serviceContext, final ConfiguredKsqlPlan plan) {
     try {
+      if (!plan.getPlan().isPresent()) {
+        return ExecuteResult.of("should not execute");
```
spena, line 277: Is there another message we can use instead of `should not execute`? Btw, is this a bug that execute is called with an empty plan?

ksqldb-engine/src/main/java/io/confluent/ksql/engine/SandboxedExecutionContext.java
```diff
@@ -138,11 +138,15 @@ public ExecuteResult execute(
       final ServiceContext serviceContext,
       final ConfiguredKsqlPlan ksqlPlan
   ) {
+    if (!ksqlPlan.getPlan().isPresent()) {
+      return ExecuteResult.of("not to execute");
```
spena, line 142: Same question from KsqlEngine

ksqldb-engine/src/main/java/io/confluent/ksql/planner/plan/ConfiguredKsqlPlan.java
```diff
@@ -18,28 +18,29 @@
 import io.confluent.ksql.config.SessionConfig;
 import io.confluent.ksql.engine.KsqlPlan;
 import java.util.Objects;
+import java.util.Optional;
 
 public final class ConfiguredKsqlPlan {
 
-  private final KsqlPlan plan;
+  private final Optional<KsqlPlan> plan;
```
spena, line 25: Wouldn't it be better to have a `Optional<ConfiguredKsqlPlan>` that is passed to the KsqlEngine and SandboxedExecutionContext instead? A configured plan with no plan inside does not seem right.

ksqldb-rest-app/src/test/java/io/confluent/ksql/rest/server/computation/RecoveryTest.java
```diff
@@ -701,6 +701,16 @@ public void shouldRecoverDropWithTerminate() {
     shouldRecover(commands);
   }
 
+  @Test
+  public void shouldRecoverIfNotExists() {
```
spena, line 705: I don't think this test will pass because the restore issue is not fixed yet. The PR is only to prevent writing the CREATE IF NOT EXISTS in the command topic, isn't it?

ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/computation/ValidatedCommandFactory.java
```diff
@@ -92,17 +95,17 @@ private static Command ensureDeserializable(final Command command) {
     }
   }
 
-  private static Command createCommand(
+  private static Optional<Command> createCommand(
```
spena, line 98: There's something I don't feel good about this 'cause you create a command but it may return nothing. This is breaking the meaning of the method. Is there something else we can do?

ksqldb-engine/src/main/java/io/confluent/ksql/engine/KsqlEngine.java
wcarlson5: Refactored this out

ksqldb-engine/src/main/java/io/confluent/ksql/engine/SandboxedExecutionContext.java
wcarlson5: same answer :)

spena:
@wcarlson5 After a second look, I now feel we are complicating the code a little bit (including myself). I think the best approach would be to check the statement in the `DistributingExecutor` class. We have the info there to do it. We can check if it the statement is a `CreateSource` or `CreateAsSelect`, then check if it is a `isNotExists` type, and check the metastore if the source already exists. If all those conditions are met, then we can return a successful response (with the right message). All of this before writing the statement to the command topic. It is in one place only. We do these with other statements too, like for read-only topics or authorization denies. What do you think? It will less prone for errors.

wcarlson5:
@spena do you mean something like https://github.com/confluentinc/ksql/pull/8206. It does seems much easier. We can do the clean up that @rodesai wanted in a separate PR

spena:
@wcarlson5 Yes, that's a simpler solution that should work. I left some comments there.

wcarlson5:
https://github.com/confluentinc/ksql/pull/8206

PR #8177, opened 2021-09-24
refactor: Refactor ConnectFormatSchemaTranslator to take translator object instead of lamda function so that the translator can take in config to decide how to transform field names

lihaosky:
Current `ConnectFormatSchemaTranslator` take a Function to translate `Schema` to ksql `Schema`. In the function, it always uppercase the field names. This PR refactor the param type from Function to Class so that we can pass in config to control how field names should be translated.

Actual config and translation fix will be following PRs

### DIFF
--- a/ksqldb-serde/src/main/java/io/confluent/ksql/serde/connect/ConnectKsqlSchemaTranslator.java
+++ b/ksqldb-serde/src/main/java/io/confluent/ksql/serde/connect/ConnectKsqlSchemaTranslator.java
@@ -21,7 +21,7 @@ import io.confluent.ksql.schema.connect.SqlSchemaFormatter.Option;
 import io.confluent.ksql.util.DecimalUtil;
 import io.confluent.ksql.util.KsqlException;
 import java.util.Map;
-import java.util.function.Function;
+import java.util.function.BiFunction;
 import org.apache.kafka.connect.data.Date;
 import org.apache.kafka.connect.data.Field;
 import org.apache.kafka.connect.data.Schema;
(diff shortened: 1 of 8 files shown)


ksqldb-serde/src/main/java/io/confluent/ksql/serde/connect/ConnectKsqlSchemaTranslator.java
```diff
@@ -32,42 +32,42 @@
...
-      .put(Type.FLOAT32, s -> Schema.OPTIONAL_FLOAT64_SCHEMA)
-      .put(Type.FLOAT64, s -> Schema.OPTIONAL_FLOAT64_SCHEMA)
-      .put(Type.STRING, s -> Schema.OPTIONAL_STRING_SCHEMA)
-      .put(Type.BOOLEAN, s -> Schema.OPTIONAL_BOOLEAN_SCHEMA)
-      .put(Type.BYTES, ConnectSchemaUtil::toKsqlBytesSchema)
-      .put(Type.ARRAY, ConnectSchemaUtil::toKsqlArraySchema)
-      .put(Type.MAP, ConnectSchemaUtil::toKsqlMapSchema)
-      .put(Type.STRUCT, ConnectSchemaUtil::toKsqlStructSchema)
-      .build();
+  private final Map<Type, BiFunction<ConnectKsqlSchemaTranslator, Schema, Schema>> connectToSql =
```
vcrfxia, line 42: Why is this changing to become non-static? (Under what scenarios will this mapping depend on the instance?)

vcrfxia (commented):
Thanks @lihaosky . There are no functional changes in this PR, right? If it's a pure refactor, then it'd be nice to update the PR title from "fix" to "refactor". 

One comment inline. No major concerns as this appears to be a pretty straightforward refactor. Looking forward to seeing the PRs to follow :)

ksqldb-serde/src/main/java/io/confluent/ksql/serde/connect/ConnectKsqlSchemaTranslator.java
lihaosky: Because some methods like `toKsqlBytesSchema` are changed to non-static, to access those methods here, this map also needs to be non-static.

lihaosky:
Hi @vcrfxia , yeah there's no functional changes in the PR. Changed title to "refactor".

vcrfxia (approved):
Thanks @lihaosky ! LGTM.

PR #8178, opened 2021-09-27
feat: add detailed scalable push query metrics with type breakdown

nateab:
### Description 

fixed merge conflicts of #8144

Add some metrics on scalable push queries such as:
- Count of local/remote requests
- Count/rate of errors
- Min, max, and avg connection duration
- Request/response size
- Rows returned/processed  
- status code of the request - 2XX, 3XX, 4XX, 5XX

These are then further broken down by source and routing node type. 
Adds additional pull query metrics:

For source, we current support:

- Windowed
- Non Windowed

For routing node type, we currently support:

- source node (the initial node receiving the request)
- remote node (any node receiving a forwarded request)

### Testing done 
Unit and manual testing 

### Reviewer checklist


AlanConfluent (approved):
Already reviewed same PR, so LG

PR #8183, opened 2021-09-28
chore: fix pull and push query v2 to properly ignore ROWOFFSET/ROWPARTITION

AlanConfluent:
### Description 
Ensures that we don't provide the new pseudo column version (and therefore ROWOFFSET and ROWPARTITION) to pull queries or scalable push queries yet since they don't (and in the case of pull queries won't) support it.

Also, stores the right state store schema in `MaterializationInfo` since this breaks when comparing to the actual rows (which have the new cols) without this change.

### Testing done 
Ran RQTT tests and unit tests.

### Reviewer checklist

### DIFF
--- a/ksqldb-common/src/main/java/io/confluent/ksql/schema/ksql/SystemColumns.java
+++ b/ksqldb-common/src/main/java/io/confluent/ksql/schema/ksql/SystemColumns.java
@@ -163,7 +163,15 @@ public final class SystemColumns {
   }
 
   public static int getPseudoColumnVersionFromConfig(final KsqlConfig ksqlConfig) {
+    return getPseudoColumnVersionFromConfig(ksqlConfig, false);
+  }
+
+  public static int getPseudoColumnVersionFromConfig(
+      final KsqlConfig ksqlConfig,
+      final boolean forPullOrScalablePushQuery
+  ) {
     return ksqlConfig.getBoolean(KsqlConfig.KSQL_ROWPARTITION_ROWOFFSET_ENABLED)
+        && !forPullOrScalablePushQuery
         ? CURRENT_PSEUDOCOLUMN_VERSION_NUMBER
         : LEGACY_PSEUDOCOLUMN_VERSION_NUMBER;
   }
--- a/ksqldb-engine/src/main/java/io/confluent/ksql/planner/plan/QueryProjectNode.java
+++ b/ksqldb-engine/src/main/java/io/confluent/ksql/planner/plan/QueryProjectNode.java
@@ -100,7 +100,7 @@ public class QueryProjectNode extends ProjectNode {
     this.addAdditionalColumnsToIntermediateSchema = shouldAddAdditionalColumnsInSchema();
     this.outputSchema = buildOutputSchema(metaStore);
     this.intermediateSchema = QueryLogicalPlanUtil.buildIntermediateSchema(
-          source.getSchema(),
+          source.getSchema().withoutPseudoAndKeyColsInValue(),
           addAdditionalColumnsToIntermediateSchema,
           isWindowed,
           ksqlConfig
--- a/ksqldb-functional-tests/src/test/java/io/confluent/ksql/test/rest/RestQueryTranslationTest.java
+++ b/ksqldb-functional-tests/src/test/java/io/confluent/ksql/test/rest/RestQueryTranslationTest.java
@@ -108,6 +108,7 @@ public class RestQueryTranslationTest {
       .withProperty(KsqlConfig.KSQL_QUERY_PULL_INTERPRETER_ENABLED, true)
       .withProperty(KsqlConfig.KSQL_QUERY_PUSH_V2_REGISTRY_INSTALLED, true)
       .withProperty(KsqlConfig.KSQL_QUERY_PUSH_V2_ENABLED, true)
+      .withProperty(KsqlConfig.KSQL_ROWPARTITION_ROWOFFSET_ENABLED, true)
       .withStaticServiceContext(TEST_HARNESS::getServiceContext)
       .build();
 
--- a/ksqldb-streams/src/main/java/io/confluent/ksql/execution/streams/SourceBuilderBase.java
+++ b/ksqldb-streams/src/main/java/io/confluent/ksql/execution/streams/SourceBuilderBase.java
@@ -29,6 +29,7 @@ import io.confluent.ksql.execution.plan.KTableHolder;
 import io.confluent.ksql.execution.plan.PlanInfo;
 import io.confluent.ksql.execution.plan.SourceStep;
 import io.confluent.ksql.execution.runtime.RuntimeBuildContext;
+import io.confluent.ksql.schema.ksql.LogicalSchema;
 import io.confluent.ksql.schema.ksql.PhysicalSchema;
 import java.util.Collection;
 import java.util.function.Function;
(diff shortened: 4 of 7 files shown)


agavra (commented):
two comments: 
- do we need any test coverage?
- can we change the PR title to say "ignore ROWOFFSET/ROWPARTITION with pull and push query v2"? title makes it seem like we are supporting these features, when we're really just ignoring them :)

ksqldb-common/src/main/java/io/confluent/ksql/schema/ksql/SystemColumns.java
```diff
@@ -163,7 +163,15 @@ public static boolean mustBeMaterializedForTableJoins(final ColumnName columnNam
...
   public static int getPseudoColumnVersionFromConfig(final KsqlConfig ksqlConfig) {
+    return getPseudoColumnVersionFromConfig(ksqlConfig, false);
+  }
+
+  public static int getPseudoColumnVersionFromConfig(
+      final KsqlConfig ksqlConfig,
+      final boolean forPullOrScalablePushQuery
+  ) {
     return ksqlConfig.getBoolean(KsqlConfig.KSQL_ROWPARTITION_ROWOFFSET_ENABLED)
+        && !forPullOrScalablePushQuery
```
vcrfxia, line 174: Hacking around the pseudocolumn version in this way feels brittle. We should update this to use the new `SystemColumns#isDisallowedInPullQueries()` method being added in https://github.com/confluentinc/ksql/pull/8135. I'm fine with merging this PR first and having @Sullivan-Patrick update this in his PR in order to unblock the cyclic dependency, but I definitely don't think we should hacking pull queries to use the old pseudocolumn version as in this PR when the feature goes out.

ksqldb-engine/src/main/java/io/confluent/ksql/planner/plan/QueryProjectNode.java
```diff
@@ -100,7 +100,7 @@ public QueryProjectNode(
     this.addAdditionalColumnsToIntermediateSchema = shouldAddAdditionalColumnsInSchema();
     this.outputSchema = buildOutputSchema(metaStore);
     this.intermediateSchema = QueryLogicalPlanUtil.buildIntermediateSchema(
-          source.getSchema(),
+          source.getSchema().withoutPseudoAndKeyColsInValue(),
```
vcrfxia, line 103: What's the purpose of this change? Is this fixing an existing bug, or something introduced with the KLIP-50 work?

ksqldb-functional-tests/src/test/java/io/confluent/ksql/test/rest/RestQueryTranslationTest.java
```diff
@@ -108,6 +108,7 @@
       .withProperty(KsqlConfig.KSQL_QUERY_PULL_INTERPRETER_ENABLED, true)
       .withProperty(KsqlConfig.KSQL_QUERY_PUSH_V2_REGISTRY_INSTALLED, true)
       .withProperty(KsqlConfig.KSQL_QUERY_PUSH_V2_ENABLED, true)
+      .withProperty(KsqlConfig.KSQL_ROWPARTITION_ROWOFFSET_ENABLED, true)
```
vcrfxia, line 111: This does not look like it's meant to be checked in.

ksqldb-streams/src/main/java/io/confluent/ksql/execution/streams/SourceBuilderBase.java
```diff
@@ -91,11 +92,14 @@
...
 
+    final LogicalSchema stateStoreSchema = source.getSourceSchema()
+        .withPseudoColumnsToMaterialize(source.getPseudoColumnVersion());
+
     return KTableHolder.materialized(
         ktable,
         buildSchema(source, false),
         ExecutionKeyFactory.unwindowed(buildContext),
-        MaterializationInfo.builder(stateStoreName, physicalSchema.logicalSchema())
+        MaterializationInfo.builder(stateStoreName, stateStoreSchema)
```
vcrfxia, line 102: Thanks for this bug fix, @AlanConfluent !

ksqldb-engine/src/main/java/io/confluent/ksql/planner/plan/QueryProjectNode.java
AlanConfluent: The purpose of this change is so that when there's no WHERE clause, which already does this call (In `QueryFilterNode`), we don't automatically include pseudo columns in our intermediate schema. I guess you could say this was a preexisting bug, but I don't think it was possible to surface it before the new pseudo columns because nothing unexpected existed in the source schema. The issue is that we build up the intermediate row which must agree with the intermediate schema, including the pseudo columns, but pull queries have no notion of ROWOFFSET or ROWPARTITION. So rather than trying to add support for it in the intermediate row, I've removed it from the schema.

ksqldb-functional-tests/src/test/java/io/confluent/ksql/test/rest/RestQueryTranslationTest.java
AlanConfluent: Well, I added it to verify that this works with the new feature enabled.  I could avoid checking it in, but then something could break again and we wouldn't know.  I'm ok to enable it by default here unless you think it's not a good idea.

ksqldb-streams/src/main/java/io/confluent/ksql/execution/streams/SourceBuilderBase.java
AlanConfluent: No problem.  Glad to help.

ksqldb-common/src/main/java/io/confluent/ksql/schema/ksql/SystemColumns.java
AlanConfluent: I agree.  This part does feel a little brittle and was a bit of a stab at getting it working, but I agree I like that method better.  Happy to let him resolve it if that works.

AlanConfluent:
> do we need any test coverage?

Well, we have all of the existing RQTT tests, some of which fail with the flag set and without the fixes.

For the unit tests, I updated some of them to include this slightly new behavior.

> can we change the PR title to say "ignore ROWOFFSET/ROWPARTITION with pull and push query v2"? title makes it seem like we are supporting these features, when we're really just ignoring them :)

Yes!

ksqldb-engine/src/main/java/io/confluent/ksql/planner/plan/QueryProjectNode.java
vcrfxia: Got it; thanks for explaining!

ksqldb-functional-tests/src/test/java/io/confluent/ksql/test/rest/RestQueryTranslationTest.java
vcrfxia: OK, if the RQTTs are passing with this enabled then I'm happy enabling it if you are. We're planning to flip the feature flag default to true within the next week anyway.

vcrfxia (approved):
Thanks @AlanConfluent ! LGTM besides the discussion on using `SystemColumns#isDisallowedInPullQueries()` rather than hacking the pseudocolumn version. I believe you and @Sullivan-Patrick are already coordinating about that offline since https://github.com/confluentinc/ksql/pull/8135 is also ready to merge as soon as the build passes.

I've also updated the PR title from `feat` to `chore` since I don't think this PR should appear in the changelog -- it's part of the larger ROWOFFSET/ROWPARTITION feature that is still WIP.

PR #8186, opened 2021-09-28
test: QTT coverage for ROWPARTITION/ROWOFFSET

Sullivan-Patrick:
### Description 
QTTs for [KLIP-50](https://github.com/confluentinc/ksql/blob/master/design-proposals/klip-50-partition-and-offset-in-ksqldb.md).

### Reviewer checklist

### DIFF
--- a/ksqldb-functional-tests/src/test/resources/query-validation-tests/row-offset-and-partition.json
+++ b/ksqldb-functional-tests/src/test/resources/query-validation-tests/row-offset-and-partition.json
@@ -76,34 +76,6 @@
       }
     },
     {
-      "name": "verify ROWPARTITION is not queryable with feature disabled",
-      "properties": {
-        "ksql.rowpartition.rowoffset.enabled": false
-      },
-      "statements": [
-        "CREATE STREAM input (id INT KEY, val INTEGER) WITH (kafka_topic='test_topic', value_format='JSON');",
-        "CREATE STREAM output AS SELECT ROWPARTITION FROM input EMIT CHANGES;"
-      ],
-      "expectedException": {
-        "type": "io.confluent.ksql.util.KsqlStatementException",
-        "message": "SELECT column 'ROWPARTITION' cannot be resolved."
-      }
-    },
-    {
-      "name": "verify ROWOFFSET is not queryable with feature disabled",
-      "properties": {
-        "ksql.rowpartition.rowoffset.enabled": false
-      },
-      "statements": [
-        "CREATE STREAM input (id INT KEY, val INTEGER) WITH (kafka_topic='test_topic', value_format='JSON');",
-        "CREATE STREAM output AS SELECT ROWOFFSET FROM input EMIT CHANGES;"
-      ],
-      "expectedException": {
-        "type": "io.confluent.ksql.util.KsqlStatementException",
-        "message": "SELECT column 'ROWOFFSET' cannot be resolved."
-      }
-    },
-    {
       "comments": "Should be removed after rowpartition.rowoffset.enabled feature flag is removed",
       "name": "create ROWOFFSET and ROWPARTITION sink user columns as STRINGs- ROWOFFSET/ROWPARTITION enabled",
       "properties": {
@@ -167,12 +139,13 @@
         "ksql.rowpartition.rowoffset.enabled": true
       },
       "statements": [
-        "CREATE STREAM input (id INT KEY, val INT) WITH (kafka_topic='test_topic', value_format='JSON');",
-        "INSERT INTO input (id, val, rowpartition) VALUES (1, 2, 3);"
+        "CREATE STREAM a (id INT KEY, val INT) WITH (kafka_topic='test_topic', value_format='JSON');",
+        "CREATE STREAM b (id INT KEY) WITH (kafka_topic='test_topic_2', value_format='JSON');",
+        "INSERT INTO a SELECT id, ROWPARTITION from b EMIT CHANGES;"
       ],
       "expectedException": {
         "type": "io.confluent.ksql.util.KsqlStatementException",
-        "message": "Failed to insert values into 'INPUT'. Inserting into column `ROWPARTITION` is not allowed."
+        "message": "Reserved column name in select: `ROWPARTITION`. Please remove or alias the column."
       }
     },
     {
@@ -181,12 +154,13 @@
         "ksql.rowpartition.rowoffset.enabled": true
       },
       "statements": [
-        "CREATE STREAM input (id INT KEY, val INT) WITH (kafka_topic='test_topic', value_format='JSON');",
-        "INSERT INTO input (id, val, rowoffset) VALUES (1, 2, 3);"
+        "CREATE STREAM a (id INT KEY, val BIGINT) WITH (kafka_topic='test_topic', value_format='JSON');",
+        "CREATE STREAM b (id INT KEY) WITH (kafka_topic='test_topic_2', value_format='JSON');",
+        "INSERT INTO a SELECT id, ROWOFFSET from b EMIT CHANGES;"
       ],
       "expectedException": {
         "type": "io.confluent.ksql.util.KsqlStatementException",
-        "message": "Failed to insert values into 'INPUT'. Inserting into column `ROWOFFSET` is not allowed."
+        "message": "Reserved column name in select: `ROWOFFSET`. Please remove or alias the column."
       }
     },
     {
--- a/ksqldb-functional-tests/src/test/resources/query-validation-tests/row-offset-and-partition.json
+++ b/ksqldb-functional-tests/src/test/resources/query-validation-tests/row-offset-and-partition.json
@@ -189,32 +189,6 @@
       }
     },
     {
-      "name": "Select ROWPARTITION/ROWOFFSET from CSAS",
-      "properties": {
-        "ksql.rowpartition.rowoffset.enabled": true
-      },
-      "statements": [
-        "CREATE STREAM a (id INT KEY, foo STRING) WITH (kafka_topic='a', value_format='JSON');",
-        "CREATE STREAM b AS SELECT id, CONCAT(foo, 'z') AS foo from a;",
-        "CREATE STREAM c AS SELECT *, ROWPARTITION as rp, ROWOFFSET as ro from b;"
-      ],
-      "inputs": [
-        {"topic":  "a", "key": 1, "value": {"foo": "foo"}, "timestamp": 5},
-        {"topic":  "a", "key": 2, "value": {"foo": "bar"}, "timestamp": 5},
-        {"topic":  "a", "key": 3, "value": {"foo": "baz"}, "timestamp": 5}
-      ],
-      "outputs": [
-        {"topic":  "C", "key": 1, "value": {"foo": "fooz", "RP": 0, "RO": 0}, "timestamp": 5},
-        {"topic":  "C", "key": 2, "value": {"foo": "barz", "RP": 0, "RO": 1}, "timestamp": 5},
-        {"topic":  "C", "key": 3, "value": {"foo": "bazz", "RP": 0, "RO": 2}, "timestamp": 5}
-      ],
-      "post": {
-        "sources": [
-          {"name": "C", "type": "stream", "schema": "id INT KEY, foo STRING, RP INT, RO BIGINT"}
-        ]
-      }
-    },
-    {
       "name": "Select ROWPARTITION/ROWOFFSET with aliases - table",
       "properties": {
         "ksql.rowpartition.rowoffset.enabled": true
@@ -222,32 +196,6 @@
       "statements": [
         "CREATE TABLE input (id INT PRIMARY KEY, foo STRING) WITH (kafka_topic='test_topic', value_format='JSON');",
         "CREATE TABLE output AS SELECT id, ROWPARTITION AS rp, ROWOFFSET AS ro FROM input;"
-      ],
-      "inputs": [
-        {"topic":  "test_topic", "key": 1, "value": {"foo": "a"}, "timestamp": 5},
-        {"topic":  "test_topic", "key": 2, "value": {"foo": "b"}, "timestamp": 6},
-        {"topic":  "test_topic", "key": 3, "value": {"foo": "c"}, "timestamp":  7}
-      ],
-      "outputs": [
-        {"topic":  "OUTPUT", "key": 1, "value": {"RP": 0, "RO": 0}, "timestamp": 5},
-        {"topic":  "OUTPUT", "key": 2, "value": {"RP": 0, "RO": 1}, "timestamp": 6},
-        {"topic":  "OUTPUT", "key": 3, "value": {"RP": 0, "RO": 2}, "timestamp": 7}
-      ],
-      "post": {
-        "sources": [
-          {"name": "OUTPUT", "type": "table", "schema": "ID INT KEY, RP INT, RO BIGINT"}
-        ]
-      }
-    },
-    {
-      "name": "Select ROWPARTITION/ROWOFFSET from CTAS",
-      "properties": {
-        "ksql.rowpartition.rowoffset.enabled": true
-      },
-      "statements": [
-        "CREATE TABLE input (id INT PRIMARY KEY, foo STRING) WITH (kafka_topic='test_topic', value_format='JSON');",
-        "CREATE TABLE queryable AS SELECT id, foo FROM input;",
-        "CREATE TABLE output AS SELECT id, ROWPARTITION AS rp, ROWOFFSET AS ro FROM queryable;"
       ],
       "inputs": [
         {"topic":  "test_topic", "key": 1, "value": {"foo": "a"}, "timestamp": 5},
@@ -392,6 +340,7 @@
         "topics": {
           "blacklist": ".*-repartition"
         },
+        "comments": "The schema shouldn't contain a key column named ROWOFFSET, this is a bug with PARTITION BY. See https://github.com/confluentinc/ksql/issues/8193",
         "sources": [
           {"name": "OUTPUT", "type": "stream", "schema": "ROWOFFSET BIGINT KEY, val INT, id INT"}
         ]
@@ -421,6 +370,7 @@
         "topics": {
           "blacklist": ".*-repartition"
         },
+        "comments": "The schema shouldn't contain a key column named ROWPARTITION, this is a bug with PARTITION BY. See https://github.com/confluentinc/ksql/issues/8193",
         "sources": [
           {"name": "OUTPUT", "type": "stream", "schema": "ROWPARTITION INT KEY, val INT, id INT"}
         ]
--- a/ksqldb-functional-tests/src/test/resources/query-validation-tests/row-offset-and-partition.json
+++ b/ksqldb-functional-tests/src/test/resources/query-validation-tests/row-offset-and-partition.json
@@ -404,7 +404,7 @@
       }
     },
     {
-      "name": "PARTITION BY ROWOFFSET - buggy?",
+      "name": "PARTITION BY ROWOFFSET",
       "properties": {
         "ksql.rowpartition.rowoffset.enabled": true
       },
@@ -432,7 +432,7 @@
       }
     },
     {
-      "name": "PARTITION BY ROWPARTITION - buggy?",
+      "name": "PARTITION BY ROWPARTITION",
       "comments": "Not the most useful PARTITION BY!",
       "properties": {
         "ksql.rowpartition.rowoffset.enabled": true
--- a/ksqldb-functional-tests/src/test/resources/query-validation-tests/row-offset-and-partition.json
+++ b/ksqldb-functional-tests/src/test/resources/query-validation-tests/row-offset-and-partition.json
@@ -1,6 +1,6 @@
 {
   "comments": [
-    "Tests ensuring ROWPARTITON and ROWOFFSET work as user column names before KLIP-50 is integrated"
+    "Tests covering ROWPARTITION and ROWOFFSET pseudocolumns"
   ],
   "tests": [
     {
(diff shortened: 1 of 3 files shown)


ksqldb-functional-tests/src/test/resources/query-validation-tests/row-offset-and-partition.json
```diff
@@ -119,6 +150,622 @@
...
+        {"topic":  "test_topic_2", "key": 6, "value": {"bar": "f"}, "timestamp":  10}
+      ],
+      "outputs": [
+        {"topic":  "OUTPUT", "key": 0, "value": {"foo": "a", "bar": "d","roa": 0, "rob": 0}, "timestamp": 8},
+        {"topic":  "OUTPUT", "key": 0, "value": {"foo":  "b", "bar": "e", "roa": 1, "rob": 1}, "timestamp": 9},
+        {"topic":  "OUTPUT", "key": 0, "value": {"foo":  "c", "bar": "f","roa": 2, "rob": 2}, "timestamp": 10}
+      ],
+      "post": {
+        "sources": [
+          {"name": "OUTPUT", "type": "stream", "schema": "FOO STRING, BAR STRING, ROA BIGINT, ROB BIGINT"}
```
Sullivan-Patrick, line 352: This test fails because ROA is not in the projection, but ROB is. On manual testing, both are in the projection. Any ideas what this might be?

ksqldb-functional-tests/src/test/resources/query-validation-tests/row-offset-and-partition.json
```diff
@@ -119,6 +150,622 @@
...
+      }
+    },
+    {
+      "name": "GROUP BY ROWPARTITION",
+      "properties": {
+        "ksql.rowpartition.rowoffset.enabled": true
+      },
+      "statements": [
+        "CREATE STREAM L (A INT KEY, B INT) WITH (kafka_topic='LEFT', value_format='JSON');",
+        "CREATE TABLE OUTPUT AS SELECT COUNT(B) AS bc, ROWPARTITION AS rp FROM L GROUP BY ROWPARTITION;"
```
Sullivan-Patrick, line 725: These ended up not being buggy like I mentioned they might have been @vcrfxia. Turns out I just needed a fresh look at the tests to realize what I was doing wrong.

ksqldb-functional-tests/src/test/resources/query-validation-tests/row-offset-and-partition.json
```diff
@@ -80,34 +80,65 @@
...
+        "CREATE STREAM input (id INT KEY, val INTEGER) WITH (kafka_topic='test_topic', value_format='JSON');",
+        "CREATE STREAM output AS SELECT ROWOFFSET FROM input EMIT CHANGES;"
+      ],
+      "expectedException": {
+        "type": "io.confluent.ksql.util.KsqlStatementException",
+        "message": "SELECT column 'ROWOFFSET' cannot be resolved."
+      }
+    },
+    {
+      "comments": "Should be removed after rowpartition.rowoffset.enabled feature flag is removed",
```
vcrfxia, line 111: I don't think this test (and the next three) should be removed after the feature flag is removed.

ksqldb-functional-tests/src/test/resources/query-validation-tests/row-offset-and-partition.json
```diff
@@ -80,34 +80,65 @@
       }
     },
     {
+      "name": "verify ROWPARTITION is not queryable with feature disabled",
```
vcrfxia, line 83: I don't think this test and the one below are necessary. This behavior is good to check while developing the feature flag but doesn't feel necessary to check in for posterity. They will be removed once the feature flag is removed anyway.

ksqldb-functional-tests/src/test/resources/query-validation-tests/row-offset-and-partition.json
```diff
@@ -119,6 +150,622 @@
...
+      }
+    },
+    {
+      "name": "INSERT INTO ROWOFFSET should fail",
+      "properties": {
+        "ksql.rowpartition.rowoffset.enabled": true
+      },
+      "statements": [
+        "CREATE STREAM input (id INT KEY, val INT) WITH (kafka_topic='test_topic', value_format='JSON');",
+        "INSERT INTO input (id, val, rowoffset) VALUES (1, 2, 3);"
```
vcrfxia, line 189: This is not the right test framework to test this. Only RQTT supports INSERT VALUES. See insert-values.json.

ksqldb-functional-tests/src/test/resources/query-validation-tests/row-offset-and-partition.json
```diff
@@ -119,6 +150,622 @@
...
+        {"topic":  "OUTPUT", "key": 3, "value": {"RP": 0, "RO": 2, "RT": 7}, "timestamp": 7}
+      ],
+      "post": {
+        "sources": [
+          {"name": "OUTPUT", "type": "stream", "schema": "ID INT KEY, RP INT, RO BIGINT, RT BIGINT"}
+        ]
+      }
+    },
+    {
+      "name": "Select ROWPARTITION/ROWOFFSET from CSAS",
```
vcrfxia, line 222: I don't think this test is necessary beyond the test above. Also, QTT does not support multiple persistent queries in the same test. I think the fact that this test passes is a fluke?

ksqldb-functional-tests/src/test/resources/query-validation-tests/row-offset-and-partition.json
```diff
@@ -119,6 +150,622 @@
...
+        {"topic":  "C", "key": 3, "value": {"foo": "baz", "RP": 0, "RO": 2}, "timestamp": 5}
+      ],
+      "post": {
+        "sources": [
+          {"name": "C", "type": "stream", "schema": "id INT KEY, foo STRING, RP INT, RO BIGINT"}
+        ]
+      }
+    },
+    {
+      "name": "Select ROWPARTITION/ROWOFFSET from CT",
```
vcrfxia, line 248:
```suggestion
      "name": "Select ROWPARTITION/ROWOFFSET with aliases - table",
```
and rename `Select ROWPARTITION and ROWOFFSET columns with aliases` above to `Select ROWPARTITION/ROWOFFSET with aliases - stream` for consistency.

ksqldb-functional-tests/src/test/resources/query-validation-tests/row-offset-and-partition.json
```diff
@@ -119,6 +150,622 @@
...
+        {"topic":  "OUTPUT", "key": 3, "value": {"RP": 0, "RO": 2}, "timestamp": 7}
+      ],
+      "post": {
+        "sources": [
+          {"name": "OUTPUT", "type": "table", "schema": "ID INT KEY, RP INT, RO BIGINT"}
+        ]
+      }
+    },
+    {
+      "name": "Select ROWPARTITION/ROWOFFSET from CTAS",
```
vcrfxia, line 273: Let's remove this, for the same reason as above.

ksqldb-functional-tests/src/test/resources/query-validation-tests/row-offset-and-partition.json
```diff
@@ -119,6 +150,622 @@
...
+        {"topic":  "test_topic_2", "key": 6, "value": {"foo": "f"}, "timestamp":  10}
+      ],
+      "outputs": [
+        {"topic":  "OUTPUT", "key": 1, "value": {"RP": 0, "RO": 0, "RT": 5}, "timestamp": 5},
+        {"topic":  "OUTPUT", "key": 2, "value": {"RP": 0, "RO": 1, "RT": 6}, "timestamp": 6},
+        {"topic":  "OUTPUT", "key": 3, "value": {"RP": 0, "RO": 2, "RT": 7}, "timestamp": 7}
+      ],
+      "post": {
+        "sources": [
+          {"name": "OUTPUT", "type": "stream", "schema": "ID INT KEY, RP INT, RO BIGINT, RT BIGINT"}
```
vcrfxia, line 323: Does this pass? This doesn't look like the output schema based on the CSAS above. Also, the CSAS above doesn't look like valid stream-stream join syntax?

ksqldb-functional-tests/src/test/resources/query-validation-tests/row-offset-and-partition.json
vcrfxia: Is ROA in the key schema instead of the value schema? (Just throwing out a guess.)

ksqldb-functional-tests/src/test/resources/query-validation-tests/row-offset-and-partition.json
```diff
@@ -119,6 +150,622 @@
...
+        ]
+      }
+    },
+    {
+      "name": "join on ROWOFFSET - INCOMPLETE",
```
vcrfxia, line 328: What does "INCOMPLETE" mean in the name here?

ksqldb-functional-tests/src/test/resources/query-validation-tests/row-offset-and-partition.json
```diff
@@ -119,6 +150,622 @@
...
+        {"topic": "OUTPUT", "key": 0, "value": {"ID": 10, "VAL":  5}},
+        {"topic": "OUTPUT", "key": 1, "value": {"ID": 9, "VAL":  6}},
+        {"topic": "OUTPUT", "key": 2, "value": {"ID": 8, "VAL":  7}}
+      ],
+      "post": {
+        "topics": {
+          "blacklist": ".*-repartition"
+        },
+        "sources": [
+          {"name": "OUTPUT", "type": "stream", "schema": "ROWOFFSET BIGINT KEY, val INT, id INT"}
```
vcrfxia, line 430: Why is this allowed? This looks like a user column with name `ROWOFFSET`?

ksqldb-functional-tests/src/test/resources/query-validation-tests/row-offset-and-partition.json
```diff
@@ -119,6 +150,622 @@
...
+        {"topic": "OUTPUT", "key": 0, "value": {"ID": 10, "VAL":  5}},
+        {"topic": "OUTPUT", "key": 0, "value": {"ID": 9, "VAL":  6}},
+        {"topic": "OUTPUT", "key": 0, "value": {"ID": 8, "VAL":  7}}
+      ],
+      "post": {
+        "topics": {
+          "blacklist": ".*-repartition"
+        },
+        "sources": [
+          {"name": "OUTPUT", "type": "stream", "schema": "ROWPARTITION INT KEY, val INT, id INT"}
```
vcrfxia, line 459: Same question as above.

ksqldb-functional-tests/src/test/resources/query-validation-tests/row-offset-and-partition.json
```diff
@@ -119,6 +150,622 @@
...
+        ]
+      }
+    },
+    {
+      "name": "Try to select ROWPARTITION in persistent query without aliases",
```
vcrfxia, line 464:
```suggestion
      "name": "Should fail if select ROWPARTITION without aliases",
```
and similarly below.

ksqldb-functional-tests/src/test/resources/query-validation-tests/row-offset-and-partition.json
```diff
@@ -119,6 +150,622 @@
...
+        {"topic":  "OUTPUT", "key": 4, "value": {"FOO":  "d", "RT": 8}, "timestamp": 8}
+      ],
+      "post": {
+        "sources": [
+          {"name": "OUTPUT", "type": "stream", "schema": "ID INT KEY, FOO STRING, RT BIGINT"}
+        ]
+      }
+    },
+    {
+      "name": "Filter and select using ROWPARTITION",
```
vcrfxia, line 542: I don't think this test is necessary, given that filter and select are each already tested separately. Same for the other test below.

ksqldb-functional-tests/src/test/resources/query-validation-tests/row-offset-and-partition.json
```diff
@@ -119,6 +150,622 @@
...
+        {"topic":  "OUTPUT", "key": 4, "value": {"FOO":  "d", "RP": 0, "RO": 3, "RT": 8}, "timestamp": 8}
+      ],
+      "post": {
+        "sources": [
+          {"name": "OUTPUT", "type": "stream", "schema": "ID INT KEY, FOO STRING, RT BIGINT, RP INT, RO BIGINT"}
+        ]
+      }
+    },
+    {
+      "name": "GREATEST UDF with ROWPARTITION",
```
vcrfxia, line 592: I don't think this and the other two tests (for UDFs) below are necessary, given that we're already testing `CAST()` above.

ksqldb-functional-tests/src/test/resources/query-validation-tests/row-offset-and-partition.json
```diff
@@ -119,6 +150,622 @@
...
+      "statements": [
+        "CREATE STREAM L (A INT KEY, B INT) WITH (kafka_topic='LEFT', value_format='JSON');",
+        "CREATE TABLE OUTPUT AS SELECT COUNT(B) AS bc, ROWPARTITION AS rp FROM L GROUP BY ROWPARTITION;"
+      ],
+      "inputs": [
+        {"topic": "LEFT", "key": 1, "value": {"B": 100}, "timestamp": 11},
+        {"topic": "LEFT", "key": 2, "value": {"B": 100}, "timestamp": 11},
+        {"topic": "LEFT", "key": 3, "value": {"B": 100}, "timestamp": 11}
+      ],
+      "comments": "the schema is wrong here, this may be an issue with the QTT framework not having great support for partition? In manual testing, RP would be part of the output",
```
vcrfxia, line 732: Is this a problem or not a problem?

ksqldb-functional-tests/src/test/resources/query-validation-tests/row-offset-and-partition.json
```diff
@@ -119,6 +150,622 @@
...
+      "statements": [
+        "CREATE STREAM L (A INT KEY, B INT) WITH (kafka_topic='LEFT', value_format='JSON');",
+        "CREATE TABLE OUTPUT AS SELECT COUNT(B) as bc, ROWOFFSET AS ro FROM L GROUP BY ROWOFFSET;"
+      ],
+      "inputs": [
+        {"topic": "LEFT", "key": 1, "value": {"B": 1}, "timestamp": 11},
+        {"topic": "LEFT", "key": 1, "value": {"B": 5}, "timestamp": 11},
+        {"topic": "LEFT", "key": 1, "value": {"B": 10}, "timestamp": 11}
+      ],
+      "comments": "the schema is wrong here, this may be an issue with the QTT framework not having great support for partition? In manual testing, RP would be part of the output",
```
vcrfxia, line 758: Same question as above.

ksqldb-functional-tests/src/test/resources/rest-query-validation-tests/push-queries.json
```diff
@@ -624,6 +704,29 @@
         ]}
       ]
     },
+    {
+      "name": "key, ROWPARTITION, ROWOFFSET in projection",
```
vcrfxia, line 708: I don't think this test adds anything beyond "explicit ROWPARTITION and ROWOFFSET" above. Let's remove it since RQTTs are expense.

vcrfxia (commented):
Thanks for the test coverage @Sullivan-Patrick ! Comments inline.

ksqldb-functional-tests/src/test/resources/query-validation-tests/row-offset-and-partition.json
Sullivan-Patrick: I'm confused- I thought we made the distinction between INSERT INTO and INSERT VALUES? https://github.com/confluentinc/ksql/pull/8121#discussion_r707605438

vcrfxia: The test you've added here is for INSERT VALUES, not INSERT INTO.
https://docs.ksqldb.io/en/latest/developer-guide/ksqldb-reference/insert-values/
https://docs.ksqldb.io/en/latest/developer-guide/ksqldb-reference/insert-into/

ksqldb-functional-tests/src/test/resources/query-validation-tests/row-offset-and-partition.json
Sullivan-Patrick: I thought the idea is that we change the non `feature-enabled` flags to copy this behavior, then remove these tests?

https://github.com/confluentinc/ksql/pull/8120#discussion_r712298315

vcrfxia: Oh I see. I didn't realize there was another copy of this test. In that case we need to remove the comments that say to remove the other copy, else we've got comments saying that both copies of the test should be removed: https://github.com/confluentinc/ksql/blob/e8d5132c5503ec25d579082397c67d7d24d6fcba/ksqldb-functional-tests/src/test/resources/query-validation-tests/row-offset-and-partition.json#L7

ksqldb-functional-tests/src/test/resources/query-validation-tests/row-offset-and-partition.json
Sullivan-Patrick: That was indeed the problem. Is this expected behavior? Seems a little strange.

ksqldb-functional-tests/src/test/resources/query-validation-tests/row-offset-and-partition.json
Sullivan-Patrick: I think I removed these comments in the still-unmerged pull query PR. I removed said removal of the comments when merging in master to that branch today so I could put the changes on this branch, so this will indeed be necessary to do again.

ksqldb-functional-tests/src/test/resources/query-validation-tests/row-offset-and-partition.json
Sullivan-Patrick: I made some changes to this test which seem to imply that QTTs are supporting multiple queries to me- I'll push the changes to sanity check.

ksqldb-functional-tests/src/test/resources/query-validation-tests/row-offset-and-partition.json
Sullivan-Patrick: No, neither of these passed, both were incomplete when I pushed the changes. These were the queries I pushed to ask you about- I've got them working now though.

ksqldb-functional-tests/src/test/resources/query-validation-tests/row-offset-and-partition.json
Sullivan-Patrick: They weren't working and hadn't finished them- this could have been better communicated via a GitHub comment, but I thought I relayed to you I had some nonfunctioning tests that I pushed? Sorry if I didn't do that or didn't make it clear enough that these were the queries in question, I'll comment the titles in this situation next time instead of writing it into the code.

ksqldb-functional-tests/src/test/resources/query-validation-tests/row-offset-and-partition.json
Sullivan-Patrick: Fresh look tells me this isn't a problem, just a silly error by me.

ksqldb-functional-tests/src/test/resources/query-validation-tests/row-offset-and-partition.json
vcrfxia: It's expected. ROA is the join key.

ksqldb-functional-tests/src/test/resources/query-validation-tests/row-offset-and-partition.json
vcrfxia: It's definitely a bug in that we shouldn't be allowing this, but evidently it's an existing bug if it's present with ROWTIME too, so it's not the end of the world if we don't fix it. Can you file a Github issue, and leave a comment linking to the issue in the relevant QTTs?

ksqldb-functional-tests/src/test/resources/query-validation-tests/row-offset-and-partition.json
vcrfxia: Sorry my previous comment was confusing -- QTTs with multiple persistent queries are not supported by the historic plan feature of the QTT framework (only the plan for the last persistent query is actually recorded). Execution of multiple persistent queries is fine, but QTT really isn't meant for that. I still think we should remove this test, given the limited value (in terms of additional coverage) and the fact that QTT does not support multiple persistent queries.

Sullivan-Patrick: Ah, that clears things up. SGTM.

ksqldb-functional-tests/src/test/resources/query-validation-tests/row-offset-and-partition.json
Sullivan-Patrick: My thoughts exactly, fixing it would be fun, but I don't think I have capacity in my last 3 days.

ksqldb-functional-tests/src/test/resources/query-validation-tests/row-offset-and-partition.json
```diff
@@ -119,6 +118,465 @@
...
+        {"topic":  "OUTPUT", "key": 4, "value": {"FOO":  "d", "RT": 8}, "timestamp": 8}
+      ],
+      "post": {
+        "sources": [
+          {"name": "OUTPUT", "type": "stream", "schema": "ID INT KEY, FOO STRING, RT BIGINT"}
+        ]
+      }
+    },
+    {
+      "name": "AVG UDAF with ROWOFFSET",
```
vcrfxia, line 458: Remove this? A previous comment discussed why.

vcrfxia (approved):
Thanks @Sullivan-Patrick -- LGTM besides the comment inline.

PR #8187, opened 2021-09-28
chore: changed queries which contain ROWPARTITION/ROWOFFSET to not trigger spq condition

Sullivan-Patrick:
### Description 
Scalable push queries currently piggyback on the topology of their source query, and therefore do not have access to partition and offset information of the result. This will likely change soon, but until then, we need to make it so push queries involving these new pseudocolumns are not determined to be scalable push queries.

### Testing done 
Added unit tests relevant to new behavior

### Reviewer checklist


ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/util/ScalablePushUtil.java
```diff
@@ -82,7 +88,45 @@ public static boolean isScalablePushQuery(
         // Must be reading from "latest"
         && isLatest
         // We only handle a single sink source at the moment from a CTAS/CSAS
-        && upstreamQueries.size() == 1;
+        && upstreamQueries.size() == 1
+        // ROWPARTITION and ROWOFFSET are not currently supported in SPQs
+        && !containsPartitionOrOffset(query, ksqlConfig);
+  }
+
+  private static boolean containsPartitionOrOffset(final Query query, final KsqlConfig ksqlConfig) {
```
vcrfxia, line 96: Let's not bake "partition" or "offset" into any of these method names. Doing so defeats the purpose of writing the methods `isDisallowedInScalablePushQueries()` in a more generic way.

ksqldb-common/src/main/java/io/confluent/ksql/schema/ksql/SystemColumns.java
```diff
@@ -197,34 +209,40 @@ private static void validatePseudoColumnVersion(final int pseudoColumnVersionNum
...
-        final boolean isDisallowedForInsertValues
+        final boolean isDisallowedForInsertValues,
+        final boolean isDisallowedInScalablePushQueries
     ) {
       this.name = requireNonNull(name, "name");
       this.type = requireNonNull(type, "type");
       this.version = version;
       this.mustBeMaterializedForTableJoins = mustBeMaterializedForTableJoins;
       this.isDisallowedForInsertValues = isDisallowedForInsertValues;
+      this. isDisallowedInScalablePushQueries = isDisallowedInScalablePushQueries;
```
vcrfxia, line 227: nit: spacing.

ksqldb-common/src/main/java/io/confluent/ksql/schema/ksql/SystemColumns.java
```diff
@@ -197,34 +209,40 @@ private static void validatePseudoColumnVersion(final int pseudoColumnVersionNum
...
 
     private static PseudoColumn of(
         final ColumnName name,
         final SqlType type,
         final int version,
         final boolean mustBeMaterializedForTableJoins,
-        final boolean isDisallowedForInsertValues
+        final boolean isDisallowedForInsertValues,
+        final boolean isDisallowedInScalablePushQueries
+
```
vcrfxia, line 237: nit: whitespace change.

vcrfxia (commented):
Thanks @Sullivan-Patrick -- LGTM at a high level, would be great to see tests (as discussed offline).

ksqldb-common/src/main/java/io/confluent/ksql/schema/ksql/SystemColumns.java
```diff
@@ -197,34 +209,40 @@ private static void validatePseudoColumnVersion(final int pseudoColumnVersionNum
...
+    final boolean isDisallowedInScalablePushQueries;
 
     private PseudoColumn(
         final ColumnName name,
         final SqlType type,
         final int version,
         final boolean mustBeMaterializedForTableJoins,
-        final boolean isDisallowedForInsertValues
+        final boolean isDisallowedForInsertValues,
+        final boolean isDisallowedInScalablePushQueries
```
AlanConfluent, line 220: If we're going to add this, I might add a similar thing for pull queries and then change how I asked for the logical schema to utilize this rather than pin pull queries to an older version of the set of pseudo columns,

AlanConfluent (commented):
Looks good to me.  As you had mentioned, we should probably introduce a file `ScalablePushUtilTest` and test your change to that check.

ksqldb-common/src/main/java/io/confluent/ksql/schema/ksql/SystemColumns.java
Sullivan-Patrick: What do you think is the best way of doing this?

Also, to sanity check, we'll likely remove the disallowing of these pseudocolumns pretty soon, once we start using a KafkaConsumer instead of the piggybacked processor, right? There might not be a conflict with newer pseudocolumn versions- of course, if plans change or something unforseen comes up, having a more flexible solution that doesn't rely on eventually allowing SPQs to be compatible  with ROWPARTITION/ROWOFFSET will be better.

ksqldb-rest-app/src/test/java/io/confluent/ksql/rest/util/ScalablePushUtilTest.java
```diff
@@ -33,25 +58,138 @@
...
+          overrides
+      );
+
+      // Then:
+      assert(!isScalablePush);
+    }
+  }
+
+  @Test
+  public void shouldMakeCompatibleQueryScalablePush() {
```
vcrfxia, line 176: I don't think we need this test. Conceptually it's the same as `isScalablePushQuery_true()`. It's nice to know that the `ColumnExtractor` mock doesn't break the testing but once we've confirmed that for ourselves, I don't think we need to check this in.

ksqldb-rest-app/src/test/java/io/confluent/ksql/rest/util/ScalablePushUtilTest.java
```diff
@@ -252,7 +390,60 @@ public void isScalablePushQuery_false_wrongUpstreamQueries_Two() {
...
+    when(query.getPartitionBy()).thenReturn(Optional.empty());
+    when(query.getRefinement())
+        .thenReturn(Optional.of(RefinementInfo.of(OutputRefinement.CHANGES)));
+    when(query.getFrom())
+        .thenReturn(new AliasedRelation(new Table(SourceName.of("Foo")), SourceName.of("blah")));
+    when(ksqlEngine.getQueriesWithSink(SourceName.of("Foo"))).thenReturn(
+        ImmutableSet.of(new QueryId("a")));
+  }
+
+  private void givenScalablePushQuery() {
```
vcrfxia, line 411: How is this different from `expectIsSQP()`? Let's de-dup (especially since the names as-is are rather confusing).

vcrfxia (commented):
Thanks @Sullivan-Patrick ! Comments inline.

ksqldb-rest-app/src/test/java/io/confluent/ksql/rest/util/ScalablePushUtilTest.java
Sullivan-Patrick: Maybe we should bring in @AlanConfluent to talk about this since it's his code.

ksqldb-rest-app/src/test/java/io/confluent/ksql/rest/util/ScalablePushUtilTest.java
Sullivan-Patrick: This presents an interesting issue- if we enable the feature flag, we need to mock the static `ColumnExtractor` using a try block, and this needs to be done in every test (unless there is a better way to mock a classes' static behavior that I'm not aware of, which is a strong possibility).

One workaround is disabling the flag for all tests which aren't testing the last condition, so true is immediately returned from `containsDisallowedColumns()` and it never tries to use the `ColumnExtractor`. Seems a bit hacky though, WDYT?

vcrfxia: How about enabling the feature flag just for the relevant tests? The default (for the rest of the tests) can be to leave it off.

ksqldb-rest-app/src/test/java/io/confluent/ksql/rest/util/ScalablePushUtilTest.java
vcrfxia: Alan's welcome to chime in but I think the question of whether this newly added test is valuable is separate from the test coverage he added in his PR? Or maybe I'm misunderstanding the question.

ksqldb-rest-app/src/test/java/io/confluent/ksql/rest/util/ScalablePushUtilTest.java
vcrfxia: Ah I see you've already done this in your recent commit. Is there still an open question here?

ksqldb-rest-app/src/test/java/io/confluent/ksql/rest/util/ScalablePushUtilTest.java
Sullivan-Patrick: Sorry, I didn't recognize this was my own test.. You're right.

ksqldb-rest-app/src/test/java/io/confluent/ksql/rest/util/ScalablePushUtilTest.java
Sullivan-Patrick: Nope, just wanted to check if that was a reasonable course of action.

ksqldb-rest-app/src/test/java/io/confluent/ksql/rest/util/ScalablePushUtilTest.java
```diff
@@ -2,257 +2,409 @@
...
     assertThat(ScalablePushUtil.isScalablePushQuery(query, ksqlEngine, ksqlConfig,
-        ImmutableMap.of(ConsumerConfig.AUTO_OFFSET_RESET_CONFIG, "latest")),
+            ImmutableMap.of(ConsumerConfig.AUTO_OFFSET_RESET_CONFIG, "latest")),
         equalTo(false));
   }
-}
+
+  private void expectIsSPQ() {
+    when(ksqlConfig.getBoolean(KsqlConfig.KSQL_QUERY_PUSH_V2_ENABLED)).thenReturn(true);
+    // to avoid static mocking of ColumnExtractor for each class
```
vcrfxia, line 372: This looks like it was meant to be removed with the latest change?

vcrfxia (approved):
Thanks @Sullivan-Patrick -- LGTM besides one minor thing inline. Let's merge this PR when the build passes, and you can pick up that fix in one of your follow-up PRs instead 🎉

PR #8189, opened 2021-09-28
fix: allow more time for consumer group cleanup

vvcephei:
During local testing, I've noticed that consumer group cleanup for my transient queries 
is frequently timing out and leaving defunct groups lying around. This patch increases the amount
of time to wait for the group to become empty.

### Reviewer checklist

### DIFF
--- a/ksqldb-engine/src/main/java/io/confluent/ksql/services/KafkaConsumerGroupClientImpl.java
+++ b/ksqldb-engine/src/main/java/io/confluent/ksql/services/KafkaConsumerGroupClientImpl.java
@@ -25,7 +25,6 @@ import java.util.Collections;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
-import java.util.concurrent.atomic.AtomicInteger;
 import java.util.function.Supplier;
 import java.util.stream.Collectors;
 import org.apache.kafka.clients.admin.Admin;
@@ -108,7 +107,6 @@ public class KafkaConsumerGroupClientImpl implements KafkaConsumerGroupClient {
 
   @Override
   public void deleteConsumerGroups(final Set<String> groups) {
-    final AtomicInteger retryCount = new AtomicInteger(0);
     try {
       // it takes heartbeat.interval.ms after a consumer is closed for the broker
       // to recognize that there are no more consumers in the consumer group - for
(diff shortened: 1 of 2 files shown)


ksqldb-engine/src/main/java/io/confluent/ksql/services/KafkaConsumerGroupClientImpl.java
```diff
@@ -117,10 +115,11 @@ public void deleteConsumerGroups(final Set<String> groups) {
       ExecutorUtil.executeWithRetries(
           () -> adminClient.get().deleteConsumerGroups(groups).all().get(),
           e -> (e instanceof RetriableException)
-              || (e instanceof GroupNotEmptyException && retryCount.getAndIncrement() < 5),
-          () -> Duration.of(3, ChronoUnit.SECONDS)
+              || (e instanceof GroupNotEmptyException),
+          (retry) -> Duration.of(3L * retry, ChronoUnit.SECONDS),
```
vvcephei, line 119: First change: add linear backoff so that we will wait 3, 6, 9, and so forth seconds between each retry instead of just waiting 3 seconds every time.

ksqldb-engine/src/main/java/io/confluent/ksql/services/KafkaConsumerGroupClientImpl.java
```diff
@@ -117,10 +115,11 @@ public void deleteConsumerGroups(final Set<String> groups) {
...
+          10
```
vvcephei, line 120: Second change: allow more retries. It was previously hard-coded in ExecutorUtil to 5, now it will be 10. So we would previously only wait up to 12 seconds before throwing an exception, and now we will wait for 135 seconds.

ksqldb-engine/src/main/java/io/confluent/ksql/services/KafkaConsumerGroupClientImpl.java
nateab: Just for my own curiosity, how did we come up with using 3 seconds?

nateab (approved):
LGTM!

ksqldb-engine/src/main/java/io/confluent/ksql/services/KafkaConsumerGroupClientImpl.java
vvcephei: There's a comment there that says that it's the default heartbeat interval. I'm not sure if the person who made this choice was aware that we're actually waiting for the heartbeat interval + the session timeout, though.

Come to think of it, I bet that this thing has just recently started timing out as another casualty of the KIP to change the default session interval to 30 seconds.

nateab: Ah i only saw the first half of the comment so I missed the explanation 😅  And that makes sense, good catch!

vvcephei:
Thanks, @nateab !

ksqldb-engine/src/main/java/io/confluent/ksql/services/KafkaConsumerGroupClientImpl.java
agavra: > I'm not sure if the person who made this choice was aware that we're actually waiting for the heartbeat interval + the session timeout, though.

hehe, guilty

agavra (approved):

ksqldb-engine/src/main/java/io/confluent/ksql/services/KafkaConsumerGroupClientImpl.java
vvcephei: Aha! Smoked you out!

Anyway, I think it was fine until we bumped that default session interval up so high.

PR #8191, opened 2021-09-29
feat: turn on feature flag for KLIP-50

Sullivan-Patrick:
### Description 
Currently a draft PR just to check build as I'm having some test failures locally. Still definitely need to merge my test fix ticket and a bugfix by @AlanConfluent before the build will be green.

### Testing done 

### Reviewer checklist


vcrfxia (commented):
Thanks @Sullivan-Patrick -- I only reviewed two of the five commits (not the two merge commits or the commit for generating new historic plans). LGTM besides the minor things inline and two more things:
- we should update JoinNodeTest.java to stop explicitly setting the feature flag to false (we want our test coverage to match the default, which is now for the feature to be enabled)
- we should update PlannedTestUtils.java to start requiring historic QTTs for tests that enable the feature flag, as we want these tests to be part of the main test coverage now.

ksqldb-rest-app/src/test/java/io/confluent/ksql/rest/util/ScalablePushUtilTest.java
```diff
@@ -45,47 +44,39 @@
...
   @Mock
   private Select select;
+  @Mock
+  private KsqlConfig ksqlConfig;
 
   @Before
   public void setUp() {
-    // Otherwise we'd need to mock static ColumnExtractor in each passing test to avoid NPEs
-    when(ksqlConfig.getBoolean(KsqlConfig.KSQL_ROWPARTITION_ROWOFFSET_ENABLED)).thenReturn(false);
+    when(ksqlConfig.getBoolean(KsqlConfig.KSQL_ROWPARTITION_ROWOFFSET_ENABLED)).thenReturn(true);
```
vcrfxia: This is the default now, right? Can we omit this?

ksqldb-rest-app/src/test/java/io/confluent/ksql/rest/util/ScalablePushUtilTest.java
```diff
@@ -45,47 +44,39 @@
 @RunWith(MockitoJUnitRunner.class)
 public class ScalablePushUtilTest {
 
-  private static final Map<String, Object> overrides = ImmutableMap.of(
-      KsqlConfig.KSQL_QUERY_PUSH_V2_ENABLED, true,
```
vcrfxia: Why is this being removed?

ksqldb-functional-tests/src/test/resources/query-validation-tests/partition-by.json
```diff
@@ -279,11 +277,7 @@
       }
     },
     {
-      "comments": "Should be removed with rowpartition.rowoffset.enabled feature flag",
```
vcrfxia: I'm confused -- why is this comment being removed if the test is still here? Same question for below.

vcrfxia (commented):
Thanks @Sullivan-Patrick ! Some minor questions inline.

ksqldb-functional-tests/src/test/resources/query-validation-tests/partition-by.json
vcrfxia: If we remove the test(s), let's also make sure to remove any historic QTTs that may have been generated for them.

ksqldb-rest-app/src/test/java/io/confluent/ksql/rest/util/ScalablePushUtilTest.java
Sullivan-Patrick: It wasn't being used anywhere AFAICT.

vcrfxia: Let's leave it in since conceptually it's important -- that feature flag enables scalable push queries. Otherwise this entire test file is kind of moot.

Sullivan-Patrick: This is in `expectIsSPQ()` and wasn't being used in any of Alan's tests- we're just passing it in through the main config, not the override. Given that, do you still think it's worth putting in? I'd want to understand the utility better before adding it.

ksqldb-rest-app/src/test/java/io/confluent/ksql/rest/util/ScalablePushUtilTest.java
Sullivan-Patrick: For posterity: we communicated and Victoria didn't realize I was using a mocked config- I'll correct behavior of relevant tests to use non-mocked configs or return true as necessary.

ksqldb-engine/src/test/java/io/confluent/ksql/planner/plan/QueryProjectNodeTest.java
```diff
@@ -72,6 +72,24 @@
       .valueColumn(K, SqlTypes.STRING)
       .build();
 
+  private static final LogicalSchema INTERMEDIATE_SCHEMA = LogicalSchema.builder()
```
Sullivan-Patrick, line 75: I know this is super hacky, but I have like an hour left of my internship. @AlanConfluent WDYT about this workaround to the tests? It's related to your schema changes for SPQs and pull queries.

PR #8192, opened 2021-09-29
test: fix race condition in PushPhysicalPlanTest

mjsax:
### Description 
Observed a failing test run with error: `Cannot subscribe to failed publisher.`

The reason is a race condition because the subscriber subscribes to the publisher after the publisher background thread is started. This PR changes the order to register the subscriber before we start the publisher thread.

### Testing done 
Using a breakpoint on the old `publisher.subscribe()` line allows to 100% reproduce the race. Thus no additional testing is needed as we call `publisher.subscribe()` earlier now (and setting a breakpoint at the new line verifies that the issue is resolved).

### Reviewer checklist

### DIFF
--- a/ksqldb-engine/src/main/java/io/confluent/ksql/physical/scalablepush/PushPhysicalPlan.java
+++ b/ksqldb-engine/src/main/java/io/confluent/ksql/physical/scalablepush/PushPhysicalPlan.java
@@ -30,7 +30,6 @@ import java.util.Optional;
 import org.reactivestreams.Subscriber;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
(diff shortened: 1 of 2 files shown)


nateab (approved):
LGTM!

AlanConfluent:
One alternative is to remove the check for `isFailed` in `subscribe` in `BasePublisher` and let it fall back into `doSubscribe` where it does the check again, but this time it is willing to send the error to the subscriber rather than throw an exception immediately.  This way, it would always succeed in subscribing, but would immediately send an error if it was already failed.  That way, an error is always handled the same way.  The subscriber shouldn't have to handle the error differently, depending on a race, in my opinion.  (This is similar to how CompletableFutures behave).

What do you think @mjsax ?

AlanConfluent (approved):

PR #8202, opened 2021-09-30
fix: make RQTT test insensitive to row ordering

patrickstuedi:
### Description 
Currently RQTT tests are sensitive to the order of the output rows, which makes it difficult to run the tests deterministically in case of concurrency or multiple partitions. This PR makes RQTT tests insensitive to row ordering. 

### Testing done 
Changed order of output rows for various tests and made sure tests pass regardless of the row order. 

### Reviewer checklist


ksqldb-functional-tests/src/test/java/io/confluent/ksql/test/rest/RestTestExecutor.java
```diff
@@ -768,12 +777,19 @@ private static void matchResponseFields(
...
-          assertThat("Response mismatch at " + newPath, actualValue, is(expectedValue));
+          incrementKey(actualRecords, actualValue.toString());
+          incrementKey(expectedRecords, expectedValue.toString());
         }
       }
     }
   }
 
+  private static void incrementKey(final HashMap<String, Integer> records, final String key){
+    int val = records.containsKey(key) ? records.get(key) : 0;
```
vpapavas, line 788: Nit: 
```
records.putIfAbsent(key,0);
records.put(key,  map.get(key) + 1);
```
Apparently, you can do this in one line as well:
```
map.merge(records, 1, Integer::sum);
```

ksqldb-functional-tests/src/test/java/io/confluent/ksql/test/rest/RestTestExecutor.java
```diff
@@ -729,12 +725,23 @@ private static void threadYield() {
     }
   }
 
+  private static void verifyResponseFields(final HashMap<String, Integer> actualRecords, final HashMap<String, Integer> expectedRecords, final int idx){
```
vpapavas, line 728: Just to understand: You count how many times you have seen each row in these maps?
If so, shouldn't you check that the maps have the same size and that also `expectedRecords` is contained in `actualRecords`?

patrickstuedi: yes that assertion is still there, line 875: 

`      assertThat(
          "row count mismatch."
              + System.lineSeparator()
              + "Expected: "
              + expectedRows.stream()
              .map(Object::toString)
              .collect(Collectors.joining(INDENT, INDENT, ""))
              + System.lineSeparator()
              + "Got: "
              + rows.stream()
              .map(Object::toString)
              .collect(Collectors.joining(INDENT, INDENT, ""))
              + System.lineSeparator(),
          rows,
          hasSize(expectedRows.size())
      );`

ksqldb-functional-tests/src/test/java/io/confluent/ksql/test/rest/RestTestExecutor.java
patrickstuedi: Nice one liner!

vpapavas (approved):
Maybe add a comment describing why the change and what it achieves for others to know. Other than that, LGTM! Thank you @patrickstuedi!

ksqldb-functional-tests/src/test/java/io/confluent/ksql/test/rest/RestTestExecutor.java
patrickstuedi: Thanks, added a description above.

patrickstuedi:
Thanks @vpapavas, I added a short description above.

ksqldb-functional-tests/src/test/java/io/confluent/ksql/test/rest/RestTestExecutor.java
```diff
@@ -768,7 +789,14 @@ private static void matchResponseFields(
           assertThat("Response mismatch at " + newPath,
               (String) actualValue, startsWith((String) expectedValue));
         } else {
-          assertThat("Response mismatch at " + newPath, actualValue, is(expectedValue));
+          final String reason = "Response mismatch at " + newPath;
+          if (newPath.contains("query") && newPath.contains("row") && expectedKey == "columns"){
```
AlanConfluent, line 793: Nit: Maybe make some constants  for the field names?

ksqldb-functional-tests/src/test/java/io/confluent/ksql/test/rest/RestTestExecutor.java
```diff
@@ -759,7 +777,10 @@ private static void matchResponseFields(
...
             statements,
             idx,
-            newPath
+            newPath,
+            actualRecords,
+            expectedRecords,
+            pathIndex
         );
       } else {
         if (operator == MatchOperator.STARTS_WITH) {
```
AlanConfluent, line 786: Do we want to add some operator to add to the query name or something like that, which will enable this behavior on a per test basis or do we just always want it to behave this way? I think we probably care about ordering in some cases, like push queries where at least in some scenarios there a well defined ordering.  The same is true of pull queries, and especially will be when we add a ORDER BY clause.  Or it could be inverted to be order agnostic unless specified.

AlanConfluent (commented):
Awesome change.  We probably want to at least have some ability to choose when we care or don't care about ordering,

ksqldb-functional-tests/src/test/java/io/confluent/ksql/test/rest/RestTestExecutor.java
patrickstuedi: Good point. 

I guess if as a next step we want to [parallelize partition fetching ](https://github.com/confluentinc/ksql/pull/8161) we would need to enable it by default (otherwise pretty much all current functional tests that don't have partitions=1 will fail) but add an option to disable it in case ordering is needed. Can you explain what you meant by adding an operator? Alternatively do you think it would be possible to control this via some internal property that can be set in the functional test in the property section? e.g.,

```
      "properties": {
        "ksql.rqtt.ordering.enforce":  true
      }
```

patrickstuedi: Added OutputConditions to control whether or not ordering is enforced, thanks for the suggestion @AlanConfluent . Default is ordering is not enforced.

ksqldb-functional-tests/src/test/java/io/confluent/ksql/test/rest/RestTestExecutor.java
```diff
@@ -181,7 +182,9 @@ void buildAndExecuteQuery(final RestTestCase testCase) {
           .build();
 
       verifyOutput(testCase);
-      verifyResponses(responses, testCase.getExpectedResponses(), testCase.getStatements());
+      final boolean verifyOutput = testCase.getOutputConditions().isPresent()
```
AlanConfluent, line 185: Nit: verifyOrder?

AlanConfluent (approved):

PR #8208, opened 2021-09-30
fix: close command runner when command topic is deleted

lct45:
### Description 
We were seeing some command topics go into a `degraded` state after a cluster was terminated because the command runner deleted the command topic but doesn't close itself. This has the command runner `close` right after seeing a `terminate` command so we don't get spurious alerts.

### Testing done 
unit test and manual test.

manual test:
created ksql cluster, rolled to this image
issued a terminate command
`
curl -X "POST" "endpoint/ksql/terminate" \
     -H "Content-Type: application/vnd.ksql.v1+json; charset=utf-8" \
     -d $'{
}' -v --http2 -H "Authorization: Basic $AUTH"
`
saw the following in datadog:
image

### Reviewer checklist


stevenpyzhang (approved):
lgtm

PR #8213, opened 2021-10-01
uppercase fix 2

lihaosky:
- fix: Add policy config to control how to translate field name
- fix: pass policy to translator and more tests


lihaosky:
It's not fully wired up. I have following up PRs to enable them. I can add or modify integration test there. Let me rename to chore. Good to know fix is user-visible.

Issue #7701, opened 2021-06-18
Add metric to report active connections/sessions


mikebin:
**Is your feature request related to a problem? Please describe.**
Not a problem, but a requested improvement to observability.

**Describe the solution you'd like**
To better understand and monitor use of a ksqlDB cluster, it would be useful to have metric(s) to report the number of active client connections/sessions, similar to how other databases provide this information. While this might not be especially beneficial for persistent queries (which are detached from a specific user session), it can be helpful for interactive push/pull queries, especially now that pull queries can execute more expensive table scans.

As a follow-on to this, it might be useful to provide a way for administrators to terminate client sessions if needed.

vvcephei:
Hey, thanks for this request, @mikebin !

Matrics would indeed be nice, but I wonder if the thing that DBAs really need is a way to list all the running queries and to be able to terminate them. In my prior experience, we used MySQL, and I used the PROCESSLIST in conjunction with KILL for this purpose (https://www.devart.com/dbforge/mysql/studio/show-running-queries-in-processlist.html is a nice brief overview).

In particular, it's nice to be able to see the client who issued the query, the time the query has been executing, and what the query is actually doing.

mikebin:
Yes, that makes sense, and I think that would provide even more useful functionality than just metric(s).

PR #8217, opened 2021-10-01
fix: don't register stream pull queries as push

vvcephei:
Previously, stream pull queries would be included in the `show queries;` output,
but they would be listed as "push" queries. Also, they would be counted against
the push query concurrency limit. Both of those behaviors are unintended
consequences of the implementation details (that stream pull queries are
currently specialized transient push queries).

This patch fixes the situation by simply not registering stream pull queries.
I will file a separate issue to register them properly, as pull queries, so they
will be listed appropriately and not count against the push query limit.

### Reviewer checklist

### DIFF
--- a/ksqldb-engine/src/main/java/io/confluent/ksql/query/QueryRegistryImpl.java
+++ b/ksqldb-engine/src/main/java/io/confluent/ksql/query/QueryRegistryImpl.java
@@ -153,8 +153,7 @@ public class QueryRegistryImpl implements QueryRegistry {
       final LogicalSchema schema,
       final OptionalInt limit,
       final Optional<WindowInfo> windowInfo,
-      final boolean excludeTombstones,
-      final Optional<ImmutableMap<TopicPartition, Long>> endOffsets) {
+      final boolean excludeTombstones) {
     // CHECKSTYLE_RULES.ON: ParameterNumberCheck
     final QueryBuilder queryBuilder = queryBuilderFactory.create(
           config,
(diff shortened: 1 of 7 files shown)


ksqldb-engine/src/main/java/io/confluent/ksql/query/QueryRegistryImpl.java
```diff
@@ -176,12 +175,59 @@ public TransientQueryMetadata createTransientQuery(
...
+        limit,
+        windowInfo,
+        excludeTombstones,
+        new ListenerImpl(),
+        new StreamsBuilder(),
+        Optional.of(endOffsets)
+    );
+    query.initialize();
+    // We don't register it as a transient query, so it won't show up in `show queries;`,
+    // nor will it count against the push query limit.
```
vvcephei, line 227: Compare to L182 when we create a transient query, in which we do register it.

This omission right here is why we pushed the stream-pull/transient query split farther down. Since we did that, we were also able to drop a couple of levels of optional end offsets, since stream pull queries always specify them and transient queries never do.

ksqldb-engine/src/main/java/io/confluent/ksql/query/QueryRegistryImpl.java
```diff
@@ -370,8 +416,10 @@ private void registerTransientQuery(
       final MetaStore metaStore,
       final TransientQueryMetadata query
   ) {
-    // Initialize the query before it's exposed to other threads via {@link allLiveQueries}.
-    query.initialize();
+    if (!query.isInitialized()) {
+      throw new IllegalStateException("Transient query must be initialized before it might"
+          + " be exposed to other threads via allLiveQueries");
+    }
```
vvcephei, line 422: We just needed to initialize before we register, it wasn't a requirement to actually do it in the same method.

ksqldb-engine/src/main/java/io/confluent/ksql/query/QueryRegistryImpl.java
```diff
@@ -153,8 +153,7 @@ public TransientQueryMetadata createTransientQuery(
       final LogicalSchema schema,
       final OptionalInt limit,
       final Optional<WindowInfo> windowInfo,
-      final boolean excludeTombstones,
-      final Optional<ImmutableMap<TopicPartition, Long>> endOffsets) {
```
vvcephei, line 157: Don't need this arg anymrore, since it will now always be empty.

agavra (commented):
code LGTM, missing some test coverage though 😉

vvcephei:
Here's the issue for adding the ability to list pull queries (it falls within the scope of an existing ticket): https://github.com/confluentinc/ksql/issues/7701

vvcephei:
Thanks @agavra , I was also mulling over the testing question.

Both the stream pull and the transient query paths are already well tested, so if those tests continue to work, we can believe we haven't broken them.

I haven't checked whether there's already a test on `SHOW QUERIES;`, I hope so. If there is one and it hasn't broken, then we can also feel good about not having broken it for supported, existing use cases.

The only thing that seems to be missing is a test to make sure that `SHOW QUERIES;` does not include stream pull queries. This would be nice to do, after all that's what this PR is trying to accomplish. However, I'm not having any amazing inspiration for how to do it. What I did manually was just set up a stream with a lot of data, run a pull query on it, and then list queries. But in a test, I think we need more of an assurance that the query is still running by the time we do the evaluation. Especially since it's a negative test: if the query does end before the evaluation, the test would just wrongly pass.

I have a couple of ideas that they're all kind of weird. For example, maybe we hack in a way to create a stream pull query that never terminates.

The only alternative might be to just write a unit test that checks that creating a stream pull query doesn't register it as a "live query" in the registry, but that's not very satisfactory because it presumes the implementation of `show queries`. It might be better than nothing, though.

agavra (approved):
Thanks @vpapavas for the additional test coverage! LGTM

> The only alternative might be to just write a unit test that checks that creating a stream pull query doesn't register it as a "live query" in the registry, but that's not very satisfactory because it presumes the implementation of show queries. It might be better than nothing, though.

I think this is satisfactory for me. Thanks for the detailed discussion.

guozhangwang (commented):
LGTM.

PR #8219, opened 2021-10-02
refactor: move runtime selection to QueryPlan and initialize runtimes up front

ableegoldman:
Importantly this also updates the code to use the latest API for shared runtimes, as https://github.com/apache/kafka/pull/11272 made several breaking changes to the NamedTopology et al. APIs. Due to the nature of the upstream changes it wasn't possible to tease out only the build fixes and leave the larger refactoring for a separate PR (due to the topology builder now being tied to a specific KafkaStreams runtime, rather than a free-floating entity with hidden internal shared state)

Also lays most of the groundwork for the new config framework, and fixes up the query validation for shared runtimes


ableegoldman:
Haven't fixed up the testing code yet so it won't compile, but the non-testing code is ready for review. @guozhangwang @rodesai  @wcarlson5  please take a look

ksqldb-engine/src/main/java/io/confluent/ksql/query/QueryRegistryImpl.java
```diff
@@ -406,6 +470,54 @@ private void unregisterQuery(final QueryMetadata query) {
     notifyDeregister(query);
   }
 
+  private SharedKafkaStreamsRuntime createNewSharedKafkaStreamsRuntime(final int runtimeIndex) {
+    final SharedKafkaStreamsRuntime runtime =
+        new SharedKafkaStreamsRuntimeImpl(kafkaStreamsBuilder, buildServerStreamsProperties(runtimeIndex));
```
wcarlson5, line 475: WHere do we do that validation runtimes now?

pom.xml
```diff
@@ -146,6 +146,7 @@
         <netty-codec-http2-version>4.1.65.Final</netty-codec-http2-version>
         <!-- Brought in by both schema registry and connect, we need to pick a version -->
         <jersey-common>2.34</jersey-common>
+        <kafka.version>3.1.0-SNAPSHOT</kafka.version>
```
wcarlson5, line 149: probably don't want to merge this?

wcarlson5 (commented):
I think that my biggest question is how we are handling validation now. I can see you removed the old way but I don't fully understand the new strategy

ksqldb-engine/src/main/java/io/confluent/ksql/util/QueryMetadataImpl.java
```diff
@@ -315,10 +314,6 @@ public void setCorruptionQueryError() {
     corruptionCommandTopic = true;
   }
 
-  protected boolean isClosed() {
```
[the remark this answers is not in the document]
ableegoldman: yeah -- I'm currently working on splitting this up into smaller PRs, I can move it out into a refactoring PR or just leave it alone if you think it might at some later time be useful (didn't seem to be right now)

PR #8221, opened 2021-10-04
fix: update error msg

wcarlson5:
update error msg to remove replicas

### Reviewer checklist


ksqldb-engine/src/main/java/io/confluent/ksql/topic/TopicCreateInjector.java
```diff
@@ -110,7 +110,7 @@ public TopicCreateInjector(
     } else if (!properties.getPartitions().isPresent()) {
       final CreateSource example = createSource.copyWith(
           createSource.getElements(),
-          properties.withPartitionsAndReplicas(2, (short) 1));
+          properties.withPartitionsAndReplicas(2));
```
mjsax, line 113: It seems we use this method only at this place. Should we rename it to ` properties.withPartitions(...)` ?

wcarlson5: done

mjsax (approved):

PR #8233, opened 2021-10-06
chore: add validations for schema inference

lihaosky:
### Validations:
* if Schema id presents:
  * format must support schema inference.


ksqldb-engine/src/main/java/io/confluent/ksql/schema/ksql/inference/DefaultSchemaInjector.java
```diff
@@ -176,6 +185,25 @@ private SchemaAndId getSchema(
...
+     * Do validation when schemaId presents or we need to infer. Conditions to meet:
+     *  1. If schema id is provided, format must support schema inference
+     */
+
+    final String schemaIdName =
+        isKey ? CommonCreateConfigs.KEY_SCHEMA_ID : CommonCreateConfigs.VALUE_SCHEMA_ID;
+    if (!formatSupportsSchemaInference(formatInfo)) {
+      final String msg = (isKey ? CommonCreateConfigs.KEY_FORMAT_PROPERTY
+          : CommonCreateConfigs.VALUE_FORMAT_PROPERTY) + " should support schema inference "
+              + "when " + schemaIdName + " is provided!";
```
vcrfxia, line 202: This logic assumes that if we hit this error, then `SCHEMA_ID` was provided. What happens in the case where `SCHEMA_ID` is not provided but the format does not support schema inference? I'm guessing the code is set up in such a way that we will not reach this code path if that's the case (do we have tests for this?), but the logic in `getKeySchema()` and `getValueSchema()` above does not reflect this. This code is confusing as a result. Can we clean this up? 

Asked differently, what happens today when no key/value columns are provided and the key/value format doesn't support schema inference? Is `DefaultSchemaInjector` ever hit? If not, let's clean up the logic above to reflect that. If so, then this error message needs to be updated.

ksqldb-engine/src/main/java/io/confluent/ksql/schema/ksql/inference/DefaultSchemaInjector.java
```diff
@@ -176,6 +185,25 @@ private SchemaAndId getSchema(
     return result.schemaAndId.get();
   }
 
+  private static void validateSchemaInference(
+      final FormatInfo formatInfo,
+      final boolean isKey
+  ) {
+    /*
+     * Do validation when schemaId presents or we need to infer. Conditions to meet:
+     *  1. If schema id is provided, format must support schema inference
```
vcrfxia, line 194: If this is the only validation that's being done, then the PR description is out of date, right?

ksqldb-engine/src/main/java/io/confluent/ksql/schema/ksql/inference/DefaultSchemaInjector.java
```diff
@@ -117,17 +118,22 @@ public DefaultSchemaInjector(final TopicSchemaSupplier schemaSupplier) {
     final CreateSourceProperties props = csStmt.getProperties();
     final FormatInfo keyFormat = SourcePropertiesUtil.getKeyFormat(props, csStmt.getName());
 
-    if (hasKeyElements(statement) || !formatSupportsSchemaInference(keyFormat)) {
+    // until we support user-configuration of single key wrapping/unwrapping, we choose
+    // to have key schema inference always result in an unwrapped key
+    final SerdeFeatures serdeFeatures = SerdeFeaturesFactory.buildKeyFeatures(
```
vcrfxia, line 123: nit: move this below line 130 so we only create it if it gets used.

ksqldb-engine/src/main/java/io/confluent/ksql/schema/ksql/inference/SchemaRegistryTopicSchemaSupplier.java
```diff
@@ -172,18 +172,22 @@ private static SchemaResult incorrectFormat(
             + System.lineSeparator()
             + "expected format: " + expectedFormat
             + System.lineSeparator()
-            + "actual format: " + actualFormat
+            + "actual format from Schema Registry: " + actualFormat
     ));
   }
 
-  private static SchemaResult notFound(final String topicName, final boolean isKey) {
+  private static SchemaResult notFound(final String topicName, final boolean isKey,
```
vcrfxia, line 179: nit: let's have each parameter on its own line, if we need multiple lines.

vcrfxia (commented):
Thanks @lihaosky -- LGTM besides the confusion inline. Also, can we rename the PR title from `feat` to `chore`? Even though this is a new feature, it's a new feature that's only relevant for a yet-to-be-released feature (`SCHEMA_ID`) so each individual feature that contributes to the larger feature does not need to appear in the changelog; just one line item for the entire feature when it's ready to be released is enough.

ksqldb-engine/src/main/java/io/confluent/ksql/schema/ksql/inference/DefaultSchemaInjector.java
lihaosky: Yeah. Let me update that

ksqldb-engine/src/main/java/io/confluent/ksql/schema/ksql/inference/DefaultSchemaInjector.java
lihaosky: There's this check before the validation in line 149:

```
    if (!props.getValueSchemaId().isPresent() && (hasValueElements(statement)
        || !formatSupportsSchemaInference(valueFormat))) {
```
So `DefaultSchemaInjector` will do nothing if no columns are provided and format doesn't support schema inference. https://github.com/confluentinc/ksql/blob/master/ksqldb-parser/src/main/java/io/confluent/ksql/parser/tree/TableElements.java#L81-L83 will hit when no key and value columns are provided somewhere later in the code. 

Let me check if no column and format doesn't support inference is tested and add test if not.

lihaosky (commented):
Thanks @vcrfxia for reviewing. Will update title and description.

ksqldb-engine/src/main/java/io/confluent/ksql/schema/ksql/inference/DefaultSchemaInjector.java
lihaosky: I guess I can put the check in line 149 in this function and let this function return boolean indicating should do schema inference or not?

lihaosky: The case is covered by:
https://github.com/confluentinc/ksql/blob/master/ksqldb-functional-tests/src/test/resources/query-validation-tests/elements.json#L1-L19
https://github.com/confluentinc/ksql/blob/master/ksqldb-engine/src/test/java/io/confluent/ksql/schema/ksql/inference/DefaultSchemaInjectorTest.java#L226-L248

ksqldb-engine/src/main/java/io/confluent/ksql/schema/ksql/inference/DefaultSchemaInjector.java
```diff
@@ -185,10 +177,24 @@ private SchemaAndId getSchema(
...
   ) {
+    final boolean hasTableElements =
+        isKey ? hasKeyElements(statement) : hasValueElements(statement);
+
+    /*
+     * Conditions for schema inference:
+     *   1. key_schema_id or value_schema_id property exist or
+     *   2. Table elements doesn't exist and format support schema inference
+     */
+    if (!schemaId.isPresent() && (hasTableElements || !formatSupportsSchemaInference(formatInfo))) {
```
vcrfxia, line 194: Can we break this apart now that it's in its own method (thanks for doing that!) so it's more readable? For example:
```
if (schemaId.isPresent()) {
    // do validation specific to this case
    return true;
} else {
    // ... etc ...
}
```

ksqldb-engine/src/main/java/io/confluent/ksql/schema/ksql/inference/DefaultSchemaInjector.java
```diff
@@ -199,9 +205,11 @@ private static void validateSchemaInference(
     if (!formatSupportsSchemaInference(formatInfo)) {
       final String msg = (isKey ? CommonCreateConfigs.KEY_FORMAT_PROPERTY
           : CommonCreateConfigs.VALUE_FORMAT_PROPERTY) + " should support schema inference "
-              + "when " + schemaIdName + " is provided!";
+          + "when " + schemaIdName + " is provided!";
```
vcrfxia, line 208: Can we add the format into this error message so the error message is more helpful to users? I think this is especially important for the `C*AS` case where we might be inheriting the format from the source. The user might not realize what format is being used otherwise.

Also (a nit), can we replace the exclamation mark with a period? Very few ksqlDB error messages use exclamation marks so this one reads a little strangely IMO.

vcrfxia (commented):
Thanks @lihaosky ! LGTM besides the two minor comments inline.

vcrfxia (approved):
Thanks @lihaosky ! LGTM.

PR #8242, opened 2021-10-07
chore: throw error when table elements and schema id both present

lihaosky:
## Description
Throw error when `*_schema_id` and table elements both present

## Test done
Added new unit test


spena:
@lihaosky What's the advantage of defining a set of columns manually + using the `schema_id` in the `with` clause?

If a user is using the `schema_id` then I assume is because they want to retrieve the schema from SR. But if they define it manually, then why would they want to set the ID? I feel this validation logic may not be necessary.

lihaosky:
@spena , I think the main question is should table elements and `schema_id` be mutually exclusive in the statement. The main reason I made them coexist is I'm thinking `schema_id` is a override mechanism in addition to table elements. Like in `c*as` statement, we plan to override the schema from query with schema from `schema_id`. So here, I'm also trying to override the schema in table elements from schema in `schema_id`. We have 3 options if table elements and schema id both present:

1. Throw error. Only allow one of them exists.
2. Use schema in table elements and ignore schema id
3. Use schema in schema id and make sure it's superset of schema in table elements.

 I feel 3 is consistent with what we plan to do with `c*as` and 2 is too ambiguous and make schema id useless. I think 1 is also fine. I don't have strong preference. @colinhicks can chime in from product side I think

colinhicks:
@lihaosky @spena, for CT/CS statements I think the first option above is best:

> 1. Throw error. Only allow one of them exists.

We only need to check for a superset schema in the case of `C*AS` statements.

spena (approved):

PR #8253, opened 2021-10-13
fix: Specify source type in DROP referential integrity errors

### LINKED ISSUE #1224: The error message for referential integrity check failure on DROP should use STREAM/TABLE
Currently the error message uses the term "data source". We should use the term STREAM or TABLE instead.

Gerrrr:
### Description 

Fixes #1224.

Without this patch a DROP command for a non-existent table or stream resulted in a generic error, e.g. `Source BAAZ does not exist`.

This patch specifies the source type in the error message. For the example above, it might be `TABLE BAAZ does not exist` if the command was `DROP TABLE BAAZ`.

### Testing done 

I modified an existing test case to cover the `STREAM` case and added another case for `TABLE`.

### Reviewer checklist


Gerrrr:
FWIW the CI seems to fail because I am not a part of the organization (yet).

PR #8261, opened 2021-10-16
feat: Refactors to use KeyValueMetadata which includes RowMetadata

AlanConfluent:
### Description 
This PR is largely a refactoring to use `KeyValueMetadata` which wraps `KeyValue` and includes `RowMetadata`. `RowMetadata` includes offset information which will be used to allowing checkpointing in push queries v2, as well as consistency implementation for push queries.

It requires piping through this data from the datasource -- in the case of push v2, it's going to be a `KafkaConsumer` providing offsets. In the case of pull queries, it'll be the materialized state store. Since local data was previously passed around using a `TableRow`, which is part of the materialization package, it made sense to introduce a layer of abstraction for the query layer -- after all, push queries are processing data that's not even materialized nor always a table.

It also introduces `RowOffsets` which are written along with the rows in the various endpoints.  For push queries it will represent a range of offsets which the previous data covered, allowing checkpointing.  For pull queries, there can just be one at the end (or beginning) which represents the version of the rows returned.

### Testing done 
Ran all the tests.

### Reviewer checklist

### DIFF
--- a/ksqldb-rest-app/src/test/java/io/confluent/ksql/rest/server/resources/streaming/PullQueryPublisherTest.java
+++ b/ksqldb-rest-app/src/test/java/io/confluent/ksql/rest/server/resources/streaming/PullQueryPublisherTest.java
@@ -54,6 +54,7 @@ import io.confluent.ksql.schema.ksql.types.SqlTypes;
 import io.confluent.ksql.services.ServiceContext;
 import io.confluent.ksql.statement.ConfiguredStatement;
 import io.confluent.ksql.util.KeyValue;
+import io.confluent.ksql.util.KeyValueMetadata;
 import io.confluent.ksql.util.KsqlConfig;
 import java.util.Collection;
 import java.util.List;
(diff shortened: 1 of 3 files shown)


AlanConfluent:
> Thanks, @AlanConfluent !
> 
> This PR looks about the way I'd expect. The only change I'd recommend after a quick review is to be a bit more cagey about the structure of the consistency tokens.
> 
> In addition to what I suggested below, @vpapavas and I were also planning to use headers instead of changing the row format, to keep it out of users' way. What are your thoughts on the pros/cons there?

I think we've talked about this a fair bit offline. The main reasons I wanted to do this are to be able to stick metadata in some particular order within the rows. I need to periodically send them down, and at least the classic headers are a one-shot thing.

I agree that it might not be necessary for you to use this mechanism for pull queries if you only need to return a token once, though as we talked about, it might make sense to return it at the end if you want the user to only consider the token if everything has been successful.

> Oh, one other thing, I was planning to use a sparse vector to represent offset updates, since only one vector component will change in each record, we don't need to re-transmit the whole thing. For the consistency work, I was thinking of using a nested map of `topic -> partition -> offset`, but I guess for scalable push queries, the topic is always determined by the query, right?

Yeah, as we also talked about offline, I don't have strong feelings about the wire format, though I think it might be smaller for me to do a dense format.  Also, for SPQ, there's only one topic, so it's not really necessarily to have it, though it wouldn't hurt.  It might be totally reasonable to share an object, though I might want to add a dense output format method.  I think for the first shot, I'll probably just do something straightforward since I know that @patrickstuedi is also creating such an object and a followup can merge representations.

PR #8264, opened 2021-10-18
feat: Add consistency vector handling to CLI and Java client

vpapavas:
### Description 

Add consistency vector handling to the CLI and Java clients. The usage is hidden behind a config flag that is false by default. Both the client and the server need to enable the config. The client needs to both enable the config and add the consistency vector in the request properties. 

The consistency vector is added to the queue that stores the query results after the query has been completed successfully.  This design is not needed for http1 as we could pass the token back to the client using the headers of the `StreamedRow`. However, this is not possible to do in http2 (at least not without heavy refactoring) since the `QueryResponseMetadata` is sent back to the client before even the query executes.  Moreover, the only way for the Publisher to communicate/send something to the Subscriber is through the queue. Hence, and in order to keep the design the same in both endpoints, the vector is added to the queue as the last item. 

The vector is serialized before being written to the wire and the clients only "see" the serialized version. In practice, the CLI does not print the vector so it is hidden from the user. 

### Testing done 
I created the `ConsistencyOffsetVectorFunctionalTest` that tests that the consistency vector is correctly sent back and forth and also not sent when the flag is not enabled. 

I did manual tests for the CLI to make sure the vector is not printed but used.

With the following curl command, one can see the vector: 
```
curl -X POST   \
http://localhost:8088/query \
 -H 'content-type: application/vnd.ksql.v1+json; charset=utf-8' \
 -d '{"ksql":"SELECT * FROM tab1 where id=1;", "streamsProperties": {
      "ksql.streams.auto.offset.reset": "earliest"},"requestProperties": {
      "ksql.query.pull.consistency.token.enabled":"true", "request.ksql.query.pull.consistency.token":""}
    }'
```

```
[{"header":{"queryId":"query_1635785940946","schema":"`ID` INTEGER KEY, `KSQL_COL_0` BIGINT, `KSQL_COL_1` INTEGER"}},
{"row":{"columns":[1,1,3]}},
{"consistencyToken":{"consistencyToken":"eyJ2ZXJzaW9uIjoyLCJvZmZzZXRWZWN0b3IiOnsiZHVtbXkiOnsiNSI6NSwiNiI6NiwiNyI6N319fQ=="}}]
```

There is no way to avoid this since the response contains the vector and there is no client where I could remove it as I do with the CLI.

### Reviewer checklist


vvcephei (approved):
Thanks @vpapavas !

I suspect we might want to tweak some of the serialization stuff before the next release, but it seems like this PR has accomplished what we wanted from it.

ksqldb-api-client/src/test/java/io/confluent/ksql/api/client/integration/ConsistencyOffsetVectorFunctionalTest.java
```diff
@@ -0,0 +1,372 @@
...
+  }
+
+  private static void verifyConsistencyVector(final String serializedCV) {
+    final ConsistencyOffsetVector cvResponse = ConsistencyOffsetVector.deserialize(serializedCV);
+    assertThat(cvResponse.getVersion(), is(2));
+    assertThat(cvResponse.getOffsetVector().keySet(), hasSize(1));
+    assertThat(cvResponse.getTopicOffsets("dummy").keySet(), hasSize(3));
+    assertThat(cvResponse.getTopicOffsets("dummy").get(5), is(5L));
+    assertThat(cvResponse.getTopicOffsets("dummy").get(6), is(6L));
+    assertThat(cvResponse.getTopicOffsets("dummy").get(7), is(7L));
```
patrickstuedi, line 327: Similarly here one could test using equal

ksqldb-common/src/main/java/io/confluent/ksql/util/ConsistencyOffsetVector.java
```diff
@@ -27,17 +34,73 @@
...
+  }
+
+  public int getVersion() {
+    return version;
+  }
+
+  @VisibleForTesting
+  public ImmutableMap<String, Map<Integer, Long>> getOffsetVector() {
+    return ImmutableMap.copyOf(offsetVector);
+  }
```
patrickstuedi, line 83: Eventually I would think we don't want to expose all the internal of the OffsetVector, otherwise there would be no need to create at type.

patrickstuedi (commented):
Just leaving a few comments, nothing that's a showstopper for this PR.

ksqldb-api-client/src/main/java/io/confluent/ksql/api/client/impl/ClientImpl.java
```diff
@@ -418,6 +441,17 @@ public void undefine(final String variable) {
...
 
+  public void setRequestProperty(final String key, final Object value) {
+    Objects.requireNonNull(value, "value");
+    final Object parsed = parser.parse(key, value);
+    requestProperties.put(key, parsed);
+  }
+
+  @VisibleForTesting
+  public String getSerializedConsistencyVector() {
+    return serializedConsistencyVector.get();
```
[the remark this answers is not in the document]
vpapavas: It is not used in a test that is in the same package. The test is an integration test and uses functionality that is available in that specific package

ksqldb-common/src/main/java/io/confluent/ksql/util/ConsistencyOffsetVector.java
vpapavas: This needs to be public for Json serialization/deserialization to work

patrickstuedi (approved):

ksqldb-api-client/src/main/java/io/confluent/ksql/api/client/impl/ExecuteQueryResponseHandler.java
```diff
@@ -61,14 +67,31 @@ protected void handleMetadata(final QueryResponseMetadata queryResponseMetadata)
...
-    final JsonArray values = new JsonArray(buff);
-    if (rows.size() < maxRows) {
-      rows.add(new RowImpl(columnNames, columnTypes, values, columnNameToIndex));
+    final Row row;
+    final Object json = buff.toJson();
+
+    if (json instanceof JsonObject) {
+      final JsonObject jsonObject = (JsonObject) json;
+      // This is the serialized consistency vector
+      // Don't add it to the result list since the user should not see it
```
AlanConfluent, line 76: We talked before about supporting an older client.  It seems like the server will just return the consistency token if it's enabled on the server.  We should only send it down if it's enabled on the client which can be done by indicating that it's enabled in the request property.

ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/resources/streaming/WSQueryEndpoint.java
```diff
@@ -298,7 +301,19 @@ private void handleQuery(final RequestContext info, final Query query,
           );
       final DataSource dataSource = analysis.getFrom().getDataSource();
       final DataSource.DataSourceType dataSourceType = dataSource.getDataSourceType();
-
+      final Map<String, Object> requestProperties = info.request.getRequestProperties();
+      Optional<ConsistencyOffsetVector> consistencyOffsetVector = Optional.empty();
+      if (ConsistencyOffsetVector.isConsistencyVectorEnabled(requestProperties)
```
AlanConfluent, line 306: Is there any reason why this check is different that in other endpoints?
```
if (ksqlConfig.getBoolean(KsqlConfig.KSQL_QUERY_PULL_CONSISTENCY_OFFSET_VECTOR_ENABLED)
```

ksqldb-common/src/main/java/io/confluent/ksql/util/ConsistencyOffsetVector.java
```diff
@@ -165,12 +225,34 @@ public int hashCode() {
...
+      return Boolean.TRUE.equals(overrides.get(
+          KsqlConfig.KSQL_QUERY_PULL_CONSISTENCY_OFFSET_VECTOR_ENABLED));
+    } else {
+      return ksqlConfig.getBoolean(KsqlConfig.KSQL_QUERY_PULL_CONSISTENCY_OFFSET_VECTOR_ENABLED);
+    }
   }
-}
 
+  public static boolean isConsistencyVectorEnabled(final Map<String, Object> requestProperties) {
+    final Object consistencyEnabled
```
AlanConfluent, line 249: Normally request properties are defined in `KsqlRequestConfig`.  We don't usually send KsqlConfigs up as a request property, but rather a normal property.

ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/resources/streaming/StreamedQueryResource.java
```diff
@@ -489,6 +491,18 @@ private EndpointResponse handleTablePullQuery(
       pullBandRateLimiter.allow(KsqlQueryType.PULL);
 
       final Optional<Decrementer> optionalDecrementer = Optional.ofNullable(decrementer);
+      Optional<ConsistencyOffsetVector> consistencyOffsetVector = Optional.empty();
+      if (ksqlConfig.getBoolean(KsqlConfig.KSQL_QUERY_PULL_CONSISTENCY_OFFSET_VECTOR_ENABLED)
```
AlanConfluent, line 495: Just to make sure I understand, if the client doesn't send up anything, the feature is not enabled, right?

There are four main cases:
- The feature is disabled on the server
- The feature is enabled on the server, but not the client
- The feature is enabled on the server and the client, but the client has provided no consistency token
- The feature is enabled on the server and client, and the client provides a consistency token for the pull query.

This seems to cover all cases, but I just want to be sure.  2 is covered because the client indicates that the feature is enabled but no consistency token is provided by using an empty string?

ksqldb-api-client/src/main/java/io/confluent/ksql/api/client/impl/ClientImpl.java
```diff
@@ -136,13 +153,19 @@ public BatchedQueryResult executeQuery(
       final String sql,
       final Map<String, Object> properties
   ) {
+    if (ConsistencyOffsetVector.isConsistencyVectorEnabled(requestProperties)) {
```
AlanConfluent, line 156: Are you meaning to check this using `properties` rather than `requestProperties` since it's defined in `KsqlConfig`?

ksqldb-api-client/src/main/java/io/confluent/ksql/api/client/impl/ClientImpl.java
```diff
@@ -108,21 +116,30 @@ private ClientImpl(final ClientOptions clientOptions, final Vertx vertx,
...
 
   @Override
   public CompletableFuture<StreamedQueryResult> streamQuery(
       final String sql,
       final Map<String, Object> properties
   ) {
+    if (ConsistencyOffsetVector.isConsistencyVectorEnabled(requestProperties)) {
+      requestProperties.put(
+          KsqlRequestConfig.KSQL_REQUEST_QUERY_PULL_CONSISTENCY_OFFSET_VECTOR,
+          serializedConsistencyVector.get());
```
AlanConfluent, line 137: Can this be null rather than empty string if it's been unset?

vpapavas: Yeah, I had it as null before but the code for `JsonObject.put` throws an NPE

ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/resources/streaming/WSQueryEndpoint.java
vpapavas: My bad, left over after refactoring

ksqldb-common/src/main/java/io/confluent/ksql/util/ConsistencyOffsetVector.java
vpapavas: I wanted to avoid creating two different configs for the same thing so reused the server config in the client as well. I can change it if it is too ugly not the right way to do it

ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/resources/streaming/StreamedQueryResource.java
vpapavas: The client needs to set the config to true and then only will the consistency token be included in the request properties. So, if it is disabled on the client, no consistency token will be added to the request properties and the server won't do anything.
The empty string is used only when it is enabled and signals an empty, not yet initialized consistency token.

AlanConfluent (approved):
We talked about things offline.  A couple of those smaller things would be great to get in this PR, but if not a followup is good too.

ksqldb-api-client/src/main/java/io/confluent/ksql/api/client/impl/ClientImpl.java
vpapavas: Added to properties

PR #8277, opened 2021-10-21
fix: fix_application_id_to_work_with_acls

wcarlson5:
### Description 

### Testing done 

### Reviewer checklist


ableegoldman:
It would be great if we could get this in to 0.22 @mjsax (though I wouldn't cut a new RC for it if we don't otherwise need one)

Definitely should be in 0.23 @nateab

PR #8282, opened 2021-10-22
fix: Print only failed line on parsing exception

### LINKED ISSUE #6850: Only the failed SQL statement should be displayed in the CLI instead of all the executed statements
When executing a list of statements in one batch in the server, if one of the statements fail, the server returns all the statements that are in the batch and not only the statement that fail. This is not a problem for a few statements. However, if a user uses a SQL script with several statements, then they'll see a verbose error in the client result.

i.e.

I have a file with two statements (the create will fail with a syntax error):
```
$ cat /tmp/test.sql 
drop stream if exists s1;
create stream if not exist s1(id int) with (kafka_topic='s1', value_format='json', partitions=1);
```
When I execute the above file, I get the following error:
```
$ ./ksql --file /tmp/test.sql
line 2:8: no viable alternative at input 'create if'
Statement: drop stream if exists s1;
create if not exists stream s1(id int) with (kafka_topic='s1', value_format='json', partitions=1);
Caused by: line 2:8: no viable alternative at input 'create stream if not exist'
Caused by: org.antlr.v4.runtime.NoViableAltException
```

or take this another example with the same statements but using the CLI `-e` parameter:
```
./ksql -e "drop stream if exists s1; create stream if not exist s1(id int) with (kafka_topic='s1', value_format='json', partitions=1);"
line 1:48: no viable alternative at input 'create stream if not exist'
Statement: drop stream if exists s1; create stream if not exist s1(id int) with (kafka_topic='s1', value_format='json', partitions=1);
Caused by: line 1:48: no viable alternative at input 'create stream if not exist'
Caused by: org.antlr.v4.runtime.NoViableAltException
```

The error shows the right error in the first line, but it then displays all the statements again. If a user has 30 statements, then this line will be too large.

The error should only include the failed statement, not all of them.

i.e.
```
./ksql -e "drop stream if exists s1; create stream if not exist s1(id int) with (kafka_topic='s1', value_format='json', partitions=1);"
line 1:48: no viable alternative at input 'create stream if not exist'
Statement: create stream if not exist s1(id int) with (kafka_topic='s1', value_format='json', partitions=1);
Caused by: line 1:48: no viable alternative at input 'create stream if not exist'
Caused by: org.antlr.v4.runtime.NoViableAltException
```

None

**Additional context**
N/A

Gerrrr:
### Description 

With this patch CLI displays only the failed SQL statement out of all the executed statements in the batch. Fixes #6850. See comments starting from https://github.com/confluentinc/ksql/issues/6850#issuecomment-948798751 for the root cause analysis and rationale for the fix.

### Testing done 

I added an integration test that does `runScript` and asserts that the error message contains only the failed statement.

### Reviewer checklist

### DIFF
--- a/ksqldb-parser/src/main/java/io/confluent/ksql/parser/DefaultKsqlParser.java
+++ b/ksqldb-parser/src/main/java/io/confluent/ksql/parser/DefaultKsqlParser.java
@@ -47,6 +47,10 @@ public class DefaultKsqlParser implements KsqlParser {
           .map(DefaultKsqlParser::parsedStatement)
           .collect(Collectors.toList());
 
+    } catch (final ParsingException e) {
+      // ParsingException counts lines starting from 1
+      final String failedLine =  sql.split(System.lineSeparator())[e.getLineNumber() - 1];
+      throw new ParseFailedException(e.getMessage(), failedLine, e);
     } catch (final Exception e) {
       throw new ParseFailedException(e.getMessage(), sql, e);
     }
--- a/ksqldb-cli/src/test/java/io/confluent/ksql/cli/CliTest.java
+++ b/ksqldb-cli/src/test/java/io/confluent/ksql/cli/CliTest.java
@@ -1217,6 +1217,27 @@ public class CliTest {
   }
 
   @Test
+  public void shouldPrintOnlyFailedStatementFromScriptFile() throws Exception {
+    // Given:
+    final File scriptFile = TMP.newFile("script.sql");
+    Files.write(scriptFile.toPath(), (""
+        + "drop stream if exists s1;\n"
+        + "create stream if not exist s1(id int) with (kafka_topic='s1', value_format='json', "
+        + "partitions=1);\n").getBytes(StandardCharsets.UTF_8));
+
+    // When:
+    localCli.runScript(scriptFile.getPath());
+
+    // Then:
+    final String out = terminal.getOutputString();
+    final String expected = "Statement: create stream if not exist s1(id int) "
+        + "with (kafka_topic='s1', value_format='json', partitions=1);\n"
+        + "Caused by: line 2:22: no viable alternative at input 'create stream if not";
+    assertThat(out, containsString(expected));
+    assertThat(out, not(containsString("drop stream if exists")));
+  }
+
+  @Test
   public void shouldUpdateCommandSequenceNumber() throws Exception {
     // Given:
     final String statementText = "create stream foo;";


ksqldb-parser/src/main/java/io/confluent/ksql/parser/DefaultKsqlParser.java
```diff
@@ -47,6 +48,15 @@
           .map(DefaultKsqlParser::parsedStatement)
           .collect(Collectors.toList());
 
+    } catch (final ParsingException e) {
+      // ParsingException counts lines starting from 1
+      final String failedLine =  sql.split(System.lineSeparator())[e.getLineNumber() - 1];
+      // Since we return only the failed statement, the line number for it is always 1
```
spena, line 54: Sorry, but after thinking a little bit more, I think you had the right number in the previous commit. If a user submit > 1 statement, then they should know the line of the set of statements they submit. There are uses that call the REST API directly with several statements.

Could you go back to the original line number?

Gerrrr: Makes sense! I reverted the change in 441f6ba.

spena (approved):

Gerrrr:
CI test failure looks unrelated:

```
Error Message
Error issuing POST to KSQL server. path:/query
Stacktrace
io.confluent.ksql.rest.client.KsqlRestClientException: Error issuing POST to KSQL server. path:/query
	at io.confluent.ksql.api.integration.ScalablePushBandwidthThrottleIntegrationTest.makeQueryRequest(ScalablePushBandwidthThrottleIntegrationTest.java:240)
	at io.confluent.ksql.api.integration.ScalablePushBandwidthThrottleIntegrationTest.makeRequestAndSetupSubscriber(ScalablePushBandwidthThrottleIntegrationTest.java:220)
	at io.confluent.ksql.api.integration.ScalablePushBandwidthThrottleIntegrationTest.scalablePushBandwidthThrottleTestHTTP1(ScalablePushBandwidthThrottleIntegrationTest.java:144)
Caused by: java.util.concurrent.ExecutionException: io.vertx.core.VertxException: Connection was closed
	at io.confluent.ksql.api.integration.ScalablePushBandwidthThrottleIntegrationTest.makeQueryRequest(ScalablePushBandwidthThrottleIntegrationTest.java:240)
	at io.confluent.ksql.api.integration.ScalablePushBandwidthThrottleIntegrationTest.makeRequestAndSetupSubscriber(ScalablePushBandwidthThrottleIntegrationTest.java:220)
	at io.confluent.ksql.api.integration.ScalablePushBandwidthThrottleIntegrationTest.scalablePushBandwidthThrottleTestHTTP1(ScalablePushBandwidthThrottleIntegrationTest.java:144)
Caused by: io.vertx.core.VertxException: Connection was closed
```

PR #8286, opened 2021-10-22
chore: stop uppercase field name during schema inference

lihaosky:
### Description 
* Put schema id to format properties to stop upper case field name during translation
* Throw exceptions instead of silent ignore unsupported types

### Testing done 
* Added new unit test
* Fixed broken unit test

### Reviewer checklist

### DIFF
--- a/ksqldb-engine/src/main/java/io/confluent/ksql/schema/ksql/inference/SchemaRegistryTopicSchemaSupplier.java
+++ b/ksqldb-engine/src/main/java/io/confluent/ksql/schema/ksql/inference/SchemaRegistryTopicSchemaSupplier.java
@@ -129,11 +129,16 @@ public class SchemaRegistryTopicSchemaSupplier implements TopicSchemaSupplier {
     final Format format = formatFactory.apply(expectedFormat);
 
     // Put schema id to properties so that translator will not uppercase field name
-    final Map<String, String> propertiesWithId = new ImmutableMap.Builder<String, String>().putAll(
-            expectedFormat.getProperties())
-        .put(isKey ? ConnectFormat.KEY_SCHEMA_ID : ConnectFormat.VALUE_SCHEMA_ID,
-            String.valueOf(id))
-        .build();
+    boolean schemaIdInProperty =
+        isKey ? expectedFormat.getProperties().containsKey(ConnectFormat.KEY_SCHEMA_ID)
+            : expectedFormat.getProperties().containsKey(ConnectFormat.VALUE_SCHEMA_ID);
+    final Map<String, String> propertiesWithId = schemaIdInProperty ? expectedFormat.getProperties()
+        : new ImmutableMap.Builder<String, String>().putAll(
+                expectedFormat.getProperties())
+            .put(isKey ? ConnectFormat.KEY_SCHEMA_ID : ConnectFormat.VALUE_SCHEMA_ID,
+                String.valueOf(id))
+            .build();
+
     final SchemaTranslator translator = format.getSchemaTranslator(propertiesWithId);
 
     if (!parsedSchema.schemaType().equals(translator.name())) {


ksqldb-engine/src/main/java/io/confluent/ksql/schema/ksql/inference/SchemaRegistryTopicSchemaSupplier.java
```diff
@@ -124,7 +127,19 @@ private SchemaResult fromParsedSchema(
...
+
+    // Put schema id to properties so that translator will not uppercase field name
+    boolean schemaIdInProperty =
+        isKey ? expectedFormat.getProperties().containsKey(ConnectFormat.KEY_SCHEMA_ID)
+            : expectedFormat.getProperties().containsKey(ConnectFormat.VALUE_SCHEMA_ID);
+    final Map<String, String> propertiesWithId = schemaIdInProperty ? expectedFormat.getProperties()
+        : new ImmutableMap.Builder<String, String>().putAll(
+                expectedFormat.getProperties())
+            .put(isKey ? ConnectFormat.KEY_SCHEMA_ID : ConnectFormat.VALUE_SCHEMA_ID,
+                String.valueOf(id))
```
vcrfxia, line 139: Does this mean we're changing the default behavior (even when no explicit schema ID is supplied in the WITH clause properties) to no longer uppercase field names? Was this discussed as part of the original KLIP? I was under the impression that we were going to leave the default behavior as is and allow users to explicitly specify SCHEMA_ID in the WITH clause properties if they wanted to preserve casing, but maybe I've misunderstood.

ksqldb-engine/src/test/java/io/confluent/ksql/schema/ksql/inference/DefaultSchemaInjectorFunctionalTest.java
```diff
@@ -400,10 +401,15 @@ public void shouldIgnoreConnectMapWithUnsupportedKey() {
...
-    );
+    final Exception e = assertThrows(
+        KsqlStatementException.class,
+        () -> shouldInferConnectType(
+            schema,
+            SchemaBuilder.struct().field("foo", Schema.OPTIONAL_STRING_SCHEMA).optional().build()
+        ));
+
+    // Then:
+    assertThat(e.getMessage(), containsString("Unsupported type for map key: bytes"));
```
vcrfxia, line 412: Update the unit test name to reflect the new expected behavior.

ksqldb-engine/src/test/java/io/confluent/ksql/schema/ksql/inference/SchemaRegistryTopicSchemaSupplierTest.java
```diff
@@ -535,11 +537,17 @@ public void shouldRequestCorrectSchemaOnGetKeySchemaWithId() throws Exception {
 
   @Test
   public void shouldPassRightSchemaToFormat() {
+    // Given:
+    Map<String, String> properties = new ImmutableMap.Builder<String, String>()
+        .putAll(formatProperties)
+        .put("VALUE_SCHEMA_ID", "12")
```
vcrfxia, line 543: What is the purpose of adding this in this unit test? Is this used in the test?

ksqldb-functional-tests/src/test/java/io/confluent/ksql/test/SchemaTranslationTest.java
```diff
@@ -180,6 +180,7 @@ TestCase buildTest(final Path originalFileName, final TestLocation location) {
         final String csasStatement = schema.rawSchema().getFields()
             .stream()
             .map(Schema.Field::name)
+            .map(name -> "`" + name + "`")
```
vcrfxia, line 183: Why is this change needed as a result of this PR?

ksqldb-serde/src/main/java/io/confluent/ksql/serde/connect/ConnectKsqlSchemaTranslator.java
```diff
@@ -184,6 +184,7 @@ private Schema toKsqlStructSchema(final Schema schema) {
         }
       } catch (final UnsupportedTypeException e) {
         log.error("Error inferring schema at field {}: {}", field.name(), e.getMessage());
+        throw e;
```
vcrfxia, line 187: Is this change related to KLIP-56? What's the motivation for this change?

Also, this change is backwards incompatible, right? I think the impact of this change is that fields with certain types of incompatible schemas were ignored while now the entire statement would be rejected. Is that correct? AFAICT this method is only ever called from the DefaultSchemaInjector which should only be called when new statements are submitted, and not when existing statements are executed from the command topic. Can you confirm? (This is important so we don't break compatibility for existing statements that have already been written to the command topic.)

ksqldb-serde/src/main/java/io/confluent/ksql/serde/protobuf/ProtobufFormat.java
```diff
@@ -59,7 +59,7 @@ public String name() {
   protected ConnectSchemaTranslator getConnectSchemaTranslator(
       final Map<String, String> formatProps
   ) {
-    FormatProperties.validateProperties(name(), formatProps, ImmutableSet.of());
+    FormatProperties.validateProperties(name(), formatProps, getSupportedProperties());
```
vcrfxia, line 62: This is a fix from a previous PR, unrelated to this one, right?

ksqldb-functional-tests/src/test/resources/historical_plans/protobuf_-_protobuf_inference_-_other_scalars/7.1.0_1635272037962/plan.json
```diff
@@ -0,0 +1,177 @@
+{
```
vcrfxia, line 1: Is it true that the only QTTs which needed to be updated are those that use schema inference without explicitly specifying KEY_SCHEMA_ID or VALUE_SCHEMA_ID in the WITH clause properties?

Also, is it true that the only changes to the historic plans is that KEY_SCHEMA_ID and/or VALUE_SCHEMA_ID is now being written into the plan?

(Want to check my understanding of the changes in this PR. Thanks.)

vcrfxia (commented):
Thanks @lihaosky ! Questions inline.

lihaosky (commented):
Thanks @vcrfxia for the review!

ksqldb-serde/src/main/java/io/confluent/ksql/serde/connect/ConnectKsqlSchemaTranslator.java
lihaosky: Should be related to KLIP-56 as this could be thrown during schema inference.

Yeah. I can only find this is called in `DefaultSchemaInjector` before command is written to command topic. 

> I think the impact of this change is that fields with certain types of incompatible schemas were ignored while now the entire statement would be rejected. Is that correct?

Yes. I want to make the unsupported type error explicit instead of silently ignored and make user think their schemas are inferred correctly but in fact not

ksqldb-serde/src/main/java/io/confluent/ksql/serde/protobuf/ProtobufFormat.java
lihaosky: Correct

ksqldb-engine/src/main/java/io/confluent/ksql/schema/ksql/inference/SchemaRegistryTopicSchemaSupplier.java
vcrfxia: Even if there are no backwards compatibility concerns for statements that have already been written to the command topic, the change is still backwards incompatible in that the same (new) statement issued on different server versions will produce different results. Working with streams with lowercase field names in ksqlDB is not very intuitive/convenient: the field names must be escaped in all subsequent statements. This is a rather large change for users to adjust to. That's not to say that we shouldn't make it, just that I think it warrants justification in a KLIP and I'd want someone more user-facing onboard cc @colinhicks

ksqldb-engine/src/test/java/io/confluent/ksql/schema/ksql/inference/SchemaRegistryTopicSchemaSupplierTest.java
vcrfxia: Ah I see. I'd rename `properties` to `expectedProperties`. Either moving it to be right before line 550 or leaving it here is fine (I've seen both in the codebase).

My confusion was that I saw this was in the `Given` and thought it was being passed in as part of the `When` and then immediately checked again in the `Then`, but I now realize I misread. Clarifying its usage by renaming would've prevented my misread.

ksqldb-engine/src/main/java/io/confluent/ksql/schema/ksql/inference/SchemaRegistryTopicSchemaSupplier.java
lihaosky: Sounds good. @colinhicks please take a look whether we want to have this behavior for schema inference

lihaosky:
Abandon as we want to keep existing behavior

Issue #1940, opened 2018-09-24
Expose Kafka message headers as new implicit column


blueedgenick:
Similar to how we expose the key and timestamp of the kafka message via special in-built pseudo-columns `rowkey` and `rowtime`, we can expose the message headers via a new pseudo-column, of type `map`, named `rowheaders`.

See also #658 which is looking to add "message routing based on headers".

big-andy-coates:
With ROWTIME now a pseudo column, the door is now open for this change. However, it will require a KLIP.

jeqo:
I have to correct my previous comment.

I was assuming headers would have to be stored, as in a `i.c.k.execution.streams.materialization.ks.KsMaterializedTable`, which would require StateStores to support headers as well. 
But if headers are mapped as a column then there is no need to change StateStores.

I took a deeper look into the current implementation and I'd like to propose the following features:

- Read Headers: Similar to `ROWTIME`, create a `ROWHEADERS` system column to map record headers into a `MAP`. This will allow users to get headers as columns.
- Set Headers: Add `WITH HEADERS valueExpression (, valueExpression)?` to queries to allow users to set headers in streams.

Here is an example on how these features will play out: 

```sql
--Given
CREATE STREAM orders (
    ORDERID INT KEY,
    ORDERUNITS double
)
WITH (
    kafka_topic='test_topic',
    partitions=12,
    replicas=1,
    value_format='JSON'
);
-- Set header from column
CREATE STREAM order_with_headers AS
SELECT
    *
    END AS case_result
FROM orders
WITH HEADERS orderid
EMIT CHANGES;
-- Read headers
CREATE STREAM order_and_headers AS
SELECT
    ROWHEADERS AS row_headers,
    *
FROM order_with_headers
EMIT CHANGES;
```

If this make sense, I'd be happy to propose a KLIP.

big-andy-coates:
Hi @jeqo,

Thanks for getting involved. Some of thoughts...

1. Headers in Kafka can have multiple values for the same key, so in KsqlDB this would probably be best represented as `MAP<STRING, ARRAY<something>>`, i.e. a map of string header to an array of elements.
2. The value of the header is a problem - it's just a byte array.   KsqlDB would need a [BYTES](https://github.com/confluentinc/ksql/issues/1742) type before we can expose headers.
3. KsqlDB normally uses the `WITH()` clause to control serialization options so it might be more standard to set headers via this clause, e.g. 

```sql
[...]

PR #8293, opened 2021-10-25
docs: klip-57: kafka headers

jzaralim:
### Description 
KLIP for exposing kafka headers in ksqlDB. 

GitHub issue: https://github.com/confluentinc/ksql/issues/1940

### Reviewer checklist


design-proposals/klip-57-kafka-headers.md
```diff
@@ -0,0 +1,132 @@
...
+* Including header columns in a data source's schema
+* Functions to help decode byte strings to primitive types (BYTES to STRING already exists, new functions would be conversions to INT, BIGINT and DOUBLE)
+
+## What is not in scope
+* Writing headers to output topics - we might do this in the future, but that is a separate discussion
+* Advanced functions to help decode byte strings (e.g. functions to deserialize complex serialized data.) - if a user wants to convert something that we don't have a function for, then they can either create their own UDF or use `CREATE FUNCTION` when that becomes available.
+
+## Public APIS
+
+### Syntax
```
mjsax, line 25: Can we point out that there can be only a single column defined as `HEADERS` ?

design-proposals/klip-57-kafka-headers.md
```diff
@@ -0,0 +1,132 @@
...
+```
+If the key is not present in the header, then the value in the column will be `null`.
+
+Sources created with schema inferencing can have header columns, they would just have to be specified. For example, the statement `CREATE STREAM A (my_headers ARRAY<STRUCT<key STRING, value BYTES>> HEADERS) WITH (kafka_topic='a', value_format='avro');` will create a stream with columns from Schema Registry, as well as one header column.
+
+### New byte-conversion functions
+
+There will also be new byte conversion functions to help decode header data:
+
+* `INT_FROM_BYTES(bytes, byte_order='big', signed=true)` - The input byte array must be exactly 4 bytes long. Signed conversions will by handled by `ByteBuffer`'s `getInt` method. Unsigned conversions will be manually calculated. If the result does not fit into an INTEGER then the method will throw an error.
```
mjsax, line 48: > then the method will throw an error.

What does this imply? Don't think we want to crash/stop a persistent query for this case?

design-proposals/klip-57-kafka-headers.md
```diff
@@ -0,0 +1,132 @@
...
+* `BIGINT_FROM_BYTES(bytes, byte_order='big', signed=true)` - The input byte array must be exactly 8 bytes long. Signed conversions will by handled by `ByteBuffer`'s `getLong` method. Unsigned conversions will be manually calculated. If the result does not fit into a BIGINT then the method will throw an error.
+* `DOUBLE_FROM_BYTES(bytes, byte_order='big')` - The input byte array must be exactly 8 bytes long. The conversion will by handled by `ByteBuffer`'s `getDouble` method.
+
+### Queries
+
+The headers columns will usable in any query like any other column, except `INSERT` query will be blocked. However, when persistent queries project header-backed columns, the header values are copied into the output row’s value (or key), not to headers in the output record.
```
mjsax, line 54: nit `will [be] usable`

design-proposals/klip-57-kafka-headers.md
```diff
@@ -0,0 +1,132 @@
...
+
+### Schema
+
+Header columns will fully be a part of the stream/table's schema. When running `SELECT *`, header columns will be included in the projection, and they will also appear when describing a source.
+
+## Design
+Columns in the logical schema will be represented by a new type of column `HeaderColumn` which will extend `Column` and contain an optional String named `key`. They will be in a new `HEADER` namespace, which functions identically to the `VALUE` namespace, except that the `StreamBuilder` will populate those columns with header values.
```
mjsax, line 61: What is the optional `key` field in `HeaderColumn` about?

design-proposals/klip-57-kafka-headers.md
```diff
@@ -0,0 +1,132 @@
...
+
+## What is not in scope
+* Writing headers to output topics - we might do this in the future, but that is a separate discussion
+* Advanced functions to help decode byte strings (e.g. functions to deserialize complex serialized data.) - if a user wants to convert something that we don't have a function for, then they can either create their own UDF or use `CREATE FUNCTION` when that becomes available.
+
+## Public APIS
+
+### Syntax
+
+Unlike other pseudocolumns, if a user wants to access headers, they will create a column with the `HEADERS` keyword. This column will be populated with the full list of header keys and values.
```
mjsax, line 27: Can we add a short reasoning why we handle headers different to timestamp/offset/partition?

design-proposals/klip-57-kafka-headers.md
```diff
@@ -0,0 +1,132 @@
...
+Columns in the logical schema will be represented by a new type of column `HeaderColumn` which will extend `Column` and contain an optional String named `key`. They will be in a new `HEADER` namespace, which functions identically to the `VALUE` namespace, except that the `StreamBuilder` will populate those columns with header values. 
+
+## Test plan
+
+The following test cases will be added:
+* Creating `STREAM`/`TABLE` with header columns
+* `SELECT * FROM ...` queries include header columns in the output
+* `DESCRIBE SOURCE ...` includes any header columns
+* Push and pull queries may select headers
+* Queries can use header columns in UDFs, filters, group by and partition by clauses
```
mjsax, line 70: Guess we could also join using headers?

design-proposals/klip-57-kafka-headers.md
```diff
@@ -0,0 +1,132 @@
...
+## Alternatives Considered
+### Use a property in the WITH clause to identify the column to be interpreted as headers
+Example: 
+```
+-- populate `my_headers` from each row's record headers
+CREATE STREAM FOO (my_headers ARRAY<STRUCT<key VARCHAR, value VARCHAR>>)
+  WITH (headers='my_headers');
+```
+* Supporting multiple header-backed columns would likely be awkward in WITH-based approach.
+* `HEADER(<key>)` would not work well in this approach.
```
mjsax, line 112: Can we add
```
Aligns to existing syntax of `KEY` and `PRIMARY KEY` that tell ksqlDB to populate a row from the record key, instead the record value (default).
```

jzaralim: added

design-proposals/klip-57-kafka-headers.md
jzaralim: Added a rejected alternative

design-proposals/klip-57-kafka-headers.md
jzaralim: Added some clarification

design-proposals/klip-57-kafka-headers.md
jzaralim: When UDFs in ksqlDB throw errors, the error is logged and `null` is written to the output topic. So this won't crash persistent queries.

design-proposals/klip-57-kafka-headers.md
jzaralim: Leaving this out. While having multiple `HEADERS` columns is pretty useless, it doesn't break anything.

design-proposals/klip-57-kafka-headers.md
```diff
@@ -0,0 +1,138 @@
...
+
+There will also be new byte conversion functions to help decode header data:
+
+* `INT_FROM_BYTES(bytes, byte_order='big', signed=true)` - The input byte array must be exactly 4 bytes long. Signed conversions will by handled by `ByteBuffer`'s `getInt` method. Unsigned conversions will be manually calculated. If the result does not fit into an INTEGER then the method will throw an error.
+* `BIGINT_FROM_BYTES(bytes, byte_order='big', signed=true)` - The input byte array must be exactly 8 bytes long. Signed conversions will by handled by `ByteBuffer`'s `getLong` method. Unsigned conversions will be manually calculated. If the result does not fit into a BIGINT then the method will throw an error.
+* `DOUBLE_FROM_BYTES(bytes, byte_order='big')` - The input byte array must be exactly 8 bytes long. The conversion will by handled by `ByteBuffer`'s `getDouble` method.
+
+### Queries
+
+The headers columns will be usable in any query like any other column, except `INSERT` query will be blocked. However, when persistent queries project header-backed columns, the header values are copied into the output row’s value (or key), not to headers in the output record.
```
spena, line 54: What does `blocked` mean? Will the `INSERT SELECT` and `INSERT VALUES` fail if trying to insert into a headers column or just blocked/not used?

design-proposals/klip-57-kafka-headers.md
```diff
@@ -0,0 +1,138 @@
...
+* Writing headers to output topics - we might do this in the future, but that is a separate discussion
+* Advanced functions to help decode byte strings (e.g. functions to deserialize complex serialized data.) - if a user wants to convert something that we don't have a function for, then they can either create their own UDF or use `CREATE FUNCTION` when that becomes available.
+
+## Public APIS
+
+### Syntax
+
+Unlike other pseudocolumns, if a user wants to access headers, they will create a column with the `HEADERS` keyword. This column will be populated with the full list of header keys and values.
+```
+ksql> CREATE STREAM A (my_headers ARRAY<STRUCT<key STRING, value BYTES>> HEADERS);
```
spena, line 29: Can I use `HEADERS` on different columns?

```
CREATE STREAM A (
  my_headers1 ARRAY<STRUCT<key STRING, value BYTES>> HEADERS,
  my_headers2 ARRAY<STRUCT<key STRING, value BYTES>> HEADERS
);
```

design-proposals/klip-57-kafka-headers.md
```diff
@@ -0,0 +1,138 @@
...
+* `BIGINT_FROM_BYTES(bytes, byte_order='big', signed=true)` - The input byte array must be exactly 8 bytes long. Signed conversions will by handled by `ByteBuffer`'s `getLong` method. Unsigned conversions will be manually calculated. If the result does not fit into a BIGINT then the method will throw an error.
+* `DOUBLE_FROM_BYTES(bytes, byte_order='big')` - The input byte array must be exactly 8 bytes long. The conversion will by handled by `ByteBuffer`'s `getDouble` method.
+
+### Queries
+
+The headers columns will be usable in any query like any other column, except `INSERT` query will be blocked. However, when persistent queries project header-backed columns, the header values are copied into the output row’s value (or key), not to headers in the output record.
+
+### Schema
+
+Header columns will fully be a part of the stream/table's schema. When running `SELECT *`, header columns will be included in the projection, and they will also appear when describing a source.
```
spena, line 58: I assume the `HEADERS` attribute won't be copied to the sink schema either, right?

design-proposals/klip-57-kafka-headers.md
```diff
@@ -0,0 +1,138 @@
...
+ksql> CREATE STREAM B (my_headers ARRAY<BYTES> HEADERS);
+Error: Columns specified with the HEADERS keyword must be typed as ARRAY<STRUCT<key STRING, value BYTES>>.
+```
+Users can also mark a column with `HEADER(<key>)`, which will populate the column with the last header that matches the key. The data type must be `BYTES`
+```
+ksql> CREATE STREAM B (value BYTES HEADER(<key>));
+```
+If the key is not present in the header, then the value in the column will be `null`.
+
+Sources created with schema inferencing can have header columns, they would just have to be specified. For example, the statement `CREATE STREAM A (my_headers ARRAY<STRUCT<key STRING, value BYTES>> HEADERS) WITH (kafka_topic='a', value_format='avro');` will create a stream with columns from Schema Registry, as well as one header column.
```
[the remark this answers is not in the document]
colinhicks: > I don't think we should support schema inference this way

@spena to me the KLIP essentially suggests that schema inference is left alone if you define a column with the `HEADER(<key>)` or `HEADERS` keywords. Any column specified with that keyword is ignored from the perspective of schema registry: we're not trying to apply a SR schema to it or derive something in the output SR schema from it.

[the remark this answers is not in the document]
spena: I'm not sure I follow this section then. My understanding is that this would happen:

AVRO SR = (id INT, name STRING)
```
# with headers
ksql> CREATE STREAM A (my_headers ARRAY<STRUCT<key STRING, value BYTES>> HEADERS) WITH (kafka_topic='a', value_format='avro');
ksql> DESCRIBE A;
(id INT, name STRING, my_headers ARRAY<STRUCT<key STRING, value BYTES>> HEADERS)

# without headers
ksql> CREATE STREAM A (my_headers ARRAY<STRUCT<key STRING, value BYTES>>) WITH (kafka_topic='a', value_format='avro');
ksql> DESCRIBE A;
(my_headers ARRAY<STRUCT<key STRING, value BYTES>>)
```

Is the above correct? If so, that's what I think shouldn't happen. I think it will make the user confused about it. I'm not sure how it should be, but I'm just concerned that will generate other problems. I think too that we could solve it in the same way we will solve (not sure how) other column attributes in the column definition. For example:

AVRO SR = (id INT, name STRING)
```
CREATE STREAM A (id INT, name STRING NOT NULL) WITH (kafka_topic='a', value_format='avro');
```

 The above is just an example that if we have to include attributes that are not in SR, then we have to define the schema manually. Or have another alternative for SR inference with extra column attributes in specific columns.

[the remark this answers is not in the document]
jzaralim: @spena That is correct. The behaviour would be similar to [partial injections](https://docs.ksqldb.io/en/latest/operate-and-deploy/schema-registry-integration/#with-partial-schema-inference).

The difference between HEADERS and other attributes like NOT NULL is that the data in headers are not included in SR, while a column with the NOT NULL attribute can still be included in a schema from SR.

design-proposals/klip-57-kafka-headers.md
jzaralim: They will fail, added that to the klip

design-proposals/klip-57-kafka-headers.md
jzaralim: Yes, added to klip

design-proposals/klip-57-kafka-headers.md
jzaralim: Added a clarification

design-proposals/klip-57-kafka-headers.md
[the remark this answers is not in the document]
mjsax: I agree that partial projection can be confusing to users, but I don't see that adding headers make the situation worse? @spena can you elaborate on your concern? Why does adding headers make it more confusing for users from your POV?

Also not sure how, `NOT NULL` et al. relate? I don't see any connection? Can you elaborate?

design-proposals/klip-57-kafka-headers.md
mjsax: > and null is written to the output topic

For the headers case, we are on the read path. So would we have a header `<k,null>` or drop the header if we cannot parse the header-value? Maybe having `<k,null>` is better to at least indicate to the user that there _is_ a header?

jzaralim: That's not a situation that would come up to begin with, since doing something like `FUNCTION(COLUMN_NAME)` creates a new column, rather than writing to `COLUMN_NAME`. So there is no way a UDF would write anything to the header field of an output topic.

design-proposals/klip-57-kafka-headers.md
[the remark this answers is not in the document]
spena: I see. I wasn't aware of the partial injections feature. In that case, then the UX is the same with HEADERS than with having KEY as the only column and infer the value columns. 

@mjsax The relation with `NOT NULL` it was just an example of adding attributes next to the column. In this case we have 3 type of attributes that specify a different scope of the column (KEY, HEADERS and VALUE). I was just getting confused.

design-proposals/klip-57-kafka-headers.md
```diff
@@ -0,0 +1,140 @@
...
+## Public APIS
+
+### Syntax
+
+Unlike other pseudocolumns, if a user wants to access headers, they will create a column with the `HEADERS` keyword. This column will be populated with the full list of header keys and values. There can be any number of HEADERS columns.
+```
+ksql> CREATE STREAM A (
+    my_headers ARRAY<STRUCT<key STRING, value BYTES>> HEADERS,
+    my_headers_2 ARRAY<STRUCT<key STRING, value BYTES>> HEADERS);
+```
```
spena, line 32: I think we might have issues with multiple HEADERS section in the future when we support inserts. For instance, trying to insert the same keys with different values in both headers. Which one would be the last key written? I think we should allow only one HEADER section. @colinhicks what do you think?

design-proposals/klip-57-kafka-headers.md
```diff
@@ -0,0 +1,140 @@
...
+
+### New byte-conversion functions
+
+There will also be new byte conversion functions to help decode header data:
+
+* `INT_FROM_BYTES(bytes, byte_order='big', signed=true)` - The input byte array must be exactly 4 bytes long. Signed conversions will by handled by `ByteBuffer`'s `getInt` method. Unsigned conversions will be manually calculated. If the result does not fit into an INTEGER then the method will throw an error.
+* `BIGINT_FROM_BYTES(bytes, byte_order='big', signed=true)` - The input byte array must be exactly 8 bytes long. Signed conversions will by handled by `ByteBuffer`'s `getLong` method. Unsigned conversions will be manually calculated. If the result does not fit into a BIGINT then the method will throw an error.
+* `DOUBLE_FROM_BYTES(bytes, byte_order='big')` - The input byte array must be exactly 8 bytes long. The conversion will by handled by `ByteBuffer`'s `getDouble` method.
+
+### Queries
```
spena, line 54: Will headers be supported in predicates?

design-proposals/klip-57-kafka-headers.md
mjsax: For the write path, I agree that we need to limit to a single header. But it seems orthogonal to allow `HEADERS` multiple times to define an "input" schema.

For the "input" schema, you would get multiple columns reading the same headers (which is redundant and not useful) but also not harmful.

For the write path, we don't have a way to write into the header atm, and if a header is selected right now, it would be copied into the record _value_ (so no issue so far). My proposal would be to add a new function `AS_HEADER()` (similar to `AS_VALUE()` that copies a key column into the record value), and we can limit that `AS_HEADER()` is used at max once in the `SELECT` clause.

design-proposals/klip-57-kafka-headers.md
jzaralim: Yup. Added a clarification

design-proposals/klip-57-kafka-headers.md
jzaralim: Added a note with explanation on how error handling will work for UDFs using headers

design-proposals/klip-57-kafka-headers.md
colinhicks: Yeah, I support what @mjsax is suggesting. I think for insert-into-values, too, users would need to use a function like `AS_HEADER()`, rather than assigning the headers based on the logical schema's column name. Maybe we would even provide an `AS_HEADER(<header-key>, <header-value>)` signature to make this more convenient ...

```
INSERT INTO A () VALUES (AS_HEADER('my-header-key', 'my-header-value'));
```

We can hash out the details when the time comes to support writes.

spena: Yes, multiple HEADERS columns is not harmful, but I think we should not allow multiple headers yet if we haven't hash out the details for the insert statements. The idea of the AS_HEADER sounds good, but what if there are other recommendations later? others that will conflict with the multiple headers and we cannot use them because users already using multiple headers? I prefer to not support that until we have a good design for how we'll insert headers.

design-proposals/klip-57-kafka-headers.md
spena: There's an optimization we can do if we allow predicates with headers. Headers are not serialized and are not part of the key/value schema. So, if a header is not found in a row, then we can prevent deserializing the key/value schema. This will save time when processing records. Is this something you have considered? is it out of scope or can it be done in this klip? @colinhicks @mjsax, thoughts?

jzaralim: Yeah that would speed things up. It's a complicated refactor though, so I don't think the effort is worth it for this klip.

design-proposals/klip-57-kafka-headers.md
mjsax: `INSERT INTO` is actually an excellent point. So far, for the write path, we only discussed persistent queries (CSAS/CTAS). However, particularly `INSERT INTO VALUES` is a different write path that we did not discuss yet. (Not 100% if `INSERT INTO AS SELECT` would fall into the CSAS/CTAS bucket or the `INSERT INTO VALUES` bucket?)

I guess we could actually support `INSERT INTO VALUES` directly in the first version: in the end, the type of the header column is well defined, and thus, if a user provides as with a valid `STRUCT`, there is no reason not to write it into the header field of the record. It's of course more work for (v1) but it might not be too complex? If we don't want to support it, the question is if a stream/table with a header column would need to reject _all_ `INSERT INTO VALUES` statements? Or could we require the that header column is `NULL` as long as we don't support `INSERT INTOE VALUES`?

In any case, it would be weird to have two header columns for `INSERT INTO VALUES`, because it would not be well defined how we "merge" both into a single record header field. As allowing multiple HEADER columns does not really buy us anything, but disallow multiple avoid the issue for `INSERT INTO VALUES`, I am convinced now, that we should strictly enforce a single HEADER column in a schema definition.

design-proposals/klip-57-kafka-headers.md
mjsax: Sounds like an implementation detail to me. Don't think we need to cover it in the KLIP.

design-proposals/klip-57-kafka-headers.md
mjsax: As discussed below, we should not allow it, but restrict a schema definition to at most one HEADER column.

design-proposals/klip-57-kafka-headers.md
jzaralim: That makes sense. So to recap:

1. We'll limit to one `HEADERS` column.
2. If there is a `HEADERS` column, then we cannot have any `HEADER(key)` columns.
3. There can still be multiple `HEADER(key)` columns, since merging them all isn't as confusing.

From an implementation standpoint, `INSERT INTO VALUES` might not be too hard, and i would testing more convenient. I think it should be a stretch goal though, since this klip only promises reading kafka headers.

spena: Should prevent multiple `HEADER(key)` that uses the same `key` too?

jzaralim: I was thinking that since it's possible for a header to repeat keys, merging won't be a problem if we can decide on how to order them. Though now that I think of it, this is basically the same problem as before, but on a smaller scale, so it makes sense to prevent the same `key` from being used multiple times. I'll update the KLIP with the restrictions.

spena (approved):
Thanks. The rest looks good to me

design-proposals/klip-57-kafka-headers.md
```diff
@@ -0,0 +1,141 @@
...
+## Public APIS
+
+### Syntax
+
+Unlike other pseudocolumns, if a user wants to access headers, they will create a column with the `HEADERS` keyword. This column will be populated with the full list of header keys and values. There can only be one HEADERS column.
+```
+ksql> CREATE STREAM A (
+    my_headers ARRAY<STRUCT<key STRING, value BYTES>> HEADERS);
+```
+If there are multiple HEADERS columns, then an error will be thrown. The type of a header backed column must be exactly `ARRAY<STRUCT<key STRING, value BYTES>>`. An error will be thrown if that is not the case.
```
mjsax, line 32: > If there are multiple HEADERS columns, then an error will be thrown. 

Should we include the error message example, similar to the second case of an incorrect data type?

mjsax (approved):

design-proposals/klip-57-kafka-headers.md
```diff
@@ -0,0 +1,141 @@
...
+
+Sources created with schema inferencing can have header columns, they would just have to be specified. For example, the statement `CREATE STREAM A (my_headers ARRAY<STRUCT<key STRING, value BYTES>> HEADERS) WITH (kafka_topic='a', value_format='avro');` will create a stream with columns from Schema Registry, as well as one header column.
+
+### New byte-conversion functions
+
+There will also be new byte conversion functions to help decode header data:
+
+* `INT_FROM_BYTES(bytes, byte_order='big', signed=true)` - The input byte array must be exactly 4 bytes long. Signed conversions will by handled by `ByteBuffer`'s `getInt` method. Unsigned conversions will be manually calculated. If the result does not fit into an INTEGER then the method will throw an error.
+* `BIGINT_FROM_BYTES(bytes, byte_order='big', signed=true)` - The input byte array must be exactly 8 bytes long. Signed conversions will by handled by `ByteBuffer`'s `getLong` method. Unsigned conversions will be manually calculated. If the result does not fit into a BIGINT then the method will throw an error.
```
spena, line 50: ksql does not support unsigned integers, so I don't think we can use the `signed` parameter. In java, for instance, unsigned integers are stored in long values in order to fit the unsigned value there.

Is this `signed` a requirement?

A similar question about the byte order. Seems any java API to convert to integer uses big endian. Are we expecting users to use different orders than just using the API to convert from int -> byte array?

jzaralim: Users are limited in what they are able to convert their bytes to. The problem with signed/unsigned is that the byte representations of the two are completely different, so if we only support signed, then a user who wants to do an unsigned conversion would have no way to do it. We could get around this by adding byte arithmetic, but that would still be a big headache to do.

spena: The ksql `INT` type cannot hold an unsigned integer. When ksql manages `INT` types, it uses the `java.lang.Integer` which holds only `-214748364` to `2147483647`. An unsigned integer should hold `0` to `4294967295`, which is not what `INT` can support.

The option should be to support `INT UNSIGNED` types, which is a bigger scope.

jzaralim: The `If the result does not fit into an INTEGER then the method will throw an error` is supposed to cover that. I can see how it's awkward.

@colinhicks are `UNSIGNED` numbers or bit manipulation on the roadmap? Either of those would give users a workaround.

As for the byte order, it's easy for us to flip a byte array and some applications do use little endian, so I think we should keep it.

PR #8303, opened 2021-10-28
docs: update klip-56

lihaosky:
### Description 
Update klip-56 with more detailed design.

### Testing done 
test in prod

### Reviewer checklist

### DIFF
--- a/design-proposals/klip-56-schema-id-in-create-statements.md
+++ b/design-proposals/klip-56-schema-id-in-create-statements.md
@@ -116,4 +116,5 @@ Add newly supported properties to:
 ## Security Implications
 No
 
+## References
 [^1]: [Avro Kafka Serializer](https://github.com/confluentinc/schema-registry/blob/802dc1a4dfbca7919375e28dfe2cbe793ae57e0c/avro-serializer/src/main/java/io/confluent/kafka/serializers/AbstractKafkaAvroSerializer.java#L116)
--- a/design-proposals/klip-56-schema-id-in-create-statements.md
+++ b/design-proposals/klip-56-schema-id-in-create-statements.md
@@ -54,7 +54,8 @@ partitions=1, key_format='avro', value_format='avro');
 * `CS/CT` command
   * Corresponding `key_format`/`value_format` or `format` property must exist and the format must support `SCHEMA_INFERENCE` (protobuf, avro, json_sr format currently).
   * The fetched schema format from _schema_registry_ must match specified format in `WITH` clause. For example, if schema format for schema ID in _schema_registry_ is `avro` but specified format in `WITH` clause is `protobuf`, an exception will be thrown.
-  * Schema with specified ID MUST exist in _schema_registry_, otherwise an exception will be thrown. Format of schema fetched from _schema registry_ must match the defined format in statement.
+  * Schema with specified ID MUST exist in _schema_registry_, otherwise an exception will be thrown. 
+  * Corresponding key/value _table elements_ must be empty if `key_schema_id` or `value_schema_id` is provided.
 
 * `CAS` command
   * For `key_format` and `value_format` properties, if `*_schema_id` is provided:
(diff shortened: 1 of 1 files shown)


design-proposals/klip-56-schema-id-in-create-statements.md
```diff
@@ -57,7 +57,7 @@ partitions=1, key_format='avro', value_format='avro');
   * Compatibility Checks. See section below.
 
 * `C*AS` command
-  * `key_format` and `value_format` are optional. If specified, they must match format fetched from _schema_registry_.
+  * Corresponding `key_format` or `value_format` must exist when `*_schema_id` property is present and the format must support `SCHEMA_INFERENCE` (protobuf, avro, json_sr format currently).
```
vcrfxia, line 60: Sorry, what's the motivation for making this change? Requiring `key_format` and/or `value_format` under these circumstances is a departure from all other types of statements where they are optional as long as there are acceptable server defaults.

lihaosky: The main reasons are:
1. Consistent with requirement of CS/CT command where it’s required which can make it explicit what format user is expecting.
2. Key SerdeFeatures which needed for translation are built from format: https://github.com/confluentinc/ksql/blob/master/ksqldb-engine/src/main/java/io/confluent/ksql/schema/ksql/inference/DefaultSchemaInjector.java#L130. it will also be inferred if we use SR to get format.  This is also implicit.

I agree if there's defaults these could be optional.

vcrfxia: I didn't realize this change was specifically for `C*AS` statements -- the server defaults I was referring to are primarily for `CS/CT` statements instead. For `C*AS` statements, the default is to inherit the key and value format from the source stream/table. Is there a reason we can't do that even when `schema_id` is provided?

vcrfxia: In other words, is there a technical/implementation reason for requiring this, or are we doing it because we think it's more intuitive for users? If the latter, I actually think it's less intuitive for users since users are never required to specify formats for `C*AS` statements otherwise.

lihaosky: If `schema_id` is provided, I guess we want to use the format defined in schema_id's schema instead of format from source stream/table? cc @colinhicks

colinhicks: My understanding had been that there was a technical/implementation reason for requiring the format to also be specified when schema_id is present. If that's not the case we have more options. But let's talk through the full range of scenarios.

For `C*AS` statements, where `{key, value}_schema_id` is provided:
* If no corresponding `{key, value}_format` is specified via `WITH`, the format from the source is used. This format must match the format of the schema stored in Schema Registry.
* If `{key, value}_format` is also specified via `WITH`, the explicit format is used. This format must match the format of the schema stored in Schema Registry.
* If the format of `{key, value}_schema_id` differs from the source format, the corresponding, explicit format must be set via `WITH`.
  * e.g. given a source stream `A` of value_format='json' and an Avro schema id=123, I must specify `CREATE STREAM B WITH(value_schema_id=123, value_format='avro') AS SELECT * FROM A;`. If the value_format is not provided in this case, I should see an error when created the persistent query.

vcrfxia: What @colinhicks laid out above is the behavior I would have expected as a user. This differs from the proposed update in the KLIP which says that `{key, value}_format` will always be required if `{key, value}_schema_id` is provided in the WITH clause of a `C*AS`.

lihaosky: OK. I will try to get the format from source.

colinhicks (approved):

spena (approved):

PR #8307, opened 2021-11-01
chore: add deprecation warning/notice to left/outer joins without grace period

spena:
### Description 
Return and display a deprecation notice for old left/outer stream-stream join semantics. The warning message will be displayed only for CREATE_AS statements. Push queries stream data from Websockets, which do not return warning statements.

```
ksql> create stream s3 as select * from s1 left join s2 within 1 second on s1.id = s2.id;

 Message                          
----------------------------------
 Created query with ID CSAS_S3_29 
----------------------------------
WARNING: DEPRECATION NOTICE: Left/Outer stream-stream joins statements without a GRACE PERIOD will not be accepted in a future ksqlDB version.
Please use the GRACE PERIOD clause as specified in https://docs.ksqldb.io/en/latest/developer-guide/ksqldb-reference/select-push-query/
```

### Testing done 

### Reviewer checklist


ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/computation/DistributingExecutor.java
```diff
@@ -241,6 +251,64 @@ public StatementExecutorResponse execute(
...
+
+    return deprecatedWarnings;
+  }
+
+  private boolean isLeftOrOuterStreamStreamJoinWithoutGrace(
+      final MetaStore metaStore,
+      final Statement statement
+  ) {
+    // Check only CREATE_AS statements.
+    // Pull and push queries are not written in the command topic.
```
mjsax, line 281: Not sure if I can follow? While I understand that we cannot send back a warning for push queried right now (because we use web sockets), I don't see the relationship to the command topic?

Also, if we deprecate something else, we should also be able to send a warning back for pull queries (and after we moved off web sockets also for push queries)?

ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/computation/DistributingExecutor.java
```diff
@@ -214,7 +223,8 @@ public StatementExecutorResponse execute(
           injected.getStatementText(),
           queuedCommandStatus.getCommandId(),
           commandStatus,
-          queuedCommandStatus.getCommandSequenceNumber()
+          queuedCommandStatus.getCommandSequenceNumber(),
+          getDeprecatedWarnings(executionContext.getMetaStore(), injected)
```
mjsax, line 227: Why do we need to write the warning into the command topic? -- I thought we want to send it back to the client that issue the statement?

ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/computation/DistributingExecutor.java
```diff
@@ -241,6 +251,64 @@ public StatementExecutorResponse execute(
...
+      final Statement statement
+  ) {
+    // Check only CREATE_AS statements.
+    // Pull and push queries are not written in the command topic.
+    if (!(statement instanceof CreateAsSelect)) {
+      return false;
+    }
+
+    if (((CreateAsSelect) statement).getQuery().getFrom() instanceof Join) {
+      final Join join = (Join) ((CreateAsSelect) statement).getQuery().getFrom();
```
mjsax, line 287: What about n-way joins?

ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/computation/DistributingExecutor.java
```diff
@@ -241,6 +251,64 @@ public StatementExecutorResponse execute(
...
+
+      // Check join sources are streams only
```
mjsax, line 289: What about n-way joins: There could we one stream-stream join followed by a stream-table join.

Gerrrr:
Just for my understanding, what is the reason we deprecate left/outer joins on streams?

ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/computation/DistributingExecutor.java
spena: Sorry, I didn't explain very well in that comment. The `DistributingExecutor` is the one that distributes commands through the command topic; waits for an answer; then responds back to the client. Only DDL commands are distributed because they change the metastore state of every node. Push and pull queries do not follow that path, so the reason this code only handles CREATE_AS operations. I think it was lazy to write all the above lines. Just wanted to mention this code path does not handle push/pull queries and it won't as that belongs somewhere else.

ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/computation/DistributingExecutor.java
spena: This is the `StatementExecutorResponse`. It is not written to the command topic but sent back to the client.

mjsax:
We don't deprecate the join :) -- Atm, we allow user to write:
```sql
FROM leftStream JOIN rightStream WITHIN ... ON ...
```
ie, there is no `GRACE PERIOD` clause that we introduced.

We want to make `GRACE PERIOD` mandatory, thus, we want inform user that they use old syntax with no grace clause and that they should add a the new clause.

ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/computation/DistributingExecutor.java
mjsax: What I am still wondering: why can not have a single place in the code that would add warning for all query types? Seems, if we add the code here, we would need to duplicate it elsewhere in the future?

spena: Done. While fixing the n-way joins, I moved the code to a different class `DeprecatedStatementsChecker` that checks if a statement is deprecated or not. We can reuse this class for push/pull queries later.

ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/computation/DistributingExecutor.java
spena: Done

Gerrrr:
Thank you for the answer @mjsax!

ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/computation/DeprecatedStatementsChecker.java
```diff
@@ -0,0 +1,122 @@
...
+            return true;
+          }
+        }
+      }
+    }
+
+    return false;
+  }
+
+  private boolean isStream(final Relation relation) {
```
mjsax, line 107: Not sure if I can follow?

If the left join node is a stream-table join, one source would be a table, but the result would still be a stream?

ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/computation/DeprecatedStatementsChecker.java
```diff
@@ -0,0 +1,122 @@
...
+
+    return false;
+  }
+
+  private boolean isLeftOrOuterStreamStreamJoinWithoutGraceQuery(final Query query) {
+    if (query.getFrom() instanceof Join) {
+      final Join join = (Join) query.getFrom();
+
+      // check left joined source is a stream
+      if (!isStream(join.getLeft())) {
```
mjsax, line 85: Why can we exit early, if the left input is not a stream? -- Even if we don't support "busy" plans right now, it would be good to write generic logic IMHO to avoid future bugs.

ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/computation/DeprecatedStatementsChecker.java
```diff
@@ -0,0 +1,122 @@
...
+        return false;
+      }
+
+      for (final JoinedSource joinedSource : join.getRights()) {
+        final JoinedSource.Type joinType = joinedSource.getType();
+        if (joinType == JoinedSource.Type.LEFT || joinType == JoinedSource.Type.OUTER) {
```
mjsax, line 91: Why do we care bout the inner/left/outer? I think GRACE should be mandatory for all cases?

ksqldb-rest-app/src/test/java/io/confluent/ksql/rest/server/computation/DeprecatedStatementsCheckerTest.java
```diff
@@ -0,0 +1,357 @@
...
+    // When
+    final Optional<DeprecatedStatementsChecker.Deprecations> deprecations =
+        statementsChecker.checkStatement(statement);
+
+    // Then
+    assertThat(deprecations, is(Optional.of(DEPRECATED_STREAM_STREAM_OUTER_JOIN_WITH_NO_GRACE)));
+  }
+
+  @Test
+  public void shouldDeprecateStreamStreamLeftJoinWithNoGraceOnMultiStreamStreamTableJoin() {
```
mjsax, line 115: Can we replicate this test with different join order? We should be able to detect both cases and warn for each.

ksqldb-rest-app/src/test/java/io/confluent/ksql/rest/server/computation/DeprecatedStatementsCheckerTest.java
```diff
@@ -0,0 +1,357 @@
...
+    // When
+    final Optional<DeprecatedStatementsChecker.Deprecations> deprecations =
+        statementsChecker.checkStatement(statement);
+
+    // Then
+    assertThat(deprecations, is(Optional.of(DEPRECATED_STREAM_STREAM_OUTER_JOIN_WITH_NO_GRACE)));
+  }
+
+  @Test
+  public void shouldDeprecateStreamStreamLeftJoinWithNoGraceOnMultiStreamStreamJoin() {
```
mjsax, line 95: Can we replicate this test with different join order? We should be able to detect both cases and warn for each.

ksqldb-rest-app/src/test/java/io/confluent/ksql/rest/server/computation/DeprecatedStatementsCheckerTest.java
```diff
@@ -0,0 +1,357 @@
...
+    // When
+    final Optional<DeprecatedStatementsChecker.Deprecations> deprecations =
+        statementsChecker.checkStatement(statement);
+
+    // Then
+    assertThat(deprecations, is(Optional.empty()));
+  }
+
+  @Test
+  public void shouldDeprecateStreamStreamOuterJoinWithoutGrace() {
```
mjsax, line 193: If we treat all three joins (inner/left/outer) equal as proposed about, those tests seems to be redundant?

spena:
@mjsax I addressed all your comments. Could you take a look?

ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/computation/DeprecatedStatementsChecker.java
```diff
@@ -0,0 +1,118 @@
...
+import io.confluent.ksql.parser.tree.Relation;
+import io.confluent.ksql.parser.tree.Statement;
+import io.confluent.ksql.parser.tree.WithinExpression;
+import java.util.Objects;
+import java.util.Optional;
+
+public class DeprecatedStatementsChecker {
+  public enum Deprecations {
+    DEPRECATED_STREAM_STREAM_OUTER_JOIN_WITH_NO_GRACE(
+        "DEPRECATION NOTICE: Left/Outer stream-stream joins statements without a GRACE PERIOD "
```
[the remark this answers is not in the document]
spena: Done.

ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/computation/DeprecatedStatementsChecker.java
```diff
@@ -0,0 +1,118 @@
...
+import io.confluent.ksql.parser.tree.Query;
+import io.confluent.ksql.parser.tree.Relation;
+import io.confluent.ksql.parser.tree.Statement;
+import io.confluent.ksql.parser.tree.WithinExpression;
+import java.util.Objects;
+import java.util.Optional;
+
+public class DeprecatedStatementsChecker {
+  public enum Deprecations {
+    DEPRECATED_STREAM_STREAM_OUTER_JOIN_WITH_NO_GRACE(
```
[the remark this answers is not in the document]
spena: Done

ksqldb-rest-app/src/test/java/io/confluent/ksql/rest/server/computation/DeprecatedStatementsCheckerTest.java
```diff
@@ -0,0 +1,269 @@
...
+              + "%s JOIN %s AS r WITHIN 1 SECOND ON m.K = r.K "
+              + "%s JOIN %s AS m WITHIN 1 SECOND GRACE PERIOD 1 SECOND ON l.K = m.K;",
+          STREAM_1.text(),
+          (joinType == JoinType.OUTER) ? "FULL OUTER" : joinType,
+          STREAM_2.text(),
+          (joinType == JoinType.OUTER) ? "FULL OUTER" : joinType,
+          STREAM_3.text()
+      ));
+    }
+  }
```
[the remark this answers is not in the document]
spena: There's a test that does already
```
@Test
  public void shouldNotDeprecateStreamStreamJoinWithGraceOnMultiStreamStreamJoin() {
    for (final JoinType joinType : JoinType.values()) {
      checkNoDeprecatedStatement(String.format(
          "SELECT * FROM %s AS l "
              + "%s JOIN %s AS m WITHIN 1 SECOND GRACE PERIOD 1 SECOND ON l.K = m.K "
              + "%s JOIN %s AS r WITHIN 1 SECOND GRACE PERIOD 1 SECOND ON m.K = r.K;",
          STREAM_1.text(),
          (joinType == JoinType.OUTER) ? "FULL OUTER" : joinType,
          STREAM_2.text(),
          (joinType == JoinType.OUTER) ? "FULL OUTER" : joinType,
          STREAM_3.text()
      ));
    }
  }
```

ksqldb-rest-app/src/test/java/io/confluent/ksql/rest/server/computation/DeprecatedStatementsCheckerTest.java
```diff
@@ -0,0 +1,269 @@
...
+
+  @Test
+  public void shouldDeprecateStreamStreamJoinsWithNoGrace() {
+    for (final JoinType joinType : JoinType.values()) {
+      checkDeprecatedStatement(String.format(
+          "SELECT * FROM %s AS l "
+              + "%s JOIN %s AS r WITHIN 1 SECOND ON l.K = r.K;",
+          STREAM_1.text(),
+          (joinType == JoinType.OUTER) ? "FULL OUTER" : joinType,
```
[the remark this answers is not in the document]
spena: Right, `OUTER` is not supported, only `FULL OUTER`.

ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/computation/DeprecatedStatementsChecker.java
```diff
@@ -0,0 +1,118 @@
...
+    // DataSourceExtractor must be initialized everytime we need to extract data from a node
+    // because it changes its internal state to keep all sources found
+    final DataSourceExtractor dataSourceExtractor = new DataSourceExtractor(metaStore, false);
+
+    // The relation object should have only one joined source, but the extract sources method
+    // returns a list. This loop checks all returned values are a Stream just to prevent throwing
+    // an exception if more than one value is returned.
+    for (final AliasedDataSource source : dataSourceExtractor.extractDataSources(relation)) {
+      if (source.getDataSource().getDataSourceType() != DataSource.DataSourceType.KSTREAM) {
+        return false;
```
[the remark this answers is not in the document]
spena: For what I understand, the `relation` references one join only, no multiple. But, the code is generic and this `relation` has a parent that may be another join. However, this shouldn't happen at this point but that's how the code is. I was going to use `Iterables.getOnlyElement`, but I didn't want to throw an exception if I misunderstood the code or a bug is introduce in the future. Being this just a checker for a deprecation notice, I felt I should do this weird recursive work that will never have more joins on it. It's confusing. I'm not sure what's the best approach to do here.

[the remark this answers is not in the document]
mjsax: > this `relation` has a parent that may be another join.

Yes, but isn't the parent is a `relation`, too?

> However, this shouldn't happen at this point but that's how the code is.

Not sure if I can follow?

ksqldb-rest-app/src/test/java/io/confluent/ksql/rest/server/computation/DeprecatedStatementsCheckerTest.java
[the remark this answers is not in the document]
mjsax: Filed https://github.com/confluentinc/ksql/issues/8478

ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/computation/DeprecatedStatementsChecker.java
```diff
@@ -0,0 +1,118 @@
...
+
+  private boolean isJoinedSourceStream(final Relation relation) {
+    // DataSourceExtractor must be initialized everytime we need to extract data from a node
+    // because it changes its internal state to keep all sources found
+    final DataSourceExtractor dataSourceExtractor = new DataSourceExtractor(metaStore, false);
+
+    // The relation object should have only one joined source, but the extract sources method
+    // returns a list. This loop checks all returned values are a Stream just to prevent throwing
+    // an exception if more than one value is returned.
+    for (final AliasedDataSource source : dataSourceExtractor.extractDataSources(relation)) {
```
mjsax, line 110: I agree to the sentiment to not crash. But I am wondering if we should log a WARN if there is more than one source found?

ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/computation/DeprecatedStatementsChecker.java
```diff
@@ -0,0 +1,118 @@
...
+        if (!joinedSource.getWithinExpression().flatMap(WithinExpression::getGrace).isPresent()) {
+          return true;
+        }
+      }
+    }
+
+    return false;
+  }
+
+  private boolean isJoinedSourceStream(final Relation relation) {
```
mjsax, line 102: Nit: maybe rename to `isStream`, because the actual comparison is `source.getDataSource().getDataSourceType() != DataSource.DataSourceType.KSTREAM`?

ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/computation/DeprecatedStatementsChecker.java
```diff
@@ -0,0 +1,118 @@
...
+
+      // check left joined source is a stream
+      if (!isJoinedSourceStream(join.getLeft())) {
+        return false;
+      }
+
+      for (final JoinedSource joinedSource : join.getRights()) {
+        // check right joined source is a stream
+        if (!isJoinedSourceStream(joinedSource.getRelation())) {
+          break;
```
mjsax, line 90: Should this be `continue;` ? If we have a stream-table joins followed by a stream-stream join, the first `JoinedSource` in the list could not contain a STREAM but a TABLE as input, but we still need to check the second join, ie, third input, too?

Seems there is test for this case, but I think the test is setup incorrect (cf comment below)? Or I miss something?

ksqldb-rest-app/src/test/java/io/confluent/ksql/rest/server/computation/DeprecatedStatementsCheckerTest.java
```diff
@@ -0,0 +1,269 @@
...
+  @Test
+  public void shouldDeprecateStreamStreamJoinsWithNoGraceOnMultiStreamStreamTableJoinOrder2() {
+    for (final JoinType joinType : JoinType.values()) {
+      checkDeprecatedStatement(String.format(
+          "SELECT * FROM %s AS l "
+              + "%s JOIN %s AS r ON l.K = r.K "
+              + "%s JOIN %s AS m WITHIN 1 SECOND ON l.K = m.K;",
+          STREAM_1.text(),
+          (joinType == JoinType.OUTER) ? "FULL OUTER" : joinType,
+          STREAM_2.text(),
```
mjsax, line 147: Should this be `TABLE_1` ?

ksqldb-rest-app/src/test/java/io/confluent/ksql/rest/server/computation/DeprecatedStatementsCheckerTest.java
```diff
@@ -0,0 +1,269 @@
...
+          (joinType == JoinType.OUTER) ? "FULL OUTER" : joinType,
+          TABLE_1.text()
```
mjsax, line 149: If we pass in `TABLE_1` as last argument, should it not crash, as a stream-table join should not have a `WITHIN` clause?

ksqldb-rest-app/src/test/java/io/confluent/ksql/rest/server/computation/DeprecatedStatementsCheckerTest.java
```diff
@@ -0,0 +1,269 @@
...
+          (joinType == JoinType.OUTER) ? "FULL OUTER" : joinType,
+          STREAM_2.text(),
+          (joinType == JoinType.OUTER) ? "FULL OUTER" : joinType,
+          STREAM_3.text()
+      ));
+    }
+  }
+
+  @Test
+  public void shouldDeprecateStreamStreamJoinsWithNoGraceOnMultiStreamStreamTableJoinOrder1() {
```
mjsax, line 123: Should we replicate this test and the next one as "shouldNotWarn" and use GRACE in the stream-stream join?

spena: There was one that I renamed to `shouldNotDeprecateStreamStreamJoinsWithGraceOnMultiStreamStreamTableJoinOrder1`. I added another with the name `shouldNotDeprecateStreamStreamJoinsWithGraceOnMultiStreamStreamTableJoinOrder2`.

ksqldb-rest-app/src/test/java/io/confluent/ksql/rest/server/computation/DeprecatedStatementsCheckerTest.java
spena: Yes. done

ksqldb-rest-app/src/test/java/io/confluent/ksql/rest/server/computation/DeprecatedStatementsCheckerTest.java
spena: I just fixed the deprecation checker and changed this parameter.

ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/computation/DeprecatedStatementsChecker.java
spena: True. I switched to continue. The new test I added verifies this.

ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/computation/DeprecatedStatementsChecker.java
spena: Done

ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/computation/DeprecatedStatementsChecker.java
spena: Added

mjsax (approved):
Thanks for the PR and sticking with me... LGTM.

Seems Jenkins failed with a flaky test. Triggered a re-build.

PR #6720, opened 2020-12-04
docs: adds border to cards

MichaelDrogalis:
### Description 

Minor change to make the cards look easier on the eyes.

PR #8310, opened 2021-11-03
fix: dont throw error on processing local commands, just log 

nateab:
### Description 
We had multiple on call incidents where a race condition occurs when writing to the local commands file which corrupts it. Processing this corrupt local commands file on startup kills the ksqldb app. The local commands is meant to only be a best effort attempt at cleaning up orphaned transient topics #6714 and abandoned state stores #6720, so we should not kill the ksqldb app if we fail to clean these up. Instead of throwing an error we now simply just log the error and continue on. 

Follow up work will be done to log the race conditions that result in the corrupted local commands file. 

Updated unit tests

### DIFF
--- a/ksqldb-rest-app/src/test/java/io/confluent/ksql/rest/server/LocalCommandsTest.java
+++ b/ksqldb-rest-app/src/test/java/io/confluent/ksql/rest/server/LocalCommandsTest.java
@@ -187,7 +187,7 @@ public class LocalCommandsTest {
   }
 
   @Test
-  public void shouldFailToCleanup() throws IOException {
+  public void shouldNotFailToCleanup() throws IOException {
     // Given
     final File dir = commandsDir.newFolder();
     LocalCommands localCommands = LocalCommands.open(ksqlEngine, dir);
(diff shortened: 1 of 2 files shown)


ksqldb-rest-app/src/test/java/io/confluent/ksql/rest/server/LocalCommandsTest.java
```diff
@@ -187,7 +187,7 @@ public void shouldFailToWrite_error() throws IOException {
   }
 
   @Test
-  public void shouldFailToCleanup() throws IOException {
+  public void shouldNotFailToCleanup() throws IOException {
```
AlanConfluent, line 190: Might want to simulate errors on calling LocalCommandFile.readRecords() which I think is what happened in practice.

AlanConfluent (approved):

ksqldb-rest-app/src/test/java/io/confluent/ksql/rest/server/LocalCommandsTest.java
nateab: added

PR #8312, opened 2021-11-03
chore: passing consistency token using request properties in HARouting

patrickstuedi:
Following the practice introduced in #8264 this PR is using request properties to pass the consistency token between servers

### DIFF
--- a/ksqldb-engine/src/main/java/io/confluent/ksql/physical/pull/HARouting.java
+++ b/ksqldb-engine/src/main/java/io/confluent/ksql/physical/pull/HARouting.java
@@ -61,6 +61,7 @@ import java.util.stream.Collectors;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+@SuppressWarnings({"checkstyle:ClassDataAbstractionCoupling", "checkstyle:CyclomaticComplexity"})
 public final class HARouting implements AutoCloseable {
 
   private static final Logger LOG = LoggerFactory.getLogger(HARouting.class);
@@ -340,10 +341,15 @@ public final class HARouting implements AutoCloseable {
         .map(location -> Integer.toString(location.getPartition()))
         .collect(Collectors.joining(","));
     // Add skip forward flag to properties
-    final Map<String, Object> requestProperties = ImmutableMap.of(
-        KsqlRequestConfig.KSQL_REQUEST_QUERY_PULL_SKIP_FORWARDING, true,
-        KsqlRequestConfig.KSQL_REQUEST_INTERNAL_REQUEST, true,
-        KsqlRequestConfig.KSQL_REQUEST_QUERY_PULL_PARTITIONS, partitions);
+    final ImmutableMap.Builder<String, Object> builder = new ImmutableMap.Builder<String, Object>()
+        .put(KsqlRequestConfig.KSQL_REQUEST_QUERY_PULL_SKIP_FORWARDING, true)
+        .put(KsqlRequestConfig.KSQL_REQUEST_INTERNAL_REQUEST, true)
+        .put(KsqlRequestConfig.KSQL_REQUEST_QUERY_PULL_PARTITIONS, partitions);
+    if (consistencyOffsetVector.isPresent()) {
+      builder.put(KsqlRequestConfig.KSQL_REQUEST_QUERY_PULL_CONSISTENCY_OFFSET_VECTOR,
+          consistencyOffsetVector.get().serialize());
+    }
+    final Map<String, Object> requestProperties = builder.build();
     final RestResponse<Integer> response;
 
     try {
@@ -355,8 +361,7 @@ public final class HARouting implements AutoCloseable {
               statement.getSessionConfig().getOverrides(),
               requestProperties,
               streamedRowsHandler(owner, pullQueryQueue, rowFactory, outputSchema),
-              shouldCancelRequests,
-              consistencyOffsetVector.map(ConsistencyOffsetVector::serialize)
+              shouldCancelRequests
           );
     } catch (Exception e) {
       // If we threw some explicit exception, then let it bubble up. All of the row handling is
@@ -438,7 +443,11 @@ public final class HARouting implements AutoCloseable {
           }
 
           if (!row.getRow().isPresent()) {
-            throw new KsqlException("Missing row data on row " + i + " of chunk");
+            if (row.getConsistencyToken().isPresent()) {
+              continue;
+            } else {
+              throw new KsqlException("Missing row data on row " + i + " of chunk");
+            }
           }
 
           final List<?> r = row.getRow().get().getColumns();
--- a/ksqldb-rest-app/src/test/java/io/confluent/ksql/rest/integration/PullQueryConsistencyFunctionalTest.java
+++ b/ksqldb-rest-app/src/test/java/io/confluent/ksql/rest/integration/PullQueryConsistencyFunctionalTest.java
@@ -1,0 +1,472 @@
+/*
+ * Copyright 2020 Confluent Inc.
+ *
+ * Licensed under the Confluent Community License (the "License"); you may not use
+ * this file except in compliance with the License.  You may obtain a copy of the
+ * License at
+ *
+ * http://www.confluent.io/confluent-community-license
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OF ANY KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations under the License.
+ */
+
+package io.confluent.ksql.rest.integration;
+
+import static io.confluent.ksql.rest.integration.HighAvailabilityTestUtil.extractQueryId;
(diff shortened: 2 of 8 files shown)


ksqldb-engine/src/main/java/io/confluent/ksql/physical/pull/HARouting.java
```diff
@@ -336,10 +336,14 @@ private static void forwardTo(
         .map(location -> Integer.toString(location.getPartition()))
         .collect(Collectors.joining(","));
     // Add skip forward flag to properties
-    final Map<String, Object> requestProperties = ImmutableMap.of(
-        KsqlRequestConfig.KSQL_REQUEST_QUERY_PULL_SKIP_FORWARDING, true,
-        KsqlRequestConfig.KSQL_REQUEST_INTERNAL_REQUEST, true,
-        KsqlRequestConfig.KSQL_REQUEST_QUERY_PULL_PARTITIONS, partitions);
+    final Map<String, Object> requestProperties = new HashMap<>();
```
vpapavas, line 339: You can leave this as an `ImmutableMap` and use the `builder` to add an extra entry

vvcephei: +1

ksqldb-engine/src/main/java/io/confluent/ksql/physical/pull/HARouting.java
```diff
@@ -340,10 +341,14 @@ private static void forwardTo(
...
+    requestProperties.put(KsqlRequestConfig.KSQL_REQUEST_QUERY_PULL_SKIP_FORWARDING, true);
+    requestProperties.put(KsqlRequestConfig.KSQL_REQUEST_INTERNAL_REQUEST, true);
+    requestProperties.put(KsqlRequestConfig.KSQL_REQUEST_QUERY_PULL_PARTITIONS, partitions);
+    if (consistencyOffsetVector.isPresent()) {
+      requestProperties.put(KsqlRequestConfig.KSQL_REQUEST_QUERY_PULL_CONSISTENCY_OFFSET_VECTOR,
```
vvcephei, line 349: I think we also have to add the "consistency vector enabled" config, right, @vpapavas ?

Which raises the question: is this logic already being tested by the existing tests from your prior PR? (If so, and we are missing the other config, then how is that test passing?)

vvcephei (commented):
Thanks, @patrickstuedi !

Just one thing to add on top of @vpapavas 's comment.

ksqldb-engine/src/main/java/io/confluent/ksql/physical/pull/HARouting.java
vpapavas: On the server side we check that:

1. The server config KsqlConfig.KSQL_QUERY_PULL_CONSISTENCY_OFFSET_VECTOR_ENABLED is set to true
2. The request properties contain a consistency token (may be an empty string the first time a request is made)

In the tests, @patrickstuedi has set the server property to true hence only the offset needs to be added.

We don't additionally need to check if the client has set the "consistency vector enabled" config, since the only way to add the vector to the request properties, is if this config is set. In this case here, the Optional will be non-empty, only if consistency is enabled

ksqldb-rest-app/src/test/java/io/confluent/ksql/rest/integration/PullQueryConsistencyFunctionalTest.java
```diff
@@ -0,0 +1,460 @@
...
+/**
+ * Test to ensure pull queries route across multiple KSQL nodes correctly.
+ *
+ * <p>For tests on general syntax and handled see RestQueryTranslationTest's
+ * materialized-aggregate-static-queries.json
+ */
+@SuppressWarnings("OptionalGetWithoutIsPresent")
+@Category({IntegrationTest.class})
+@RunWith(MockitoJUnitRunner.class)
+public class PullQueryConsistencyFunctionalTest {
```
vpapavas, line 98: General comment about this class: There is a bunch of code that is copied from `PullQueryRoutingFunctionalTest`. How about moving the common code to the `HighAvailabilityTestUtil` class?

vvcephei: This is objectively good advice.

Just FYI, I could also get behind keeping them independent so that the tests can evolve independently without affecting each other. I think the real deciding factor isn't how much code is similar between the two classes, but how much hassle do you anticipate in the future if we keep them independent or if we create a common util class?

For example, if there's a lot of common code, but it's all pretty straightforward, and you think it'll just sit there not hurting anyone for the next 5 years, then I'd say don't bother combining them. But if it looks like the common code is pretty tricky and you think any change to one test would imply we should also change the other test, then it's better to combine the common code.

In other words, you've got my opinion, and I'm happy whichever way you decide to go.

vvcephei (approved):
LGTM, mod @vpapavas 's open question.

ksqldb-rest-app/src/test/java/io/confluent/ksql/rest/integration/PullQueryConsistencyFunctionalTest.java
patrickstuedi: Ok, thanks for both comments @vpapavas @vvcephei . Btw, one reason I didn't extend PullQueryRoutingFunctionalTest in the first place is because all those tests expect 1 partition, and I needed multiple partitions and there is no way to mock the "before" part depending on which tests is executed.

ksqldb-rest-app/src/test/java/io/confluent/ksql/rest/integration/PullQueryConsistencyFunctionalTest.java
```diff
@@ -0,0 +1,460 @@
...
+      clusterFormation.router.getApp(),
+      clusterFormation.standBy.getHost(),
+      HighAvailabilityTestUtil::remoteServerIsUp, USER_CREDS);
+
+    // When:
+    final List<StreamedRow> rows_0 = makeRequestIncludingConsistencyToken(clusterFormation.router.getApp(),
+      sqlTableScan, null, USER_CREDS);
+
+    // Then:
+    assertThat(rows_0, hasSize(HEADER + 5));
```
guozhangwang, line 319: Just curious, could we check the actual row content, more than just the size?

ksqldb-rest-app/src/test/java/io/confluent/ksql/rest/integration/PullQueryConsistencyFunctionalTest.java
```diff
@@ -0,0 +1,460 @@
...
+
+    public KsqlHostInfoEntity getHost() {
+      return host;
+    }
+
+    public TestKsqlRestApp getApp() {
+      return app;
+    }
+
+    public Shutoffs getShutoffs() {
```
guozhangwang, line 430: This seems not used?

ksqldb-rest-app/src/test/java/io/confluent/ksql/rest/integration/PullQueryConsistencyFunctionalTest.java
patrickstuedi: @vpapavas I was playing around with moving common parts around, but it's tricky, you move one thing then we need to move another thing et etc, eventually we need to move TestApps and the actual number of it, etc, all things that we might want to change in the future and certain things are all already changed

I think it would be best to keep the copy of those infra calls currently

ksqldb-rest-app/src/test/java/io/confluent/ksql/rest/integration/PullQueryConsistencyFunctionalTest.java
patrickstuedi: Good catch @guozhangwang . The thing is, this test class is where lots of the consistency tests will eventually go in. Currently is just checking that forwarding and returning of the token works, but eventually we want to test consistency tokens are handled as expected, also in the presence of shutdown/rebalance/etc, so I prefer to keep those infra calls for the moment.

ksqldb-rest-app/src/test/java/io/confluent/ksql/rest/integration/PullQueryConsistencyFunctionalTest.java
```diff
@@ -0,0 +1,466 @@
...
+      .skip(HEADER)
+      .filter(sr -> sr.getRow().isPresent())
+      .map(sr -> {
+        return sr.getRow().get().getColumns();
+      })
+      .collect(Collectors.toList());
+
+    // Then:
+    assertThat(rows, hasSize(HEADER + 6));
+    assertThat(rows.get(HEADER+5).getConsistencyToken(), is(not(Optional.empty())));
```
vpapavas, line 331: What you actually want to check is that the consistency token contains the information you expect i.e., it did not get garbled after all the fan out and response.

vpapavas (approved):
Thank you @patrickstuedi ! LGTM just add the check in the test that the CV is actually correct

PR #8313, opened 2021-11-03
chore: bump command version from 10 to 11 for v0.23

mjsax:
### Description 
Command version in 0.22.0 release is 10. We need to bump it to 11 for 0.23.


nateab (approved):
LGTM!

PR #8314, opened 2021-11-03
feat: add HEADERS column type to syntax

jzaralim:
### Description 
Adds `HEADERS` as a column type (like `KEY` or `PRIMARY KEY`) as a part of https://github.com/confluentinc/ksql/pull/8293

At some point in the future, changes will be made to
1. Reject `HEADERS` columns that are not of type `ARRAY<STRUCT<key STRING, value BYTES>>`
2. Add `HEADER(key)` to the syntax

### Testing done 
Unit tests

### Reviewer checklist

### DIFF
--- a/ksqldb-parser/src/main/antlr4/io/confluent/ksql/parser/SqlBase.g4
+++ b/ksqldb-parser/src/main/antlr4/io/confluent/ksql/parser/SqlBase.g4
@@ -127,7 +127,11 @@ tableElements
     ;
 
 tableElement
-    : identifier type ((PRIMARY)? KEY)?
+    : identifier type (((PRIMARY)? KEY) | headerType)?
+    ;
+
+headerType
+    : HEADERS
     ;
 
 tableProperties
@@ -538,6 +542,7 @@ ADD: 'ADD';
 ALTER: 'ALTER';
 VARIABLES: 'VARIABLES';
 PLUGINS: 'PLUGINS';
+HEADERS: 'HEADERS';
 
 IF: 'IF';
 
--- a/ksqldb-parser/src/main/java/io/confluent/ksql/parser/tree/TableElement.java
+++ b/ksqldb-parser/src/main/java/io/confluent/ksql/parser/tree/TableElement.java
@@ -52,6 +52,15 @@ public final class TableElement extends AstNode {
     public boolean isKey() {
       return this == KEY || this == PRIMARY_KEY;
     }
+
+    public static Namespace of(final SqlBaseParser.TableElementContext context) {
+      if (context.headerType() != null) {
+        return Namespace.HEADERS;
+      }
+      return context.KEY() == null
+          ? Namespace.VALUE
+          : context.PRIMARY() == null ? Namespace.KEY : Namespace.PRIMARY_KEY;
+    }
   }
 
   private final Namespace namespace;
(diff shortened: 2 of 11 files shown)


ksqldb-parser/src/main/java/io/confluent/ksql/parser/tree/TableElement.java
```diff
@@ -128,4 +133,13 @@ public String toString() {
         + ", namespace=" + namespace
         + '}';
   }
+
+  public static Namespace getNamespaceFromContext(final SqlBaseParser.TableElementContext context) {
+    if (context.headerType() != null) {
```
mjsax, line 138: Would it be simpler to use a `switch` statement?

ksqldb-parser/src/test/java/io/confluent/ksql/parser/AstBuilderTest.java
```diff
@@ -866,4 +868,17 @@ public void shouldFailOnBigIntWithParameter() {
             () -> givenQuery("CREATE STREAM INPUT (K BIGINT (KEY)) WITH (kafka_topic='input',value_format='JSON');")
     );
   }
+
+  @Test
+  public void shouldSupportHeadersColumns() {
+    // Given:
+    final SingleStatementContext stmt
+        = givenQuery("CREATE STREAM INPUT (K BIGINT HEADERS) WITH (kafka_topic='input',value_format='JSON');");
```
mjsax, line 876: This statement raises interesting questions:
- should we allow a "header only" schema definition (I think, yes, why not?)
- if yes, why would we need a `value_format` property for this case?

mjsax (commented):
Not super familiar with making changes to the grammar. A few high level comments.

ksqldb-parser/src/test/java/io/confluent/ksql/parser/AstBuilderTest.java
jzaralim: That is a good point. Right now, [all sources](https://docs.ksqldb.io/en/latest/developer-guide/ksqldb-reference/create-stream/#stream-properties) sources require the key_format and value_format properties, even if it doesn't actually have a key/value. It would be nice if they are not needed, but that's a separate issue to fix.

ksqldb-parser/src/main/java/io/confluent/ksql/parser/tree/TableElement.java
jzaralim: How these `...Context` objects work is for each possible token that could be in the element, there is a field that either contains an object or is `null`. So we have to go through each field to see if it's there.

ksqldb-parser/src/main/antlr4/io/confluent/ksql/parser/SqlBase.g4
```diff
@@ -127,7 +127,11 @@ tableElements
     ;
 
 tableElement
-    : identifier type ((PRIMARY)? KEY)?
+    : identifier type (((PRIMARY)? KEY) | headerType)?
+    ;
+
+headerType
+    : HEADERS
     ;
```
spena, line 135: This type seems lonely. Could the `HEADERS` keyword be part of the `tableElement` instead of `headerType` (like the PRIMARY KEY)? Or are there other reasons to keep this?

ksqldb-parser/src/main/java/io/confluent/ksql/parser/tree/TableElement.java
```diff
@@ -128,4 +133,13 @@ public String toString() {
         + ", namespace=" + namespace
         + '}';
   }
+
+  public static Namespace getNamespaceFromContext(final SqlBaseParser.TableElementContext context) {
```
spena, line 137: Should this method be part of the `Namespace` enum instead? Perhaps a `Namespace.of(context)`

jzaralim: oh nice thats clean. moved

spena (approved):
+1

PR #8322, opened 2021-11-04
fix: if not exists return type

wcarlson5:
```
ksql> CREATE STREAM IF NOT EXISTS BAR AS SELECT * FROM FOO EMIT CHANGES;

 Message                                                              
----------------------------------------------------------------------
 Cannot add stream `BAR`: A stream with the same name already exists. 
----------------------------------------------------------------------
ksql> CREATE STREAM FOO (id INT KEY, bar VARCHAR) WITH(value_format='json', kafka_topic='foo', partitions=6);
Cannot add stream 'FOO': A stream with the same name already exists
ksql> 

```

### DIFF
--- a/ksqldb-api-client/src/main/java/io/confluent/ksql/api/client/impl/DdlDmlResponseHandlers.java
+++ b/ksqldb-api-client/src/main/java/io/confluent/ksql/api/client/impl/DdlDmlResponseHandlers.java
@@ -37,6 +37,10 @@ final class DdlDmlResponseHandlers {
       final JsonObject ksqlEntity,
       final CompletableFuture<ExecuteStatementResult> cf
   ) {
+    if (isIfNotExistsWarning(ksqlEntity)) {
+      cf.complete(new ExecuteStatementResultImpl(Optional.empty()));
+      return;
+    }
     if (!isCommandStatusEntity(ksqlEntity)) {
       handleUnexpectedEntity(ksqlEntity, cf);
       return;
@@ -66,6 +70,13 @@ final class DdlDmlResponseHandlers {
   private static boolean isCommandStatusEntity(final JsonObject ksqlEntity) {
     return ksqlEntity.getString("commandId") != null
         && ksqlEntity.getJsonObject("commandStatus") != null;
+  }
+
+  private static boolean isIfNotExistsWarning(final JsonObject ksqlEntity) {
+    return ksqlEntity.getString("statementText") != null
+        && ksqlEntity.getString("statementText").contains("IF NOT EXISTS")
+        && ksqlEntity.getString("message") != null
+        && ksqlEntity.getString("message").contains("A stream with the same name already exists.");
   }
 
   // CHECKSTYLE_RULES.OFF: CyclomaticComplexity
--- a/ksqldb-api-client/src/main/java/io/confluent/ksql/api/client/impl/DdlDmlResponseHandlers.java
+++ b/ksqldb-api-client/src/main/java/io/confluent/ksql/api/client/impl/DdlDmlResponseHandlers.java
@@ -38,7 +38,7 @@ final class DdlDmlResponseHandlers {
       final CompletableFuture<ExecuteStatementResult> cf
   ) {
     if (isIfNotExistsWarning(ksqlEntity)) {
-      cf.complete(new ExecuteStatementResultImpl(Optional.of(ksqlEntity.getString("message"))));
+      cf.complete(new ExecuteStatementResultImpl(Optional.empty()));
       return;
     }
     if (!isCommandStatusEntity(ksqlEntity)) {
--- a/ksqldb-api-client/src/main/java/io/confluent/ksql/api/client/impl/DdlDmlResponseHandlers.java
+++ b/ksqldb-api-client/src/main/java/io/confluent/ksql/api/client/impl/DdlDmlResponseHandlers.java
@@ -38,7 +38,7 @@ final class DdlDmlResponseHandlers {
       final CompletableFuture<ExecuteStatementResult> cf
   ) {
     if (isIfNotExistsWarning(ksqlEntity)) {
-      cf.complete(new ExecuteStatementResultImpl(Optional.empty()));
+      cf.complete(new ExecuteStatementResultImpl(Optional.of(ksqlEntity.getString("message"))));
       return;
     }
     if (!isCommandStatusEntity(ksqlEntity)) {
@@ -75,9 +75,8 @@ final class DdlDmlResponseHandlers {
   private static boolean isIfNotExistsWarning(final JsonObject ksqlEntity) {
     return ksqlEntity.getString("statementText") != null
         && ksqlEntity.getString("statementText").contains("IF NOT EXISTS")
-        && ksqlEntity.getString("message") != null
-        && ksqlEntity.getString("message").contains("with the same name already exists.")
-        && ksqlEntity.getString("warnings") != null;
+        && ksqlEntity.getString("@type") != null
+        && ksqlEntity.getString("@type").equals("warning_entity");
   }
 
   // CHECKSTYLE_RULES.OFF: CyclomaticComplexity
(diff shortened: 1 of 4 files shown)


wcarlson5:
@nae701  Can this get picked to 0.23?

ksqldb-api-client/src/main/java/io/confluent/ksql/api/client/impl/DdlDmlResponseHandlers.java
```diff
@@ -68,6 +72,13 @@ private static boolean isCommandStatusEntity(final JsonObject ksqlEntity) {
         && ksqlEntity.getJsonObject("commandStatus") != null;
   }
 
+  private static boolean isIfNotExistsWarning(final JsonObject ksqlEntity) {
+    return ksqlEntity.getString("statementText") != null
+        && ksqlEntity.getString("statementText").contains("IF NOT EXISTS")
+        && ksqlEntity.getString("message") != null
+        && ksqlEntity.getString("message").contains("A stream with the same name already exists.");
```
jzaralim, line 79: This could also apply to tables and connectors. If there's some way to determine if the ksqlEntity is a warning rather than an error, then that would be the most inclusive way to catch everything.

wcarlson5: good call

jzaralim: There's also `DROP STREAM|TABLE|CONNECTOR IF EXISTS` which also returns a warning. I think just checking for the warning and making sure it's not an error should be enough.

wcarlson5: Are you saying don't check the msg contents?

jzaralim: Yeah. As long as we check after trying every other entity type, it should be accurate. It also definitely won't overlap with `ErrorEntity` because those have a field called `errorMessage` instead of just `message`.

ksqldb-api-client/src/main/java/io/confluent/ksql/api/client/impl/DdlDmlResponseHandlers.java
```diff
@@ -68,6 +72,14 @@ private static boolean isCommandStatusEntity(final JsonObject ksqlEntity) {
         && ksqlEntity.getJsonObject("commandStatus") != null;
   }
 
+  private static boolean isIfNotExistsWarning(final JsonObject ksqlEntity) {
+    return ksqlEntity.getString("statementText") != null
+        && ksqlEntity.getString("statementText").contains("IF NOT EXISTS")
+        && ksqlEntity.getString("message") != null
+        && ksqlEntity.getString("message").contains("with the same name already exists.")
+        && ksqlEntity.getString("warnings") != null;
```
jzaralim, line 80:
```
{"@type":"warning_entity","statementText":"CREATE STREAM IF NOT EXISTS high_value_stock_trades\n    WITH (KAFKA_TOPIC='high_value_stock_trades') AS\nSELECT U.ID USERID,\n    U.REGIONID REGION,\n    T.SIDE SIDE,\n    T.QUANTITY QUANTITY,\n    T.SYMBOL SYMBOL,\n    T.PRICE PRICE,\n    T.ACCOUNT ACCOUNT\nFROM STOCKTRADES T\nINNER JOIN USERS U ON (T.USERID = U.ID)\nWHERE ((T.PRICE > 500) AND (T.QUANTITY > 2500))\nEMIT CHANGES;","message":"Cannot add stream `HIGH_VALUE_STOCK_TRADES`: A stream with the same name already exists.","warnings":[]}
```

Checking for a warnings field does not actually check if the entity is a `WarningEntity`

ksqldb-api-client/src/main/java/io/confluent/ksql/api/client/impl/DdlDmlResponseHandlers.java
```diff
@@ -37,6 +37,10 @@ static void handleExecuteStatementResponse(
       final JsonObject ksqlEntity,
       final CompletableFuture<ExecuteStatementResult> cf
   ) {
+    if (isIfNotExistsWarning(ksqlEntity)) {
+      cf.complete(new ExecuteStatementResultImpl(Optional.empty()));
```
jzaralim, line 41: Let's return the warning message instead of just empty.

ksqldb-api-client/src/test/java/io/confluent/ksql/api/client/ClientTest.java
```diff
@@ -789,6 +791,23 @@ public void shouldHandleErrorResponseFromExecuteStatement() {
     assertThat(e.getCause().getMessage(), containsString("something bad"));
   }
 
+  @Test
+  public void shouldHandleIfNotExistsWarningResponseFromExecuteStatement() throws Exception {
```
jzaralim, line 795: Can you also add a test to `ClientIntegrationTest`?

ksqldb-api-client/src/main/java/io/confluent/ksql/api/client/impl/DdlDmlResponseHandlers.java
wcarlson5: ```{"statementText":"CREATE STREAM IF NOT EXISTS ","message":"Cannot add stream `HIGH_VALUE_STOCK_TRADES`: A stream with the same name already exists.","warnings":[]}

ksqldb-api-client/src/main/java/io/confluent/ksql/api/client/impl/DdlDmlResponseHandlers.java
wcarlson5: sounds good

ksqldb-api-client/src/main/java/io/confluent/ksql/api/client/impl/DdlDmlResponseHandlers.java
wcarlson5: sadly we don't check the Entity types and we just check to see if it has a cmd seq. If it doesn't we consider it an error. Since the entire point is avoid putting these on the cmd topic is doesn't really work

ksqldb-api-client/src/test/java/io/confluent/ksql/api/client/ClientTest.java
wcarlson5: yes

ksqldb-api-client/src/main/java/io/confluent/ksql/api/client/impl/DdlDmlResponseHandlers.java
jzaralim: Hm yeah that's weird too. One way around this would be to add a warning field to `ExecuteStatementResult`. Otherwise, leaving the QueryID blank is what we do for `DROP` statements, so perhaps it's not the worst thing to leave it blank.

ksqldb-api-client/src/main/java/io/confluent/ksql/api/client/impl/DdlDmlResponseHandlers.java
jzaralim: ohh okay. this solution works to just check for the `IF NOT EXISTS` case,  but`WarningEntity` is also used for `DROP CONNECTOR IF EXISTS` and `UNDEFINE` (neither of those should be accepted by the `executeStatement` method, but they get executed anyway.)

ksqldb-api-client/src/main/java/io/confluent/ksql/api/client/impl/DdlDmlResponseHandlers.java
wcarlson5: I think we should leave it blank. For the UI and a CLI a warning will print still. This is only for the java client and that might as well just check the logs.

ksqldb-api-client/src/main/java/io/confluent/ksql/api/client/impl/DdlDmlResponseHandlers.java
wcarlson5: Maybe we can do that in a follow up. I think we should keep this targeted to avoid secondary effects for this release

jzaralim (approved):

PR #8326, opened 2021-11-05
test: add BytesLiteral as immutable type everywhere

jzaralim:
### Description 
Add BytesLiteral to all the immutability tests in packages that depend on `ksqldb-execution` so that `mvn test` and intellij runs would pass.

### Reviewer checklist


wcarlson5 (approved):
I had to do this one place a while back. LGTM

Issue #8065, opened 2021-08-27
Having EXPLODE in CASE block leads to unexpected behavior


kaijunxu:
**Describe the bug**
Extending from #5782, when `EXPLODE` is used in any `WHEN` clause within a `CASE` block, the output of *every* clause will output as many rows as the number of elements in the largest array being `EXPLODE`d.

**To Reproduce**
ksqlDB 0.20.0.

Sample source data
```
+------------------------+-----------------------------------------------------------------------------------------------------------------
|ROWKEY                  |STATUS                                                                                                           
+------------------------+-----------------------------------------------------------------------------------------------------------------
|1TESTPLATE              |{TYPE=entry, CURR_CARPARK_NO=TEST1, CURR_TIMESTAMP=1630037554207, LAST_CARPARK_NO=TEST1, LAST_TIMESTAMP=1630037548761, TAILGATING=true}                                                                                           |
|1TESTPLATE              |{TYPE=multi, CURR_CARPARK_NO=TEST2, CURR_TIMESTAMP=1630037576185, LAST_CARPARK_NO=TEST1, LAST_TIMESTAMP=1630037571713, TAILGATING=true}                                                                                           |
|1TESTPLATE              |{TYPE=exit, CURR_CARPARK_NO=TEST3, CURR_TIMESTAMP=1630037599613, LAST_CARPARK_NO=TEST1, LAST_TIMESTAMP=1630037595696, TAILGATING=true}                                                                                            |
```

Statements run (1)
```sql
SELECT
  ROWKEY AS NUM_PLATE,
  CASE
    WHEN STATUS->TYPE = 'entry' THEN STATUS->LAST_CARPARK_NO
    WHEN STATUS->TYPE = 'exit' THEN STATUS->CURR_CARPARK_NO
    WHEN STATUS->TYPE = 'multi' THEN EXPLODE(ARRAY[STATUS->LAST_CARPARK_NO, STATUS->CURR_CARPARK_NO])
  END AS CAR_PARK_NO,
  STATUS->TYPE AS TYPE,
  STATUS->LAST_TIMESTAMP AS ALERT_START,
  STATUS->CURR_TIMESTAMP AS ALERT_END
FROM TAILGATE_STREAM_RAW
PARTITION BY NULL
EMIT CHANGES;
```

Statements run (2)
```sql
SELECT
  ROWKEY AS NUM_PLATE,
  CASE
[...]

PR #8327, opened 2021-11-05
fix: return error message if table functions are used inside a CASE

spena:
### Description 
Fixes https://github.com/confluentinc/ksql/issues/8065#event-5563432289

The last comment on https://github.com/confluentinc/ksql/issues/8065#event-5563432289 suggested that table functions should not be allowed in CASE functions. 

Today, we support one table function only (`EXPLODE`) which returns one column and several rows. But if in the future we support other tables functions (created with SQL UDTF for example) that return different columns, then mixing UDTF with CASE will complicate the resulted schema.

A quick example why not to support this case. The CASE will attempt to return 1 column for IDs < then 10, but 8 columns for > 10. This is not possible to do.
```
SELECT CASE
    WHEN s1.id >= 10 THEN
        tableFunctionReturns1Column()
    ELSE
       tableFunctionReturns2Columns()
   END
FROM ...
```

The following error is displayed when using this case:
```
ksql> select case when false then explode(array[1.2]) end from arraytest2 emit changes limit 1;
Table functions cannot be used in CASE: (CASE WHEN false THEN EXPLODE(ARRAY[1.2]) END)
```

### Testing done 

### Reviewer checklist

### DIFF
--- a/ksqldb-engine/src/main/java/io/confluent/ksql/planner/plan/FlatMapNode.java
+++ b/ksqldb-engine/src/main/java/io/confluent/ksql/planner/plan/FlatMapNode.java
@@ -98,8 +98,9 @@ public class FlatMapNode extends SingleSourcePlanNode {
     final ImmutableMap.Builder<Integer, Expression> buildContext = ImmutableMap
         .builder();
 
-    for (int idx = 0; idx < analysis.getSelectItems().size(); idx++) {
-      final SelectItem selectItem = analysis.getSelectItems().get(idx);
+    final List<SelectItem> selectItems = analysis.getSelectItems();
+    for (int idx = 0; idx < selectItems.size(); idx++) {
+      final SelectItem selectItem = selectItems.get(idx);
       if (!(selectItem instanceof SingleColumn)) {
         continue;
       }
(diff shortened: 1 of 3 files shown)


ksqldb-engine/src/main/java/io/confluent/ksql/planner/plan/FlatMapNode.java
```diff
@@ -98,8 +98,9 @@ public Expression resolveSelect(final int idx, final Expression expression) {
     final ImmutableMap.Builder<Integer, Expression> buildContext = ImmutableMap
         .builder();
 
-    for (int idx = 0; idx < analysis.getSelectItems().size(); idx++) {
-      final SelectItem selectItem = analysis.getSelectItems().get(idx);
+    final List<SelectItem> selectItems = analysis.getSelectItems();
+    for (int idx = 0; idx < selectItems.size(); idx++) {
+      final SelectItem selectItem = selectItems.get(idx);
```
spena, line 103: Minor refactor: the `getSelectItems` walks through each item and creates another SelectItem object everytime is called. Here I just calling it once.

colinhicks (approved):
Approved from the product behavior perspective. Thanks for the detailed reasoning here and in the other issue comment.

> Today, we support one table function only (EXPLODE)

Nit: we do also support `cube_explode`.

jzaralim (approved):

PR #8330, opened 2021-11-06
fix: make parse_date able to parse partial dates

### LINKED ISSUE #8328: `parse_date` will not parse a year.
**Describe the bug**
`parse_date('2021', 'yyyy')` returns `null`. I would expect it to return 01/01/2021.

**To Reproduce**
```sql
create stream inputs ( date_string VARCHAR) with (kafka_topic='inputs', value_format = 'avro', partitions=3);
insert into inputs (date_string) values ('2004');
insert into inputs (date_string) values ('2012');
```

```sql
set 'auto.offset.reset' = 'earliest';
select parse_date(date_string, 'yyyy')
from inputs
emit changes;
```

```
+-----------+
|KSQL_COL_0 |
+-----------+
|null       |
|null       |
```

jzaralim:
### Description 
Fixes #8328

Reason it didn't work is because `LocalDate` doesn't do partial parsing. `LocalTime` does partial parsing for `hh:mm`, but not just `hh`. I think that is fine though.

`StringToDate` also has this problem, but I did not fix that because it is deprecated.

### Testing done 
Unit + QTT

### Reviewer checklist

### DIFF
--- a/ksqldb-engine/src/main/java/io/confluent/ksql/function/udf/datetime/ParseDate.java
+++ b/ksqldb-engine/src/main/java/io/confluent/ksql/function/udf/datetime/ParseDate.java
@@ -26,6 +26,8 @@ import io.confluent.ksql.function.udf.UdfParameter;
 import io.confluent.ksql.util.KsqlConstants;
 import java.sql.Date;
 import java.time.LocalDate;
+import java.time.Year;
+import java.time.YearMonth;
 import java.time.format.DateTimeFormatter;
 import java.time.format.DateTimeFormatterBuilder;
 import java.time.temporal.ChronoField;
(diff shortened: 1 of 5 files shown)


ksqldb-engine/src/main/java/io/confluent/ksql/function/udf/datetime/ParseDate.java
```diff
@@ -73,8 +75,18 @@ public Date parseDate(
...
+            TimeUnit.DAYS.toMillis(LocalDate.from(ta).toEpochDay()));
+      } else if (ta.isSupported(ChronoField.YEAR) && ta.isSupported(ChronoField.MONTH_OF_YEAR)) {
+        return new Date(
+            TimeUnit.DAYS.toMillis(YearMonth.from(ta).atDay(1).toEpochDay()));
+      } else if (ta.isSupported(ChronoField.YEAR) && !ta.isSupported(ChronoField.DAY_OF_MONTH)) {
+        return new Date(
+            TimeUnit.DAYS.toMillis(Year.from(ta).atDay(1).toEpochDay()));
+      } else {
+        throw new KsqlFunctionException("Cannot create DATE out of the fields provided.");
+      }
```
spena, line 89: So we are limited to use specific string formats. Seems not very flexible. Is there another library that allows us to use any string format we want?

jzaralim: Swapped it to `SimpleDateFormat`, it does what we want. The difference is that it does even more partial parsing than we promised - it parses `PARSE_DATE("2021-04", "yyyy-dd")` to `2021-01-04`.

jzaralim: Actually `SimpleDateFormat` isn't thread safe. `FastDateFormat` from `apache-commons`'s `lang3` library works though.

spena (approved):
Awesome, thanks for the change.

PR #8332, opened 2021-11-06
Kci 1191/extend list properties

tolgadur:
### Description 
We want to specify whether a property is internal and which level (server or query) it is for the `list properties` command:

Screenshot 2021-11-06 at 15 59 55
 

### Testing done 
Unit testing 

### Reviewer checklist


wcarlson5 (approved):
LGTM. Only question is do we want to have properties default to internal or not? Right now it is not which is fine

PR #8333, opened 2021-11-07
feat: pull query LIMIT clause

cprasad1:
### Description 
Introduces `LIMIT` clause for pull queries by implementing [KLIP-58](https://github.com/confluentinc/ksql/pull/8298/)

Users of pull queries don't have any way of restricting the number of rows returned by a pull query. The LIMIT clause will enable users to save on computation, time and bandwidth costs by giving them an easy way to restrict the number of rows returned.

Syntax evolution:
```SQL
SELECT select_expr [, ...]                 SELECT select_expr [, ...] 
  FROM table/stream                          FROM table/stream        
  [ WHERE where_condition ]    ────────►     [ WHERE where_condition ]
  [ AND window_bounds ];                     [ AND window_bounds ]    
                                           + [ LIMIT num_records ];  
```

Usage:
For pull queries over tables:
```SQL
SELECT distanceinmiles, riders 
FROM ridersNearMountainView 
WHERE distanceinmiles < 115 
LIMIT 3;

+-------------------------------+-------------------------------+
|DISTANCEINMILES                |RIDERS                         |
+-------------------------------+-------------------------------+
|0.0                            |[4ab5cbad, 8b6eae59, 4a7c7b41] |
|10.0                           |[18f4ea86]                     |
|50.0                           |[c2309eec, 4ddad000]           |
Limit Reached
Query terminated
```

For pull queries over streams:
```SQL
SELECT profileid, latitude 
FROM riderlocations 
WHERE latitude > 38 
LIMIT 5;

+-------------------------------+-------------------------------+
|PROFILEID                      |LATITUDE                       |
+-------------------------------+-------------------------------+
|c2309eeca                      |38.7877                        |
|18f4ea86a                      |38.3903                        |
|4ab5cbada                      |38.3952                        |
|8b6eae59a                      |38.3944                        |
|4a7c7b41a                      |38.4049                        |
Limit Reached
Query terminated
```

### Testing done 

### Reviewer checklist


ksqldb-engine/src/main/java/io/confluent/ksql/planner/LogicalPlanner.java
```diff
@@ -219,6 +225,10 @@ public OutputNode buildQueryLogicalPlan(
       }
     }
 
+    if (!isScalablePush && analysis.getLimitClause().isPresent()) {
```
vvcephei, line 228: What's the significance of this exclusion?

ksqldb-engine/src/main/java/io/confluent/ksql/planner/plan/QueryLimitNode.java
```diff
@@ -0,0 +1,46 @@
...
+
+public class QueryLimitNode extends SingleSourcePlanNode {
+  private final int limit;
+
+  public QueryLimitNode(final PlanNodeId id,
+                        final PlanNode source,
+                        final OptionalInt limit
+  ) {
+    super(id, source.getNodeOutputType(), source.getSourceName(), source);
+    this.limit = limit.getAsInt();
```
vvcephei, line 30: What's up with the parameter being Optional, but then we require it to be present here?

ksqldb-engine/src/test/java/io/confluent/ksql/planner/plan/QueryProjectNodeTest.java
```diff
@@ -44,10 +37,16 @@
 import io.confluent.ksql.util.KsqlConfig;
 import java.util.List;
 import java.util.Optional;
+import static org.hamcrest.MatcherAssert.assertThat;
```
vvcephei, line 40: This change probably doesn't need to be checked in.

ksqldb-functional-tests/src/test/resources/rest-query-validation-tests/pull-queries-against-streams.json
```diff
@@ -31,18 +31,6 @@
...
-      "statements": [
-        "CREATE STREAM S1 (MYKEY INT KEY, MYVALUE INT) WITH (kafka_topic='test_topic', value_format='JSON');",
-        "SELECT * FROM S1 LIMIT 1;"
-      ],
-      "expectedError": {
-        "type": "io.confluent.ksql.rest.entity.KsqlStatementErrorMessage",
-        "message": "Pull queries don't support LIMIT clauses. See https://cnfl.io/queries for more info.\nAdd EMIT CHANGES if you intended to issue a push query.",
-        "status": 400
-      }
-    },
```
vvcephei, line 45: Would it be better to convert this to a test that the query works on streams?

vvcephei: nvm, I see you moved it to the new file.

ksqldb-rest-app/src/test/java/io/confluent/ksql/rest/integration/PullQueryRoutingFunctionalTest.java
```diff
@@ -45,10 +29,18 @@
 import io.confluent.ksql.rest.entity.KsqlErrorMessage;
 import io.confluent.ksql.rest.entity.KsqlHostInfoEntity;
 import io.confluent.ksql.rest.entity.StreamedRow;
+import static io.confluent.ksql.rest.entity.StreamedRowMatchers.matchersRowsAnyOrder;
```
vvcephei, line 32: side note: it seems like your import ordering configuration needs to be updated so you don't just scramble the imports when you edit files. Not a big deal, but it makes the diff a little harder to read.

vvcephei (commented):
Very nice, @cprasad1 !

Just a few small comments.

ksqldb-engine/src/main/java/io/confluent/ksql/physical/pull/operators/LimitOperator.java
```diff
@@ -0,0 +1,103 @@
...
+  public LimitOperator(final QueryLimitNode logicalNode, final ProcessingLogger logger) {
+    this(logicalNode, logger, logicalNode.getLimit());
+  }
+
+  @VisibleForTesting
+  LimitOperator(final QueryLimitNode logicalNode,
+                final ProcessingLogger logger,
+                final int limit
+  ) {
+    this.logger = Objects.requireNonNull(logger, "logger");
```
AlanConfluent, line 44: This isn't being used, and really is for persistent queries so can be removed.

ksqldb-engine/src/main/java/io/confluent/ksql/physical/pull/PullPhysicalPlanBuilder.java
```diff
@@ -191,6 +197,16 @@ private SelectOperator translateFilterNode(final QueryFilterNode logicalNode) {
     return new SelectOperator(logicalNode, logger);
   }
 
+  private LimitOperator translateLimitNode(final QueryLimitNode logicalNode) {
+    final ProcessingLogger logger = processingLogContext
```
AlanConfluent, line 201: You can get rid of this unless you're going to use it.  I don't think we actually use this for pull queries, except where we share APIs with persistent queries and must, namely the SelectValueMapper which is shared.  The reason is that we can err immediately and surface that error whereas persistent queries cannot.

ksqldb-engine/src/main/java/io/confluent/ksql/planner/LogicalPlanner.java
AlanConfluent: It's true that we could also do this for scalable push, though then we'd probably want to account for it in the physical plan builder as well.  I think it would just ignore it at the moment if you didn't add support there too, which might be fine if we just want to avoid exclusions like this.

If you were going to add support for scalable push, it would probably make sense to do that in a separate PR.

ksqldb-rest-app/src/test/java/io/confluent/ksql/rest/integration/PullQueryRoutingFunctionalTest.java
```diff
@@ -292,6 +292,63 @@ public void cleanUp() {
     APP_SHUTOFFS_2.reset();
   }
 
+  @Test
+  public void shouldReturnLimitRowsMultiHostSetup() {
```
AlanConfluent, line 296: You should probably do this in its own file since these tests are very specific to routing.

Also, this has only one partition in these tests, which honestly is not a very interesting mult-node case.  Would be great to have a tests with a couple nodes and at least a couple partitions.

ksqldb-engine/src/main/java/io/confluent/ksql/planner/plan/QueryLimitNode.java
cprasad1: You are right. it should always be there

ksqldb-engine/src/main/java/io/confluent/ksql/planner/LogicalPlanner.java
vvcephei: Ok, so reading between the lines a little, if we're planning a scalable push query, we simply won't include the limit node in the logical or physical plan. Or are you saying that limit clauses are actually not allowed for scalable push queries at all?

In any case, I'm fine with splitting stuff off into follow-on PRs, but if the clause is supported in SPQs, then I'm concerned about omitting them from the logical plan.

vvcephei (approved):
Looks good to me @cprasad1 !

I just had one small concern, but it's either not valid (because scalable push queries don't support LIMIT) or we can address it in another PR.

ksqldb-engine/src/main/java/io/confluent/ksql/planner/LogicalPlanner.java
AlanConfluent: They are allowed (and encouraged!), but it would take more code to fully support this logical node.  That's all I meant.

ksqldb-rest-app/src/test/java/io/confluent/ksql/rest/integration/PullQueryLimitFunctionalTest.java
```diff
@@ -0,0 +1,491 @@
...
+
+        // issue pull query with limit after partitioning active
+        final List<StreamedRow> rows_3 = makePullQueryRequest(clusterFormation.router.getApp(),
+                sqlLimit, null, USER_CREDS);
+
+        // check that we only got limit number of rows
+        assertThat(rows_3, hasSize(limitSubsetSize));
+    }
+
+    private ClusterFormation findClusterFormation(
```
AlanConfluent, line 380: I don't really think that most of this is necessary for your test.  I guess I understand a simplified version of this where you ensure that you have an active host for all of your partitions.

ksqldb-rest-app/src/test/java/io/confluent/ksql/rest/integration/PullQueryLimitFunctionalTest.java
```diff
@@ -0,0 +1,491 @@
...
+    @BeforeClass
+    public static void setUpClass() {
+        FaultyKafkaConsumer.FaultyKafkaConsumer0.setPauseOffset(APP_SHUTOFFS_0::getKafkaPauseOffset);
+        FaultyKafkaConsumer.FaultyKafkaConsumer1.setPauseOffset(APP_SHUTOFFS_1::getKafkaPauseOffset);
+        FaultyKafkaConsumer.FaultyKafkaConsumer2.setPauseOffset(APP_SHUTOFFS_2::getKafkaPauseOffset);
+    }
+
+    @Before
+    public void setUp() {
+        //Create topic with 4 partition to control who is active and standby
```
AlanConfluent, line 214: Heh, the cluster formation logic works because we have a single partition and it's well defined who's active and who's standby.  With 4 partitions, you're going to have each node as active for something.

ksqldb-rest-app/src/test/java/io/confluent/ksql/rest/integration/PullQueryLimitFunctionalTest.java
```diff
@@ -0,0 +1,491 @@
...
+        ClusterFormation clusterFormation = new ClusterFormation();
+        ClusterStatusResponse clusterStatusResponse
+                = HighAvailabilityTestUtil.sendClusterStatusRequest(testApp0.getApp(), USER_CREDS);
+        ActiveStandbyEntity entity0 = clusterStatusResponse.getClusterStatus().get(testApp0.getHost())
+                .getActiveStandbyPerQuery().get(queryId);
+        ActiveStandbyEntity entity1 = clusterStatusResponse.getClusterStatus().get(testApp1.getHost())
+                .getActiveStandbyPerQuery().get(queryId);
+
+        // find active
+        if (!entity0.getActiveStores().isEmpty() && !entity0.getActivePartitions().isEmpty()) {
```
AlanConfluent, line 391: getActivePartitions() is a set of topic partitions.  This only works with 1 partition.

ksqldb-rest-app/src/test/java/io/confluent/ksql/rest/integration/PullQueryLimitFunctionalTest.java
```diff
@@ -0,0 +1,491 @@
...
+
+    @Rule
+    public final Timeout timeout = Timeout.builder()
+            .withTimeout(4, TimeUnit.MINUTES)
+            .withLookingForStuckThread(true)
+            .build();
+
+    @BeforeClass
+    public static void setUpClass() {
+        FaultyKafkaConsumer.FaultyKafkaConsumer0.setPauseOffset(APP_SHUTOFFS_0::getKafkaPauseOffset);
```
AlanConfluent, line 207: You don't utilize any of these shutoffs in your test to simulate failures.

ksqldb-rest-app/src/test/java/io/confluent/ksql/rest/integration/PullQueryLimitFunctionalTest.java
```diff
@@ -0,0 +1,491 @@
...
+
+    private static final String PROPS_JAAS_REALM = "KsqlServer-Props";
+    private static final String KSQL_RESOURCE = "ksql-user";
+    private static final String USER_WITH_ACCESS = "harry";
+    private static final String USER_WITH_ACCESS_PWD = "changeme";
+    private static final Optional<BasicCredentials> USER_CREDS
+            = Optional.of(BasicCredentials.of(USER_WITH_ACCESS, USER_WITH_ACCESS_PWD));
+
+    @ClassRule
+    public static final TestBasicJaasConfig JAAS_CONFIG = TestBasicJaasConfig
```
AlanConfluent, line 112: It's fine to include auth in this test, though not entirely necessary since we have pull query tests with it elsewhere and you could simplify this.

ksqldb-rest-app/src/test/java/io/confluent/ksql/rest/integration/PullQueryLimitFunctionalTest.java
```diff
@@ -0,0 +1,491 @@
...
+
+        // issue pull query with limit
+        final List<StreamedRow> rows_1 = makePullQueryRequest(clusterFormation.router.getApp(),
+                sqlLimit, null, USER_CREDS);
+
+        // check that we only got limit number of rows
+        assertThat(rows_1, hasSize(limitSubsetSize));
+
+        // Partition off the active
+        clusterFormation.active.getShutoffs().shutOffAll();
```
AlanConfluent, line 291: This is kinda cool to verify, but this is really an HA test of a sort.  Not sure if we want to include that in a limit test.

AlanConfluent (approved):
This all looks good to me.  I would just ask that you simplify the functional test.  It does some stuff that's more related to HA routing rather than testing the limit functionality.  And the cluster formation logic doesn't really make sense in the multi partition scenario.

ksqldb-rest-app/src/test/java/io/confluent/ksql/rest/integration/PullQueryLimitFunctionalTest.java
cprasad1: removed

PR #8334, opened 2021-11-08
Kci 1081/implement endpoint for query scaling server configs

tolgadur:
### Description 
Due to the changes in the Query Scaling project queries will share a Kafka Streams runtime instead of each having its own. This means that the majority of configs will be set at the server-level instead of the query-level. For this we need a new api endpoints to set new configurations at the server-level:

```
ksql> alter system 'ksql.udfs.enabled'='test';
Could not write the statement 'alter system 'ksql.udfs.enabled'='test';' into the command topic.
Caused by: Invalid value null for configuration Failed to set ksql.udfs.enabled
	to test. Caused by: Not recognizable as ksql, streams, consumer, or producer
	property: ksql.udfs.enabled 

ksql> alter system 'test'='test';
Could not write the statement 'alter system 'test'='test';' into the command topic.
Caused by: Invalid value null for configuration Failed to set test to test.
	Caused by: Not recognizable as ksql, streams, consumer, or producer property:
	test 

ksql> alter system 'ksql.persistent.prefix'='test';

 Message                  

 System property altered. 

ksql> list properties;

 Property                                               | Scope | Default override | Effective Value                                                                                                                            

 ksql.access.validator.enable                           | KSQL  |                  | auto                                                                                                                                       
 ksql.authorization.cache.expiry.time.secs              | KSQL  |                  | 30                                                                                                                                         
 ksql.authorization.cache.max.entries                   | KSQL  |                  | 10000                                                                                                                                      
 ksql.cast.strings.preserve.nulls                       | KSQL  |                  | true                                                                                                                                       
 ksql.connect.url                                       | KSQL  |                  | http://localhost:8083                                                                                                                      
 ksql.connect.worker.config                             | KSQL  |                  |                                                                                                                                            
 ksql.create.or.replace.enabled                         | KSQL  |                  | true                                                                                                                                       
 ksql.error.classifier.regex                            | KSQL  |                  |                                                                                                                                            
 ksql.extension.dir                                     | KSQL  |                  | ext                                                                                                                                        
 ksql.hidden.topics                                     | KSQL  |                  | _confluent.*,__confluent.*,_schemas,__consumer_offsets,__transaction_state,connect-configs,connect-offsets,connect-status,connect-statuses 
 ksql.insert.into.values.enabled                        | KSQL  |                  | true                                                                                                                                       
 ksql.internal.topic.min.insync.replicas                | KSQL  |                  | 1                                                                                                                                          
 ksql.internal.topic.replicas                           | KSQL  |                  | 1                                                                                                                                          
 ksql.lambdas.enabled                                   | KSQL  |                  | true                                                                                                                                       
 ksql.metastore.backup.location                         | KSQL  |                  |                                                                                                                                            
 ksql.metrics.extension                                 | KSQL  |                  | NULL                                                                                                                                       
 ksql.metrics.tags.custom                               | KSQL  |                  |                                                                                                                                            
 ksql.nested.error.set.null                             | KSQL  |                  | true                                                                                                                                       
 ksql.output.topic.name.prefix                          | KSQL  |                  |                                                                                                                                            
 ksql.persistence.default.format.key                    | KSQL  |                  | KAFKA                                                                                                                                      
 ksql.persistence.default.format.value                  | KSQL  |                  | NULL                                                                                                                                       
 ksql.persistence.wrap.single.values                    | KSQL  |                  | NULL                                                                                                                                       
 ksql.persistent.prefix                                 | KSQL  | SERVER           | test                                                                                                                                       
 [...]
```

### Disclaimer
Due to many local copies of `KsqlConfig` there may be instances where somewhere in the codebase a copy of `KsqlConfig` does not have the new overriden properties. We plan to have a larger refactor for this in future.

### Testing done 
Unit testing 

### Reviewer checklist


ksqldb-engine/src/main/java/io/confluent/ksql/engine/KsqlEngine.java
```diff
@@ -324,6 +324,15 @@ public TransientQueryMetadata executeTransientQuery(
     }
   }
 
+  public KsqlConfig getKsqlConfig() {
```
rodesai, line 327: you should add this method to `KsqlExecutionContext`, and implement it there by getting the config from the passed in `EngineContext`. This way we can get the config out of the sandboxed engine. While you're at it, can you remove the `ksqlConfig` member from `KsqlEngine`? It should just use the config in `primaryContext`

ksqldb-common/src/main/java/io/confluent/ksql/util/KsqlConfig.java
```diff
@@ -1318,6 +1318,10 @@ private KsqlConfig(final ConfigGeneration generation,
     this.ksqlStreamConfigProps = ksqlStreamConfigProps;
   }
 
+  public boolean keyExistsInConfigMap(final String key) {
```
rodesai, line 1321: what's this for? it seems like it's never used outside your unit test.

ksqldb-common/src/main/java/io/confluent/ksql/util/KsqlConfig.java
```diff
@@ -1466,6 +1470,13 @@ private KsqlConfig(final ConfigGeneration generation,
     return Collections.unmodifiableMap(allPropsCleaned);
   }
 
+  public KsqlConfig cloneWithoutOverride() {
```
rodesai, line 1473: why do we need this method?

ksqldb-engine/src/main/java/io/confluent/ksql/engine/KsqlEngine.java
```diff
@@ -324,6 +324,15 @@ public TransientQueryMetadata executeTransientQuery(
...
+    return this.ksqlConfig.cloneWithoutOverride();
```
rodesai, line 328: why do you have to clone it? `KsqlConfig` is immutable.

ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/validation/RequestValidator.java
```diff
@@ -113,6 +113,7 @@ public int validate(
 
     final KsqlExecutionContext ctx = requireSandbox(snapshotSupplier.apply(serviceContext));
     final Injector injector = injectorFactory.apply(ctx, serviceContext);
+    final KsqlConfig ksqlConfig = ksqlEngine.getKsqlConfig();
```
rodesai, line 116: don't use a passed in engine here. instead, get the config out of the sandbox (`ctx`). You'll have to add an interface for getting the config (see my other comment).

ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/execution/RequestHandler.java
```diff
@@ -93,7 +89,8 @@ public KsqlEntityList execute(
               : Collections.emptyMap())
       );
       final ConfiguredStatement<?> configured = ConfiguredStatement.of(prepared,
-          SessionConfig.of(ksqlConfig, sessionProperties.getMutableScopedProperties())
+          SessionConfig.of(this.ksqlEngine.getKsqlConfig(),
```
rodesai, line 92: this is probably not the right place to generate a `ConfiguredStatement` anymore, because the configuration can change between this line and the point where the command is actually executed from `executeStatement`. We should instead leave this as a `PreparedStatement` and assemble the final `ConfiguredStatement` either after `commandQueueSync.waitFor` or in the executor implementation itself (the latter would require a bit more refactoring).

ksqldb-engine/src/main/java/io/confluent/ksql/engine/KsqlEngine.java
rodesai, line 327: this needs to be synchronized with `alterSystemProperty`

ksqldb-rest-model/src/main/java/io/confluent/ksql/rest/entity/PropertiesList.java
```diff
@@ -159,6 +254,112 @@
       FAIL_ON_DESERIALIZATION_ERROR_CONFIG
   );
 
+  public static final List<String> EditablePropertyList = ImmutableList.of(
```
rodesai, line 257: why are these all editable? only a handful of these should be editable right? can you add a link to the document that classifies the configs?

ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/validation/CustomValidators.java
```diff
@@ -78,7 +79,7 @@
         throw new KsqlRestException(Errors.queryEndpoint(statement.getStatementText()));
       }),
   PRINT_TOPIC(PrintTopic.class, PrintTopicValidator::validate),
-
+  ALTER_SYSTEM_PROPERTY(AlterSystemProperty.class, StatementValidator.NO_VALIDATION),
```
rodesai, line 82: this should not be NO_VALIDATION - you should just not set anything for AlterSystemProperty and the validator should then use the ValidatedCommandFactory to validate `ALTER SYNTAX`

ksqldb-engine/src/main/java/io/confluent/ksql/engine/KsqlEngine.java
tolgadur: My build was failing when I exposed `KsqlConfig` without cloning it as "internal implementation is exposed". Is there another way to do this?

ksqldb-common/src/main/java/io/confluent/ksql/util/KsqlConfig.java
tolgadur: See comment below.

ksqldb-rest-model/src/main/java/io/confluent/ksql/rest/entity/PropertiesList.java
tolgadur: I think @ableegoldman was saying that all configs in `StreamsConfig`, `ConsumerConfig`, `ProducerConfig`, `KsqlConfig` and `AdminClient` that aren't in the `internal` list above should be editable, which are all the configs i listed here. 

@ableegoldman do you have more insight here? Should any of these configs not be editable?

ksqldb-common/src/main/java/io/confluent/ksql/util/KsqlConfig.java
tolgadur: Good catch. I was using this for my validation before we agreed on the allow list.

ksqldb-engine/src/main/java/io/confluent/ksql/engine/KsqlEngine.java
tolgadur: With synchronization you mean using the `ksqlConfig` of the `EngineContext` when doing this: `this.ksqlConfig = this.ksqlConfig.cloneWithPropertyOverwrite(overrides);`, right?

ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/computation/ValidatedCommandFactory.java
```diff
@@ -105,9 +108,31 @@ private static Command createCommand(
       return createForTerminateQuery(statement, context);
     }
 
+    if (statement.getStatement() instanceof AlterSystemProperty) {
+      return createForAlterSystemQuery(statement, context);
+    }
+
     return createForPlannedQuery(statement, serviceContext, context);
```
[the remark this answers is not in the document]
tolgadur: Do you mean updating the `statement` for `createForPlannedQuery` or for `createForAlterSystemQuery`? Because I am not using the config in my validation?

ksqldb-rest-model/src/main/java/io/confluent/ksql/rest/entity/PropertiesList.java
ableegoldman: See my response on slack, in sum though we should remove all of the ksql (and admin) configs from the allowlist.

Side note: the  `SINK_WINDOW_CHANGE_LOG_ADDITIONAL_RETENTION_MS_PROPERTY,` seems to actually be useless, it's never, I don't understand why there's a separate ksql config for this since it just corresponds to a StreamsConfig, in fact it's never handled by ksql so given the extra `SINK` prefix in the name relative to the  actual Streams config, I believe it wouldn't even work anyways. I guess I'll just file a ticket to clean that up on the side

tolgadur: I tried to write a method to get all the config names from the classes directly instead of specifying them one by one (see line 362) but this list weirdly includes properties such as `sasl.mechanism` or `ssl.protocol` that are turned by `ConsumerConfig.configNames()`. I am assuming that we don't want them to be editable? Is there a way to exclude them perhaps or should we stick to the manual definition?

ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/validation/CustomValidators.java
tolgadur: When I remove this line I get:

```
Do not know how to validate statement of type: class io.confluent.ksql.parser.tree.AlterSystemProperty
Known types: [class io.confluent.ksql.parser.tree.ListConnectorPlugins, class io.confluent.ksql.parser.tree.ListTables, class io.confluent.ksql.parser.tree.ListTopics, class io.confluent.ksql.parser.tree.Query, class io.confluent.ksql.parser.tree.DropConnector, class io.confluent.ksql.parser.tree.ListQueries, class io.confluent.ksql.parser.tree.DescribeStreams, class io.confluent.ksql.parser.tree.UndefineVariable, class io.confluent.ksql.parser.tree.DefineVariable, class io.confluent.ksql.parser.tree.CreateConnector, class io.confluent.ksql.parser.tree.PrintTopic, class io.confluent.ksql.parser.tree.ShowColumns, class io.confluent.ksql.parser.tree.DescribeConnector, class io.confluent.ksql.parser.tree.ListConnectors, class io.confluent.ksql.parser.tree.ListFunctions, class io.confluent.ksql.parser.tree.ListProperties, class io.confluent.ksql.parser.tree.SetProperty, class io.confluent.ksql.parser.tree.Explain, class io.confluent.ksql.parser.tree.DescribeFunction, class io.confluent.ksql.parser.tree.UnsetProperty, class io.confluent.ksql.parser.tree.ListStreams, class io.confluent.ksql.parser.tree.ListVariables, class io.confluent.ksql.parser.tree.InsertValues, class io.confluent.ksql.parser.tree.ListTypes, class io.confluent.ksql.parser.tree.DescribeTables]
```
When I do keep this line it still uses the `ValidatedCommandFactory` to validate `ALTER SYNTAX`. Is the CustomValidators specification just extra maybe?

ksqldb-engine/src/main/java/io/confluent/ksql/engine/KsqlEngine.java
rodesai: I mean synchronzation from a concurrency perspective. The ksql config object (both here and in primaryContext) is not `volatile`, so you might read the value while it's being written and get back a corrupted value. To fix this you can either make it volatile, or make `alterSystemProperty` and `getKsqlConfig` synchronized.

ksqldb-engine/src/main/java/io/confluent/ksql/engine/EngineContext.java
```diff
@@ -159,6 +159,14 @@ QueryRegistry getQueryRegistry() {
     return queryRegistry;
   }
 
+  KsqlConfig getKsqlConfig() {
```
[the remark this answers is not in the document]
tolgadur: does the synchronized keyword work or do we need something else here?
```
    synchronized (this) {
      this.ksqlConfig = this.ksqlConfig.cloneWithPropertyOverwrite(overrides);
    }
```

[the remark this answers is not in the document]
rodesai: this method also needs to be synchronized:
```
synchronized KsqlConfig getKsqlConfig() {...
```

ksqldb-engine/src/main/java/io/confluent/ksql/engine/EngineContext.java
```diff
@@ -159,6 +159,16 @@ QueryRegistry getQueryRegistry() {
...
+    return ksqlConfig;
+  }
+
+  void alterSystemProperty(final Map<String, String> overrides) {
+    synchronized (this) {
```
rodesai, line 167: just use the `synchronized` keyword in the method signature

ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/computation/CommandIdAssigner.java
```diff
@@ -142,6 +145,14 @@ private static CommandId getAlterSourceCommandId(final AlterSource alterSource)
     );
   }
 
+  private static CommandId getAlterSystemCommandId(final AlterSystemProperty alterSystemProperty) {
+    return new CommandId(
+        Type.CLUSTER,
+        alterSystemProperty.toString(),
```
rodesai, line 151: why are you using the string form of alter system property here?

rodesai: I don't think it will parse correctly when the command id is serialized/deserialized because the command id always serializes to something like "<type>/<name>/<action>". In this case type is CLUSTER and action is ALTER. But the name is just the stringified version of AlterSystemProperty. We can't really come up with a name for this operation. Can you just use a uuid here to ensure it's unique?

ksqldb-rest-model/src/main/java/io/confluent/ksql/rest/entity/PropertiesList.java
```diff
@@ -161,14 +127,98 @@
...
+      GROUP_ID_CONFIG,
```
rodesai, line 131: this should not be editable (cc @ableegoldman)

ksqldb-rest-model/src/main/java/io/confluent/ksql/rest/entity/PropertiesList.java
```diff
@@ -161,14 +127,98 @@
...
+      GROUP_INSTANCE_ID_CONFIG,
```
rodesai, line 132: this should not be editable (cc @ableegoldman)

ksqldb-rest-model/src/main/java/io/confluent/ksql/rest/entity/PropertiesList.java
```diff
@@ -161,14 +127,98 @@
...
+      MAX_POLL_RECORDS_CONFIG,
+      MAX_POLL_INTERVAL_MS_CONFIG,
+      SESSION_TIMEOUT_MS_CONFIG,
+      HEARTBEAT_INTERVAL_MS_CONFIG,
+      BOOTSTRAP_SERVERS_CONFIG,
```
rodesai, line 137: this should not be editable (cc @ableegoldman)

ksqldb-rest-model/src/main/java/io/confluent/ksql/rest/entity/PropertiesList.java
```diff
@@ -161,14 +127,98 @@
...
+      CLIENT_DNS_LOOKUP_CONFIG,
```
rodesai, line 138: this should not be editable (cc @ableegoldman)

ksqldb-rest-model/src/main/java/io/confluent/ksql/rest/entity/PropertiesList.java
```diff
@@ -161,14 +127,98 @@
...
+      PARTITION_ASSIGNMENT_STRATEGY_CONFIG,
+      AUTO_OFFSET_RESET_CONFIG,
+      FETCH_MIN_BYTES_CONFIG,
+      FETCH_MAX_BYTES_CONFIG,
+      FETCH_MAX_WAIT_MS_CONFIG,
+      METADATA_MAX_AGE_CONFIG,
+      MAX_PARTITION_FETCH_BYTES_CONFIG,
+      SEND_BUFFER_CONFIG,
+      RECEIVE_BUFFER_CONFIG,
+      CLIENT_ID_CONFIG,
```
rodesai, line 150: this should not be editable (cc @ableegoldman)

rodesai (approved):
LGTM

PR #8306, opened 2021-10-29
8296: Debug issue with ConsumerCollector on Linux ARM64

martin-g:
Fixes #8296 

Re-initialize `ConsumerCollector#totalBytesSum` after re-initializing `MetricCollectors#metrics`

### Description 

`ConsumerCollectorTest` fails consistently on Linux ARM64.
The reason is that `MetricCollectorsTest` resets `MetricCollectors` after each of its tests and that `ConsumerCollector#totalBytesSum` is initialized in `static {...}` class initializer.
So, if `ConsumerCollector` class is loaded before any of the `MetricCollectorsTest`'s test finishes then `ConsumerCollector#totalBytesSum` will be lost because it uses an obsolete instance of `MetricCollectors#metrics`

### Testing done 

`MetricCollectorsTest` and `ConsumerCollectorTest` now pass also on Linux ARM64.

### Reviewer checklist
- [ ] Ensure docs are updated if necessary. (eg. if a user visible feature is being added or changed).
- [ ] Ensure relevant issues are linked (description should include text like "Fixes #<issue number>")

PR #8335, opened 2021-11-08
Compare Double test expectations with error epsilon

martin-g:
### Description 

io.confluent.ksql.function.udf.math.ExpTest fails on Linux ARM64 with:

```
[INFO] Running io.confluent.ksql.function.udf.math.ExpTest
[ERROR] Tests run: 4, Failures: 1, Errors: 0, Skipped: 0, Time elapsed: 0.054 s <<< FAILURE! - in io.confluent.ksql.function.udf.math.ExpTest
[ERROR] io.confluent.ksql.function.udf.math.ExpTest.shouldHandlePositive  Time elapsed: 0.013 s  <<< FAILURE!
java.lang.AssertionError: 

Expected: is <2.718281828459045>
     but: was <2.7182818284590455>
	at io.confluent.ksql.function.udf.math.ExpTest.shouldHandlePositive(ExpTest.java:56)

[INFO] 
[INFO] Results:
[INFO] 
[ERROR] Failures: 
[ERROR]   ExpTest.shouldHandlePositive:56 
Expected: is <2.718281828459045>
     but: was <2.7182818284590455>
[INFO] 
[ERROR] Tests run: 4, Failures: 1, Errors: 0, Skipped: 0
```

### Testing done 

Updated the tests to use approximation. Testing `double`s for equality is not stable.

### Reviewer checklist

### DIFF
--- a/ksqldb-engine/src/test/java/io/confluent/ksql/function/udf/math/ExpTest.java
+++ b/ksqldb-engine/src/test/java/io/confluent/ksql/function/udf/math/ExpTest.java
@@ -16,6 +16,7 @@ package io.confluent.ksql.function.udf.math;
 
 import static org.hamcrest.MatcherAssert.assertThat;
 import static org.hamcrest.Matchers.is;
+import static org.hamcrest.Matchers.closeTo;
 import static org.hamcrest.Matchers.nullValue;
 
 import org.junit.Before;
@@ -39,22 +40,22 @@ public class ExpTest {
 
   @Test
   public void shouldHandleNegative() {
-    assertThat(udf.exp(-1), is(0.36787944117144233));
-    assertThat(udf.exp(-1L), is(0.36787944117144233));
-    assertThat(udf.exp(-1.0), is(0.36787944117144233));
+    assertThat(udf.exp(-1), is(closeTo(0.36787944117144233, 0.00000000000000005)));
+    assertThat(udf.exp(-1L), is(closeTo(0.36787944117144233, 0.00000000000000005)));
+    assertThat(udf.exp(-1.0), is(closeTo(0.36787944117144233, 0.00000000000000005)));
   }
 
   @Test
   public void shouldHandleZero() {
-    assertThat(udf.exp(0), is(1.0));
-    assertThat(udf.exp(0L), is(1.0));
-    assertThat(udf.exp(0.0), is(1.0));
+    assertThat(udf.exp(0), is(closeTo(1.0, 0.1)));
+    assertThat(udf.exp(0L), is(closeTo(1.0, 0.1)));
+    assertThat(udf.exp(0.0), is(closeTo(1.0, 0.1)));
   }
 
   @Test
   public void shouldHandlePositive() {
-    assertThat(udf.exp(1), is(2.718281828459045));
-    assertThat(udf.exp(1L), is(2.718281828459045));
-    assertThat(udf.exp(1.0), is(2.718281828459045));
+    assertThat(udf.exp(1), is(closeTo(2.718281828459045, 0.00000000000005)));
+    assertThat(udf.exp(1L), is(closeTo(2.718281828459045, 0.00000000000005)));
+    assertThat(udf.exp(1.0), is(closeTo(2.718281828459045, 0.00000000000005)));
   }
 }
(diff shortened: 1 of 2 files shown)


martin-g:
Related-to: #8306 - the error occurs only on Linux ARM64.

martin-g:
Any feedback ?

ksqldb-engine/src/test/java/io/confluent/ksql/function/udf/math/ExpTest.java
```diff
@@ -39,22 +40,22 @@ public void shouldHandleNull() {
...
+    assertThat(udf.exp(0L), is(closeTo(1.0, 0.1)));
+    assertThat(udf.exp(0.0), is(closeTo(1.0, 0.1)));
   }
 
   @Test
   public void shouldHandlePositive() {
-    assertThat(udf.exp(1), is(2.718281828459045));
-    assertThat(udf.exp(1L), is(2.718281828459045));
-    assertThat(udf.exp(1.0), is(2.718281828459045));
+    assertThat(udf.exp(1), is(closeTo(2.718281828459045, 0.00000000000005)));
```
Gerrrr, line 57: Can you please use scientific notation for errors tolerance? For example:

```java
assertThat(udf.exp(1), is(closeTo(2.718281828459045, 5E-17)))
```

 is more readable than 

```java
assertThat(udf.exp(1), is(closeTo(2.718281828459045, 0.00000000000005)))
```

ksqldb-engine/src/test/java/io/confluent/ksql/function/udf/math/ExpTest.java
```diff
@@ -39,22 +40,22 @@ public void shouldHandleNull() {
...
+    assertThat(udf.exp(-1L), is(closeTo(0.36787944117144233, 0.00000000000000005)));
+    assertThat(udf.exp(-1.0), is(closeTo(0.36787944117144233, 0.00000000000000005)));
   }
 
   @Test
   public void shouldHandleZero() {
-    assertThat(udf.exp(0), is(1.0));
-    assertThat(udf.exp(0L), is(1.0));
-    assertThat(udf.exp(0.0), is(1.0));
+    assertThat(udf.exp(0), is(closeTo(1.0, 0.1)));
```
Gerrrr, line 50: Do these udfs really have epsilon around 0.1? Can't we use 5E-17 similar to other cases?

Gerrrr (commented):
Thank you for your contribution! I left a couple of comments. Also, I think that it makes sense to apply the same approach in `SqrtTest` where applicable.

Meanwhile, I'll try to find a CI that can run these tests on arm64.

ksqldb-engine/src/test/java/io/confluent/ksql/function/udf/math/ExpTest.java
martin-g: Thanks for the feedback, @Gerrrr !
I've improved it with your suggestion! Fully agreed that it is easier to read now! 
I've changed it to `1E-15`. Otherwise someone later may wonder why `5` and not `3` or `8`.

martin-g:
> Meanwhile, I'll try to find a CI that can run these tests on arm64

I guess you might have some internal ARM64 systems which you could use.
In case you don't then you could use Oracle Cloud. The provide Ampere Altra 80 4 x vCPU @ 2.8GHz and 24GB RAM in their Free Plan.
If Oracle Cloud is not an option then my employer could also provide a VM that you could use as a CI agent long term. Please let me know if you decide to go with that option!

Gerrrr:
So I've tried 2 options - Oracle Cloud and Amazon EC2 Graviton:
* Oracle Cloud does not let me create a free account because it does not like my credit card.
* Amazon EC2 Graviton instance created via a quick click-through does not let me connect to it via ssh - both from the terminal and web interface.

Unfortunately, I won't have time this week to debug EC2 instance configuration. I hope to return to this PR a bit later either when I have access to ARM64 VMs or time to go through AWS documentation.

martin-g:
> * Oracle Cloud does not let me create a free account because it does not like my credit card.

They do not work with pre-paid cards, like Revolut. Maybe this is your issue ...


> * Amazon EC2 Graviton instance created via a quick click-through does not let me connect to it via ssh - both from the terminal and web interface.

Maybe you forgot to upload your public SSH key first ?!

Just ping me at mgrigorov at apache org if you decide to use the donated VM!

PR #8342, opened 2021-11-09
feat: Continuation tokens and ALOS for SPQs 

AlanConfluent:
### Description 
This PR introduces a new type of consumer for SPQs, the `CatchupConsumer`.  It is a kafka consumer like the `LatestConsumer` except that it can seek back in the past and catch up to the latest.  When a `CatchupConsumer` is running it has the same partition assignment as the `LatestConsumer`, and when it catches up, offset for offset, the ProcessingQueue of the catchup is registered with the latest, and the `CatchupConsumer` is stopped.

This ability is now used both to implement ALOS for the client (enforced between ksqlDB nodes) as well as the ability to pick up where the client left off in case of a network fault.

For example, you could issue a query:

```
$ curl -X "POST" "http://localhost:8088/query-stream" --http2 -H "Content-Type: application/json; charset=utf-8" -d $'{ "sql": "SELECT * FROM RATINGS_STREAM emit changes;", "properties": {"ksql.query.push.v2.enabled": true, "auto.offset.reset": "latest", "ksql.query.push.v2.continuation.tokens.enabled": true} }'
{"queryId":"SCALABLE_PUSH_QUERY_3529861242791321853","columnNames":["ROUTE_ID","KEY","RATING_ID","USER_ID","STARS","RATING_TIME","CHANNEL","MESSAGE"],"columnTypes":["BIGINT","INTEGER","BIGINT","INTEGER","INTEGER","BIGINT","STRING","STRING"]}
[1044,185,1044,10,2,1636497872404,"iOS","your team here rocks!"]
{"continuationToken":"eyJzIjp7Im8iOls1Mjg0LDUzMjksNTEwMiw1MjU3LDUyOTUsNTM2N119LCJlIjp7Im8iOls1Mjg1LDUzMjksNTEwMiw1MjU3LDUyOTUsNTM2N119LCJ2IjowfQ=="}
[1045,838,1045,16,1,1636497873404,"iOS","thank you for the most friendly, helpful experience today at your new lounge"]
{"continuationToken":"eyJzIjp7Im8iOls1Mjg1LDUzMjksNTEwMiw1MjU3LDUyOTUsNTM2N119LCJlIjp7Im8iOls1Mjg1LDUzMjksNTEwMiw1MjU3LDUyOTUsNTM2OF19LCJ2IjowfQ=="}
....
```

Then, after a fault or at a later time, you can issue the following command to pick back up:
```
asheinberg-MBP15:wrk asheinberg$ curl -X "POST" "http://localhost:8088/query-stream" --http2 -H "Content-Type: application/json; charset=utf-8" -d $'{ "sql": "SELECT * FROM RATINGS_STREAM emit changes;", "properties": {"ksql.query.push.v2.enabled": true, "auto.offset.reset": "latest", "ksql.query.push.v2.continuation.tokens.enabled": true}, "requestProperties": {"request.ksql.query.push.start.offsets": "eyJzIjp7Im8iOls1Mjg2LDUzMjksNTEwMyw1MjU4LDUyOTUsNTM2OF19LCJlIjp7Im8iOls1Mjg2LDUzMjksNTEwMyw1MjU4LDUyOTYsNTM2OF19LCJ2IjowfQ=="} }'
{"queryId":"SCALABLE_PUSH_QUERY_3594658901457706443","columnNames":["ROUTE_ID","KEY","RATING_ID","USER_ID","STARS","RATING_TIME","CHANNEL","MESSAGE"],"columnTypes":["BIGINT","INTEGER","BIGINT","INTEGER","INTEGER","BIGINT","STRING","STRING"]}
[1055,726,1055,-1,2,1636497883404,"web","thank you for the most friendly, helpful experience today at your new lounge"]
[1056,604,1056,0,4,1636497884405,"android","(expletive deleted)"]
[1050,612,1050,5,2,1636497878404,"android","(expletive deleted)"]
[1057,116,1057,15,4,1636497885402,"iOS","meh"]
[1053,440,1053,0,4,1636497881404,"web","more peanuts please"]
[1058,442,1058,9,1,1636497886403,"android","why is it so difficult to keep the bathrooms clean ?"]
[1051,561,1051,18,3,1636497879404,"web","more peanuts please"]
[1054,280,1054,4,1,1636497882404,"android","thank you for the most friendly, helpful experience today at your new lounge"]
[1059,914,1059,2,3,1636497887401,"ios","thank you for the most friendly, helpful experience today at your new lounge"]
[1052,367,1052,4,4,1636497880404,"iOS-test","worst. flight. ever. #neveragain"]
{"continuationToken":"eyJzIjp7Im8iOls1Mjg2LDUzMjksNTEwMyw1MjU4LDUyOTYsNTM2OF19LCJlIjp7Im8iOls1Mjg3LDUzMzAsNTEwNiw1MjYwLDUyOTcsNTM3MF19LCJ2IjowfQ=="}
```

### Testing done 
Many unit tests + functional tests.  Will write a few more as well.

### Reviewer checklist


ksqldb-engine/src/main/java/io/confluent/ksql/physical/scalablepush/ScalablePushRegistry.java
```diff
@@ -167,33 +191,40 @@ public synchronized boolean isClosed() {
...
-    final LatestConsumer latestConsumer = this.latestConsumer.get();
-    if (latestConsumer != null && !latestConsumer.isClosed()) {
-      if (latestConsumer.getNumRowsReceived() > 0
-          && newNodeContinuityEnforced && expectingStartOfRegistryData) {
-        LOG.warn("Request missed data with new node added to the cluster, rows {}",
-            latestConsumer.getNumRowsReceived());
-        throw new KsqlException("New node missed data");
+    try {
+      if (catchupMetadata.isPresent()) {
+        if (catchupCounter.get() >=
```
agavra, line 205: I think I left this comment elsewhere as well, but I think it's a bit misleading to use `AtomicInteger` here when everything is synchronized anyway. When I saw below us `getAndIncrement` my first thought was "hold up, there's probably a race with whatever is checking it's contents that might let us get two into this block", then I noticed it's synchronized so we don't actually need to worry.

I'm worried that future engineers will make false assumptions (e.g. "oh I don't need synchronized because it's atomic")

ksqldb-engine/src/main/java/io/confluent/ksql/physical/scalablepush/ScalablePushRegistry.java
```diff
@@ -364,4 +457,129 @@ private void runLatest(final LatestConsumer latestConsumerToRun) {
...
+   * @param catchupMetadata The catchup metadata
+   */
+  private synchronized void startCatchup(
+      final ProcessingQueue processingQueue,
+      final CatchupMetadata catchupMetadata
+  ) {
+    final CatchupConsumer catchupConsumer = createCatchupConsumer(processingQueue,
+        catchupMetadata.getPushOffsetRange(),
+        catchupMetadata.getCatchupConsumerGroup());
+    catchupConsumer.register(processingQueue);
```
agavra, line 512: super nit: seems like this should just be in `createcatchupConsumer`, no?

ksqldb-engine/src/main/java/io/confluent/ksql/physical/scalablepush/ScalablePushRegistry.java
```diff
@@ -364,4 +457,129 @@ private void runLatest(final LatestConsumer latestConsumerToRun) {
...
+      catchupConsumer.run();
+    } catch (Throwable t) {
+      LOG.error("Got error while running catchup", t);
+      catchupConsumerToRun.onError();
+    } finally {
+      catchupConsumerToRun.unregister(processingQueue);
+      catchupConsumers.remove(processingQueue.getQueryId());
+      catchupCounter.decrementAndGet();
+      // If things ended exceptionally, stop latest
+      stopLatestConsumerOnLastRequest();
```
agavra, line 532: these are littered throughout the code base. is it possible to just have the latest consumer stop itself if there are no requests registered so we don't have to do that bookkeeping?

ksqldb-engine/src/main/java/io/confluent/ksql/physical/scalablepush/ScalablePushRegistry.java
```diff
@@ -364,4 +457,129 @@ private void runLatest(final LatestConsumer latestConsumerToRun) {
...
+  private synchronized void startCatchup(
+      final ProcessingQueue processingQueue,
+      final CatchupMetadata catchupMetadata
+  ) {
+    final CatchupConsumer catchupConsumer = createCatchupConsumer(processingQueue,
+        catchupMetadata.getPushOffsetRange(),
+        catchupMetadata.getCatchupConsumerGroup());
+    catchupConsumer.register(processingQueue);
+    catchupConsumers.put(processingQueue.getQueryId(), catchupConsumer);
+    catchupCounter.incrementAndGet();
```
agavra, line 514: why have a separate counter when we can just `catchupConsumers.size()`? seems like one less thing to bookkeep

ksqldb-engine/src/main/java/io/confluent/ksql/physical/scalablepush/ScalablePushRegistry.java
```diff
@@ -167,33 +191,40 @@ public synchronized boolean isClosed() {
...
-    if (latestConsumer != null && !latestConsumer.isClosed()) {
-      if (latestConsumer.getNumRowsReceived() > 0
-          && newNodeContinuityEnforced && expectingStartOfRegistryData) {
-        LOG.warn("Request missed data with new node added to the cluster, rows {}",
-            latestConsumer.getNumRowsReceived());
-        throw new KsqlException("New node missed data");
+    try {
+      if (catchupMetadata.isPresent()) {
+        if (catchupCounter.get() >=
+            ksqlConfig.getInt(KsqlConfig.KSQL_QUERY_PUSH_V2_MAX_CATCHUP_CONSUMERS)) {
```
agavra, line 206: general point about this PR - can we at least have JMX metrics around the number of running catchup consumers, the number of active query processing queues attached to the latest queue, etc..? Basically, anything that will help us debug incidents like we've had with hitting the pull query bandwidth limit and it not going down.

ksqldb-engine/src/main/java/io/confluent/ksql/physical/scalablepush/ScalablePushRegistry.java
```diff
@@ -364,4 +457,129 @@ private void runLatest(final LatestConsumer latestConsumerToRun) {
...
+      catchupConsumer.newAssignment(newAssignment);
+    }
+  }
+
+  /**
+   * Creates the latest consumer and its underlying kafka consumer.
+   * @param processingQueue The queue on which to send an error if anything goes wrong
+   * @return The new LatestConsumer
+   */
+  private CatchupConsumer createCatchupConsumer(
```
agavra, line 474: nit: wdyt about making all non-syncrhonized methods static and passing in everything we need here to ensure that we're not abusing thread unsafe state?

ksqldb-engine/src/main/java/io/confluent/ksql/physical/scalablepush/consumer/LatestConsumer.java
```diff
@@ -154,15 +142,20 @@ private void maybeSeekToEnd() {
...
           && entry.getValue().offset() > metadata.offset()) {
+        LOG.info("ALAN: About to seek seeking to end {}", currentPositions);
         consumer.seekToEnd(topicPartitions);
-        resetCurrentPosition();
+        updateCurrentPositions();
         LOG.info("LatestConsumer seeking to end {}", currentPositions);
         return;
       }
     }
+    LOG.info("ALAN: DID NOT SEEK TO END");
```
agavra, line 159: `I AM GROOT` (a few times 😆 )

in seriousness, if these are helpful logs should we include them at a trace or debug level?

ksqldb-engine/src/main/java/io/confluent/ksql/physical/scalablepush/consumer/ScalablePushConsumer.java
```diff
@@ -121,47 +132,89 @@ public void run() {
...
-        continue;
+        computeProgressToken(Optional.of(startOffsetVector));
+        try {
+          consumer.commitSync();
+          updateCommittedOffsets();
+        } catch (CommitFailedException e) {
+          LOG.warn("Failed to commit, likely due to rebalance.  Will wait for new assignment", e);
+        }
+
+        afterCommit();
```
agavra, line 173: should we call after commit even if the commit failed? if so maybe rename it?

ksqldb-engine/src/main/java/io/confluent/ksql/physical/scalablepush/consumer/ScalablePushConsumer.java
```diff
@@ -121,47 +132,89 @@ public void run() {
...
+  }
 
-      if (newAssignment) {
-        newAssignment = false;
-        onNewAssignment();
-      }
+  private void updateCommittedOffsets() {
+    final Map<TopicPartition, OffsetAndMetadata> offsets
+        = consumer.committed(new HashSet<>(topicPartitions.get()));
+    latestCommittedOffsets.set(ImmutableMap.copyOf(offsets));
```
agavra, line 184: perhaps too early to comment this, but it seems like we're a bit wasteful on object allocation in and around this code (lots of deep copying vectors around) - not sure it makes sense to revisit it but just thought I'd throw my hat down in that ring

ksqldb-engine/src/main/java/io/confluent/ksql/physical/scalablepush/consumer/ScalablePushConsumer.java
```diff
@@ -209,6 +262,7 @@ public void closeAsync() {
     for (final ProcessingQueue processingQueue : processingQueues.values()) {
       processingQueue.close();
     }
+    consumer.wakeup();
```
agavra, line 265: is this a bug fix?

agavra (commented):
Haven't looked at tests yet, and honestly I'm not totally sure I've reviewed all production code because I jumped around a lot and it's pretty big. It'd probably be good to get another set of eyes on it as well.

Generally I think the approach looks good to me - there's only one comment that I feel like I need more clarity on (about the "stutter" approach of synchronizing the latest/catchup consumers). I don't think there's any single place, though, that documents how the algorithms governing the scalable push query distributed state machine work - I think we really need that so that someone who wasn't part of the discussions can pick up and understand this work (especially re: optimistic gap detection, etc...).

ksqldb-engine/src/main/java/io/confluent/ksql/physical/scalablepush/ScalablePushRegistry.java
```diff
@@ -364,4 +457,129 @@ private void runLatest(final LatestConsumer latestConsumerToRun) {
...
+              .getConsumerGroupClient()
+              .listConsumerGroupOffsets(consumerGroupId);
+      if (metadata == null || metadata.values().stream().allMatch(Objects::isNull)) {
+        return;
+      }
+      serviceContext
+          .getConsumerGroupClient()
+          .deleteConsumerGroups(ImmutableSet.of(consumerGroupId));
+    } catch (Throwable t) {
+      LOG.error("Failed to delete consumer group " + consumerGroupId, t);
```
vvcephei, line 560: Will this lead to resource leakage?

ksqldb-engine/src/main/java/io/confluent/ksql/physical/scalablepush/consumer/CatchupConsumer.java
```diff
@@ -0,0 +1,248 @@
...
+        consumer.seek(tp, startingOffsets.get(tp.partition()));
+      }
+    }
+  }
+
+  protected void afterOfferedRow(final ProcessingQueue processingQueue) {
+    // Since we handle only one request at a time, we can afford to block and wait for the request.
+    while (processingQueue.isAtLimit()) {
+      LOG.info("Sleeping for a bit since queue is full");
+      sleepMs.accept(1000L);
```
vvcephei, line 174: What do you think about sleeping for a much shorter amount of time? It seems like this amount of pausing could put us into a sawtooth pattern, where we process a couple of batches, fill up the queue, then the consumer drains the queue a couple of ms later but we sit idle for a full second, then full up the queue again, etc.

ksqldb-engine/src/main/java/io/confluent/ksql/physical/scalablepush/consumer/CatchupConsumer.java
```diff
@@ -0,0 +1,248 @@
...
+  protected void onNewAssignment() {
+    onNewAssignment(Optional.empty());
+  }
+
+  /**
+   * The first time we start up a LatestConsumer, if it doesn't already have an assignment, we need
+   * to wait for it to give us that assignment by calling {@link #newAssignment}
+   * @return The assignment given by latest
+   */
+  protected Set<TopicPartition> waitForNewAssignmentFromLatestConsumer() {
```
vvcephei, line 136: Obviously, I'm not deeply familiar with this code (yet), but the relationship between the catchup and latest consumers looks strange to me. It seems like the latest consumer could "simply" manage the state of the catchup during rebalances and we would then not need any synchronization, atomic references, polling for assignments or state changes, etc. Maybe I'm missing something, though.

Coming back later with an edit: I'm also wondering if having separate threads for the latest and catchup consumers is worth all the hassle. It seems like we could ditch a lot of the complexity here by just taking turns polling them in the same thread, which would also be the same thread that manages all the assignment transitions. E.g., I don't think we'd need the stutter mechanism at all. Again, I might just be missing something here.

ksqldb-engine/src/main/java/io/confluent/ksql/physical/scalablepush/consumer/CatchupConsumer.java
```diff
@@ -0,0 +1,248 @@
...
+      if (lc == null) {
+        return;
+      }
+      for (final ProcessingQueue processingQueue : processingQueues.values()) {
+        lc.register(processingQueue);
+      }
+      processingQueues.clear();
+      close();
+    };
+    return catchupCoordinator.checkShouldCatchUp(signalledLatest, isCaughtUp, switchOver);
```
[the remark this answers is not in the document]
vvcephei: I haven't looked back to answer that last question, but it does remind me of the way we configure restore consumers in Streams with an "upper bound" so that they will never pass the point they are supposed to catch up to.

ksqldb-rest-app/src/test/java/io/confluent/ksql/rest/server/NetworkDisruptorClient.java
```diff
@@ -0,0 +1,151 @@
...
+import java.util.Optional;
+import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.CopyOnWriteArrayList;
+import java.util.concurrent.atomic.AtomicBoolean;
+import java.util.concurrent.atomic.AtomicReference;
+import java.util.function.Consumer;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class NetworkDisruptorClient implements SimpleKsqlClient {
```
vvcephei, line 24: Neato!

ksqldb-rest-client/src/test/java/io/confluent/ksql/rest/client/KsqlClientTest.java
```diff
@@ -739,8 +739,8 @@ public void shouldHandleArbitraryErrorsOnGetRequests() {
     List<StreamedRow> expectedResponse = new ArrayList<>();
     for (int i = 0; i < numRows; i++) {
       GenericRow row = GenericRow.genericRow("foo", 123, true);
-      StreamedRow sr = StreamedRow.pushRow(row);
-      expectedResponse.add(sr);
+//      StreamedRow sr = StreamedRow.pushRow(row);
+//      expectedResponse.add(sr);
```
vvcephei, line 743: What's up with this?

vvcephei (commented):
Awesome work, @AlanConfluent !

At a high level, I only have two concerns:
1. I'd echo @agavra 's thoughts about operability. This logic is pretty complex, and it seems likely that we're going to need to debug problems with it at some point, but I'm not sure we will have the logs and metrics we need to see what is happening.
2. I know this is late in the game, and I'd understand if you're not open to significant refactoring at this point, but I have the sense that we could simplify the whole catchup/latest consumer system by switching to a single-threaded model.

Smaller concern: there are a few instances of commented code and testing-looking log messages. It might be a good idea to make a final pass from top to bottom and make sure there's no cruft in the PR before merging.

Thanks!

ksqldb-common/src/main/java/io/confluent/ksql/util/PushOffsetVector.java
```diff
@@ -18,31 +18,79 @@
...
+  private final AtomicReference<List<Long>> offsets = new AtomicReference<>();
 
   @SuppressFBWarnings(value = "EI_EXPOSE_REP2")
   @JsonCreator
   public PushOffsetVector(final @JsonProperty(value = "o") List<Long> offsets) {
-    this.offsets = offsets;
+    this.offsets.set(ImmutableList.copyOf(offsets));
+  }
+
+  public PushOffsetVector() {
```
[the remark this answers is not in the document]
AlanConfluent: I agree that there's some overlap here.  We basically started off by saying that we would try to create a common interface and then we could use different implementations since I might want to store mine densely and serialize them differently.  In retrospect, after looking at them, they're similar enough that they should probably be merged at some point.

I added some of these thoughts to the documentation. 

Also, there are more issues with the "consistent pull query then read" in that the offsets are tracking different things.  For example: `CREATE table RATINGS_TABLE as SELECT * from SOURCE; SELECT * FROM RATINGS_TABLE WHERE COL > 10;`. It's my understanding that the offsets stored in the materialization and used for consistency are going to be the offsets of `SOURCE` whereas with SPQ, the offsets used for continuation are those of `RATINGS_TABLE`.  We can still potentially figure out some way to do this, such as just ask streams what are the current offsets written to `RATINGS_TABLE` before a pull query, but it's not going to be quite as straightforward as we were abstractly thinking.

ksqldb-engine/src/main/java/io/confluent/ksql/physical/scalablepush/consumer/CatchupCoordinatorImpl.java
```diff
@@ -0,0 +1,74 @@
...
+
+  @Override
+  public synchronized void checkShouldWaitForCatchup() {
+      while (catchupJoiners.get() > 0) {
+        try {
+          latestWaiting.set(true);
+          this.wait();
+        } catch (InterruptedException e) {
+          LOG.error("Caught InterruptedException during catchup waiting", e);
+          Thread.currentThread().interrupt();
```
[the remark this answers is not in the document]
AlanConfluent: Good call.  Yeah, made it bubble up an exception which should stop the thread.

ksqldb-engine/src/main/java/io/confluent/ksql/physical/scalablepush/consumer/LatestConsumer.java
AlanConfluent: Eek.  Added one of them since I think it's helpful.

ksqldb-engine/src/main/java/io/confluent/ksql/physical/scalablepush/consumer/ScalablePushConsumer.java
AlanConfluent: Changed it to `afterBatchProcessed`

ksqldb-engine/src/main/java/io/confluent/ksql/physical/scalablepush/consumer/ScalablePushConsumer.java
AlanConfluent: I deleted this particular one, but I still have the currentPositions.

Well, some of these objects are shared between latest and catchup, namely currentPositions, and it's important that they be threadsafe, hence the copying.  Not sure that there's a much better way of doing this while giving a snapshot effect.  Happy to address this if it's an issue, but given that it happens once a poll, that's really not too bad, and at least with latest is shared between all queries.

ksqldb-engine/src/main/java/io/confluent/ksql/physical/scalablepush/consumer/ScalablePushConsumer.java
AlanConfluent: I moved this to an other PR so shouldn't be a diff here.

ksqldb-rest-client/src/test/java/io/confluent/ksql/rest/client/KsqlClientTest.java
AlanConfluent: No idea.  Not sure why I would do such a thing.  Undone!

ksqldb-rest-app/src/test/java/io/confluent/ksql/rest/server/NetworkDisruptorClient.java
AlanConfluent: :-)

ksqldb-common/src/main/java/io/confluent/ksql/util/KsqlRequestConfig.java
```diff
@@ -56,12 +56,18 @@
   private static final String KSQL_REQUEST_QUERY_PUSH_SKIP_FORWARDING_DOC =
       "Controls whether a ksql host forwards a push query request to another host";
 
-  public static final String KSQL_REQUEST_QUERY_PUSH_REGISTRY_START =
-      "request.ksql.query.push.registry.start";
-  public static final boolean KSQL_REQUEST_QUERY_PUSH_REGISTRY_START_DEFAULT = false;
-  private static final String KSQL_REQUEST_QUERY_PUSH_REGISTRY_START_DOC =
-      "Indicates whether a connecting node expects to be at the start of the registry data. After a"
-          + "rebalance, this ensures we don't miss any data.";
```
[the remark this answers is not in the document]
AlanConfluent: These aren't part of the official configs, they're internal request only, so I don't think this is required.

ksqldb-common/src/main/java/io/confluent/ksql/util/PushOffsetVector.java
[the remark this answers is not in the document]
vvcephei: That's correct. This is the main technical work we'll have to tackle before we can support "query and subscribe".

ksqldb-engine/src/main/java/io/confluent/ksql/physical/scalablepush/PushPhysicalPlanBuilder.java
```diff
@@ -162,16 +166,27 @@ private SelectOperator translateFilterNode(final QueryFilterNode logicalNode) {
...
+    return new PeekStreamOperator(scalablePushRegistry, logicalNode, queryId, catchupMetadata);
+  }
+
+  private String getConsumerGroupId(final Optional<String> catchupConsumerGroupFromSource) {
+    final ScalablePushRegistry scalablePushRegistry =
+        persistentQueryMetadata.getScalablePushRegistry()
+            .orElseThrow(() -> new IllegalStateException("Scalable push registry cannot be found"));
+    return catchupConsumerGroupFromSource.orElse(
+        scalablePushRegistry.getCatchupConsumerId(
+            "" + Math.abs(ThreadLocalRandom.current().nextLong())));
```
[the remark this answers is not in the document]
vvcephei: Yeah, I recall a paper/post/thing from Microsoft "proving" that the likelihood of a collision from a random UUID is lower than the likelihood of a collision from an atomic counter getting its state messed up by a cosmic ray... Might be hyperbolic, but, my understanding is that if you're using a UUID, you really don't need to check for collisions.

ksqldb-engine/src/main/java/io/confluent/ksql/engine/EngineExecutor.java
```diff
@@ -345,13 +348,21 @@ ScalablePushQueryMetadata executeScalablePushQuery(
...
+              offsetRange,
+              catchupConsumerGroup
+          );
+      final Optional<PushOffsetRange> offsetRange = pushRoutingOptions.getContinuationToken()
+          .map(PushOffsetRange::deserialize);
+      final Optional<String> catchupConsumerGroup = pushRoutingOptions.getCatchupConsumerGroup();
+      final PushPhysicalPlanManager physicalPlanManager = new PushPhysicalPlanManager(
+          pushPhysicalPlanCreator, catchupConsumerGroup, offsetRange);
+      final PushPhysicalPlan physicalPlan = physicalPlanManager.getPhysicalPlan();
+      plan = physicalPlan;
```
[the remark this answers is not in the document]
vvcephei: Sounds good. Thanks for taking a look.

ksqldb-engine/src/main/java/io/confluent/ksql/physical/scalablepush/consumer/CatchupConsumer.java
vvcephei: Sounds like a good plan to me. This isn't the kind of idea worth stopping the train for, but good to keep in mind if things get hard to manage in practice.

ksqldb-engine/src/main/java/io/confluent/ksql/physical/scalablepush/PushPhysicalPlanBuilder.java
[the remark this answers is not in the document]
vvcephei: I notice that line 194 changed, but not this one... Was that intentional?

ksqldb-engine/src/main/java/io/confluent/ksql/physical/scalablepush/ScalablePushRegistry.java
```diff
@@ -139,6 +161,15 @@ public synchronized void close() {
...
       executorService.shutdownNow();
     }
+    executorServiceCatchup.shutdown();
+    try {
+      if (!executorServiceCatchup.awaitTermination(5000, TimeUnit.MILLISECONDS)) {
+        executorServiceCatchup.shutdownNow();
+      }
+    } catch (InterruptedException e) {
+      LOG.error("Interrupted during shutdown", e);
+      executorServiceCatchup.shutdownNow();
```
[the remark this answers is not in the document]
vvcephei: bump

ksqldb-engine/src/main/java/io/confluent/ksql/physical/scalablepush/ScalablePushRegistry.java
```diff
@@ -230,9 +272,14 @@ public synchronized boolean isLatestRunning() {
     return latestConsumer != null && !latestConsumer.isClosed();
   }
 
+  @VisibleForTesting
+  public synchronized boolean anyCatchupsRunning() {
+    return catchupConsumers.size() > 0;
+  }
+
   @VisibleForTesting
   public int numRegistered() {
```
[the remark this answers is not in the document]
vvcephei: bump

ksqldb-engine/src/main/java/io/confluent/ksql/physical/scalablepush/ScalablePushRegistry.java
vvcephei: bump

vvcephei (approved):
Thanks, @AlanConfluent , I had a couple of questions that didn't seem to be resolved, but nothing worth delaying this PR for if you're ready to merge.

ksqldb-engine/src/main/java/io/confluent/ksql/physical/scalablepush/ScalablePushRegistry.java
```diff
@@ -139,6 +161,15 @@ public synchronized void close() {
...
     }
+    executorServiceCatchup.shutdown();
+    try {
+      if (!executorServiceCatchup.awaitTermination(5000, TimeUnit.MILLISECONDS)) {
+        executorServiceCatchup.shutdownNow();
+      }
+    } catch (InterruptedException e) {
+      LOG.error("Interrupted during shutdown", e);
+      executorServiceCatchup.shutdownNow();
+    }
```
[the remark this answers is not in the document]
AlanConfluent: Done!

ksqldb-engine/src/main/java/io/confluent/ksql/physical/scalablepush/ScalablePushRegistry.java
```diff
@@ -167,33 +191,40 @@ public synchronized boolean isClosed() {
...
-          && newNodeContinuityEnforced && expectingStartOfRegistryData) {
-        LOG.warn("Request missed data with new node added to the cluster, rows {}",
-            latestConsumer.getNumRowsReceived());
-        throw new KsqlException("New node missed data");
+    try {
+      if (catchupMetadata.isPresent()) {
+        if (catchupCounter.get() >=
+            ksqlConfig.getInt(KsqlConfig.KSQL_QUERY_PUSH_V2_MAX_CATCHUP_CONSUMERS)) {
+          processingQueue.onError();
+          throw new KsqlException("Too many catchups registered");
```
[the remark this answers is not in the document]
AlanConfluent: Done

ksqldb-engine/src/main/java/io/confluent/ksql/physical/scalablepush/ScalablePushRegistry.java
```diff
@@ -167,33 +191,40 @@ public synchronized boolean isClosed() {
...
+        }
+        // Latest much be running for a catchup consumer to exist.
```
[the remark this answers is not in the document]
AlanConfluent: Done

ksqldb-engine/src/main/java/io/confluent/ksql/physical/scalablepush/ScalablePushRegistry.java
AlanConfluent: Yeah, I agree.  The issue is that the decrement was happening outside of synchronized, which I considered a fine thing (maybe it was below the limit by the time the error was thrown).  In this case, I agree it's easier to just use synchronized.

ksqldb-engine/src/main/java/io/confluent/ksql/physical/scalablepush/ScalablePushRegistry.java
AlanConfluent: I'm just doing normal camelCase.  Would you normally not capitalize in this case?

ksqldb-engine/src/main/java/io/confluent/ksql/physical/scalablepush/ScalablePushRegistry.java
AlanConfluent: The issue is that it needs to have latest running even if the only queue is for a catchup.  You might want to stop if if there's an error creating the catchup or if you unregister the last queue.  Not sure there's a way around it.

ksqldb-engine/src/main/java/io/confluent/ksql/physical/scalablepush/ScalablePushRegistry.java
AlanConfluent: Good call.  Not sure why I didn't see that.

ksqldb-engine/src/main/java/io/confluent/ksql/physical/scalablepush/PushPhysicalPlanBuilder.java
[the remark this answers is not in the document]
AlanConfluent: Nope.  Changed!

ksqldb-engine/src/main/java/io/confluent/ksql/physical/scalablepush/ScalablePushRegistry.java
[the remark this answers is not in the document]
AlanConfluent: This  has been replaced with Almogs suggestion of MoreExecutors.shutdownAndAwaitTermination

ksqldb-engine/src/main/java/io/confluent/ksql/physical/scalablepush/ScalablePushRegistry.java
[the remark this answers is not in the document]
AlanConfluent: That's a reasonable comment.  I think some things were synchronized and other were objects which are threadsafe.  I added synchronized to a bunch of other places to simplify this and only keep it unsynchronized in private methods where it's done intentionally (e.g. while running a consumer).

ksqldb-engine/src/main/java/io/confluent/ksql/physical/scalablepush/ScalablePushRegistry.java
AlanConfluent: Well, eventually the consumer groups will be GCed on their own by Kafka.  But yes, they theoretically could.

AlanConfluent:
@vvcephei Sorry for not addressing some of those comments.  I had a hard time finding them after I did a rebase, but I've addressed them now.

PR #8351, opened 2021-11-10
fix: Makes PullQueryRoutingFunctionalTest  more reliable and adds back test

AlanConfluent:
### Description 
Adds back the test `shouldQueryActiveWhenActiveAliveQueryIssuedToStandby` which was removed when it was being flaky.  Also, attempts to address flakiness by using a static task assigner for streams which should consistently give the same tasks over and over again to the same servers, so rebalancing won't be an issue.

### Testing done 
Ran tests!

### Reviewer checklist


agavra (approved):
lgtm! thanks for the fix

PR #8355, opened 2021-11-11
chore: refactor SharedKafkaStreamsRuntime into abstract class

ableegoldman:
Pulled out all the basic refactorings into a separate PR with no logical/behavioral changes. This PR includes

1. renamed a few classes to make them more consistent with the super classes or naming conventions in ksql
2. renamed a few fields in preparation for the next set of changes
2. changed `SharedKafkaStreamsRuntime` from an interface to an abstract class, since the two classes previously implementing it shared a large amount of common logic/method implementations. Only the identical methods were moved to the abstract class in this PR, everything else 
3. removed unused `closed` field in query metadata


ksqldb-engine/src/main/java/io/confluent/ksql/util/ValidationSharedKafkaStreamsRuntimeImpl.java
```diff
@@ -15,192 +15,103 @@
...
+        new QueryMetadataImpl.TimeBoundedQueue(Duration.ofHours(1), 0)
     );
-    kafkaStreamsBuilder = sharedKafkaStreamsRuntime.getKafkaStreamsBuilder();
-    kafkaStreams = kafkaStreamsBuilder.buildNamedTopologyWrapper(streamsProperties);
-    metadata = new ConcurrentHashMap<>(sharedKafkaStreamsRuntime.getMetadata());
-    sources = new ConcurrentHashMap<>(sharedKafkaStreamsRuntime.getSourcesMap());
-    for (PersistentQueriesInSharedRuntimesImpl queryMetadata : metadata.values()) {
+
+    for (BinPackedPersistentQueryMetadata query : sharedRuntime.collocatedQueries.values()) {
       //kafkaStreams.addNamedTopology(queryMetadata.getTopology());
```
wcarlson5, line 45: This was commented out because adding named topologies  was breaking the tests. I think it might have been because we were reusing the topology. We need to be able to copy the topology. Do we have this work tracked?

ksqldb-engine/src/main/java/io/confluent/ksql/util/ValidationSharedKafkaStreamsRuntimeImpl.java
```diff
@@ -15,192 +15,103 @@
...
 
-  public void markSources(final QueryId queryId, final Set<SourceName> sourceNames) {
-    sources.put(queryId, sourceNames);
+  private static Map<String, Object> getSandboxStreamsProperties(
+      final SharedKafkaStreamsRuntime sharedKafkaStreamsRuntime
+  ) {
+    final Map<String, Object> sandboxStreamsProperties =
+        new ConcurrentHashMap<>(sharedKafkaStreamsRuntime.getStreamProperties());
+    sandboxStreamsProperties.put(
+        StreamsConfig.APPLICATION_ID_CONFIG,
```
wcarlson5, line 67: We fixed the format that we make appIds for the normal runtimes here https://github.com/confluentinc/ksql/blob/03af146beb1cc68bea95779c8360b70f10708b01/ksqldb-engine/src/main/java/io/confluent/ksql/query/QueryBuilder.java#L504

Can you make this match? it was for metrics so it shouldn't be a huge deal but I would rather have the runtime Ids match as much as possible

ksqldb-engine/src/main/java/io/confluent/ksql/util/ValidationSharedKafkaStreamsRuntimeImpl.java
```diff
@@ -15,192 +15,103 @@
...
-
-  public boolean isError(final QueryId queryId) {
-    return false;
+    return StreamThreadExceptionResponse.REPLACE_THREAD;
   }
 
+  @Override
   public void stop(final QueryId queryId) {
-
+    log.trace("");
```
wcarlson5, line 104: Maybe fill this out?, I don't think this should be called

wcarlson5 (commented):
I think this is really close. I just have a few minor things

ksqldb-engine/src/main/java/io/confluent/ksql/util/ValidationSharedKafkaStreamsRuntimeImpl.java
ableegoldman: whoops, thanks for the catch

ksqldb-engine/src/main/java/io/confluent/ksql/util/QueryMetadataImpl.java
```diff
@@ -315,10 +313,6 @@ public void setCorruptionQueryError() {
     corruptionCommandTopic = true;
   }
 
-  protected boolean isClosed() {
```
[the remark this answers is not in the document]
ableegoldman: Yep

ksqldb-engine/src/main/java/io/confluent/ksql/util/SharedKafkaStreamsRuntime.java
```diff
@@ -25,57 +28,141 @@
...
+        "Marking source {} for query {} the mapping for this runtime is {}",
+        sourceNames,
+        queryId.toString(),
+        sourceNames
+    );
+  }
+
+  public abstract void register(
       QueryErrorClassifier errorClassifier,
-      Map<String, Object> streamsProperties,
```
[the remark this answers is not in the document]
ableegoldman: This was unused in either implementation, we don't seem to need to query props for `register` (at least not in the current code)

ksqldb-engine/src/main/java/io/confluent/ksql/util/ValidationSharedKafkaStreamsRuntimeImpl.java
ableegoldman: This is covered in my next PR, which includes all the work related to the query validation and sandboxed runtimes that I'm splitting out of the initial mega-PR

So short answer is, yes. Ticket is https://confluentinc.atlassian.net/browse/KCI-909

ksqldb-engine/src/main/java/io/confluent/ksql/query/QueryBuilder.java
```diff
@@ -458,7 +458,7 @@ PersistentQueryMetadata buildPersistentQueryInSharedRuntime(
         serviceContext
     );
 
-    return new PersistentQueriesInSharedRuntimesImpl(
+    return new BinPackedPersistentQueryMetadata(
```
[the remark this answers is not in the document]
ableegoldman: We should follow the naming convention for the QueryMetadata classes. I figured it was fine without the `Impl` since there's no interface to confuse it with, but I'm also fine with adding `Impl` if you have a preference. Not sure what the naming convention is when it comes to classses who don't strictly need the `Impl` suffix?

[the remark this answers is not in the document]
wcarlson5: `public class PersistentQueryMetadataImpl
    extends QueryMetadataImpl implements PersistentQueryMetadata` We do have this. I think its useful to have the `impl` because then when dealing out side the abstraction you can use `PersistentQueryMetadata` without having to check. But its not a hill i am willing to die on

ksqldb-engine/src/main/java/io/confluent/ksql/util/SharedKafkaStreamsRuntime.java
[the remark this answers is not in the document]
wcarlson5: So we will need to add dome form of it back sometime when we receive the query specific props

wcarlson5 (approved):
LGTM, if you can get [this](https://github.com/confluentinc/ksql/pull/8355/files#r750733900) in the PR or one of the follow ups that would be nice but otherwise Im convinced

ksqldb-engine/src/main/java/io/confluent/ksql/query/QueryBuilder.java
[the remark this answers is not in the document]
ableegoldman: 🤔  but in that example the `Impl` _is_ necessary due to the interface that would otherwise have the same name. Whereas here there's no other interface, the `BinPackedPersistentQueryMetadata` just directly implements `PersistentQueryMetadata`

But fair enough, I'll leave the `Impl` suffix in the name

ksqldb-engine/src/main/java/io/confluent/ksql/util/ValidationSharedKafkaStreamsRuntimeImpl.java
ableegoldman: Will do. Actually this right here should just be tacking on the `"-validation"` suffix to the application id of the original/real, but thanks for reminding/pointing me to that --  I noticed there's a `QueryApplicationId` builder class already so I factored the shared runtimes id logic into a separate builder method in that class so we can keep all the application id logic together. See [here ](https://github.com/confluentinc/ksql/pull/8355/files#diff-fd36d70d052543249b4a009833353071168d34ea1631d26d49f32d8f41abce41R28-R40)

ksqldb-common/src/main/java/io/confluent/ksql/util/QueryApplicationId.java
```diff
@@ -25,10 +25,35 @@
...
+  public static String buildSharedRuntimeId(
+      final KsqlConfig config,
+      final boolean persistent,
+      final int sharedRuntimeIndex
+  ) {
+    final String queryAppId = buildPrefix(config, persistent) + "-" + sharedRuntimeIndex;
+    if (persistent) {
+      return queryAppId;
+    } else {
+      return addTimeSuffix(queryAppId);
```
ableegoldman, line 37: @walker I changed the application id for shared runtimes very slightly, to use this "time suffix" instead of a UUID, which is what we seem to be doing for dedicated KafkaStreams applications currently and should be sufficient for getting a unique app id. This'll probably end up being removed in the next PR or two but I'm trying to keep this PR logically the same so it's easy to merge, so just swapping a UUID for a timestamp seemed like a trivial enough change for this PR

ksqldb-engine/src/main/java/io/confluent/ksql/query/QueryBuilder.java
```diff
@@ -493,35 +492,23 @@ private SharedKafkaStreamsRuntime getKafkaStreamsInstance(
...
     final SharedKafkaStreamsRuntime stream;
     final KsqlConfig ksqlConfig = config.getConfig(true);
-    final String queryPrefix = ksqlConfig
-        .getString(KsqlConfig.KSQL_PERSISTENT_QUERY_NAME_PREFIX_CONFIG);
-    final String serviceId = ksqlConfig.getString(KsqlConfig.KSQL_SERVICE_ID_CONFIG);
     if (real) {
       stream = new SharedKafkaStreamsRuntimeImpl(
           kafkaStreamsBuilder,
-          config.getConfig(true).getInt(KsqlConfig.KSQL_QUERY_ERROR_MAX_QUEUE_SIZE),
+          ksqlConfig.getInt(KsqlConfig.KSQL_QUERY_ERROR_MAX_QUEUE_SIZE),
```
ableegoldman, line 498: Just some minor cleanup to use the `ksqlConfig` variable everywhere and more importantly, factor the shared runtime id logic into a single builder method in `QueryApplicationId` (where we currently have the method for building regular application ids)

ableegoldman:
I believe I've addressed all your comments and requests. Unfortunately the PR builder is broken and may be for the next few days (due to the 7.2 branch cut, see here) , so I'm just going to run the tests locally and make sure everything looks good then merge it so I can continue the work building on top of this PR

ableegoldman:
Tests passed locally, going to merge this so I can rebase my other PR(s). Thanks for the review Walker!

ksqldb-common/src/main/java/io/confluent/ksql/util/QueryApplicationId.java
wcarlson5: Thats not my tag :). but yeah I don't mind that change

PR #8357, opened 2021-11-12
feat: support custom headers for connector requests

vcrfxia:
### Description 

Two changes in this PR:
* introduces `KsqlPrincipal` to track additional properties along with a user `Principal`
* adds a new extension point (`ConnectRequestHeadersExtension`) to allow configuration of custom headers to be sent with ksqlDB connector requests

This PR is still a WIP -- tests need to be added.

### Testing done 

Manual. Automated tests will be added soon.

### Reviewer checklist


ksqldb-engine/src/main/java/io/confluent/ksql/security/KsqlPrincipal.java
```diff
@@ -0,0 +1,30 @@
...
+ * WARRANTIES OF ANY KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations under the License.
+ */
+
+package io.confluent.ksql.security;
+
+import java.security.Principal;
+import java.util.Map;
+
+public interface KsqlPrincipal extends Principal {
```
Gerrrr, line 21: This class should be used instead of `Principal` pretty much everywhere in the ksql codebase, right? If so, we should probably document that in the javadoc.

ksqldb-engine/src/main/java/io/confluent/ksql/security/KsqlSecurityContext.java
```diff
@@ -16,33 +16,32 @@
...
+      final Optional<KsqlPrincipal> userPrincipal,
       final ServiceContext serviceContext
   ) {
     this.userPrincipal = userPrincipal;
     this.serviceContext = serviceContext;
   }
 
   /**
-   * Returns a {@code java.security.Principal} object containing the name of the current
+   * Returns a {@code KsqlPrincipal} object containing the name and properties of the current
```
Gerrrr, line 37: nit: `{@code KsqlPrincipal}` should actually be `{@link KsqlPrincipal}`. Then, one can click on it to open the referred class.

ksqldb-engine/src/main/java/io/confluent/ksql/security/KsqlSecurityContext.java
```diff
@@ -16,33 +16,32 @@
...
    * authenticated user. If the user has not been authenticated, the method returns
    * {@code Optional.empty}.
    *
-   * @return a {@code java.security.Principal} containing the name of the user making this request;
-   *         {@code Optional.empty} if the user has not been authenticated
+   * @return a {@code KsqlPrincipal} containing the name and properties of the user making
```
Gerrrr, line 41: nit: same as above

ksqldb-engine/src/main/java/io/confluent/ksql/services/DefaultConnectClient.java
```diff
@@ -285,10 +291,25 @@ public DefaultConnectClient(
     return authHeader;
   }
 
+  @VisibleForTesting
+  Map<String, String> getAdditionalRequestHeaders() {
+    return additionalRequestHeaders;
+  }
+
   private Header[] headers() {
```
Gerrrr, line 299: nit: it appears to me that both `authHeader` and `additionalRequestHeaders` do not change over time. Is this right? If so, we can probably create `Header[]` once and just serve it from this method.

ksqldb-rest-app/src/main/java/io/confluent/ksql/api/auth/JaasAuthProvider.java
```diff
@@ -128,7 +130,16 @@ private void getUser(
     // We do the actual authorization here not in the User class
     final boolean authorized = validateRoles(lc, allowedRoles);
 
-    promise.complete(new JaasUser(username, password, authorized));
+    // if the subject from the login context is already a KsqlPrincipal, use the subject
+    // directly rather than creating a new one
+    final Optional<KsqlPrincipal> ksqlPrincipal = lc.getSubject().getPrincipals().stream()
```
Gerrrr, line 135: Do we ever expect more than 1 principal here?

ksqldb-engine/src/main/java/io/confluent/ksql/security/KsqlPrincipal.java
```diff
@@ -0,0 +1,30 @@
...
+
+  Map<String, Object> getUserProperties();
```
Gerrrr, line 23: This like a good candidate for a default implementation returning an empty map.

Gerrrr (commented):
Overall looks good, but I feel that I miss the full picture here (you can probably tell from my comments). I'd suggest @spena to take a look as well.

ksqldb-rest-app/src/main/java/io/confluent/ksql/api/auth/JaasAuthProvider.java
```diff
@@ -128,7 +130,16 @@ private void getUser(
...
+        .filter(p -> p instanceof KsqlPrincipal)
+        .map(p -> (KsqlPrincipal)p)
+        .findFirst();
+    final JaasUser user = ksqlPrincipal.isPresent()
+        ? new JaasUser(ksqlPrincipal.get(), authorized)
+        : new JaasUser(username, password, authorized);
```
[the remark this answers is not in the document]
spena: The password is used in the confluent-security-plugins to create the impersonated user context. It is important to make autho. requests with the basic credentials.

If you plan to add it as a required param in `KsqlPrincipal`, then does it make sense to delete the `JaasPrincipal`? It won't be used anymore as KsqlPrincipal will have the same info.

ksqldb-engine/src/main/java/io/confluent/ksql/connect/ConnectRequestHeadersExtension.java
```diff
@@ -0,0 +1,40 @@
...
+   * Creates custom headers to be included with all connector requests made by ksqlDB.
+   * Note that this is in addition to any headers already sent by ksqlDB by default,
+   * such as those required for authenticating with Connect.
+   *
+   * @param userPrincipal principal associated with the user who submitted the connector
+   *                      request to ksqlDB, if present (i.e., if user authentication
+   *                      is enabled)
+   * @return additional headers to be included with connector requests made by ksqlDB
+   */
+  Map<String, String> getHeaders(Optional<KsqlPrincipal> userPrincipal);
```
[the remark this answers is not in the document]
vcrfxia: The javadoc a few lines up contains
```
   * Note that this is in addition to any headers already sent by ksqlDB by default,
   * such as those required for authenticating with Connect.
```
Do you think that's clear enough, or should I add more?

ksqldb-rest-app/src/main/java/io/confluent/ksql/api/auth/JaasAuthProvider.java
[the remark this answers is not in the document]
vcrfxia: We still need to keep some KsqlPrincipal implementation in the ksql repo, since users can use basic authN without a custom auth plugin that returns a KsqlPrincipal. For this reason, I think we still want to keep JaasPrincipal.

ksqldb-engine/src/main/java/io/confluent/ksql/services/DefaultConnectClient.java
vcrfxia: Yeah, this is true. The implementation was already this way (separate `headers()` method on each request) before my changes so I didn't touch it. I can look into refactoring, either as part of this PR or as a follow-up.

ksqldb-rest-app/src/main/java/io/confluent/ksql/api/auth/JaasAuthProvider.java
vcrfxia: The ksql codebase is only really set up for a single principal based on the interfaces used throughout the engine, so I think it's fine to only expect one.

ksqldb-engine/src/main/java/io/confluent/ksql/services/DefaultConnectClient.java
```diff
@@ -285,10 +291,25 @@ public DefaultConnectClient(
     return authHeader;
   }
 
+  @VisibleForTesting
+  Map<String, String> getAdditionalRequestHeaders() {
+    return additionalRequestHeaders;
+  }
+
```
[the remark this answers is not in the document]
vcrfxia: Unit testing is simpler by exposing these two components separately but, based on the other conversation about building `headers` in the constructor, I'll get rid of these accessors when I make that refactor.

ksqldb-engine/src/main/java/io/confluent/ksql/security/KsqlPrincipal.java
vcrfxia: Sure, I'll add a javadoc.

ksqldb-engine/src/main/java/io/confluent/ksql/services/DefaultConnectClient.java
vcrfxia: Made this change as part of this PR.

ksqldb-engine/src/main/java/io/confluent/ksql/security/KsqlPrincipal.java
vcrfxia: Done, though I left the no-op overrides in the two implementation classes (JaasPrincipal and DefaultKsqlPrincipal) for clarity. I can remove the redundancy if preferred.

ksqldb-engine/src/main/java/io/confluent/ksql/services/DefaultConnectClient.java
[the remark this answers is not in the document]
vcrfxia: Made the change above for refactoring `headers`. I replaced the previous two `VisibleForTesting` accessors with a new one to expose `requestHeaders`.

vcrfxia:
Hey @Gerrrr @spena thanks for your reviews! Made the requested changes, including the ones from the previous PR. Can y'all give this another review when you get a chance? Thanks.

ksqldb-engine/src/main/java/io/confluent/ksql/services/DefaultConnectClient.java
Gerrrr: Thank you!

Gerrrr (approved):
LGTM!

ksqldb-engine/src/main/java/io/confluent/ksql/connect/ConnectRequestHeadersExtension.java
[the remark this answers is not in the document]
spena: Perhaps just about what takes precedence . If I add an authHeader, what header tkaes precedence, the default or the new one I am adding?

ksqldb-engine/src/main/java/io/confluent/ksql/services/DefaultConnectClient.java
```diff
@@ -344,6 +343,25 @@ private String resolveUri(final String relativePath) {
...
+    authHeader.ifPresent(header -> headers.add(
+        new BasicHeader(HttpHeaders.AUTHORIZATION.toString(), authHeader.get())));
+
+    if (!additionalRequestHeaders.isEmpty()) {
+      final List<Header> additionalHeaders = additionalRequestHeaders.entrySet().stream()
+          .map(e -> new BasicHeader(e.getKey(), e.getValue()))
+          .collect(Collectors.toList());
+      headers.addAll(additionalHeaders);
+    }
```
spena, line 360: This is related to the comment of the headers precedence. If I set an authHeader in additionalHeaders, then this code is going to override it. Should we keep the default ksql header or allow the user to override it?

spena (approved):
Looks good. Just the comment about not overriding the authheader?

ksqldb-engine/src/main/java/io/confluent/ksql/services/DefaultConnectClient.java
vcrfxia: Hm good question. I could see the argument going either way. Currently, I think it's nice for users to be able to override the auth header if they want. This could be used as a workaround for supporting custom forms of authN other than basic auth, since that's currently the only form we support today. However, I could see this override behavior being confusing in the future, if more ksql headers are introduced.

I'm leaning towards leaving the behavior for now, unless you feel otherwise.

ksqldb-engine/src/main/java/io/confluent/ksql/connect/ConnectRequestHeadersExtension.java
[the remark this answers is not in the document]
vcrfxia: Ah got'cha -- I left some thoughts in https://github.com/confluentinc/ksql/pull/8357#discussion_r770172332. I will definitely update the javadoc to include this info as you suggested, but I'd like to do it in a follow-up PR in order to merge this one sooner to unblock some other PRs. I'm going to need a follow-up PR to update the use of `token` in `JaasPrincipal` (as we discussed) anyway, so I can slip the change into there.

PR #8367, opened 2021-11-16
fix: Makes tests wait only 500ms for consumer group cleanup

AlanConfluent:
### Description 
Currently, the `QueryCleanupService` waits 30s on shutdown.  For some hard to debug reason, in tests, I constantly get `GroupNotEmptyException`s for query ids of queries that have been deleted in tests.  This makes mosts functional tests timeout at 30s.

This PR makes that value a config and sets it to 500ms in functional tests.  After all, when a server is shut down in a test, the state is going to be wiped out anyway.  This makes many functional tests go from 40s to 10s in my testing which makes it much easier to deal with and will hopefully speed up our test runs.

### Testing done 
Ran tests.

### Reviewer checklist


agavra (approved):
LGTM, happy to try it out and see if it causes any flakiness in the tests.

PR #8381, opened 2021-11-17
KSE-599: Add token field to JaasPrincipal

Gerrrr:
### Description 

This patch adds `token` field to the `JaasPrincipal` and a `password` to the constructor. We need the token to authenticate to MDS with that principal. See https://github.com/confluentinc/confluent-security-plugins/pull/417 and https://github.com/confluentinc/confluent-cloud-plugins/pull/239 for more details.

### Testing done 

I ran relevant unit tests locally to make sure that changes in constructor don't break anything. Then, I built the `ksqldb-rest-app` jar and tested https://github.com/confluentinc/confluent-security-plugins/pull/417 with it. After that, I built `confluent-ksql-security-plugin` and tested https://github.com/confluentinc/confluent-cloud-plugins/pull/239 with it.

### Reviewer checklist

### DIFF
--- a/ksqldb-rest-app/src/main/java/io/confluent/ksql/api/auth/JaasPrincipal.java
+++ b/ksqldb-rest-app/src/main/java/io/confluent/ksql/api/auth/JaasPrincipal.java
@@ -15,18 +15,32 @@
 
 package io.confluent.ksql.api.auth;
 
+import java.nio.charset.StandardCharsets;
 import java.security.Principal;
+import java.util.Base64;
 import java.util.Objects;
 
 /**
  * Principal implementation created when authenticating with the JaasAuthProvider
  */
-class JaasPrincipal implements Principal {
+public class JaasPrincipal implements Principal {
 
   private final String name;
 
-  JaasPrincipal(final String name) {
+  private final String token;
+
+  public JaasPrincipal(final String name, final String password) {
     this.name = Objects.requireNonNull(name);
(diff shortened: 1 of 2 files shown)


ksqldb-rest-app/src/main/java/io/confluent/ksql/api/auth/JaasPrincipal.java
```diff
@@ -15,18 +15,32 @@
...
+public class JaasPrincipal implements Principal {
 
   private final String name;
 
-  JaasPrincipal(final String name) {
+  private final String token;
+
+  public JaasPrincipal(final String name, final String password) {
     this.name = Objects.requireNonNull(name);
+    this.token = createToken(name, Objects.requireNonNull(password));
```
spena, line 34: Ok, I was hoping to pass the token directly to the constructor. But I see the password is already decoded from the JaasAuthProvider, so it make sense to pass the password. In this case, just keep the password in the field. I thought we were going to get access to the HTTP authorization header, which has the encoded string, and pass it to the Principal. 

The MDS API accept the user/pass too, so we can use that and avoid to encode the string there

Gerrrr: 1b28358

spena (approved):
I just left one minor change. 

The code looks good, +1

PR #8382, opened 2021-11-17
docs: fix docs to reflect that `fail.on.*.error` configs cannot be applied per query

vcrfxia:
### Description 

The configs `ksql.fail.on.production.error` and `ksql.fail.on.deserialization.error` can only be set at a server level, and cannot be applied per-query. This PR fixes an error in the docs to reflect this.

Once merged, I'll cherry-pick to the (recent) live ksqlDB docs branches as well.

### Testing done 

Docs-only change.

### Reviewer checklist

### DIFF
--- a/docs/reference/server-configuration.md
+++ b/docs/reference/server-configuration.md
@@ -73,7 +73,7 @@ an external `ksql.connect.url`.
 
 ## `ksql.fail.on.deserialization.error`
 
-**Per query:** yes
+**Per query:** no
 
 Indicates whether to fail if corrupt messages are read. ksqlDB decodes
 messages at runtime when reading from a Kafka topic. The decoding that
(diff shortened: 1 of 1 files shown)


jzaralim (approved):

JimGalasyn (approved):
LGTM, but to make `pint merge` happy, you'll need to retarget to `6.2.0-post`.

ableegoldman:
Interesting -- @vcrfxia does this mean we do/can/should prevent users from setting the corresponding Streams configs for queries? That would be `default.deserialization.exception.handler` and `default.production.exception.handler`

vcrfxia:
> does this mean we do/can/should prevent users from setting the corresponding Streams configs for queries?

From a quick inspection of the code, it doesn't look like we prevent users from setting the corresponding Streams configs. If they set those at a per-query level, then I think they would override anything configured with the ksql configs. As for whether we should prevent users from doing that, I don't see any particular reason why we should for users running their own ksqlDB deployments. 

The reason that the ksql configs can only be set at a server-level today has to do with where they're loaded/transformed into the corresponding Streams configs. Today it happens only at server start time, so trying to apply them at a per-query level would have no effect and is therefore disallowed. We could change this in the future, though.

PR #6160, opened 2020-09-04
fix: delete zombie consumer groups 🧟

agavra:
fixes #1283

### Description 

Before this PR, we don't clean up consumer groups that are leftover which makes it really annoying for users to look into their existing consumer groups. This PR adds cleanup on close for persistent and transient queries.

The unfortunate part, is that the consumer group API takes a while to update when the last consumer has left the group (specifically, the `heartbeat.interval.ms` which is default to 3 seconds, needs to pass after the last consumer leaves the group). To account for this, I've taken the external resource cleanup out of the main execution path so that the UI doesn't hang when we terminate a query.

### Testing done 

Unit testing, and manual testing to make sure the resources are cleaned up.

### Reviewer checklist
- [ ] Ensure docs are updated if necessary. (eg. if a user visible feature is being added or changed).
- [ ] Ensure relevant issues are linked (description should include text like "Fixes #<issue number>")

PR #8383, opened 2021-11-17
feat: clean up based on query id for shared runtimes

wcarlson5:
look for shared runtimes when cleaning up and remove only the queries internal topics

### Testing done 

### Reviewer checklist

### DIFF
--- a/ksqldb-engine/src/main/java/io/confluent/ksql/engine/KsqlEngine.java
+++ b/ksqldb-engine/src/main/java/io/confluent/ksql/engine/KsqlEngine.java
@@ -594,9 +594,9 @@ public class KsqlEngine implements KsqlExecutionContext, Closeable {
         final QueryMetadata query
     ) {
       final String applicationId = query.getQueryApplicationId();
-      String queryID = "";
+      Optional<String> queryID = Optional.empty();
       if (ksqlConfig.getBoolean(KsqlConfig.KSQL_SHARED_RUNTIME_ENABLED)) {
-        queryID = query.getQueryId().toString();
+        queryID = Optional.of(query.getQueryId().toString());
       }
       if (query.hasEverBeenStarted()) {
         cleanupService.addCleanupTask(
--- a/ksqldb-engine/src/main/java/io/confluent/ksql/engine/KsqlEngine.java
+++ b/ksqldb-engine/src/main/java/io/confluent/ksql/engine/KsqlEngine.java
@@ -594,11 +594,16 @@ public class KsqlEngine implements KsqlExecutionContext, Closeable {
         final QueryMetadata query
     ) {
       final String applicationId = query.getQueryApplicationId();
+      String queryID = "";
+      if (ksqlConfig.getBoolean(KsqlConfig.KSQL_SHARED_RUNTIME_ENABLED)) {
+        queryID = query.getQueryId().toString();
+      }
       if (query.hasEverBeenStarted()) {
         cleanupService.addCleanupTask(
             new QueryCleanupService.QueryCleanupTask(
                 serviceContext,
                 applicationId,
+                queryID,
                 query instanceof TransientQueryMetadata,
                 ksqlConfig.getKsqlStreamConfigProps()
                     .getOrDefault(
--- a/ksqldb-engine/src/main/java/io/confluent/ksql/engine/KsqlEngine.java
+++ b/ksqldb-engine/src/main/java/io/confluent/ksql/engine/KsqlEngine.java
@@ -594,16 +594,16 @@ public class KsqlEngine implements KsqlExecutionContext, Closeable {
         final QueryMetadata query
     ) {
       final String applicationId = query.getQueryApplicationId();
-      Optional<String> queryID = Optional.empty();
+      Optional<String> topologyName = Optional.empty();
       if (ksqlConfig.getBoolean(KsqlConfig.KSQL_SHARED_RUNTIME_ENABLED)) {
-        queryID = Optional.of(query.getQueryId().toString());
+        topologyName = Optional.of(query.getQueryId().toString());
       }
       if (query.hasEverBeenStarted()) {
         cleanupService.addCleanupTask(
             new QueryCleanupService.QueryCleanupTask(
                 serviceContext,
                 applicationId,
-                queryID,
+                topologyName,
                 query instanceof TransientQueryMetadata,
                 ksqlConfig.getKsqlStreamConfigProps()
                     .getOrDefault(
--- a/ksqldb-engine/src/main/java/io/confluent/ksql/engine/QueryCleanupService.java
+++ b/ksqldb-engine/src/main/java/io/confluent/ksql/engine/QueryCleanupService.java
@@ -154,7 +154,7 @@ public class QueryCleanupService extends AbstractExecutionThreadService {
       } else {
         tryRun(
             () -> SchemaRegistryUtil.cleanupInternalTopicSchemas(
-                appId,
+                appId + "-" + queryId ,
                 serviceContext.getSchemaRegistryClient(),
                 isTransient),
             "internal topic schemas"
--- a/ksqldb-engine/src/main/java/io/confluent/ksql/engine/QueryCleanupService.java
+++ b/ksqldb-engine/src/main/java/io/confluent/ksql/engine/QueryCleanupService.java
@@ -23,13 +23,10 @@ import io.confluent.ksql.services.ServiceContext;
 import java.io.File;
 import java.nio.file.Path;
 import java.nio.file.Paths;
-import java.util.List;
 import java.util.Objects;
 import java.util.Set;
 import java.util.concurrent.BlockingQueue;
 import java.util.concurrent.LinkedBlockingDeque;
-import java.util.stream.Collectors;
-
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
@@ -145,7 +142,8 @@ public class QueryCleanupService extends AbstractExecutionThreadService {
             "internal topic schemas"
         );
 
-        tryRun(() -> serviceContext.getTopicClient().deleteInternalTopics(appId), "internal topics");
+        tryRun(() -> serviceContext.getTopicClient().deleteInternalTopics(appId),
+            "internal topics");
         tryRun(
             () -> serviceContext
                 .getConsumerGroupClient()
@@ -160,8 +158,8 @@ public class QueryCleanupService extends AbstractExecutionThreadService {
             "internal topic schemas"
         );
         tryRun(
-          () -> serviceContext.getTopicClient().deleteInternalTopics(appId + "-" + queryId)
-          ,"internal topics"
+            () -> serviceContext.getTopicClient().deleteInternalTopics(appId + "-" + queryId),
+            "internal topics"
         );
 
       }
--- a/ksqldb-engine/src/main/java/io/confluent/ksql/engine/QueryCleanupService.java
+++ b/ksqldb-engine/src/main/java/io/confluent/ksql/engine/QueryCleanupService.java
@@ -92,77 +92,64 @@ public class QueryCleanupService extends AbstractExecutionThreadService {
 
   public static class QueryCleanupTask implements Runnable {
     private final String appId;
-    private final String ;
+    private final String queryTopicPrefix;
+    private final Optional<String> topologyName;
+    private final String pathName;
     private final boolean isTransient;
     private final ServiceContext serviceContext;
-    private String stateDir;
 
     public QueryCleanupTask(
         final ServiceContext serviceContext,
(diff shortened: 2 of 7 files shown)


wcarlson5:
@rodesai @guozhangwang @tolgadur @ableegoldman Can I get a quick look at this?

wcarlson5:
There are going to be a few test failures until https://github.com/confluentinc/ksql/pull/8388 is merged because we don't clean up the topology directory yet

ksqldb-engine/src/main/java/io/confluent/ksql/engine/QueryCleanupService.java
```diff
@@ -111,35 +114,55 @@ public String getAppId() {
...
-      try {
-        final Path pathName = Paths.get(stateDir + "/" + appId);
-        final File directory = new File(String.valueOf(pathName.normalize()));
-        if (directory.exists()) {
-          FileUtils.deleteDirectory(directory);
-          LOG.warn("Deleted local state store for non-existing query {}. "
-                  + "This is not expected and was likely due to a "
-                  + "race condition when the query was dropped before.",
-              appId);
+      if (queryId.equals("")) {
```
ableegoldman, line 123: nit: can you define a constant for this instead of hard-coded `""` around the codebase so that it's clear what the meaning of this is? Otherwise I think it might be a bit confusing for someone who isn't familiar with the shared runtimes/NamedTopology stuff since it's not obvious why the `queryId` would ever be empty, eg
```
public static final String DEDICATED_RUNTIME_QUERY = "";
```

ksqldb-engine/src/main/java/io/confluent/ksql/engine/QueryCleanupService.java
```diff
@@ -111,35 +114,55 @@ public String getAppId() {
...
+      } else {
+        tryRun(
+            () -> SchemaRegistryUtil.cleanupInternalTopicSchemas(
+                appId + "-" + queryId ,
+                serviceContext.getSchemaRegistryClient(),
+                isTransient),
+            "internal topic schemas"
+        );
+        tryRun(
+            () -> serviceContext.getTopicClient().deleteInternalTopics(appId + "-" + queryId),
```
ableegoldman, line 161: Do we need to have separate branches like this, or could we just define a `queryTopicPrefix` String based on the `queryId` and then just plug that in to the original logic wherever it previously used `appId`? It seems like most of the logic should be the same, if not all of it

Similarly, I think we probably want to keep the part about checking if the directory still exists or not, if there's a race condition that can cause it to stick around in the current architecture then it probably exists for the shared runtime architecture as well. We should be able to do the same thing here, ie just define the `pathName` / `directory` based on whether it's a binpacked query or not, and then plug that in rather than branch.

FYI for NamedTopologies, instead of checking/deleting the whole state directory you would just need to get the NamedTopology directory, which is of the form <state_dir>/<application_id>/`__<queryId>__`

ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/util/PersistentQueryCleanupImpl.java
```diff
@@ -59,6 +59,7 @@ public void cleanupLeakedQueries(final List<PersistentQueryMetadata> persistentQ
           new QueryCleanupService.QueryCleanupTask(
           serviceContext,
           appId,
+          "",
```
ableegoldman, line 62: Not quite sure what "leaked" queries are, but should we also check this for having shared runtimes enabled and potentially pass in the actual query id? @agavra can you explain the context here a bit?

ableegoldman (commented):
> There are going to be a few test failures until #8388 is merged because we don't clean up the topology directory yet

@wcarlson5 is there any way to rewrite these tests, temporarily, so that they don't fail? If not, can you just mark them as `@Ignore` and file a quick ticket to remove this once #8388  goes through? I'd like to get this PR merged as soon as it's ready so we can close tickets and make sure the Epic reflects the actual work remaining as much as possible

ksqldb-engine/src/main/java/io/confluent/ksql/engine/QueryCleanupService.java
wcarlson5: I can use optional, empty for dedicated runtimes

ksqldb-engine/src/main/java/io/confluent/ksql/engine/QueryCleanupService.java
wcarlson5: I can probably pull that off. We don't want ot clean up the consumerGroups tho right?

wcarlson5 (commented):
@ableegoldman I don't think that I can write it so those tests pass yet. The point of them are to drop queries. What used to happen is that it would wipe out the whole runtime and then it would get recreated. It probably shouldn't have been passing in the first place. I can disable the shared runtimes for those tests though for now.

ksqldb-engine/src/main/java/io/confluent/ksql/engine/QueryCleanupService.java
ableegoldman: Whatever you prefer, as long as it's clear enough to someone who might read this code in the future without detailed knowledge of the shared runtime architecture

ksqldb-engine/src/main/java/io/confluent/ksql/engine/QueryCleanupService.java
ableegoldman: > We don't want ot clean up the consumerGroups tho right

Right

ksqldb-engine/src/main/java/io/confluent/ksql/engine/QueryCleanupService.java
```diff
@@ -120,7 +121,7 @@ public String getQueryId() {
 
     @Override
     public void run() {
-      if (queryId.equals("")) {
+      if (!queryId.isPresent()) {
```
ableegoldman, line 124: can we at least leave a comment here explaining what it means for the queryId to be empty?

wcarlson5: I think I am going to change the name from `queryId` to `topologyName` one its more accurate and two it seems to be pretty confusing.

ksqldb-engine/src/test/java/io/confluent/ksql/integration/EndToEndIntegrationTest.java
```diff
@@ -110,7 +110,7 @@
   @Parameterized.Parameters(name = "{0}")
   public static Collection<Boolean> data() {
     return Arrays.asList(
-        true, false
```
wcarlson5, line 113: I have a ticket to remember to clean this up later https://confluentinc.atlassian.net/browse/KCI-1240

ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/util/PersistentQueryCleanupImpl.java
agavra: initially I introduced the QueryCleanupService to cleanup consumer groups from terminated queries (see https://github.com/confluentinc/ksql/pull/6160) asynchronously from the main clean up. It looks like since then it was also augmented to delete directories on startup that weren't properly deleted (due to some race conditions? If I recall this can happen if streams takes too long to close). 

I'm not familiar with how the shared runtimes work - if they share a state store then it makes sense not to run the query cleanup in that scenario.

I'm not entirely sure of this part though, looks like @lct45 and @vvcephei were involved in those changes, so if you're not confident we should pull them in.

ksqldb-engine/src/main/java/io/confluent/ksql/engine/QueryCleanupService.java
```diff
@@ -91,20 +92,27 @@ public void addCleanupTask(final QueryCleanupTask task) {
...
     public QueryCleanupTask(
         final ServiceContext serviceContext,
         final String appId,
+        final Optional<String> topologyName,
         final boolean isTransient,
         final String stateDir
     ) {
       this.serviceContext = Objects.requireNonNull(serviceContext, "serviceContext");
       this.appId = Objects.requireNonNull(appId, "appId");
+      this.topologyName = Objects.requireNonNull(topologyName, "topologyName");
```
lct45, line 110: I don't know how `requireNonNull` and `Optional` play together -> will it only require non-null if one exists?

ksqldb-engine/src/main/java/io/confluent/ksql/engine/KsqlEngine.java
```diff
@@ -594,11 +594,16 @@ public void onClose(
...
       final String applicationId = query.getQueryApplicationId();
+      Optional<String> topologyName = Optional.empty();
+      if (ksqlConfig.getBoolean(KsqlConfig.KSQL_SHARED_RUNTIME_ENABLED)) {
+        topologyName = Optional.of(query.getQueryId().toString());
+      }
       if (query.hasEverBeenStarted()) {
         cleanupService.addCleanupTask(
             new QueryCleanupService.QueryCleanupTask(
                 serviceContext,
                 applicationId,
```
lct45, line 605: I'm probably just missing context here, but if shared runtimes is on, what's the application id here?

ksqldb-engine/src/main/java/io/confluent/ksql/engine/QueryCleanupService.java
```diff
@@ -91,20 +92,27 @@ public void addCleanupTask(final QueryCleanupTask task) {
...
+      queryTopicPrefix = topologyName.map(s -> appId + "-" + s).orElse(appId);
```
lct45, line 111: Would you add a comment here with some context? Just something short about why we concatenate

ksqldb-engine/src/main/java/io/confluent/ksql/engine/QueryCleanupService.java
```diff
@@ -91,20 +92,27 @@ public void addCleanupTask(final QueryCleanupTask task) {
...
       this.isTransient = isTransient;
-      this.stateDir = stateDir;
+      pathName = topologyName
```
lct45, line 113: Do we need to use `topologyName` here or can we use the `queryTopicPrefix` we created above? it looks like there's some duplication

ksqldb-engine/src/main/java/io/confluent/ksql/engine/QueryCleanupService.java
```diff
@@ -114,32 +122,35 @@ public String getAppId() {
...
 
-      tryRun(() -> serviceContext.getTopicClient().deleteInternalTopics(appId), "internal topics");
-      tryRun(
-          () -> serviceContext
-              .getConsumerGroupClient()
-              .deleteConsumerGroups(ImmutableSet.of(appId)),
-          "internal consumer groups");
+      tryRun(() -> serviceContext.getTopicClient().deleteInternalTopics(queryTopicPrefix),
+          "internal topics");
+      if (!topologyName.isPresent()) {
```
lct45, line 147: Are we not cleaning up consumer groups because there might be shared runtimes that are still using them? Is there a scenario where there's only one runtime left so we would want to clean up consumer groups?

lct45 (approved):
a few q's but overall lgtm, thanks @wcarlson5

ksqldb-engine/src/main/java/io/confluent/ksql/engine/QueryCleanupService.java
wcarlson5: it requires the optional object to not be null

ksqldb-engine/src/main/java/io/confluent/ksql/engine/KsqlEngine.java
wcarlson5: it not specific to any query

ksqldb-engine/src/main/java/io/confluent/ksql/engine/QueryCleanupService.java
wcarlson5: One is for state stores the other is to topics and other things

ksqldb-engine/src/main/java/io/confluent/ksql/engine/QueryCleanupService.java
wcarlson5: yes. That is not really a seniario, the number of runtimes should be fixed and we should not clean them when we are sharing them

ksqldb-engine/src/main/java/io/confluent/ksql/engine/QueryCleanupService.java
wcarlson5: I can add a comment

PR #8392, opened 2021-11-19
refactor: Consolidates the logic for launching queries in QueryExecutor

AlanConfluent:
### Description 
Much of the logic for starting queries was sprinkled throughout our three query endpoints.  This consolidates the logic for launching queries to a single class `QueryExecutor`.  This is one of the first steps in consolidating endpoints and simplifying our query logic.

### Testing done 
Ran unit tests functional tests

### Reviewer checklist


ksqldb-rest-app/src/test/java/io/confluent/ksql/api/perf/PullQueryRunner.java
```diff
@@ -127,11 +127,11 @@ protected void endRun() throws Exception {
         final Context context,
         final WorkerExecutor workerExecutor,
         final ApiSecurityContext apiSecurityContext,
-        final MetricsCallbackHolder metricsCallbackHolder) {
-//      PullQueryPublisher publisher = new PullQueryPublisher(context, DEFAULT_ROWS);
-//      publishers.add(publisher);
-//      return CompletableFuture.completedFuture(publisher);
-      return null;
```
vvcephei, line 134: What in the world? Any idea why it was like this in master?

ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/resources/streaming/PushQueryPublisher.java
```diff
@@ -135,30 +66,10 @@ public synchronized void subscribe(final Flow.Subscriber<Collection<StreamedRow>
...
-    // Note: we are not recording response size in this case because it is not
-    // accessible in the websocket endpoint.
-    metrics.recordConnectionDuration(startTimeNanos, sourceType, routingNodeType);
-    metrics.recordRowsReturned(metadata.getTotalRowsReturned(),
-        sourceType, routingNodeType);
-    metrics.recordRowsProcessed(metadata.getTotalRowsProcessed(),
-        sourceType, routingNodeType);
-  }
+    final WebSocketSubscriber<StreamedRow> webSocketSubscriber =
+        (WebSocketSubscriber<StreamedRow>) subscriber;
```
vvcephei, line 70: first thought... Woah, this whole method is really only used for the websocket?

second pass: oh, I guess the whole "Flow.Pub/Sub" ecosystem is only for websocket, iirc.

Maybe in another pass, we can just remove those interfaces and then hoist this type into the parameters for this method, since it seems like the Pub/Sub interfaces are actively impeding our ability to make this code more maintainable.

vvcephei (approved):
Thanks, @AlanConfluent !

I made a relatively quick pass, since I think we're going to continue iterating from here. This chunk of work looks really good. I'm happy to see all that complex logic in one place with a clear contract for each endpoint to follow.

My only concern is whether we are still covered in all the error-handling edge cases. As long as the tests are passing (especially the integration/functional tests; the unit tests that we have modified can't really be trusted to catch regressions), I think we can merge this.

AlanConfluent:
> My only concern is whether we are still covered in all the error-handling edge cases. As long as the tests are passing (especially the integration/functional tests; the unit tests that we have modified can't really be trusted to catch regressions), I think we can merge this.

I agree.  I tried testing it a fair bit with the unit tests.  It's hard to simulate a lot of these manually or even in functional tests. I think the unit tests + existing functional tests will have to do and we can go from there.  I'm happy to fix anything that comes up from this -- that's somewhat expected from large refactorings.  Obviously new tests will be written for anything that gets turned up.

PR #8399, opened 2021-11-20
chore: add raw schema to session override

lihaosky:
### Description 
* Add raw schema to schema result
* Populate raw schema to session config to be used later in Schema Injector
* Don't populate schema id to property if implicitly used

### Testing done 
* Unit test. Will add more unit test

### Reviewer checklist

### DIFF
--- a/ksqldb-engine/src/test/java/io/confluent/ksql/schema/ksql/inference/DefaultSchemaInjectorTest.java
+++ b/ksqldb-engine/src/test/java/io/confluent/ksql/schema/ksql/inference/DefaultSchemaInjectorTest.java
@@ -18,8 +18,11 @@ package io.confluent.ksql.schema.ksql.inference;
 import static io.confluent.ksql.schema.ksql.inference.TopicSchemaSupplier.SchemaAndId.schemaAndId;
 import static org.hamcrest.MatcherAssert.assertThat;
 import static org.hamcrest.Matchers.containsString;
+import static org.hamcrest.Matchers.hasKey;
 import static org.hamcrest.Matchers.is;
+import static org.hamcrest.Matchers.not;
 import static org.hamcrest.Matchers.sameInstance;
+import static org.junit.Assert.assertFalse;
 import static org.junit.Assert.assertThrows;
 import static org.mockito.ArgumentMatchers.any;
 import static org.mockito.ArgumentMatchers.eq;
@@ -30,6 +33,8 @@ import static org.mockito.Mockito.when;
 
 import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableMap;
+import io.confluent.kafka.schemaregistry.ParsedSchema;
+import io.confluent.kafka.schemaregistry.avro.AvroSchema;
 import io.confluent.ksql.config.SessionConfig;
 import io.confluent.ksql.execution.expression.tree.BooleanLiteral;
 import io.confluent.ksql.execution.expression.tree.IntegerLiteral;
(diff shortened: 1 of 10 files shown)


ksqldb-engine/src/test/java/io/confluent/ksql/schema/ksql/inference/DefaultSchemaInjectorTest.java
```diff
@@ -299,7 +328,7 @@ public void shouldInjectForCsStatement() {
             + "`mapField` MAP<STRING, BIGINT>, "
             + "`structField` STRUCT<`s0` BIGINT>, "
             + "`decimalField` DECIMAL(4, 2)) "
-            + "WITH (KAFKA_TOPIC='some-topic', KEY_FORMAT='protobuf', KEY_SCHEMA_ID=18, VALUE_FORMAT='avro', VALUE_SCHEMA_ID=5);"
+            + "WITH (KAFKA_TOPIC='some-topic', KEY_FORMAT='protobuf', VALUE_FORMAT='avro');"
```
mjsax, line 331: Not sure if I fully understand the test. If I understand correct, we test the case that the user did provide both key and value schema-id. So why are they removed? (The code above change the statement to replace the plain schema-id with the full "raw schema"...)

Seems I am missing something.

In the end, the result statement is what we write into the command topic, right? So should it not contain the user specified schema-ids? Looking in the next PR, it add `KEY_AVRO_SCHEMA_FULL_NAME` and `VALUE_AVRO_SCHEMA_FULL_NAME` but I am wondering why we don't have both -- IDs and NAMEs -- in the end?

ksqldb-functional-tests/src/test/resources/query-validation-tests/elements.json
```diff
@@ -1184,7 +1184,7 @@
       ],
       "expectedException": {
         "type": "io.confluent.ksql.util.KsqlException",
-        "message": "Tables require a PRIMARY KEY. Please define the PRIMARY KEY.\nUse a partial schema to define the primary key and still load the value columns from the Schema Registry, for example:\n\tCREATE TABLE INPUT (ID INT PRIMARY KEY) WITH (...);"
```
mjsax, line 1187: Why do we change this error message?

ksqldb-serde/src/main/java/io/confluent/ksql/serde/connect/ConnectFormat.java
```diff
@@ -54,6 +54,9 @@
  */
 public abstract class ConnectFormat implements Format {
 
+  public static String SR_KEY_SCHEMA_ID = "SR_KEY_SCHEMA_ID";
```
mjsax, line 57: Why the `SR_` prefix? Just curious? Isn't a schema-id not alway from SR?

mjsax (commented):
Just a few clarification questions for my own understanding.

ksqldb-functional-tests/src/test/resources/query-validation-tests/elements.json
lihaosky: In https://github.com/confluentinc/ksql/blob/master/ksqldb-engine/src/main/java/io/confluent/ksql/ddl/commands/CreateSourceFactory.java#L176, because we don't populate schema id unless explicitly provided user anymore.

ksqldb-serde/src/main/java/io/confluent/ksql/serde/connect/ConnectFormat.java
lihaosky: Good point. I changed this to `KEY_SCHEMA_ID` in next PR. I'll change it in this PR

ksqldb-engine/src/test/java/io/confluent/ksql/schema/ksql/inference/DefaultSchemaInjectorTest.java
mjsax: Ah. I think I mixed something up. Thanks for clarifying.

mjsax (approved):

Issue #8363, opened 2021-11-15
The AVG function throws NPE if the value is null


spena:
**Describe the bug**
When a `null` value is passed to the `AVG` function, the following exception is logged in the KSQL log:
```
Uncaught exception in query org.apache.kafka.streams.errors.StreamsException: Exception caught in process. taskId=1_4, processor=KSTREAM-SOURCE-0000000009, topic=_confluent-ksql-pksqlc-zg3m0query_CTAS_OBEM_DEV_COV_FLOOR_OCCUPANCY_FINAL_2193-Aggregate-GroupBy-repartition, partition=4, offset=5, stacktrace=java.lang.NullPointerException
	at io.confluent.ksql.function.udaf.average.AverageUdaf.lambda$averageDouble$11(AverageUdaf.java:89)
	at io.confluent.ksql.function.udaf.average.AverageUdaf$1.undo(AverageUdaf.java:144)
	at io.confluent.ksql.function.udaf.average.AverageUdaf$1.undo(AverageUdaf.java:101)
	at io.confluent.ksql.function.UdafTableAggregateFunction.undo(UdafTableAggregateFunction.java:45)
	at io.confluent.ksql.execution.function.udaf.KudafUndoAggregator.apply(KudafUndoAggregator.java:68)
	at io.confluent.ksql.execution.function.udaf.KudafUndoAggregator.apply(KudafUndoAggregator.java:27)
	at org.apache.kafka.streams.kstream.internals.KTableAggregate$KTableAggregateProcessor.process(KTableAggregate.java:92)
	at org.apache.kafka.streams.kstream.internals.KTableAggregate$KTableAggregateProcessor.process(KTableAggregate.java:59)
	at org.apache.kafka.streams.processor.internals.ProcessorAdapter.process(ProcessorAdapter.java:71)
	at org.apache.kafka.streams.processor.internals.ProcessorNode.process(ProcessorNode.java:146)
	at org.apache.kafka.streams.processor.internals.ProcessorContextImpl.forwardInternal(ProcessorContextImpl.java:253)
	at org.apache.kafka.streams.processor.internals.ProcessorContextImpl.forward(ProcessorContextImpl.java:232)
	at org.apache.kafka.streams.processor.internals.ProcessorContextImpl.forward(ProcessorContextImpl.java:191)
	at org.apache.kafka.streams.processor.internals.SourceNode.process(SourceNode.java:84)
[...]

PR #8400, opened 2021-11-20
fix: handle null input for average udaf

jzaralim:
### Description 
This PR fixes an NPE which occurs in the average udaf when removing a null value from the average.

### Testing done 
added a QTT and a unit test

### Reviewer checklist


nateab (approved):
LGTM! Thanks @jzaralim

jnh5y:
This addresses https://github.com/confluentinc/ksql/issues/8363.

Issue #7925, opened 2021-07-30
Support specifying fully-qualified record name for PROTOBUF and JSON_SR


mikebin:
**Is your feature request related to a problem? Please describe.**
ksqlDB currently supports specifying a top-level fully-qualified record name for the `AVRO` format with `VALUE_AVRO_SCHEMA_FULL_NAME`, but this isn't currently supported for `PROTOBUF` and `JSON_SR`. So you end up with a default record name in the schema like `ConnectDefault1`, which might be undesirable for application(s) which need to consume the data downstream.

**Describe the solution you'd like**
Add equivalent support for specifying a top-level fully-qualified record name with `PROTOBUF` and `JSON_SR` - perhaps rename `VALUE_AVRO_SCHEMA_FULL_NAME` to `VALUE_SCHEMA_FULL_NAME`.

**Describe alternatives you've considered**

**Additional context**
Would we need similar support for specifying fully-qualified record names for for key schemas as well, now that ksqlDB supports keys in Avro, Protobuf, and JSON Schema format?

colinhicks:
Older issue with the same property-renaming suggestion as above: https://github.com/confluentinc/ksql/issues/4656

PR #8401, opened 2021-11-21
chore: add schema id to format property

lihaosky:
### Description 
* Add schema id to format property so it will sync down to command topic and metastore for serialization use.

### Testing done 
* Unit test.

### Reviewer checklist


ksqldb-engine/src/test/java/io/confluent/ksql/schema/ksql/inference/DefaultSchemaInjectorTest.java
```diff
@@ -475,8 +510,8 @@ public void shouldGetSchemasWithoutSchemaId() {
     final ConfiguredStatement<CreateTable> result = injector.inject(ctStatement);
 
     // Then:
-    assertThat(result.getStatementText(), containsString("KEY_SCHEMA_ID=18"));
-    assertThat(result.getStatementText(), containsString("VALUE_SCHEMA_ID=5"));
+    assertThat(result.getStatementText(), not(containsString("KEY_SCHEMA_ID=18")));
+    assertThat(result.getStatementText(), not(containsString("VALUE_SCHEMA_ID=5")));
```
mjsax, line 514: Given the changes from above, should we add `containsString("KEY_AVRO_SCHEMA_FULL_NAME='myrecord'")` and similar for value?

ksqldb-engine/src/test/java/io/confluent/ksql/schema/ksql/inference/DefaultSchemaInjectorTest.java
```diff
@@ -618,8 +653,7 @@ public void shouldThrowIfKeyFormatDoesNotSupportSchemaIdInference() {
 
     // Then:
     assertThat(e.getMessage(),
-        containsString("KEY_FORMAT should support schema inference when KEY_SCHEMA_ID is provided. "
-            + "Current format is KAFKA."));
+        containsString("KAFKA does not support the following configs: [KEY_SCHEMA_ID]"));
```
mjsax, line 656: I could not find where this change originates (ie, why does this error message change)?

ksqldb-parser/src/main/java/io/confluent/ksql/parser/properties/with/CreateSourceAsProperties.java
```diff
@@ -119,17 +120,26 @@ public SerdeFeatures getValueSerdeFeatures() {
...
-      builder.put(AvroFormat.FULL_SCHEMA_NAME, AvroFormat.getKeySchemaName(name));
+      final String schemaName = props.getString(CommonCreateConfigs.KEY_AVRO_SCHEMA_FULL_NAME);
+      if (schemaName != null) {
+        // Set if schema name is explicitly set
+        builder.put(AvroFormat.FULL_SCHEMA_NAME, schemaName);
+      } else {
+        // ensure that the schema name for the key is unique to the sink - this allows
+        // users to always generate valid, non-conflicting avro record definitions in
+        // generated Java classes (https://github.com/confluentinc/ksql/issues/6465)
+        builder.put(AvroFormat.FULL_SCHEMA_NAME, AvroFormat.getKeySchemaName(name));
```
mjsax, line 131: For my own education: why don't we have a similar schema name generation for the value below? And should we really introduce `KEY_AVRO_SCHEMA_FULL_NAME` ?

The linked tickets say:
> We could introduce KEY_AVRO_SCHEMA_FULL_NAME. Quickest solution... and lame.

Not sure if I understand the full context. Maybe @agavra knows?

mjsax (commented):
A few clarification questions.

ksqldb-engine/src/test/java/io/confluent/ksql/schema/ksql/inference/DefaultSchemaInjectorTest.java
lihaosky: This test actually shouldn't `containsString("KEY_AVRO_SCHEMA_FULL_NAME='myrecord'")` since we only change the schema name when schema id is explicitly provided which is covered for example on line 438.

ksqldb-engine/src/test/java/io/confluent/ksql/schema/ksql/inference/DefaultSchemaInjectorTest.java
lihaosky: It's from:

https://github.com/confluentinc/ksql/blob/master/ksqldb-engine/src/main/java/io/confluent/ksql/schema/ksql/inference/DefaultSchemaInjector.java#L241

and then

https://github.com/confluentinc/ksql/blob/5dd40134f6ad892cf092fd29ac2b0a3f60f37868/ksqldb-serde/src/main/java/io/confluent/ksql/serde/FormatFactory.java#L50

Since we put schema_id property to `FormatInfor` property and `kafka` format doesn't support this property, the validation will fail

ksqldb-parser/src/main/java/io/confluent/ksql/parser/properties/with/CreateSourceAsProperties.java
lihaosky: > why don't we have a similar schema name generation for the value below?

Because the purpose of this is to avoid clashing with `KsqlDataSourceSchema` which is the default name for value schema name? So if no name is provided, default value name will be `KsqlDataSourceSchema` and this will make sure key schema name will (may) be different from value schema name. So there's no need to generate value schema name? 

> We could introduce KEY_AVRO_SCHEMA_FULL_NAME. Quickest solution... and lame.

Maybe the original purpose is to get around clashing names instead of introduce ways to set key name for user? Not sure  exactly... @agavra can comment

agavra: This is the suggestion in the original ticket that was implemented

> I propose we default the name of the record in the key schema to `<sourceNameInCamelCase>Key`, e.g. for a source named `BOB`, the record in the key schema would be called `BobKey`.

The reason Andy thought introducing `KEY_AVRO_SCHEMA_FULL_NAME` is "lame" is because it basically makes it a required property for anyone that's using AVRO for both keys/values (which isn't very SQLy), and is also trying to generate class names. It seems like better UX to just generate a good name (e.g. `BobKey`) from the beginning.

> For my own education: why don't we have a similar schema name generation for the value below?

@mjsax this is historical for backwards compatibility - I think it would make sense for us to generate the key/value both in the same way in the default path but we'd have to think through the implications to the users.

-----

All that being said, I'm not totally opposed to introducing `KEY_AVRO_SCHEMA_FULL_NAME` _if_ we have evidence that users aren't happy with our generated names. That gives them the control to potentially introduce also their own naming convention to the record names. But absent that user request, I'd err against introducing the added complexity.

lihaosky: > it basically makes it a required property for anyone that's using AVRO for both keys/values

I don't fully understand this. It will be optional and the key schema name will be generated as `<sourceNameInCamelCase>Key` if not provided. Why it's required?

> we have evidence that users aren't happy with our generated names. That gives them the control to potentially introduce also their own naming convention to the record names. But absent that user request, I'd err against introducing the added complexity.

The context is that we want to control the key schema name and make it same as the avro schema's name we fetch from schema registry using provided schema id. Not sure if I'm a user :) Also I thought there's an existing `VALUE_AVRO_SCHEMA_FULL_NAME` and it doesn't hurt to provide `KEY_AVRO_SCHEMA_FULL_NAME` to add the flexibility as well. If there's downside of exposing `KEY_AVRO_SCHEMA_FULL_NAME` to user and there's cleaner ways to set the key schema name for avro format, it's also ok. What would be a cleaner way to set format property internally in general?

agavra: > I don't fully understand this. It will be optional and the key schema name will be generated as <sourceNameInCamelCase>Key if not provided. Why it's required?

Sorry I wasn't totally clear, I meant that was the motivation for introducing the generated key name - now that we have that, it wouldn't be required. I just wanted to give context on the motivation.

> If there's downside of exposing KEY_AVRO_SCHEMA_FULL_NAME to user and there's cleaner ways to set the key schema name for avro format, it's also ok. 

I don't know of any other way, I don't think it's a serious downside other than just adding more bloat. I think we had imagined one day getting rid of `VALUE_AVRO_SCHEMA_FULL_NAME` and make it just the generated name, but I don't really have a strong opinion on whether that's the right thing to do or not.

tl;dr, I think it's fine to just keep your PR as is and introduce `KEY_AVRO_SCHEMA_FULL_NAME`.

ksqldb-engine/src/test/java/io/confluent/ksql/schema/ksql/inference/DefaultSchemaInjectorTest.java
mjsax: On the first PR, you said:
> `givenKeyAndValueInferenceSupported` doesn't specify schema id.

This method is use in this test, but also in `shouldInjectForCsStatement()` above, and in the above test, `KEY_AVRO_SCHEMA_FULL_NAME` and `VALUE_AVRO_SCHEMA_FULL_NAME` are both added.

What do I miss?

ksqldb-engine/src/test/java/io/confluent/ksql/schema/ksql/inference/DefaultSchemaInjectorTest.java
```diff
@@ -299,7 +328,8 @@ public void shouldInjectForCsStatement() {
             + "`mapField` MAP<STRING, BIGINT>, "
             + "`structField` STRUCT<`s0` BIGINT>, "
             + "`decimalField` DECIMAL(4, 2)) "
-            + "WITH (KAFKA_TOPIC='some-topic', KEY_FORMAT='protobuf', KEY_SCHEMA_ID=18, VALUE_FORMAT='avro', VALUE_SCHEMA_ID=5);"
+            + "WITH (KAFKA_TOPIC='some-topic', KEY_AVRO_SCHEMA_FULL_NAME='myrecord', "
+            + "KEY_FORMAT='protobuf', VALUE_AVRO_SCHEMA_FULL_NAME='myrecord', VALUE_FORMAT='avro');"
```
mjsax, line 332: Nor sure why `KEY_AVRO_SCHEMA_FULL_NAME` and `KEY_AVRO_SCHEMA_FULL_NAME` are added here. The test setup seems to use `givenKeyAndValueInferenceSupported` that does not specify a explicit schema-ids ?

ksqldb-parser/src/main/java/io/confluent/ksql/parser/properties/with/CreateSourceAsProperties.java
mjsax: > I don't fully understand this. It will be optional and the key schema name will be generated as <sourceNameInCamelCase>Key if not provided. Why it's required?

I _think_ @agavra meant it would be required if we only have the property without any name generation.

> this is historical for backwards compatibility - I think it would make sense for us to generate the key/value both in the same way in the default path but we'd have to think through the implications to the users.

Thanks for the background. I guess the confusing part is, that we have a "value" property, that we also use to generate a key-name. It would be simpler, if the property would be name `AVRO_SCHEMA_FULL_NAME` (without the `VALUE_` prefix) and use `BobValue` and `BobKey` as names (to built upon the other example). But I understand why we ended up where we are.

> The context is that we want to control the key schema name and make it same as the avro schema's name we fetch from schema registry using provided schema id. Not sure if I'm a user :) 

That a new context and "you are not a user" (seems you are a program? :trollface:). For this new context it makes sense to add `KEY_AVRO_SCHEMA_FULL_NAME` now, but it did maybe not back in the days.

So I think we are good for now. Maybe a good follow up (out of scope for this PR) would be, to change existing behavior a little bit:
- If `KEY_AVRO_SCHEMA_FULL_NAME` is specified, use as-is (only for the key)
- If `VALUE_AVRO_SCHEMA_FULL_NAME` is specified, use as-is (only for the value -- _this is a change_)
- Add `AVRO_SCHEMA_FULL_NAME` and use for key and value with `XxxKey` and `XxxValue` names to avoid conflict

The tricky question is backward compatibility...

agavra: lgtm

ksqldb-engine/src/test/java/io/confluent/ksql/schema/ksql/inference/DefaultSchemaInjectorTest.java
lihaosky: thumbup

good catch. Let me fix it.

ksqldb-parser/src/main/java/io/confluent/ksql/parser/properties/with/CreateSourceAsProperties.java
mjsax: > I don't know of any other way, I don't think it's a serious downside other than just adding more bloat. I think we had imagined one day getting rid of VALUE_AVRO_SCHEMA_FULL_NAME and make it just the generated name, but I don't really have a strong opinion on whether that's the right thing to do or not.

Even better. I guess we could try to use a combination of `cluster-id` and stream/table name as default: `MyClusterMyTableKey` and `MyClusterMyTableValue` and use `AVRO_SCHEMA_FULL_NAME` to let people overwrite it... Anyway. Not relevant for this PR.

ksqldb-engine/src/main/java/io/confluent/ksql/schema/ksql/inference/DefaultSchemaInjector.java
```diff
@@ -266,7 +267,24 @@ private static CreateSource addSchemaFields(
     final TableElements elements = buildElements(preparedStatement, keySchema, valueSchema);
 
     final CreateSource statement = preparedStatement.getStatement();
-    return statement.copyWith(elements, statement.getProperties());
+    final CreateSourceProperties properties = statement.getProperties();
+
+    Optional<String> avroKeySchemaName = Optional.empty();
```
mjsax, line 272: Nit: should we make this variable `final` and add an `else` block below to assign `Optional.empty()` ? (same for `avroValueSchemaName`)

ksqldb-engine/src/main/java/io/confluent/ksql/schema/ksql/inference/DefaultSchemaInjector.java
```diff
@@ -266,7 +267,24 @@ private static CreateSource addSchemaFields(
...
+    Optional<String> avroValueSchemaName = Optional.empty();
+
+    // Only populate avro key and value schema names when schema ids are explicitly provided
+    // and fetched schema is avro format
```
mjsax, line 276: Why only AVRO but not Protobuf/JSON_SR? (Just for my own education.)

ksqldb-engine/src/test/java/io/confluent/ksql/schema/ksql/inference/DefaultSchemaInjectorTest.java
```diff
@@ -520,7 +520,15 @@ public void shouldGetValueSchemaWithSchemaId() {
     final ConfiguredStatement<CreateStream> result = injector.inject(csStatement);
 
     // Then:
-    assertThat(result.getStatementText(), containsString("VALUE_SCHEMA_ID=42"));
+    assertThat(result.getStatementText(), is(
+        "CREATE STREAM `cs` ("
+        + "`intField` INTEGER, `bigIntField` BIGINT, "
```
mjsax, line 525: nit formatting: hard to read, can we do one line per column?

ksqldb-engine/src/test/java/io/confluent/ksql/schema/ksql/inference/DefaultSchemaInjectorTest.java
```diff
@@ -520,7 +520,15 @@ public void shouldGetValueSchemaWithSchemaId() {
...
+        + "`doubleField` DOUBLE, `stringField` STRING, `booleanField` BOOLEAN, "
+        + "`arrayField` ARRAY<INTEGER>, `mapField` MAP<STRING, BIGINT>, "
+        + "`structField` STRUCT<`s0` BIGINT>, `decimalField` DECIMAL(4, 2)) WITH "
+        + "(KAFKA_TOPIC='some-topic', KEY_FORMAT='kafka', VALUE_AVRO_SCHEMA_FULL_NAME='myrecord', "
+        + "VALUE_FORMAT='json_sr', VALUE_SCHEMA_ID=42);"));
```
mjsax, line 530: In alignment to my question above. We use `json_sr` format and add `VALUE_AVRO_SCHEMA_FULL_NAME`. Seems to contradict the comment in `DefaultSchemaInjector`?

OrL if we support `VALUE_AVRO_SCHEMA_FULL_NAME` only for AVRO, it should not be added for this test case?

ksqldb-parser/src/main/java/io/confluent/ksql/parser/properties/with/CreateSourceProperties.java
```diff
@@ -143,17 +144,25 @@ public String getKafkaTopic() {
   private Map<String, String> getKeyFormatProperties(final String keyFormat, final String name) {
```
mjsax, line 144: Just a general question: seems this method is the same as in `CreateSourceAsProperties` ? Wondering if both should share code? Not sure why both are different?

mjsax (commented):
Few follow up question.

ksqldb-engine/src/main/java/io/confluent/ksql/schema/ksql/inference/DefaultSchemaInjector.java
lihaosky: Sure.

ksqldb-engine/src/main/java/io/confluent/ksql/schema/ksql/inference/DefaultSchemaInjector.java
lihaosky: Good question. let me check whether protobuf and json_sr also require this to be backward compatible. ksqlDB doesn't populate name for protobuf and json_sr currently not sure the reason...

ksqldb-engine/src/test/java/io/confluent/ksql/schema/ksql/inference/DefaultSchemaInjectorTest.java
lihaosky: Because the returned raw schema was mocked to be `avro` and we check raw schema type to populate the name, that's where's discrepancy is from. We do have check in schema supplier that the format from property must be consistent with the raw format fetched.

ksqldb-parser/src/main/java/io/confluent/ksql/parser/properties/with/CreateSourceProperties.java
lihaosky: Good question. Looks like the they share lots of code and could be refactored to share some base class or common util code. Not sure the reason why they are this way now. Could be good follow up task to refactor?

ksqldb-engine/src/main/java/io/confluent/ksql/schema/ksql/inference/DefaultSchemaInjector.java
lihaosky: protobuf and json_sr also needs name to be same to be backward compatible. See https://github.com/confluentinc/ksql/issues/7925. Currently, ksqlDb doesn't provide a way to set it and default is `ConnectDefault1`...

ksqldb-engine/src/test/java/io/confluent/ksql/schema/ksql/inference/DefaultSchemaInjectorTest.java
mjsax: > Because the returned raw schema was mocked to be `avro`

For this case, shouldn't the expected statement not use `VALUE_FORMAT='avro'`? Where does the `json_sr` come from?

ksqldb-parser/src/main/java/io/confluent/ksql/parser/properties/with/CreateSourceProperties.java
mjsax: > Could be good follow up task to refactor?

Yeah. Maybe. I guess there might be a reason for the current "split" of both. Maybe @agavra or @vcrfxia would know?

ksqldb-engine/src/test/java/io/confluent/ksql/schema/ksql/inference/DefaultSchemaInjectorTest.java
lihaosky: `givenValueButNotKeyInferenceSupported` is hardcoded to populate `json_sr` format for value

ksqldb-parser/src/main/java/io/confluent/ksql/parser/properties/with/CreateSourceProperties.java
agavra: they're different object models, and not all the properties are the same - we could probably refactor into a base class (I think I even suggested this once) but Andy was against it at the time and was able to convince me. I don't entirely remember why if I'm being honest 🤷

ksqldb-serde/src/test/java/io/confluent/ksql/serde/FormatFactoryTest.java
```diff
@@ -135,10 +135,15 @@ public void shouldThrowWhenCreatingFromUnsupportedProperty() {
   @Test
   public void shouldNotThrowWhenCreatingFromSupportedProperty() {
     // Given:
-    final FormatInfo formatInfo = FormatInfo.of("AVRO", ImmutableMap.of("fullSchemaName", "1"));
+    final FormatInfo formatInfo = FormatInfo.of("AVRO", ImmutableMap.of("schemaId", "1", "fullSchemaName", "2"));
+    final FormatInfo protobufInfo = FormatInfo.of("PROTOBUF",
+        ImmutableMap.of("schemaId", "1"));
+    final FormatInfo jsonSRFormatInfo = FormatInfo.of("JSON_SR",
+        ImmutableMap.of("schemaId", "123"));
```
mjsax, line 142: Is "fullSchemaName" missing for protobuf and json?

ksqldb-serde/src/test/java/io/confluent/ksql/serde/FormatFactoryTest.java
```diff
@@ -135,10 +135,15 @@ public void shouldThrowWhenCreatingFromUnsupportedProperty() {
   @Test
   public void shouldNotThrowWhenCreatingFromSupportedProperty() {
     // Given:
-    final FormatInfo formatInfo = FormatInfo.of("AVRO", ImmutableMap.of("fullSchemaName", "1"));
+    final FormatInfo formatInfo = FormatInfo.of("AVRO", ImmutableMap.of("schemaId", "1", "fullSchemaName", "2"));
```
mjsax, line 138: nit: `avroFormatInfo` ?

ksqldb-serde/src/test/java/io/confluent/ksql/serde/FormatFactoryTest.java
```diff
@@ -135,10 +135,15 @@ public void shouldThrowWhenCreatingFromUnsupportedProperty() {
...
+    final FormatInfo protobufInfo = FormatInfo.of("PROTOBUF",
```
mjsax, line 139: nit: `protobufFormatInfo` ?

mjsax (approved):

ksqldb-serde/src/test/java/io/confluent/ksql/serde/FormatFactoryTest.java
lihaosky: Let me add that

PR #8406, opened 2021-11-22
fix: synchronizing writes to localcommand file

patrickstuedi:
Attempt to fix race condition when writing local command file: https://confluentinc.atlassian.net/browse/KQE-532

### DIFF
--- a/ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/LocalCommandsFile.java
+++ b/ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/LocalCommandsFile.java
@@ -59,7 +59,7 @@ public final class LocalCommandsFile  implements Closeable {
     return new LocalCommandsFile(file, true);
   }
 
-  public void write(final LocalCommand localCommand) throws IOException {
+  public synchronized void write(final LocalCommand localCommand) throws IOException {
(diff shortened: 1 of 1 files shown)


agavra (approved):
thanks!

ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/LocalCommandsFile.java
```diff
@@ -59,7 +59,7 @@ public static LocalCommandsFile createWriteable(final File file) {
     return new LocalCommandsFile(file, true);
   }
 
-  public void write(final LocalCommand localCommand) throws IOException {
+  public synchronized void write(final LocalCommand localCommand) throws IOException {
```
AlanConfluent, line 62: Do we want to do this on close as well, to ensure that we don't close it partway through writing a record?

AlanConfluent (approved):
Nice!  Looks good

ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/LocalCommandsFile.java
patrickstuedi: Good point.

PR #8410, opened 2021-11-23
fix: Fixes flaky ScalablePushBandwidthThrottleIntegrationTest caused by timeout

AlanConfluent:
### Description 
Because `ScalablePushBandwidthThrottleIntegrationTest` creates a server per test, the startup time is attributed to the test, so although there's a 2 minute limit, most of it can be taken up by this startup.  When the test then runs, the server gets terminated part way through and you get a funny connection closed error rather than a timeout, leading to confusion.

This ups the timeout of this test to prevent this.

### Testing done 
Ran the test and simulated slow startup and was able to reproduce flaky stacktraces.

### Reviewer checklist


agavra (approved):
that's a super long time to start, is there any way we can make it faster?

nateab (approved):
LGTM!

AlanConfluent:
> that's a super long time to start, is there any way we can make it faster?

@agavra I agree.  This is not ideal, but the test itself takes maybe 30 seconds at times to run which means that if the startup takes way too long, it can hit 2 minutes in really slow CI runs.  I think the ideal solution would include speeding up the server startup time.  That would be a good followup.

PR #8411, opened 2021-11-23
chore: schema id serde

lihaosky:
1. Use schema id to do serialization if presented in metastore. 
2. Set schema name to Logical schema which is used to build value (e.g. `Struct`).


ksqldb-serde/src/main/java/io/confluent/ksql/serde/avro/AvroFormat.java
```diff
@@ -81,9 +82,12 @@ protected ConnectSchemaTranslator getConnectSchemaTranslator(
     // Ensure schema is compatible by converting to Avro schema:
     getConnectSchemaTranslator(formatProps).fromConnectSchema(connectSchema);
 
-    final String schemaFullName = new AvroProperties(formatProps).getFullSchemaName();
+    final AvroProperties properties = new AvroProperties(formatProps);
```
mjsax, line 85: For JSON_SR and PB we don't have a corresponding `XxxProperties` helper class, and do the schema-id parsing "manually". Would it be worth to introduce such helpers (also to unify the code pattern)?

lihaosky: Yeah. That's the plan

lihaosky:
Test failing because depends on https://github.com/confluentinc/ksql/pull/8421. Will merge https://github.com/confluentinc/ksql/pull/8421 first before fixing this.

ksqldb-serde/src/main/java/io/confluent/ksql/serde/connect/ConnectSchemas.java
```diff
@@ -34,7 +34,8 @@ private ConnectSchemas() {
    * @param columns the list of columns.
    * @return the Struct schema.
    */
-  public static ConnectSchema columnsToConnectSchema(final List<? extends SimpleColumn> columns) {
+  public static ConnectSchema columnsToConnectSchema(final List<? extends SimpleColumn> columns,
+      final String name) {
```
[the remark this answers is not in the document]
lihaosky: It's schemaName which can be set

ksqldb-serde/src/test/java/io/confluent/ksql/serde/connect/ConnectSchemasTest.java
```diff
@@ -24,7 +24,7 @@ public void shouldConvertColumnsToStructSchema() {
         .build();
 
     // When:
-    final ConnectSchema result = ConnectSchemas.columnsToConnectSchema(schema.value());
+    final ConnectSchema result = ConnectSchemas.columnsToConnectSchema(schema.value(), null);
```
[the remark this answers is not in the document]
lihaosky: Optional kind of feel if it's empty, don't set it. Make it string and pass null will set null to schema name inside this method. That might be the main reason. But I'm ok either way.

ksqldb-functional-tests/src/test/resources/query-validation-tests/elements.json
```diff
@@ -76,7 +76,7 @@
       ],
       "expectedException": {
         "type": "io.confluent.ksql.util.KsqlException",
-        "message": "KEY_FORMAT should support schema inference when KEY_SCHEMA_ID is provided. Current format is KAFKA."
+        "message": "KAFKA does not support the following configs: [schemaId]"
```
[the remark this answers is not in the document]
lihaosky: Because `FormatFactory.of` was called and throw early. I agree what you said, let me catch it and rethrow.

ksqldb-serde/src/main/java/io/confluent/ksql/serde/json/KsqlJsonSerdeFactory.java
```diff
@@ -40,13 +43,19 @@
...
-   * @param useSchemaRegistryFormat whether or not to require the magic byte and
-   *                                schemaID as part of the JSON message
+   * @param properties JsonSchemaFormat properties
    */
-  KsqlJsonSerdeFactory(final boolean useSchemaRegistryFormat) {
-    this.useSchemaRegistryFormat = useSchemaRegistryFormat;
+  KsqlJsonSerdeFactory(final JsonSchemaProperties properties) {
+    this.useSchemaRegistryFormat = true;
+    this.properties = Objects.requireNonNull(properties, "properties");
   }
```
spena, line 59: I have a suggestion for better readability of this class. What do you think about keeping the old constructor as private and create these two static methods:
```
// returns a KsqlJsonSerdeFactory with SR format enabled
final KsqlJsonSerdeFactory factory = KsqlJsonSerdeFactory.withSchemaRegistryProperties(properties);

// returns a instance of KsqlJsonSerdeFactory without SR disabled
final KsqlJsonSerdeFactory factory = KsqlJsonSerdeFactory.get();
```

ksqldb-engine/src/main/java/io/confluent/ksql/schema/ksql/inference/DefaultSchemaInjector.java
```diff
@@ -213,28 +214,36 @@ private static boolean shouldInferSchema(
...
+    final String formatPropMsg = String.format("%s should support schema inference when %s is "
+        + "provided. Current format is %s.", formatProp, schemaIdName, formatInfo.getFormat());
+
+    final Format format;
+    try {
+      format = FormatFactory.of(formatInfo);
+    } catch (KsqlException e) {
+      if (e.getMessage().contains("does not support the following configs: [schemaId]")) {
+        throw new KsqlException(formatPropMsg);
+      }
```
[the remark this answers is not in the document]
lihaosky: Because before https://github.com/confluentinc/ksql/pull/8401 is merged, we don't put `schemaId` to format properties and `FormatFactory.of()` doesn't throw. 

https://github.com/lihaosky/ksql/blob/305ae88b0ffd2c34be8e1f0f0f18dc80f5f3d40a/ksqldb-parser/src/main/java/io/confluent/ksql/parser/properties/with/CreateSourceProperties.java#L190-L193

> A typo in the future this condition to throw a different exception

We have unit test covering this case if it helps. Otherwise, it's hard to check `FormatFactory.of` throws because of schema id problem... Unless we create different exception type?

ksqldb-serde/src/main/java/io/confluent/ksql/serde/json/KsqlJsonSerdeFactory.java
lihaosky: Sounds good to me!

ksqldb-serde/src/main/java/io/confluent/ksql/serde/connect/ConnectDataTranslator.java
```diff
@@ -41,8 +43,15 @@
...
 
+  private final Optional<Schema> phyiscalSchema;
+
   public ConnectDataTranslator(final Schema schema) {
+    this(schema, null);
+  }
+
+  public ConnectDataTranslator(final Schema schema, final Schema physicalSchema) {
     this.schema = Objects.requireNonNull(schema, "schema");
+    this.phyiscalSchema = Optional.ofNullable(physicalSchema);
```
spena, line 54: If null is accepted and the meaning is just that the parameter is optional, then let's pass `Optional` directly. It is a common practice we have followed in the ksql code.

ksqldb-serde/src/main/java/io/confluent/ksql/serde/connect/ConnectDataTranslator.java
```diff
@@ -41,8 +43,15 @@
 
   private final Schema schema;
 
+  private final Optional<Schema> phyiscalSchema;
+
   public ConnectDataTranslator(final Schema schema) {
+    this(schema, null);
+  }
+
+  public ConnectDataTranslator(final Schema schema, final Schema physicalSchema) {
```
spena, line 52: Seems that only one schema is used. Why do we need to keep the two schemas in this class? I don't see other code that requires the original even if the physical is present. For instance, `ProtobufSerdeFactory` creates a different `ConnectDataTranslator` depending of the type of schema. Can' we just keep one schema? or if the reason is that you want to know whether this schema is physical or not, then should I different class that extends from `ConnectDataTransaltor` be sufficient?

lihaosky: Good point. Let me refactor and create a separate class extending `ConnectDataTranslator`

spena (approved):
Thanks, looks good +1

jaceklaskowski:
It looks like the change breaks the build on openjdk version "11.0.13" 2021-10-19 due to the following in [ksqldb-serde/src/test/java/io/confluent/ksql/serde/avro/KsqlAvroSerializerTest.java](https://github.com/confluentinc/ksql/blob/8a11ac218923ab7617ec7853385c52034a4662e4/ksqldb-serde/src/test/java/io/confluent/ksql/serde/avro/KsqlAvroSerializerTest.java#L42):

> import com.sun.rmi.rmid.ExecOptionPermission

lihaosky:
https://github.com/confluentinc/ksql/pull/8451 to fix

PR #8415, opened 2021-11-25
fix: tolerate out-of-order execution on HARouting tests

patrickstuedi:
Now that partition fetch operations are executed in parallel the number of fetch operations that are being executed until the coordinator aborts is no longer deterministic. This PR relaxes the assumption with regard to the number of partition fetch calls being made in a flaky test.


jzaralim (approved):
thanks for the quick fix @patrickstuedi

PR #8417, opened 2021-11-27
feat: block inserting into sources with headers

jzaralim:
### Description 
Makes the following statements fail:
* `INSERT INTO ... SELECT ...` if the target source has header columns
* `INSERT VALUES ...` if one of the target columns is a header column

### Testing done 
Manually tested and added unit tests. Integration tests will be in another PR.

### Reviewer checklist

### DIFF
--- a/ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/computation/DistributingExecutor.java
+++ b/ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/computation/DistributingExecutor.java
@@ -282,7 +282,8 @@ public class DistributingExecutor {
     }
 
     if (!dataSource.getSchema().headers().isEmpty()) {
-      throw new KsqlException("Cannot insert into a source with header columns");
+      throw new KsqlException("Cannot insert into " + insertInto.getTarget().text()
+          + " because it has header columns");
     }
   }
 }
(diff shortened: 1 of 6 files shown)


ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/execution/InsertValuesExecutor.java
```diff
@@ -144,6 +146,10 @@ public void execute(
 
     final DataSource dataSource = getDataSource(config, metaStore, insertValues);
 
+    if (!verifyInsertValuesAllowed(insertValues.getColumns(), dataSource)) {
+      throw new KsqlException("Cannot insert into a HEADER column");
```
spena, line 150: It'd be good to display the column header name. Perhaps throwing the exception from inside the method would be helpful so you can customize the message. What about renaming the method too `validateInsert()` or `throwIfInsertOnHeadersColumns()`?

ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/computation/DistributingExecutor.java
```diff
@@ -280,5 +280,9 @@ private void throwIfInsertOnReadOnlyTopic(
       throw new KsqlException("Cannot insert into read-only topic: "
           + dataSource.getKafkaTopicName());
     }
+
+    if (!dataSource.getSchema().headers().isEmpty()) {
+      throw new KsqlException("Cannot insert into a source with header columns");
```
spena, line 285: Instead of `source`, use the name of the source. The `dataSource.getDataSourceType()` returns the type.

ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/execution/InsertValuesExecutor.java
jzaralim: Updated the messages

spena (approved):

PR #8418, opened 2021-11-28
chore: add key/value schema name prop and refactor

lihaosky:
### Description 
* Add `KEY_SCHEMA_FULL_NAME` and `VALUE_SCHEMA_FULL_NAME` to set schema name explicitly for avro, protobuf and json_sr format
* Add properties for protobuf, json, json_sr and introduce base class `ConnectProperties`
* Refactor `key_schema_id`, `value_schema_id` in `ConnectFormat` to be just `schema_id` in `ConnectProperties` since it exists separately in key and value format properties so using same name is fine.

### Testing done 
Unit test

### Reviewer checklist

### DIFF
--- a/ksqldb-engine/src/main/java/io/confluent/ksql/ddl/commands/CreateSourceFactory.java
+++ b/ksqldb-engine/src/main/java/io/confluent/ksql/ddl/commands/CreateSourceFactory.java
@@ -229,7 +229,12 @@ public final class CreateSourceFactory {
       final KsqlConfig ksqlConfig
   ) {
     final FormatInfo keyFormat = SourcePropertiesUtil.getKeyFormat(props, name);
+    // Validate format name and property
+    FormatFactory.of(keyFormat);
+
     final FormatInfo valueFormat = SourcePropertiesUtil.getValueFormat(props);
+    // Validate format name and property
+    FormatFactory.of(valueFormat);
 
     final SerdeFeatures keyFeatures = keySerdeFeaturesSupplier.build(
         schema,
--- a/ksqldb-engine/src/main/java/io/confluent/ksql/ddl/commands/CreateSourceFactory.java
+++ b/ksqldb-engine/src/main/java/io/confluent/ksql/ddl/commands/CreateSourceFactory.java
@@ -229,12 +229,7 @@ public final class CreateSourceFactory {
       final KsqlConfig ksqlConfig
   ) {
     final FormatInfo keyFormat = SourcePropertiesUtil.getKeyFormat(props, name);
-    // Validate format name and property
-    FormatFactory.of(keyFormat);
-
     final FormatInfo valueFormat = SourcePropertiesUtil.getValueFormat(props);
-    // Validate format name and property
-    FormatFactory.of(valueFormat);
 
     final SerdeFeatures keyFeatures = keySerdeFeaturesSupplier.build(
         schema,
--- a/ksqldb-parser/src/test/java/io/confluent/ksql/parser/properties/with/CreateSourceAsPropertiesTest.java
+++ b/ksqldb-parser/src/test/java/io/confluent/ksql/parser/properties/with/CreateSourceAsPropertiesTest.java
@@ -170,11 +170,12 @@ public class CreateSourceAsPropertiesTest {
             .build());
 
     // Then:
-    assertThat(properties.getValueFormatProperties().get(ConnectProperties.FULL_SCHEMA_NAME), is("schema"));
+    assertThat(properties.getValueFormatProperties(),
+        hasEntry(ConnectProperties.FULL_SCHEMA_NAME, "schema"));
   }
 
   @Test
-  public void shouldSetFullKeySchemaName() {
+  public void shouldSetKeyFullSchemaName() {
     // Given:
     final CreateSourceAsProperties props = CreateSourceAsProperties
         .from(ImmutableMap.<String, Literal>builder()
@@ -182,7 +183,7 @@ public class CreateSourceAsPropertiesTest {
             .put(KEY_SCHEMA_FULL_NAME, new StringLiteral("KeySchema"))
             .build());
 
-    // When / Then:
+    // Then:
     assertThat(props.getKeyFormatProperties("json_sr", "foo"),
         hasEntry(ConnectProperties.FULL_SCHEMA_NAME, "KeySchema"));
   }
--- a/ksqldb-parser/src/test/java/io/confluent/ksql/parser/properties/with/CreateSourceAsPropertiesTest.java
+++ b/ksqldb-parser/src/test/java/io/confluent/ksql/parser/properties/with/CreateSourceAsPropertiesTest.java
@@ -19,10 +19,14 @@ import static com.google.common.collect.ImmutableMap.of;
 import static io.confluent.ksql.parser.properties.with.CreateSourceAsProperties.from;
 import static io.confluent.ksql.properties.with.CommonCreateConfigs.FORMAT_PROPERTY;
 import static io.confluent.ksql.properties.with.CommonCreateConfigs.KEY_FORMAT_PROPERTY;
+import static io.confluent.ksql.properties.with.CommonCreateConfigs.KEY_SCHEMA_FULL_NAME;
 import static io.confluent.ksql.properties.with.CommonCreateConfigs.TIMESTAMP_FORMAT_PROPERTY;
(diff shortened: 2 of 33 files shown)


ksqldb-parser/src/test/java/io/confluent/ksql/parser/properties/with/CreateSourceAsPropertiesTest.java
```diff
@@ -149,10 +154,37 @@ public void shouldNotSetAvroSchemaNameForNonAvroKey() {
...
+            .put(CommonCreateConfigs.VALUE_FORMAT_PROPERTY, new StringLiteral("Protobuf"))
+            .put(CommonCreateConfigs.VALUE_SCHEMA_FULL_NAME, new StringLiteral("schema"))
+            .build());
+
+    // Then:
+    assertThat(properties.getValueFormatProperties().get(ConnectProperties.FULL_SCHEMA_NAME), is("schema"));
+  }
+
+  @Test
+  public void shouldSetFullKeySchemaName() {
```
mjsax, line 177: nit: `shouldSetKeyFullSchemaName()`

ksqldb-parser/src/test/java/io/confluent/ksql/parser/properties/with/CreateSourceAsPropertiesTest.java
```diff
@@ -149,10 +154,37 @@ public void shouldNotSetAvroSchemaNameForNonAvroKey() {
...
+    // Given:
+    final CreateSourceAsProperties props = CreateSourceAsProperties
+        .from(ImmutableMap.<String, Literal>builder()
+            .put(FORMAT_PROPERTY, new StringLiteral("Json_sr"))
+            .put(KEY_SCHEMA_FULL_NAME, new StringLiteral("KeySchema"))
+            .build());
+
+    // When / Then:
```
mjsax, line 185: nit: Why `When / Then`  and not just `Then` ?

ksqldb-parser/src/test/java/io/confluent/ksql/parser/properties/with/CreateSourceAsPropertiesTest.java
```diff
@@ -149,10 +154,37 @@ public void shouldNotSetAvroSchemaNameForNonAvroKey() {
...
+    assertThat(props.getKeyFormatProperties("json_sr", "foo"),
+        hasEntry(ConnectProperties.FULL_SCHEMA_NAME, "KeySchema"));
```
mjsax, line 187: nit: why `hasEntry` compared to `...get(...), is(...)` as above?

ksqldb-serde/src/main/java/io/confluent/ksql/serde/json/JsonFormat.java
```diff
@@ -55,6 +55,11 @@ protected ConnectSchemaTranslator getConnectSchemaTranslator(
     throw new UnsupportedOperationException(name() + " does not implement Schema Registry support");
   }
 
+  @Override
+  public Set<String> getSupportedProperties() {
+    return JsonProperties.SUPPORTED_PROPERTIES;
```
mjsax, line 60: Why this change? I though JSON won't support this, only JSON_SR ?

Might be easier to just `return ImmutableSet.of();` ?

ksqldb-serde/src/main/java/io/confluent/ksql/serde/json/JsonProperties.java
```diff
@@ -0,0 +1,46 @@
...
+package io.confluent.ksql.serde.json;
+
+import com.google.common.collect.ImmutableSet;
+import edu.umd.cs.findbugs.annotations.SuppressFBWarnings;
+import io.confluent.ksql.serde.connect.ConnectProperties;
+import java.util.Map;
+
+public class JsonProperties extends ConnectProperties {
+
+  static final ImmutableSet<String> SUPPORTED_PROPERTIES = ImmutableSet.of();
```
mjsax, line 25: Why do we need to introduce this class if the set of supported properties is empty?

ksqldb-serde/src/test/java/io/confluent/ksql/serde/json/JsonPropertiesTest.java
```diff
@@ -0,0 +1,80 @@
...
+    // When:
+    final Exception e = assertThrows(UnsupportedOperationException.class,
+        properties::getFullSchemaName);
+
+    // Then:
+    assertThat(e.getMessage(), is("JSON does not implement Schema Registry support"));
+  }
+
+  @Test
+  public void shouldGetDefaultFullSchemaName() {
```
mjsax, line 58: Seem this test is a duplicate of the previous one?

ksqldb-serde/src/main/java/io/confluent/ksql/serde/json/JsonSchemaProperties.java
```diff
@@ -0,0 +1,43 @@
...
+ * specific language governing permissions and limitations under the License.
+ */
+
+package io.confluent.ksql.serde.json;
+
+import com.google.common.collect.ImmutableSet;
+import edu.umd.cs.findbugs.annotations.SuppressFBWarnings;
+import java.util.Map;
+
+public class JsonSchemaProperties extends JsonProperties {
```
mjsax, line 22: Should we directly `extend ConnectProperties` ? Not sure if having `JsonProperties` at all make sense?

ksqldb-engine/src/main/java/io/confluent/ksql/ddl/commands/CreateSourceFactory.java
```diff
@@ -229,7 +229,12 @@ private Formats buildFormats(
       final KsqlConfig ksqlConfig
   ) {
     final FormatInfo keyFormat = SourcePropertiesUtil.getKeyFormat(props, name);
+    // Validate format name and property
+    FormatFactory.of(keyFormat);
```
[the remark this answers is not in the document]
lihaosky: Not sure exactly. Might be validated previously in other places like `DefaultSchemaInjector` and `SchemaRegisterInjector` implicitly. I don't fully recall why I added this, might be because I wrote some unit test for other PR previously and found incompatible properties can be set... 

Should be good to have a central place to do all validations explicitly so the preconditions are clear. 

I'm ok to remove this given this could be checked in other code path.

ksqldb-parser/src/test/java/io/confluent/ksql/parser/properties/with/CreateSourceAsPropertiesTest.java
lihaosky: Good catch. Don't recall this. Will change previous one to `hasEntry`

ksqldb-serde/src/main/java/io/confluent/ksql/serde/json/JsonFormat.java
lihaosky: Because I introduced `JsonProperties`, so make it consistent with other format properties class to have `SUPPORTED_PROPERTIES` field. So we can modify `SUPPORTED_PROPERTIES` when new properties are added later.

ksqldb-serde/src/main/java/io/confluent/ksql/serde/json/JsonProperties.java
lihaosky: 1. Json extended `ConnectFormat`, so I thought it might make sense to create `JsonProperties` extends `ConnectProperties`?
2. Both Json and JsonSchema format call https://github.com/confluentinc/ksql/blob/master/ksqldb-serde/src/main/java/io/confluent/ksql/serde/json/KsqlJsonSerdeFactory.java#L40 to create serializer atm. I would still like to maintain the behavior and pass a property object of Json or JsonSchema to that class later. That's why I create `JsonProperties` and let `JsonSchemaProperties` extend it so that I can pass both `JsonProperties` and `JsonSchemaProperties` to `KsqlJsonSerdeFactory`

That being said, I guess 2 is a stronger reason but there should be other ways to solve it as well.

ksqldb-serde/src/main/java/io/confluent/ksql/serde/json/JsonSchemaProperties.java
lihaosky: See comment above, I think extends `ConnectProperties` is fine.

mjsax (approved):

PR #8420, opened 2021-11-29
docs: Update packaging build number for artifacts

colinhicks:
### Description 
When we concatenate the packaging URL for Confluent Platform dependencies in the documentation (e.g. the sample Java client pom), both the artifact version and the build number must be accurate. This updates the build number for the 0.22.0 release.


JimGalasyn (approved):
LGTM, I'll cherry-pick to master.

PR #8421, opened 2021-11-29
chore: register raw parsed schema fetched from schema id if exist

lihaosky:
### Description 
Register raw schema fetched using schema id

### Testing done 
* Unit test.

### Reviewer checklist

### DIFF
--- a/ksqldb-engine/src/main/java/io/confluent/ksql/schema/ksql/inference/SchemaRegisterInjector.java
+++ b/ksqldb-engine/src/main/java/io/confluent/ksql/schema/ksql/inference/SchemaRegisterInjector.java
@@ -19,12 +19,7 @@ import static io.confluent.ksql.util.KsqlConstants.getSRSubject;
 
 import io.confluent.kafka.schemaregistry.ParsedSchema;
 import io.confluent.kafka.schemaregistry.client.SchemaRegistryClient;
-import io.confluent.kafka.schemaregistry.client.rest.exceptions.RestClientException;
-import io.confluent.kafka.schemaregistry.protobuf.ProtobufSchema;
-import io.confluent.kafka.serializers.protobuf.AbstractKafkaProtobufSerializer;
-import io.confluent.kafka.serializers.subject.DefaultReferenceSubjectNameStrategy;
 import io.confluent.ksql.KsqlExecutionContext;
-import io.confluent.ksql.exception.KsqlSchemaAuthorizationException;
 import io.confluent.ksql.execution.ddl.commands.CreateSourceCommand;
 import io.confluent.ksql.parser.properties.with.SourcePropertiesUtil;
 import io.confluent.ksql.parser.tree.CreateAsSelect;
@@ -48,13 +43,12 @@ import io.confluent.ksql.services.ServiceContext;
 import io.confluent.ksql.statement.ConfiguredStatement;
 import io.confluent.ksql.statement.Injector;
 import io.confluent.ksql.util.KsqlConfig;
+import io.confluent.ksql.util.KsqlException;
 import io.confluent.ksql.util.KsqlSchemaRegistryNotConfiguredException;
 import io.confluent.ksql.util.KsqlStatementException;
 import io.confluent.ksql.util.Pair;
-import java.io.IOException;
 import java.util.List;
 import java.util.Objects;
-import org.apache.kafka.common.acl.AclOperation;
 
 public class SchemaRegisterInjector implements Injector {
 
@@ -167,7 +161,10 @@ public class SchemaRegisterInjector implements Injector {
       final String statementText,
       final boolean registerIfSchemaExists
   ) {
-    if (kvRawSchema.left == null) {
+    final boolean registerRawKey = kvRawSchema.left != null;
+    final boolean registerRawValue = kvRawSchema.right != null;
+
+    if (!registerRawKey) {
       registerSchema(
           schema.key(),
           kafkaTopic,
@@ -179,18 +176,8 @@ public class SchemaRegisterInjector implements Injector {
           getSRSubject(kafkaTopic, true),
           true
       );
-    } else {
-      registerRawSchema(
-          kvRawSchema.left,
-          kafkaTopic,
-          keyFormat,
-          config,
-          statementText,
-          getSRSubject(kafkaTopic, true),
-          true);
     }
-
-    if (kvRawSchema.right == null) {
+    if (!registerRawValue) {
       registerSchema(
           schema.value(),
           kafkaTopic,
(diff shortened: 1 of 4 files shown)


ksqldb-engine/src/main/java/io/confluent/ksql/schema/ksql/inference/SchemaRegisterInjector.java
```diff
@@ -102,8 +105,15 @@ private void registerForCreateSource(final ConfiguredStatement<? extends CreateS
         cs.getSessionConfig().getConfig(false)
     );
 
+    final SchemaAndId rawKeySchema = (SchemaAndId) cs.getSessionConfig().getOverrides()
+        .get(CommonCreateConfigs.KEY_SCHEMA_ID);
```
spena, line 109: I saw a different code in `DefaultSchemaInjector.java` where `KEY_SCHEMA_ID` and `VALUE_SCHEMA_ID` are strings.
```
final String schemaIdName =
            isKey ? CommonCreateConfigs.KEY_SCHEMA_ID : CommonCreateConfigs.VALUE_SCHEMA_ID;
```

Why is the casting to `SchemaAndId`? Doesn't it fail?

ksqldb-engine/src/main/java/io/confluent/ksql/schema/ksql/inference/SchemaRegisterInjector.java
```diff
@@ -155,80 +167,121 @@ private void registerSchemas(
...
       final FormatInfo formatInfo,
-      final SerdeFeatures serdeFeatures,
       final KsqlConfig config,
       final String statementText,
-      final boolean registerIfSchemaExists,
       final String subject,
-      final boolean isKey
+      final Boolean isKey
   ) {
+    final String schemaIdPropStr = isKey ? "key_schema_id" : "value_schema_id";
```
spena, line 226: Aren't these two strings part of the `CommonCreateConfigs` class?

ksqldb-engine/src/main/java/io/confluent/ksql/schema/ksql/inference/SchemaRegisterInjector.java
```diff
@@ -155,80 +167,121 @@ private void registerSchemas(
...
-        final ParsedSchema parsedSchema = translator.toParsedSchema(
-            PersistenceSchema.from(schema, serdeFeatures)
+  private static int registerRawSchema(
+      final SchemaRegistryClient srClient,
+      final ParsedSchema parsedSchema,
+      final String topic,
+      final String subject,
+      final String statementText,
+      final boolean isKey
+  ) {
```
spena, line 268: This method has the potential to be moved to `SchemaRegistryUtil`. I see different methods there that deal with authentication errors (like this method). What do you think? You can write unit tests for that method in `SchemaRegistryUtilTest`.

If you move it, the only parameter I think we should not move is the `statementText`, which is only used to customize the error message. We should catch the exception when calling the method instead, and re-throw it with the statement text in this class.

ksqldb-engine/src/main/java/io/confluent/ksql/schema/ksql/inference/SchemaRegisterInjector.java
```diff
@@ -155,80 +167,121 @@ private void registerSchemas(
...
   ) {
+    final String schemaIdPropStr = isKey ? "key_schema_id" : "value_schema_id";
     final Format format = FormatFactory.of(formatInfo);
-    if (!format.supportsFeature(SerdeFeature.SCHEMA_INFERENCE)) {
-      return;
+    if (!canRegister(format, config, topic)) {
+      throw new KsqlStatementException(schemaIdPropStr + " is provided but format "
+          + format.name() + " doesn't support registering in Schema Registry",
+          statementText);
     }
```
spena, line 232: Shouldn't this check be called earlier in the `registerSchemas` method to avoid registering schemas without completing the create execution?

For instance, say the key_schema_id is registered, but then the value_schema_id fails because the format does not support schema inference. The CREATE will fail leaving a schema registered in SR, right?

I think you should check for schema inference support and other validations earlier before attempting to register all the schemas.

ksqldb-engine/src/main/java/io/confluent/ksql/schema/ksql/inference/SchemaRegisterInjector.java
lihaosky: This is casting value to `SchemaAndId` type: `.get(CommonCreateConfigs.KEY_SCHEMA_ID);`

ksqldb-engine/src/main/java/io/confluent/ksql/schema/ksql/inference/SchemaRegisterInjector.java
lihaosky: Make sense. For original code path, it doesn't throw, but for raw schema exist case, it will throw. I can check before registering for raw schema exist case

ksqldb-engine/src/main/java/io/confluent/ksql/schema/ksql/inference/SchemaRegisterInjector.java
lihaosky: Yeah. Let me take a look and put it there

spena (approved):

ksqldb-engine/src/main/java/io/confluent/ksql/schema/ksql/inference/SchemaRegisterInjector.java
lihaosky: `SchemaAndId` was put to `SessionConfig.overrides` here: https://github.com/confluentinc/ksql/blob/master/ksqldb-engine/src/main/java/io/confluent/ksql/schema/ksql/inference/DefaultSchemaInjector.java#L120-L128

It's not put to `CreateSourceProperties`

ksqldb-engine/src/main/java/io/confluent/ksql/schema/ksql/inference/SchemaRegisterInjector.java
```diff
@@ -155,45 +161,130 @@ private void registerSchemas(
...
+    final int id;
+    try {
+      id = SchemaRegistryUtil.registerSchema(serviceContext.getSchemaRegistryClient(),
+          schemaAndId.rawSchema, topic, subject, isKey);
+    } catch (KsqlException e) {
+      throw new KsqlStatementException("Could not register schema for topic: " + e.getMessage(),
+          statementText, e);
+    }
+
+    if (id != schemaAndId.id) {
```
lihaosky, line 264: One thing I'm about update here is to check if `ServiceContext` is sandboxservicecontext and make it pass validation because sand box sr client always return `123` which will fail this. Do you think it's a good way to bypass this?

spena: I'm now wondering, why do we have to register a schema that should have already be registered? If a schema_id, then we just need to fetch the scheme and use it, right?

I was going to say, what if you do this check inside the `SchemaRegistryUtil.registerSchema`, so it does not interfer with the sandbox, but I got wonder about the above question.

spena (approved):

ksqldb-engine/src/main/java/io/confluent/ksql/schema/ksql/inference/SchemaRegisterInjector.java
lihaosky: Discussed offline. Fetched raw schema can be in any subject. Reregister here under the correct subject to have proper compatibility check in SR. Also during serialization time using schema id, serializer will look up the schema under correct subject.

PR #8423, opened 2021-11-29
fix: make to_bytes work with lowercase hex

### LINKED ISSUE #8302: Function TO_BYTES does not accept HEX values in lowercase
**Describe the bug**
The `to_bytes` function returns null when passing a string with lowercase hexadecimal values.

```
sql> select to_bytes('7b0a', 'hex'), to_bytes('7B0A', 'hex') from b emit changes;
+--------------------------------------------------------------------+--------------------------------------------------------------------+
|KSQL_COL_0                                                          |KSQL_COL_1                                                          |
+--------------------------------------------------------------------+--------------------------------------------------------------------+
|null                                                                |ewo=                                                                |
^CQuery terminated
```

**Expected behavior**
`to_bytes` should allow hex values in lower/upper case.

**Additional context**
The problem is in the `BytesUtils` class which uses the following function to code values:
```
private static byte[] hexDecoding(final String value) {
    return BaseEncoding.base16().decode(value);
  }
```

I debugged the code and found the following exception thrown:
```
com.google.common.io.BaseEncoding$DecodingException: Unrecognized character: b
```

`BaseEncoding` only supports A-B and 0-9 values (all uppercase). We should find another library that accepts lower cases. 

An idea is to convert `value` to upper case, but this is adding extra overhead to the hexDecoding() function. Say a user wants to decode a long long string (first parse would be to upper case, second parse would be to decode).

jzaralim:
### Description 
Fixes #8302

Replaces Google's BaseEncoding class with `javax.xml.bind.DatatypeConverter ` for Hex conversions.

### Testing done 
Added unit and QTT tests

### Reviewer checklist

### DIFF
--- a/ksqldb-engine/src/test/java/io/confluent/ksql/function/udf/string/ToBytesTest.java
+++ b/ksqldb-engine/src/test/java/io/confluent/ksql/function/udf/string/ToBytesTest.java
@@ -42,6 +42,8 @@ public class ToBytesTest {
             is(ByteBuffer.wrap(new byte[]{})));
         assertThat(udf.toBytes("1a2B", "hex"),
             is(ByteBuffer.wrap(new byte[]{26, 43})));
+        assertThat(udf.toBytes("1A2B", "hex"),
+            is(ByteBuffer.wrap(new byte[]{26, 43})));
     }
 
     @Test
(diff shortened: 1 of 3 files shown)


ksqldb-engine/src/test/java/io/confluent/ksql/function/udf/string/ToBytesTest.java
```diff
@@ -40,6 +40,8 @@ public void shouldConvertHexStringToBytes() {
             is(ByteBuffer.wrap(new byte[]{33})));
         assertThat(udf.toBytes("", "hex"),
             is(ByteBuffer.wrap(new byte[]{})));
+        assertThat(udf.toBytes("1a2B", "hex"),
+            is(ByteBuffer.wrap(new byte[]{26, 43})));
```
spena, line 44: can you add a test with same hex but all uppercase? Just to make sure that both return the same value.

spena (approved):
I left one minor comment

PR #8431, opened 2021-12-01
test: Attempts to speed up ScalablePushBandwidthThrottleIntegrationTest

AlanConfluent:
### Description 
I think this test is slow because the iterations can take a while to do, so this boils down the iterations to just 5 total rather than 12.

### Testing done 
Ran test locally.

### Reviewer checklist


nateab (approved):
LGTM! Any reason to not reduce iterations even further down to 2 for example?

AlanConfluent:
> LGTM! Any reason to not reduce iterations even further down to 2 for example?

No real reason.  Turning down to 2!

PR #8434, opened 2021-12-01
feature: allow shared runtimes to be restarted to pick up config changes

wcarlson5:
create new topologies and use those to get a new kafka streams runtimes

### Testing done 

### Reviewer checklist

### DIFF
--- a/ksqldb-engine/src/main/java/io/confluent/ksql/util/BinPackedPersistentQueryMetadataImpl.java
+++ b/ksqldb-engine/src/main/java/io/confluent/ksql/util/BinPackedPersistentQueryMetadataImpl.java
@@ -138,15 +138,15 @@ public class BinPackedPersistentQueryMetadataImpl implements PersistentQueryMeta
     this.materializationProviderBuilder =
         requireNonNull(materializationProviderBuilder, "materializationProviderBuilder");
     this.listener = requireNonNull(listener, "listener");
+    this.namedTopologySupplier = requireNonNull(namedTopologySupplier, "namedTopologySupplier");
     this.materializationProvider = materializationProviderBuilder
             .flatMap(builder -> builder.apply(
                     this.sharedKafkaStreamsRuntime.getKafkaStreams(),
-                    getTopology()
+                    topology
             ));
     this.classifier = requireNonNull(classifier, "classifier");
     this.streamsProperties = requireNonNull(streamsProperties, "streamsProperties");
     this.scalablePushRegistry = requireNonNull(scalablePushRegistry, "scalablePushRegistry");
-    this.namedTopologySupplier = requireNonNull(namedTopologySupplier, "namedTopologySupplier");
   }
 
 
--- a/ksqldb-engine/src/main/java/io/confluent/ksql/util/BinPackedPersistentQueryMetadataImpl.java
+++ b/ksqldb-engine/src/main/java/io/confluent/ksql/util/BinPackedPersistentQueryMetadataImpl.java
@@ -147,7 +147,6 @@ public class BinPackedPersistentQueryMetadataImpl implements PersistentQueryMeta
     this.classifier = requireNonNull(classifier, "classifier");
     this.streamsProperties = requireNonNull(streamsProperties, "streamsProperties");
     this.scalablePushRegistry = requireNonNull(scalablePushRegistry, "scalablePushRegistry");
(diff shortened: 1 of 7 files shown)


wcarlson5:
@guozhangwang @ableegoldman I have ran into issues reusing topologies or builds before so I just made it from scratch. Will that cause other issues? I think the offsets are based on group ID so that shouldn't be an issue i think

ksqldb-engine/src/main/java/io/confluent/ksql/util/BinPackedPersistentQueryMetadataImpl.java
```diff
@@ -290,10 +295,9 @@ public String getQueryApplicationId() {
     return applicationId;
   }
 
-  @SuppressFBWarnings(value = "EI_EXPOSE_REP", justification = "topology is for reference")
   @Override
   public NamedTopology getTopology() {
-    return topology;
+    return namedTopologySupplier.get();
```
ableegoldman, line 300: Was  `getTopology` ever being called before this? If so then it probably wants the actual NamedTopology for that query, which is important. How about adding a separate method for this, eg `getTopologyCopy`

ksqldb-engine/src/main/java/io/confluent/ksql/util/SharedKafkaStreamsRuntimeImpl.java
```diff
@@ -167,4 +168,22 @@ public void start(final QueryId queryId) {
...
+        .buildNamedTopologyWrapper(streamsProperties);
+    kafkaStreams = kafkaStreamsNamedTopologyWrapper;
+    for (final BinPackedPersistentQueryMetadataImpl binPackedPersistentQueryMetadata
+        : collocatedQueries.values()) {
+      kafkaStreamsNamedTopologyWrapper.addNamedTopology(
+          binPackedPersistentQueryMetadata.getTopology()
+      );
+    }
+    kafkaStreams.close();
+    kafkaStreams.cleanUp();
```
ableegoldman, line 184: We should avoid calling cleanup to keep this lightweight, otherwise each config change will restore everything from scratch 😬

ableegoldman (commented):
Two things to address but otherwise LGTM

ksqldb-engine/src/main/java/io/confluent/ksql/util/SharedKafkaStreamsRuntimeImpl.java
wcarlson5: good call

ksqldb-engine/src/main/java/io/confluent/ksql/util/BinPackedPersistentQueryMetadataImpl.java
wcarlson5: I don't think that it really needs the original. But lets add a separate method just incase

ksqldb-engine/src/main/java/io/confluent/ksql/util/SandboxedSharedKafkaStreamsRuntimeImpl.java
```diff
@@ -42,7 +43,7 @@ public SandboxedSharedKafkaStreamsRuntimeImpl(
     );
 
     for (BinPackedPersistentQueryMetadataImpl query : sharedRuntime.collocatedQueries.values()) {
-      //kafkaStreams.addNamedTopology(query.getTopology());
```
ableegoldman, line 45: I think this might still be an issue for the validation runtimes because we want to build the topology with a new NamedTopologyBuilder that we got from the validation runtime, not the one based on the original query's actual runtime. But if the tests pass for now this is ok with me and we can clean it up later (it might technically work but is a bit awkward)

wcarlson5: yeah I have an idea how to clean it up, I think we might need to have the copy method take a parameter. but we should do that as a follow up

ableegoldman: sorry yes to clarify I'm not asking for any changes to this in the current PR

ableegoldman (approved):
LGTM, please give me a ping when you've merged it btw!

PR #8445, opened 2021-12-03
feat: Validate connector config before creating it

Gerrrr:
### Description 

With this patch `ConnectExecutor` validates the connector configuration before submitting a create request. 

#### Why?

Why do we need an additional request for validation? Isn't it enough to just return the error from the create request. Turns out, the errors returned by the create requests might not contain enough information to understand what is the actual issue. Here is an example error response:

```json
{
  "error_code": 400,
  "message": "Connector configuration is invalid and contains the following 1 error(s):\nA value is required\nYou can also find the above list of errors at the endpoint `/connector-plugins/{connectorType}/config/validate`"
}
```

From that message it is not clear what field is actually missing. In turn, suggested `/config/validate` endpoint provides structured per-field information. See https://docs.confluent.io/platform/current/connect/references/restapi.html#put--connector-plugins-(string-name)-config-validate for more details.

#### What?

* `io.confluent.ksql.services.ConnectClient` interface has a new method `validate` that calls `/config/validate` for the given plugin and configuration.
* `io.confluent.ksql.rest.server.execution.ConnectExecutor` has a new private `validate` method that returns a list of error messages, one string per field. If a field has multiple error messages, they are joined together. If the validate request fails (as in, 502 or 404), this failure is returned.
*  `io.confluent.ksql.rest.server.execution.ConnectExecutor#execute` calls `validate` before creating the connector. If `validate` returns at least one error, the method returns said error(s) and does not execute the create statement.

### Testing done 

* Added unit tests for `DefaultConnectClient#validate`.
* Added unit tests for validation errors in `ConnectExecutorTest`.

### Reviewer checklist

### DIFF
--- a/ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/execution/ConnectExecutor.java
+++ b/ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/execution/ConnectExecutor.java
@@ -15,6 +15,7 @@
 
 package io.confluent.ksql.rest.server.execution;
 
+import com.google.common.collect.ImmutableList;
 import com.google.common.collect.Maps;
 import io.confluent.ksql.KsqlExecutionContext;
 import io.confluent.ksql.connect.supported.Connectors;
@@ -28,9 +29,14 @@ import io.confluent.ksql.services.ConnectClient;
 import io.confluent.ksql.services.ConnectClient.ConnectResponse;
 import io.confluent.ksql.services.ServiceContext;
 import io.confluent.ksql.statement.ConfiguredStatement;
+import io.confluent.ksql.util.ParserUtil;
+import java.util.HashMap;
 import java.util.List;
+import java.util.Map;
 import java.util.Optional;
+import java.util.stream.Collectors;
 import org.apache.commons.lang3.StringUtils;
+import org.apache.kafka.connect.runtime.rest.entities.ConfigInfos;
 import org.apache.kafka.connect.runtime.rest.entities.ConnectorInfo;
 
 public final class ConnectExecutor {
@@ -46,6 +52,13 @@ public final class ConnectExecutor {
   ) {
     final CreateConnector createConnector = statement.getStatement();
     final ConnectClient client = serviceContext.getConnectClient();
+
+    final List<String> errors = validate(createConnector, client);
+    if (!errors.isEmpty()) {
+      final String errorMessage = "Validation error: " + String.join("\n", errors);
+      return StatementExecutorResponse.handled(Optional.of(new ErrorEntity(
+          statement.getStatementText(), errorMessage)));
+    }
 
     final Optional<KsqlEntity> connectorsResponse = handleIfNotExists(
         statement, createConnector, client);
@@ -79,6 +92,45 @@ public final class ConnectExecutor {
 
     return StatementExecutorResponse.handled(response.error()
         .map(err -> new ErrorEntity(statement.getStatementText(), err)));
+  }
+
+  private static List<String> validate(final CreateConnector createConnector,
+      final ConnectClient client) {
+    final Map<String, String> config = new HashMap<>(createConnector.getConfig().size());
+    createConnector.getConfig().forEach((k, v) ->
+        // Parsing the statement wraps string fields with single quotes which breaks the
+        // validation.
+        config.put(k, ParserUtil.unquote(v.toString(),"'")));
+    config.put("name", createConnector.getName());
+
+    final String connectorType = config.get("connector.class");
+    // Request with an empty connector type in the url results in 404.
+    // Request with a connector type that contains only spaces results in 405.
+    // In both cases it is user-friendlier to return the actual error.
+    if (connectorType == null) {
+      // Return error message identical to the one produced by the connector
+      return ImmutableList.of(String.format(
+          "Connector config %s contains no connector type", config));
+    } else if (connectorType.trim().isEmpty()) {
+      return ImmutableList.of("Connector type cannot be empty");
+    }
+
+    final ConnectResponse<ConfigInfos> response = client.validate(connectorType, config);
+    if (response.error().isPresent()) {
+      return ImmutableList.of(response.error().get());
+    } else if (response.datum().isPresent()) {
+      return response
+          .datum()
+          .get()
+          .values()
+          .stream()
+          .filter(configInfo -> !configInfo.configValue().errors().isEmpty())
+          .map(configInfo -> configInfo.configValue().name()
+              + " - "
+              + String.join(". ", configInfo.configValue().errors()))
+          .collect(Collectors.toList());
+    }
+    return ImmutableList.of();
   }
 
   private static Optional<KsqlEntity> handleIfNotExists(
--- a/ksqldb-rest-app/src/test/java/io/confluent/ksql/rest/server/execution/ConnectExecutorTest.java
+++ b/ksqldb-rest-app/src/test/java/io/confluent/ksql/rest/server/execution/ConnectExecutorTest.java
@@ -18,6 +18,7 @@ package io.confluent.ksql.rest.server.execution;
 import static org.hamcrest.MatcherAssert.assertThat;
 import static org.hamcrest.Matchers.hasEntry;
 import static org.hamcrest.Matchers.instanceOf;
+import static org.hamcrest.Matchers.is;
 import static org.mockito.ArgumentMatchers.anyMap;
 import static org.mockito.ArgumentMatchers.anyString;
 import static org.mockito.ArgumentMatchers.eq;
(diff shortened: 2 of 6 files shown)


ksqldb-rest-app/src/test/java/io/confluent/ksql/rest/server/execution/ConnectExecutorTest.java
```diff
@@ -161,18 +191,107 @@ public void shouldReturnErrorIfConnectorExists() {
...
+        ImmutableList.of()),
+        new ConfigValueInfo("port", null, ImmutableList.of(), ImmutableList.of(), true));
+    when(connectClient.validate(anyString(), anyMap()))
+        .thenReturn(ConnectResponse.success(
+            new ConfigInfos(
+                "foo",
+                2,
+                ImmutableList.of(),
+                ImmutableList.of(configInfo1, configInfo2, configInfo3)),
+            HttpStatus.SC_BAD_REQUEST));
```
vcrfxia, line 284: Does a validation failure actually result in a 400 return code?

ksqldb-rest-app/src/test/java/io/confluent/ksql/rest/server/execution/ConnectExecutorTest.java
```diff
@@ -118,13 +128,31 @@ public void shouldReturnConnectorInfoEntityOnSuccess() {
...
+    givenValidationError();
+    givenCreationSuccess();
+
+    // When:
+    final Optional<KsqlEntity> entity = ConnectExecutor
+        .execute(CREATE_CONNECTOR_CONFIGURED, mock(SessionProperties.class), null,
+            serviceContext).getEntity();
 
     // Then:
     assertThat("Expected non-empty response", entity.isPresent());
```
vcrfxia, line 158: Is it difficult to add a check for the actual error message returned to the user in this test? It'd be nice to see an example of the error message produced to ensure that the formatting is sane, and to prevent accidental regressions in the future. (Looks fine from the code you added but it's always nice to double-check in a test if possible.)

ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/execution/ConnectExecutor.java
```diff
@@ -81,6 +92,44 @@ public static StatementExecutorResponse execute(
...
+    // Request with an empty connector type in the url results in 404.
+    // Request with a connector type that contains only spaces results in 405.
+    // In both cases it is user-friendlier to return the actual error.
+    if (connectorType == null) {
+      // Return error message identical to the one produced by the connector
+      return ImmutableList.of(String.format(
+          "Connector config %s contains no connector type", config));
+    } else if (connectorType
+        .replace("'", "")
+        .replace("\"", "")
```
vcrfxia, line 109: Out of curiosity, why replace quotes in particular (and not other special characters as well)?

vcrfxia (commented):
Thanks @Gerrrr -- this is fantastic! A few minor questions inline.

ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/execution/ConnectExecutor.java
Gerrrr: Consider `shouldReturnErrorIfConnectorTypeIsEmpty`:

https://github.com/confluentinc/ksql/blob/d2db98c9ac464d60a82c1475e7a23c64c03c5528/ksqldb-rest-app/src/test/java/io/confluent/ksql/rest/server/execution/ConnectExecutorTest.java#L215-L235

In this case `connectorType` will look something like `' '` - a whitespace surrounded by single quotes. 

I didn't remove other special characters because the purpose of these transformations is to hedge against sending an empty connector type(i.e. `""` or a string containing only whitespaces). As for different unicode space characters, Java's `String#trim` is supposed to remove them. For a peace of mind, I tested a couple of those and it worked fine :).

ksqldb-rest-app/src/test/java/io/confluent/ksql/rest/server/execution/ConnectExecutorTest.java
Gerrrr: Added in 77f6de69656c41aac1f463d2c993e797dc1ba4f0.

ksqldb-rest-app/src/test/java/io/confluent/ksql/rest/server/execution/ConnectExecutorTest.java
Gerrrr: Good catch! It's actually 200. Fixed in 77f6de69656c41aac1f463d2c993e797dc1ba4f0.

ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/execution/ConnectExecutor.java
```diff
@@ -47,6 +52,12 @@ public static StatementExecutorResponse execute(
     final CreateConnector createConnector = statement.getStatement();
     final ConnectClient client = serviceContext.getConnectClient();
 
+    final List<String> errors = validate(createConnector, client);
+    if (!errors.isEmpty()) {
+      return StatementExecutorResponse.handled(Optional.of(new ErrorEntity(
+          statement.getStatementText(), String.join("\n", errors))));
```
vcrfxia, line 58: Can we add a sentence to this error message to explain that the connector configuration failed validation? Otherwise the specific errors themselves are a bit hard to understand, without this addition context.

(Thanks for the additional test coverage for error messages, BTW. Makes it much easier to reason about these types of things :) )

vcrfxia (commented):
Thanks @Gerrrr ! One more suggestion for an error message improvement. Otherwise LGTM.

ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/execution/ConnectExecutor.java
Gerrrr: Done in 1bdc6f3.

vcrfxia (approved):
Thanks @Gerrrr ! LGTM, though I think unit tests might need to be updated for the latest error message change?

ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/execution/ConnectExecutor.java
```diff
@@ -81,6 +93,45 @@ public static StatementExecutorResponse execute(
         .map(err -> new ErrorEntity(statement.getStatementText(), err)));
   }
 
+  private static List<String> validate(final CreateConnector createConnector,
+      final ConnectClient client) {
+    final Map<String, String> config = new HashMap<>(createConnector.getConfig().size());
+    createConnector.getConfig().forEach((k, v) ->
+        // Parsing the statement wraps string fields with single quotes which breaks the
+        // validation.
+        config.put(k, v.toString().replace("'", "")));
```
vcrfxia, line 102: Is it safe to simply replace all single-quotes like this? It might be safer to instead use something similar to https://github.com/confluentinc/ksql/blob/e4c229517173d1f3feac6b29c1b1f88f9f4f8a43/ksqldb-parser/src/main/java/io/confluent/ksql/util/ParserUtil.java#L98

Gerrrr: Ah cool, I didn't know about this one. Thanks for pointing it out!

ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/execution/ConnectExecutor.java
```diff
@@ -81,6 +94,45 @@ public static StatementExecutorResponse execute(
         .map(err -> new ErrorEntity(statement.getStatementText(), err)));
   }
 
+  private static List<String> validate(final CreateConnector createConnector,
+      final ConnectClient client) {
+    final Map<String, String> config = new HashMap<>(createConnector.getConfig().size());
+    createConnector.getConfig().forEach((k, v) ->
+        // Parsing the statement wraps string fields with single quotes which breaks the
```
vcrfxia, line 101: The unit tests added in the first version of this PR didn't catch this bug (that we weren't stripped quotes from configs before passing them for validation), right? Is there test coverage somewhere in the codebase that prevents a regression here? Maybe ConnectIntegrationTest.java?

Gerrrr: Yes, this bug was caught by `ConnectIntegrationTest`. `ConnectExecutorTest` didn't catch it because we mock the connector there.

PR #8449, opened 2021-12-04
feat: Adds support to /query-stream for http1.1 and StreamedRow json format

AlanConfluent:
### Description 
This is part of an effort to consolidate endpoints.  What this aims to do is add support for StreamedRow to `/query-stream`.  This should allow a followup change to install the same handler for `/query`.  Then we can support multiple paths under the same handling logic.

Also, this allows handling of http 1.1 requests using `/query-stream`.

### Testing done 
Ran tests and manually tested.

### Reviewer checklist

### DIFF
--- a/ksqldb-api-client/src/test/java/io/confluent/ksql/api/client/ClientTest.java
+++ b/ksqldb-api-client/src/test/java/io/confluent/ksql/api/client/ClientTest.java
@@ -215,7 +215,7 @@ public class ClientTest extends BaseApiTest {
     // Then
     assertThat(streamedQueryResult.columnNames(), is(DEFAULT_COLUMN_NAMES));
     assertThat(streamedQueryResult.columnTypes(), is(DEFAULT_COLUMN_TYPES));
-    assertThat(streamedQueryResult.queryID(), is(nullValue()));
+    assertThat(streamedQueryResult.queryID(), is(notNullValue()));
 
     shouldReceiveRows(streamedQueryResult, true);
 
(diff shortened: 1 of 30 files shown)


vvcephei (commented):
Thanks for this, it's very cool, and very elegant, @AlanConfluent !

The change seems pretty straightforward, but I'm wondering about the testing. I think the existing tests are sufficient to demonstrate that we haven't introduced a regression, but how can we be sure that `/query-stream` is correctly sending StreamedRows, or that it actually works properly with http 1.1?

It seems like there should be some way to verify this using the same set of tests that we currently have. For example, can we paramterize the API test for `/query-stream` so that it first runs with http 2 and then with http 1.1? Likewise, can we parameterize the test also to run requesting the StreamedRow format, and then verify the same response invariants hold while also verifying that the correct format is actually used?

AlanConfluent:
@vvcephei 

> The change seems pretty straightforward, but I'm wondering about the testing. I think the existing tests are sufficient to demonstrate that we haven't introduced a regression, but how can we be sure that `/query-stream` is correctly sending StreamedRows, or that it actually works properly with http 1.1?

I added a couple functional tests specifically for this case of http 1.1 with StreamedRow on `/query-stream`.  Once we actually switch over to `/query` using this same endpoint, it'll effectively inherit all existing tests.

> It seems like there should be some way to verify this using the same set of tests that we currently have. For example, can we paramterize the API test for `/query-stream` so that it first runs with http 2 and then with http 1.1? Likewise, can we parameterize the test also to run requesting the StreamedRow format, and then verify the same response invariants hold while also verifying that the correct format is actually used?

At the moment, we just have a bunch of tests.  I can potentially look into a nicer parameterizing of the tests once we switch over officially to using the newer handler with `/query`.  I'm honestly not sure how much use there will strictly be in testing every combination of http_version x format x endpoint.  If you see the code itself, there's not too much interplay between these things and the bunch of tests we currently have are probably sufficient.  Also, if we were interested, I could always artificially limit some of these combinations to be disallowed since they're currently very limited at the moment.  Tell me what you think.

vvcephei (approved):
Thanks, @AlanConfluent . My preference would still be to parameterize the rest API test, but I don't have a specific reason for it; just general paranoia. If you don't think it's worth it, then we don't have to.

AlanConfluent:
> Thanks, @AlanConfluent . My preference would still be to parameterize the rest API test, but I don't have a specific reason for it; just general paranoia. If you don't think it's worth it, then we don't have to.

It would be nice if we had junit5, so that we could use ParameterizedTest.  Since we're on junit4 and `RestApiTest` doesn't use Parmeterized, I just have a test that loops through all combinations.  It should be as easy as adding in the new endpoint when I enable this for `/query`.

PR #8452, opened 2021-12-06
feat: verify no other query is running before putting into command topic

tolgadur:
### Description 
We want to make sure that no other query is running before changing `processing.guarantee` with the new `ALTER SYSTEM` query

```
ksql>  alter system 'processing.guarantee'='exactly_once';
Could not write the statement 'alter system 'processing.guarantee'='exactly_once';' into the command topic.
Caused by: Invalid value null for configuration Failed to set
	processing.guarantee to exactly_once. Please terminate all running queries
	before attempting to set processing.guarantee
```

### Testing done 
unit test + manual testing 

### Reviewer checklist

### DIFF
--- a/ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/computation/ValidatedCommandFactory.java
+++ b/ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/computation/ValidatedCommandFactory.java
@@ -15,8 +15,12 @@
 
 package io.confluent.ksql.rest.server.computation;
 
+import static org.apache.kafka.streams.StreamsConfig.PROCESSING_GUARANTEE_CONFIG;
+
 import com.fasterxml.jackson.core.JsonProcessingException;
 import io.confluent.ksql.KsqlExecutionContext;
+import io.confluent.ksql.config.ConfigItem;
+import io.confluent.ksql.config.KsqlConfigResolver;
 import io.confluent.ksql.engine.KsqlPlan;
 import io.confluent.ksql.execution.json.PlanJsonMapper;
 import io.confluent.ksql.parser.tree.AlterSystemProperty;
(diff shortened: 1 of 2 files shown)


ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/computation/ValidatedCommandFactory.java
```diff
@@ -133,6 +138,17 @@ private static Command createForAlterSystemQuery(
               propertyName, propertyValue, propertyName), null);
     }
 
+    // verify that no persistent query is running when attempting to change 'processing.guarantee'
+    final KsqlConfigResolver resolver = new KsqlConfigResolver();
+    final Optional<ConfigItem> resolvedItem = resolver.resolve(propertyName, false);
+    if (resolvedItem.isPresent()
+        && Objects.equals(resolvedItem.get().getPropertyName(), PROCESSING_GUARANTEE_CONFIG)
+        && !context.getPersistentQueries().isEmpty()) {
```
wcarlson5, line 146: do we know this will be true when the command is executed? can there be queries in the cmd topic that have not been started yet or something?

wcarlson5 (commented):
just one question from me :)

ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/computation/ValidatedCommandFactory.java
tolgadur: cc @rodesai

wcarlson5 (approved):
lets just get this merged and we can add more features later

ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/computation/ValidatedCommandFactory.java
rodesai: Nope - this should be a sandbox copy of the engine, with the copy being taken during a transaction. So if any other queries slipped in the transaction would fail to commit.

rodesai (approved):
LGTM

PR #8459, opened 2021-12-06
chore: enable tests 

wcarlson5:
there were a few tests that were broken that are fixed now that we have blocking add and remove topologies merged

### Testing done 

### Reviewer checklist

### DIFF
--- a/ksqldb-engine/src/test/java/io/confluent/ksql/integration/EndToEndIntegrationTest.java
+++ b/ksqldb-engine/src/test/java/io/confluent/ksql/integration/EndToEndIntegrationTest.java
@@ -110,7 +110,7 @@ public class EndToEndIntegrationTest {
   @Parameterized.Parameters(name = "{0}")
   public static Collection<Boolean> data() {
     return Arrays.asList(
-        false
+        false, true
     );
   }
 
--- a/ksqldb-engine/src/test/java/io/confluent/ksql/integration/EndToEndIntegrationTest.java
+++ b/ksqldb-engine/src/test/java/io/confluent/ksql/integration/EndToEndIntegrationTest.java
@@ -308,14 +308,14 @@ public class EndToEndIntegrationTest {
     );
 
     // Then:
-    final String valSbuejct = KsqlConstants.getSRSubject(topicName, false);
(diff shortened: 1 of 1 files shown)


lct45 (approved):
lgtm when build passes

ableegoldman:
Looks like we still have a few failing tests:
```
io.confluent.ksql.integration.EndToEndIntegrationTest.shouldSupportDroppingAndRecreatingJoinQuery[true]
io.confluent.ksql.integration.EndToEndIntegrationTest.shouldRegisterCorrectPrimitiveSchemaForCreateAsStatements[true]
io.confluent.ksql.integration.EndToEndIntegrationTest.shouldCleanUpAvroSchemaOnDropSource[true]
```

@wcarlson5 did the build run since merging the changes to use blocking add/remove or could this be stale?

wcarlson5:
@ableegoldman I did run it. Its passing locally though every time, I am looking into it though. I am not sure why Jenkins and local would be different.

ableegoldman:
I replayed the tests, maybe it just needs to be re-run since merging the PR?

ableegoldman:
I do see some seemingly weird things in the logs btw, for example in `shouldCleanUpAvroSchemaOnDropSource` we only ever see one copy of a KafkaStreams app being created before hitting this error (eg the line `Kafka Streams version:` is only printed once, we see it come up (without being started) then shut down to NOT_RUNNING -- then some time after that, all of a sudden we hit this error. Could we be reusing the validation runtime after it was shut down? regardless of whether this particular test is failing, that seems very odd to me

wcarlson5:
That is really odd. The Validation runtimes should not be the issue since the PR that has them use the topologies is not merged yet. Also they are never started so they shouldn't be able to make a stateDir

Maybe Jenkins is running these three tests at once? 

Are those the full logs because that doesn't quite seem like what I would expect at all?

ableegoldman:
Idk,  but I see that the same tests failed again.

@wcarlson5 I think it might make sense for me to hand off the one large remaining ksql ticket on my plate ([KCI-859](https://confluentinc.atlassian.net/browse/KCI-859)) to you, and then tackle any new or existing debugging/investigational work as my top priority,  alongside reviewing and the planning work for MD -- let's discuss tomorrow

PR #8460, opened 2021-12-07
docs: Add http protocol to the default ksql.connect.url value

Gerrrr:
### Description 

Default value for `ksql.connect.url` used in the docs - `localhost:8083` is not correct. Setting this value in `ksql-server.properties` leads to the following error:

```
ksql> CREATE SOURCE CONNECTOR `jdbc-connector` WITH(
>  "connector.class"='io.confluent.connect.jdbc.JdbcSourceConnector',
>  "connection.url"='jdbc:postgresql://localhost:5432/user1',
>  "mode"='bulk',
>  "topic.prefix"='jdbc-',
>  "key"='username',
>  "connection.username"='user1');
io.confluent.ksql.util.KsqlServerException: org.apache.hc.client5.http.ClientProtocolException: Target host is not specified
Caused by: org.apache.hc.client5.http.ClientProtocolException: Target host is
	not specified
Caused by: Target host is not specified
```

Instead, the value should be `http://localhost:8083` - the same as the default in the code:

https://github.com/confluentinc/ksql/blob/5c9be20917f1f88b0587310134cc3b48b396e345/ksqldb-common/src/main/java/io/confluent/ksql/util/KsqlConfig.java#L180

### Testing done 

Manual testing.

### Reviewer checklist

### DIFF
--- a/docs/concepts/connectors.md
+++ b/docs/concepts/connectors.md
@@ -30,7 +30,7 @@ There are two ways to deploy the ksqlDB-Connect integration:
 
 1.  **External**: If a {{ site.kconnect }} cluster is available, set the
     `ksql.connect.url` property in your ksqlDB Server configuration file.
-    The default value for this is `localhost:8083`.
+    The default value for this is `http://localhost:8083`.
 2.  **Embedded**: ksqlDB can double as a {{ site.kconnect }} server and
     will run a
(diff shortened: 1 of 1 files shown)


JimGalasyn (approved):
LGTM, thanks for the fix!

PR #8466, opened 2021-12-08
chore: get query-specific errors from StreamsException

ableegoldman:
Now that we have the NamedTopology info in the StreamsException class, we should check whether a given error has originated from a specific query and add it to that query's error queue rather than the runtime error queue instead

### DIFF
--- a/ksqldb-engine/src/main/java/io/confluent/ksql/util/SharedKafkaStreamsRuntimeImpl.java
+++ b/ksqldb-engine/src/main/java/io/confluent/ksql/util/SharedKafkaStreamsRuntimeImpl.java
@@ -20,13 +20,16 @@ import io.confluent.ksql.query.KafkaStreamsBuilder;
 import io.confluent.ksql.query.QueryError;
 import io.confluent.ksql.query.QueryErrorClassifier;
 import io.confluent.ksql.query.QueryId;
+import io.confluent.ksql.util.QueryMetadataImpl.TimeBoundedQueue;
 import java.time.Duration;
 import java.util.Collection;
 import java.util.Map;
 import java.util.concurrent.ExecutionException;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.TimeoutException;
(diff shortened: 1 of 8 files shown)


ksqldb-engine/src/main/java/io/confluent/ksql/util/SharedKafkaStreamsRuntimeImpl.java
```diff
@@ -94,20 +99,54 @@ public void register(
...
+      final String queryId;
+      if (e instanceof StreamsException && ((StreamsException) e).taskId().isPresent()) {
+        queryId = ((StreamsException) e).taskId().get().topologyName();
+      } else {
+        queryId = null;
+      }
+
+      if (queryId == null) {
+        for (BinPackedPersistentQueryMetadataImpl query : collocatedQueries.values()) {
+          query.getListener().onError(collocatedQueries.get(query.getQueryId()), queryError);
```
wcarlson5, line 112: why not just use the query's `setQueryError` ? Then the `getQueryErrors` is much simpler because you already have all the relevant errors in the queue

wcarlson5 (commented):
I think this is good. But it would be nice to add a test for the new handler that makes sure the errors end up in the right queue

ableegoldman:
> it would be nice to add a test

I'm working on it, IntelliJ is giving me a hard time about compiling :/ But thanks for the quick review!

ksqldb-engine/src/main/java/io/confluent/ksql/util/SharedKafkaStreamsRuntimeImpl.java
ableegoldman: You do mean just remove the `runtimeErrors` queue altogether and add any runtime-wide errors to all query error queues, yes? SGTM, the errors should be pretty small so not worth trying to de-duplicate

wcarlson5 (approved):
Once that test is added this is good to go!

PR #8472, opened 2021-12-08
feat: new runtime selection

wcarlson5:
Move the runtime selection into the query plan. The runtime should be assigned and then validated. If validated correctly it will be assigned the same runtime and put on the cmd topic. If not validated the actual assignor will not not be affected and will move on to the next query. 

When receiving the plan to build the query the query build will find the runtime with that ID. If it doesn't exist it will make a new one.

Right now its doesn't check for config compatibility which is how it currently is. There needs to be some more work for config checking, but that should be a follow up PR. 

### Testing done 

### Reviewer checklist

### DIFF
--- a/ksqldb-common/src/main/java/io/confluent/ksql/util/KsqlConfig.java
+++ b/ksqldb-common/src/main/java/io/confluent/ksql/util/KsqlConfig.java
@@ -547,7 +547,7 @@ public class KsqlConfig extends AbstractConfig {
           + "CREATE SOURCE [TABLE|STREAM] statements will continue being read-only.";
 
   public static final String KSQL_SHARED_RUNTIME_ENABLED = "ksql.runtime.feature.shared.enabled";
-  public static final Boolean KSQL_SHARED_RUNTIME_ENABLED_DEFAULT = false;
+  public static final Boolean KSQL_SHARED_RUNTIME_ENABLED_DEFAULT = true;
   public static final String KSQL_SHARED_RUNTIME_ENABLED_DOC =
       "Feature flag for sharing streams runtimes. "
           + "Default is false. If false, persistent queries will use separate "
--- a/ksqldb-common/src/main/java/io/confluent/ksql/util/KsqlConfig.java
+++ b/ksqldb-common/src/main/java/io/confluent/ksql/util/KsqlConfig.java
@@ -547,7 +547,7 @@ public class KsqlConfig extends AbstractConfig {
           + "CREATE SOURCE [TABLE|STREAM] statements will continue being read-only.";
 
   public static final String KSQL_SHARED_RUNTIME_ENABLED = "ksql.runtime.feature.shared.enabled";
-  public static final Boolean KSQL_SHARED_RUNTIME_ENABLED_DEFAULT = true;
+  public static final Boolean KSQL_SHARED_RUNTIME_ENABLED_DEFAULT = false;
   public static final String KSQL_SHARED_RUNTIME_ENABLED_DOC =
       "Feature flag for sharing streams runtimes. "
           + "Default is false. If false, persistent queries will use separate "
--- a/ksqldb-engine/src/main/java/io/confluent/ksql/engine/RuntimeAssignor.java
+++ b/ksqldb-engine/src/main/java/io/confluent/ksql/engine/RuntimeAssignor.java
@@ -1,13 +1,27 @@
+/*
+ * Copyright 2021 Confluent Inc.
+ *
+ * Licensed under the Confluent Community License (the "License"); you may not use
+ * this file except in compliance with the License.  You may obtain a copy of the
+ * License at
+ *
+ * http://www.confluent.io/confluent-community-license
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OF ANY KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations under the License.
+ */
+
 package io.confluent.ksql.engine;
+
+import static io.confluent.ksql.util.QueryApplicationId.buildSharedRuntimeId;
 
 import io.confluent.ksql.name.SourceName;
 import io.confluent.ksql.query.QueryId;
 import io.confluent.ksql.util.BinPackedPersistentQueryMetadataImpl;
 import io.confluent.ksql.util.KsqlConfig;
 import io.confluent.ksql.util.PersistentQueryMetadata;
-
-import static io.confluent.ksql.util.QueryApplicationId.buildSharedRuntimeId;
-
 import java.util.Collection;
 import java.util.HashMap;
 import java.util.HashSet;
@@ -25,9 +39,9 @@ public class RuntimeAssignor {
 
   public RuntimeAssignor(final KsqlConfig config, final int runtimes) {
     runtimesToSources = new HashMap<>();
-    idToRuntime= new HashMap<>();
-    for(int i=0; i < runtimes; i++) {
-      String runtime = buildSharedRuntimeId(config, true, i);
+    idToRuntime = new HashMap<>();
+    for (int i = 0; i < runtimes; i++) {
+      final String runtime = buildSharedRuntimeId(config, true, i);
       runtimesToSources.put(runtime, new HashSet<>());
     }
   }
@@ -35,7 +49,8 @@ public class RuntimeAssignor {
   private RuntimeAssignor(final RuntimeAssignor other) {
     this.runtimesToSources = new HashMap<>();
     this.idToRuntime = new HashMap<>();
-    for (Map.Entry<String, Collection<SourceName>> runtime : other.runtimesToSources.entrySet()) {
+    for (Map.Entry<String, Collection<SourceName>> runtime
+        : other.runtimesToSources.entrySet()) {
       this.runtimesToSources.put(runtime.getKey(), new HashSet<>(runtime.getValue()));
     }
     this.idToRuntime.putAll(other.idToRuntime);
@@ -45,7 +60,9 @@ public class RuntimeAssignor {
     return new RuntimeAssignor(this);
   }
 
-  public String getRuntime(final QueryId queryId, final Collection<SourceName> sources, final KsqlConfig config) {
+  public String getRuntime(final QueryId queryId,
+                           final Collection<SourceName> sources,
+                           final KsqlConfig config) {
     if (idToRuntime.containsKey(queryId)) {
       return idToRuntime.get(queryId);
     }
@@ -69,7 +86,8 @@ public class RuntimeAssignor {
 
   public void dropQuery(final PersistentQueryMetadata queryMetadata) {
     if (queryMetadata instanceof BinPackedPersistentQueryMetadataImpl) {
-      runtimesToSources.get(queryMetadata.getQueryApplicationId()).removeAll(queryMetadata.getSourceNames());
+      runtimesToSources.get(queryMetadata.getQueryApplicationId())
+          .removeAll(queryMetadata.getSourceNames());
       idToRuntime.remove(queryMetadata.getQueryId());
     }
   }
@@ -84,7 +102,8 @@ public class RuntimeAssignor {
   public void rebuildAssignment(final Collection<PersistentQueryMetadata> queries) {
     for (PersistentQueryMetadata queryMetadata: queries) {
       if (queryMetadata instanceof BinPackedPersistentQueryMetadataImpl) {
-        runtimesToSources.put(queryMetadata.getQueryApplicationId(), queryMetadata.getSourceNames());
+        runtimesToSources.put(queryMetadata.getQueryApplicationId(),
+            queryMetadata.getSourceNames());
         idToRuntime.put(queryMetadata.getQueryId(), queryMetadata.getQueryApplicationId());
       }
     }
--- a/ksqldb-engine/src/main/java/io/confluent/ksql/engine/RuntimeAssignor.java
+++ b/ksqldb-engine/src/main/java/io/confluent/ksql/engine/RuntimeAssignor.java
@@ -39,6 +39,7 @@ public class RuntimeAssignor {
   }
 
   public RuntimeAssignor createSandbox() {
+
     return new RuntimeAssignor(this);
   }
 
--- a/ksqldb-engine/src/main/java/io/confluent/ksql/engine/RuntimeAssignor.java
+++ b/ksqldb-engine/src/main/java/io/confluent/ksql/engine/RuntimeAssignor.java
@@ -18,7 +18,7 @@ import java.util.stream.Collectors;
 public class RuntimeAssignor {
   private final Map<String, Collection<SourceName>> runtimesToSources;
 
-  RuntimeAssignor(final KsqlConfig config) {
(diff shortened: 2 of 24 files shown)


wcarlson5:
Wow this really got big. Its mostly test changes though. I am still working on my testing for the `RuntimeAssignor` but I think this is ready for a review of the concept.

ksqldb-engine/src/main/java/io/confluent/ksql/engine/RuntimeAssignor.java
```diff
@@ -0,0 +1,124 @@
...
+      this.runtimesToSources.put(runtime.getKey(), new HashSet<>(runtime.getValue()));
+    }
+    this.idToRuntime.putAll(other.idToRuntime);
+  }
+
+  public RuntimeAssignor createSandbox() {
+    return new RuntimeAssignor(this);
+  }
+
+  public String getRuntime(final QueryId queryId,
```
rodesai, line 63: `getX` often implies that the method is read-only. Consider using a name like `getAndMaybeAddRuntimeForQuery`

ksqldb-engine/src/main/java/io/confluent/ksql/engine/RuntimeAssignor.java
```diff
@@ -0,0 +1,124 @@
...
+                           final Collection<SourceName> sources,
+                           final KsqlConfig config) {
+    if (idToRuntime.containsKey(queryId)) {
+      return idToRuntime.get(queryId);
+    }
+    String runtime;
```
rodesai, line 69: you can make this final

ksqldb-engine/src/main/java/io/confluent/ksql/engine/RuntimeAssignor.java
```diff
@@ -0,0 +1,124 @@
...
+    final List<String> possibleRuntimes = runtimesToSources.entrySet()
+        .stream()
+        .filter(t -> t.getValue().stream().noneMatch(sources::contains))
+        .map(Map.Entry::getKey)
+        .collect(Collectors.toList());
+    if (possibleRuntimes.isEmpty()) {
```
rodesai, line 75: this chunk of code might be cleaner like:
```
final String runtime;
if (possibleRuntimes.isEmpty()) {
    runtime = makeNewRuntime(config);
} else {
    runtime = possibleRuntimes.get(Math.abs(queryId.hashCode()) % possibleRuntimes.size());
}
runtimesToSources.get(runtime).addAll(sources);
idToRuntime.put(queryId, runtime);
```

Additionally, this would mean you should add this to `makeNewRuntime`:
```
runtimesToSources.put(runtime, new HashSet<>());
```

ksqldb-engine/src/main/java/io/confluent/ksql/engine/RuntimeAssignor.java
```diff
@@ -0,0 +1,124 @@
...
+  }
+
+
+  private String makeNewRuntime(final KsqlConfig config) {
+    final String runtime = buildSharedRuntimeId(config, true, runtimesToSources.size());
+    runtimesToSources.put(runtime, new HashSet<>());
+    return runtime;
+  }
+
+  public void rebuildAssignment(final Collection<PersistentQueryMetadata> queries) {
```
rodesai, line 102: what's this for? it's only used in tests. Can we clean it up?

ksqldb-engine/src/main/java/io/confluent/ksql/query/QueryBuilder.java
```diff
@@ -521,43 +521,44 @@ public NamedTopology getNamedTopology(final SharedKafkaStreamsRuntime sharedRunt
...
+          final QueryId queryId,
           final MetricCollectors metricCollectors) {
     for (final SharedKafkaStreamsRuntime sharedKafkaStreamsRuntime : streams) {
-      if (sharedKafkaStreamsRuntime.getSources().stream().noneMatch(sources::contains)
-          || sharedKafkaStreamsRuntime.getQueries().contains(queryID)) {
-        sharedKafkaStreamsRuntime.markSources(queryID, sources);
+      if (sharedKafkaStreamsRuntime.getApplicationId().equals(applicationId)
+          || (sharedKafkaStreamsRuntime.getApplicationId().equals(applicationId + "-validation")
+          && !real)) {
+        sharedKafkaStreamsRuntime.markSources(queryId, sources);
```
rodesai, line 532: why do we need to bookkeep sources in `SharedKafkaStreamsRuntime` now?

ksqldb-engine/src/test/java/io/confluent/ksql/engine/RuntimeAssignorTest.java
```diff
@@ -0,0 +1,100 @@
...
+        sources1,
+        KSQL_CONFIG
+    );
+    when(queryMetadata.getQueryApplicationId()).thenReturn(firstRuntime);
+    when(queryMetadata.getQueryId()).thenReturn(query1);
+    when(queryMetadata.getSourceNames()).thenReturn(new HashSet<>(sources1));
+  }
+
+  @Test
+  public void shouldCreateSandboxAndDroppingAQueryWillNotChangeReal() {
```
rodesai, line 51: add similar tests for adding a new runtime, or adding a query to an existing runtime

ksqldb-engine/src/test/java/io/confluent/ksql/engine/RuntimeAssignorTest.java
```diff
@@ -0,0 +1,100 @@
...
+    final String runtime = runtimeAssignor.getRuntime(
+        query2,
+        sources2,
+        KSQL_CONFIG
+    );
+    assertThat(runtime, not(equalTo(firstRuntime)));
+  }
+
+  @Test
+  public void shouldDropQueryThenUseSameRuntimeForTheSameSources() {
```
rodesai, line 79: orthogonal question: when we do this, do we make sure to reset offsets in the consumer in streams?

ksqldb-engine/src/test/java/io/confluent/ksql/engine/RuntimeAssignorTest.java
```diff
@@ -0,0 +1,100 @@
...
+  }
+
+  @Test
+  public void shouldCreateSandboxAndDroppingAQueryWillNotChangeReal() {
+    final RuntimeAssignor sandbox = runtimeAssignor.createSandbox();
+    sandbox.dropQuery(queryMetadata);
+    assertThat("Was changed by sandbox.", runtimeAssignor.getIdToRuntime().containsKey(query1));
+    assertThat("The query was not removed.", !sandbox.getIdToRuntime().containsKey(query1));
+  }
+
```
rodesai, line 57: add a test that verifies we use the same runtime if the sources are different, and a test that verifies that we use a different runtime if the sources are the same.

ksqldb-common/src/main/java/io/confluent/ksql/util/KsqlConfig.java
```diff
@@ -512,7 +512,7 @@
           + "CREATE SOURCE [TABLE|STREAM] statements will continue being read-only.";
 
   public static final String KSQL_SHARED_RUNTIME_ENABLED = "ksql.runtime.feature.shared.enabled";
-  public static final Boolean KSQL_SHARED_RUNTIME_ENABLED_DEFAULT = false;
+  public static final Boolean KSQL_SHARED_RUNTIME_ENABLED_DEFAULT = true;
```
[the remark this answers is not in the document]
wcarlson5: I just had this changed for testing. I will remove it before merging

ksqldb-engine/src/main/java/io/confluent/ksql/engine/RuntimeAssignor.java
```diff
@@ -0,0 +1,124 @@
...
+    idToRuntime = new HashMap<>();
+    for (int i = 0; i < runtimes; i++) {
+      final String runtime = buildSharedRuntimeId(config, true, i);
+      runtimesToSources.put(runtime, new HashSet<>());
+    }
+  }
+
+  private RuntimeAssignor(final RuntimeAssignor other) {
+    this.runtimesToSources = new HashMap<>();
+    this.idToRuntime = new HashMap<>();
```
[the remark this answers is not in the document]
wcarlson5: good call

ksqldb-engine/src/main/java/io/confluent/ksql/engine/RuntimeAssignor.java
wcarlson5: I guess I can

ksqldb-engine/src/main/java/io/confluent/ksql/engine/RuntimeAssignor.java
wcarlson5: Yes, that is a better name

ksqldb-engine/src/main/java/io/confluent/ksql/engine/RuntimeAssignor.java
```diff
@@ -0,0 +1,124 @@
...
+    for (PersistentQueryMetadata queryMetadata: queries) {
+      if (queryMetadata instanceof BinPackedPersistentQueryMetadataImpl) {
+        runtimesToSources.put(queryMetadata.getQueryApplicationId(),
+            queryMetadata.getSourceNames());
+        idToRuntime.put(queryMetadata.getQueryId(), queryMetadata.getQueryApplicationId());
+      }
+    }
+  }
+
+  public void close() {
```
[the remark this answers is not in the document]
wcarlson5: We probably don't

ksqldb-engine/src/main/java/io/confluent/ksql/engine/RuntimeAssignor.java
wcarlson5: Right now it is used only in tests but when we restore from the cmd topic we need to rebuild the assignor. We can do that from the list of queries that are running. I need to actually do this. I have a follow up https://confluentinc.atlassian.net/browse/KCI-1319

ksqldb-engine/src/test/java/io/confluent/ksql/engine/RuntimeAssignorTest.java
wcarlson5: Yes, we have a reset offset option when removing a topology. We always reset the offsets except when doing a create or replace

ksqldb-engine/src/main/java/io/confluent/ksql/engine/RuntimeAssignor.java
wcarlson5: That is cleaner. I already have line in makeNewRuntime

ksqldb-engine/src/main/java/io/confluent/ksql/query/QueryBuilder.java
wcarlson5: We might not now. I would like to keep doing it for now because it should help with the restore. If I find a better way when doing that I can remove it then

ksqldb-engine/src/test/java/io/confluent/ksql/engine/RuntimeAssignorTest.java
wcarlson5: `shouldCreateSandboxAndAddingAQueryAndRuntimeWillNotChangeReal` and `shouldCreateSandboxAndAddingAQueryWillNotChangeReal`

ksqldb-engine/src/test/java/io/confluent/ksql/engine/RuntimeAssignorTest.java
wcarlson5: `shouldAddingAQueryWithDifferentSourcesWillNotAddARuntime` and `shouldAddingAQueryWithSameSourcesWillAddARuntime`

wcarlson5 (commented):
Thanks for the review @rodesai . I got to all your comments and they made sense. The only one that I really had a comment on was about the keeping track of sources in the runtime as well. explained more in the comment but I would like to keep that for now

ksqldb-engine/src/main/java/io/confluent/ksql/query/QueryBuilder.java
```diff
@@ -521,43 +521,44 @@ public NamedTopology getNamedTopology(final SharedKafkaStreamsRuntime sharedRunt
   }
 
   private SharedKafkaStreamsRuntime getKafkaStreamsInstance(
+          final String applicationId,
           final Set<SourceName> sources,
-          final QueryId queryID,
+          final QueryId queryId,
           final MetricCollectors metricCollectors) {
     for (final SharedKafkaStreamsRuntime sharedKafkaStreamsRuntime : streams) {
```
ableegoldman, line 528: Why do we need to track the `streams` again in the `QueryBuilder` class if we only ever need to access a single SharedKafkaStreamsRuntime per query (and there's one QueryBuilder per query, IIUC)? It seems like we could remove that field entirely, as well as the `real` flag, and then just look up the correct runtime inside QueryRegistryImpl and pass it in as a parameter to the QueryBuilder

(Also, instead of looping through all of the streams can't we use a map keyed by the application/runtime id?)

ksqldb-engine/src/main/java/io/confluent/ksql/util/PersistentQueryMetadata.java
```diff
@@ -58,6 +58,8 @@
 
   void stop();
 
+  void stop(boolean resetOffsets);
```
ableegoldman, line 61: I think we should give this a default implementation here that just invokes  `stop()` and then only override the `stop(boolean resetOffsets)` for the binpacked queries to avoid leaking into the other query types too much

ableegoldman (approved):
LGTM, please address my comments in a followup PR but feel free to merge for now

ksqldb-engine/src/main/java/io/confluent/ksql/engine/RuntimeAssignor.java
```diff
@@ -0,0 +1,117 @@
...
+
+public class RuntimeAssignor {
+  private final Map<String, Collection<SourceName>> runtimesToSources;
+  private final Map<QueryId, String> idToRuntime;
+
+  public RuntimeAssignor(final KsqlConfig config) {
+    this(config, 8);
+  }
+
+  public RuntimeAssignor(final KsqlConfig config, final int runtimes) {
```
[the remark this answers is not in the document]
wcarlson5: https://github.com/confluentinc/ksql/pull/8525

ksqldb-engine/src/main/java/io/confluent/ksql/engine/RuntimeAssignor.java
```diff
@@ -0,0 +1,117 @@
...
+import java.util.HashSet;
+import java.util.List;
+import java.util.Map;
+import java.util.stream.Collectors;
+
+public class RuntimeAssignor {
+  private final Map<String, Collection<SourceName>> runtimesToSources;
+  private final Map<QueryId, String> idToRuntime;
+
+  public RuntimeAssignor(final KsqlConfig config) {
```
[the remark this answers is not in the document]
wcarlson5: https://github.com/confluentinc/ksql/pull/8525

PR #8473, opened 2021-12-08
chore: clarify reasoning behind preventing live updates to `processing.guarantee` config

ableegoldman:
We decided to prevent users from modifying the server-level `processing.guarantee` config unless there are no persistent queries running at the time which would be affected by this change, to avoid mixed processing of exactly-once and at-least-once semantics. We should make sure to state clearly the reasoning behind this

### DIFF
--- a/ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/computation/ValidatedCommandFactory.java
+++ b/ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/computation/ValidatedCommandFactory.java
@@ -37,8 +37,14 @@ import io.confluent.ksql.util.KsqlConstants;
 import io.confluent.ksql.util.KsqlServerException;
 import io.confluent.ksql.util.KsqlStatementException;
 import io.confluent.ksql.util.PersistentQueryMetadata;
+import io.confluent.ksql.util.QueryMetadata;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import java.util.Collection;
 import java.util.Objects;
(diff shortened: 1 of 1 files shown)


rodesai (approved):
LGTM

ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/computation/ValidatedCommandFactory.java
```diff
@@ -144,9 +152,19 @@ private static Command createForAlterSystemQuery(
...
+          context.getPersistentQueries()
+              .stream()
+              .map(QueryMetadata::getQueryId)
+              .collect(Collectors.toList());
+      LOG.error("Failed to set {} to {} due to the {} persistent queries currently running: {}",
+                propertyName, propertyValue, runningQueries.size(), runningQueries);
       throw new ConfigException(
-          String.format("Failed to set %s to %s. Please terminate all running queries "
-              + "before attempting to set %s", propertyName, propertyValue, propertyName), null);
+          String.format("Unable to set %s to %s due to existing persistent queries. The %s may"
```
derekjn, line 163: I think this can be consolidated a bit to something along the lines of:

> ${propertyName} may not be changed for running persistent queries which have already processed data using a different ${propertyName}. To modify ${propertyName} you must first terminate all running persistent queries.

Or a slightly different version:

> ${propertyName} may not be changed for running persistent queries which are already processing data using a different ${propertyName}. To modify ${propertyName} you must first terminate all running persistent queries.

PR #8474, opened 2021-12-08
fix: use correct runtime to build copy of query's NamedTopology

ableegoldman:
When building up a copy of a query's NamedTopology for use when restarting the runtime or validating a new query, we need to make sure to use a NamedTopologyBuilder that corresponds to the correct/new runtime. Otherwise the copied topology will still be tied to the original topology and its runtime.

See https://github.com/confluentinc/ksql/pull/8434/files#r764394827

### DIFF
--- a/ksqldb-engine/src/main/java/io/confluent/ksql/query/QueryBuilder.java
+++ b/ksqldb-engine/src/main/java/io/confluent/ksql/query/QueryBuilder.java
@@ -477,28 +477,19 @@ final class QueryBuilder {
         classifier,
         queryOverrides,
         scalablePushRegistry,
-        () -> getNamedTopology(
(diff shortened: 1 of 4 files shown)


wcarlson5 (approved):
LGTM when the build passes

ableegoldman:
One test failure but I'd definitely call it unrelated: PullQueryLimitHARoutingTest.shouldReturnLimitRowsMultiHostSetupTable
```Stacktrace
java.lang.AssertionError: 

Expected: a collection with size <301>
     but: collection size was <302>
	at io.confluent.ksql.rest.integration.PullQueryLimitHARoutingTest.shouldReturnLimitRowsMultiHostSetupTable(PullQueryLimitHARoutingTest.java:291)
```

Will merge

PR #8475, opened 2021-12-08
feat: indicate header columns in  source descriptions

jzaralim:
### Description 
`DESCRIBE` statements will indicate which columns are headers, for example

```
 Field  | Type                                                      
--------------------------------------------------------------------
 ROWKEY | VARCHAR(STRING)  (primary key)                            
 f_0    | BOOLEAN                                                   
 f_1    | INTEGER                                                   
 f_2    | BIGINT                                                    
 f_3    | DOUBLE                                                    
 f_4    | VARCHAR(STRING)                                           
 f_5    | ARRAY<VARCHAR(STRING)>                                    
 f_6    | MAP<STRING, BIGINT>                                       
 f_7    | STRUCT<a DOUBLE>                                          
 HEAD   | ARRAY<STRUCT<KEY VARCHAR(STRING), VALUE BYTES>> (headers) 
--------------------------------------------------------------------

```
### Testing done 
Manual, unit, console tests
### Reviewer checklist

### DIFF
--- a/ksqldb-cli/src/main/java/io/confluent/ksql/cli/console/Console.java
+++ b/ksqldb-cli/src/main/java/io/confluent/ksql/cli/console/Console.java
@@ -503,6 +503,16 @@ public class Console implements Closeable {
   ) {
     final FieldType possibleFieldType = field.getType().orElse(null);
 
+    if (possibleFieldType == FieldType.HEADER) {
+      final String headerType;
+      if (field.getHeaderKey().isPresent()) {
+        headerType = "(header('" + field.getHeaderKey().get() + "'))";
+      } else {
+        headerType = "(headers)";
+      }
+      return String.format("%-16s %s", field.getSchema().toTypeString(), headerType);
+    }
+
     if (possibleFieldType == FieldType.KEY) {
       final String wt = windowType
           .map(v -> " (Window type: " + v + ")")
--- a/ksqldb-cli/src/main/java/io/confluent/ksql/cli/console/Console.java
+++ b/ksqldb-cli/src/main/java/io/confluent/ksql/cli/console/Console.java
@@ -504,12 +504,9 @@ public class Console implements Closeable {
     final FieldType possibleFieldType = field.getType().orElse(null);
 
     if (possibleFieldType == FieldType.HEADER) {
-      final String headerType;
-      if (field.getHeaderKey().isPresent()) {
-        headerType = "(header('" + field.getHeaderKey().get() + "'))";
-      } else {
-        headerType = "(headers)";
-      }
+      final String headerType = field.getHeaderKey()
+          .map(k -> "(header('" + k + "'))")
+          .orElse("(headers)");
       return String.format("%-16s %s", field.getSchema().toTypeString(), headerType);
     }
 
--- a/ksqldb-cli/src/test/java/io/confluent/ksql/cli/console/ConsoleTest.testPrintSourceDescriptionWithExtractedHeader.approved.tabular
+++ b/ksqldb-cli/src/test/java/io/confluent/ksql/cli/console/ConsoleTest.testPrintSourceDescriptionWithExtractedHeader.approved.tabular
@@ -1,0 +1,43 @@
+
+Name                 : TestSource
+Type                 : TABLE
+Timestamp field      : 2000-01-01
+Key format           : kafka
+Value format         : avro
+Kafka topic          : kafka-topic (partitions: 1, replication: 1)
+Statement            : sql statement
+
+ Field  | Type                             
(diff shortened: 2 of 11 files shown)


ksqldb-rest-model/src/main/java/io/confluent/ksql/rest/entity/FieldInfo.java
```diff
@@ -31,22 +31,26 @@
 
   public enum FieldType {
     SYSTEM, // To be removed in the future. 0.9 saw this value no longer used.
```
spena, line 33: This can be removed, right? I don't see being used either.

ksqldb-rest-model/src/main/java/io/confluent/ksql/rest/entity/FieldInfo.java
```diff
@@ -31,22 +31,26 @@
...
   private final Optional<FieldType> type;
+  private final Optional<String> headerKey;
 
   @JsonCreator
   public FieldInfo(
       @JsonProperty(value = "name", required = true) final String name,
       @JsonProperty(value = "schema", required = true) final SchemaInfo schema,
-      @JsonProperty("fieldType") final Optional<FieldType> type
+      @JsonProperty("fieldType") final Optional<FieldType> type,
+      @JsonProperty("headerKey") final Optional<String> headerKey
```
spena, line 48: Seems this field is optional in different places. What if you create another constructor that defaults to Optional.empty()?

ksqldb-cli/src/test/java/io/confluent/ksql/cli/console/ConsoleTest.testPrintSourceDescriptionWithExtractedHeader.approved.tabular
```diff
@@ -0,0 +1,43 @@
...
+ f_4    | VARCHAR(STRING)                  
+ f_5    | ARRAY<VARCHAR(STRING)>           
+ f_6    | MAP<STRING, BIGINT>              
+ f_7    | STRUCT<a DOUBLE>                 
+ HEAD   | BYTES            (header('abc')) 
+-------------------------------------------
+
+Queries that read from this TABLE
+-----------------------------------
+readId (ERROR) : read query
```
spena, line 26: is this ERROR ok?

ksqldb-cli/src/main/java/io/confluent/ksql/cli/console/Console.java
```diff
@@ -503,6 +503,16 @@ private static String formatFieldType(
   ) {
     final FieldType possibleFieldType = field.getType().orElse(null);
 
+    if (possibleFieldType == FieldType.HEADER) {
+      final String headerType;
+      if (field.getHeaderKey().isPresent()) {
+        headerType = "(header('" + field.getHeaderKey().get() + "'))";
+      } else {
+        headerType = "(headers)";
+      }
```
spena, line 512: Tip

```
final String headerType = field.getHeaderKey()
  .map(k -> "(header('" + k + "'))")
  .orElse(() -> "(headers)");
```

ksqldb-rest-model/src/main/java/io/confluent/ksql/rest/entity/FieldInfo.java
jzaralim: The purpose of keeping this was so that pre-0.9 servers could still work with newer CLIs. I'm not sure how many people are still using 0.9, but I think this is easy enough to just keep anyway. This does raise a good point though - the changes in this PR would make HEADERS require the newest CLI. Are we okay with that?

ksqldb-rest-model/src/main/java/io/confluent/ksql/rest/entity/FieldInfo.java
jzaralim: Done

ksqldb-cli/src/main/java/io/confluent/ksql/cli/console/Console.java
jzaralim: Done

ksqldb-cli/src/test/java/io/confluent/ksql/cli/console/ConsoleTest.testPrintSourceDescriptionWithExtractedHeader.approved.tabular
jzaralim: Yeah, all the source description tests have errors to test printing errors

spena (approved):

PR #8476, opened 2021-12-08
feat: custom auth configs for ksql connector requests

vcrfxia:
### Description 

This PR broadens the types of Connect deployments that ksqlDB's connector integration can work with. Specifically, this PR:
* adds new configs for specifying custom HTTP basic auth credentials to be sent with ksql connector requests. If not specified, then ksqlDB will continue to forward the incoming request auth header as it does today.
* adds support for automatically reloading file-based basic auth credentials (see above), if configured. This functionality is off by default in order to avoid the background thread that monitors for file changes.
* adds a configuration to specify whether missing/malformed credentials (see above) should result in an error or not. By default, an error will be thrown, but this PR adds a config to control this behavior to support use cases where, e.g., the credentials file may not exist at first but will be created after the ksqlDB server has already started.
* switches the Connect API paths to be relative rather than absolute, to support integrating with Connect deployments where the raw endpoints are nested under other paths.

### Testing done 

Unit tests + manual.

### Reviewer checklist

### DIFF
--- a/ksqldb-common/src/main/java/io/confluent/ksql/util/KsqlConfig.java
+++ b/ksqldb-common/src/main/java/io/confluent/ksql/util/KsqlConfig.java
@@ -811,7 +811,8 @@ public class KsqlConfig extends AbstractConfig {
             false,
             ConfigDef.Importance.LOW,
             "If true, basic auth credentials for connector auth will automatically reload "
-                + "on file change."
+                + "on file change (creation or modification). File deletion is not monitored and "
+                + "old credentials will continue to be used in this case."
         ).define(
             CONNECT_BASIC_AUTH_FAIL_ON_UNREADABLE_CREDENTIALS,
             ConfigDef.Type.BOOLEAN,
--- a/ksqldb-common/src/main/java/io/confluent/ksql/util/KsqlConfig.java
+++ b/ksqldb-common/src/main/java/io/confluent/ksql/util/KsqlConfig.java
@@ -91,9 +91,31 @@ public class KsqlConfig extends AbstractConfig {
 
   public static final String SCHEMA_REGISTRY_URL_PROPERTY = "ksql.schema.registry.url";
 
-  public static final String CONNECT_URL_PROPERTY = "ksql.connect.url";
+  public static final String KSQL_CONNECT_PREFIX = "ksql.connect.";
 
-  public static final String CONNECT_WORKER_CONFIG_FILE_PROPERTY = "ksql.connect.worker.config";
+  public static final String CONNECT_URL_PROPERTY = KSQL_CONNECT_PREFIX + "url";
+
+  public static final String CONNECT_WORKER_CONFIG_FILE_PROPERTY =
+      KSQL_CONNECT_PREFIX + "worker.config";
+
+  public static final String CONNECT_BASIC_AUTH_CREDENTIALS_SOURCE_PROPERTY =
+      KSQL_CONNECT_PREFIX + "basic.auth.credentials.source";
+  public static final String BASIC_AUTH_CREDENTIALS_SOURCE_NONE = "NONE";
+  public static final String BASIC_AUTH_CREDENTIALS_SOURCE_FILE = "FILE";
+  private static final ConfigDef.ValidString BASIC_AUTH_CREDENTIALS_SOURCE_VALIDATOR =
+      ConfigDef.ValidString.in(
+          BASIC_AUTH_CREDENTIALS_SOURCE_NONE,
+          BASIC_AUTH_CREDENTIALS_SOURCE_FILE
+      );
+  public static final String BASIC_AUTH_CREDENTIALS_USERNAME = "username";
+  public static final String BASIC_AUTH_CREDENTIALS_PASSWORD = "password";
+
+  public static final String CONNECT_BASIC_AUTH_CREDENTIALS_FILE_PROPERTY =
+      KSQL_CONNECT_PREFIX + "basic.auth.credentials.file";
+  public static final String CONNECT_BASIC_AUTH_CREDENTIALS_RELOAD_PROPERTY =
+      KSQL_CONNECT_PREFIX + "basic.auth.credentials.reload";
+  public static final String CONNECT_BASIC_AUTH_FAIL_ON_UNREADABLE_CREDENTIALS =
+      KSQL_CONNECT_PREFIX + "basic.auth.credentials.fail.on.unreadable";
 
   public static final String KSQL_ENABLE_UDFS = "ksql.udfs.enabled";
 
@@ -764,6 +786,39 @@ public class KsqlConfig extends AbstractConfig {
                 + "will prevent connect from starting up embedded within KSQL. For more information"
                 + " on configuring connect, see "
                 + "https://docs.confluent.io/current/connect/userguide.html#configuring-workers."
+        ).define(
+            CONNECT_BASIC_AUTH_CREDENTIALS_SOURCE_PROPERTY,
+            ConfigDef.Type.STRING,
+            BASIC_AUTH_CREDENTIALS_SOURCE_NONE,
+            BASIC_AUTH_CREDENTIALS_SOURCE_VALIDATOR,
+            Importance.LOW,
+            "If providing explicit basic auth credentials for ksqlDB to use when sending connector "
+                + "requests, this config specifies how credentials should be loaded. Valid "
+                + "options are 'FILE' in order to specify the username and password in a "
+                + "properties file, or 'NONE' to indicate that custom basic auth should "
+                + "not be used. If 'NONE', ksqlDB will forward the auth header, if present, "
+                + "on the incoming ksql request to Connect."
+        ).define(
+            CONNECT_BASIC_AUTH_CREDENTIALS_FILE_PROPERTY,
+            ConfigDef.Type.STRING,
+            "",
+            Importance.LOW,
+            "If '" + CONNECT_BASIC_AUTH_CREDENTIALS_SOURCE_PROPERTY + "' is set to 'FILE', "
+                + "then this config specifies the path to the credentials file."
+        ).define(
+            CONNECT_BASIC_AUTH_CREDENTIALS_RELOAD_PROPERTY,
+            ConfigDef.Type.BOOLEAN,
+            false,
+            ConfigDef.Importance.LOW,
+            "If true, basic auth credentials for connector auth will automatically reload "
+                + "on file change."
+        ).define(
+            CONNECT_BASIC_AUTH_FAIL_ON_UNREADABLE_CREDENTIALS,
+            ConfigDef.Type.BOOLEAN,
+            true,
+            ConfigDef.Importance.LOW,
+            "If true, failure to load basic auth credentials for connector auth will result "
+                + "in an error. If false, failure will result in a warn log and empty credentials."
         ).define(
             KSQL_ENABLE_UDFS,
             ConfigDef.Type.BOOLEAN,
--- a/ksqldb-engine/src/main/java/io/confluent/ksql/services/DefaultConnectClientFactory.java
+++ b/ksqldb-engine/src/main/java/io/confluent/ksql/services/DefaultConnectClientFactory.java
@@ -1,0 +1,140 @@
+/*
+ * Copyright 2021 Confluent Inc.
+ *
+ * Licensed under the Confluent Community License (the "License"); you may not use
+ * this file except in compliance with the License.  You may obtain a copy of the
+ * License at
+ *
+ * http://www.confluent.io/confluent-community-license
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OF ANY KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations under the License.
+ */
+
+package io.confluent.ksql.services;
+
+import io.confluent.ksql.services.ConnectClient.ConnectClientFactory;
+import io.confluent.ksql.util.FileWatcher;
+import io.confluent.ksql.util.KsqlConfig;
+import java.io.FileInputStream;
+import java.io.IOException;
+import java.nio.charset.Charset;
+import java.nio.file.Paths;
+import java.util.Base64;
+import java.util.Objects;
+import java.util.Optional;
+import java.util.Properties;
+import org.apache.kafka.common.config.ConfigException;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class DefaultConnectClientFactory implements ConnectClientFactory {
+
+  private static final Logger log = LoggerFactory.getLogger(DefaultConnectClientFactory.class);
+
+  private final KsqlConfig ksqlConfig;
+  private Optional<String> connectAuthHeader;
+  private FileWatcher credentialsFileWatcher;
+
+  public DefaultConnectClientFactory(
+      final KsqlConfig ksqlConfig
+  ) {
+    this.ksqlConfig = Objects.requireNonNull(ksqlConfig, "ksqlConfig");
+  }
+
+  @Override
+  public DefaultConnectClient get(final Optional<String> ksqlAuthHeader) {
+    if (connectAuthHeader == null) {
+      connectAuthHeader = buildAuthHeader();
+    }
+
+    return new DefaultConnectClient(
+        ksqlConfig.getString(KsqlConfig.CONNECT_URL_PROPERTY),
+        // if no explicit header specified, then forward incoming request header
+        connectAuthHeader.isPresent() ? connectAuthHeader : ksqlAuthHeader
+    );
+  }
+
+  @Override
+  public void close() {
+    if (credentialsFileWatcher != null) {
+      credentialsFileWatcher.shutdown();
+    }
+  }
+
+  private Optional<String> buildAuthHeader() {
+    // custom basic auth credentials
+    if (ksqlConfig.getString(KsqlConfig.CONNECT_BASIC_AUTH_CREDENTIALS_SOURCE_PROPERTY)
+        .equalsIgnoreCase(KsqlConfig.BASIC_AUTH_CREDENTIALS_SOURCE_FILE)) {
+      final String credentialsFile =
+          ksqlConfig.getString(KsqlConfig.CONNECT_BASIC_AUTH_CREDENTIALS_FILE_PROPERTY);
+      final boolean failOnUnreadableCreds =
+          ksqlConfig.getBoolean(KsqlConfig.CONNECT_BASIC_AUTH_FAIL_ON_UNREADABLE_CREDENTIALS);
+
+      if (ksqlConfig.getBoolean(KsqlConfig.CONNECT_BASIC_AUTH_CREDENTIALS_RELOAD_PROPERTY)) {
+        startBasicAuthFileWatcher(credentialsFile, failOnUnreadableCreds);
+      }
+
+      return buildBasicAuthHeader(credentialsFile, failOnUnreadableCreds);
+    } else {
+      return Optional.empty();
+    }
+  }
+
+  private void startBasicAuthFileWatcher(
+      final String filePath,
+      final boolean failOnUnreadableCreds
+  ) {
+    try {
+      credentialsFileWatcher = new FileWatcher(Paths.get(filePath), () -> {
+        connectAuthHeader = buildBasicAuthHeader(filePath, failOnUnreadableCreds);
+      });
+      credentialsFileWatcher.start();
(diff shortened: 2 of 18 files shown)


ksqldb-engine/src/main/java/io/confluent/ksql/services/DefaultConnectClientFactory.java
```diff
@@ -0,0 +1,140 @@
...
+import org.apache.kafka.common.config.ConfigException;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class DefaultConnectClientFactory implements ConnectClientFactory {
+
+  private static final Logger log = LoggerFactory.getLogger(DefaultConnectClientFactory.class);
+
+  private final KsqlConfig ksqlConfig;
+  private Optional<String> connectAuthHeader;
```
Gerrrr, line 38: This field is modified and accessed from different threads, so we should make it `volatile`. FWIW we can get away with `volatile` here rather than atomic reference because we always update the field unconditionally to its previous state.

ksqldb-engine/src/main/java/io/confluent/ksql/services/DefaultConnectClientFactory.java
```diff
@@ -0,0 +1,140 @@
...
+
+  public DefaultConnectClientFactory(
+      final KsqlConfig ksqlConfig
+  ) {
+    this.ksqlConfig = Objects.requireNonNull(ksqlConfig, "ksqlConfig");
+  }
+
+  @Override
+  public DefaultConnectClient get(final Optional<String> ksqlAuthHeader) {
+    if (connectAuthHeader == null) {
```
Gerrrr, line 49: I would probably modify this class such that the header and the watcher are created in the constructor, and we get rid of this `if`.  Right now, this class is not thread-safe in a way that allows starting multiple watchers, then losing references to all the old watchers, probably other phenomena as well.

ksqldb-engine/src/main/java/io/confluent/ksql/services/DefaultConnectClientFactory.java
```diff
@@ -0,0 +1,140 @@
...
+      return Optional.empty();
+    }
+  }
+
+  private void startBasicAuthFileWatcher(
+      final String filePath,
+      final boolean failOnUnreadableCreds
+  ) {
+    try {
+      credentialsFileWatcher = new FileWatcher(Paths.get(filePath), () -> {
```
Gerrrr, line 91: AFAIU the file watcher does not notify on the file deletion, so if the credentials file is removed, we'll still use the most recent header. Is this OK for our purposes? An alternative would be to erase the header on file deletion. I'd probably add a short comment addressing this scenario in any case.

ksqldb-engine/src/test/java/io/confluent/ksql/services/DefaultConnectClientFactoryTest.java
```diff
@@ -0,0 +1,206 @@
...
+import org.junit.Before;
+import org.junit.Rule;
+import org.junit.Test;
+import org.junit.rules.TemporaryFolder;
+import org.junit.runner.RunWith;
+import org.mockito.Mock;
+import org.mockito.junit.MockitoJUnitRunner;
+
+@RunWith(MockitoJUnitRunner.class)
+public class DefaultConnectClientFactoryTest {
```
Gerrrr, line 44: If it is not too much trouble, I'd probably test that credentials are reloaded on file change.

ksqldb-rest-app/src/test/java/io/confluent/ksql/api/impl/DefaultKsqlSecurityContextProviderTest.java
```diff
@@ -74,15 +77,16 @@ public void setup() {
         userServiceContextFactory,
         ksqlConfig,
         () -> schemaRegistryClientFactory,
+        (authHeader) -> connectClient, // TODO: fix
```
Gerrrr, line 80: nit: redundant TODO?

ksqldb-engine/src/main/java/io/confluent/ksql/services/DefaultConnectClientFactory.java
vcrfxia: We can't create the header and watcher in the constructor -- the DefaultConnectClientFactory constructor is called every time a service context is created, i.e., on every ksql request, regardless of whether it's a connector request or not. We don't want to load credentials (and potentially start the watcher) on every request; we only want to do this if a connector request is issued. 

I can synchronize this `get()` method in order to address the concurrency concern you raised -- `get()` should only ever be called from a single thread anyway.

ksqldb-engine/src/main/java/io/confluent/ksql/services/DefaultConnectClientFactory.java
vcrfxia: Yes, this is fine for our purposes. I considered updating the FileWatcher implementation to watch for this but it seems preferable to continue using outdated credentials/TLS certs rather than no credentials/certs in the event of file deletion. I will add some comments as you suggested.

ksqldb-engine/src/test/java/io/confluent/ksql/services/DefaultConnectClientFactoryTest.java
vcrfxia: I didn't add one initially since I thought it'd have to be an integration test, rather than a unit test, which would be a fair bit of work but now that I'm thinking about it again I think we can get away with a unit test. If so, I'll add one. If not, I'd rather not block on this. (We already have plans for integration tests for this functionality elsewhere where the configs will be exercised.)

ksqldb-engine/src/main/java/io/confluent/ksql/services/DefaultConnectClientFactory.java
Gerrrr: Makes sense! FWIW we should also synchronize `close`.

ksqldb-engine/src/test/java/io/confluent/ksql/services/DefaultConnectClientFactoryTest.java
vcrfxia: Added unit tests for reload in the latest commit.

ksqldb-engine/src/main/java/io/confluent/ksql/services/DefaultConnectClientFactory.java
```diff
@@ -0,0 +1,150 @@
...
+        connectAuthHeader = buildBasicAuthHeader(filePath, failOnUnreadableCreds);
+      });
+      credentialsFileWatcher.start();
+      log.info("Enabled automatic connector credentials reload for location: " + filePath);
+    } catch (java.io.IOException e) {
+      log.error("Failed to enable automatic connector credentials reload.");
```
Gerrrr, line 97: We should log the `IOException` here, otherwise we don't know why the automatic reload failed:

```java
log.error("Failed to enable automatic connector credentials reload.", e);
```

ksqldb-engine/src/main/java/io/confluent/ksql/services/DefaultConnectClientFactory.java
```diff
@@ -0,0 +1,150 @@
...
+import java.nio.file.Paths;
+import java.util.Base64;
+import java.util.Objects;
+import java.util.Optional;
+import java.util.Properties;
+import org.apache.kafka.common.config.ConfigException;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class DefaultConnectClientFactory implements ConnectClientFactory {
```
Gerrrr, line 33: State changes for `connectAuthHeader` a bit non-trivial, so I would probably add a short javadoc describing it. Something along the lines:
* `connectAuthHeader` is set on the first `get`.
* If credentials are in a file, we read them from the file AND start a file watcher thread that monitors file changes and updates the header accordingly.

Gerrrr (approved):
LGTM! I left a couple of very optional nits/suggestions.

ksqldb-common/src/main/java/io/confluent/ksql/util/KsqlConfig.java
```diff
@@ -764,6 +786,40 @@ private static ConfigDef buildConfigDef(final ConfigGeneration generation) {
...
+            "If true, basic auth credentials for connector auth will automatically reload "
+                + "on file change (creation or modification). File deletion is not monitored and "
+                + "old credentials will continue to be used in this case."
+        ).define(
+            CONNECT_BASIC_AUTH_FAIL_ON_UNREADABLE_CREDENTIALS,
+            ConfigDef.Type.BOOLEAN,
+            true,
+            ConfigDef.Importance.LOW,
+            "If true, failure to load basic auth credentials for connector auth will result "
+                + "in an error. If false, failure will result in a warn log and empty credentials."
```
spena, line 822: Why do we need this failure optional? If an error occurs with Connect, then it's a good warning to the user that Connect/KSQL settings are not well configured, right? 

Cloud users don't see the logs, just the client errors.

ksqldb-common/src/main/java/io/confluent/ksql/util/KsqlConfig.java
```diff
@@ -91,9 +91,31 @@
...
-  public static final String CONNECT_WORKER_CONFIG_FILE_PROPERTY = "ksql.connect.worker.config";
+  public static final String CONNECT_URL_PROPERTY = KSQL_CONNECT_PREFIX + "url";
+
+  public static final String CONNECT_WORKER_CONFIG_FILE_PROPERTY =
+      KSQL_CONNECT_PREFIX + "worker.config";
+
+  public static final String CONNECT_BASIC_AUTH_CREDENTIALS_SOURCE_PROPERTY =
+      KSQL_CONNECT_PREFIX + "basic.auth.credentials.source";
+  public static final String BASIC_AUTH_CREDENTIALS_SOURCE_NONE = "NONE";
+  public static final String BASIC_AUTH_CREDENTIALS_SOURCE_FILE = "FILE";
```
spena, line 104: Any other sources that you have in mind?

ksqldb-rest-app/src/main/java/io/confluent/ksql/api/server/Server.java
```diff
@@ -27,6 +27,7 @@
 import io.confluent.ksql.rest.server.KsqlRestConfig;
 import io.confluent.ksql.rest.server.state.ServerState;
 import io.confluent.ksql.security.KsqlSecurityExtension;
+import io.confluent.ksql.util.FileWatcher;
```
spena, line 30: Is this used?

ksqldb-common/src/main/java/io/confluent/ksql/util/KsqlConfig.java
```diff
@@ -91,9 +91,31 @@
...
-  public static final String CONNECT_URL_PROPERTY = "ksql.connect.url";
+  public static final String KSQL_CONNECT_PREFIX = "ksql.connect.";
 
-  public static final String CONNECT_WORKER_CONFIG_FILE_PROPERTY = "ksql.connect.worker.config";
+  public static final String CONNECT_URL_PROPERTY = KSQL_CONNECT_PREFIX + "url";
+
+  public static final String CONNECT_WORKER_CONFIG_FILE_PROPERTY =
+      KSQL_CONNECT_PREFIX + "worker.config";
+
+  public static final String CONNECT_BASIC_AUTH_CREDENTIALS_SOURCE_PROPERTY =
```
spena, line 101: Can users override these configs from the client?

ksqldb-engine/src/main/java/io/confluent/ksql/services/DefaultConnectClientFactory.java
```diff
@@ -0,0 +1,150 @@
...
+    try {
+      inputStream = new FileInputStream(credentialsPath);
+      credentials.load(inputStream);
+
+      if (credentials.containsKey(KsqlConfig.BASIC_AUTH_CREDENTIALS_USERNAME)
+          && credentials.containsKey(KsqlConfig.BASIC_AUTH_CREDENTIALS_PASSWORD)) {
+        final String userInfo = credentials.getProperty(KsqlConfig.BASIC_AUTH_CREDENTIALS_USERNAME)
+            + ":" + credentials.getProperty(KsqlConfig.BASIC_AUTH_CREDENTIALS_PASSWORD);
+        return Optional.of("Basic " + Base64.getEncoder()
+            .encodeToString(userInfo.getBytes(Charset.defaultCharset())));
```
spena, line 123: Seems we need a common class to encode user/pass. This is used in `JaasPrincipal` and also here. 
i.e.
```
// Returns a base64 encode user/password
String KsqlCredentials.basic(user, pass);   
```

ksqldb-common/src/main/java/io/confluent/ksql/util/KsqlConfig.java
vcrfxia: Not at the moment. The structure for the config is chosen to be consistent with schema registry: https://github.com/confluentinc/schema-registry/blob/7c9a7d88d9a2aba083f29423d4ae017127941662/client/src/main/java/io/confluent/kafka/schemaregistry/client/SchemaRegistryClientConfig.java#L27

ksqldb-rest-app/src/main/java/io/confluent/ksql/api/server/Server.java
vcrfxia: Yes. This line was added because I moved the package that FileWatcher was previously in.

ksqldb-common/src/main/java/io/confluent/ksql/util/KsqlConfig.java
vcrfxia: No, the credentials are loaded once upfront while starting the server (and creating the service context provider).

ksqldb-engine/src/main/java/io/confluent/ksql/services/DefaultConnectClientFactory.java
vcrfxia: Yeah, it's used all over the place. Also in RestIntegrationTestUtil.java and ClientImpl.java. Probably other places too.

It's a small enough piece of code and it doesn't look like there's an existing util file that makes sense for a helper method; that's probably why we haven't created one yet. Do you think it's worth adding a new file such as BasicAuthUtil into ksqldb-common just for this? My instinct says no but I can add one if you think it's better.

ksqldb-engine/src/main/java/io/confluent/ksql/services/DefaultConnectClientFactory.java
```diff
@@ -94,7 +111,7 @@ private void startBasicAuthFileWatcher(
       credentialsFileWatcher.start();
       log.info("Enabled automatic connector credentials reload for location: " + filePath);
     } catch (java.io.IOException e) {
-      log.error("Failed to enable automatic connector credentials reload.");
+      log.error("Failed to enable automatic connector credentials reload: " + e.getMessage());
```
Gerrrr, line 114: My 2c, feel free to ignore as I am probably splitting hairs here: it is more informative to pass an exception as the second argument to the server logs rather than concatting its message because you also get a stack trace. Having the stack trace in the logs makes debugging so much easier. Consider a (made up) example where I added a runtime exception somewhere in the along the call chain:

```java
  @Test
  public void test() {
    try {
      final KsqlEntity response = executeListVariables("LIST VARIABLES;").get();
    } catch (Exception e) {
      logger.error("logging with the second argument", e);
      logger.error("logging by concating the message: " + e.getMessage());
    }
  }
```

Test output:

```
[2021-12-13 19:23:45,956] ERROR logging with the second argument (io.confluent.ksql.rest.server.execution.ListVariablesExecutorTest:95)
java.lang.RuntimeException: :(
	at io.confluent.ksql.rest.server.execution.ListVariablesExecutor.execute(ListVariablesExecutor.java:42)
	at io.confluent.ksql.rest.server.execution.CustomExecutors.execute(CustomExecutors.java:124)
	at io.confluent.ksql.rest.server.execution.ListVariablesExecutorTest.executeListVariables(ListVariablesExecutorTest.java:55)
	at io.confluent.ksql.rest.server.execution.ListVariablesExecutorTest.test(ListVariablesExecutorTest.java:93)
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.lang.reflect.Method.invoke(Method.java:498)
	at org.junit.runners.model.FrameworkMethod$1.runReflectiveCall(FrameworkMethod.java:59)
	at org.junit.internal.runners.model.ReflectiveCallable.run(ReflectiveCallable.java:12)
	at org.junit.runners.model.FrameworkMethod.invokeExplosively(FrameworkMethod.java:56)
	at org.junit.internal.runners.statements.InvokeMethod.evaluate(InvokeMethod.java:17)
	at org.junit.internal.runners.statements.RunBefores.evaluate(RunBefores.java:26)
	at org.mockito.internal.runners.DefaultInternalRunner$1$1.evaluate(DefaultInternalRunner.java:54)
	at org.junit.runners.ParentRunner$3.evaluate(ParentRunner.java:306)
	at org.junit.runners.BlockJUnit4ClassRunner$1.evaluate(BlockJUnit4ClassRunner.java:100)
	at org.junit.runners.ParentRunner.runLeaf(ParentRunner.java:366)
	at org.junit.runners.BlockJUnit4ClassRunner.runChild(BlockJUnit4ClassRunner.java:103)
	at org.junit.runners.BlockJUnit4ClassRunner.runChild(BlockJUnit4ClassRunner.java:63)
	at org.junit.runners.ParentRunner$4.run(ParentRunner.java:331)
	at org.junit.runners.ParentRunner$1.schedule(ParentRunner.java:79)
	at org.junit.runners.ParentRunner.runChildren(ParentRunner.java:329)
	at org.junit.runners.ParentRunner.access$100(ParentRunner.java:66)
	at org.junit.runners.ParentRunner$2.evaluate(ParentRunner.java:293)
	at org.junit.runners.ParentRunner$3.evaluate(ParentRunner.java:306)
	at org.junit.runners.ParentRunner.run(ParentRunner.java:413)
	at org.mockito.internal.runners.DefaultInternalRunner$1.run(DefaultInternalRunner.java:99)
	at org.mockito.internal.runners.DefaultInternalRunner.run(DefaultInternalRunner.java:105)
	at org.mockito.internal.runners.StrictRunner.run(StrictRunner.java:40)
	at org.mockito.junit.MockitoJUnitRunner.run(MockitoJUnitRunner.java:163)
	at org.junit.runner.JUnitCore.run(JUnitCore.java:137)
	at com.intellij.junit4.JUnit4IdeaTestRunner.startRunnerWithArgs(JUnit4IdeaTestRunner.java:69)
	at com.intellij.rt.junit.IdeaTestRunner$Repeater.startRunnerWithArgs(IdeaTestRunner.java:33)
	at com.intellij.rt.junit.JUnitStarter.prepareStreamsAndStart(JUnitStarter.java:235)
	at com.intellij.rt.junit.JUnitStarter.main(JUnitStarter.java:54)
[2021-12-13 19:23:45,963] ERROR logging by concating the message: :( (io.confluent.ksql.rest.server.execution.ListVariablesExecutorTest:96)
```

Even though the first option is way more verbose and contains a bunch of entries we do not care about, the top N entries are pure gold - they show us where _exactly_ the exception has happened.

Gerrrr (approved):
LGTM! Thank you for addressing the feedback.

ksqldb-engine/src/main/java/io/confluent/ksql/services/DefaultConnectClientFactory.java
vcrfxia: Makes sense -- thanks for the tip! I will make this update in my follow-up (dependent) PR (https://github.com/confluentinc/ksql/pull/8357) rather than this PR in order to avoid waiting for another build before merging.

ksqldb-common/src/main/java/io/confluent/ksql/util/KsqlConfig.java
vcrfxia: Discussed offline -- we're going to remove this config and default the behavior to false, i.e., failure to load credentials will not throw an error from the engine. Instead, the error will be logged to the server log and the user will receive a generic 401 from the failed request, rather than a message indicating that the credentials file could be read. We think this behavior is fine from the user's POV since they can access the server logs to view the error, and we generally want to minimize the number of server configs introduced.

I'm going to make these changes in my follow-up PR (https://github.com/confluentinc/ksql/pull/8357) in order to merge this one sooner to unblock the other one.

PR #8477, opened 2021-12-09
fix: clean up transient queries when using shared runtimes

wcarlson5:
we shouldn't be trying to clean up named topologies for transient queries when shared runtimes are on. 

### Testing done 

### Reviewer checklist


lct45 (approved):
@wcarlson5 can you add a bit of context in the PR description before merging? this lgtm but I can't tell exactly what logic you're editing

PR #8480, opened 2021-12-09
feat: Allow to plug-in custom error handling for Connect server errors

Gerrrr:
### Description 

This patch enables pluggable error handling for ksqlDB <-> Connect integration. Specifically, it allows to customize behaviour on `UNAUTHORIZED` and `FORBIDDEN` server errors.

The patch introduces new configuration option -  `ksql.connect.error.handler` that allows to override `DefaultConnectServerErrors` by implementing `ConnectServerErrors` interface. The latter allows to return `Optional<KsqlEntity>` for a specific error code. Currently, the interface has custom handlers for the 2 codes mentioned above - 401 and 403 as well as a fall-back handler for other errors. In the future we can extend this interface with more handlers for specific errors.

Specifically, I made the following changes:
* `ConnectExecutor`, `DescribeConnectExecutor`, `DropConnectExecutor`, `ListConnectorPluginsExecutor`, `ListConnectorsExecutor`  have an immutable error handler as part of their state.
* For the executors mentioned in the previous point, `execute` is an instance method now.
* `CustomExecutors` is a class now rather than an `enum` of static methods.

The first 2 changes follow one another. As we change executors' behavior based on the loaded plugin, the best place to store it is in their state.

I considered an alternative, adding the error handler to a static method signature, but concluded that it is suboptimal because:

* it is a "garbage argument" for all but 5 executors - the chances are slim that other executors will ever use it.
* With the project's tendency to plugin custom behavior, there is a good chance that some other executor might need a similar extension, and we'll have to add another such argument.

### Testing done 

I added new cases to all changed executors' test suites that check server errors are handled by the actual handler supplied in the passed `KsqlConfig`.

### Reviewer checklist

### DIFF
--- a/ksqldb-common/src/main/java/io/confluent/ksql/util/KsqlConfig.java
+++ b/ksqldb-common/src/main/java/io/confluent/ksql/util/KsqlConfig.java
@@ -130,6 +130,12 @@ public class KsqlConfig extends AbstractConfig {
       "Custom extension to allow for more fine-grained control of connector requests made by "
           + "ksqlDB. Extensions should implement the ConnectRequestHeadersExtension interface.";
 
+  public static final String KSQL_CONNECT_SERVER_ERROR_HANDLER =
+      KSQL_CONNECT_PREFIX + "error.handler";
+  public static final String KSQL_CONNECT_SERVER_ERROR_HANDLER_DEFAULT = null;
+  private static final String KSQL_CONNECT_SERVER_ERROR_HANDLER_DOC =
+      "A class that allows the ksqlDB server to customize error handling from connector requests.";
+
   public static final String KSQL_ENABLE_UDFS = "ksql.udfs.enabled";
 
   public static final String KSQL_EXT_DIR = "ksql.extension.dir";
@@ -1329,6 +1335,13 @@ public class KsqlConfig extends AbstractConfig {
             KSQL_HEADERS_COLUMNS_ENABLED_DEFAULT,
             Importance.LOW,
             KSQL_HEADERS_COLUMNS_ENABLED_DOC
+        )
+        .define(
+            KSQL_CONNECT_SERVER_ERROR_HANDLER,
+            Type.CLASS,
+            KSQL_CONNECT_SERVER_ERROR_HANDLER_DEFAULT,
+            Importance.LOW,
+            KSQL_CONNECT_SERVER_ERROR_HANDLER_DOC
         )
         .withClientSslSupport();
 
--- a/ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/execution/ConnectServerErrors.java
+++ b/ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/execution/ConnectServerErrors.java
@@ -1,0 +1,85 @@
+/*
+ * Copyright 2021 Confluent Inc.
+ *
+ * Licensed under the Confluent Community License (the "License"); you may not use
+ * this file except in compliance with the License.  You may obtain a copy of the
+ * License at
+ *
+ * http://www.confluent.io/confluent-community-license
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OF ANY KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations under the License.
+ */
+
+package io.confluent.ksql.rest.server.execution;
+
+import static io.netty.handler.codec.http.HttpResponseStatus.FORBIDDEN;
+import static io.netty.handler.codec.http.HttpResponseStatus.UNAUTHORIZED;
+
+import io.confluent.ksql.parser.tree.Statement;
+import io.confluent.ksql.rest.entity.KsqlEntity;
+import io.confluent.ksql.services.ConnectClient.ConnectResponse;
+import io.confluent.ksql.statement.ConfiguredStatement;
+import java.util.Optional;
+
+/**
+ * An interface that allows to plug-in custom error handling for Connect server errors, such as 403
+ * Forbidden or 401 Unauthorized.
+ */
+public interface ConnectServerErrors {
+
+  /**
+   * Handles error response for a create connector request. This method dispatches to specific
+   * methods based on error codes.
+   *
+   * @param statement the executed statement
+   * @param response the failed response
+   * @return the optional {@link KsqlEntity} that represents server error
+   */
+  default Optional<KsqlEntity> handle(
+      final ConfiguredStatement<? extends Statement> statement,
+      final ConnectResponse<?> response) {
+    if (response.httpCode() == FORBIDDEN.code()) {
+      return handleForbidden(statement, response);
+    } else if (response.httpCode() == UNAUTHORIZED.code()) {
+      return handleUnauthorized(statement, response);
+    } else {
+      return handleDefault(statement, response);
+    }
+  }
+
+  /**
+   * This method allows altering error response on 403 Forbidden.
+   *
+   * @param statement the executed statement
+   * @param response the failed response
+   * @return the optional {@code KsqlEntity} that represents server error
+   */
+  Optional<KsqlEntity> handleForbidden(
+      ConfiguredStatement<? extends Statement> statement,
+      ConnectResponse<?> response);
+
+  /**
+   * This method allows altering error response on 401 Unauthorized.
+   *
+   * @param statement the executed statement
+   * @param response the failed response
+   * @return the optional {@code KsqlEntity} that represents server error
+   */
+  Optional<KsqlEntity> handleUnauthorized(
+      ConfiguredStatement<? extends Statement> statement,
+      ConnectResponse<?> response);
+
+  /**
+   * This method is a fall-back for errors that are not handled by the error-specific methods.
+   *
+   * @param statement the executed statement
+   * @param response the failed response
+   * @return the optional {@code KsqlEntity} that represents server error
+   */
+  Optional<KsqlEntity> handleDefault(
+      ConfiguredStatement<? extends Statement> statement,
+      ConnectResponse<?> response);
+}
--- a/ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/execution/CustomExecutors.java
+++ b/ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/execution/CustomExecutors.java
@@ -15,8 +15,7 @@
 
 package io.confluent.ksql.rest.server.execution;
 
-import com.google.common.collect.ImmutableMap;
-import io.confluent.ksql.KsqlExecutionContext;
+import com.google.common.annotations.VisibleForTesting;
 import io.confluent.ksql.parser.tree.CreateConnector;
 import io.confluent.ksql.parser.tree.DefineVariable;
 import io.confluent.ksql.parser.tree.DescribeConnector;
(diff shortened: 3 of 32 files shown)


ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/KsqlRestConfig.java
```diff
@@ -602,6 +610,12 @@
             DefaultErrorMessages.class,
             Importance.LOW,
             KSQL_SERVER_ERRORS_DOC
+        ).define(
+            KSQL_CONNECT_SERVER_ERROR_HANDLER,
```
vcrfxia, line 614: Let's make add this new config into KsqlConfig instead of KsqlRestConfig. KsqlRestConfig is meant for configs relevant to ksql's REST server. The new config being added more configures the behavior of ksql's engine, rather than its REST server.

ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/execution/StatementExecutor.java
```diff
@@ -35,12 +36,14 @@
    * @param sessionProperties the session properties
    * @param executionContext the context in which to execute it
    * @param serviceContext the services to use to execute it
+   * @param restConfig the rest config allows to plug-in custom behavior, e.g. error handling
```
vcrfxia, line 39: Assuming we move the new config from KsqlRestConfig into KsqlConfig instead, then we don't need this new argument. KsqlConfig is already captured as part of ConfiguredStatement (the first param).

vcrfxia (commented):
Thanks @Gerrrr -- left a couple design comments inline.

Building on the inline comments, if we were to make the two suggested changes, then I think the new functionality would work but we'd be loading the custom extension on each connector request which isn't great. Ideally we'd just load the extension once upfront when starting the server, and pass that to the ConnectExecutor. I think we have two options for doing this. Both involve changes where we load the custom executors in KsqlResource ([link](https://github.com/confluentinc/ksql/blob/aa7bc9134eddc4b5eebfa2141ac87c6ae0632a87/ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/resources/KsqlResource.java#L190)).

One option is to update CustomExecutors to accept a ConnectServerErrors and pipe that through to ConnectExecutor. Unfortunately this means that ConnectExecutor (and CustomExecutors) will no longer be able to be final as they are today. If this pollutes the code/makes things unnecessarily complicated, then let's not.

Another option is to pull the ConnectExecutor functionality out of CustomExectors entirely, update the RequestHandler constructor to accept a ConnectExecutor, and update RequestHandler to send connector requests to the ConnectExecutor. Then ConnectExecutor can be configured at creation time, similar to how the DistributingExecutor is configured today (same place in KsqlResource). 

My preference is for the former unless it's ugly, in which case we can go with the latter. Give both a try and see which one you think is cleaner?

Gerrrr:
@vcrfxia I agree with your assessment that it is a bad idea to create a plugin instance on every request, and it is much better to load it once on executor creation or pass the created plugin to the static `execute` method. In 4d4ea68b41 I chose the former option. Specifically, I made the following changes:
* `ConnectExecutor` has an immutable error handler as part of its state.
* `ConnectExecutor#execute` is an instance method now.
* `CustomExecutors` is a class now rather than an enum of static methods.

The first 2 changes follow one another. As we change `ConnectExecutor`'s behavior based on the loaded plugin, the best place to store it is in its state.

I considered the alternative mentioned above, adding the error handler to a static method signature, but concluded that it is suboptimal because:
1. it is a "garbage argument" for all but 1 executor - the chances are slim that any other executor will ever use it.
2. With the project's tendency to plugin custom behavior, there is a good chance that some other executor might need a similar extension, and we'll have to add another such argument.

WDYT?

ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/execution/StatementExecutor.java
Gerrrr: Good idea! Done in 4d4ea68.

ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/KsqlRestConfig.java
Gerrrr: Done in 4d4ea68.

ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/execution/ConnectServerErrors.java
```diff
@@ -0,0 +1,79 @@
...
+import org.apache.kafka.connect.runtime.rest.entities.ConnectorInfo;
+
+/**
+ * An interface that allows to plug-in custom error handling for Connect server errors, such as 403
+ * Forbidden or 401 Unauthorized.
+ */
+public interface ConnectServerErrors {
+
+  /**
+   * Handles error response for a create connector request. This method dispatches to specific
```
vcrfxia, line 35: This new plugin should be applied to all connector requests, not just CREATE CONNECTOR.

ksqldb-rest-app/src/test/java/io/confluent/ksql/rest/server/execution/ListTopicsExecutorTest.java
```diff
@@ -59,6 +63,7 @@ public void setUp() {
           engine.getServiceContext().getSchemaRegistryClientFactory(),
           engine.getServiceContext().getConnectClient()
     );
+    customExecutors = new CustomExecutors(new DefaultConnectServerErrors());
```
vcrfxia, line 66: nit: most of the updated tests introduce
```
private static final CustomExecutors CUSTOM_EXECUTORS
```
instead, which I think is cleaner. Can we keep these test updates consistent?

ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/execution/CustomExecutors.java
```diff
@@ -42,91 +41,156 @@
...
+    return (StatementExecutor<DescribeFunction>) EXECUTOR_MAP.get(DescribeFunction.class);
+  }
+
+  @VisibleForTesting
+  public StatementExecutor<ListQueries> listQueries() {
+    return (StatementExecutor<ListQueries>) EXECUTOR_MAP.get(ListQueries.class);
+  }
+
+  @VisibleForTesting
+  public StatementExecutor<ListVariables> listVariables() {
```
vcrfxia, line 192: Do these test-only methods need to be public or is package-private sufficient?

ksqldb-rest-app/src/test/java/io/confluent/ksql/rest/server/computation/RecoveryTest.java
```diff
@@ -105,6 +106,9 @@
   @Mock
   private Errors errorHandler = mock(Errors.class);
 
+  @Mock
```
vcrfxia, line 109: nit: unused

ksqldb-rest-app/src/test/java/io/confluent/ksql/rest/server/execution/ConnectExecutorTest.java
```diff
@@ -243,6 +257,34 @@ public void shouldReturnErrorIfConnectorTypeIsEmpty() {
     assertThat(((ErrorEntity) entity.get()).getErrorMessage(), is(expectedError));
   }
 
+  @Test
+  public void shouldReturnPluggableErrors() {
```
vcrfxia, line 261: Thanks for the test coverage! I think it's worth adding separate unit tests for each of the three cases (403, 401, other). Hopefully it's quick.

ksqldb-common/src/main/java/io/confluent/ksql/util/KsqlConfig.java
```diff
@@ -126,6 +126,12 @@
   public static final String CONNECT_BASIC_AUTH_FAIL_ON_UNREADABLE_CREDENTIALS =
       KSQL_CONNECT_PREFIX + "basic.auth.credentials.fail.on.unreadable";
 
+  public static final String KSQL_CONNECT_SERVER_ERROR_HANDLER =
+      KSQL_CONNECT_PREFIX + "error.handler";
+  public static final String KSQL_CONNECT_SERVER_ERROR_HANDLER_DEFAULT = null;
+  private static final String KSQL_CONNECT_SERVER_ERROR_HANDLER_DOC =
+      "A class that allows the KSQL server to customize connect error handling.";
```
vcrfxia, line 133: nit:
```suggestion
      "A class that allows the ksqlDB server to customize error handling from connector requests.";
```

ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/execution/ConnectServerErrors.java
```diff
@@ -0,0 +1,79 @@
+/*
+ * Copyright 2018 Confluent Inc.
```
vcrfxia, line 2: nit: year, and also for the other newly added file.

ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/execution/ConnectServerErrors.java
```diff
@@ -0,0 +1,79 @@
...
+   *
+   * @param statement the executed statement
+   * @param response the failed response
+   * @return the optional {@code KsqlEntity} that represents server error
+   */
+  Optional<KsqlEntity> handleUnauthorized(
+      ConfiguredStatement<CreateConnector> statement,
+      ConnectResponse<ConnectorInfo> response);
+
+  Optional<KsqlEntity> handleDefault(
```
vcrfxia, line 76: nit: add javadoc

vcrfxia (commented):
Thanks @Gerrrr -- this is looking really great! One main comment plus a bunch of minor ones. Also, can you update the PR description to reflect the new approach?

ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/execution/ConnectServerErrors.java
Gerrrr: Done in https://github.com/confluentinc/ksql/pull/8480/commits/7ac1fe2c1b271aec0498c7f1315b34f0681bb7a4#diff-24e5d395473a57f7588579718e1f2209f549729a9d975c232ebedb3c0e52f905R81-R113.

ksqldb-rest-app/src/test/java/io/confluent/ksql/rest/server/execution/ListTopicsExecutorTest.java
Gerrrr: Done in https://github.com/confluentinc/ksql/pull/8480/commits/7ac1fe2c1b271aec0498c7f1315b34f0681bb7a4#diff-4a4e423d225fa849b4fab3a3e83448875660ab295021d824c67562715b987b0a.

ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/execution/CustomExecutors.java
Gerrrr: You are right! These methods can be package-private. Changed in https://github.com/confluentinc/ksql/pull/8480/commits/7ac1fe2c1b271aec0498c7f1315b34f0681bb7a4#diff-24e5d395473a57f7588579718e1f2209f549729a9d975c232ebedb3c0e52f905R116-R199.

ksqldb-rest-app/src/test/java/io/confluent/ksql/rest/server/computation/RecoveryTest.java
Gerrrr: Removed in https://github.com/confluentinc/ksql/pull/8480/commits/7ac1fe2c1b271aec0498c7f1315b34f0681bb7a4#diff-48e54df909b0174a4501759566450a4f8988e57588e9b4dc0ff56999dfab88aa.

ksqldb-rest-app/src/test/java/io/confluent/ksql/rest/server/execution/ConnectExecutorTest.java
Gerrrr: Done in https://github.com/confluentinc/ksql/pull/8480/commits/7ac1fe2c1b271aec0498c7f1315b34f0681bb7a4#diff-89adae5ad10f6b2fbed946a5e2cf663e25b4dc129525f3a0fff774428af3a08cR262-R344.

ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/execution/ConnectServerErrors.java
Gerrrr: Fixed in https://github.com/confluentinc/ksql/pull/8480/commits/7ac1fe2c1b271aec0498c7f1315b34f0681bb7a4#diff-583e52f2f5ace65057d66ec857d4f304326c56d5b542d0c14c2e685c97c875c0R2, https://github.com/confluentinc/ksql/pull/8480/commits/7ac1fe2c1b271aec0498c7f1315b34f0681bb7a4, https://github.com/confluentinc/ksql/pull/8480/commits/7ac1fe2c1b271aec0498c7f1315b34f0681bb7a4#diff-d412eea1cfe4a271bb0654c2a543c2ad7d7cb53febd8e522c9c04dcad5f22169R2.

ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/execution/ConnectServerErrors.java
Gerrrr: Added in https://github.com/confluentinc/ksql/pull/8480/commits/7ac1fe2c1b271aec0498c7f1315b34f0681bb7a4#diff-583e52f2f5ace65057d66ec857d4f304326c56d5b542d0c14c2e685c97c875c0R75-R81.

ksqldb-rest-app/src/test/java/io/confluent/ksql/rest/server/execution/ListConnectorsExecutorTest.java
```diff
@@ -212,4 +226,94 @@ public void shouldListInvalidConnectorWithNoInfo() {
...
   }
 
+  @Test
+  public void shouldReturnPluggableForbiddenError() {
+    //Given:
+    when(connectClient.connectors())
+        .thenReturn(
+            ConnectResponse.failure("FORBIDDEN", HttpStatus.SC_FORBIDDEN));
+
+    final KsqlConfig config = new KsqlConfig(ImmutableMap.of(
```
vcrfxia, line 236: nit: we can reduce code repetition and improve readability of unit tests by pulling out common setup code like this into a helper method such as `givenCustomErrorHander();`. No need to go through and update all your newly added tests if it's too much toil but definitely good to keep in mind for the future.

ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/validation/CustomValidators.java
```diff
@@ -99,7 +100,8 @@
   SHOW_COLUMNS(ShowColumns.class, ListSourceExecutor::columns),
   EXPLAIN(Explain.class, ExplainExecutor::execute),
   DESCRIBE_FUNCTION(DescribeFunction.class, DescribeFunctionExecutor::execute),
-  DESCRIBE_CONNECTOR(DescribeConnector.class, new DescribeConnectorExecutor()::execute),
+  DESCRIBE_CONNECTOR(DescribeConnector.class,
+      new DescribeConnectorExecutor(new DefaultConnectServerErrors())::execute),
```
vcrfxia, line 104: Unrelated to your PR but what is this validation actually doing? AFAICT `DescribeConnectorExecutor::execute()` never actually throws an error. Does this validation do anything or is it a no-op?

vcrfxia (approved):
Thanks @Gerrrr ! LGTM with a couple non-blocking comments inline.

Thanks for updating the PR description as well. Can you update it to reflect that custom errors (besides 401 and 403) are also customizable? Otherwise it looks fantastic -- comprehensive and concise :)

ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/validation/CustomValidators.java
Gerrrr: I agree with your judgement that `DescribeConnectorExecutor::execute()` never throws, so it's a no-op. 

After a brief look at this class I think that only very limited number of executors ever throw, consequently only these validators actually do any useful work, namely `InsertValuesExecutor()::execute`, `ListSourceExecutor::columns`, and `ExplainExecutor::execute`. 

Maybe there is a reason for this, but it appears to me that executor behavior is inconsistent - either all executors should throw on errors or return an error entity instead. WDYT? If you agree, I'll open an issue.

ksqldb-rest-app/src/test/java/io/confluent/ksql/rest/server/execution/ListConnectorsExecutorTest.java
Gerrrr: Good idea! Done in dcc2852ac34031b32089af45f6b8ebfa71e1edd8.

PR #8481, opened 2021-12-09
feat: restart runtime after alter system query

tolgadur:
### Description 
Once the system configs are altered we need to restart the runtimes to pick up the change. This PR is dependend on https://github.com/confluentinc/ksql/pull/8434

### Testing done 
unit tests 

### Reviewer checklist

### DIFF
--- a/ksqldb-engine/src/main/java/io/confluent/ksql/engine/KsqlEngine.java
+++ b/ksqldb-engine/src/main/java/io/confluent/ksql/engine/KsqlEngine.java
@@ -64,6 +64,7 @@ import io.confluent.ksql.util.KsqlStatementException;
 import io.confluent.ksql.util.PersistentQueryMetadata;
 import io.confluent.ksql.util.QueryMetadata;
 import io.confluent.ksql.util.ScalablePushQueryMetadata;
+import io.confluent.ksql.util.SharedKafkaStreamsRuntime;
 import io.confluent.ksql.util.StreamPullQueryMetadata;
 import io.confluent.ksql.util.TransientQueryMetadata;
 import io.vertx.core.Context;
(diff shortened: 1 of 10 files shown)


ksqldb-engine/src/main/java/io/confluent/ksql/engine/KsqlEngine.java
```diff
@@ -213,6 +213,11 @@ public boolean hasActiveQueries() {
     return !primaryContext.getQueryRegistry().getPersistentQueries().isEmpty();
   }
 
+  public void restartStreamsRuntime() {
```
wcarlson5, line 216: Why do we need this at this level? would it be better to just call this whenever the alter system is called? Or should we leave them decoupled? WDYT

ksqldb-engine/src/main/java/io/confluent/ksql/query/QueryRegistryImpl.java
```diff
@@ -221,6 +223,23 @@ public TransientQueryMetadata createStreamPullQuery(
     return query;
   }
 
+  @Override
+  public void restartStreamsRuntime(final KsqlConfig config) {
```
wcarlson5, line 227: Can you change the name to include this is picking up the new configs?

wcarlson5 (commented):
its on its way

ksqldb-engine/src/main/java/io/confluent/ksql/engine/KsqlEngine.java
tolgadur: I think we need to seperate it as we call the method in the `ValidatedCommandFactory` for validation purposes. If we didn't the streams runtime.

wcarlson5: ah yeah thats a good point

wcarlson5 (approved):
LGTM

PR #8483, opened 2021-12-09
fix: Fixes StreamPublisher to use reset rather than closing connection

AlanConfluent:
### Description 
Currently, on `StreamPublisher` we explicitly call `request.connection().close()` when we want to close the publisher, which closes the connection regardless of whether http1 or http2 is in use.  What this means is that under http2, where multiple requests maybe be getting multiplexed, this has the effect of killing other live streams sharing the same connection, in a way that doesn't seem to trigger any callbacks at the moment on the server side unfortunately.  And even if it did, we're still breaking good requests.

`request.reset()` is the equivalent of `request.connection().close()` in http1, but only ends the stream in http2, which is what we want.

The only downside at the moment is that the last request on an http2 connection seems to have a delay of about 30s after sending a goAway to actually close the connection, which is waiting for any existing requests to complete.  We just want it to end immediately, as it would with http1, so we have it do that.  I couldn't find a better Vertx setting or method, but this works as intended.

This change enables this behavior only with certain rest requests as the moment, only where http2 is known to be used.

### Testing done 
Manually tests many times with multiple connections to the server.

### Reviewer checklist

### DIFF
--- a/ksqldb-rest-client/src/main/java/io/confluent/ksql/rest/client/KsqlTarget.java
+++ b/ksqldb-rest-client/src/main/java/io/confluent/ksql/rest/client/KsqlTarget.java
@@ -371,7 +371,7 @@ public final class KsqlTarget {
         (resp, vcf) -> {
           if (resp.statusCode() == 200) {
             pubRef.set(new StreamPublisher<>(Vertx.currentContext(),
-                resp, mapper, vcf, true));
+                resp, mapper, vcf, true, true));
             vcf.complete(new ResponseWithBody(resp));
           } else {
             resp.bodyHandler(body -> vcf.complete(new ResponseWithBody(resp, body)));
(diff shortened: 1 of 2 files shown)


agavra (approved):

ksqldb-rest-app/src/main/java/io/confluent/ksql/api/server/QueryStreamHandler.java
```diff
@@ -82,6 +83,14 @@ public void handle(final RoutingContext routingContext) {
     final Optional<Boolean> internalRequest = ServerVerticle.isInternalRequest(routingContext);
     final MetricsCallbackHolder metricsCallbackHolder = new MetricsCallbackHolder();
     final long startTimeNanos = Time.SYSTEM.nanoseconds();
+
+    // For http2, if we get a goAway, meaning that the connection is going to be shutdown, just end
+    // it immediately, otherwise there will be a delay of 30s until we get a call on the endhandler.
+    // Consider merging this logic into ConnectionQueryManager used by push queries.
```
vvcephei, line 89: It might be nice to track this with a ticket, if I'm reading it correctly as a TODO.

vvcephei (dismissed):
LGTM! Thanks, @AlanConfluent

PR #8487, opened 2021-12-10
chore: exclude dependency to log4j2

cadonna:
### Description 

ksql has a dependency to log4j2 although it uses log4j1 for
logging. The dependency to log4j2 comes transitively from
confluent-log4j.

This PR excludes the log4j2 dependency from confluent-log4j.

### Reviewer checklist

### DIFF
--- a/ksqldb-parser/src/main/java/io/confluent/ksql/util/ParserUtil.java
+++ b/ksqldb-parser/src/main/java/io/confluent/ksql/util/ParserUtil.java
@@ -18,6 +18,7 @@ package io.confluent.ksql.util;
 import static io.confluent.ksql.parser.SqlBaseParser.DecimalLiteralContext;
 import static java.util.Objects.requireNonNull;
 
+import com.google.common.base.Strings;
 import io.confluent.ksql.execution.expression.tree.DecimalLiteral;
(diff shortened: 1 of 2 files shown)


ksqldb-parser/src/main/java/io/confluent/ksql/util/ParserUtil.java
```diff
@@ -256,7 +256,7 @@ public static ColumnConstraints getColumnConstraints(
     } else if (context.HEADERS() != null) {
       builder = builder.headers();
     } else if (context.HEADER() != null) {
-      builder = builder.header(Strings.trimToNull(unquote(context.STRING().getText(), "'")));
```
spena, line 259: Can you trim the text after the `unquote` call? The case was that a user could use `header('  k  ')`, and that key should be `k`. The `trimToNull` justs makes sure that `header('   ')` is null.

cadonna: I see. Thanks for the catch!

spena (approved):

nateab:
Im gonna merge this in @cadonna so I can cherry pick it on the release stabilization branch for ksql, thanks!

PR #8490, opened 2021-12-10
fix: make HEADER and HEADERS non-reserved

### LINKED ISSUE #8446: Column name 'headers' throws an error that it is reserved name
**Describe the bug**
When using `headers` as a column name, even if I didn't specify headers as column type, an error is thrown that is a reserved keyword. However, this error is not thrown when using `key` as a column name.

```
create stream s1(key string key, message string, headers array<struct<key string, value bytes>>) with (kafka_topic='topic1', value_format
='delimited');
line 1:50: "headers" is a reserved keyword and it can't be used as an identifier. You can use it as an identifier by escaping it as 'headers' 
Statement: create stream s1(key string key, message string, headers array<struct<key string, value bytes>> headers) with (kafka_topic='topic1', value_f
ormat='delimited');
Caused by: line 1:50: "headers" is a reserved keyword and it can't be used as an
	identifier. You can use it as an identifier by escaping it as 'headers'
```

jzaralim:
### Description 
Fixes #8446 

### Testing done 
Changed some QTTs to use these as column names

### Reviewer checklist

### DIFF
--- a/ksqldb-functional-tests/src/test/resources/historical_plans/headers_-_Can_use_header_columns_in_filters/7.2.0_1639450823104/plan.json
+++ b/ksqldb-functional-tests/src/test/resources/historical_plans/headers_-_Can_use_header_columns_in_filters/7.2.0_1639450823104/plan.json
@@ -1,0 +1,179 @@
+{
+  "plan" : [ {
+    "@type" : "ksqlPlanV1",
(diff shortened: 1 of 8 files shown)


spena (approved):

guozhangwang (approved):
LGTM

PR #8492, opened 2021-12-10
feat: throw exception if feature flag is off

tolgadur:
### Description 
The `Alter System` work work as expected in the old runtime. We want to disallow this if the feature flag is not set.

### Testing done 
Existing tests

### Reviewer checklist

### DIFF
--- a/ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/computation/ValidatedCommandFactory.java
+++ b/ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/computation/ValidatedCommandFactory.java
@@ -136,6 +136,12 @@ public final class ValidatedCommandFactory {
     final String propertyName = alterSystemProperty.getPropertyName();
     final String propertyValue = alterSystemProperty.getPropertyValue();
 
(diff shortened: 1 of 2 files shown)


ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/computation/ValidatedCommandFactory.java
```diff
@@ -136,6 +136,11 @@ private static Command createForAlterSystemQuery(
     final String propertyName = alterSystemProperty.getPropertyName();
     final String propertyValue = alterSystemProperty.getPropertyValue();
 
+    // raise exception if feature flag is set
+    if (!context.getKsqlConfig().getBoolean(KsqlConfig.KSQL_SHARED_RUNTIME_ENABLED)) {
+      throw new KsqlServerException("This feature is not yet enabled.");
```
guozhangwang, line 141: Could we be more specific in the error message, like "Cannot alter system configs when KSQL_SHARED_RUNTIME_ENABLED is turned off."

guozhangwang (approved):
One nit, otherwise LGTM!

PR #8493, opened 2021-12-10
fix: close all runtimes in sandbox

wcarlson5:
this is about cleaning up the validation runtimes. when creating the sandbox there was some issues where the runtimes would get made at creation but the sandbox execution context doesn’t get cleaned up. Now I don’t make the validation runtimes until a query is issued. And ensue that all the runtimes get cleaned up.

### Testing done 

### Reviewer checklist

### DIFF
--- a/ksqldb-engine/src/main/java/io/confluent/ksql/engine/SandboxedExecutionContext.java
+++ b/ksqldb-engine/src/main/java/io/confluent/ksql/engine/SandboxedExecutionContext.java
@@ -173,7 +173,7 @@ final class SandboxedExecutionContext implements KsqlExecutionContext {
     if (getKsqlConfig().getBoolean(KsqlConfig.KSQL_SHARED_RUNTIME_ENABLED)) {
       result.getQuery().map(QueryMetadata::getKafkaStreams).ifPresent(streams -> streams.close());
     } else {
-      engineContext.getQueryRegistry().getPersistentQueries().values().stream().map(QueryMetadata::getKafkaStreams).forEach(KafkaStreams::close);
+      engineContext.getQueryRegistry().closeRuntimes();
     }
     return result;
   }
--- a/ksqldb-engine/src/main/java/io/confluent/ksql/engine/SandboxedExecutionContext.java
+++ b/ksqldb-engine/src/main/java/io/confluent/ksql/engine/SandboxedExecutionContext.java
@@ -47,7 +47,6 @@ import io.confluent.ksql.util.Sandbox;
 import io.confluent.ksql.util.ScalablePushQueryMetadata;
 import io.confluent.ksql.util.TransientQueryMetadata;
 import io.vertx.core.Context;
-import org.apache.kafka.streams.KafkaStreams;
 import java.util.List;
 import java.util.Map;
 import java.util.Optional;
--- a/ksqldb-engine/src/main/java/io/confluent/ksql/engine/SandboxedExecutionContext.java
+++ b/ksqldb-engine/src/main/java/io/confluent/ksql/engine/SandboxedExecutionContext.java
@@ -47,6 +47,7 @@ import io.confluent.ksql.util.Sandbox;
 import io.confluent.ksql.util.ScalablePushQueryMetadata;
 import io.confluent.ksql.util.TransientQueryMetadata;
 import io.vertx.core.Context;
+import org.apache.kafka.streams.KafkaStreams;
 import java.util.List;
 import java.util.Map;
 import java.util.Optional;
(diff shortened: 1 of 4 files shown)


ksqldb-engine/src/main/java/io/confluent/ksql/engine/SandboxedExecutionContext.java
```diff
@@ -160,7 +161,11 @@ public ExecuteResult execute(
     ).execute(ksqlPlan.getPlan());
 
     // Having a streams running in a sandboxed environment is not necessary
-    result.getQuery().map(QueryMetadata::getKafkaStreams).ifPresent(streams -> streams.close());
+    if (getKsqlConfig().getBoolean(KsqlConfig.KSQL_SHARED_RUNTIME_ENABLED)) {
+      result.getQuery().map(QueryMetadata::getKafkaStreams).ifPresent(streams -> streams.close());
+    } else {
+      engineContext.getQueryRegistry().getPersistentQueries().values().stream().map(QueryMetadata::getKafkaStreams).forEach(KafkaStreams::close);
```
ableegoldman, line 167: This is going to close the actual runtime -- note that when we create the `SandboxedBinPackedPersistentQueryMetadataImpl` it saves a reference to the runtime of the original `BinPackedPersistentQueryMetadataImpl`. This is one of the things we need to fix

ableegoldman: It seems like we should only ever be creating SandboxedSharedKafkaStreamsRuntimeImpl in one place, specifically we should create them in the `QueryRegistryImpl` constructor like we do now. But we then need to connect the `SandboxedBinPackedPersistentQueryMetadataImpl` to the sandboxed runtime rather than let it have a reference to the actual runtime, to avoid it closing that

ableegoldman: But I think this definitely explains why the validation runtimes aren't being closed

wcarlson5: That seems more like an issue with the sandbox version of the query registry

wcarlson5: but I have a work around

ksqldb-engine/src/main/java/io/confluent/ksql/engine/SandboxedExecutionContext.java
```diff
@@ -160,7 +160,11 @@ public ExecuteResult execute(
     ).execute(ksqlPlan.getPlan());
 
     // Having a streams running in a sandboxed environment is not necessary
-    result.getQuery().map(QueryMetadata::getKafkaStreams).ifPresent(streams -> streams.close());
+    if (!getKsqlConfig().getBoolean(KsqlConfig.KSQL_SHARED_RUNTIME_ENABLED)) {
```
wcarlson5, line 163: We need make sure all the runtimes for this sandbox gets closed not just the one for the new query. They will not be used again

ksqldb-engine/src/main/java/io/confluent/ksql/query/QueryBuilder.java
```diff
@@ -446,33 +447,41 @@ PersistentQueryMetadata buildPersistentQueryInSharedRuntime(
...
+                queryId,
+                applicationId,
+                queryOverrides,
+                physicalPlan
+            )
     );
+    if (real) {
+      return binPackedPersistentQueryMetadata;
+    } else {
+      return SandboxedBinPackedPersistentQueryMetadataImpl.of(
```
wcarlson5, line 481: the new validation query was a normal query in  a sandbox env which was weird. This should prevent that

ksqldb-engine/src/main/java/io/confluent/ksql/query/QueryRegistryImpl.java
```diff
@@ -125,9 +126,7 @@ private QueryRegistryImpl(final QueryRegistryImpl original) {
         .filter(Optional::isPresent)
         .map(Optional::get)
         .collect(Collectors.toList());
-    streams.addAll(original.streams.stream()
-        .map(SandboxedSharedKafkaStreamsRuntimeImpl::new)
-        .collect(Collectors.toList()));
+    sourceStreams.addAll(original.streams);
```
wcarlson5, line 129: We used to copy these over when the sandbox was made. But the sandbox isn't always used for persistant queries and isn't cleaned up so we should wait to copy them over until we are sure they are needed and will be cleaned up

guozhangwang: I'm wondering: since we now always `closeRuntimes` at the end even if no persistent queries are created during the lifetime of the sandbox, does that still matter to do the late-copy and keep a separate `sourceStreams` from `streams`?

wcarlson5: We actually don't always close the sandbox. the only persistant resource used in the sandbox that I can tell is the runtimes. So if we don't late copy the sourceStreams they will be left behind quite often

ableegoldman (approved):
LGTM, I hope we can advocate for some time to clean up some of this tech debt in the near future though because it's super confusing

ableegoldman:
@wcarlson5 once this is merged can you also retrigger the build for PR that reactivates the shared runtimes parameter in the EndToEndIntegrationTest? I have a feeling this could be related

PR #8494, opened 2021-12-10
chore: Clean up build

vvcephei:
* Fix deprecation warnings
* Fix ConsumerCollectorTest fails for some people in Maven

I think the failure is because Maven chooses to run the tests in a different
order for some people, which causes the validations not to work due to the
way that ConsumerCollector creates one of the metrics in a static initialization
block.

It looks like it was written that way because we wanted to have a unique sensor
across all interceptors. One way to do that is to pass it in
to all instances via the config map in `configure`. Another way is to simply leverage
the fact that sensor creation and metric registration are idempotent in the Metrics
library. Note: they weren't idempotent at the time this class was written.

In this PR, I went with the latter because it keeps the sensor definition local to
the class.

I also cleaned up some synchronization that is also no longer necessary, given
the idempotence of registration.

### Reviewer checklist

### DIFF
--- a/ksqldb-common/src/main/java/io/confluent/ksql/metrics/ConsumerCollector.java
+++ b/ksqldb-common/src/main/java/io/confluent/ksql/metrics/ConsumerCollector.java
@@ -23,7 +23,6 @@ import java.util.Collection;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
-import java.util.Map.Entry;
 import java.util.function.Function;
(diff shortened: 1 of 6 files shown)


wcarlson5 (approved):
Thanks for fixing the build!

vvcephei:
I'm closing this in favor of https://github.com/confluentinc/ksql/pull/8498 which actually fixes the structural problem.

Issue #2701, opened 2019-04-17
Breaking Change: Inconsistency between GROUP BY and PARTITION BY's handling of aliases


big-andy-coates:
There is currently an inconsistency between `GROUP BY`'s and `PARTITION BY`'s handling of aliases:
 - `PARTITION BY` requires the statement to use the name of the column as it is in the output schema, i.e. the alias, e.g.

```sql
SELECT field AS alias FROM whatever PARTITION BY alias;
```

- `GROUP BY` requires the statement to use the name of the column as it is in the source schema, e.g.

```sql
SELECT field AS alias FROM whatever GROUP BY field;
```

This is inconsistent and we should look to standardize on one.

agavra:
Confirming postgres behavior:
```sql
almog.gavra=# CREATE TABLE users (id VARCHAR, name VARCHAR);
CREATE TABLE
almog.gavra=# INSERT INTO users (id, name) VALUES ('1', 'albert');
INSERT 0 1
almog.gavra=# INSERT INTO users (id, name) VALUES ('1', 'alex');
INSERT 0 1
almog.gavra=# SELECT substring(name, 1, 1) as name, count(*) FROM users GROUP by name;
 name | count
------+-------
 a    |     1
 a    |     1
(2 rows)
```
It groups based on the _original_ columns (which is what we do for `GROUP BY`). I think we should stick to this behavior.

This will, though, cause some confusion. If you partition by something that you don't project it will be the rowkey but it will not exist in the value. If I did:
```
CREATE STREAM foo AS SELECT all_caps(name) as name FROM bar PARTITION BY name;
```
The `ROWKEY` would be the old name, but the `name` column would be the new one... 😨 It might make sense to have two different behaviors for each one, but I can also see the use in unifying them.

big-andy-coates:
For me, `GROUP BY` is implicitly also a `PARTITION BY`.  The data ends up partitioned-by the group-by key.  So there is no different.

I find thinking of the operations like this helps me, (using parentheses to show scope):

```
SELECT A (FROM X GROUP BY Y);
SELECT A (FROM X PARTITION BY Y);
```

we just don't materialize, (i.e. write to kafka), the grouped / partitioned source in the parenthesis.  Instead, we delay materialization, apply the select, then materialize

PR #8495, opened 2021-12-10
chore: improve error message on unknown PARTITION BY column

agavra:
### Description 
it seems that users are generally confused by the behavior of `PARTITION BY` (that it functions on source columns, no sink columns). While I admit to that having been a questionable decision to begin with (see discussion on https://github.com/confluentinc/ksql/issues/2701) we're kinda stuck with it - so may as well build the explanation into the product.

### Testing done 

QTT

### Reviewer checklist


mjsax:
Not sure if it's strictly a bug, but I agree it's not user friendly. Basically, if you want to apply a `PARTITION BY` on the output, you need to repeat the expression used in the `SELECT` in the `PARTITION BY`:

```sql
SELECT substring(foo, 1, 1) as bar FROM s1 PARTITION BY substring(foo, 1, 1);
```

It's an annoying SQL trait (that also applies to `GROUP BY`). But I think it would be possible to fix if we wanted to. There are still couple of open tickets about it.

agavra:
+1 thanks for the review @mjsax - generally I agree with you and the workaround is generally acceptable

PR #8498, opened 2021-12-11
refactor: eliminate static metrics registry

vvcephei:
### Description 
Previously, metrics were kept in a static registry, causing tests to interfere with each other. This change eliminates the static metrics registry and allows each ksql server to create its own registry and pass it to all components.

### Testing done 
Running the existing test suite.

### Reviewer checklist


ksqldb-common/src/test/java/io/confluent/ksql/metrics/MetricCollectorsTest.java
```diff
@@ -292,12 +467,16 @@ public void shouldNotIncludeRestoreConsumersWhenComputingPerQueryStats() {
...
+        metricCollectors
+    );
     for (int i = 0; i < 2000; i++) {
-      StreamsErrorCollector.recordError("test-application", TEST_TOPIC);
+      streamsErrorCollector.recordError(TEST_TOPIC);
     }
     // we have 2000 errors in one sample out of a 100. So the effective error rate computed
     // should be 20 for this run.
-    assertEquals(20.0, Math.floor(MetricCollectors.currentErrorRate()), 0.1);
-    StreamsErrorCollector.notifyApplicationClose("test-application");
```
vvcephei, line 301: Instead of unregistering the collector, we can just let it go out of scope now.

ksqldb-common/src/test/java/io/confluent/ksql/metrics/StreamsErrorCollectorTest.java
```diff
@@ -56,73 +52,108 @@ public void setUp() {
     applicationId = buildApplicationId();
   }
 
-  @SuppressFBWarnings("ST_WRITE_TO_STATIC_FROM_INSTANCE_METHOD")
-  @After
-  public void tearDown() {
```
vvcephei, line 61: not needed anymore!

ksqldb-common/src/test/java/io/confluent/ksql/metrics/StreamsErrorCollectorTest.java
```diff
@@ -56,73 +52,108 @@ public void setUp() {
...
     final String otherTopicId = "other-topic-id";
     final int nmsgs = 3;
-    IntStream.range(0, nmsgs).forEach(i -> recordError(applicationId, TOPIC_NAME));
-    IntStream.range(0, nmsgs + 1).forEach(i -> recordError(otherAppId, otherTopicId));
+    IntStream.range(0, nmsgs).forEach(i -> collector.recordError(TOPIC_NAME));
+    IntStream.range(0, nmsgs + 1).forEach(i -> otherCollector.recordError(otherTopicId));
 
     // When:
-    notifyApplicationClose(otherAppId);
+    otherCollector.cleanup();
```
vvcephei, line 136: The essence of the test is still the same. When the "other collector" gets cleaned up, it should de-register and drop its collectors.

ksqldb-common/src/test/java/io/confluent/ksql/metrics/StreamsErrorCollectorTest.java
```diff
@@ -56,73 +52,108 @@ public void setUp() {
...
 
     // Then:
+    final Collection<Stat> statsFor = metricCollectors.getStatsFor(TOPIC_NAME, true);
+    assertThat(statsFor, hasSize(2));
+    assertThat(getStat(statsFor, "consumer-failed-messages"), of(equalTo(nmsgs * 1.0)));
+    assertThat(getStat(statsFor, "consumer-failed-messages-per-sec"), of(greaterThan(0.0)));
+
+    final Collection<Stat> otherStatsFor = metricCollectors.getStatsFor(otherTopicId, true);
+    assertThat(otherStatsFor, hasSize(0));
+
```
vvcephei, line 146: I kept a few checks I had added while I was trying to figure out what the test was trying to do.

ksqldb-engine/src/main/java/io/confluent/ksql/KsqlExecutionContext.java
```diff
@@ -73,6 +74,8 @@
    */
   KsqlConfig getKsqlConfig();
 
+  MetricCollectors metricCollectors();
```
vvcephei, line 77: This doesn't seem like a horrible expansion of the responsibility of the "execution contexct" to also expose the metrics registry. Adding this here allows us to save on adding the registry to the constructors of like 50 more classes.

ksqldb-engine/src/main/java/io/confluent/ksql/embedded/KsqlContext.java
```diff
@@ -74,13 +75,19 @@
    */
   public static KsqlContext create(
       final KsqlConfig ksqlConfig,
-      final ProcessingLogContext processingLogContext
+      final ProcessingLogContext processingLogContext,
+      final MetricCollectors metricCollectors
```
vvcephei, line 79: Just plumbing the registry through everywhere it's needed.

ksqldb-engine/src/main/java/io/confluent/ksql/embedded/KsqlContext.java
```diff
@@ -74,13 +75,19 @@
...
     Objects.requireNonNull(ksqlConfig, "ksqlConfig cannot be null.");
     final ServiceContext serviceContext =
         ServiceContextFactory.create(ksqlConfig, DisabledKsqlClient::instance);
     final MutableFunctionRegistry functionRegistry = new InternalFunctionRegistry();
-    UserFunctionLoader.newInstance(ksqlConfig, functionRegistry, ".").load();
+    UserFunctionLoader.newInstance(
+        ksqlConfig,
+        functionRegistry,
+        ".",
+        metricCollectors.getMetrics()
```
vvcephei, line 89: Rather than just throwing in the kitchen sink, I opted to only pass the Metrics instead of MetricCollectors through to the locations that just need the reduced scope.

ksqldb-engine/src/main/java/io/confluent/ksql/engine/KsqlEngine.java
```diff
@@ -400,7 +414,7 @@ public StreamPullQueryMetadata createStreamPullQuery(
   private TopicDescription getTopicDescription(final Admin admin, final String sourceTopicName) {
     final KafkaFuture<TopicDescription> topicDescriptionKafkaFuture = admin
         .describeTopics(Collections.singletonList(sourceTopicName))
-        .values()
+        .topicNameValues()
```
vvcephei, line 417: deprecation warning

ksqldb-engine/src/main/java/io/confluent/ksql/engine/KsqlEngine.java
```diff
@@ -628,7 +642,12 @@ public void onClose(
             ));
       }
 
-      StreamsErrorCollector.notifyApplicationClose(applicationId);
+      final Object o =
+          query.getStreamsProperties()
+              .get(KsqlConfig.KSQL_INTERNAL_STREAMS_ERROR_COLLECTOR_CONFIG);
+      if (o instanceof StreamsErrorCollector) {
+        ((StreamsErrorCollector) o).cleanup();
+      }
```
vvcephei, line 650: This is a bit mysterious, but since we are no longer relying on a static map of application ids to find the error collector, we need to get the reference to the specific collector we passed in to this Streams app. The clean thing to do would have been to add the error collector to the QueryMetadata, but I was staring to get pretty concerned about increasing the scope of changes in this PR.

Instead, we can just get the instance out of the config map the exact same way that Streams does to configure the deserialization exception handler.

ksqldb-engine/src/main/java/io/confluent/ksql/internal/StorageUtilizationMetricsReporter.java
```diff
@@ -49,38 +48,30 @@
       = LoggerFactory.getLogger(StorageUtilizationMetricsReporter.class);
   private static final String METRIC_GROUP = "ksqldb_utilization";
 
-  private final Map<String, Map<String, TaskStorageMetric>> metricsSeen;
-  private final Metrics metricRegistry;
-  static AtomicBoolean registeredNodeMetrics = new AtomicBoolean(false);
```
vvcephei, line 54: It looks like this flag was a hack to fix some of the tests that were clashing on the static metrics registry. It's not needed anymore.

ksqldb-engine/src/main/java/io/confluent/ksql/internal/StorageUtilizationMetricsReporter.java
```diff
@@ -49,38 +48,30 @@
...
-    this.metricRegistry = Objects.requireNonNull(metricRegistry);
   }
 
   @Override
   public void init(final List<KafkaMetric> list) {
   }
 
   @Override
-  public void configure(final Map<String, ?> map) {
+  public synchronized void configure(final Map<String, ?> map) {
```
vvcephei, line 63: Added synchronized to resolve an "inconsistent synchronization" warning. I'm pretty sure it's not necessary, but it's also not harmful.

ksqldb-engine/src/main/java/io/confluent/ksql/query/QueryBuilder.java
```diff
@@ -638,6 +658,18 @@ private RuntimeBuildContext buildContext(
...
+    // Passing shared state into managed components
+    newStreamsProperties.put(KsqlConfig.KSQL_INTERNAL_METRIC_COLLECTORS_CONFIG, metricCollectors);
+    newStreamsProperties.put(
+        KsqlConfig.KSQL_INTERNAL_METRICS_CONFIG,
+        metricCollectors.getMetrics()
+    );
+    newStreamsProperties.put(
+        KsqlConfig.KSQL_INTERNAL_STREAMS_ERROR_COLLECTOR_CONFIG,
+        new StreamsErrorCollector(applicationId, metricCollectors)
+    );
```
vvcephei, line 671: Ok, here's where we're adding our shared objects to the Streams config. AK will pass them through to the Streams, Consumer, and Producer interceptors and handlers that need them.

ksqldb-engine/src/test/java/io/confluent/ksql/connect/supported/JdbcSourceTest.java
```diff
@@ -21,7 +21,7 @@
 import com.google.common.collect.ImmutableMap;
 import io.confluent.ksql.connect.Connector;
 import io.confluent.ksql.metastore.model.DataSource.DataSourceType;
-import io.confluent.ksql.metastore.model.MetaStoreMatchers.OptionalMatchers;
+import io.confluent.ksql.test.util.OptionalMatchers;
```
vvcephei, line 24: I moved this matcher up the project dependency graph so I could use it in another test.

ksqldb-engine/src/test/java/io/confluent/ksql/function/UdfLoaderTest.java
```diff
@@ -635,7 +636,12 @@ public void shouldUseConfigForExtDir() {
         .build();
     final KsqlConfig config
         = new KsqlConfig(configMap);
-    UserFunctionLoader.newInstance(config, functionRegistry, "").load();
+    UserFunctionLoader.newInstance(
+        config,
+        functionRegistry,
+        "",
+        new Metrics()
```
vvcephei, line 643: I'm just liberally creating new registries in the tests where it doesn't really matter what they are. We might be able to share some instances, but I'm in the mood to keep it simple at this point in the PR.

ksqldb-engine/src/test/java/io/confluent/ksql/internal/StorageUtilizationMetricsReporterTest.java
```diff
@@ -54,17 +53,17 @@
 
   @Before
   public void setUp() {
-    listener = new StorageUtilizationMetricsReporter(metrics);
+    listener = new StorageUtilizationMetricsReporter();
+    listener.configure(
+        ImmutableMap.of(
+            KsqlConfig.KSQL_INTERNAL_METRICS_CONFIG, metrics
+        )
+    );
```
vvcephei, line 61: Rather than creating special constructors for only the tests to use, why not just have the tests call `configure` the same way the production code does?

ksqldb-engine/src/test/java/io/confluent/ksql/internal/StorageUtilizationMetricsReporterTest.java
```diff
@@ -54,17 +53,17 @@
...
     when(metrics.metricName(any(), any(), (Map<String, String>) any())).thenAnswer(
       a -> new MetricName(a.getArgument(0), a.getArgument(1), "", a.getArgument(2)));
     StorageUtilizationMetricsReporter.setTags(BASE_TAGS);
   }
 
-  @After
-  public void cleanup() {
-    StorageUtilizationMetricsReporter.reset();
```
vvcephei, line 65: I don't remember if I pointed it out before, but you'll notice lots of tests had `initialize()` and `reset()` calls that were an attempt to fix static registry clashes. None of those are needed anymore.

ksqldb-engine/src/test/java/io/confluent/ksql/internal/StorageUtilizationMetricsReporterTest.java
```diff
@@ -94,17 +93,6 @@ public void shouldAddNodeMetricsOnConfigure() throws IOException {
...
-  public void shouldOnlyAddNodeMetricsOnce() {
-    // Given:
-
-    // When:
-    listener.configure(ImmutableMap.of("state.dir", "/tmp//"));
-
-    // Then:
-    // did not throw error for duplicate metrics
-  }
-
```
vvcephei, line 107: I think at some point in the past, this was verifying that two listeners wouldn't try to register the same metrics in the static registry. At the time of this change, though, I guess the implementation of the listener had changed, and this test was actually verifying nothing. This is a good reminder to add some kind of positive check to such tests to make sure they're really validating what they think they are.

Regardless, the whole test is a truism now that we don't have a static registry and additionally now that sensor registration is idempotent.

ksqldb-metastore/src/test/java/io/confluent/ksql/metastore/model/MetaStoreMatchers.java
```diff
@@ -83,36 +83,5 @@ protected SqlType featureValueOf(final Column actual) {
     }
   }
 
-  public static final class OptionalMatchers {
```
vvcephei, line 86: moved to test-utils

ksqldb-rest-app/src/test/java/io/confluent/ksql/rest/entity/SourceDescriptionFactoryTest.java
```diff
@@ -97,49 +99,52 @@ public void shouldReturnLocalStatsBasedOnKafkaTopic() {
...
-          Collections.emptyList(),
-          Collections.emptyList(),
-          Optional.empty(),
-          Collections.emptyList(),
-          Collections.emptyList(),
-          Stream.empty(),
-          Stream.empty(),
-          localhost
-      );
+    Mockito.when(mock.getAndFormatStatsFor(anyString(), anyBoolean()))
```
vvcephei, line 104: The static mock just turns into a regular one.

ksqldb-rest-app/src/test/java/io/confluent/ksql/rest/entity/SourceDescriptionFactoryTest.java
```diff
@@ -161,39 +166,39 @@ public void testShouldIncludeRemoteStatsIfProvided() {
             x)
         ).collect(Collectors.toList());
 
-    try (MockedStatic<MetricCollectors> mc = mockStatic(MetricCollectors.class)) {
```
vvcephei, line 164: switch to a non-static mock

ksqldb-streams/src/main/java/io/confluent/ksql/execution/streams/metrics/RocksDBMetricsCollector.java
```diff
@@ -87,20 +87,14 @@
...
-  private final Metrics metrics;
-
-  public RocksDBMetricsCollector() {
-    this(MetricCollectors.getMetrics());
-  }
-
-  @VisibleForTesting
-  RocksDBMetricsCollector(final Metrics metrics) {
-    this.metrics = Objects.requireNonNull(metrics);
-  }
```
vvcephei, line 99: The metrics object is just set in `configure` now. All test and production code uses the default 0-arg constructor.

ksqldb-test-util/src/main/java/io/confluent/ksql/test/util/KafkaEmbedded.java
```diff
@@ -325,7 +325,8 @@ void waitForTopicsToBeAbsent(final String... topicNames) {
   public Map<String, Integer> getPartitionCount(final Collection<String> topics) {
     try (AdminClient adminClient = adminClient()) {
       final DescribeTopicsResult describeTopicsResult = adminClient.describeTopics(topics);
-      final Map<String, TopicDescription> topicDescriptionMap = describeTopicsResult.all().get();
+      final Map<String, TopicDescription> topicDescriptionMap =
+          describeTopicsResult.allTopicNames().get();
```
vvcephei, line 329: deprecation warning

ksqldb-test-util/src/main/java/io/confluent/ksql/test/util/OptionalMatchers.java
```diff
@@ -0,0 +1,54 @@
...
+ */
+
+package io.confluent.ksql.test.util;
+
+import java.util.Optional;
+import org.hamcrest.Description;
+import org.hamcrest.Matcher;
+import org.hamcrest.TypeSafeDiagnosingMatcher;
+
+public final class OptionalMatchers {
```
vvcephei, line 23: moved from MetaStoreMatchers

vvcephei (commented):
Hey @confluentinc/ksql ,

I spent hours debugging a failing test in two PRs yesterday that both turned out not to be problems with the PR, but just metrics getting cached in my JVM. I was getting problems running on the CLI with Maven (because the classloader is running tests in a different order than Jenkins, I guess) and also with IDEA (because the loaded classes were not all unloaded between test runs, so independent runs would pollute each others' static state).

This is the third occasion I've bumped into this problem, so I think it's time to just fix it.

I was planning to break this up into several PRs if it was too big. While this PR is borderline, it's not insane, so I figured it makes more sense in one chunk.

ksqldb-rest-app/src/test/java/io/confluent/ksql/rest/server/execution/DescribeConnectorExecutorTest.java
```diff
@@ -173,6 +176,7 @@ public void shouldDescribeKnownConnector() {
 
     // Then:
     verify(engine).getMetaStore();
+    verify(engine).metricCollectors();
```
vvcephei, line 179: it's a bit silly that we have to verify interactions we don't care about, but I've fought enough battles for today.

ksqldb-rest-app/src/test/java/io/confluent/ksql/rest/server/execution/PullQueryMetricsTest.java
```diff
@@ -99,9 +98,9 @@ public void shouldRecordNumberOfLocalRequests() {
 
     // Then:
     assertThat(value, equalTo(1.0));
-    assertThat(rate, closeTo(0.03, 0.001));
+    assertThat(rate, greaterThan(0.0));
```
vvcephei, line 101: These tests were failing because the measured rates were slightly outside of the tolerances of the prior test. All we really care about is that the rate metric got registered and recorded, so `>0` is sufficient. I think the measurement changed just due to the difference in the amount of time that passed between the creation of the static registry before and this test's registry instance now.

ksqldb-examples/src/main/java/io/confluent/ksql/embedded/EmbeddedKsql.java
```diff
@@ -28,7 +29,8 @@ public static void main(final String[] args) {
 
     final KsqlContext ksqlContext = KsqlContext.create(
         new KsqlConfig(Collections.emptyMap()),
-        ProcessingLogContext.create());
+        ProcessingLogContext.create(),
+        new MetricCollectors());
```
vvcephei, line 33: Hmm, this file is in `ksqldb-examples`, which raises the question of whether this is a public API. I can easily keep the old constructor for backward compatibility. Should I?

Also, is there anything else here that's a public API?

ksqldb-engine/src/main/java/io/confluent/ksql/query/QueryBuilder.java
```diff
@@ -605,7 +624,7 @@ private RuntimeBuildContext buildContext(
     );
   }
 
-  public static Map<String, Object> updateListProperties(
+  public static void updateListProperties(
```
vvcephei, line 627: It's confusing to return the same map that's provided as an argument, since it implies we're not modifying the argument map when we in fact are.

ksqldb-engine/src/main/java/io/confluent/ksql/query/QueryBuilder.java
vvcephei: @tolgadur / @wcarlson5 , 

These are necessary to be able to start up Streams. I couldn't tell from the code in https://github.com/confluentinc/ksql/pull/8481 whether you're re-using the same config from the running app or building a new one from scratch. If it's the latter, then you're going to need this block as well. Which leads me to the question of why we would duplicate the contents of `buildStreamsProperties` to begin with instead of just calling it.

I didn't see any discussion of that point in #8481.

ksqldb-engine/src/main/java/io/confluent/ksql/query/QueryRegistryImpl.java
```diff
@@ -231,9 +239,8 @@ public void restartStreamsRuntime(final KsqlConfig config) {
       final Map<String, Object> newStreamsProperties =
           new HashMap<>(config.getKsqlStreamConfigProps(applicationId));
       newStreamsProperties.put(StreamsConfig.APPLICATION_ID_CONFIG, applicationId);
-      final Map<String, Object> properties = QueryBuilder
-          .updateListProperties(newStreamsProperties);
-      stream.overrideStreamsProperties(properties);
+      QueryBuilder.updateListProperties(newStreamsProperties);
+      stream.overrideStreamsProperties(newStreamsProperties);
```
vvcephei, line 243: note that previously, `properties` was the exact same object as `newStreamsProperties`.

vvcephei (commented):
Hey @tolgadur , @wcarlson5  , I've changed some stuff you just added in https://github.com/confluentinc/ksql/pull/8481 . 

Can you take a look and make sure I didn't break something? I re-ran the tests from your PR, but I didn't see any integration tests for restarting Streams.

agavra (approved):
LGTM! Huge thanks @vvcephei this makes life way better ( even if the passing in the metrics collectors via config is a bit magical ). Can we test that a simple local deployment still deploys some of the streams metrics via JMX? If that works then I'm confident in this change.

Otherwise, only comment is re: sandboxing but I think that can be addressed in a different PR since I think this behavior is identical to what it was previously with the static registry.

ksqldb-common/src/main/java/io/confluent/ksql/metrics/StreamsErrorCollector.java
```diff
@@ -38,36 +39,29 @@
...
-        .increment(null, true);
+  @SuppressFBWarnings(
+      value = "EI_EXPOSE_REP2",
+      justification = "metrics"
+  )
+  public StreamsErrorCollector(final String applicationId, final MetricCollectors collectors) {
+    this.metricCollectors = collectors;
+    this.metrics = collectors.getMetrics();
+    this.time = collectors.getTime();
+    this.id = collectors.addCollector(applicationId, this);
```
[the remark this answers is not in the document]
vvcephei: That's a much better idea. I'm not sure what I was thinking. I'll fix that in a follow-up (to avoid waiting for ITs again) unless there's another change I have to do in this PR.

ksqldb-engine/src/main/java/io/confluent/ksql/query/QueryBuilder.java
wcarlson5: @vvcephei Thanks for letting us know about this. I think your changes shouldn't break our feature. There are still more changes coming on top of this so don't hold of merging on our account

ksqldb-engine/src/main/java/io/confluent/ksql/engine/SandboxedExecutionContext.java
```diff
@@ -59,11 +60,14 @@
...
   private final EngineContext engineContext;
+  private final MetricCollectors metricCollectors;
 
   SandboxedExecutionContext(
       final EngineContext sourceContext,
-      final ServiceContext serviceContext
+      final ServiceContext serviceContext,
+      final MetricCollectors metricCollectors
   ) {
+    this.metricCollectors = metricCollectors;
```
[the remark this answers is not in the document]
vvcephei: Yeah, we can do that in a follow-up. The dependency that forced me to add the metrics to the execution context in the first place does imply that sandbox usages could affect the runtime metrics.

guozhangwang (approved):
LGTM (I only looked through the non-testing part of it though.. shamelessly trusting that @agavra have had them all). I like the refactoring especially when it can help removing those synchronizations.

vvcephei:
Gah. There's another merge conflict. Thanks for the reviews, all! I'm going to resolve those conflicts and then merge as soon as the tests pass.

vvcephei:
Also, just to double-check, I did run the ksql server locally and went through the quickstart, then checked the metrics beans in JConsole. Everything looks like it still works end-to-end. Regardless, I'll be extra vigilent as we go through the release process for the next release.

agavra (approved):
Excited enough for this change I thought it was worth green checking twice. Thanks John!

vvcephei:
Thanks again for the reviews, everyone!

ksqldb-engine/src/main/java/io/confluent/ksql/engine/EngineContext.java
```diff
@@ -96,7 +98,7 @@ static EngineContext create(
         new DefaultKsqlParser(),
         cleanupService,
         ksqlConfig,
-        new QueryRegistryImpl(registrationListeners)
+        new QueryRegistryImpl(registrationListeners, metricCollectors)
```
rodesai, line 101: why does this need the whole collector?

ksqldb-engine/src/main/java/io/confluent/ksql/internal/StorageUtilizationMetricsReporter.java
```diff
@@ -49,38 +48,30 @@
...
+    this.metricRegistry = (Metrics) requireNonNull(
+        map.get(KsqlConfig.KSQL_INTERNAL_METRICS_CONFIG)
+    );
+    this.metricsSeen = new HashMap<>();
```
rodesai, line 67: nit: can we just initialize this from the constructor and make it final?

rodesai (commented):
LGTM FWIW (couple small nits inline)

ksqldb-engine/src/main/java/io/confluent/ksql/engine/EngineContext.java
rodesai: nvm - I see why now.

ksqldb-engine/src/main/java/io/confluent/ksql/query/QueryBuilder.java
tolgadur: We are rebuilding the config from scratch. I will open another PR to add it.

PR #8501, opened 2021-12-13
chore: Revert "fix: explicitly set confluent log4j version to 1.2.17-cp5"

nateab:
Reverts confluentinc/ksql#8497

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -632,7 +632,7 @@
         <dependency>
             <groupId>io.confluent</groupId>
             <artifactId>confluent-log4j</artifactId>
-            <version>1.2.17-cp5</version>
+            <version>1.2.17-cp6</version>
             <exclusions>
                 <exclusion>
                     <groupId>org.apache.logging.log4j</groupId>
--- a/pom.xml
+++ b/pom.xml
@@ -613,35 +613,37 @@
                 <version>${wiremock.version}</version>
                 <scope>test</scope>
             </dependency>
-
-            <!-- Every child project uses slf4j/log4j for logging, might as well just include them here automatically -->
-            <dependency>
-                <groupId>org.slf4j</groupId>
-                <artifactId>slf4j-api</artifactId>
-            </dependency>
-
-            <dependency>
-                <groupId>org.slf4j</groupId>
-                <artifactId>slf4j-log4j12</artifactId>
-            </dependency>
-
-            <!-- Use a repackaged version of log4j with security patches. Default log4j v1.2 is a transitive dependency of slf4j-log4j12, but it is excluded in common/pom.xml -->
-            <dependency>
-                <groupId>io.confluent</groupId>
-                <artifactId>confluent-log4j</artifactId>
-                <exclusions>
-                    <exclusion>
-                        <groupId>org.apache.logging.log4j</groupId>
-                        <artifactId>log4j-core</artifactId>
-                    </exclusion>
-                    <exclusion>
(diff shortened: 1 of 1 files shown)


cadonna:
@nateab Why do you want to revert the change? It seemed reasonable to me. If we want to explicitly exclude log4j2 then `dependencyManagement` seems the right way to do that.

https://maven.apache.org/guides/introduction/introduction-to-dependency-mechanism.html

nateab:
@cadonna i want to revert my change which reverted your change 😅  because I thought based on the thread with MIlo that exclusion didnt work. But so this PR brings back your change, and is what is in `0.23.1-rc9` right now.

cadonna:
According to https://maven.apache.org/guides/introduction/introduction-to-optional-and-excludes-dependencies.html:

> Exclusions work on the entire dependency graph below the point where they are declared.

So I guess this PR might be OK. However, I am wondering whether it would be an improvement to put the logging dependencies into the `<dependencyManagement>` tag as all other dependencies. Apparently, that tag takes precedence on dependencies specified in the child projects.

nateab:
Yeah I am fine with moving this into `dependencyManagement` if you think thats better

nateab:
Hmm when I try to move it, jenkins complains for some reason 
` [ERROR] Failed to execute goal org.codehaus.mojo:versions-maven-plugin:2.8.1:resolve-ranges (default-cli) on project ksqldb-parent: Execution default-cli of goal org.codehaus.mojo:versions-maven-plugin:2.8.1:resolve-ranges failed. NullPointerException`
No idea why. Maybe we should just keep it as it was before then?

cadonna:
Could you please try with
```
            <!-- Use a repackaged version of log4j with security patches. Default log4j v1.2 is a transitive dependency of slf4j-log4j12, but it is excluded in common/pom.xml -->
            <dependency>
                <groupId>io.confluent</groupId>
                <artifactId>confluent-log4j</artifactId>
                <version>${confluent-log4j.version}</version>
                <exclusions>
                    <exclusion>
                        <groupId>org.apache.logging.log4j</groupId>
                        <artifactId>log4j-core</artifactId>
                    </exclusion>
                    <exclusion>
                        <groupId>org.apache.logging.log4j</groupId>
                        <artifactId>log4j-api</artifactId>
                    </exclusion>
                </exclusions>
            </dependency>
```
(put it where all Confluent dependencies are defined, before `<!-- End Confluent dependencies -->`)

and
```
            <!-- Every child project uses slf4j/log4j for logging, might as well just include them here automatically -->
            <dependency>
                <groupId>org.slf4j</groupId>
                <artifactId>slf4j-api</artifactId>
                <version>${slf4j.version}</version>
            </dependency>

            <dependency>
                <groupId>org.slf4j</groupId>
                <artifactId>slf4j-log4j12</artifactId>
                <version>${slf4j.version}</version>
            </dependency> 
```

cadonna:
@nateab It seems my proposal fixes the issue you posted but it is also seems a bit more of a hassle because this affects dependency specifications in the child projects. Let's not moving the logger dependecies into `dependencyManagement` for now. I will try to move it in a separate PR.

agavra (approved):

agavra:
@nateab are we going to merge this?

nateab:
@agavra oh huh i didnt see that jenkins passed. Though i have already added this change to another open PR  #8571  (along with some more exclusions I had to do in order to fully exclude the log4j 2 jars). I can close this one out though and just merge the other one

Issue #8505, opened 2021-12-13
Allow Scalable Push Queries to take advantage of http2 multiplexing


AlanConfluent:
**Is your feature request related to a problem? Please describe.**
Yes.  We had to disable multiplexing because there's not currently a way to complete/cancel a request once it's been started from the client side.  The codebase currently does `response.request().connection().close();` (in `StreamPublisher`) which is fine for http1, but closes any shared connections in http2, failing unrelated queries.  I tried a change to make it `response.request().reset()`, but this also seems to shutdown the connection if it's for the first request using the http2 connection (though seemingly not others).  

**Describe the solution you'd like**
We need to figure out how to ensure that we can cancel a request from the client side after the request has been initiated.  Maybe something like `response.request().reset()`, but in a way that doesn't close the connection.

**Describe alternatives you've considered**
- A separate endpoint `/close` which will allow async closing of existing connections.

**Additional context**

PR #8507, opened 2021-12-13
fix: Prevents internal http2 requests from having shared connections closed

AlanConfluent:
### Description 
Related to #8505.  Since we currently close connections upon being finished with a request, this disables multiplexing so that shared connections are not closed, affecting other requests.

### Testing done 
Ran locally.

### Reviewer checklist

### DIFF
--- a/ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/KsqlRestConfig.java
+++ b/ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/KsqlRestConfig.java
@@ -383,7 +383,7 @@ public class KsqlRestConfig extends AbstractConfig {
 
   public static final String KSQL_INTERNAL_HTTP2_MAX_POOL_SIZE_CONFIG
       = "ksql.internal.http2.max.pool.size";
-  public static final int KSQL_INTERNAL_HTTP2_MAX_POOL_SIZE_DEFAULT = 1000;
+  public static final int KSQL_INTERNAL_HTTP2_MAX_POOL_SIZE_DEFAULT = 3000;
   public static final String KSQL_INTERNAL_HTTP2_MAX_POOL_SIZE_DOC =
       "The maximum connection pool size used by Vertx for http2 internal connections";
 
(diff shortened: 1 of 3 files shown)


ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/KsqlRestConfig.java
```diff
@@ -381,6 +381,12 @@
           + " otherwise be verbose. Note that this works on the entire URI, respecting the "
           + KSQL_ENDPOINT_LOGGING_LOG_QUERIES_CONFIG + " configuration";
 
+  public static final String KSQL_INTERNAL_HTTP2_MAX_POOL_SIZE_CONFIG
+      = "ksql.internal.http2.max.pool.size";
+  public static final int KSQL_INTERNAL_HTTP2_MAX_POOL_SIZE_DEFAULT = 1000;
```
agavra, line 386: whats the risk of setting this even higher, even (say) 10k? I know we've been defensive before and it's often just bit us unnecessarily. I'm wondering if this could be the case here.

agavra (approved):

ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/KsqlRestConfig.java
vvcephei: There are other limits (the port range that the kernel sets aside for listening connections, the file handle limit, etc.). We wanted to shoot for a limit that's high enough to permit our use case, but low enough that we will hit our own limit before some weird and hard-to-debug subsystem limit.

Personally, I really hope we can find a "better" solution to this problem sooner than later, but I also agree we should patch this problem asap while we seek a "proper" solution.

vvcephei (approved):
Thanks, @AlanConfluent !

Not to be "that guy", but would it be appropriate to have a test for this logic? It seems like it might be a bummer to try and set this config in an emergency later only to discover that it doesn't work for some silly reason.

ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/KsqlRestConfig.java
AlanConfluent: It depends on the settings of the machine it runs on.  I checked some for reference and they were about 1M files open allowed, so I think we could go higher. Have some value in mind?

AlanConfluent:
> Thanks, @AlanConfluent !
> 
> Not to be "that guy", but would it be appropriate to have a test for this logic? It seems like it might be a bummer to try and set this config in an emergency later only to discover that it doesn't work for some silly reason.

Wrote a test for it!

ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/KsqlRestConfig.java
agavra: Let's keep it at 3k, seems reasonable to allow 1.5k push queries on a 3 node cluster

PR #8509, opened 2021-12-14
fix: invoke per-query listeners upon state change in shared runtime

ableegoldman:
Fix for the query-level metrics such as [Query Status](https://ccloud-production.datadoghq.com/dashboard/m59-7gm-x4j/ksql-node-health?fullscreen_end_ts=1639468089988&fullscreen_paused=false&fullscreen_section=overview&fullscreen_start_ts=1636876089988&fullscreen_widget=8638173689799481&from_ts=1636876064877&to_ts=1639468064877&live=true) which aren't coming through due to the fact that we aren't registering a `StateListener` on the shared `KafkaStreams` runtime. This patch sets a `StateListener` that loops over all queries in the runtime and invokes the `#onStateChange` callback on their individual query listeners.

Also includes a miscellaneous set of minor fixes/cleanups in the BinPackedPersistentQueryMetadataImpl and SharedKafkaStreamsRuntime interface/subclasses

### DIFF
--- a/ksqldb-engine/src/main/java/io/confluent/ksql/util/BinPackedPersistentQueryMetadataImpl.java
+++ b/ksqldb-engine/src/main/java/io/confluent/ksql/util/BinPackedPersistentQueryMetadataImpl.java
@@ -236,11 +236,6 @@ public class BinPackedPersistentQueryMetadataImpl implements PersistentQueryMeta
   }
 
   @Override
-  public Optional<MaterializationProvider> getMaterializationProvider() {
-    return materializationProvider;
-  }
-
-  @Override
   public Optional<ScalablePushRegistry> getScalablePushRegistry() {
     return scalablePushRegistry;
   }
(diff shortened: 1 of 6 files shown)


ksqldb-engine/src/main/java/io/confluent/ksql/util/BinPackedPersistentQueryMetadataImpl.java
```diff
@@ -234,11 +235,6 @@ public void stop() {
                                                 + " individual queries in shared runtime");
   }
 
-  @Override
```
wcarlson5, line 237: Was this just unused? We may want to look into why if so. that might affect some functionality

wcarlson5 (approved):
This is what we talked about. TY!

Once the build passes you should be good to merge

ksqldb-engine/src/main/java/io/confluent/ksql/util/BinPackedPersistentQueryMetadataImpl.java
ableegoldman: I'll investigate quickly to make sure, but the fact that it was also unused in `QueryMetadataImpl` leads me to believe it was just an unused getter for a field that is only needed within the class (which jives with my understanding of what the `materializationProvider` is/does)

PR #8510, opened 2021-12-14
fix: refactoring properties replacement before restarting runtime

tolgadur:
### Description 
After https://github.com/confluentinc/ksql/pull/8498 we needed to do some refactoring for the restart runtime.

### Testing done 
Unit tests

### Open questions 
- Is there any way we can do an end to end test to verify this works?

### Reviewer checklist

### DIFF
--- a/ksqldb-engine/src/main/java/io/confluent/ksql/query/QueryBuilder.java
+++ b/ksqldb-engine/src/main/java/io/confluent/ksql/query/QueryBuilder.java
@@ -193,8 +193,10 @@ final class QueryBuilder {
 
     final Map<String, Object> streamsProperties = buildStreamsProperties(
         applicationId,
-        queryId,
-        metricCollectors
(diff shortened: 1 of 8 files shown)


ksqldb-engine/src/main/java/io/confluent/ksql/query/QueryBuilder.java
```diff
@@ -191,10 +191,12 @@ TransientQueryMetadata buildTransientQuery(
         streamsBuilder
     );
 
-    final Map<String, Object> streamsProperties = buildStreamsProperties(
+    final Map<String, Object> streamsProperties = QueryBuilder.buildStreamsProperties(
```
wcarlson5, line 194: super nit: You don't need the `QueryBuilder.` thing. but at least have it consistent with line 280

wcarlson5 (approved):
Just one nit

PR #8511, opened 2021-12-14
fix: reject incorrectly typed HEADER columns

### LINKED ISSUE #8447: Creating a stream with HEADERS does not fail if I specify a key/value type other than STRING/BYTES
**Describe the bug**
In the following CREATE statement, I used a `hdrs array<struct<key int, value int>> headers` column that has `value` as int. An error should have been thrown 'cause we support only bytes in headers columns.

```
ksql> create stream s1(key string key, message string, hdrs array<struct<key int, value int>> headers) with (kafka_topic='topic1', value_format='delimited');

 Message        
----------------
 Stream created 
----------------
ksql> select key, message, hdrs as appId from s1 emit changes;
+-------------------------------------------------------------+-------------------------------------------------------------+-------------------------------------------------------------+
|KEY                                                          |MESSAGE                                                      |APPID                                                        |
+-------------------------------------------------------------+-------------------------------------------------------------+-------------------------------------------------------------+
|1                                                            |message1                                                     |[{KEY=appName, VALUE=a2Fma2FjYXQ=}, {KEY=appId, VALUE=MQ==}] |
|2                                                            |message2                                                     |[{KEY=appName, VALUE=a2Fma2FjYXQ=}, {KEY=appId, VALUE=MQ==}] |
|3                                                            |messag3                                                      |[{KEY=appName, VALUE=a2Fma2FjYXQ=}, {KEY=appId, VALUE=Mg==}] |
|4                                                            |message4                                                     |[{KEY=appId, VALUE=Mw==}]                                    |
|5                                                            |message5                                                     |[]                                                           |
|6                                                            |message6                                                     |[{KEY=appId, VALUE=Mw==}]                                    |
|7                                                            |message7                                                     |[{KEY=appId, VALUE=Mw==}, {KEY=appId, VALUE=NA==}]           |
^CQuery terminated
```

jzaralim:
### Description 
Fixes #8447 

Throws an error if a user tries to create a source with a HEADER column with an incorrect type

### Testing done 
QTT and unit tests

### Reviewer checklist

### DIFF
--- a/ksqldb-parser/src/main/java/io/confluent/ksql/parser/tree/TableElements.java
+++ b/ksqldb-parser/src/main/java/io/confluent/ksql/parser/tree/TableElements.java
@@ -15,17 +15,21 @@
 
 package io.confluent.ksql.parser.tree;
 
+import static io.confluent.ksql.schema.ksql.SystemColumns.HEADERS_TYPE;
+
 import com.google.common.collect.ImmutableList;
 import com.google.errorprone.annotations.Immutable;
 import io.confluent.ksql.name.ColumnName;
 import io.confluent.ksql.schema.ksql.LogicalSchema;
 import io.confluent.ksql.schema.ksql.LogicalSchema.Builder;
 import io.confluent.ksql.schema.ksql.types.SqlType;
+import io.confluent.ksql.schema.ksql.types.SqlTypes;
 import io.confluent.ksql.util.KsqlException;
 import java.util.Iterator;
 import java.util.List;
 import java.util.Map.Entry;
 import java.util.Objects;
+import java.util.Optional;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
 import java.util.stream.StreamSupport;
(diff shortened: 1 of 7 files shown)


ksqldb-parser/src/main/java/io/confluent/ksql/parser/tree/TableElements.java
```diff
@@ -91,6 +95,9 @@ public LogicalSchema toLogicalSchema() {
       if (tableElement.getConstraints().isKey() || tableElement.getConstraints().isPrimaryKey()) {
         builder.keyColumn(fieldName, fieldType);
       } else if (tableElement.getConstraints().isHeaders()) {
+        throwOnIncorrectHeaderColumnType(
```
spena, line 98: I think this check should be done earlier in the `AstBuilder` where the `TableElement` is created. My opinion is that once `TableElement` is created, then `toLogicalSchema` shouldn't fail. This method just converts from one object to another (unsure why isEmpty() fails, but it's already there). Also, the `AstBuilder` may check for invalid syntax, so why I think it should live there. What do you think?

jzaralim: That makes sense. Moved.

spena (approved):
LGTM

Issue #8512, opened 2021-12-14
Unable to use separate certificate for Kafka mTLS and https


piotrsmolinski:
**Describe the bug**
In my scenario the Kafka brokers are configured with mTLS and HTTPS is mandatory for all REST services. The user in mTLS authentication is extracted from CN field of the certificate subject. The HTTPS certificates on the other hand are delivered for each service instance separately and wildcard certificates are not allowed. The CN of HTTPS certificate is enforced to be FQDN. These requirements impose separate certificates for both connections

For this configuration the natural approach is to set *ssl.keystore.location* to HTTPS certificate file and *ksql.streams.ssl.keystore.location* to Kafka client certificate. The problem is that the result Kafka settings are not overridden with *ksql.streams.* prefixed values, but rather the last value for the final unprefixed key is taken. If the input *originals* happen to offer the *ksql.streams.ssl.keystore.location* last, the final Kafka config seems to be correct. If the prefixed entry comes first in the loop, *ksql.streams.* value is ignored.

**To Reproduce**
1. Use ksqlDB 0.22.0
2. Generate separate certificates for HTTPS and for Kafka. Use different CA and separate trusted chains.
3. Set *ksql.streams.ssl.keystore.location* and *ksql.streams.ssl.truststore.location* (and related passwords) values to Kafka accepted mTLS client certificate.
4. Set *ssl.keystore.location* and *ssl.truststore.location* (and related passwords) values to HTTPS server certificate.

The result is not deterministic. In some configurations the connection to Kafka is established with the correct certificate, while slightly different config leads to HTTPS certificate offered as mTLS client certificate or Kafka server certificate being rejected due to missing trusted chain.

**Expected behavior**
When *ksql.streams.ssl.keystore.location* is set, this value is always used for connections to Kafka. If the value is not set, *ssl.keystore.location* should be used. This applies to other settings as well.

**Actual behaviour**
In the case when we discovered the behaviour, the trusted chains were same for Kafka and HTTPS. The Kafka brokers accepted the offered HTTPS server certificate and claimed authorization failures accessing the requested topics. Authorizer log has shown failure for the Subject of the HTTPS server certificate. Sandboxed tests have shown the behaviour is not consistent and depends on order in which the configuration entries are presented in KsqlConfig#applyStreamsConfig method.

https://github.com/confluentinc/ksql/blob/v0.22.0-ksqldb/ksqldb-common/src/main/java/io/confluent/ksql/util/KsqlConfig.java#L1172-L1182

PR #8513, opened 2021-12-14
Bugfix for #8512

piotrsmolinski:
### Description 
The bugfix introduces a helper method **originalsWithPrefixOverride** similar to existing **valuesWithPrefixOverride**. Contrary to the latter it does not limit the result to defined configuration entries. The new method is used instead of **originals** to provide the values for Kafka configuration in KsqlConfig.

### Testing done 
Unit test added to KsqlConfigTest failing in unmodified class.


piotrsmolinski:
closed as the PR was created from the wrong branch

PR #8523, opened 2021-12-15
feat: Drop query from assignor

wcarlson5:
Then a query is dropped we also want to remove it from the assignor so that runtime can use the sources again. We need to remove it form the KsqlContext

### Reviewer checklist

### DIFF
--- a/ksqldb-engine/src/main/java/io/confluent/ksql/embedded/KsqlContext.java
+++ b/ksqldb-engine/src/main/java/io/confluent/ksql/embedded/KsqlContext.java
@@ -189,7 +189,10 @@ public class KsqlContext implements AutoCloseable {
   }
 
   public void terminateQuery(final QueryId queryId) {
-    ksqlEngine.getPersistentQuery(queryId).ifPresent(QueryMetadata::close);
+    ksqlEngine.getPersistentQuery(queryId).ifPresent(t -> {
+      t.close();
+      ksqlEngine.removeQueryFromAssignor(t);
+    });
   }
 
   private static ExecuteResult execute(
--- a/ksqldb-engine/src/main/java/io/confluent/ksql/embedded/KsqlContext.java
+++ b/ksqldb-engine/src/main/java/io/confluent/ksql/embedded/KsqlContext.java
@@ -188,6 +188,7 @@ public class KsqlContext implements AutoCloseable {
     serviceContext.close();
   }
 
+  @VisibleForTesting
   public void terminateQuery(final QueryId queryId) {
     ksqlEngine.getPersistentQuery(queryId).ifPresent(t -> {
       t.close();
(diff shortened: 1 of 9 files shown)


ableegoldman (approved):
LGTM

ableegoldman:
Mm on a related note, I think we should attempt to drop/delete runtimes if we ever remove all queries from them && have more than the `num.initial.shared.runtimes` config -- not necessarily in this PR, just leaving a note here since you'll also need to drop them from the assignor

ableegoldman:
Retriggered the tests since the spotBugs issue seems to have been fixed

ableegoldman:
Oh I guess I needed to hit `rebuild` instead of `replay`, anyways should get a green build now.. 🤞

ableegoldman:
Nevermind there's a new build issue, Natea is looking into it -- hopefully we can merge this soon since we got all the others in... 😅

ksqldb-engine/src/main/java/io/confluent/ksql/embedded/KsqlContext.java
```diff
@@ -189,7 +189,10 @@ public void close() {
   }
 
   public void terminateQuery(final QueryId queryId) {
```
ableegoldman, line 191: It looks like this actually isn't being called from anywhere outside of test code...at least that's what my IDE is saying. Seems like a problem? Not sure where/how we actually terminate queries, or why we have this method if it's only for tests -- maybe should be marked with `@VisibleForTesting` since it's very much not clear that it isn't used  for "real" queries 😬 

Was poking around because it looks like we aren't handling the CREATE OR REPLACE case properly for https://github.com/confluentinc/ksql/pull/8538, ie we need to remove the old source topics from the RuntimeAssignor when a query is replaced -- we should just need to make sure that we invoke `#dropQuery` and then re-register that query in the `RuntimeAssignor`. Could you look into this?

ableegoldman: If you want to test it out, I added a CREATE OR REPLACE test to EndToEndIntegrationTest in my PR, plus you'll need [this fix] (https://github.com/confluentinc/ksql/pull/8538/files#diff-28490c7f037eb219dd2e44fabbf194d2d8805f895becd6daf48f79bf3ce4f09aR123-R126)for a bug that was causing RuntimeAssignor#stopQuery to throw an exception. Although given that I found the bug in #stopQuery at all, when this is only called from testing code, makes me wonder how useful those integration tests really are for replicating the real thing...

wcarlson5: For now this isn't being used. This PR is what is using this method. Until it is merged then that won't be used. see the file below.

wcarlson5 (commented):
As I found the real call and added it there

wcarlson5:
I just ran the tests locally and they all passed. there is a build issue right not that is not released to this PR so I am just going to merge it for now

ksqldb-engine/src/main/java/io/confluent/ksql/embedded/KsqlContext.java
ableegoldman: Sorry I meant `terminateQuery` not `stopQuery`

PR #8524, opened 2021-12-15
fix: Restore runtime assignor

wcarlson5:
If the validation node goes down or needs to replay the cmd topic we need to pick up the running queries from cmd topic an put them in the assignor. This will make sure that the runtime assignor will always be in sync with the runtimes using the Engine Context. 

### Reviewer checklist


ableegoldman:
> If the validation node goes down

What does this mean? We don't have a separate node running the validation stuff, the API threads and command runner threads are both on the server AFAIK, do you just mean if the server itself goes down? Why do we only need to rebuild for a validation assignor and not the "real" one?

wcarlson5:
Yes I mean if the server the validation is being done on goes down.

Also we rebuild the real assignor not the validation one. We just do it right before the validation one is made because then we will be checking before we build each query

ableegoldman (approved):

PR #8527, opened 2021-12-16
fix: header column names must be unique

jzaralim:
### Description 
Headers columns cannot have the same name as another value or key column, because that introduces ambiguities in some parts of the code. Internally, headers are treated very similarly to pseudocolumns, so this is similar to not being able to have key/value columns named `ROWOFFSET`.

### Testing done 
QTT + unit tests

### Reviewer checklist


jzaralim:
This is already covered in `TableElements`

PR #8529, opened 2021-12-16
fix: only use uuid as alter system command id

tolgadur:
### Description 
The command id for the `alter system` query was using a uuid plus the query in string form. This was leading to the server to crash as the `/` is not allowed as a command id due to the following line `final String[] splitOnSlash = fromString.split("/", 3);`. 

This changes the command id to only use a random uuid as originally planned. 

### Testing done 
manual testing 

### Reviewer checklist

### DIFF
--- a/ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/computation/CommandIdAssigner.java
+++ b/ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/computation/CommandIdAssigner.java
@@ -149,7 +149,7 @@ public class CommandIdAssigner {
   private static CommandId getAlterSystemCommandId(final AlterSystemProperty alterSystemProperty) {
     return new CommandId(
         Type.CLUSTER,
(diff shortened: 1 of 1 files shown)


wcarlson5 (approved):

ableegoldman:
@tolgadur there was a merge conflict when cherrypicking this to the 7.1.x branch, can you do the cherrypicking?

tolgadur:
7.1.x doesn't have any of the `ALTER SYSTEM` changes, including stuff like https://github.com/confluentinc/ksql/pull/8334 which was merged 14 days ago. Are we sure this is the branch we want?

wcarlson5:
https://github.com/confluentinc/ksql/pull/8531 you should check that out to see what is going on

agavra:
@wcarlson5 @tolgadur @ableegoldman the merge hasn't gone in yet, we'll need to wait for the bulk merge to go in and then we can cherry-pick the commits onto 7.1.x and pint merge them back

PR #8531, opened 2021-12-16
merge master into 7.1.x

agavra:
### Description 

### Testing done 

### Reviewer checklist


vcrfxia (commented):
We need to back out the pom.xml version bumps, else this merge will break 7.1.x.

agavra:
@vcrfxia good call out! thanks

PR #8532, opened 2021-12-16
feat: don't book keep sources in runtime as well

wcarlson5:
Now that we have the runtime assignor we don't need to keep checking the sources in the runtime. There will not be queries assigned to `SharedKafkaStreamRuntime` with the same sources. If by some bug there is the add will fail.
 
### Reviewer checklist

### DIFF
--- a/ksqldb-engine/src/main/java/io/confluent/ksql/query/QueryBuilder.java
+++ b/ksqldb-engine/src/main/java/io/confluent/ksql/query/QueryBuilder.java
@@ -430,8 +430,7 @@ final class QueryBuilder {
         valueFormat.getFeatures()
     );
 
-    final NamedTopologyBuilder namedTopologyBuilder =
-        ((KafkaStreamsNamedTopologyWrapper) sharedKafkaStreamsRuntime.getKafkaStreams())
+    final NamedTopologyBuilder namedTopologyBuilder = sharedKafkaStreamsRuntime.getKafkaStreams()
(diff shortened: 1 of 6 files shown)


ableegoldman (approved):
LGTM

ableegoldman:
compiler error, need to remove calls to `markSources` in SharedKafkaStreamsRuntimeImplTest and SandboxedSharedKafkaStreamsRuntimeImplTest

> If by some bug there is the add will fail.

Is this true? Where do we do this checking now -- or do you mean we just rely on the `#addNamedTopology` to fail? I'm not sure that actually works like that -- although I can put that into one of my PRs, to complete the future exceptionally if topics of the same name exist

ableegoldman:
Anyways I think we will need to do some bookkeping of source topics anyways in order to throw a clear error message in the CREATE OR REPLACE case, although even so I would agree it makes more sense to do so in the RuntimeAssignor than in the actual runtime itself.

wcarlson5:
If you want I can hold off on merging this until we finish the Create or replace testing?

ableegoldman:
Oh sorry I just checked the code and yeah, I guess we do throw an exception from `#addNamedTopology` directly which will still cause it to fail. So nevermind about that. 

If you can just add a test for CREATE OR REPLACE to make sure it still fails if the new topic(s) overlap then I'm happy with merging this now -- we should have test coverage to make sure before doing so

PR #8537, opened 2021-12-17
chore: Switch over to using QueryStreamHandler for all queries on /query

AlanConfluent:
### Description 
Switches `/query` to be handled by the handler `QueryStreamHandler` when the config `ksql.endpoint.migrate.query` is set.  The defaults should be the same in this scenario, including format.

### Testing done 
Ran unit and functional tests.

### Reviewer checklist


cprasad1 (approved):
LGTM!

guozhangwang (approved):
LGTM!

PR #8538, opened 2021-12-17
chore: display error with workaround when CREATE OR REPLACE fails due to overlapping sources

ableegoldman:
A limitation in V1 of the query scaling engine will prevent users from issuing CREATE OR REPLACE commands that update the source topics of a query if those topics happen to already be consumed by another query in the same runtime. This should be a very rare case but if a user does hit it, we should make sure to explain why this failed and suggest a workaround for updating the query in the error message.

### DIFF
--- a/ksqldb-engine/src/main/java/io/confluent/ksql/engine/RuntimeAssignor.java
+++ b/ksqldb-engine/src/main/java/io/confluent/ksql/engine/RuntimeAssignor.java
@@ -120,8 +120,10 @@ public class RuntimeAssignor {
   public void rebuildAssignment(final Collection<PersistentQueryMetadata> queries) {
     for (PersistentQueryMetadata queryMetadata: queries) {
       if (queryMetadata instanceof BinPackedPersistentQueryMetadataImpl) {
-        runtimesToSources.put(queryMetadata.getQueryApplicationId(),
-            queryMetadata.getSourceNames());
+        runtimesToSources.put(
+            queryMetadata.getQueryApplicationId(),
+            new HashSet<>(queryMetadata.getSourceNames())
+        );
         idToRuntime.put(queryMetadata.getQueryId(), queryMetadata.getQueryApplicationId());
       }
     }
--- a/ksqldb-engine/src/main/java/io/confluent/ksql/engine/RuntimeAssignor.java
+++ b/ksqldb-engine/src/main/java/io/confluent/ksql/engine/RuntimeAssignor.java
@@ -23,20 +23,26 @@ import io.confluent.ksql.query.QueryId;
 import io.confluent.ksql.util.BinPackedPersistentQueryMetadataImpl;
 import io.confluent.ksql.util.KsqlConfig;
 import io.confluent.ksql.util.PersistentQueryMetadata;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 import java.util.Collection;
 import java.util.HashMap;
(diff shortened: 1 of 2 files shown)


wcarlson5 (commented):
Looks good. Im not sure we need ConcurrentHashMaps but its can't hurt. 

Can you add a test for this in the RuntimeAssginorTest?

ableegoldman:
Blocked on [this bug]( https://github.com/confluentinc/ksql/pull/8523/files#r771100239)  and https://github.com/confluentinc/ksql/pull/8523 in general

wcarlson5:
I merged #8523 with the fix to that bug

wcarlson5 (approved):
Assuming the build is green (or locally green given the jenkins issue) and you test it locally to see there error message form the cli I think you are good to go

ksqldb-engine/src/main/java/io/confluent/ksql/engine/RuntimeAssignor.java
```diff
@@ -57,11 +63,28 @@ public RuntimeAssignor createSandbox() {
     return new RuntimeAssignor(this);
   }
 
-  public String getRuntimeAndMaybeAddRuntime(final QueryId queryId,
-                                             final Collection<SourceName> sources,
-                                             final KsqlConfig config) {
-    if (idToRuntime.containsKey(queryId)) {
-      return idToRuntime.get(queryId);
+  synchronized public String getRuntimeAndMaybeAddRuntime(final QueryId queryId,
```
guozhangwang, line 66: Why only synchronize this function but not others like `dropQuery` and `rebuildAssignment` ? It seems both `runtimesToSources` and `idToRuntime` are updated there too.

guozhangwang (commented):
LGTM overall, just a question regarding the concurrency mechanisms here.

ksqldb-engine/src/main/java/io/confluent/ksql/engine/RuntimeAssignor.java
```diff
@@ -57,11 +63,28 @@ public RuntimeAssignor createSandbox() {
...
+      if (!overlappingSources.isEmpty()) {
+        log.error("Attempt to replace query {} in runtime {} with new input topics {} failed since"
+                      + " the runtime is already consuming from the following input topics: {}\n"
+                      + "Overlapping sources: {}",
+                  queryId, runtimeIdForExistingQuery, sources, existingSources, overlappingSources);
+        throw new IllegalStateException(String.format(
+            "Unable to update query %s to read from stream(s) %s as there is another query in this"
+                + " runtime already consuming from these topic(s). This restriction will be lifted"
+                + " in the near future, but you can work around it for now by dropping the query"
+                + "  and restarting it instead of using CREATE OR REPLACE",
```
agavra, line 84: FWIW, this isn't a workaround - I'm not sure we should suggest it. You can't do this if any downstream query is using the source (which is the only reason to really use COR in first place)

agavra (commented):
>  CREATE OR REPLACE commands that update the source topics of a query

I'm not sure I understand the PR, this isn't possible? You can't change the source topic of a query (e.g. `CSAS WITH (kafka_topic='foo')` and then `COR WITH (kafka_topic='bar')` will fail. At the moment, the `WITH` properties need to be identical)

ableegoldman:
>  this isn't possible? 

Thanks for that @agavra, that's even better. I'll close this out then

PR #8542, opened 2021-12-17
fix: Allow disabling Scalable Push Query ALOS to be set via config

AlanConfluent:
### Description 
Allows scalable push queries to be set via sercer config, not just via the request.

### Testing done 
Tested the change manually and ran tests.

### Reviewer checklist


guozhangwang:
Just to clarify, the config change still needs a bounce to be made effective right? Whereas a request would not.

guozhangwang (approved):

AlanConfluent:
> Just to clarify, the config change still needs a bounce to be made effective right? Whereas a request would not.

@guozhangwang Yes, that's right.  They can both be used with this change, so they can still enable/disabled it with the request if they want, but by default, it will use the server config which requires a roll to change.

AlanConfluent:
Not sure why that top build link didn't get updated, it still shows the failed first build when there's a number 2 that passed.  I'm going to merge.

Issue #8517, opened 2021-12-15
Pull queries with LIMIT clause might return more rows than specified by the limit


Gerrrr:
**Describe the bug**

`io.confluent.ksql.rest.integration.PullQueryLimitHARoutingTest.shouldReturnLimitRowsMultiHostSetupTable` test sometimes fails on the following assertion:

https://github.com/confluentinc/ksql/blob/9c8e8a559da18e17509ab3cedd5ca810543f34ae/ksqldb-rest-app/src/test/java/io/confluent/ksql/rest/integration/PullQueryLimitHARoutingTest.java#L258-L259

This happens because `rows_1` contains 1 more records (301 instead 300) than it should. I believe that this might be a genuine bug if `LimitOperator#next` is accessed from multiple threads as this method is not thread-safe.

**To Reproduce**

Repeatedly run `PullQueryLimitHARoutingTest.shouldReturnLimitRowsMultiHostSetupTable`.

**Expected behavior**

Limit queries should always return the number of rows that is less than or equal to the set limit.

**Actual behaviour**

The test shows that Limit queries can return more rows than expected.

Gerrrr:
#8544 indeed fixes the original issue, but at a significantly higher cost than it should. I am not sure how much impact it has on existing benchmarks, but since the fix is rather straightforward, in my opinion, we should make a follow-up straight-away.

There are multiple issues with the current approach:
* First of all, `PullQueryQueue #acceptRow` is `synchronized` for all kinds of pull queries, not just the ones that have `LIMIT` clause.
* The entire method is a critical section which is very much suboptimal given, as @patrickstuedi pointed out, we run a callback there. The only practical usage of that callback is to publish messages, so we'd rather not do that under a lock.
* AFAIU a lock is not even needed in that case - we just need to update `totalRowsQueued` with CAS. Without contention there is no  additional overhead required to ensure thread-safety.

PR #8544, opened 2021-12-17
fix: make adding to pull query queue threadsafe

cprasad1:
### Description 
fixes https://github.com/confluentinc/ksql/issues/8517 by making adding rows to pull query queue threadsafe

### Testing done 
I ran `io.confluent.ksql.rest.integration.PullQueryLimitHARoutingTest.shouldReturnLimitRowsMultiHostSetupTable` locally ~90 times and it never failed. Before this fix, it used to fail before ~25 runs

### Reviewer checklist

### DIFF
--- a/ksqldb-engine/src/main/java/io/confluent/ksql/query/PullQueryQueue.java
+++ b/ksqldb-engine/src/main/java/io/confluent/ksql/query/PullQueryQueue.java
@@ -217,7 +217,7 @@ public class PullQueryQueue implements BlockingRowQueue {
    * Enqueues a row on the queue.  Blocks until the row can be accepted.
    * @param row The row to enqueue.
    */
-  public boolean acceptRow(final PullQueryRow row) {
+  public synchronized boolean acceptRow(final PullQueryRow row) {
     try {
       if (row == null) {
         return false;


vvcephei (approved):
Thanks, @cprasad1 !

AlanConfluent:
If we're going to stick synchronized on that method, we might want to consider if we need other datastructures to be be synchronized themselves, since it can be a little misleading if we're not relying on that fact.

cprasad1:
@AlanConfluent the 2 datastructures that I was concerned about are the `BlockingQueue` and the `AtomicLong totalRowsQueued`. Both of them seem to be thread-safe

vvcephei:
Chiming in: both of those have readers from different threads, so they'd need a memory barrier of some kind anyway. However, they don't need mutual exclusion from the writer, so we don't need to synchronize them on the same mutex.

Gerrrr:
As an idea, we probably don't have to make the entire method `synchronized` At least, we might want to have a fast-path for the case without `limit`.

AFAIU the original issue was caused by 2 threads passing the first `totalRowsQueued.get() >= limit.getAsInt()` check, and then both adding rows to the queue. If this is correct, then it should be sufficient to condition on CAS `totalRowsQueued`. Something along the lines of:

```java
  /**
   * Enqueues a row on the queue.  Blocks until the row can be accepted.
   * @param row The row to enqueue.
   */
  public boolean acceptRow(final PullQueryRow row) {
    try {
      if (row == null) {
        return false;
      }

      // fast-path for pull queries without limit
      if (!limit.isPresent()) {
        if (doAcceptRow(row)) {
          totalRowsQueued.incrementAndGet();
          return true;
        }
        return false;
      }

      while (true) {
        final long rowsQueued = totalRowsQueued.get();
        if (rowsQueued >= limit.getAsInt()) {
          closeInternal(true);
          return false;
        } else if (totalRowsQueued.compareAndSet(rowsQueued, rowsQueued + 1)) {
          final boolean accepted = doAcceptRow(row);
          if (rowsQueued + 1 >= limit.getAsInt()) {
            closeInternal(true);
          }
          return accepted;
        }
      }
    } catch (final InterruptedException e) {
      // Forced shutdown?
      LOG.error("Interrupted while trying to offer row to queue", e);
    }
    return false;
  }

  private boolean doAcceptRow(final PullQueryRow row) throws InterruptedException {
    while (!closed.get()) {
      if (rowQueue.offer(row, offerTimeoutMs, TimeUnit.MILLISECONDS)) {
        queuedCallback.run();
        return true;
      }
    }
    return false;
  }
```

patrickstuedi (commented):
I'd be a bit concerned about us running the callback under the lock. Also, can we get stuck in the loop there?

PR #8549, opened 2021-12-21
feat: add max task usage and num stateful tasks metrics

lct45:
### Description 
We're adding a metric (`max_task_storage_used_bytes`) that takes the maximum task disk usage on a node and emits that. This allows users to see the maximum disk space a single task is taking up without having to filter through all the task metrics on their side.

We're also adding a metric `num_stateful_tasks` that records the number of stateful tasks on a node. These two metrics combined will allow users to better understand their task work distribution and understand how to tune their system for their needs. 

### Testing done 
unit tests, tested on jconsole

### Reviewer checklist

### DIFF
--- a/ksqldb-engine/src/main/java/io/confluent/ksql/internal/StorageUtilizationMetricsReporter.java
+++ b/ksqldb-engine/src/main/java/io/confluent/ksql/internal/StorageUtilizationMetricsReporter.java
@@ -30,6 +30,7 @@ import java.util.Map;
 import java.util.Optional;
 import java.util.Set;
 import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.atomic.AtomicInteger;
 import java.util.function.Supplier;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
@@ -52,6 +53,7 @@ public class StorageUtilizationMetricsReporter implements MetricsReporter {
   private Map<String, Map<String, TaskStorageMetric>> metricsSeen;
   private static Metrics metricRegistry;
   private static Map<String, String> customTags = new HashMap<>();
+  private static AtomicInteger numberStatefulTasks;
 
   public StorageUtilizationMetricsReporter() {
   }
@@ -66,6 +68,7 @@ public class StorageUtilizationMetricsReporter implements MetricsReporter {
         map.get(KsqlConfig.KSQL_INTERNAL_METRICS_CONFIG)
     );
     this.metricsSeen = new HashMap<>();
+    this.numberStatefulTasks = new AtomicInteger(0);
   }
 
   public static void configureShared(
@@ -85,7 +88,8 @@ public class StorageUtilizationMetricsReporter implements MetricsReporter {
         metricRegistry.metricName("storage_utilization", METRIC_GROUP, customTags);
     final MetricName maxTaskPerNode = 
         metricRegistry.metricName("max_task_storage_used_bytes", METRIC_GROUP, customTags);
-
+    final MetricName numStatefulTasks =
+        metricRegistry.metricName("num_stateful_tasks", METRIC_GROUP, customTags);
     metricRegistry.addMetric(
         nodeAvailable,
         (Gauge<Long>) (config, now) -> baseDir.getFreeSpace()
@@ -107,6 +111,10 @@ public class StorageUtilizationMetricsReporter implements MetricsReporter {
     metricRegistry.addMetric(
         maxTaskPerNode,
         (Gauge<BigInteger>) (config, now) -> (getMaxTaskUsage())
+    );
+    metricRegistry.addMetric(
+        numStatefulTasks,
+        (Gauge<AtomicInteger>) (config, now) -> (numberStatefulTasks)
     );
   }
 
@@ -183,6 +191,7 @@ public class StorageUtilizationMetricsReporter implements MetricsReporter {
     final TaskStorageMetric newMetric;
     // We haven't seen a metric for this query's task before
     if (!metricsSeen.get(queryId).containsKey(taskId)) {
+      numberStatefulTasks.getAndIncrement();
       // create a new task level metric to track state store storage usage
       newMetric = new TaskStorageMetric(
         metricRegistry.metricName(
@@ -212,6 +221,7 @@ public class StorageUtilizationMetricsReporter implements MetricsReporter {
   ) {
     // remove storage metric for this task
     taskMetric.remove(metric);
+    numberStatefulTasks.getAndDecrement();
     if (taskMetric.metrics.size() == 0) {
       // no more storage metrics for this task, can remove task gauge
       metricRegistry.removeMetric(taskMetric.metricName);
--- a/ksqldb-engine/src/main/java/io/confluent/ksql/internal/StorageUtilizationMetricsReporter.java
+++ b/ksqldb-engine/src/main/java/io/confluent/ksql/internal/StorageUtilizationMetricsReporter.java
@@ -27,6 +27,7 @@ import java.util.Collection;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.Optional;
 import java.util.Set;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.function.Supplier;
@@ -49,7 +50,7 @@ public class StorageUtilizationMetricsReporter implements MetricsReporter {
   private static final String METRIC_GROUP = "ksqldb_utilization";
 
   private Map<String, Map<String, TaskStorageMetric>> metricsSeen;
-  private Metrics metricRegistry;
+  private static Metrics metricRegistry;
   private static Map<String, String> customTags = new HashMap<>();
 
   public StorageUtilizationMetricsReporter() {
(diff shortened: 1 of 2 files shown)


ksqldb-engine/src/main/java/io/confluent/ksql/internal/StorageUtilizationMetricsReporter.java
```diff
@@ -101,6 +108,14 @@ public static void configureShared(
         (((double) baseDir.getTotalSpace() - (double) baseDir.getFreeSpace()) 
             / (double) baseDir.getTotalSpace())
     );
+    metricRegistry.addMetric(
```
lct45, line 111: actually getting NPEs from these two, it seems like `configureShared` is called before `configure` to the `metricRegistry` and `numberStatefulTasks` aren't initialized before they're called in the gauges

lct45: updated the gauges to handle the if the items are null, moving them to `configure` means they won't get set at start-up and you'll have to check to see if it's already been registered so we don't get errors about double registering metrics
`ksql> CREATE TABLE currentLocation AS
>  SELECT profileId,
>         LATEST_BY_OFFSET(latitude) AS la,
>         LATEST_BY_OFFSET(longitude) AS lo
>  FROM riderlocations
>  GROUP BY profileId
>  EMIT CHANGES;
Could not write the statement 'CREATE TABLE currentLocation AS
  SELECT profileId,
         LATEST_BY_OFFSET(latitude) AS la,
         LATEST_BY_OFFSET(longitude) AS lo
  FROM riderlocations
  GROUP BY profileId
  EMIT CHANGES;' into the command topic.
Caused by: A metric named 'MetricName [name=max_task_storage_used_bytes,
	group=ksqldb_utilization, description=, tags={}]' already exists, can't register
	another one.`

ksqldb-engine/src/main/java/io/confluent/ksql/internal/StorageUtilizationMetricsReporter.java
```diff
@@ -49,8 +51,9 @@
   private static final String METRIC_GROUP = "ksqldb_utilization";
 
   private Map<String, Map<String, TaskStorageMetric>> metricsSeen;
-  private Metrics metricRegistry;
+  private static Metrics metricRegistry;
```
rodesai, line 54: This should not be static - just leave it local to each member. We intentionally made it not-static to avoid re-using the instance across test runs.

ksqldb-engine/src/main/java/io/confluent/ksql/internal/StorageUtilizationMetricsReporter.java
```diff
@@ -65,6 +68,7 @@ public synchronized void configure(final Map<String, ?> map) {
         map.get(KsqlConfig.KSQL_INTERNAL_METRICS_CONFIG)
     );
     this.metricsSeen = new HashMap<>();
+    this.numberStatefulTasks = new AtomicInteger(0);
```
rodesai, line 71: This is a static, so we shouldn't re-initialize for each instance. We should just initialize this statically (e.g. `private static AtomicInteger numberStatefulTasks = new AtomicInteger(0);`) or from `configureShared`

ksqldb-engine/src/main/java/io/confluent/ksql/internal/StorageUtilizationMetricsReporter.java
```diff
@@ -227,6 +244,24 @@ private BigInteger computeQueryMetric(final String queryId) {
     }
     return queryMetricSum;
   }
+  
+  public static synchronized BigInteger getMaxTaskUsage() {
+    if (metricRegistry == null) {
+      return BigInteger.ZERO;
+    }
+    final Collection<KafkaMetric> taskMetrics = metricRegistry
```
rodesai, line 252: This is kind of expensive. Once this is shipped can you just see what happens to the cpu of an idle instance? I don't think it will make much difference because it only runs every 30 seconds. But better to check anyway. Alternatively we could keep a static set of the tasks and a max that's updated on every new or removed task.

ksqldb-engine/src/test/java/io/confluent/ksql/internal/StorageUtilizationMetricsReporterTest.java
```diff
@@ -253,6 +267,58 @@ public void shouldIgnoreNonSSTMetrics() {
     // Then:
     assertThrows(AssertionError.class, () -> verifyAndGetRegisteredMetric(TASK_STORAGE_METRIC, TASK_ONE_TAGS));
   }
+  
+  @Test
```
rodesai, line 271: can you add a case for max task when there are 0 tasks?

ksqldb-engine/src/main/java/io/confluent/ksql/internal/StorageUtilizationMetricsReporter.java
lct45: Is there a way to still reference it from a static context?

ksqldb-engine/src/main/java/io/confluent/ksql/internal/StorageUtilizationMetricsReporter.java
lct45: just changed it to be initialized statically, updated the gauge as well

ksqldb-engine/src/main/java/io/confluent/ksql/internal/StorageUtilizationMetricsReporter.java
lct45: idle instance as in an instance with no queries or instance with queries but no data? I can check after shipping, it's definitely ... unideal, but storing the tasks / max had a longer LOE so went with this first.

Would the static max update as the underlying task metrics do? I wanted to make sure we were getting the max of all the most recent gauge values

ksqldb-engine/src/test/java/io/confluent/ksql/internal/StorageUtilizationMetricsReporterTest.java
lct45: yup

ksqldb-engine/src/main/java/io/confluent/ksql/internal/StorageUtilizationMetricsReporter.java
lct45: I didn't change this because I'm not sure how to handle it, `configureShared` is called from a static context so it has to be static, meaning `getMaxTaskUsage` has to be static which calls `metricRegistry.metrics() which intellij is also telling me needs to be static to be called from this context. I can create a helper function to get `metricRegistry.metrics()` but it seems like it would have the same issue

ksqldb-engine/src/main/java/io/confluent/ksql/internal/StorageUtilizationMetricsReporter.java
rodesai: > idle instance as in an instance with no queries or instance with queries but no data? I can check after shipping, it's definitely ... unideal, but storing the tasks / max had a longer LOE so went with this first.

Yep. An idle cluster will still have loads of metrics, so seeing if the cpu usage for an idle cluster is higher than normal is a good test.

rodesai: > Would the static max update as the underlying task metrics do? I wanted to make sure we were getting the max of all the most recent gauge values

Good point - the space used is not known at the time the metric is added. If the current impl turns out too expensive we can just maintain a static mapping of task to sst file used metrics and then take the max every time this is called.

ksqldb-engine/src/main/java/io/confluent/ksql/internal/StorageUtilizationMetricsReporter.java
rodesai: `configureShared` is passed a `MetricsRegistry` instance that you can register the metrics in.

ksqldb-engine/src/main/java/io/confluent/ksql/internal/StorageUtilizationMetricsReporter.java
guozhangwang: Not sure I understand the above issue: what object the NPE is referring to? The metric names like `numberStatefulTasks` should be inside the registry, and here we are just adding a new metric under that name, and `numberStatefulTasks` are already static so it should have been initialized at the time this line is executed.

lct45: The NPE was when the gauge was being sampled, so when `getMaxTaskUsage` was called it tried to access the metrics registry before it had been initialized since it gets initialized in `configure`. This actually shouldn't happen now that I pass the metrics registry into `getMaxTaskUsage`. The other NPE was for `numberStatefulTasks` which was initialized in `configure` so when the gauge called `numberStatefulTasks.get()` it returned a NPE. This is now initialized to 0 so that NPE is also resolved

ksqldb-engine/src/main/java/io/confluent/ksql/internal/StorageUtilizationMetricsReporter.java
```diff
@@ -227,6 +243,21 @@ private BigInteger computeQueryMetric(final String queryId) {
     }
     return queryMetricSum;
   }
+  
+  public static synchronized BigInteger getMaxTaskUsage(final Metrics metricRegistry) {
+    final Collection<KafkaMetric> taskMetrics = metricRegistry
+        .metrics()
+        .entrySet()
+        .stream()
+        .filter(e -> e.getKey().name().contains("task_storage_used_bytes"))
```
rodesai, line 252: put the metric name in a constant

rodesai (approved):
LGTM with one nit inline

PR #8550, opened 2021-12-22
docs: klip-59: JSON functions

Gerrrr:
### Description 

KLIP for adding new JSON functions.

### Reviewer checklist

### DIFF
--- a/design-proposals/klip-59-json-functions.md
+++ b/design-proposals/klip-59-json-functions.md
@@ -1,0 +1,264 @@
+# KLIP-59 - JSON functions
+
+**Authors**: Aleksandr Sorokoumov (@gerrrr), Colin Hicks (@colinhicks)
+**Release Target**: TBD
+**Status**: _Approved_ |
+**Discussion**: [GitHub PR](https://github.com/confluentinc/ksql/pull/8550)
+
+## Motivation and background
+
+Many ksqlDB users leverage JSON. ETL pipelines often deal in large JSON blobs from, e.g., MongoDB.
+As the de-facto format of the web, it is often the format of analytics events produced to Kafka, as
+well as a choice for materialized views consumed by web apps.
+
+We want to position ksqlDB as a JSON-friendly database. Support for reading, writing, and
+traversing data as JSON presents an important value-add. And users of traditional databases like
+PostgreSQL already enjoy powerful support for JSON.
+
+Today, ksqlDB can:
+* Read JSON-formatted keys and values in input records.
+* Serialize keys and values as JSON in output records.
+* Extract a string value at a given JSON path.
+
+
+## What is in scope
+
+* Validate if a given string contains JSON value
+* Calculate the length of a JSON array encoded in a string
+* Extract keys and records from a given JSON-formatted string
+* Concatenate JSON-formatted strings
+* Convert a ksqlDB record to JSON-formatted string
+
+## What is not in scope
+
+* A JSON datatype
+* Binary JSON support
+
+## Public APIs
+
+Users can use ksqlDB’s varchar type on columns that they know are JSON formatted. A JSON-formatted
+string can also be nested within a map or struct.
+
+### is_json_string
+
+```
+is_json_string(json_string) -> Boolean
+```
+
+Given a string, returns true if it can be parsed as a valid JSON value, false otherwise.
+
+#### Examples
+
+```
+is_json_string("[1, 2, 3]") // returns true
+is_json_string("{}") // returns true
+is_json_string("1") // returns true
+is_json_string("\"abc\"") // returns true
+is_json_string("null") // returns true
+is_json_string("") // returns false
+is_json_string("abc") // returns false
+is_json_string(NULL) // returns false
+```
+
+
+### json_array_length
+
+```
+json_array_length(json_string) -> Integer
+```
+
+Given a string, parses it as a JSON value and returns the length of the top-level array. Returns
+`NULL` if the string can't be interpreted as a JSON array, i.e., it is `NULL` or it does not contain
+valid JSON, or the JSON value is not an array.
+
+#### Examples
+
+```
+json_array_length("[1, 2, 3]") // returns 3
+json_array_length("[1, [1, [2]], 3]") // returns 3
+json_array_length("[]") // returns 0
+json_array_length("{}") // returns NULL
+json_array_length("123") // returns NULL
+json_array_length("abc") // returns NULL
+json_array_length(NULL) // returns NULL
+```
+
+### json_keys
+
+```
+json_keys(json_string) -> Array<String>
+```
+
+Given a string, parses it as a JSON object and returns a ksqlDB array of strings representing the
+top-level keys. Returns `NULL` if the string can't be interpreted as a JSON object, i.e., it is
+`NULL` or it does not contain valid JSON, or the JSON value is not an object.
+
+#### Examples
+
+```
+json_keys("{\"a\": \"abc\", \"b\": { \"c\": \"a\" }, \"d\": 1}") // returns ["a", "b", "d"]
+json_keys("{}") // returns []
+json_keys("[]") // returns NULL
+json_keys("") // returns NULL
+json_keys("123") // returns NULL
+json_keys("abc") // returns NULL
+json_keys(NULL) // returns NULL
+```
+
+### json_records
+
+```
+json_records(json_string) -> Array<Struct<json_key:String, json_value:String>>
+```
+
+Given a string, parses it as a JSON object and returns a ksqlDB array of structs containing 2 
+strings - `json_key` and `json_value` representing the top-level keys and values. Returns `NULL` if 
+the string can't be interpreted as a JSON object, i.e. it is `NULL` or it does not contain valid 
+JSON, or the JSON value is not an object.
+
+#### Examples
+
+```
+json_records("{\"a\": \"abc\", \"b\": { \"c\": \"a\" }, \"d\": 1}") // returns [Struct{json_key="a", json_value="\"abc\""}, Struct{json_key="b", json_value="{ \"c\": \"a\" }}, Struct{json_key="d", json_value="1"}]
+json_records("{}") // returns []
+json_records("[]") // returns NULL
+json_records("") // returns NULL
+json_records("123") // returns NULL
+json_records("abc") // returns NULL
+json_records(NULL) // returns NULL
+```
+
+#### Rejected alternatives
+
+**Remove redundant quotes around values**. For example:
+
+```
+json_records("{\"a\": \"abc\"}) // returns [["a", "abc"]]
+```
+
+This makes the output less awkward, but removes the type information in certain cases. In the following
+case it is not possible to derive whether `1` is a string or a number:
+
+```
+json_records("{\"a\": \"1\"}) // returns [Struct{json_key="a", json_value="1"}]
+```
+
+**Return array of arrays rather than an array of structs**. For example:
+```
+json_records("{\"a\": \"abc\"}) // returns [["a", "abc"]]
+```
+
+A struct is a better return type in this case as it enables meaningful field names and reduces
+the number of potential sanity checks on the end-user side (e.g., array length).
+
+### json_concat
+
+```
+json_concat(json_string, json_string) -> String
+```
+
+Given 2 strings, parse them as JSON values and return a string representing their concatenation.
+Concatenation rules are identical to PostgreSQL's [|| operator](https://www.postgresql.org/docs/14/functions-json.html):
+* If both strings deserialize into JSON objects, then return an object with a union of the input key,
+  taking values from the second object in the case of duplicates.
+* If both strings deserialize into JSON arrays, then return the result of array concatenation.
+* If at least one of the deserialized values is not an object, then convert non-array inputs to a
+  single-element array and return the result of array concatenation.
+* If at least one of the input strings is `NULL` or can't be deserialized as JSON, then return `NULL`.
+
+Akin to PostgreSQL's `||` operator, this function merges only top-level object keys or arrays.
+
+#### Examples
+
+```
+json_concat("{\"a\": 1}", "{\"b\": 2}") // returns "{\"a\": 1, \"b\": 2}"
+json_concat("{\"a\": {\"5\": 6}}", "{\"a\": {\"3\": 4}}") // returns "{\"a\": {\"3\": 4}}"
+json_concat("{}", "{}") // returns "{}"
+json_concat("[1, 2]", "[3, 4]") // returns "[1,2,3,4]"
+json_concat("[1, [2]]", "[[[3]], [[[4]]]]") // returns "[ 1, [2], [[3]], [[[4]]] ]"
+json_concat("null", "null") // returns [null, null]
+json_concat("[1, 2]", "{\"a\": 1}") // returns "[1, 2, {\"a\": 1}]"
+json_concat("[1, 2]", "3") // returns "[1, 2, 3]"
+json_concat("1", "2") // returns "[1, 2]"
+json_concat("[]", "[]") // returns []
+json_concat("abc", "[1]") // returns NULL
+json_concat(NULL, "[1]") // returns NULL
+```
+
+#### Rejected alternatives
+
+If one argument represents a valid JSON and the other one does not, then return the valid string
+as a result of concatenation.
+
+Given that concatenation of invalid JSON strings is an exceptional situation, we should return
+`NULL` if any input is not a JSON. This way, we avoid ambiguity if a user needs to trace back how
+they got certain results.
+
+If this function returns the valid record in the case when the other one is invalid, from the user
+perspective, there are 3 possibilities how they ended up with this result:
+* The second object could have been empty.
+* The second object could have had the same keys/values.
+* The second object is invalid JSON.
+
+By always returning `NULL` given an invalid JSON string we remove the third, the least obvious,
+possibility which reduces cognitive load from the user.
+
+### to_json_string
+
+```
+to_json_string(val) -> String
+```
+
+Given any ksqlDB type returns the equivalent JSON string.
+
+#### Examples
+
+**Primitives types**
+
+```
+to_json_string(1) // returns "1"
+to_json_string(15.3) // returns "15.3"
+to_json_string("abc") // returns "\"abc\""
+to_json_string(true) // returns "true"
+to_json_string(2021-10-11) // DATE type, returns "\"2021-10-11\""
+to_json_string(13:25) // TIME type, returns "\"13:25\""
+to_json_string(2021-06-31T12:18:39.446) // TIMESTAMP type, returns "\"2021-06-31T12:18:39.446\""
+to_json_string(NULL) // returns "null"
+```
+
+**Compound types**
+```
+to_json_string(Array[1, 2, 3]) // returns "[1, 2, 3]"
+to_json_string(Struct{id=1,name=A}) // returns "{\"id\": 1, \"name\": \"a\"}"
+to_json_string(Map('c' := 2, 'd' := 4)) // returns "{\"c\": 2, \"d\": \"4\"}"
+to_json_string(Array[Struct{json_key=1 json_value=Map('c' := 2, 'd' := true)}]) // returns "[{\"json_key\": 1, \"json_value\": {\"c\": 2, \"d\": true}}]"
+```
+
+Custom types are processed in the same way as the types they alias.
+
+#### Rejected alternatives
+
+Return `NULL` given `NULL` as input. Although, it might make sense from the `NULL` propagation point
+of view, there are several arguments why we should map SQL `NULL` to JSON `null`:
+* Akin to SQL's `NULL`, JSON's `null` is a marker of absent value, so semantics does not change.
+* Returning SQL `NULL` would make it impossible to serialize any input to JSON `null`.
+
+## Design
+
+Akin to the existing `EXTRACTJSONFIELD` and `JSON_ARRAY_CONTAINS`, the general workflow for all
+functions is to parse the given string(s) into jackson's `JsonNode`s and perform manipulations according
+to the functions' specifications.
+
+This approach might be suboptimal from the performance point of view since every function invocation
+completely deserializes the entire structure and then serializes the result back. However, the best
+option for performance-sensitive JSON manipulation is the introduction of native JSON types that
+avoids the serde step altogether. In turn, string manipulation is easy to implement, and it enables
+the majority of common use cases.
+
+## Test Plan
+
+Add unit tests, QTT tests for the new functions.
+
+## Documentation Updates
+
+Add functions description to [Scalar functions docs](https://github.com/confluentinc/ksql/blob/master/docs/developer-guide/ksqldb-reference/scalar-functions.md).
--- a/design-proposals/klip-59-json-functions.md
+++ b/design-proposals/klip-59-json-functions.md
@@ -108,18 +108,17 @@ json_keys(NULL) // returns NULL
 ### json_records
 
 ```
-json_records(json_string) -> Array<Struct<json_key:String, json_value:String>>
+json_records(json_string) -> Map<String, String>
 ```
 
-Given a string, parses it as a JSON object and returns a ksqlDB array of structs containing 2 
-strings - `json_key` and `json_value` representing the top-level keys and values. Returns `NULL` if 
-the string can't be interpreted as a JSON object, i.e. it is `NULL` or it does not contain valid 
-JSON, or the JSON value is not an object.
+Given a string, parses it as a JSON object and returns a map representing the top-level keys and 
+values. Returns `NULL` if the string can't be interpreted as a JSON object, i.e. it is `NULL` or 
+it does not contain valid JSON, or the JSON value is not an object.
 
 #### Examples
 
 ```
-json_records("{\"a\": \"abc\", \"b\": { \"c\": \"a\" }, \"d\": 1}") // returns [Struct{json_key="a", json_value="\"abc\""}, Struct{json_key="b", json_value="{ \"c\": \"a\" }}, Struct{json_key="d", json_value="1"}]
+json_records("{\"a\": \"abc\", \"b\": { \"c\": \"a\" }, \"d\": 1}") // {"a": "\"abc\"", "b": "{ \"c\": \"a\" }", "d": "1"}
 json_records("{}") // returns []
 json_records("[]") // returns NULL
 json_records("") // returns NULL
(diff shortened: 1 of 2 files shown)


design-proposals/klip-59-json-functions.md
```diff
@@ -0,0 +1,244 @@
...
+```
+
+### json_records
+
+```
+json_records(json_string)
+```
+
+Given a string, parses it as a JSON object and returns a ksqlDB array of two-element arrays
+representing the top-level keys and values. Returns `NULL` if the string can't be interpreted as a
```
agavra, line 115: json objects don't have to have the same value type, right? what would happen if we passed in `{"a": 1, "b": "X"}` (subtext: ksqlDB arrays are strongly typed)

design-proposals/klip-59-json-functions.md
```diff
@@ -0,0 +1,244 @@
...
+* Validate if a given string contains JSON value
+* Calculate the length of a JSON array encoded in a string
+* Extract keys and records from a given JSON-formatted string
+* Concatenate JSON-formatted strings
+* Convert a ksqlDB record to JSON-formatted string
+
+## What is not in scope
+
+* A JSON datatype
+* Binary JSON support
```
agavra, line 35: I don't immediately see any problems with omitting this from scope, but if we do ever introduce these would we expect the semantics of any of these to change at all (e.g. return a `JSON` type instead of a string or somethign else)?

agavra (approved):
Nice! Thanks @Gerrrr

design-proposals/klip-59-json-functions.md
colinhicks: Based on the example below, I think the result in ksql types would be: `Array[Array['a',  '1'], Array['b', '"X"']]`. Note: double-quotes surrounding the value of `b`. Essentially each inner value (meaning the second item in the nested arrays) in the result is equivalent to calling `to_json_string` on the JSON value input.

I assume the relevant the UDF's return type is `Array<Array<string>>`.

Gerrrr: Agree, the return type should be `Array<Array<string>>`.

design-proposals/klip-59-json-functions.md
Gerrrr: I would rather introduce a separate set of functions or operators (see PostgreSQL syntax) for the `JSON` type instead of changing the return type of the existing functions. The reasons are:
* We can provide functions for built-in `JSON` type with more elegant syntax AND avoid wrapping the output in strings.
* Having a function that behaves differently given different argument types might confuse the users. Consider `json_concat` that kinda does it already. Its behavior description is three times longer than descriptions for any other proposed function, which is a good proxy for cognitive complexity.

colinhicks (approved):

design-proposals/klip-59-json-functions.md
```diff
@@ -0,0 +1,244 @@
...
+Given a string, returns true if it can be parsed as a valid JSON value, false otherwise.
+
+#### Examples
+
+```
+is_json("[1, 2, 3]") // returns true
+is_json("{}") // returns true
+is_json("1") // returns true
+is_json("\"abc\"") // returns true
+is_json("null") // returns true
```
mjsax, line 57: For my own education. What is `null` in JSON? Is the an "empty" JSON (is it considered a primitive or complex type)?

design-proposals/klip-59-json-functions.md
```diff
@@ -0,0 +1,244 @@
...
+is_json("") // returns false
```
mjsax, line 58: For my own education: why does this return `false`?

design-proposals/klip-59-json-functions.md
mjsax: If we would add a `JSON` type, how would we name functions that accept `JSON` compared to the function names that are proposed in this KLIP? Just curious.

design-proposals/klip-59-json-functions.md
```diff
@@ -0,0 +1,244 @@
...
+```
+
+Given a string, parses it as a JSON object and returns a ksqlDB array of strings representing the
+top-level keys. Returns `NULL` if the string can't be interpreted as a JSON object, i.e., it is
+`NULL` or it does not contain valid JSON, or the JSON value is not an object.
+
+#### Examples
+
+```
+json_keys("{\"a\": \"abc\", \"b\": { \"c\": \"a\" }, \"d\": 1}") // returns ["a", "b", "c"]
```
mjsax, line 99: Should this return `["a", "b", "d"]` ? `c` is the nested key?

design-proposals/klip-59-json-functions.md
mjsax: We should update the KLIP to call out the exact type of the result array.

mjsax: Btw: Why do we return an array of 2-element arrays? Would it be better to return an array of structs with two string field instead, ie, `Array<Struct<json_key:String,json_value:String>>`

design-proposals/klip-59-json-functions.md
```diff
@@ -0,0 +1,244 @@
...
+json_records("{\"a\": \"abc\", \"b\": { \"c\": \"a\" }, \"d\": 1}") // returns [["a", "\"abc\""], ["b", "{ \"c\": \"a\" }"], ["d", "1"]]
+json_records("{}") // returns []
+json_records("[]") // returns NULL
+json_records("") // returns NULL
+json_records("123") // returns NULL
+json_records("abc") // returns NULL
+json_records(NULL) // returns NULL
+```
+
+#### Rejected alternatives
```
mjsax, line 130: I might be better to put all rejected alternative as the end of the KLIP instead of spreading it out for each function.

design-proposals/klip-59-json-functions.md
```diff
@@ -0,0 +1,244 @@
...
+* If at least one of the input strings is `NULL` or can't be deserialized as JSON, then return `NULL`.
+
+Akin to PostgreSQL's `||` operator, this function merges only top-level object keys or arrays.
+
+#### Examples
+
+```
+json_concat("{\"a\": 1}", "{\"b\": 2}") // returns "{\"a\": 1, \"b\": 2}"
+json_concat("{\"a\": {\"3\": 4}}", "{\"a\": {\"5\": 6}}") // returns "{\"a\": {\"5\": 6}}"
+json_concat("{\"a\": {\"5\": 6}}", "{\"a\": {\"3\": 4}}") // returns "{\"a\": {\"3\": 4}}"
```
mjsax, line 166: Example seems to be redundant to the previous one in L165?

design-proposals/klip-59-json-functions.md
```diff
@@ -0,0 +1,244 @@
...
+json_concat("{}", "{}") // returns "{}"
+json_concat("[1, 2]", "[3, 4]") // returns "[1,2,3,4]"
```
mjsax, line 168: Seems this case is described via second bullet-point above -- but it's not really explicit. Should we update the bullet points above to be easier to follow?

design-proposals/klip-59-json-functions.md
```diff
@@ -0,0 +1,244 @@
...
+json_concat("[1, [2]]", "[[[3]], [[[4]]]]") // returns "[ 1, [2], [[3]], [[[4]]] ]"
+json_concat("null", "null") // returns [null, null]
+json_concat("[1, 2]", "{\"a\": 1}") // returns "[1, 2, {\"a\": 1}]"
+json_concat("[1, 2]", "3") // returns "[1, 2, 3]"
+json_concat("[3]", "[1, 2]") // returns "[3, 1, 2]"
```
mjsax, line 173: Seems to be redundant to example above in L168?

design-proposals/klip-59-json-functions.md
```diff
@@ -0,0 +1,244 @@
...
+json_concat("abc", "[1]") // returns NULL
+json_concat(NULL, "[1]") // returns NULL
+```
+
+#### Rejected alternatives
+
+If one argument represents a valid JSON and the other one does not, then return the valid string
+as a result of concatenation.
+
+Given that concatenation of invalid JSON strings is an exceptional situation, I believe that we
```
mjsax, line 185: nit: remove `I believe that`

design-proposals/klip-59-json-functions.md
```diff
@@ -0,0 +1,244 @@
...
+
+### to_json_string
+
+```
+to_json_string(val)
+```
+
+Given any ksqlDB type returns the equivalent JSON string.
+
+#### Examples
```
mjsax, line 206: Can we add examples for all types? Missing: boolean, time, date, timestamp.

What about custom types? (Guess it's straightforward, but maybe worth to mention).

Might also be good to have a few example with nested complex types?

design-proposals/klip-59-json-functions.md
```diff
@@ -0,0 +1,244 @@
...
+* Returning SQL `NULL` would make it impossible to serialize any input to JSON `null`.
+
+## Design
+
+Akin to the existing `EXTRACTJSONFIELD` and `JSON_ARRAY_CONTAINS`, the general workflow for all
+functions is to parse the given string(s) into jackson's `JsonNode`s and perform manipulations according
+to the functions' specifications.
+
+This approach might be suboptimal from the performance point of view since every function invocation
+completely deserializes the entire structure and then serializes the result back. However, I would argue
```
mjsax, line 232: nit: remove `I would argue`

design-proposals/klip-59-json-functions.md
suhas-satish: I prefer keeping it in-line with each function closer to the reader's context

design-proposals/klip-59-json-functions.md
suhas-satish: This was non-intuitive to me. I would have expected union of json values as the result. ie, "{\"a\": [{\"3\": 4}, {\"5\": 6}] }". Is this too difficult and error-prone to implement and not how its done in postgres?

design-proposals/klip-59-json-functions.md
Gerrrr: Thinking about your question, I realized that function names we picked in this KLIP would actually be perfect for the `JSON` type. I propose to rename functions here in the following way:

* `is_json` -> `is_json_string`. The former name would better correspond to a function that returns `true` if the argument is of `JSON` type.
* `json_array_length` -> `json_string_array_length`
* `json_keys` -> `json_string_keys`
* `json_records` -> `json_string_records`
* `json_concat` -> `json_string_concat`

Thoughts @mjsax @agavra @colinhicks @suhas-satish ?

design-proposals/klip-59-json-functions.md
Gerrrr: JSON `null` is a primitive type. See [RFC-4627](https://datatracker.ietf.org/doc/html/rfc4627):

> JSON can represent four primitive types (strings, numbers, booleans,
>   and null) and two structured types (objects and arrays).

design-proposals/klip-59-json-functions.md
Gerrrr: An empty string does not contain valid JSON because it can't be decoded into any of the primitive or structured types JSON can represent.

design-proposals/klip-59-json-functions.md
colinhicks: What is the motivation for having different function names depending on the argument type? From an implementation perspective, if we were to implement a JSON type, wouldn't this just mean adding an overload that accepts an argument of that type?

design-proposals/klip-59-json-functions.md
Gerrrr: @mjsax removed in https://github.com/confluentinc/ksql/pull/8550/commits/41e8e75e867b3e9cc884c8577df7daeed9154dc1.

@suhas-satish In my experience, this behavior for map union is widespread. Some examples are:
* [Python union operator](https://www.python.org/dev/peps/pep-0584/)
* [Java Map's putAll](https://docs.oracle.com/javase/8/docs/api/java/util/Map.html#putAll-java.util.Map-)
* [Javascript's union](https://stackoverflow.com/questions/32000865/simplest-way-to-merge-es6-maps-sets)

design-proposals/klip-59-json-functions.md
Gerrrr: Removed in https://github.com/confluentinc/ksql/pull/8550/commits/9672f21f932c16bf173529889608035ce467270c.

design-proposals/klip-59-json-functions.md
Gerrrr: Removed in https://github.com/confluentinc/ksql/pull/8550/commits/9672f21f932c16bf173529889608035ce467270c.

design-proposals/klip-59-json-functions.md
Gerrrr: Good catch! Fixed in https://github.com/confluentinc/ksql/pull/8550/commits/5185e67e123fd03445abba27e8ade743b57eb07c.

design-proposals/klip-59-json-functions.md
Gerrrr: Good ideas @mjsax! I specified the return type of all functions in https://github.com/confluentinc/ksql/pull/8550/commits/145d932752706b0bf995d1ff7899d0d244dc14c1.

design-proposals/klip-59-json-functions.md
Gerrrr: Agree with @suhas-satish. In my view it is easier to reason about individual functions this way.

design-proposals/klip-59-json-functions.md
Gerrrr: Added a separate bullet point for the case when both strings deserialize into lists in https://github.com/confluentinc/ksql/pull/8550/commits/f519ea910cbc971e02230b89a528a79f9156da33.

design-proposals/klip-59-json-functions.md
Gerrrr: Removed in https://github.com/confluentinc/ksql/pull/8550/commits/6b53d3e5cd42f8e7c2e41c1096c68477137001ec.

design-proposals/klip-59-json-functions.md
colinhicks: I don't think we should optimize for an unplanned feature. But if we were to add a JSON type in the future, having for example a `json_keys` function that accepts both a JSON-typed value or a string value is what I would expect as a user.

Taking this a step further, if we were to also introduce a binary JSON type, we would not add `bson_keys` and so on, for a similar reason that we don't have `floor_float`, `floor_decimal`, `floor_big_decimal`, etc. (by the way we also implement floor for integers to preserve the semantics even though this is a no-op). I get that it's slightly different in the JSON string-argument case, where the coercion of the argument value implies a more expensive operation that we assume we might better optimize if and when we deliver JSON types. But I don't think that means we need a distinct name for the string-argument case.

agavra: I have a slight preference toward @colinhicks' suggestion, but wouldn't rock the boat if we decided otherwise. I think overloading is pretty common in ksql. Is there any precedence in existing systems (postgres, sqlserver, etc...)?

Gerrrr: Thank you for elaborating your point @colinhicks!

While writing down my specific concerns with having the same function names I realized that they only apply to  `is_json` and `json_concat`. I agree that the rest of the functions can be overloaded later with native json types. 

First, I still think that `is_json` should be renamed to `is_json_string` so that a function for sure has only 1 job - to check if a given string can be decoded as JSON.

As for overloading `json_concat`, my primary concern is that Its return type would depend on its arguments' type. As this is not exactly 1-1 conversion like with float and decimal, I'd rather have 2 functions for JSON concatenation in the future - one that accepts strings and returns a string and another one that accepts native JSON and return its result of the same type. 

A strawman argument to avoid overloading `json_concat` is mixing types, e.g. `json_concat(json, string)`. It would be challenging to come up with the most intuitive return type for that. It would also require extensive documentation to enumerate all cases when such function can return `NULL`.

colinhicks: > First, I still think that is_json should be renamed to is_json_string so that a function for sure has only 1 job - to check if a given string can be decoded as JSON.

Sounds good!

> I'd rather have 2 functions for JSON concatenation in the future - one that accepts strings and returns a string and another one that accepts native JSON and return its result of the same type.

Great point. I agree, @Gerrrr

design-proposals/klip-59-json-functions.md
mjsax: Would be good to hear from @colinhicks what he thinks about the return type? Pros/Cons for both ideas? Would also be worth to add to rejected alternatives.

design-proposals/klip-59-json-functions.md
mjsax: Great discussion. -- Not sure if I understand the argument about `json_concat`? If we have `json_concat(string, string) -> string` and `json_concat(json, json) -> json` it seems to be well defined? And we don't need to support `json_concat(string, json)` or `json_concat(json, string)`.

Or maybe I missed the point of the argument?

design-proposals/klip-59-json-functions.md
mjsax: So it's just not in the json spec? (I guess the spec _could_ define what it mean if they wanted to, eg, `""` == `"null"`).

design-proposals/klip-59-json-functions.md
mjsax: @suhas-satish I agree that it not totally intuitive, but we don't need to have two example for the same case. L165 and L166 are just the same.

design-proposals/klip-59-json-functions.md
Gerrrr: Yeah, it is not in the spec.

design-proposals/klip-59-json-functions.md
Gerrrr: Added in https://github.com/confluentinc/ksql/pull/8550/commits/4cd79269dcb217e5b43aa47ee420a85b720f72fb.

design-proposals/klip-59-json-functions.md
Gerrrr: > If we have json_concat(string, string) -> string and json_concat(json, json) -> json it seems to be well defined?

I think that it might be confusing for users if a function returns a different type based on its arguments type. 

Besides, this would be inconsistent with our current practice - e.g., we have [ARRAY_UNION](https://docs.ksqldb.io/en/latest/developer-guide/ksqldb-reference/scalar-functions/#array_union) and [MAP_UNION](https://docs.ksqldb.io/en/latest/developer-guide/ksqldb-reference/scalar-functions/#map_union) rather than a single function `UNION` that is overloaded for different containers.

design-proposals/klip-59-json-functions.md
Gerrrr: Meanwhile I added a line on why an array of structs is a better alternative than an arrays of arrays - https://github.com/confluentinc/ksql/pull/8550/commits/391c9e04421a14c2dadfb0d4811ad11988b87773.

design-proposals/klip-59-json-functions.md
mjsax: > I think that it might be confusing for users if a function returns a different type based on its arguments type.

Not really (at least not for me). And we do this already in ksqlDB. For example `FLOOR()` (https://github.com/confluentinc/ksql/blob/master/ksqldb-engine/src/main/java/io/confluent/ksql/function/udf/math/Floor.java) is defined for int,long,double,decimal and it's return type matches it's input type (many other function work similar).

I am not sure why we have `ARRAY_UNION` and `MAP_UNION`. It does not make too much sense to me to have both, instead of `UNION()` (even if `UNION()` is a little bit tricky as there is also the SQL `UNION` keyword -- even if we don't support `UNION`).

blueedgenick: as the original author of MAP_UNION and ARRAY_UNION i will chime in to say that is exactly (one of) the reasons why: i.e. we don't want to introduce a function whose name is the same as a SQL keyword. Secondary reason is that it follows the naming convention established in other database-like systems with comparable functionality (Presto in this case, IIRC)

mjsax: So are you saying you would prefer an overloaded `json_concat` or to add `json_string_concat`?

Gerrrr:
While implementing the first functions from this KLIP, I noticed that [existing functions throw on deserialization errors](https://github.com/confluentinc/ksql/blob/b2b6a73c5100f7ab90a3185059a5b4cf1665ee3a/ksqldb-engine/src/main/java/io/confluent/ksql/function/udf/json/JsonExtractString.java#L88). I updated the KLIP so that new functions are consistent with the existing ones - 515f8f5476806cce66953c025037b8debce1e8dc.

mjsax:
I guess from a user point of view, it does not really make a difference if a function returns `null` or throws an exception? SQL does not support exceptions, and errors (like exceptions) result in `NULL` for the user? -- For the purpose of a KLIP, it seems that throwing an exception is an implementation detail?

design-proposals/klip-59-json-functions.md
Gerrrr: I agree that we should override `json_concat` for json and string types. Thank you for the discussion!

Gerrrr:
Good point @mjsax! I will revert the change.

mjsax (approved):

design-proposals/klip-59-json-functions.md
jnh5y: @colinhicks @mjsax @agavra Apologies for a little bit of rehashing things.  While reviewing the PR for this, I asked if it'd make sense for the return type to be Map<String, String> instead of Array<Struct<Key,Key>>.  The types are isomorphic / contain the same information. 

My slight preference is for a Map, but that's based on how I think about JSON objects.  What do you all think?

Gerrrr: FWIW I don't have strong opinion on this. As Jim pointed out, the types are isomorphic, so I don't have any preference or arguments one way or another.

colinhicks: I'm pro-Map. Thanks for the suggestion, @jnh5y!

Key-iteration order is also not guaranteed for JSON objects, so choosing Map seems better in that it would also eliminate any chance folks would expect to always see keys in the same order. Another benefit with map is that users can "pluck" a single value by key. Meanwhile, the results can still be iterated over with transform, filter, and reduce.

mjsax: I don't think `Map<String,String>` would work, as we could have duplicate keys that the `Map` cannot store?

mjsax: What about duplicate keys? A `Map` could not represent it.

jnh5y: This SO answer helped me see that things could be tricky: https://stackoverflow.com/a/53529204

I don't know what to recommend here.  This is definitely a sharp edge.

Gerrrr: > What about duplicate keys?

Duplicat keys is either not an issue or a way bigger issue, depending on how we look at them. Jackson does not support duplicate keys, as `ObjectNode` maintains its fields in `LinkedHashMap` that does not allow duplicates ([link](https://github.com/FasterXML/jackson-databind/blob/279d2001db6435ea1e92c48d9e796baa74403dd7/src/main/java/com/fasterxml/jackson/databind/node/ObjectNode.java#L26)) and it is unlikely that they will support them in the near future (https://github.com/FasterXML/jackson-databind/issues/2809).

jnh5y: It seems like the options are:
* Go with Array<Struct<String,String>> (since that's the existing code and it'd cover things if there was a desire to support the JSON spec to the letter).
* Switch to Map<String, String> (since this is what the library supports).

Thoughts?  At this point, I am unsure what to suggest.

mjsax: Might be a product call, if we should follow the spec or the limitations of Jackson... \cc @colinhicks

design-proposals/klip-59-json-functions.md
```diff
@@ -0,0 +1,264 @@
...
+json_records("{\"a\": \"abc\"}) // returns [["a", "abc"]]
+```
+
+A struct is a better return type in this case as it enables meaningful field names and reduces
+the number of potential sanity checks on the end-user side (e.g., array length).
+
+### json_concat
+
+```
+json_concat(json_string, json_string) -> String
```
jnh5y, line 157: (Aleks suggested I comment here.). Should/could this method take varargs?

As one thing to think about, should we have a UDAF to concat multiple entries?  (Which would/should likely be a separate function.)

Gerrrr: I think this is great idea. ksqlDB's UDFs support varargs, so we can just change this method to `json_concat(json_strings...)` without any downsides.

mjsax: If the UDF frameworks supports it, sure!

design-proposals/klip-59-json-functions.md
Gerrrr: We discussed this offline and agreed not to support duplicate keys for now and proceed with `Map<String, String>` return type here. I updated the doc in 85de9ab.

Gerrrr:
I am going to merge the PR; happy to continue the discussion here and create follow-up PRs if necessary.

design-proposals/klip-59-json-functions.md
```diff
@@ -0,0 +1,279 @@
...
+traversing data as JSON presents an important value-add. And users of traditional databases like
+PostgreSQL already enjoy powerful support for JSON.
+
+Today, ksqlDB can:
+* Read JSON-formatted keys and values in input records.
+* Serialize keys and values as JSON in output records.
+* Extract a string value at a given JSON path.
+
+
+## What is in scope
```
IvanVas, line 24: It'd be great to cover another use case.
You receive a json string (escaped, varchar) and would like to flatten it into the output stream (format=JSON).
I'm not sure it would be possible without introducing a JSON type (union type) or making Map and Array types generic.

Possible alternative would be to have a separate serialiser (say "JSON_FLATTENED").

Previous discussion: https://confluentcommunity.slack.com/archives/C6UJNMY67/p1644308477110799?thread_ts=1644232788.546249&cid=C6UJNMY67

jnh5y: @IvanVas could you check and see if there's an issue for this and if not, open one?  I wouldn't want this idea to get lost since this PR is closed.

IvanVas: Done, thanks. 
https://github.com/confluentinc/ksql/issues/8747

PR #8555, opened 2022-01-03
Using IQv2 API for pull queries on window stores

patrickstuedi:
This PR is using the recently added IQv2 based Window and Session queries to implement window and session key lookups and table scans.

### DIFF
--- a/ksqldb-rest-app/src/test/java/io/confluent/ksql/rest/integration/PullQueryIQv2FunctionalTest.java
+++ b/ksqldb-rest-app/src/test/java/io/confluent/ksql/rest/integration/PullQueryIQv2FunctionalTest.java
@@ -15,6 +15,8 @@
 
 package io.confluent.ksql.rest.integration;
 
+import static io.confluent.ksql.rest.entity.StreamedRowMatchers.matchersRows;
+import static io.confluent.ksql.rest.entity.StreamedRowMatchers.matchersRowsAnyOrder;
 import static io.confluent.ksql.util.KsqlConfig.KSQL_STREAMS_PREFIX;
 import static org.hamcrest.MatcherAssert.assertThat;
 import static org.hamcrest.Matchers.containsInAnyOrder;
@@ -28,7 +30,9 @@ import io.confluent.ksql.integration.IntegrationTestHarness;
 import io.confluent.ksql.integration.Retry;
 import io.confluent.ksql.name.ColumnName;
 import io.confluent.ksql.rest.client.BasicCredentials;
+import io.confluent.ksql.rest.entity.KsqlHostInfoEntity;
 import io.confluent.ksql.rest.entity.StreamedRow;
+import io.confluent.ksql.rest.server.KsqlRestConfig;
 import io.confluent.ksql.rest.server.TestKsqlRestApp;
 import io.confluent.ksql.schema.ksql.LogicalSchema;
 import io.confluent.ksql.schema.ksql.PhysicalSchema;
@@ -44,6 +48,7 @@ import io.confluent.ksql.util.UserDataProvider;
 import java.io.IOException;
 import java.util.List;
 import java.util.Optional;
+import java.util.Set;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicLong;
 import java.util.stream.Collectors;
@@ -357,6 +362,146 @@ public class PullQueryIQv2FunctionalTest {
     assertThat(rowsNoHeader, containsInAnyOrder(
         ImmutableList.of(key0, 1), ImmutableList.of(key1, 1), ImmutableList.of(key2, 1),
         ImmutableList.of(key3, 1), ImmutableList.of(key4, 1)));
+  }
+
+  @Test
+  public void shouldGetSingleWindowedKey() {
+    // Given:
+    final String key = USER_PROVIDER.getStringKey(0);
+
+    makeAdminRequest(
+      "CREATE TABLE " + output + " AS"
+        + " SELECT " +  USER_PROVIDER.key() + ", COUNT(1) AS COUNT FROM " + USERS_STREAM
+        + " WINDOW TUMBLING (SIZE 1 SECOND)"
+        + " GROUP BY " + USER_PROVIDER.key() + ";"
+    );
+
+    waitForTableRows();
+
+    final String sql = "SELECT * FROM " + output
+      + " WHERE USERID = '" + key + "'"
+      + " AND WINDOWSTART = " + BASE_TIME + ";";
+
+    // When:
+    final List<StreamedRow> rows_0 = makePullQueryRequest(REST_APP_0, sql);
+
+    // Then:
+    assertThat(rows_0, hasSize(HEADER + 1));
+    assertThat(rows_0.get(1).getRow(), is(not(Optional.empty())));
+    assertThat(rows_0.get(1).getRow().get().getColumns(), is(ImmutableList.of(
+      key,                    // USERID
+      BASE_TIME,              // WINDOWSTART
+      BASE_TIME + ONE_SECOND, // WINDOWEND
+      1                       // COUNT
+    )));
+  }
+
+  @SuppressWarnings("unchecked")
+  @Test
+  public void shouldGetMultipleWindowedKeys() {
+    // Given:
+    final String key0 = USER_PROVIDER.getStringKey(0);
+    final String key1 = USER_PROVIDER.getStringKey(1);
+    final String key2 = USER_PROVIDER.getStringKey(2);
+
+    makeAdminRequest(
+      "CREATE TABLE " + output + " AS"
+        + " SELECT " +  USER_PROVIDER.key() + ", COUNT(1) AS COUNT FROM " + USERS_STREAM
+        + " WINDOW TUMBLING (SIZE 1 SECOND)"
+        + " GROUP BY " + USER_PROVIDER.key() + ";"
+    );
+
+    waitForTableRows();
+
+    final String sql = "SELECT * FROM " + output
+      + " WHERE USERID IN ('" + key0 + "'," + "'" + key1 + "'," + "'" + key2 + "') "
+      + " AND WINDOWSTART = " + BASE_TIME + ";";
+
+    // When:
+    final List<StreamedRow> rows_0 = makePullQueryRequest(REST_APP_0, sql);
+
+    // Then:
+    assertThat(rows_0, hasSize(HEADER + 3));
+    assertThat(rows_0.get(1).getRow(), is(not(Optional.empty())));
+    List<List<?>> rows = rows_0.subList(1, rows_0.size()).stream()
+      .map(sr -> sr.getRow().get().getColumns())
+      .collect(Collectors.toList());
+    assertThat(rows, containsInAnyOrder(ImmutableList.of(
+      key0,                    // USERID
+      BASE_TIME,              // WINDOWSTART
+      BASE_TIME + ONE_SECOND, // WINDOWEND
+      1                       // COUNT
+      ),
+      ImmutableList.of(
+        key1,                    // USERID
+        BASE_TIME,              // WINDOWSTART
+        BASE_TIME + ONE_SECOND, // WINDOWEND
+        1                       // COUNT
+      ),
+      ImmutableList.of(
+        key2,                    // USERID
+        BASE_TIME,              // WINDOWSTART
+        BASE_TIME + ONE_SECOND, // WINDOWEND
+        1                       // COUNT
+      )));
+  }
+
+  @Test
+  public void shouldDoTableScanWindowedBothNodes() {
+    // Given:
+    makeAdminRequest(
+      "CREATE TABLE " + output + " AS"
+        + " SELECT " +  USER_PROVIDER.key() + ", COUNT(1) AS COUNT FROM " + USERS_STREAM
+        + " WINDOW TUMBLING (SIZE 1 SECOND)"
+        + " GROUP BY " + USER_PROVIDER.key() + ";"
+    );
+
+    waitForTableRows();
+
+    final String sql = "SELECT * FROM " + output + ";";
+
+    // When:
+    final List<StreamedRow> rows_0 = makePullQueryRequest(REST_APP_0, sql);
+
+    // Then:
+    assertThat(rows_0, hasSize(HEADER + 5));
+    assertThat(rows_0.get(1).getRow(), is(not(Optional.empty())));
+
+    List<List<?>> rows = rows_0.subList(1, rows_0.size()).stream()
+      .map(sr -> sr.getRow().get().getColumns())
+      .collect(Collectors.toList());
+    assertThat(rows, containsInAnyOrder(
+      ImmutableList.of(
+        USER_PROVIDER.getStringKey(0),  // USERID
+        BASE_TIME,                      // WINDOWSTART
+        BASE_TIME + ONE_SECOND,         // WINDOWEND
+        1                               // COUNT
+      ),
+      ImmutableList.of(
+        USER_PROVIDER.getStringKey(1),  // USERID
+        BASE_TIME,                      // WINDOWSTART
+        BASE_TIME + ONE_SECOND,         // WINDOWEND
+        1                               // COUNT
+      ),
+      ImmutableList.of(
+        USER_PROVIDER.getStringKey(2),  // USERID
+        BASE_TIME,                      // WINDOWSTART
+        BASE_TIME + ONE_SECOND,         // WINDOWEND
+        1                               // COUNT
+      ),
+      ImmutableList.of(
+        USER_PROVIDER.getStringKey(3),  // USERID
+        BASE_TIME,                      // WINDOWSTART
+        BASE_TIME + ONE_SECOND,         // WINDOWEND
+        1                               // COUNT
+      ),
+      ImmutableList.of(
+        USER_PROVIDER.getStringKey(4),  // USERID
+        BASE_TIME,                      // WINDOWSTART
+        BASE_TIME + ONE_SECOND,         // WINDOWEND
+        1                               // COUNT
+      )
+    ));
   }
 
 
--- a/ksqldb-streams/src/main/java/io/confluent/ksql/execution/streams/materialization/ks/KsMaterializedSessionTableIQv2.java
+++ b/ksqldb-streams/src/main/java/io/confluent/ksql/execution/streams/materialization/ks/KsMaterializedSessionTableIQv2.java
@@ -1,0 +1,134 @@
+/*
+ * Copyright 2019 Confluent Inc.
+ *
+ * Licensed under the Confluent Community License (the "License"); you may not use
+ * this file except in compliance with the License.  You may obtain a copy of the
+ * License at
+ *
+ * http://www.confluent.io/confluent-community-license
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OF ANY KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations under the License.
+ */
+
+package io.confluent.ksql.execution.streams.materialization.ks;
+
+import static org.apache.kafka.streams.query.StateQueryRequest.inStore;
+
+import com.google.common.collect.ImmutableList;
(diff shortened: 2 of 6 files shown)


ksqldb-streams/src/main/java/io/confluent/ksql/execution/streams/materialization/ks/KsMaterializedSessionTable.java
```diff
@@ -78,13 +69,19 @@
   }
 
   private List<WindowedRow> findSession(
-      final ReadOnlySessionStore<GenericKey, GenericRow> store,
       final GenericKey key,
       final Range<Instant> windowStart,
       final Range<Instant> windowEnd
   ) {
+
+    final WindowRangeQuery<GenericKey, GenericRow> query = WindowRangeQuery.withKey(key);
```
guozhangwang, line 77: I'm wondering why we cannot now pass in the `windowStart/End` as query parameters directly?

guozhangwang (commented):
I have a meta question regarding the unit tests: for the removed ones, some of those are testing `shouldThrow` etc and are no longer meaningful, but some others like `should close iterator` etc, should we still have corresponding test cases for them?

jaceklaskowski (commented):
LGTM

ksqldb-streams/src/main/java/io/confluent/ksql/execution/streams/materialization/ks/KsMaterializedSessionTableIQv2.java
```diff
@@ -0,0 +1,115 @@
...
+  ) {
+
+    final WindowRangeQuery<GenericKey, GenericRow> query = WindowRangeQuery.withKey(key);
+    final StateQueryRequest<KeyValueIterator<Windowed<GenericKey>, GenericRow>> request =
+        inStore(stateStore.getStateStoreName()).withQuery(query);
+    final StateQueryResult<KeyValueIterator<Windowed<GenericKey>, GenericRow>> result =
+        stateStore.getKafkaStreams().query(request);
+
+    try (KeyValueIterator<Windowed<GenericKey>, GenericRow> it =
+             result.getGlobalResult().getResult()) {
```
vvcephei, line 84: We won't be querying a global store, so there won't be a global result. Looking through the code from the old cache bypass logic, I think you'll want to do something like this `io.confluent.ksql.execution.streams.materialization.ks.SessionStoreCacheBypass#findFirstNonEmptyIterator` on the PartitionResults instead.

ksqldb-streams/src/main/java/io/confluent/ksql/execution/streams/materialization/ks/KsMaterializedWindowTableIQv2.java
```diff
@@ -0,0 +1,190 @@
...
+      final WindowKeyQuery<GenericKey, ValueAndTimestamp<GenericRow>> query =
+          WindowKeyQuery.withKeyAndWindowStartRange(key, lower, upper);
+      final StateQueryRequest<WindowStoreIterator<ValueAndTimestamp<GenericRow>>> request =
+          inStore(stateStore.getStateStoreName()).withQuery(query);
+      final KafkaStreams streams = stateStore.getKafkaStreams();;
+      final StateQueryResult<WindowStoreIterator<ValueAndTimestamp<GenericRow>>> result =
+          streams.query(request);
+
+      try (WindowStoreIterator<ValueAndTimestamp<GenericRow>> it
+          = result.getGlobalResult().getResult()) {
```
vvcephei, line 79: Here, you'll want to use the `partition` to get the specific partition result (as above, this isn't a global store, so there won't be a global result).

ksqldb-streams/src/main/java/io/confluent/ksql/execution/streams/materialization/ks/KsMaterializedWindowTableIQv2.java
```diff
@@ -0,0 +1,190 @@
...
+      final Instant upper = calculateUpperBound(windowStartBounds, windowEndBounds);
+
+      final WindowRangeQuery<GenericKey, ValueAndTimestamp<GenericRow>> query =
+          WindowRangeQuery.withWindowStartRange(lower, upper);
+      final StateQueryRequest<KeyValueIterator<Windowed<GenericKey>, ValueAndTimestamp<GenericRow>>>
+          request = inStore(stateStore.getStateStoreName()).withQuery(query);
+      final StateQueryResult<KeyValueIterator<Windowed<GenericKey>, ValueAndTimestamp<GenericRow>>>
+          result = stateStore.getKafkaStreams().query(request);
+      final KeyValueIterator<Windowed<GenericKey>, ValueAndTimestamp<GenericRow>> iterator =
+          result.getGlobalResult().getResult();
```
vvcephei, line 131: Same deal here.

ksqldb-streams/src/test/java/io/confluent/ksql/execution/streams/materialization/ks/KsMaterializedWindowTableIQv2Test.java
```diff
@@ -0,0 +1,306 @@
...
+  public void shouldThrowNPEs() {
+    new NullPointerTester()
+        .setDefault(KsStateStore.class, stateStore)
+        .testConstructors(KsMaterializedWindowTable.class, Visibility.PACKAGE);
+  }
+
+  @Test
+  public void shouldThrowIfGettingStateStoreFails() {
+    // Given:
+    when(stateStore.getKafkaStreams()).thenThrow(new MaterializationTimeOutException("Boom"));
```
vvcephei, line 144: It's hard to imagine that this method itself would throw an exception. Maybe we should make sure we get the right exception if the query itself throws and also if the query just returns a "failed" result?

ksqldb-streams/src/test/java/io/confluent/ksql/execution/streams/materialization/ks/KsMaterializedWindowTableIQv2Test.java
```diff
@@ -0,0 +1,306 @@
...
+    assertThat(e.getMessage(), containsString(
+        "Failed to get value from materialized table"));
+    assertThat(e.getCause(), (instanceOf(MaterializationTimeOutException.class)));
+  }
+
+  @Test
+  public void shouldThrowIfStoreFetchFails_fetchAll() {
+    // Given:
+    when(kafkaStreams.query(any()))
+      .thenThrow(new MaterializationTimeOutException("Boom"));
```
vvcephei, line 197: I think several of these tests are copied and modified from the IQv1 test (which is good), but I don't think they've been modified in a way that makes sense all the time. For example, this test is identical to the previous one.

ksqldb-streams/src/test/java/io/confluent/ksql/execution/streams/materialization/ks/KsMaterializedWindowTableIQv2Test.java
```diff
@@ -0,0 +1,306 @@
...
+    // Given:
+    final Range<Instant> start = Range.closed(
+      NOW,
+      NOW.plusSeconds(10)
+    );
+
+    when(kafkaStreams.query(any())).thenReturn(queryResult);
+    when(queryResult.getGlobalResult()).thenReturn(fetchResult);
+    when(fetchResult.getResult()).thenReturn(fetchIterator);
+    when(fetchIterator.hasNext()).thenReturn(true, true, false);
```
vvcephei, line 222: Similar concern to https://github.com/confluentinc/ksql/pull/8556#discussion_r779933662

vvcephei (commented):
Thanks, @patrickstuedi ! 

There are a few issues, but nothing crazy; it looks good to me overall!

In addition to the unit test, it'd be good to have at least one integration test to verify that the IQv2 code actually works when integrated.

Thanks!

ksqldb-streams/src/main/java/io/confluent/ksql/execution/streams/materialization/ks/KsMaterializedSessionTable.java
patrickstuedi: yea..the old code didn't do it but manually processed the records so I assumed it to be safer to do it the same way. I suggest we make it the goal of this PR to create parity with the IQv1, and then add optimizations like this as small follow-up PRs.

ksqldb-streams/src/main/java/io/confluent/ksql/execution/streams/materialization/ks/KsMaterializedSessionTableIQv2.java
patrickstuedi: Ok, makes sense, changed that part, let me know what you think.

ksqldb-streams/src/main/java/io/confluent/ksql/execution/streams/materialization/ks/KsMaterializedWindowTableIQv2.java
patrickstuedi: Makes sense, done.

ksqldb-streams/src/test/java/io/confluent/ksql/execution/streams/materialization/ks/KsMaterializedWindowTableIQv2Test.java
patrickstuedi: mmh, I can't find any duplicates

ksqldb-streams/src/test/java/io/confluent/ksql/execution/streams/materialization/ks/KsMaterializedWindowTableIQv2Test.java
patrickstuedi: I changed this one to not use the globalresult

ksqldb-streams/src/main/java/io/confluent/ksql/execution/streams/materialization/ks/KsMaterializedSessionTableIQv2.java
```diff
@@ -0,0 +1,136 @@
...
+      final GenericKey key,
+      final int partition,
+      final Range<Instant> windowStart,
+      final Range<Instant> windowEnd
+  ) {
+    try {
+      return findSession(key, partition, windowStart, windowEnd);
+    } catch (final Exception e) {
+      throw new MaterializationException("Failed to get value from materialized table", e);
+    }
```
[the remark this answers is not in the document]
vvcephei: Also applies to the other try/catch blocks.

ksqldb-streams/src/main/java/io/confluent/ksql/execution/streams/materialization/ks/KsMaterializedSessionTableIQv2.java
```diff
@@ -0,0 +1,138 @@
...
+        + result.getPartitionResults());
+    }
+
+    if (!partitionResults.containsKey(partition)) {
+      throw new MaterializationException("The query did not return the requested partition result: "
+        + result.getPartitionResults());
+    }
+
+    final QueryResult<KeyValueIterator<Windowed<GenericKey>, GenericRow>> queryResult =
+        partitionResults.get(partition);
```
vvcephei, line 100:
```suggestion
    final QueryResult<KeyValueIterator<Windowed<GenericKey>, GenericRow>> queryResult =
        result.getPartitionResults().get(partition);
```

See https://github.com/confluentinc/ksql/pull/8570

ksqldb-streams/src/main/java/io/confluent/ksql/execution/streams/materialization/ks/KsMaterializedWindowTableIQv2.java
```diff
@@ -0,0 +1,233 @@
...
+          + result.getPartitionResults());
+      }
+
+      if (!partitionResults.containsKey(partition)) {
+        throw new MaterializationException("The query did not return the requested partition "
+            + "result: " + result.getPartitionResults());
+      }
+
+      final QueryResult<WindowStoreIterator<ValueAndTimestamp<GenericRow>>> queryResult =
+          partitionResults.get(partition);
```
vvcephei, line 94: As above.

ksqldb-streams/src/main/java/io/confluent/ksql/execution/streams/materialization/ks/KsMaterializedWindowTableIQv2.java
```diff
@@ -0,0 +1,233 @@
...
+          + result.getPartitionResults());
+      }
+
+      if (!partitionResults.containsKey(partition)) {
+        throw new MaterializationException("The query did not return the requested partition "
+            + "result: " + result.getPartitionResults());
+      }
+
+      final QueryResult<KeyValueIterator<Windowed<GenericKey>,
+          ValueAndTimestamp<GenericRow>>> queryResult = partitionResults.get(partition);
```
vvcephei, line 167: As above.

ksqldb-streams/src/test/java/io/confluent/ksql/execution/streams/materialization/ks/KsMaterializedWindowTableIQv2Test.java
```diff
@@ -0,0 +1,321 @@
...
+    final Exception e = assertThrows(
+        MaterializationException.class,
+        () -> table.get(PARTITION, WINDOW_START_BOUNDS, WINDOW_END_BOUNDS)
+    );
+
+    // Then:
+    assertThat(e.getMessage(), containsString(
+        "Failed to scan materialized table"));
+    assertThat(e.getCause(), (instanceOf(MaterializationTimeOutException.class)));
+  }
```
vvcephei, line 189: I don't think we need this test (plus, it's not testing what it says it is).

I.e., it's testing what happens if `stateStore.getKafkaStreams()` throws an exception, which will never happen (because that method is just a field getter).

ksqldb-streams/src/test/java/io/confluent/ksql/execution/streams/materialization/ks/KsMaterializedWindowTableIQv2Test.java
```diff
@@ -0,0 +1,321 @@
...
+  @SuppressWarnings("UnstableApiUsage")
+  @Test
+  public void shouldThrowNPEs() {
+    new NullPointerTester()
+        .setDefault(KsStateStore.class, stateStore)
+        .testConstructors(KsMaterializedWindowTable.class, Visibility.PACKAGE);
+  }
+
+  @Test
+  public void shouldThrowIfGettingStateStoreFails() {
```
vvcephei, line 139:
```suggestion
  public void shouldThrowIfQueryThrows() {
```

ksqldb-streams/src/test/java/io/confluent/ksql/execution/streams/materialization/ks/KsMaterializedWindowTableIQv2Test.java
```diff
@@ -0,0 +1,321 @@
...
+    );
+
+    // Then:
+    assertThat(e.getMessage(), containsString(
+        "Failed to get value from materialized table"));
+    assertThat(e.getCause(), (instanceOf(MaterializationTimeOutException.class)));
+  }
+
+  @Test
+  public void shouldThrowIfGettingStateStoreFailedResult() {
```
vvcephei, line 156:
```suggestion
  public void shouldThrowIfQueryFails() {
```

ksqldb-streams/src/test/java/io/confluent/ksql/execution/streams/materialization/ks/KsMaterializedWindowTableIQv2Test.java
```diff
@@ -0,0 +1,321 @@
...
+    final Exception e = assertThrows(
+        MaterializationException.class,
+        () -> table.get(A_KEY, PARTITION, WINDOW_START_BOUNDS, WINDOW_END_BOUNDS)
+    );
+
+    // Then:
+    assertThat(e.getMessage(), containsString(
+        "Failed to get value from materialized table"));
+    assertThat(e.getCause(), (instanceOf(MaterializationTimeOutException.class)));
+  }
```
vvcephei, line 207: This test is a duplicate of `shouldThrowIfGettingStateStoreFails` above, although in retrospect, I guess that other test is a duplicate of this one, since the name if this test is more accurate.

ksqldb-streams/src/test/java/io/confluent/ksql/execution/streams/materialization/ks/KsMaterializedWindowTableIQv2Test.java
```diff
@@ -0,0 +1,321 @@
...
+import org.junit.Before;
+import org.junit.Test;
+import org.junit.runner.RunWith;
+import org.mockito.ArgumentCaptor;
+import org.mockito.Captor;
+import org.mockito.Mock;
+import org.mockito.junit.MockitoJUnitRunner;
+
+@RunWith(MockitoJUnitRunner.class)
+public class KsMaterializedWindowTableIQv2Test {
```
vvcephei, line 73: As in the other tests, I think we should preserve more of the tests from the old test class:
* shouldCloseIterator
* shouldCloseIterator_fetchAll
* shouldReturnEmptyIfKeyNotPresent
* shouldReturnValuesForClosedEndBounds
* shouldReturnValuesForClosedEndBounds_fetchAll
* shouldReturnValuesForOpenStartBounds
* shouldReturnValuesForOpenStartBounds_fetchAll
* shouldReturnValuesForOpenEndBounds
* shouldReturnValuesForOpenEndBounds_fetchAll
* shouldMaintainResultOrder
* shouldSupportRangeAll
* shouldSupportRangeAll_fetchAll

Although, as a suggestion, it looks like the original authors of those tests went a bit overboard with the mocks. I thought the technique of returning a hand-crafted StateQueryResponse worked out pretty well in your existing tests and in https://github.com/confluentinc/ksql/pull/8556 .

vvcephei (commented):
Thanks, @patrickstuedi ! I just made a full pass on the PR. These are all my comments.

ksqldb-streams/src/main/java/io/confluent/ksql/execution/streams/materialization/ks/KsMaterializedSessionTable.java
patrickstuedi: @guozhangwang agree that some of the old tests are still valuable, I added tests for iterator closing and empty sets.

ksqldb-streams/src/main/java/io/confluent/ksql/execution/streams/materialization/ks/KsMaterializedSessionTableIQv2.java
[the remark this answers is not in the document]
patrickstuedi: Makes sense

ksqldb-streams/src/test/java/io/confluent/ksql/execution/streams/materialization/ks/KsMaterializedWindowTableIQv2Test.java
patrickstuedi: Changed this to a case where the query fails

ksqldb-streams/src/test/java/io/confluent/ksql/execution/streams/materialization/ks/KsMaterializedWindowTableIQv2Test.java
patrickstuedi: Good catch! Adjusted the test

ksqldb-rest-app/src/test/java/io/confluent/ksql/rest/integration/PullQueryIQv2FunctionalTest.java
```diff
@@ -359,6 +364,146 @@ public void shouldDoRangeQueriesUpperBound() {
         ImmutableList.of(key3, 1), ImmutableList.of(key4, 1)));
   }
 
+  @Test
```
vpapavas, line 367: I think we should add a couple of cases for the Session window as well to be complete

vpapavas (approved):
Thank you @patrickstuedi ! LGTM with one small comment about adding a couple tests cases for session stores to the functional test, just to be complete and cover our bases

PR #8556, opened 2022-01-04
feat: Use IQv2 when executing range and table scan pull queries

vpapavas:
### Description 
Now that IQv2 has been merged into Kafka Streams, we can take advantage of it in ksql. This PR uses the new API for range and full table scan queries.

### Testing done 
No new tests

### Reviewer checklist

### DIFF
--- a/ksqldb-streams/src/main/java/io/confluent/ksql/execution/streams/materialization/ks/KsMaterialization.java
+++ b/ksqldb-streams/src/main/java/io/confluent/ksql/execution/streams/materialization/ks/KsMaterialization.java
@@ -35,15 +35,12 @@ public final class KsMaterialization implements Materialization {
   private final Optional<WindowInfo> windowInfo;
   private final KsStateStore stateStore;
   private final Locator locator;
-  private final KafkaStreams kafkaStreams;
 
   KsMaterialization(
-      final KafkaStreams streams,
       final Optional<WindowInfo> windowInfo,
       final Locator locator,
       final KsStateStore stateStore
   ) {
-    this.kafkaStreams = requireNonNull(streams, "streams");
     this.windowInfo = requireNonNull(windowInfo, "windowInfo");
     this.stateStore = requireNonNull(stateStore, "stateStore");
     this.locator = requireNonNull(locator, "locator");
@@ -69,7 +66,7 @@ public final class KsMaterialization implements Materialization {
     if (windowInfo.isPresent()) {
       throw new UnsupportedOperationException("Table has windowed key");
     }
-    return new KsMaterializedTable(kafkaStreams, stateStore);
+    return new KsMaterializedTable(stateStore);
   }
 
   @SuppressWarnings("OptionalGetWithoutIsPresent") // Enforced by type
--- a/ksqldb-streams/src/main/java/io/confluent/ksql/execution/streams/materialization/ks/KsMaterialization.java
+++ b/ksqldb-streams/src/main/java/io/confluent/ksql/execution/streams/materialization/ks/KsMaterialization.java
@@ -35,18 +35,15 @@ public final class KsMaterialization implements Materialization {
   private final Optional<WindowInfo> windowInfo;
   private final KsStateStore stateStore;
   private final Locator locator;
-  private final KsqlConfig ksqlConfig;
 
   KsMaterialization(
       final Optional<WindowInfo> windowInfo,
       final Locator locator,
-      final KsStateStore stateStore,
-      final KsqlConfig ksqlConfig
+      final KsStateStore stateStore
   ) {
     this.windowInfo = requireNonNull(windowInfo, "windowInfo");
     this.stateStore = requireNonNull(stateStore, "stateStore");
     this.locator = requireNonNull(locator, "locator");
-    this.ksqlConfig = requireNonNull(ksqlConfig, "ksqlConfig");
   }
 
   @Override
@@ -69,7 +66,8 @@ public final class KsMaterialization implements Materialization {
     if (windowInfo.isPresent()) {
       throw new UnsupportedOperationException("Table has windowed key");
     }
-    if (ksqlConfig.getBoolean(KsqlConfig.KSQL_QUERY_PULL_CONSISTENCY_OFFSET_VECTOR_ENABLED)) {
+    if (stateStore.getKsqlConfig().getBoolean(
+        KsqlConfig.KSQL_QUERY_PULL_CONSISTENCY_OFFSET_VECTOR_ENABLED)) {
       return new KsMaterializedTableIQv2(stateStore);
     }
     return new  KsMaterializedTable(stateStore);
--- a/ksqldb-streams/src/main/java/io/confluent/ksql/execution/streams/materialization/ks/KsMaterialization.java
+++ b/ksqldb-streams/src/main/java/io/confluent/ksql/execution/streams/materialization/ks/KsMaterialization.java
@@ -25,7 +25,6 @@ import io.confluent.ksql.model.WindowType;
 import io.confluent.ksql.schema.ksql.LogicalSchema;
 import io.confluent.ksql.serde.WindowInfo;
 import java.util.Optional;
-import org.apache.kafka.streams.KafkaStreams;
 
 /**
  * Kafka Streams impl of {@link Materialization}.
--- a/ksqldb-streams/src/main/java/io/confluent/ksql/execution/streams/materialization/ks/KsMaterialization.java
+++ b/ksqldb-streams/src/main/java/io/confluent/ksql/execution/streams/materialization/ks/KsMaterialization.java
@@ -25,6 +25,7 @@ import io.confluent.ksql.model.WindowType;
 import io.confluent.ksql.schema.ksql.LogicalSchema;
 import io.confluent.ksql.serde.WindowInfo;
 import java.util.Optional;
+import org.apache.kafka.streams.KafkaStreams;
 
 /**
  * Kafka Streams impl of {@link Materialization}.
@@ -34,12 +35,15 @@ public final class KsMaterialization implements Materialization {
   private final Optional<WindowInfo> windowInfo;
   private final KsStateStore stateStore;
   private final Locator locator;
+  private final KafkaStreams kafkaStreams;
 
   KsMaterialization(
+      final KafkaStreams streams,
       final Optional<WindowInfo> windowInfo,
       final Locator locator,
       final KsStateStore stateStore
   ) {
+    this.kafkaStreams = requireNonNull(streams, "streams");
     this.windowInfo = requireNonNull(windowInfo, "windowInfo");
     this.stateStore = requireNonNull(stateStore, "stateStore");
     this.locator = requireNonNull(locator, "locator");
@@ -65,7 +69,7 @@ public final class KsMaterialization implements Materialization {
     if (windowInfo.isPresent()) {
       throw new UnsupportedOperationException("Table has windowed key");
     }
-    return new KsMaterializedTable(stateStore);
+    return new KsMaterializedTable(kafkaStreams, stateStore);
   }
 
   @SuppressWarnings("OptionalGetWithoutIsPresent") // Enforced by type
--- a/ksqldb-streams/src/main/java/io/confluent/ksql/execution/streams/materialization/ks/KsMaterialization.java
+++ b/ksqldb-streams/src/main/java/io/confluent/ksql/execution/streams/materialization/ks/KsMaterialization.java
@@ -24,6 +24,7 @@ import io.confluent.ksql.execution.streams.materialization.MaterializedWindowedT
 import io.confluent.ksql.model.WindowType;
 import io.confluent.ksql.schema.ksql.LogicalSchema;
 import io.confluent.ksql.serde.WindowInfo;
+import io.confluent.ksql.util.KsqlConfig;
 import java.util.Optional;
 
 /**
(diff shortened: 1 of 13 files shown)


ksqldb-streams/src/main/java/io/confluent/ksql/execution/streams/materialization/ks/KsMaterialization.java
```diff
@@ -34,12 +35,15 @@
...
   private final Locator locator;
+  private final KafkaStreams kafkaStreams;
 
   KsMaterialization(
+      final KafkaStreams streams,
       final Optional<WindowInfo> windowInfo,
       final Locator locator,
       final KsStateStore stateStore
   ) {
+    this.kafkaStreams = requireNonNull(streams, "streams");
```
patrickstuedi, line 46: Do we need to pass kafkastreams here? Why not use stateStore.getKafkaStreams()?

patrickstuedi (commented):
Thanks @vpapavas, LGTM, just left one comment.

ksqldb-streams/src/main/java/io/confluent/ksql/execution/streams/materialization/ks/KsMaterialization.java
vpapavas: I don't see the method `stateStore.getKafkaStreams()`? Am I missing something?

vpapavas: Ah, I think you meant adding this method? Sure, we can do that

patrickstuedi: My mistake, sorry, I didn't remember I added it in my PR so I assumed it was available. Yea adding it probably makes sense

ksqldb-streams/src/main/java/io/confluent/ksql/execution/streams/materialization/ks/KsMaterializedTable.java
```diff
@@ -60,13 +70,19 @@
...
+          .withPartitions(ImmutableSet.of(partition));
+      final StateQueryResult<KeyValueIterator<GenericKey, ValueAndTimestamp<GenericRow>>>
+          result = stateStore.getKafkaStreams().query(request);
+      final KeyValueIterator<GenericKey, ValueAndTimestamp<GenericRow>> iterator =
+          result.getOnlyPartitionResult().getResult();
+      
       return Streams.stream(IteratorUtil.onComplete(iterator, iterator::close))
           .map(keyValue -> Row.of(stateStore.schema(), keyValue.key, keyValue.value.value(),
-              keyValue.value.timestamp()))
+                                  keyValue.value.timestamp()))
```
guozhangwang, line 85: This is not for this PR, but just a thought: could we save on two iterations from the store query result here, by having a delegating iterator of `Iterator<Row>` whose implementation calls the underlying `Iterator<GenericKey, ValueAndTimestamp<GenericRow>>`? I'm asking only because for large returned iterable sets, `Streams.stream()` will iterate and copy-paste into another set and the returns an iterator, which may take non-negligible time.

guozhangwang (approved):

ksqldb-streams/src/main/java/io/confluent/ksql/execution/streams/materialization/ks/KsMaterializedTable.java
vpapavas: Yes, that makes total sense. I will open a ticket to implement it in a follow up PR.

ksqldb-streams/src/test/java/io/confluent/ksql/execution/streams/materialization/ks/KsMaterializedTableTest.java
```diff
@@ -105,9 +118,9 @@ public void shouldThrowNPEs() {
   }
 
   @Test
-  public void shouldThrowIfGettingStateStoreFails() {
+  public void shouldThrowIfQueryFails() {
     // Given:
-    when(stateStore.store(any(), anyInt())).thenThrow(new MaterializationTimeOutException("Boom"));
+    when(stateStore.getKafkaStreams().query(any())).thenThrow(new MaterializationTimeOutException("Boom"));
```
vvcephei, line 123: We might also return a failed QueryResult. Should we add a test for that?

ksqldb-streams/src/test/java/io/confluent/ksql/execution/streams/materialization/ks/KsMaterializedTableTest.java
```diff
@@ -122,56 +135,125 @@ public void shouldThrowIfGettingStateStoreFails() {
...
 
   @Test
-  public void shouldThrowIfStoreGetFails() {
+  public void shouldKeyQueryWithCorrectParams() {
     // Given:
-    when(tableStore.get(any())).thenThrow(new MaterializationTimeOutException("Boom"));
+    when(kafkaStreams.query(any())).thenReturn(queryResult);
+    when(queryResult.getOnlyPartitionResult()).thenReturn(keyResult);
+    when(keyResult.getResult()).thenReturn(row);
+    when(row.value()).thenReturn(ROW1);
```
vvcephei, line 143: Man, this mocking stuff is brutal. Do you think it would be a good idea to add a utility method to create results like:

```
private static <R> StateQueryResult<ValueAndTimestamp<R> makeResult(final R result) {
        final StateQueryResult<ValueAndTimestamp<R>> result = new StateQueryResult<>();
        result.addResult(PARTITION, QueryResult.forResult(ValueAndTimestamp.make(result, -1)));
        return result;
}
```

And then be able to just do:
```suggestion
    when(kafkaStreams.query(any())).thenReturn(makeResult(ROW1));
```

The reason it might be nice is that if we change how the code interacts with the result, we're going to have to change every single test case as well. E.g., if you take the suggestion to get the specific partition instead of "getOnlyPartitionResult", or when we add logic to extract the Position later on. It's better if unit tests are generally able to ignore irrelevant interactions, since changing code AND tests is always riskier than just changing code and observing that all the tests still pass.

Just a suggestion, feel free to take it/leave it/leave it for future refactoring.

ksqldb-streams/src/main/java/io/confluent/ksql/execution/streams/materialization/ks/KsMaterializedTable.java
```diff
@@ -76,15 +92,30 @@
...
 
+      final StateQueryRequest<KeyValueIterator<GenericKey, ValueAndTimestamp<GenericRow>>>
+          request = inStore(stateStore.getStateStoreName())
+          .withQuery(query)
+          .withPartitions(ImmutableSet.of(partition));
+      final StateQueryResult<KeyValueIterator<GenericKey, ValueAndTimestamp<GenericRow>>>
+          result = stateStore.getKafkaStreams().query(request);
       final KeyValueIterator<GenericKey, ValueAndTimestamp<GenericRow>> iterator =
-          store.range(from, to);
+          result.getOnlyPartitionResult().getResult();
```
vvcephei, line 113: We don't really need to do "get only partition result" because we already know the partition we requested. I doesn't matter much, but it seems more direct to say:

```suggestion
          result.getPartitionResults().get(partition).getResult();
```

(Also applies to the other query)

ksqldb-streams/src/test/java/io/confluent/ksql/execution/streams/materialization/ks/KsMaterializedTableTest.java
```diff
@@ -122,56 +135,125 @@ public void shouldThrowIfGettingStateStoreFails() {
...
+    RangeQuery rangeQuery = (RangeQuery)request.getQuery();
+    assertThat(rangeQuery.getLowerBound(), is(Optional.empty()));
+    assertThat(rangeQuery.getUpperBound(), is(Optional.empty()));
   }
 
   @Test
   public void shouldReturnEmptyIfKeyNotPresent() {
+    // Given:
+    when(kafkaStreams.query(any())).thenReturn(queryResult);
+    when(queryResult.getOnlyPartitionResult()).thenReturn(keyResult);
```
vvcephei, line 255: Similar to the above, we have no guarantee that this mocked behavior accurately reflects how a real StateQueryResult works when there's no key found. Returning a real `makeResult(null)` would be more airtight.

vvcephei (commented):
Thanks, @vpapavas !

It looks good to me overall, just a few smallish comments.

Edit: Oh, I also just realized that we're going to need a feature flag. We should probably do the same thing as in https://github.com/confluentinc/ksql/pull/8555 : copy the KsMaterializedTable and test instead of modifying them, and then choose between them based on the config.

ksqldb-streams/src/main/java/io/confluent/ksql/execution/streams/materialization/ks/KsMaterializedTable.java
vvcephei: Is this right? It looks to me like the `Stream.stream` and the subsequent `Stream#iterator` are both lazy.

vvcephei:
There are a couple of test failures, by the way:     io.confluent.ksql.materialization.ks.KsMaterializationFunctionalTest.shouldQueryMaterializedTableForAggregatedStream
    io.confluent.ksql.materialization.ks.KsMaterializationFunctionalTest.shouldQueryMaterializedTableForAggregatedTable

ksqldb-streams/src/main/java/io/confluent/ksql/execution/streams/materialization/ks/KsMaterializedTable.java
vpapavas: Already on it :) Unfortunately, you reviewed a stale version

ksqldb-streams/src/test/java/io/confluent/ksql/execution/streams/materialization/ks/KsMaterializedTableTest.java
vpapavas: This is a great suggestions, thank you!

ksqldb-streams/src/main/java/io/confluent/ksql/execution/streams/materialization/ks/KsMaterialization.java
```diff
@@ -34,15 +35,18 @@
...
   private final KsStateStore stateStore;
   private final Locator locator;
+  private final KsqlConfig ksqlConfig;
 
   KsMaterialization(
       final Optional<WindowInfo> windowInfo,
       final Locator locator,
-      final KsStateStore stateStore
+      final KsStateStore stateStore,
+      final KsqlConfig ksqlConfig
```
patrickstuedi, line 44: I think you could skip adding the config as it is available in the stateStore already

vpapavas: Yes, already did that. Thanks

ksqldb-streams/src/test/java/io/confluent/ksql/execution/streams/materialization/ks/KsMaterializedTableTest.java
vpapavas: We definitely should! :)

ksqldb-streams/src/main/java/io/confluent/ksql/execution/streams/materialization/ks/KsMaterializedTableIQv2.java
```diff
@@ -0,0 +1,187 @@
...
+      if (queryResult.isFailure()) {
+        throw new MaterializationException(queryResult.getFailureMessage());
+      }
+      if (queryResult.getResult() == null) {
+        return Optional.empty();
+      } else {
+        final ValueAndTimestamp<GenericRow> row = queryResult.getResult();
+        return Optional.of(Row.of(stateStore.schema(), key, row.value(), row.timestamp()));
+      }
+    } catch (final Exception e) {
```
vvcephei, line 88:
```suggestion
    } catch (final MaterializationException e) {
        throw e;
    } catch (final Exception e) {
```

ksqldb-streams/src/test/java/io/confluent/ksql/execution/streams/materialization/ks/KsMaterializedTableIQv2Test.java
```diff
@@ -127,9 +127,9 @@ public void shouldThrowIfQueryFails() {
     );
 
     // Then:
-    assertThat(e.getMessage(), containsString(
-        "Failed to get value from materialized table"));
-    assertThat(e.getCause(), (instanceOf(MaterializationTimeOutException.class)));
+    System.out.println(e.getMessage());
```
vvcephei, line 130: Can you do a quick follow-up to remove this println?

ksqldb-streams/src/main/java/io/confluent/ksql/execution/streams/materialization/ks/KsMaterializedTableIQv2.java
```diff
@@ -0,0 +1,193 @@
...
+          result.getPartitionResults();
+      if (queryResultMap.size() != 1) {
+        throw new MaterializationException("The query did not return exactly one partition result: "
+                + result.getPartitionResults());
+      }
+      if (!queryResultMap.containsKey(partition)) {
+        throw new MaterializationException(
+            "The query did not return the requested partition result: "
+                + result.getPartitionResults());
+      }
```
vvcephei, line 76: These _should_ never happen, but I guess it doesn't hurt to check.

ksqldb-streams/src/main/java/io/confluent/ksql/execution/streams/materialization/ks/KsMaterializedTableIQv2.java
```diff
@@ -0,0 +1,193 @@
...
+      final QueryResult<ValueAndTimestamp<GenericRow>> queryResult =
+          queryResultMap.get(partition);
+      if (queryResult.isFailure()) {
+        throw new MaterializationException(queryResult.getFailureMessage());
+      }
```
vvcephei, line 81: Note: you could just not handle this case. If you try and do `getResult` on a failed result, it already throws an exception that contains the failure reason and message:

https://github.com/apache/kafka/blob/trunk/streams/src/main/java/org/apache/kafka/streams/query/internals/FailedQueryResult.java#L84-L88

vvcephei (approved):
Thanks, @vpapavas !

I had some comments related to cleanup, which I think we can just handle in a follow-on PR.

PR #8563, opened 2022-01-05
feat: un-synchronize PullQueryQueue row queuing when limit absent

### LINKED ISSUE #8517: Pull queries with LIMIT clause might return more rows than specified by the limit

cprasad1:
### Description 
As @Gerrrr mentioned in #8517, there are multiple current issues with our current mechanism:
- First of all, PullQueryQueue #acceptRow is synchronized for all pull queries, not just the ones that have LIMIT clause.
- The entire method is a critical section which is very much suboptimal given we run a callback there. The only practical usage of that callback is to publish messages, so we'd rather not do that under a lock.
- AFAIU a lock is not even needed in that case - we just need to update totalRowsQueued with CAS. Without contention there is no additional overhead required to ensure thread-safety.

This PR fixes #8517 

### Testing done 

### Reviewer checklist

### DIFF
--- a/ksqldb-engine/src/main/java/io/confluent/ksql/query/PullQueryQueue.java
+++ b/ksqldb-engine/src/main/java/io/confluent/ksql/query/PullQueryQueue.java
@@ -217,30 +217,54 @@ public class PullQueryQueue implements BlockingRowQueue {
    * Enqueues a row on the queue.  Blocks until the row can be accepted.
    * @param row The row to enqueue.
    */
-  public synchronized boolean acceptRow(final PullQueryRow row) {
+  @SuppressWarnings({"checkstyle:CyclomaticComplexity"})
+  public boolean acceptRow(final PullQueryRow row) {
     try {
       if (row == null) {
         return false;
       }
-      if (limit.isPresent() && totalRowsQueued.get() >= limit.getAsInt()) {
-        closeInternal(true);
+
+      // fast-path for pull queries without limit
+      if (!this.limit.isPresent()) {
+        if (doAcceptRow(row)) {
+          totalRowsQueued.incrementAndGet();
+          return true;
+        }
         return false;
       }
 
-      while (!closed.get()) {
-        if (rowQueue.offer(row, offerTimeoutMs, TimeUnit.MILLISECONDS)) {
-          totalRowsQueued.incrementAndGet();
-          queuedCallback.run();
-          if (limit.isPresent() && totalRowsQueued.get() >= limit.getAsInt()) {
+      final int limit = this.limit.getAsInt();
+      while (true) {
+        final long rowsQueued = totalRowsQueued.get();
+        if (rowsQueued >= limit) {
+          closeInternal(true);
+          return false;
+        } else if (totalRowsQueued.compareAndSet(rowsQueued, rowsQueued + 1)) {
+          final boolean accepted = doAcceptRow(row);
+          if (accepted && rowsQueued + 1 >= limit) {
             closeInternal(true);
+          } else if (!accepted) {
+            // we updated the counter speculatively, but failed to add the row, therefore have
+            // to decrement the counter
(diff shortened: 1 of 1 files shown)


ksqldb-engine/src/main/java/io/confluent/ksql/query/PullQueryQueue.java
```diff
@@ -245,6 +259,16 @@ public synchronized boolean acceptRow(final PullQueryRow row) {
     return false;
   }
 
+  private boolean doAcceptRow(final PullQueryRow row) throws InterruptedException {
```
vvcephei, line 262: specific ask: I'd prefer to do everything in one loop, but if it's better not to, can you either inline this method or extract a method for the other loop?

It looked to me like if we didn't have a limit, we'd only try to insert once, which is not the case.

vvcephei (changes requested):
Thanks @cprasad1 ( and @Gerrrr )

I agree with the spirit of reducing contention via the CAS instead of synchronizing. It would be good, though, if we could get an implementation that's roughly as "obviously" correct, though, which I'm not sure this one is.

The fact that there are now two different implementations of the loop, depending on whether there's a limit or not is troubling. The fact that most of the writers will exit early after the winner speculatively increments the counter, but then the winner might fail, is also troubling.

I'm not a fan of code-golfing in PR reviews, but I'm a little concerned about correctness and maintainability in this case. Do you think we can simplify the code and/or add more comments explaining why it's ok to do anything that seems sketchy?

Gerrrr:
Thank you @cprasad1 for coming back to this! There was a valid argument that the CAS solution is hard to reason about. As an idea, we can achieve pretty much the same behavior by synchronizing only the `LIMIT` and running the callback outside of the lock. For example:

```java

  public boolean acceptRow(final PullQueryRow row) {
    try {
      if (row == null) {
        return false;
      }

      final boolean accepted;
      if (!limit.isPresent()) {
        accepted = doAcceptRow(row);
      } else {
        synchronized (this) {
          if (totalRowsQueued.get() >= limit.getAsInt()) {
            closeInternal(true);
            return false;
          }
          accepted = doAcceptRow(row);
        }
      }

      if (accepted) {
        queuedCallback.run();
      }
      return accepted;
    } catch (final InterruptedException e) {
      // Forced shutdown?
      LOG.error("Interrupted while trying to offer row to queue", e);
    }
    return false;
  }

  private boolean doAcceptRow(final PullQueryRow row) throws InterruptedException {
    while (!closed.get()) {
      if (rowQueue.offer(row, offerTimeoutMs, TimeUnit.MILLISECONDS)) {
        totalRowsQueued.incrementAndGet();
        return true;
      }
    }
    return false;
  }
```

WDYT?

cprasad1:
yes, @Gerrrr I think that's a better solution. changed accordinbgly

Gerrrr (approved):

vvcephei (approved):
Thanks @cprasad1 and @Gerrrr !

PR #8566, opened 2022-01-07
fix: make writes to the backup file atomic

rodesai:
This patch makes writes to the backup file atomic. Writing a record now will either
complete in full, or not at all. This is achieved by writing the record to a dirty
copy, and then renaming the dirty copy to the live backup file. To maintain the dirty
copy without creating a new copy on each write, we create a temporary hardlink to
the live copy before renaming the dirty copy, and then rename the link to the dirty
copy. Finally, the write appends the record to the dirty copy (which was the old
live copy) to keep it in sync

### DIFF
--- a/ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/BackupReplayFile.java
+++ b/ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/BackupReplayFile.java
@@ -31,11 +31,15 @@ import java.util.ArrayList;
 import java.util.List;
 import java.util.Objects;
 import org.apache.kafka.clients.consumer.ConsumerRecord;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 /**
  * A file that is used by the backup service to replay command_topic commands.
  */
 public final class BackupReplayFile implements Closeable {
+  private static final Logger LOGGER = LoggerFactory.getLogger(BackupReplayFile.class);
+
   private static final String KEY_VALUE_SEPARATOR_STR = ":";
   private static final String NEW_LINE_SEPARATOR_STR = "\n";
   private static final String TMP_SUFFIX = ".tmp";
@@ -68,7 +72,7 @@ public final class BackupReplayFile implements Closeable {
     this.filesystem = Objects.requireNonNull(filesystem, "filesystem");
     this.writable = write;
     if (write) {
-      initDirtyCopy();
+      initFiles();
     }
   }
 
@@ -143,13 +147,14 @@ public final class BackupReplayFile implements Closeable {
   public void close() {
   }
 
-  private void initDirtyCopy() throws IOException {
+  private void initFiles() throws IOException {
     final Path dirtyFile = dirty(file).toPath();
     Files.deleteIfExists(dirtyFile);
     Files.deleteIfExists(tmp(file).toPath());
-    if (file.exists()) {
-      Files.copy(file.toPath(), dirty(file).toPath(), StandardCopyOption.REPLACE_EXISTING);
+    if (file.createNewFile()) {
+      LOGGER.info("created new backup replay file {}", file.getAbsolutePath());
     }
+    Files.copy(file.toPath(), dirty(file).toPath(), StandardCopyOption.REPLACE_EXISTING);
   }
 
   private File dirty(final File file) {
--- a/ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/BackupReplayFile.java
+++ b/ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/BackupReplayFile.java
@@ -98,12 +98,12 @@ public final class BackupReplayFile implements Closeable {
       final File file,
       final Filesystem filesystem
   ) throws IOException {
-    final FileOutputStream writer = createWriter(file, filesystem);
-    writer.write(record.key());
-    writer.write(KEY_VALUE_SEPARATOR_BYTES);
-    writer.write(record.value());
-    writer.write(NEW_LINE_SEPARATOR_BYTES);
-    writer.close();
+    try (final FileOutputStream writer = createWriter(file, filesystem)) {
+      writer.write(record.key());
+      writer.write(KEY_VALUE_SEPARATOR_BYTES);
+      writer.write(record.value());
+      writer.write(NEW_LINE_SEPARATOR_BYTES);
+    }
   }
 
   public void write(final ConsumerRecord<byte[], byte[]> record) throws IOException {
--- a/ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/BackupReplayFile.java
+++ b/ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/BackupReplayFile.java
@@ -24,7 +24,6 @@ import java.io.FileNotFoundException;
 import java.io.FileOutputStream;
 import java.io.IOException;
 import java.nio.charset.StandardCharsets;
-import java.nio.file.CopyOption;
 import java.nio.file.Files;
 import java.nio.file.Path;
 import java.nio.file.StandardCopyOption;
@@ -167,7 +166,7 @@ public final class BackupReplayFile implements Closeable {
 
   @VisibleForTesting
   interface Filesystem {
-    FileOutputStream outputStream(final File file, final boolean append)
+    FileOutputStream outputStream(File file, boolean append)
         throws FileNotFoundException;
   }
 }
--- a/ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/BackupReplayFile.java
+++ b/ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/BackupReplayFile.java
@@ -15,6 +15,7 @@
 
 package io.confluent.ksql.rest.server;
 
+import com.google.common.annotations.VisibleForTesting;
 import io.confluent.ksql.util.KsqlException;
 import io.confluent.ksql.util.Pair;
 import java.io.Closeable;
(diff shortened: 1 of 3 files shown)


ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/BackupReplayFile.java
```diff
@@ -35,36 +39,43 @@
...
+  private static final String TMP_SUFFIX = ".tmp";
+  private static final String DIRTY_SUFFIX = ".dirty";
 
   private static final byte[] KEY_VALUE_SEPARATOR_BYTES =
       KEY_VALUE_SEPARATOR_STR.getBytes(StandardCharsets.UTF_8);
   private static final byte[] NEW_LINE_SEPARATOR_BYTES =
       NEW_LINE_SEPARATOR_STR.getBytes(StandardCharsets.UTF_8);
 
   private final File file;
-  private final FileOutputStream writer;
```
rodesai, line 45: we now open/close a new writer for each append

ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/BackupReplayFile.java
```diff
@@ -35,36 +39,43 @@
...
+  private final boolean writable;
+  private final Filesystem filesystem;
```
rodesai, line 52: just a shim over filesystem calls for testing

ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/BackupReplayFile.java
```diff
@@ -79,16 +89,39 @@ public String getPath() {
...
-    writer.flush();
+    writer.close();
+  }
+
+  public void write(final ConsumerRecord<byte[], byte[]> record) throws IOException {
+    if (!writable) {
+      throw new IOException("Write permission denied.");
+    }
+    final File dirty = dirty(file);
+    final File tmp = tmp(file);
```
lct45, line 110: is the `tmp` file what you call the `live` file in the PR description? or is `tmp` just temporary file and the `file` we pass in is the `live` file

ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/BackupReplayFile.java
```diff
@@ -79,16 +89,39 @@ public String getPath() {
...
+    Files.createLink(tmp.toPath(), file.toPath());
+    Files.move(
+        dirty.toPath(),
+        file.toPath(),
+        StandardCopyOption.REPLACE_EXISTING,
+        StandardCopyOption.ATOMIC_MOVE
+    );
+    Files.move(tmp.toPath(), dirty.toPath());
+    // keep the dirty copy in sync with the live copy, which now has the write
+    appendRecordToFile(record, dirty, filesystem);
```
lct45, line 124: and this is the new dirty copy, right? Which was previously the live copy? It's a bit confusing since L124 and L 112 are the exact same

ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/BackupReplayFile.java
```diff
@@ -79,16 +89,39 @@ public String getPath() {
...
+    // atomically rename the dirty copy to the "live" copy while copying the live copy to
+    // the "dirty" copy via a temporary hard link
+    Files.createLink(tmp.toPath(), file.toPath());
+    Files.move(
+        dirty.toPath(),
+        file.toPath(),
+        StandardCopyOption.REPLACE_EXISTING,
+        StandardCopyOption.ATOMIC_MOVE
+    );
+    Files.move(tmp.toPath(), dirty.toPath());
```
lct45, line 122: After this, will we'll just have the dirty and live copies, right?

ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/BackupReplayFile.java
```diff
@@ -107,9 +140,33 @@ public void write(final ConsumerRecord<byte[], byte[]> record) throws IOExceptio
...
-    if (writer != null) {
-      writer.close();
+  public void close() {
+  }
+
+  private void initDirtyCopy() throws IOException {
+    final Path dirtyFile = dirty(file).toPath();
+    Files.deleteIfExists(dirtyFile);
+    Files.deleteIfExists(tmp(file).toPath());
+    if (file.exists()) {
```
lct45, line 150: Would we want to throw here if the file doesn't actually exist or is it okay if it doesn't exist? It doesn't seem like we'd note that we didn't do anything here

ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/CommandTopicBackupImpl.java
```diff
@@ -89,7 +89,11 @@ public void initialize() {
       LOG.warn("Failed to read the latest backup from {}. Continue with a new file. Error = {}",
           replayFile.getPath(), e.getMessage());
 
-      replayFile = newReplayFile();
+      try {
+        replayFile = newReplayFile();
+      } catch (final IOException ee) {
+        throw new RuntimeException(ee);
```
lct45, line 95: Is there any concern with adding new RuntimeExceptions?

ksqldb-rest-app/src/test/java/io/confluent/ksql/rest/server/BackupReplayFileTest.java
```diff
@@ -80,6 +94,62 @@ public void shouldWriteRecord() throws IOException {
...
+    // Given
+    final ConsumerRecord<byte[], byte[]> record1= newStreamRecord("stream1");
+    final ConsumerRecord<byte[], byte[]> record2 = newStreamRecord("stream2");
+
+    // When
+    replayFile.write(record1);
+    replayFile.write(record2);
+
+    // Then
+    final List<String> commands = Files.readAllLines(internalReplayFile.toPath());
```
lct45, line 108: Is there a way to confirm that the backup/dirty is also up to date?

ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/BackupReplayFile.java
rodesai: Not sure what you mean by the file we pass in. The live file is at the path stored in `file` - it's the file we're trying to update.

ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/BackupReplayFile.java
rodesai: Yep. After this, the "live" copy will have the update, and the "dirty" copy will not have the update, which is why we need to apply the update on line 124.

ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/CommandTopicBackupImpl.java
rodesai: Nah, the upper layers need to be able to deal with runtime exceptions - and they do by throwing to the top and exiting the process.

ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/BackupReplayFile.java
rodesai: It's ok for it not to exist - that's the case for new clusters. That said, I think we need to do some initialization in that case - not sure how this worked.

ksqldb-rest-app/src/test/java/io/confluent/ksql/rest/server/BackupReplayFileTest.java
rodesai: I suppose we could, but i'd rather not assert on the internal details of the implementation. This is basically what i was trying to capture with a test case that does multiple writes. If the backup were not updated, then we would not see the second write succeed.

ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/BackupReplayFile.java
lct45: @rodesai just confirming - is that initialization something we need to add or is this good to go?

lct45 (approved):
LGTM pending confirmation that we don't need initialization for new clusters where the file doesn't exist

ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/BackupReplayFile.java
guozhangwang: > That said, I think we need to do some initialization in that case - not sure how this worked.

Not sure I understand this sentence here: if it's a new cluster, then it's okay to have all three: live copy, dirty copy, and tmp, to be empty at initialization, since when the first append happens we would lazily created those files then, right?

guozhangwang (approved):
LGTM!

rodesai:
> Not sure I understand this sentence here: if it's a new cluster, then it's okay to have all three: live copy, dirty copy, and tmp, to be empty at initialization, since when the first append happens we would lazily created those files then, right

Currently append assumes that the live and dirty copy exist - so the init method creates them if they don't.

guozhangwang (approved):
Got it, thanks! I reviewed the latest commit on init and it LGTM.

rodesai:
> Why we need to close streams instance explicitly in this test?

Yeah actually now that I think about it more there's a real bug here in the write implementation where we don't close the handle on error. Fixing.

guozhangwang (approved):
Made another pass on the recent commits again, LGTM!

PR #8568, opened 2022-01-07
chore: expose original principal from DefaultKsqlPrincipal (MINOR)

vcrfxia:
### Description 

Minor change to expose the original Principal wrapped inside DefaultKsqlPrincipal, so that custom extensions can access the original principal even after wrapping. This PR also updates the existing code to always wrap the original principal in DefaultKsqlPrincipal, and have DefaultKsqlPrincipal pass through the custom properties of the original principal, even if the original principal is a KsqlPrincipal. This change is made because otherwise we'd have two patterns for custom extensions to access original principals:
* have the original principal implement KsqlPrincipal (comes at no cost since KsqlPrincipal has a default implementation for the only method beyond the Principal interface that it extends), which causes the original principal to be passed through (throughout the ksql engine) directly, or
* do not implement KsqlPrincipal, which means the original principal will be wrapped in DefaultKsqlPrincipal. Then access the original principal from the new `getOriginalPrincipal()` method added in this PR.

To avoid confusion from having multiple patterns, this PR removes the first one by always wrapping the incoming Principal in a DefaultKsqlPrincipal.

This change needs to be backported to 7.1.x since that's where DefaultKsqlPrincipal was first introduced.

### Testing done 

Unit tests added.

### Reviewer checklist

### DIFF
--- a/ksqldb-engine/src/main/java/io/confluent/ksql/security/DefaultKsqlPrincipal.java
+++ b/ksqldb-engine/src/main/java/io/confluent/ksql/security/DefaultKsqlPrincipal.java
@@ -37,4 +37,8 @@ public class DefaultKsqlPrincipal implements KsqlPrincipal {
   public Map<String, Object> getUserProperties() {
     return Collections.emptyMap();
   }
+
+  public Principal getOriginalPrincipal() {
+    return principal;
+  }
 }
--- a/ksqldb-engine/src/main/java/io/confluent/ksql/security/DefaultKsqlPrincipal.java
+++ b/ksqldb-engine/src/main/java/io/confluent/ksql/security/DefaultKsqlPrincipal.java
@@ -20,6 +20,11 @@ import java.util.Collections;
 import java.util.Map;
 import java.util.Objects;
 
+/**
+ * A {@link KsqlPrincipal} implementation that wraps another {@code Principal}.
+ * If the wrapped principal is a KsqlPrincipal, then its user properties are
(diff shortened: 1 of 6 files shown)


ksqldb-engine/src/main/java/io/confluent/ksql/security/DefaultKsqlPrincipal.java
```diff
@@ -37,4 +37,8 @@ public String getName() {
   public Map<String, Object> getUserProperties() {
     return Collections.emptyMap();
   }
+
+  public Principal getOriginalPrincipal() {
```
spena, line 41: Could you add a comment about where this method is used? Just to make sure is never deleted or public id removed accidentally.

spena (approved):

spena:
@vcrfxia Seems we'll need to have only one pattern. Now that ksql wraps the JwtPrincipal, then custom plugins will never get a class with it, only KsqlPrincipal.

ksqldb-engine/src/main/java/io/confluent/ksql/security/DefaultKsqlPrincipal.java
vcrfxia: Done.

vcrfxia (commented):
All right, I updated the code to always wrap the incoming Principal in a DefaultKsqlPrincipal, even if the incoming Principal is already a KsqlPrincipal. This cleans up the usage so there's only one access pattern for the original principal.

PR #8579, opened 2022-01-11
fix: use topology-specific IQ APIs

ableegoldman:
Found two more instances of IQ APIs that need to be updated to fetch metadata for specific queries within the shared runtime:

1. Instead of returning `kafkaStreams.metadataForAllClients()`, the bin-packed persistent queries' `allMetadata` method should return only the query-specific results from (new API) `KafkaStreamsNamedTopologyWrapper#streamsMetadataForTopology`
2. We seem to have lost the `.partition()` specification when constructing the named topology version of the `StoreQueryParameters`, we needed to add this to the  `NamedTopologyStoreQueryParameter` class on the Streams side and then invoke that in ksql

Blocked on https://github.com/apache/kafka/pull/11609

### DIFF
--- a/ksqldb-engine/src/main/java/io/confluent/ksql/util/SharedKafkaStreamsRuntimeImpl.java
+++ b/ksqldb-engine/src/main/java/io/confluent/ksql/util/SharedKafkaStreamsRuntimeImpl.java
@@ -65,7 +65,9 @@ public class SharedKafkaStreamsRuntimeImpl extends SharedKafkaStreamsRuntime {
       final QueryId queryId
   ) {
     collocatedQueries.put(queryId, binpackedPersistentQueryMetadata);
-    log.info("mapping {}", collocatedQueries);
+    log.info("Registered query in shared runtime: {}\n"
+                 + "Runtime {} is executing these queries: {}",
+             queryId, getApplicationId(), collocatedQueries.keySet());
   }
 
   private void setupAndStartKafkaStreams(final KafkaStreams kafkaStreams) {
@@ -164,7 +166,8 @@ public class SharedKafkaStreamsRuntimeImpl extends SharedKafkaStreamsRuntime {
 
   @Override
   public void stop(final QueryId queryId, final boolean resetOffsets) {
-    log.info("Attempting to stop Query: " + queryId.toString());
+    log.info("Attempting to stop query: {} in runtime {} with resetOffsets={}",
+             queryId, getApplicationId(), resetOffsets);
     if (collocatedQueries.containsKey(queryId)) {
       if (kafkaStreams.state().isRunningOrRebalancing()) {
         try {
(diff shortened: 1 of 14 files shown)


wcarlson5 (approved):
LGTM once you get a green run

ksqldb-engine/src/main/java/io/confluent/ksql/util/SharedKafkaStreamsRuntimeImpl.java
```diff
@@ -65,7 +65,9 @@ public void register(
       final QueryId queryId
   ) {
     collocatedQueries.put(queryId, binpackedPersistentQueryMetadata);
-    log.info("mapping {}", collocatedQueries);
+    log.info("Registered query in shared runtime: {}\n"
```
guozhangwang, line 68: Thanks for the improved logging! This is great. And ditto below :)

ksqldb-streams/src/main/java/io/confluent/ksql/execution/streams/materialization/ks/KsStateStore.java
```diff
@@ -61,28 +61,22 @@ LogicalSchema schema() {
...
-        }
-        return ((KafkaStreamsNamedTopologyWrapper) kafkaStreams).store(namedTopologyParameters);
-      } else if (ksqlConfig.getBoolean(KsqlConfig.KSQL_QUERY_PULL_ENABLE_STANDBY_READS)) {
-        // True flag allows queries on standby and replica state stores
-        return kafkaStreams.store(parameters.enableStaleStores());
-      } else {
-        // False flag allows queries only on active state store
-        return kafkaStreams.store(parameters);
-      }
+      final StoreQueryParameters<T> parameters = sharedRuntime
```
guozhangwang, line 68: Nice cleanup,    and fix!

guozhangwang (approved):
Please feel free to merge after the blocking AK one is done.

guozhangwang:
I made a pass on the latest commit, I think removing the internal reference is the right thing to do but it would indeed break tests, do you know if a follow-up PR is on its way by the database team (I think that would block this PR from being merged)?

jnh5y:
I have opened https://github.com/confluentinc/ksql/pull/8674 as a rebase of this PR.

ksqldb-streams/src/main/java/io/confluent/ksql/execution/streams/materialization/ks/KsStateStore.java
```diff
@@ -73,28 +73,22 @@ KsqlConfig getKsqlConfig() {
 
   <T> T store(final QueryableStoreType<T> queryableStoreType, final int partition) {
     try {
-      final StoreQueryParameters<T> parameters = StoreQueryParameters.fromNameAndType(
-          stateStoreName, queryableStoreType).withPartition(partition);
-      if (ksqlConfig.getBoolean(KsqlConfig.KSQL_SHARED_RUNTIME_ENABLED)
```
jnh5y, line 78: Do we still need to check the value of `ksqlConfig.getBoolean(KsqlConfig.KSQL_SHARED_RUNTIME_ENABLED`?

wcarlson5: yes

wcarlson5: no we just check if we are using a Named topology related object

jnh5y: What would have happened previously if someone had set `KSQL_SHARED_RUNTIME_ENABLED` to false?  (It seems that it would have prevented using a NamedTopology even if one were configured.

There is a logic change here.  Is it one we want?  (Sorry to ask twice; just trying to be sure about things.)

wcarlson5: That was the issue. We want this method to work with what ever store it is give regardless of `KSQL_SHARED_RUNTIME_ENABLED`

wcarlson5:
rebased here https://github.com/confluentinc/ksql/pull/8675

PR #8585, opened 2022-01-12
feat: migrate java client to use application/vnd.ksql.v1+json

cprasad1:
### Description 
Migrate the java client to use `application/vnd.ksql.v1+json` format when hitting the `/query-stream` endpoint without breaking backward compatibility

### Testing done 

### Reviewer checklist

### DIFF
--- a/ksqldb-api-client/src/main/java/io/confluent/ksql/api/client/impl/ExecuteQueryResponseHandler.java
+++ b/ksqldb-api-client/src/main/java/io/confluent/ksql/api/client/impl/ExecuteQueryResponseHandler.java
@@ -67,32 +67,35 @@ public class ExecuteQueryResponseHandler extends QueryResponseHandler<BatchedQue
 
   @Override
   protected void handleRow(final Buffer buff) {
-    final Row row;
-    final Object json = buff.toJson();
+    final JsonObject json = buff.toJsonObject();
 
-    if (json instanceof JsonObject) {
-      final JsonObject jsonObject = (JsonObject) json;
-      // This is the serialized consistency vector
-      // Don't add it to the result list since the user should not see it
-      if (jsonObject.getMap() != null && jsonObject.getMap().containsKey("consistencyToken")) {
-        log.info("Response contains consistency vector " + jsonObject);
-        serializedConsistencyVector.set((String) ((JsonObject) json).getMap().get(
-            "consistencyToken"));
+    if (!json.containsKey("finalMessage")) {
+      if (json.containsKey("row")) {
+        if (rows.size() < maxRows) {
+          rows.add(new RowImpl(
+                  columnNames,
+                  columnTypes,
+                  new JsonArray((List)((Map<?, ?>) json.getMap().get("row")).get("columns")),
+                  columnNameToIndex));
+        } else {
+          throw new KsqlClientException(
+                  "Reached max number of rows that may be returned by executeQuery(). "
+                          + "Increase the limit via ClientOptions#setExecuteQueryMaxResultRows(). "
+                          + "Current limit: " + maxRows);
+        }
+      } else if (json.getMap() != null) {
+        // This is the serialized consistency vector
+        // Don't add it to the result list since the user should not see it
+        if (json.getMap().containsKey("consistencyToken")) {
+          log.info("Response contains consistency vector " + json);
+          serializedConsistencyVector.set((String) json.getMap().get(
+              "consistencyToken"));
+        } else {
+          throw new RuntimeException("Could not decode JSON, expected consistency token: " + json);
+        }
       } else {
-        throw new RuntimeException("Could not decode JSON, expected consistency toke: " + json);
+        throw new RuntimeException("Could not decode JSON: " + json);
       }
-    } else  if (json instanceof JsonArray) {
-      final JsonArray values = new JsonArray(buff);
-      if (rows.size() < maxRows) {
-        rows.add(new RowImpl(columnNames, columnTypes, values, columnNameToIndex));
-      } else {
-        throw new KsqlClientException(
-            "Reached max number of rows that may be returned by executeQuery(). "
-                + "Increase the limit via ClientOptions#setExecuteQueryMaxResultRows(). "
-                + "Current limit: " + maxRows);
-      }
-    } else {
-      throw new RuntimeException("Could not decode JSON: " + json);
     }
   }
 
--- a/ksqldb-api-client/src/main/java/io/confluent/ksql/api/client/impl/QueryResponseHandler.java
+++ b/ksqldb-api-client/src/main/java/io/confluent/ksql/api/client/impl/QueryResponseHandler.java
@@ -15,12 +15,17 @@
 
 package io.confluent.ksql.api.client.impl;
 
+import static io.confluent.ksql.util.BytesUtils.toJsonMsg;
+
 import com.fasterxml.jackson.databind.ObjectMapper;
 import io.confluent.ksql.api.client.util.JsonMapper;
+import io.confluent.ksql.api.client.util.RowUtil;
 import io.confluent.ksql.rest.entity.QueryResponseMetadata;
 import io.vertx.core.Context;
 import io.vertx.core.buffer.Buffer;
 import io.vertx.core.parsetools.RecordParser;
+import java.util.Map;
+import java.util.Objects;
 import java.util.concurrent.CompletableFuture;
 
 abstract class QueryResponseHandler<T extends CompletableFuture<?>> extends ResponseHandler<T> {
@@ -35,10 +40,14 @@ abstract class QueryResponseHandler<T extends CompletableFuture<?>> extends Resp
 
   @Override
   protected void doHandleBodyBuffer(final Buffer buff) {
+    final Buffer strippedValidJson = toJsonMsg(buff, true);
+
     if (!hasReadArguments) {
-      handleArgs(buff);
+      handleArgs(strippedValidJson);
     } else {
-      handleRow(buff);
+      if (!Objects.equals(strippedValidJson.toString(), "")) {
+        handleRow(strippedValidJson);
+      }
     }
   }
 
@@ -59,10 +68,16 @@ abstract class QueryResponseHandler<T extends CompletableFuture<?>> extends Resp
 
   private void handleArgs(final Buffer buff) {
     hasReadArguments = true;
-
     final QueryResponseMetadata queryResponseMetadata;
     try {
-      queryResponseMetadata = JSON_MAPPER.readValue(buff.getBytes(), QueryResponseMetadata.class);
+      final Object header = buff.toJsonObject().getMap().get("header");
+      final String queryId = (String) ((Map<?, ?>) header).get("queryId");
+      final String schema = (String) ((Map<?, ?>) header).get("schema");
+      queryResponseMetadata =
+              new QueryResponseMetadata(queryId,
+                      RowUtil.colNamesFromSchema(schema),
+                      RowUtil.colTypesFromSchema(schema),
+                      null);
     } catch (Exception e) {
       cf.completeExceptionally(e);
       return;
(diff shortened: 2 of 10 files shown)


ksqldb-api-client/src/main/java/io/confluent/ksql/api/client/impl/ExecuteQueryResponseHandler.java
```diff
@@ -58,31 +62,50 @@
...
-        serializedConsistencyVector.set((String) ((JsonObject) json).getMap().get(
-            "consistencyToken"));
-      } else {
-        throw new RuntimeException("Could not decode JSON, expected consistency toke: " + json);
+      final ConsistencyToken consistencyToken = row.getConsistencyToken().get();
+      try {
+        final String consistencyTokenString = consistencyToken.getConsistencyToken();
+        log.info("Response contains consistency vector " + consistencyTokenString);
+        serializedConsistencyVector.set(consistencyTokenString);
+      } catch (Exception e) {
```
AlanConfluent, line 101: Do you need this catch here?  There is no parsing in particular that's happening and no known exception to catch.  I think the error before was that if there was a JsonObject, it must be a consistency token, but using `StreamedRow` takes care of much of the decisioning for you here.

ksqldb-api-client/src/main/java/io/confluent/ksql/api/client/impl/StreamQueryResponseHandler.java
```diff
@@ -49,52 +54,72 @@
...
+
+    if (row.getConsistencyToken().isPresent()) {
+      // This is the serialized consistency vector
+      // Don't add it to the publisher's buffer since the user should not see it
+      final ConsistencyToken consistencyToken = row.getConsistencyToken().get();
+      try {
+        final String consistencyTokenString = consistencyToken.getConsistencyToken();
+        LOG.info("Response contains consistency vector " + consistencyTokenString);
+        serializedConsistencyVector.set(consistencyTokenString);
+      } catch (Exception e) {
```
AlanConfluent, line 101: Same here

ksqldb-api-client/src/main/java/io/confluent/ksql/api/client/impl/QueryResponseHandler.java
```diff
@@ -51,23 +76,46 @@ protected void doHandleException(final Throwable t) {
...
+
     } catch (Exception e) {
       cf.completeExceptionally(e);
       return;
     }
 
     handleMetadata(queryResponseMetadata);
   }
+
+  public static Buffer toValidJsonMsg(final Buffer responseLine, final boolean stripArray) {
```
AlanConfluent, line 103: Not a big deal, but this is copied logic.  Might makes sense to put in common if you think we'll modify this much.

ksqldb-api-client/src/main/java/io/confluent/ksql/api/client/impl/ExecuteQueryResponseHandler.java
```diff
@@ -58,31 +62,50 @@
...
-  protected void handleMetadata(final QueryResponseMetadata queryResponseMetadata) {
-    cf.queryID().complete(queryResponseMetadata.queryId);
-    columnNames = queryResponseMetadata.columnNames;
-    columnTypes = RowUtil.columnTypesFromStrings(queryResponseMetadata.columnTypes);
+  protected void handleMetadata(final StreamedRow queryResponseMetadata) {
+    final String queryId = queryResponseMetadata.getHeader().get().getQueryId().toString();
+
+    cf.queryID().complete(queryId);
+
+    final List<Column> columns = queryResponseMetadata
```
AlanConfluent, line 70: Share this logic with the class below?  It's pretty simple, but still might make sense to push into the parent class.

ksqldb-api-client/src/main/java/io/confluent/ksql/api/client/impl/ExecuteQueryResponseHandler.java
cprasad1: yes, you are right. removed `try...catch` block

ksqldb-api-client/src/main/java/io/confluent/ksql/api/client/impl/StreamQueryResponseHandler.java
cprasad1: yes, you are right. removed try...catch block

ksqldb-api-client/src/main/java/io/confluent/ksql/api/client/impl/QueryResponseHandler.java
cprasad1: done

ksqldb-api-client/pom.xml
```diff
@@ -158,6 +158,12 @@
             <version>${jetty.version}</version>
             <scope>test</scope>
         </dependency>
+        <dependency>
+            <groupId>com.google.code.gson</groupId>
+            <artifactId>gson</artifactId>
+            <version>2.8.6</version>
+            <scope>compile</scope>
+        </dependency>
```
cprasad1, line 166: trying to figure out a replacement for this. Doing  `final JsonArray values = new JsonArray(dataRow.getColumns())` instead of using gson leads to `Illegal type in JsonObject: class java.math.BigDecimal`

AlanConfluent: Before your change, when we were using the Vertx json directly, what type does it use for a BigDecimal equivalent?

ksqldb-api-client/src/main/java/io/confluent/ksql/api/client/impl/ExecuteQueryResponseHandler.java
```diff
@@ -59,30 +62,32 @@
...
-    } else  if (json instanceof JsonArray) {
-      final JsonArray values = new JsonArray(buff);
+      final ConsistencyToken consistencyToken = row.getConsistencyToken().get();
+      final String consistencyTokenString = consistencyToken.getConsistencyToken();
+      log.info("Response contains consistency vector " + consistencyTokenString);
+      serializedConsistencyVector.set(consistencyTokenString);
+    } else  if (row.getRow().isPresent()) {
+      final DataRow dataRow = row.getRow().get();
+      final Gson gson = new GsonBuilder().serializeNulls().create();
+      final JsonArray values = new JsonArray(gson.toJson(dataRow.getColumns()));
```
AlanConfluent, line 90: We parse using Jackson into StreamedRow only to stringify it again, and then reparse it into Vertex json object?  That seems like it should be avoidable.   

We should either switch to using pojos everywhere rather than the Vertex json, in which case there's no need for this double parsing, or if it's more trouble than it's worth, skip parsing into the StreamedRow object with jackson. You would still request `application/vnd.ksql.v1+json`, but do `final Object json = buff.toJson();` as before and ask for fields by name.  Instead of it being the delimited format, it's the `application/vnd.ksql.v1+json` format, so the code is similar, but just is looking for slightly different fields.  This would still accomplish the main goal of just moving to the common format.  Tell me what you think.

cprasad1: using `readTree` now instead to extract the fields

ksqldb-api-client/src/main/java/io/confluent/ksql/api/client/impl/ExecuteQueryResponseHandler.java
```diff
@@ -59,30 +58,34 @@
...
-        serializedConsistencyVector.set((String) ((JsonObject) json).getMap().get(
-            "consistencyToken"));
-      } else {
-        throw new RuntimeException("Could not decode JSON, expected consistency toke: " + json);
-      }
-    } else  if (json instanceof JsonArray) {
-      final JsonArray values = new JsonArray(buff);
+
+      final String consistencyTokenString = rowNode
+              .get(CONSISTENCY_TOKEN_KEY)
```
AlanConfluent, line 81: Extra get?

ksqldb-api-client/src/main/java/io/confluent/ksql/api/client/util/RowUtil.java
```diff
@@ -50,4 +51,69 @@ private static ColumnType columnTypeFromString(final String columnType) {
...
+    }
+    return listCols;
+  }
+
+  private static String removeBackTickAndKeyTrim(final String dirtyString) {
+    return dirtyString.replace(" KEY", "").replace("`", "").trim();
+  }
+
+  private static String splitAndGetFirst(final String stringToSplit) {
+    return stringToSplit.split(" ", 2)[0];
```
AlanConfluent, line 113: Maybe this is always going to work, but you could make this more of a generic whitespace matcher rather than a single space.  Anyway, not a huge deal so long as we have unit tests that are going to catch any changes.

ksqldb-api-client/src/main/java/io/confluent/ksql/api/client/impl/QueryResponseHandler.java
```diff
@@ -35,10 +52,32 @@
...
+        rowNode = JSON_MAPPER.readTree(strippedValidJson.getBytes());
+      } catch (Exception e) {
+        cf.completeExceptionally(e);
+        return;
+      }
+
+      if (rowNode.has(FINAL_MESSAGE_KEY)) {
+        doHandleBodyEnd();
+      } else if (rowNode.has(ERROR_MESSAGE_KEY)) {
+        final KsqlException exception = new KsqlException(
```
AlanConfluent, line 75: Before, I believe this would be handled by calling the following rather than failing the request:

```
queryResult.handleError(new KsqlException(
            jsonObject.getString("message")
        ));
```

We should try to keep consistent behavior there.

ksqldb-api-client/src/main/java/io/confluent/ksql/api/client/impl/QueryResponseHandler.java
```diff
@@ -35,10 +52,32 @@
...
+      final JsonNode rowNode;
+      try {
+        rowNode = JSON_MAPPER.readTree(strippedValidJson.getBytes());
+      } catch (Exception e) {
+        cf.completeExceptionally(e);
+        return;
+      }
+
+      if (rowNode.has(FINAL_MESSAGE_KEY)) {
+        doHandleBodyEnd();
```
AlanConfluent, line 73: This is now called twice.  I'm not sure that's an issue, but we may want to avoid that.  Maybe just log a message instead?

AlanConfluent (approved):

ksqldb-api-client/src/main/java/io/confluent/ksql/api/client/impl/QueryResponseHandler.java
```diff
@@ -35,10 +52,32 @@
...
+    }
+
     if (!hasReadArguments) {
-      handleArgs(buff);
+      handleArgs(strippedValidJson);
     } else {
-      handleRow(buff);
+      final JsonNode rowNode;
+      try {
+        rowNode = JSON_MAPPER.readTree(strippedValidJson.getBytes());
```
AlanConfluent, line 66: Also, another question.  Is there any reason we used Jackson vs the Vertx json it was previously using?  Just curious if the handling of BigDecimals was the reason or something like that.

ksqldb-api-client/src/main/java/io/confluent/ksql/api/client/impl/ExecuteQueryResponseHandler.java
cprasad1: no. the `ConsistencyToken` object has a `consistencyToken`. instance variable

ksqldb-api-client/src/main/java/io/confluent/ksql/api/client/util/RowUtil.java
cprasad1: let's keep it specific for now as it is highly dependent on our schema format today (and it is made for one and only one purpose)

ksqldb-api-client/src/main/java/io/confluent/ksql/api/client/impl/QueryResponseHandler.java
cprasad1: yeah, handling `BigDecimal` becomes much cleaner with jackson

ksqldb-api-client/src/main/java/io/confluent/ksql/api/client/impl/QueryResponseHandler.java
cprasad1: done

JimGalasyn (approved):

PR #8590, opened 2022-01-12
feat: Change API to return Position from Streams

vpapavas:
### Description 
Created a new return type `KsMaterializedQueryResult` for `table.get` methods that query the streams store to return, besides the KeyValue, also the position used for consistency. Previously, the methods returned a single object, a list, an iterator but all were converted to iterators. So I changed the return type to always be an iterator. 

Note, this is only an API change. The actual using of the position will be done in a separate PR

### Testing done 
Updated all unit tests for IQv1 and IQv2

### Reviewer checklist

### DIFF
--- a/ksqldb-streams/src/main/java/io/confluent/ksql/execution/streams/materialization/ks/KsMaterializedQueryResult.java
+++ b/ksqldb-streams/src/main/java/io/confluent/ksql/execution/streams/materialization/ks/KsMaterializedQueryResult.java
@@ -1,0 +1,69 @@
+/*
+ * Copyright 2019 Confluent Inc.
+ *
+ * Licensed under the Confluent Community License (the "License"; you may not use
+ * this file except in compliance with the License. You may obtain a copy of the
+ * License at
+ *
+ * http://www.confluent.io/confluent-community-license
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OF ANY KIND, either express or implied.  See the License for the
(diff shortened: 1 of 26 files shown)


ksqldb-streams/src/main/java/io/confluent/ksql/execution/streams/materialization/ks/KsMaterializedQueryResult.java
```diff
@@ -0,0 +1,54 @@
...
+
+  public static <T extends TableRow> KsMaterializedQueryResult<T> rowIteratorWithPosition(
+      final Iterator<T> iterator, final Position position) {
+    return new KsMaterializedQueryResult<>(iterator, Optional.of(position));
+  }
+
+  private KsMaterializedQueryResult(final Iterator<T> rowIterator,
+                                    final Optional<Position> position) {
+    this.rowIterator = Objects.requireNonNull(rowIterator, "rowIterator");
+    this.position = Objects.requireNonNull(position, "position");
```
patrickstuedi, line 44: Question: in what scenarios will position be optional/empty?

patrickstuedi (approved):
LGTM, thanks @vpapavas !

ksqldb-streams/src/main/java/io/confluent/ksql/execution/streams/materialization/ks/KsMaterializedQueryResult.java
vpapavas: It's optional because for IQv1 we don't use the position. When we decide to use only IQv2, we can remove the optional part

PR #8593, opened 2022-01-13
fix: Fixes an NPE which was causing terminate to fail

AlanConfluent:
### Description 
In tests, at times, a double call was being made to `QueryStateMetricsReportingListener.onDeregister` which would cause an NPE the second time.  This was causing terminations to fail, such as in `TerminateTransientQueryFunctionalTest`.

The two calls were caused by first the termination, which calls `TransientQueryMetadata.close` and issues a complete event to be send to the client, causing it to close the connection and end the request, which in turn caused a `TransientQueryMetadata.close` to be called again.  If each was done just right, the `onDeregister` call would complete in one case and then the second would fail.  Other times, they would each run, finding the query, and not get an NPE.

This should ensure that whether the query ends entirely by the client or as a result of a termination, the logic of `TransientQueryMetadata.close` is called just once.

### Testing done 
Ran `TerminateTransientQueryFunctionalTest` many times to verify fix.

### Reviewer checklist

### DIFF
--- a/ksqldb-engine/src/main/java/io/confluent/ksql/internal/QueryStateMetricsReportingListener.java
+++ b/ksqldb-engine/src/main/java/io/confluent/ksql/internal/QueryStateMetricsReportingListener.java
@@ -86,10 +86,8 @@ public class QueryStateMetricsReportingListener implements QueryEventListener {
 
   @Override
   public void onDeregister(final QueryMetadata query) {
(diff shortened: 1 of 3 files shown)


nateab (approved):
LGTM! thanks for the quick fix

cprasad1 (approved):
makes sense. thanks @AlanConfluent !

PR #8599, opened 2022-01-14
fix: back out post-3.1 changes to fix 7.1.x build

ableegoldman:
Remove/revert the changes to ksql that rely on Streams APIs which aren't present in the 7.1.0 branch of ccs-kafka that will ship with ksql 7.1


wcarlson5 (commented):
I can't think of anything else that needs to be removed but the build needs to pass still

guozhangwang:
Thanks for the quick turnout @ableegoldman !

PR #8607, opened 2022-01-18
feat: uses internal topics prefix for shared runtimes

wcarlson5:
Without shared runtimes: `_confluent-ksql-default_query_CSAS_CART_EVENT_PRODUCT_1-Join-repartition`
shared runtimes without change: `_confluent-ksql-default_query_-1-CSAS_CART_EVENT_PRODUCT_1-Join-repartition`
shared runtimes with change: `_confluent-ksql-default_query-CSAS_CART_EVENT_PRODUCT_1-Join-repartition`

Will remove the runtime Id from internal topics so that we will be able to use topic in a different application with later work

### Testing done 

### Reviewer checklist

### DIFF
--- a/ksqldb-engine/src/main/java/io/confluent/ksql/query/QueryBuilder.java
+++ b/ksqldb-engine/src/main/java/io/confluent/ksql/query/QueryBuilder.java
@@ -568,9 +568,9 @@ final class QueryBuilder {
 
     if (config.getBoolean(KsqlConfig.KSQL_SHARED_RUNTIME_ENABLED)) {
       newStreamsProperties.put(StreamsConfig.InternalConfig.TOPIC_PREFIX_ALTERNATIVE,
-          ReservedInternalTopics.KSQL_INTERNAL_TOPIC_PREFIX +
-              config.getString(KsqlConfig.KSQL_SERVICE_ID_CONFIG) +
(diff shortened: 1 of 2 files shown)


ableegoldman (approved):
LGTTM!

guozhangwang:
@wcarlson5 should we change the logic in `KafkaTopicClientImpl` too, where `applicationId` is used as the default prefix?

wcarlson5:
@guozhangwang do you mean when deleting internal topics?

guozhangwang:
> I think we do need to but it might require a bit more thought because we will need to change how that is done

Yes I was thinking about the purging of internal topics upon terminating the query indeed. Please let me know what procedure changes we'd need to consider.

wcarlson5:
@guozhangwang  Sophie reminded me I already made that change. I just need to adapt the new thing to use the new prefix instead of the application like you mentioned. I am going to leave `KafkaTopicClientImpl's logic` the same just call it with the prefix instead of app ID

guozhangwang:
Okay, thanks!

wcarlson5:
This caused an issue for running old queries on the same server as queries in the new runtime. The fix can be found here https://github.com/confluentinc/ksql/pull/9470

Issue #4800, opened 2020-03-17
REST API - fails if CS + CT combined (Failed to guarantee existence of topic / Replication factor must be larger than 0)


rmoff:
ksqlDB 0.7.1

Docker Compose: 
[docker-compose.txt](https://github.com/confluentinc/ksql/files/4344709/docker-compose.txt)

Bring the stack up

```
docker-compose up -d
```

Prepared statements to run: 

```
➜ cat src/statements.sql
CREATE STREAM clicks (ip VARCHAR, url VARCHAR, timestamp VARCHAR)
WITH (KAFKA_TOPIC='clicks',
      TIMESTAMP='timestamp',
      TIMESTAMP_FORMAT='yyyy-MM-dd''T''HH:mm:ssX',
      PARTITIONS=1,
      VALUE_FORMAT='Avro');

CREATE TABLE IP_SESSIONS AS
SELECT IP,
       TIMESTAMPTOSTRING(WINDOWSTART,'yyyy-MM-dd HH:mm:ss') AS SESSION_START_TS,
       TIMESTAMPTOSTRING(WINDOWEND,'yyyy-MM-dd HH:mm:ss')   AS SESSION_END_TS,
       COUNT(*)                                             AS CLICK_COUNT,
       WINDOWEND - WINDOWSTART                              AS SESSION_LENGTH_MS
  FROM CLICKS
       WINDOW SESSION (5 MINUTES)
GROUP BY IP;

```

🔴 Try to create a stream over the REST API :

```
➜ statements=$(tr '\n' ' ' < src/statements.sql) && \
    echo '{"ksql":"'$statements'", "streamsProperties": {}}' | \
        curl -X "POST" "http://localhost:8088/ksql" \
             -H "Content-Type: application/vnd.ksql.v1+json; charset=utf-8" \
             -d @- | \
        jq

  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100  8007    0  7331  100   676  74050   6828 --:--:-- --:--:-- --:--:-- 80878
{
  "@type": "statement_error",
  "error_code": 50000,
  "message": "Failed to guarantee existence of topic IP_SESSIONS\nCaused by: Replication factor must be larger than 0.",
  "stackTrace": [
    "io.confluent.ksql.services.KafkaTopicClientImpl.createTopic(KafkaTopicClientImpl.java:135)",
[...]

PR #8614, opened 2022-01-19
fix: register schema within sandbox

### LINKED ISSUE #1394: UX: CSAS returns control to UI before completing, subsequent statement fails with Unable to verify the AVRO schema is compatible with KSQL. Subject not found.
I want to be able to run a set of KSQL statements, either cut & paste as part of a demo, or as a script passed to KSQL. 

I've hit a problem with this: 

```
ksql> CREATE STREAM CUSTOMERS_SRC_REKEY WITH (PARTITIONS=1) AS SELECT * FROM CUSTOMERS_SRC PARTITION BY ID;

 Message
----------------------------
 Stream created and running
----------------------------
ksql> CREATE TABLE CUSTOMERS WITH (KAFKA_TOPIC='CUSTOMERS_SRC_REKEY', VALUE_FORMAT ='AVRO', KEY='ID');
 Unable to verify the AVRO schema is compatible with KSQL. Subject not found.; error code: 40401
ksql> CREATE TABLE CUSTOMERS WITH (KAFKA_TOPIC='CUSTOMERS_SRC_REKEY', VALUE_FORMAT ='AVRO', KEY='ID');
 Unable to verify the AVRO schema is compatible with KSQL. Subject not found.; error code: 40401
ksql> CREATE TABLE CUSTOMERS WITH (KAFKA_TOPIC='CUSTOMERS_SRC_REKEY', VALUE_FORMAT ='AVRO', KEY='ID');

 Message
---------------
 Table created
---------------
ksql>
```

What I think is happening is that the CSAS results in a new Avro schema, which is not registered in time before the subsequent statement (which requires it) is executed. The second statement has to be manually retried until it succeeds (a few seconds later). 

KSQL needs to either (a) wait until the first statement has fully executed or (b) before throwing the error in the second statement, check if there is some pending change that would allow it to complete without error.

mjsax:
Closes #1394


ksqldb-engine/src/main/java/io/confluent/ksql/services/SandboxedSchemaRegistryClient.java
```diff
@@ -36,21 +46,158 @@
 
   static SchemaRegistryClient createProxy(final SchemaRegistryClient delegate) {
     Objects.requireNonNull(delegate, "delegate");
+    final SandboxSchemaRegistryCache sandboxSchemaRegistryCache = new SandboxSchemaRegistryCache(delegate);
 
     return LimitedProxyBuilder.forClass(SchemaRegistryClient.class)
-        .swallow("register", anyParams(), 123)
+        .forward("register", methodParams(String.class, ParsedSchema.class), sandboxSchemaRegistryCache)
```
mjsax, line 52: This is the "main" fix -- instead of swallowing the registration, we capture it within the sandbox.

suhas-satish: how are we planning to test this?

suhas-satish: how are we planning to test this?

suhas-satish (approved):

ksqldb-engine/src/main/java/io/confluent/ksql/services/SandboxedSchemaRegistryClient.java
mjsax: Working on tests... The PR is still WIP. Just opened if for early feedback.

config/ksql-server.properties
```diff
@@ -78,4 +78,4 @@ compression.type=snappy
 #------ Schema Registry -------
 
 # Uncomment and complete the following to enable KSQL's integration to the Confluent Schema Registry:
-# ksql.schema.registry.url=http://localhost:8081
+ ksql.schema.registry.url=http://localhost:8081
```
mjsax, line 81:
```suggestion
 # ksql.schema.registry.url=http://localhost:8081
```

config/ksql-server.properties
```diff
@@ -78,4 +78,4 @@ compression.type=snappy
 #------ Schema Registry -------
 
 # Uncomment and complete the following to enable KSQL's integration to the Confluent Schema Registry:
-# ksql.schema.registry.url=http://localhost:8081
+ # ksql.schema.registry.url=http://localhost:8081
```
mjsax, line 81:
```suggestion
# ksql.schema.registry.url=http://localhost:8081
```

ksqldb-engine/src/main/java/io/confluent/ksql/services/SandboxedSchemaRegistryClient.java
```diff
@@ -36,21 +43,193 @@
...
+      return schemaId;
+    }
+
+    @Override
+    public int register(
+        final String subject,
+        final ParsedSchema parsedSchema,
+        final int version,
+        final int id) {
+      return -1; // swallow
```
lihaosky, line 90: Why is this swallowed whereas previous `register` does actual work?

ksqldb-engine/src/main/java/io/confluent/ksql/services/SandboxedSchemaRegistryClient.java
```diff
@@ -36,21 +43,193 @@
...
+        final String schemaType,
+        final String schemaString,
+        final List<SchemaReference> references) {
+      throw new UnsupportedOperationException();
+    }
+
+    @Override
+    public int register(final String subject, final ParsedSchema parsedSchema) {
+      if (subjectCache.containsKey(subject)) {
+        throw new IllegalStateException("Subject '" + subject + "' already in use.");
```
lihaosky, line 75: Why does this throw instead of returning existing id for the subject which is the behavior of schema registry?

ksqldb-engine/src/main/java/io/confluent/ksql/services/SandboxedSchemaRegistryClient.java
```diff
@@ -36,21 +43,193 @@
...
+      }
+      final int schemaId = nextSchemaId--;
+      subjectCache.put(subject, parsedSchema);
+      subjectToId.put(subject, schemaId);
+      idCache.put(schemaId, parsedSchema);
```
lihaosky, line 80: Could you use `MockSchemaRegistryClient`? https://github.com/confluentinc/schema-registry/blob/master/client/src/main/java/io/confluent/kafka/schemaregistry/client/MockSchemaRegistryClient.java

ksqldb-engine/src/main/java/io/confluent/ksql/services/SandboxedSchemaRegistryClient.java
```diff
@@ -36,21 +43,193 @@
...
+        throws RestClientException, IOException {
+
+      try {
+        return delegate.getSchemaBySubjectAndId(subject, id);
+      } catch (final RestClientException e) {
+        if (e.getStatus() == HttpStatus.SC_NOT_FOUND) {
+          final ParsedSchema schemaByName = subjectCache.get(subject);
+          final ParsedSchema schemaById = idCache.get(id);
+          if (schemaByName != null && schemaByName == schemaById) {
+            return schemaByName;
```
lihaosky, line 120: ditto

ksqldb-engine/src/main/java/io/confluent/ksql/services/SandboxedSchemaRegistryClient.java
```diff
@@ -36,21 +43,193 @@
...
+
+    @Override
+    public int getId(final String subject, final ParsedSchema parsedSchema)
+        throws RestClientException, IOException {
+      try {
+        return delegate.getId(subject, parsedSchema);
+      } catch (final RestClientException e) {
+        if (e.getStatus() == HttpStatus.SC_NOT_FOUND && subjectToId.containsKey(subject)) {
+          return subjectToId.get(subject);
+        }
```
lihaosky, line 225: ditto

ksqldb-engine/src/main/java/io/confluent/ksql/services/SandboxedSchemaRegistryClient.java
mjsax: Should it not depend on the compatibility rules if updating the schema actually works? Not sure to what extend we need to fully mimic SR logic? -- I tired to keep it simple in the hope it would be sufficient. Not totally sure to be fair. Thoughts?

ksqldb-engine/src/main/java/io/confluent/ksql/services/SandboxedSchemaRegistryClient.java
mjsax: Maybe. What would we gain?

ksqldb-engine/src/main/java/io/confluent/ksql/services/SandboxedSchemaRegistryClient.java
mjsax: Old code swallows both: `.swallow("register", anyParams(), 123)`

I don't think ksqlDB actually uses this method? I just added it to make the existing unit tests pass.

ksqldb-engine/src/main/java/io/confluent/ksql/services/SandboxedSchemaRegistryClient.java
lihaosky: This will make the following fail because they share same subject?
`CREATE test_stream_1 (id int key, name varchar) with (kafka_topic='some_topic', partitions=1, format='avro')`
`CREATE test_stream_2 (id int key, name varchar, age int) with (kafka_topic='some_topic', partitions=1, format='avro')`

ksqldb-engine/src/main/java/io/confluent/ksql/services/SandboxedSchemaRegistryClient.java
lihaosky: 1. Make maintaining the 3 caches here easier and less error prone? `MockSchemaRegistryClient` seems to have similar logic inside.
2. Make behavior same as prod schema registry. For example, `MockSchemaRegistryClient` could have register same subject differently instead of throwing.

ksqldb-engine/src/main/java/io/confluent/ksql/services/SandboxedSchemaRegistryClient.java
vcrfxia: I think we should err on the side of false positives rather than false negatives. It'd be bad to have statements that actually work fail validation because then we might be introducing a regression (if a customer uses such statements in their workflows). In contrast, a false positive means we let some statements through validation that then fail in execution. Not great since multi-statements requests might have been partially executed, but also doesn't prevent any existing workflows.

ksqldb-engine/src/main/java/io/confluent/ksql/services/SandboxedSchemaRegistryClient.java
vcrfxia: +1 if we can use the MockSchemaRegistryClient rather than maintaining our own caches for the newly registered subjects, that'd be nice. OTOH if it's a lot of effort I think the current approach is fine too. Let's just be careful about not throwing errors where we shouldn't be (see above).

vcrfxia (commented):
Thanks @mjsax -- great test coverage! Comments inline.

ksqldb-engine/src/test/java/io/confluent/ksql/integration/DependentStatementsIntegrationTest.java
```diff
@@ -0,0 +1,238 @@
...
+  @Before
+  public void before() throws Exception {
+    TEST_HARNESS.before();
+    ksqlContext  = TEST_HARNESS.ksqlContextBuilder()
+        .withAdditionalConfig(
+            KsqlConfig.SCHEMA_REGISTRY_URL_PROPERTY,
+            "http://foo:8080")
+        .withAdditionalConfig(
+            KsqlConfig.KSQL_SHARED_RUNTIME_ENABLED,
+            sharedRuntimes)
```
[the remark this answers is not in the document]
mjsax: I just blindly copied from `EndToEndIntegrationTest` -- not even sure that the difference is. Happy to remove it.

ksqldb-engine/src/test/java/io/confluent/ksql/integration/DependentStatementsIntegrationTest.java
```diff
@@ -0,0 +1,238 @@
...
+import org.junit.rules.RuleChain;
+import org.junit.rules.Timeout;
+import org.junit.runner.RunWith;
+import org.junit.runners.Parameterized;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+@RunWith(Parameterized.class)
+@Category({IntegrationTest.class})
+public class DependentStatementsIntegrationTest {
```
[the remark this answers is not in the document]
mjsax: Should be covered (cf https://github.com/confluentinc/ksql/issues/4800)?

ksqldb-engine/src/test/java/io/confluent/ksql/integration/DependentStatementsIntegrationTest.java
```diff
@@ -0,0 +1,159 @@
...
+      String.valueOf(valueSchemaId)
+    );
+  }
+
+  private void executeStatement(
+      final String statement,
+      final String... args
+  ) {
+    final String formatted = format(statement, (Object[])args);
+    log.debug("Sending statement: {}", formatted);
```
lihaosky, line 146: nit: remove logging in test?

ksqldb-engine/src/test/java/io/confluent/ksql/services/SandboxedSchemaRegistryClientTest.java
```diff
@@ -196,15 +197,21 @@ public void shouldGetSchemaById() throws Exception {
 
     @Test
     public void shouldGetId() throws Exception {
+      // Given:
+      when(delegate.getId(anyString(), any(ParsedSchema.class)))
```
lihaosky, line 201: Do we need case where `delegate` throws `SC_NOT_FOUND` but mock client returns id?

lihaosky (commented):
LGTM overall! Some nits and please take a look at test failure.

ksqldb-engine/src/main/java/io/confluent/ksql/services/SandboxedSchemaRegistryClient.java
```diff
@@ -36,21 +43,213 @@
...
-        .forward("getVersion", methodParams(String.class, ParsedSchema.class), delegate)
-        .build();
+    return new SandboxSchemaRegistryCache(delegate);
   }
 
   private SandboxedSchemaRegistryClient() {
   }
+
+  static final class SandboxSchemaRegistryCache implements SchemaRegistryClient {
+    private final MockSchemaRegistryClient mockedClient = new MockSchemaRegistryClient();
```
vcrfxia, line 53: Can we add comments explaining the difference between `delegate` and `mockedClient` here, i.e., what each is used for/represents? Without this context, this code is very difficult to understand.

We might also want to rename `mockedClient` to `sandboxCacheClient` or similar, in order to clarify within the code itself as well.

ksqldb-engine/src/main/java/io/confluent/ksql/services/SandboxedSchemaRegistryClient.java
```diff
@@ -36,21 +43,213 @@
...
+    }
+
+    @Override
+    public ParsedSchema getSchemaById(final int id) throws RestClientException, IOException {
+      try {
+        return delegate.getSchemaById(id);
+      } catch (RestClientException e) {
+        // if we don't find the schema in SR, we try to get it from the sandbox cache
+        if (e.getStatus() == HttpStatus.SC_NOT_FOUND) {
+          return mockedClient.getSchemaById(id);
```
vcrfxia, line 97: The earlier version of this PR only returned from this line if the schema was present in the sandbox cache, and otherwise threw on the line below. This latest version of the PR now always returns from the mocked client representing the sandbox cache. Does the mocked client throw an error if the schema is not present? Otherwise there is a behavior change here (which might be fine, wondering if it's intentional).

Same comment/question for the other methods below too (`getSchemaBySubjectAndId`, `getLatestSchemaMetadata`, `getVersion`, and `getId`).

ksqldb-engine/src/test/java/io/confluent/ksql/integration/DependentStatementsIntegrationTest.java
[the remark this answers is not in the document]
vcrfxia: The PR that fixed that issue only added unit test coverage, and no integration test coverage. I meant to say that it'd be nice to add an integration test into this new DependentStatementsIntegrationTest.java file for the topic dependency case too.

vcrfxia (commented):
Thanks @mjsax ! LGTM with some minor comments/questions inline.

ksqldb-engine/src/main/java/io/confluent/ksql/services/SandboxedSchemaRegistryClient.java
lihaosky: It also throws if not found: https://github.com/confluentinc/schema-registry/blob/master/client/src/main/java/io/confluent/kafka/schemaregistry/client/MockSchemaRegistryClient.java#L123-L143.

mjsax:
Test failed a second time, but it passes locally:
```
Expected: is <3>
     but: was <2>
	at io.confluent.ksql.physical.scalablepush.locator.AllHostsLocatorTest.shouldLocate(AllHostsLocatorTest.java:56)
```

Any idea? Let's see if it fails again in the next build.

lihaosky:
Is `AllHostsLocatorTest` failure related to https://github.com/confluentinc/ksql/pull/8665? Latest build failure seems to be a different one though: 

```
KsqlResourceFunctionalTest.shouldInsertIntoValuesForAvroTopic:201->makeKsqlRequest:267 Failed to await result.msg: Failed to insert values into 'BOOKS'.        Could not serialize key: ['Metamorphosis']
```

ksqldb-engine/src/main/java/io/confluent/ksql/services/SandboxedSchemaRegistryClient.java
```diff
@@ -36,21 +43,217 @@
...
-        .build();
+    return new SandboxSchemaRegistryCache(delegate);
   }
 
   private SandboxedSchemaRegistryClient() {
   }
+
+  static final class SandboxSchemaRegistryCache implements SchemaRegistryClient {
+    // we use `MockSchemaRegistryClient` as a cache inside the sandbox to store
+    // newly registered schema (with polluting the actual SR)
```
vcrfxia, line 54:
```suggestion
    // newly registered schemas (without polluting the actual SR)
```

ksqldb-engine/src/main/java/io/confluent/ksql/services/SandboxedSchemaRegistryClient.java
```diff
@@ -36,21 +43,217 @@
...
+    // this allowd dependent statement to execute successfully inside the sandbox
```
vcrfxia, line 55:
```suggestion
    // this allows dependent statements to execute successfully inside the sandbox
```

ksqldb-engine/src/test/java/io/confluent/ksql/integration/DependentStatementsIntegrationTest.java
```diff
@@ -0,0 +1,170 @@
...
+
+  @After
+  public void after() {
+    toClose.forEach(QueryMetadata::close);
+    ksqlContext.after();
+    TEST_HARNESS.after();
+  }
+
+  @Test
+  public void shouldCreateDependentTopicWithDefaultReplicationInSandbox() {
```
vcrfxia, line 81: This test doesn't actually test a topic dependency, does it? I thought the dependency issue fixed in the previous PR was if the first CREATE STREAM creates a topic and then the second one uses the newly created topic. (Second statement should reference the topic, not the stream.)

This dependency case (CS followed by CSAS) should already be tested in other integration tests throughout the codebase.

vcrfxia (commented):
Thanks @mjsax , a couple more comments inline.

The AllHostsLocatorTest failure was unrelated (see https://github.com/confluentinc/ksql/pull/8664). Not sure about the new one.

ksqldb-engine/src/test/java/io/confluent/ksql/integration/DependentStatementsIntegrationTest.java
mjsax: No, the fix was about "inherit the replication factor". The second statement creates an output topic, and was not able to create it, because it did neither inherit the replication factor from upstream topic, nor did if fetch default replication factor from Kafka.

Happy to remove this test again -- I thought you asked to add it. I can also add a different one that use two CS statements, but it also seems redundant?

vcrfxia: Oh, huh! I totally misunderstood the original ticket for that issue then, haha. If this integration test covers the fix, then that sounds great to me. Thanks for adding it!

vcrfxia (approved):
Thanks @mjsax ! LGTM.

mjsax:
Another (different) test failure:
```
java.lang.AssertionError: Failed to await result.msg: Failed to insert values into 'BOOKS'. Could not serialize key: ['Metamorphosis']
	at io.confluent.ksql.rest.integration.KsqlResourceFunctionalTest.makeKsqlRequest(KsqlResourceFunctionalTest.java:267)
	at io.confluent.ksql.rest.integration.KsqlResourceFunctionalTest.shouldInsertIntoValuesForAvroTopic(KsqlResourceFunctionalTest.java:201)
```

Failed locally, too. Is this related to broken `master` ?

PR #8635, opened 2022-01-20
feat: allow custom connect auth header configuration

vcrfxia:
### Description 

This PR builds on the custom connect headers extension introduced in https://github.com/confluentinc/ksql/pull/8357. Three changes in this PR:
* The auth header used by ksqlDB for connector requests is now configurable as well. 
* The principal associated with the incoming ksql request is no longer piped through to the extension, as the interface for custom headers no longer accepts the KsqlPrincipal as a parameter. (We thought we had a use case for this but turns out we don't, so I've cleaned up the code.)
* Incoming request headers are now piped through to the extension instead, as the custom auth header might be interested in these.

### Testing done 

Unit tests + manual.

### Reviewer checklist

### DIFF
--- a/ksqldb-engine/src/main/java/io/confluent/ksql/connect/ConnectRequestHeadersExtension.java
+++ b/ksqldb-engine/src/main/java/io/confluent/ksql/connect/ConnectRequestHeadersExtension.java
@@ -15,8 +15,10 @@
 
 package io.confluent.ksql.connect;
 
-import io.confluent.ksql.security.KsqlPrincipal;
+import java.util.Collections;
+import java.util.List;
 import java.util.Map;
+import java.util.Map.Entry;
 import java.util.Optional;
 
 /**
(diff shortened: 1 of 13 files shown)


ksqldb-rest-app/src/main/java/io/confluent/ksql/api/auth/ApiSecurityContext.java
```diff
@@ -16,11 +16,15 @@
...
+import java.util.Map.Entry;
 import java.util.Optional;
 
 public interface ApiSecurityContext {
 
   Optional<KsqlPrincipal> getPrincipal();
 
   Optional<String> getAuthToken();
+
+  List<Entry<String, String>> getRequestHeaders();
```
vcrfxia, line 29: Passing incoming request headers as part of the ApiSecurityContext in this way means the existing `Optional<String> getAuthToken();` is not strictly needed anymore since the auth token is extracted from the auth header. I've decided to leave it in despite the redundancy since the auth token is used in multiple places (both the connect client and the internal ksql client) and passing an extra string seems fine compared to having code in multiple places for extracting the header. Happy to refactor if others disagree, though.

ksqldb-engine/src/main/java/io/confluent/ksql/connect/ConnectRequestHeadersExtension.java
```diff
@@ -25,16 +27,44 @@
...
+   * ksqlDB's default auth header for ksqlDB connector requests. This auth header
+   * configured by this extension takes precedence over any other custom auth
+   * configurations for ksqlDB connector requests, including basic auth configured in
+   * ksqlDB server properties.
+   *
+   * @return whether to use the custom auth header returned from
+   *         {@link ConnectRequestHeadersExtension#getAuthHeader(List)} instead of
+   *         ksqlDB's default auth header for ksqlDB connector requests
+   */
+  default boolean shouldUseCustomAuthHeader() {
```
Gerrrr, line 42: In what situations `ConnectRequestHeadersExtension` is configured, but it should not be used?

Gerrrr (commented):
Looks good! I just left one question to understand the purpose of `shouldUseCustomAuthHeader`.

ksqldb-engine/src/main/java/io/confluent/ksql/connect/ConnectRequestHeadersExtension.java
vcrfxia: If someone wants to only use the extension for custom additional headers (via the `getHeaders()` method) but wants to continue using ksqlDB's default auth header, then they can set `shouldUseCustomAuthHeader = false` rather than re-implementing ksqlDB's logic for computing the default auth header.

vcrfxia: I can add a line to the javadoc to clarify.

Gerrrr: Thank you! This makes sense.

Gerrrr (approved):

PR #8636, opened 2022-01-21
docs: add missing environment variable in event-driven-microservice tutorial

lrworth:
### Description 
Running `docker-compose up` with the given docker-compose.yml results in schema-registry exiting with an error about not finding bootstrap servers.

### Testing done 
I ran `docker-compose up` with the modified version and it worked.

### Reviewer checklist


JimGalasyn (approved):
Good catch, thanks for the fix!

PR #8640, opened 2022-01-21
fix: clean up custom prefixed internal topics

wcarlson5:
https://github.com/confluentinc/ksql/pull/8607 made it so the app id was not always the prefix for internal topics now we clean up based on that not the app id 

### Testing done 

### Reviewer checklist

### DIFF
--- a/ksqldb-common/src/main/java/io/confluent/ksql/util/QueryApplicationId.java
+++ b/ksqldb-common/src/main/java/io/confluent/ksql/util/QueryApplicationId.java
@@ -69,7 +69,7 @@ public final class QueryApplicationId {
   public static String buildPrefix(
       final String serviceId,
       final String queryPrefix
-   ) {
+  ) {
     return ReservedInternalTopics.KSQL_INTERNAL_TOPIC_PREFIX
         + serviceId
         + queryPrefix;
--- a/ksqldb-common/src/main/java/io/confluent/ksql/util/QueryApplicationId.java
+++ b/ksqldb-common/src/main/java/io/confluent/ksql/util/QueryApplicationId.java
@@ -30,7 +30,7 @@ public final class QueryApplicationId {
       final boolean persistent,
       final int sharedRuntimeIndex
   ) {
-    final String queryAppId = buildPrefix(config, persistent) + sharedRuntimeIndex;
+    final String queryAppId = buildInternalTopicPrefix(config, persistent) + sharedRuntimeIndex;
     if (persistent) {
       return queryAppId;
     } else {
@@ -43,7 +43,7 @@ public final class QueryApplicationId {
       final boolean persistent,
       final QueryId queryId
   ) {
-    final String queryAppId = buildPrefix(config, persistent) + queryId;
+    final String queryAppId = buildInternalTopicPrefix(config, persistent) + queryId;
     if (persistent) {
       return queryAppId;
     } else {
@@ -51,7 +51,7 @@ public final class QueryApplicationId {
     }
   }
 
-  public static String buildPrefix(
+  public static String buildInternalTopicPrefix(
       final KsqlConfig config,
       final boolean persistent
   ) {
@@ -63,10 +63,10 @@ public final class QueryApplicationId {
 
     final String queryPrefix = config.getString(configName);
 
-    return buildPrefix(serviceId, queryPrefix);
+    return buildInternalTopicPrefix(serviceId, queryPrefix);
   }
 
-  public static String buildPrefix(
+  public static String buildInternalTopicPrefix(
       final String serviceId,
       final String queryPrefix
   ) {
--- a/ksqldb-common/src/main/java/io/confluent/ksql/util/QueryApplicationId.java
+++ b/ksqldb-common/src/main/java/io/confluent/ksql/util/QueryApplicationId.java
@@ -30,7 +30,7 @@ public final class QueryApplicationId {
       final boolean persistent,
       final int sharedRuntimeIndex
   ) {
-    final String queryAppId = buildPrefix(config, persistent) + "-" + sharedRuntimeIndex;
+    final String queryAppId = buildPrefix(config, persistent) + sharedRuntimeIndex;
     if (persistent) {
       return queryAppId;
     } else {
@@ -51,7 +51,7 @@ public final class QueryApplicationId {
     }
   }
 
-  private static String buildPrefix(
+  public static String buildPrefix(
       final KsqlConfig config,
       final boolean persistent
   ) {
@@ -63,6 +63,13 @@ public final class QueryApplicationId {
 
     final String queryPrefix = config.getString(configName);
 
+    return buildPrefix(serviceId, queryPrefix);
+  }
+
+  public static String buildPrefix(
+      final String serviceId,
+      final String queryPrefix
+   ) {
     return ReservedInternalTopics.KSQL_INTERNAL_TOPIC_PREFIX
         + serviceId
         + queryPrefix;
--- a/ksqldb-engine/src/main/java/io/confluent/ksql/engine/QueryCleanupService.java
+++ b/ksqldb-engine/src/main/java/io/confluent/ksql/engine/QueryCleanupService.java
@@ -20,6 +20,7 @@ import com.google.common.util.concurrent.AbstractExecutionThreadService;
 import com.spun.util.io.FileUtils;
 import io.confluent.ksql.schema.registry.SchemaRegistryUtil;
 import io.confluent.ksql.services.ServiceContext;
+import io.confluent.ksql.util.QueryApplicationId;
 import java.io.File;
 import java.nio.file.Path;
 import java.nio.file.Paths;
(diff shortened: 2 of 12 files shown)


ksqldb-engine/src/main/java/io/confluent/ksql/engine/QueryCleanupService.java
```diff
@@ -119,6 +119,11 @@ public QueryCleanupTask(
       }
     }
 
+    private String getQueryTopicPrefix() {
+      return appId.split("query")[0]
```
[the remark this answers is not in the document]
wcarlson5: sadly no, we need the service ID from the config

ksqldb-engine/src/main/java/io/confluent/ksql/engine/QueryCleanupService.java
```diff
@@ -108,7 +108,7 @@ public QueryCleanupTask(
       this.serviceContext = Objects.requireNonNull(serviceContext, "serviceContext");
       this.appId = Objects.requireNonNull(appId, "appId");
       this.topologyName = Objects.requireNonNull(topologyName, "topologyName");
-      queryTopicPrefix = topologyName.map(s -> appId + "-" + s).orElse(appId);
+      queryTopicPrefix = topologyName.map(s -> getQueryTopicPrefix() + "-" + s).orElse(appId);
```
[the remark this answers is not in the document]
wcarlson5: good idea

ksqldb-engine/src/main/java/io/confluent/ksql/services/KafkaTopicClientImpl.java
```diff
@@ -315,12 +315,12 @@ public void deleteTopics(final Collection<String> topicsToDelete) {
   }
 
   @Override
-  public void deleteInternalTopics(final String applicationId) {
+  public void deleteInternalTopics(final String internalTopicPrefix) {
```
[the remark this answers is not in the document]
wcarlson5: We probably should. There was a little bit of work to get the restore working for shared runtimes at all but it wasn't as bad as I had feared.

ksqldb-engine/src/main/java/io/confluent/ksql/engine/QueryCleanupService.java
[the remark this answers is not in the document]
ableegoldman: Oh I see, I thought we just needed a regex-matching substring but I see now it's supposed to contain the full prefix for the topic name

ableegoldman:
Hey, this is actually related to #8607 but I only happened to notice it just now -- I think [this ](https://github.com/confluentinc/ksql/pull/8607/files#diff-07688cabdbef935830f98db31c2be835ffde5f1ea481f8ac1509d35c20840989R571-R573) should (a) be factored into a public utility method so that we have a single source of truth if we ever need to access it elsewhere rather than try to rebuild it each time and trying to keep everything in sync -- I'd recommend adding it to QueryApplicationId, yes it's not technically an app id but it's essentially analogue to one and anyways shares most of the same components plus we'd presumably want to stay in sync if e.g. we ever added yet another component to the query app id (like a new prefix config, idk 🤷‍♀️ )

and (b) consider making it completely identical to the app id for shared runtimes, just dropping the index/replacing that with the "query" thing

On that note: is the only reason we add the "query" to  this prefix to make it easier/possible to parse and extract the prefix substring? Because if we don't have a strict need for this "query" piece I believe we can just reuse the `QueryApplicationId#build` method instead of adding a new utility method at all. Does that make sense or did I miss anything?

ableegoldman:
For example could we just call QueryApplicationId#build or the new utility method if we add one rather than do this parsing [here](https://github.com/confluentinc/ksql/pull/8640/files#diff-9057f1ba8c4274a58fdb8012ca9fca1e185d716eb2f9512f375aabe38fea8919R122-R127)?

ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/restore/KsqlRestoreCommandTopic.java
```diff
@@ -394,30 +395,42 @@ private static void maybeCleanUpQuery(final byte[] command, final KsqlConfig ksq
...
         queryIdFound = true;
       }
     }
 
     // the command contains a query, clean up it's internal state store and also the internal topics
     if (queryIdFound) {
       final StreamsConfig streamsConfig = new StreamsConfig(streamsProperties);
-      final String applicationId =
-          streamsConfig.getString(StreamsConfig.APPLICATION_ID_CONFIG);
+      String topicPrefix = streamsConfig.getString(StreamsConfig.APPLICATION_ID_CONFIG);;
```
ableegoldman, line 424: nit: use ternary operator ( the xx ? yy : zz thing) to set this so we can make it `final`, or if/else with separate declaration, whatever

ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/restore/KsqlRestoreCommandTopic.java
```diff
@@ -427,13 +440,20 @@ private static void maybeCleanUpQuery(final byte[] command, final KsqlConfig ksq
         System.out.println(
             String.format(
                 "Cleaned up internal state store and internal topics for query %s",
-                applicationId));
+                topicPrefix));
       } catch (final Exception e) {
-        System.out.println(String.format("Failed to clean up query %s ", applicationId));
+        System.out.println(String.format("Failed to clean up query %s ", topicPrefix));
```
ableegoldman, line 445: Wonder why we use `System.out.println` here instead of log4j? Any of @rodesai @agavra @vcrfxia know if this is intentional?

guozhangwang:
> Hey, this is actually related to #8607 but I only happened to notice it just now -- I think this should (a) be factored into a public utility method so that we have a single source of truth if we ever need to access it elsewhere rather than try to rebuild it each time and trying to keep everything in sync -- I'd recommend adding it to QueryApplicationId, yes it's not technically an app id but it's essentially analogue to one and anyways shares most of the same components plus we'd presumably want to stay in sync if e.g. we ever added yet another component to the query app id (like a new prefix config, idk 🤷‍♀️ ) and (b) consider making it completely identical to the app id for shared runtimes, just dropping the index/replacing that with the "query" thing

I'm +100 on this.

Otherwise, the PR lgtm.

wcarlson5:
@ableegoldman The "query thing is legacy and is baked into several of our other things like the dd agent and I don't know all of them. I would rather not change it I don't have too. Another reason I have not made the topic prefix construction method a  public utility method is that it needs a config. And not all the callers have the config to use. I can make it public and make a public one to extract it form the app id too. would that work?

ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/restore/KsqlRestoreCommandTopic.java
wcarlson5: @ableegoldman The restore cmd topic tool is run via the cli I imagine that it is used there

ksqldb-engine/src/main/java/io/confluent/ksql/engine/QueryCleanupService.java
```diff
@@ -106,12 +106,13 @@ public QueryCleanupTask(
         final Optional<String> topologyName,
```
ableegoldman, line 106: super nit: rename to `queryId`? Ksql devs may not necessarily know what it means for a topology to be named or understand  how it relates/why we would even care about this name in the first place

ksqldb-common/src/main/java/io/confluent/ksql/util/QueryApplicationId.java
```diff
@@ -30,7 +30,7 @@ public static String buildSharedRuntimeId(
       final boolean persistent,
       final int sharedRuntimeIndex
   ) {
-    final String queryAppId = buildPrefix(config, persistent) + "-" + sharedRuntimeIndex;
```
ableegoldman, line 33: seriously great catch -- would have sucked to get stuck with this awkward looking thing

ksqldb-common/src/main/java/io/confluent/ksql/util/QueryApplicationId.java
```diff
@@ -51,7 +51,7 @@ public static String build(
     }
   }
 
-  private static String buildPrefix(
+  public static String buildPrefix(
```
ableegoldman, line 54: nit: rename this and the other to `buildInternalTopicPrefix` or similar to make it clear what exactly this returns the prefix for

ableegoldman (approved):
Looking good, had some code clarity suggestions that I think are worth committing since the PR build hasn't been going for very long yet

Beyond that I'd recommend running all the tests for any module that's touched on in this PR or seems potentially connected, if you aren't already. Don't want to have to wait for the entire PR build to run to discover there's some failing unit test or checkstyle issue, let's get this guy merged ASAP!

PR #8644, opened 2022-01-24
build: Update Git Commit Id plugin

jnh5y:
* This version allows for use of the flag -Dmaven.gitcommitid.nativegit=true
to allow developers to use native git rather than JGit.

### Description 
* This version allows for use of the flag -Dmaven.gitcommitid.nativegit=true
to allow developers to use native git rather than JGit.

### Reviewer checklist

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -140,7 +140,7 @@
         <!-- <compile.warnings-flag>-Werror</compile.warnings-flag> -->
         <!-- Only used to provide login module implementation for tests -->
         <jetty.version>9.4.28.v20200408</jetty.version>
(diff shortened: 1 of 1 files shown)


jnh5y:
Should this PR include some update to a markdown file to let developers know that they can use native git?  (Native git takes 2 seconds for the git-commit-it versus 12 seconds for JGit--at least on my laptop.)

spena (approved):
+1

As you said, it might be good to add some docs, but I don't find what doc should be good to have it.

jnh5y:
> +1
> 
> As you said, it might be good to add some docs, but I don't find what doc should be good to have it.

Eh, maybe I could add something to the bottom of the Contributing.MD?  That seems best and also wrong.:)

Issue #5706, opened 2020-06-26
Allow UDF functions to be executed without the FROM clause


spena:
**Is your feature request related to a problem? Please describe.**
Sometimes, we just need to test UDF functions to verify they work, i.e. `select ucase('abc');`, but that requires users to create a dummy stream, insert at least a row, set the offset to earliest, and execute the query with the UDF `select ucase('abc') from dummy_stream emit changes limit 1`. 

**Describe the solution you'd like**
Allow users to execute UDF tests in their queries without requiring the FROM clause.
i.e.
```
ksql> select ucase('abc');
col_0
----------
ABC
Query terminated
```

**Describe alternatives you've considered**
No alternatives. The only way to make it work is:
1. Create a dummy stream (including the dummy topic)
2. Insert at least a row in the stream
3. Set offset to earliest
4. Execute query with UDF

PR #8645, opened 2022-01-24
WIP: feat: Allow functions to be executed without a FROM clause

jnh5y:
Addresses: https://github.com/confluentinc/ksql/issues/5706

### Description 
This change adds a new execution pathway so that ksql console users can try out functions directly.

### Testing done 
So far, I have only tested manually. 

### Reviewer checklist

### DIFF
--- a/ksqldb-parser/src/main/antlr4/io/confluent/ksql/parser/SqlBase.g4
+++ b/ksqldb-parser/src/main/antlr4/io/confluent/ksql/parser/SqlBase.g4
@@ -46,8 +46,7 @@ singleExpression
 
 statement
     : query                                                                 #queryStatement
-    | expression                                                            #expressionEvaluate
-    | EVALUTE expression                                                    #evaluateExpression
+    | (SELECT) expression                                                   #expressionEvaluate
     | (LIST | SHOW) PROPERTIES                                              #listProperties
     | (LIST | SHOW) ALL? TOPICS EXTENDED?                                   #listTopics
     | (LIST | SHOW) STREAMS EXTENDED?                                       #listStreams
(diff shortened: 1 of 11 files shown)


jnh5y:
To be done:
[] Unit tests.
[] Deciding on syntax.
[] Generalizing type support.
[] Docs.

ksqldb-parser/src/main/antlr4/io/confluent/ksql/parser/SqlBase.g4
```diff
@@ -46,6 +46,8 @@ singleExpression
 
 statement
     : query                                                                 #queryStatement
+    | expression                                                            #expressionEvaluate
+    | EVALUTE expression                                                    #evaluateExpression
```
spena, line 50: what's the different with calling `expression` and `EVALUATE expression`? Isn't the same, both calling the expression?

Btw, all DBs I've seen use the select clause for calling an expression, i.e. `select expression;`. Can we do it like that? We could call `select expr1, expr2, expr3;` and display one row in 3 columns with one expression each.

I think you can make all clauses in the select optional, like `query: SELECT selectItem (',' selectItem)* (<the rest>)?`

spena (commented):
The code changes look right to support an expression evaluation. I think there are other places we need to change. But, I think we should keep a standard for how we execute some expressions, like `SELECT expr1;`. 

One option would be to modify the `Query` class to accept an optional from clause and then start from there? Maybe in the code, you can check if a query does not have a from clause, then we just call the code to execute the expression.

Another option is to stay with other DBs using a DUAL table like Oracle - https://docs.oracle.com/cd/B19306_01/server.102/b14200/queries009.htm. They have a special table called DUAL. They can do `SELECT expr1 FROM DUAL;`. DUAL is a dummy table just to allow users run an expression. That might reduce some code work and void making the FROM clause optional. But not sure about possible compatibility issues if users have created a DUAL stream or table.

PR #8652, opened 2022-01-25
chore: add custom metrics tags to the total bytes sum metric for the server

stevenpyzhang:
### Description 
We want to also include custom metrics tags in the total bytes consumed metric in the ConsumerCollector class. We do this by adding the value from the KsqlConfig into the streams properties map in QueryBuilder

### Testing done 
Unit tests

Added `ksql.metrics.tags.custom=test-tag:1337` to the server properties
Started up server
Opened jconsole
Looked at the bean name: `io.confluent.ksql.metrics:type=consumer-metrics,test-tag:1337`

### Reviewer checklist

### DIFF
--- a/ksqldb-common/src/main/java/io/confluent/ksql/metrics/ConsumerCollector.java
+++ b/ksqldb-common/src/main/java/io/confluent/ksql/metrics/ConsumerCollector.java
@@ -21,7 +21,6 @@ import io.confluent.ksql.metrics.TopicSensors.SensorMetric;
 import io.confluent.ksql.util.KsqlConfig;
 import java.util.ArrayList;
 import java.util.Collection;
-import java.util.Collections;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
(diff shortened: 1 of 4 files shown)


ksqldb-common/src/main/java/io/confluent/ksql/metrics/ConsumerCollector.java
```diff
@@ -69,17 +68,26 @@ public void configure(final Map<String, ?> map) {
     if (id == null) {
       id = (String) map.get(ConsumerConfig.CLIENT_ID_CONFIG);
     }
+
+    final KsqlConfig config = new KsqlConfig(map);
```
guozhangwang, line 72: Do we have to first wrap it in a `KsqlConfig` and then get by the config name? I saw that it's default value is `""` anyways so seems just getting from the `map` with static `parseStringAsMap` is fine too.

guozhangwang (approved):
LGTM!

guozhangwang (approved):
LGTM!

stevenpyzhang:
@guozhangwang I ended up reverting the last commit since it leads to some NPE in MetricsCollectorTests.java. I think it's cleaner to just cast it to KsqlConfig which helps us handle these null cases in the map passed into the config function

guozhangwang:
> @guozhangwang I ended up reverting the last commit since it leads to some NPE in MetricsCollectorTests.java. I think it's cleaner to just cast it to KsqlConfig which helps us handle these null cases in the map passed into the config function

Cool, thanks for the follow-up explanation!

guozhangwang (approved):

PR #8660, opened 2022-01-26
fix: remove usages of internal Streams API/variable that was removed upstream

ableegoldman:
Apparently we used the StreamsMetadataImpl.NOT_AVAILABLE variable throughout the ksql repo, but we very much should not have been since this is actually not in the public API of Streams. The risk of using internal APIs is that they have no public contract or compatability guarantees of any kind, and can for example be removed at any time without deprecation/warning.

Unfortunately this is exactly what just happened to `StreamsMetadataImpl.NOT_AVAILABLE`, but at least we can easily address this since it was pretty much just a convenience to begin with used for comparisons, and we no longer return any metadata for unavailable clients rather than build up this metadata and then just filter it out on the user side (ie ksql)

### DIFF
--- a/ksqldb-engine/src/main/java/io/confluent/ksql/physical/scalablepush/locator/AllHostsLocator.java
+++ b/ksqldb-engine/src/main/java/io/confluent/ksql/physical/scalablepush/locator/AllHostsLocator.java
@@ -62,7 +62,6 @@ public class AllHostsLocator implements PushLocator {
         .map(QueryMetadata::getAllMetadata)
         .filter(Objects::nonNull)
         .flatMap(Collection::stream)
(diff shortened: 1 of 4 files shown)


patrickstuedi (approved):
LGTM, thanks @ableegoldman !

ableegoldman:
Merging ASAP since this is blocking the 0.24 release and the build in general, I did test locally to make sure the patch fixes the build but if there are test failures somehow that come up later they should be addressed in a followup PR

guozhangwang (commented):
Just curious do we have a separate PR to fix the tests now?

jnh5y:
@ableegoldman @guozhangwang I peeked at this and tossed up a PR to fix the `AllHostsLocatorTest` here: https://github.com/confluentinc/ksql/pull/8664

PR #8667, opened 2022-01-27
fix: don't throw on drop

wcarlson5:
Don't hit an immutability issue. the query returns its list of sources in an immutable set so we need to add them all to the existing set instead of replace them

ksql> drop stream filtered_users;

 Message                                                      
--------------------------------------------------------------
 Source `FILTERED_USERS` (topic: FILTERED_USERS) was dropped. 
--------------------------------------------------------------

instead of

ksql> drop table users_agg_2;

 Message                              
--------------------------------------
 Source `USERS_AGG_2` does not exist. 
--------------------------------------

### Testing done 
added coverage in unit tests

### Reviewer checklist


lct45 (approved):
lgtm

PR #8668, opened 2022-01-28
fix: keep shared runtime dirs even if no query has state

wcarlson5:
we were looking at all dir for queries and removing all the ones that were being actively used. but for named topologies if there was no state then we would not see that runtime as being used and we would delete it out from under streams. Streams would get confused and when you then tried to drop the query for real everything would hang

### Testing done 
Manual testing

### Reviewer checklist


nateab (approved):
LGTM!

PR #8672, opened 2022-01-28
WIP: trying native git on CI.

jnh5y:
### Description 

### Testing done 

### Reviewer checklist


lihaosky:
noob question: why do we need this change?

jnh5y:
@lihaosky The git-commit-id plugin defaults to using JGit.  The JGit client takes 12 seconds on my laptop (and also it seems the CI servers).  The native client takes 2 seconds.  

The git-commit-id plugin runs for each of the 28 modules, so that 10 second saving should be about 5 minutes...  I hope.

From looking the CI for this job, it ran in 2 hr 9 min.  A green build from last night ran in 2h 16min.

lihaosky (commented):
I see. I found below in https://github.com/git-commit-id/git-commit-id-maven-plugin/blob/master/docs/using-the-plugin.md:

> Explanation:
This plugin ships with custom `jgit` implementation that is being used to obtain all
relevant information. If set to to `true` this plugin will use the native `git` binary
instead of the custom `jgit` implementation.
Although this should usually give your build some performance boost, it may randomly
break if you upgrade your git version and it decides to print information in a different
format suddenly. As rule of thumb, keep using the default `jgit` implementation (keep
this `false`) until you notice performance problems within your build (usually when you
have *hundreds* of maven modules).
With version *3.0.2*  you can also control it using the commandline option `-Dmaven.gitcommitid.nativegit=true`

Seems there's some risk things will break if using native git. Not sure how risky it is though...

jnh5y:
> Seems there's some risk things will break if using native git. Not sure how risky it is though...

Yes, there's some risk.  Not sure what this would do if someone didn't have a native git client installed.  

This PR: https://github.com/confluentinc/ksql/pull/8677 adds an option to the ksqldb CI build to enable.  If that runs, then we'd have the best of a few worlds.  As developers, we'd be able to turn on the native Git client, and we'd pick up the slightly faster CI build time.;)

lihaosky:
Do you mean we should get https://github.com/confluentinc/ksql/pull/8677 merged instead of this one?

jnh5y:
> Do you mean we should get #8677 merged instead of this one?

Possibly.  At the moment, I'm experimenting with the PRs.  Once I have a recommendation, I'll put a PR without a WIP marker.  Thanks for asking!

jnh5y:
In case the ksqldb build runs in locations without a native Git client, I'm preferring the other PR: https://github.com/confluentinc/ksql/pull/8677

PR #8673, opened 2022-01-28
feat: Use bounds in pull query

vpapavas:
### Description 
This PR accomplishes three things:
1) Enforce consistency bounds on all pull queries
2) Make sure NOT_UP_TO_BOUND errors are returned to the user
3) Try standbys when we get the above error

In implementing the above, I realized that we had some not-so-desired behavior in HARouting regarding error handling and forwarding to standbys. With my changes, we save exceptions we get from standbys as suppressed exceptions. This is helpful for debugging but also to return a meaningful error to the user. Moreover, we always try standbys if a query throws an exception on a node, irrespective of the exception. We don't try other nodes if the response is erroneous.  

I also tried cleaning up the error messages as we catch and throw exceptions many times and it was getting unwieldy.
 
### Testing done 
The most important tests are in  `PullQueryConsistencyFunctionalTest` that have three nodes and test that queries fail/succeed due to consistency constraints.

### Reviewer checklist


ksqldb-rest-client/src/main/java/io/confluent/ksql/rest/client/KsqlRestClient.java
```diff
@@ -208,10 +208,10 @@ public void setServerAddress(final String serverAddress) {
   ) {
     KsqlTarget target = target();
     final Map<String, Object> requestPropertiesToSend = new HashMap<>(requestProperties);
-    if (ConsistencyOffsetVector.isConsistencyVectorEnabled(localProperties.toMap())) {
+    if (ConsistencyOffsetVector.isConsistencyVectorEnabled(requestPropertiesToSend)) {
       final String serializedCV = serializedConsistencyVector.get();
       // KsqlRequest:serializeClassValues throws NPE for null value
-      requestPropertiesToSend.put(
+      requestPropertiesToSend.putIfAbsent(
```
vpapavas, line 214: If I make this change, `putIfAbsent` then I can test the behavior much easier. However, it is allowing someone write access to the consistency token. Then again, they have write access if they do a curl. @vvcephei @patrickstuedi 
 What do you think? Is it ok?

ksqldb-rest-app/src/test/java/io/confluent/ksql/rest/integration/RestApiTest.java
```diff
@@ -879,7 +898,103 @@ public void shouldExecutePullQueryOverRest() {
...
+            requestProperties
+        );
+        return Arrays.asList(response.split(System.lineSeparator()));
+      };
+
+      // Then:
+      final List<String> messages = assertThatEventually(call, hasSize(HEADER + 1));
+      System.out.println("messages = " + messages);
+      assertThat(messages, hasSize(HEADER + 1));
+      assertThat(messages.get(1), containsString("Error in processing query."));
```
vpapavas, line 993: We currently don't return any meaningful error code or message, hence the only thing I can do is check for the presence of AN error

ksqldb-rest-app/src/test/java/io/confluent/ksql/rest/integration/PullQueryConsistencyFunctionalTest.java
```diff
@@ -344,6 +365,104 @@ public void shouldPassConsistencyTokenInFanOutTableScan()
       ImmutableList.of(KEY4, 1)));
   }
 
+  @Test
```
vpapavas, line 368: If we allow the `putIfAbsent` in the client, then I can do the tests in this class without doing a raw rest request. More importantly, it will allow me to test the routing to standby functionality as well

ksqldb-rest-client/src/main/java/io/confluent/ksql/rest/client/KsqlRestClient.java
patrickstuedi: I don't think it's a problem. But can you explain why using put makes testing more difficult?

ksqldb-engine/src/main/java/io/confluent/ksql/physical/pull/HARouting.java
```diff
@@ -200,8 +206,11 @@ private void executeRounds(
       try {
         final PartitionFetchResult fetchResult = future.get();
         if (fetchResult.isError()) {
+          exceptionsPerPartition.computeIfAbsent(
```
vpapavas, line 209: The exceptions thrown by standby nodes are added as suppressed exceptions. Previously, they were lost

ksqldb-engine/src/main/java/io/confluent/ksql/physical/pull/HARouting.java
```diff
@@ -217,23 +226,34 @@ private void executeRounds(
...
     pullQueryQueue.close();
   }
 
   private KsqlPartitionLocation nextNode(final KsqlPartitionLocation partition) {
     return partition.removeHeadHost();
   }
 
-  private static KsqlNode getNodeForRound(
-      final KsqlPartitionLocation location) {
+  private KsqlNode getNodeForRound(
```
vpapavas, line 248: Only throw the `Exhausted standby....` when we are at the router

ksqldb-engine/src/main/java/io/confluent/ksql/physical/pull/HARouting.java
```diff
@@ -284,12 +304,14 @@ static PartitionFetchResult executeOrRouteQuery(
           .ifPresent(queryExecutorMetrics -> queryExecutorMetrics.recordLocalRequests(1));
         synchronized (pullPhysicalPlan) {
           pullPhysicalPlan.execute(ImmutableList.of(location), pullQueryQueue, rowFactory);
-          return new PartitionFetchResult(RoutingResult.SUCCESS, location, Optional.empty());
+          return new PartitionFetchResult(
+              RoutingResult.SUCCESS, location, Optional.empty(), Optional.empty());
         }
-      } catch (StandbyFallbackException e) {
+      } catch (StandbyFallbackException | NotUpToBoundException e) {
```
vpapavas, line 310: This is necessary to allow trying standbys

ksqldb-engine/src/main/java/io/confluent/ksql/physical/pull/HARouting.java
```diff
@@ -361,24 +385,23 @@ private static void forwardTo(
...
-      if (ksqlException != null) {
-        throw ksqlException;
-      }
       // If the exception was caused by closing the connection, we consider this intentional and
       // just return.
       if (shouldCancelRequests.isDone()) {
         LOG.warn("Connection canceled, so returning");
         return;
       }
+
```
vpapavas, line 394: All exceptions thrown by a host that a query was forwarded to, will result in a `StandbyFallbackException` so that we can keep trying other standbys. Previously, we were throwing a `KsqlException` which was preventing us from keep trying other hosts

ksqldb-engine/src/test/java/io/confluent/ksql/physical/pull/HARoutingTest.java
```diff
@@ -674,13 +675,13 @@ public void forwardingError_cancelled() throws ExecutionException, InterruptedEx
   @Test
   public void forwardingError_noRows() {
     // Given:
-    locate(location2);
+    locate(location4);
```
vpapavas, line 678: With the change where we keep trying standbys if a host throws exception, these tests were not failing :) So, had to use a location for a single node

ksqldb-rest-app/src/main/java/io/confluent/ksql/api/server/QuerySubscriber.java
```diff
@@ -93,9 +94,19 @@ private void checkMakeRequest() {
 
   @Override
   public void handleError(final Throwable t) {
-    log.error("Error in processing query", t);
-    final KsqlErrorMessage errorResponse = new KsqlErrorMessage(ERROR_CODE_SERVER_ERROR,
-        "Error in processing query. Check server logs for details.");
+    final StringBuilder stringBuilder = new StringBuilder();
+    stringBuilder.append(t);
```
vpapavas, line 98: Add all the suppressed exceptions to the message exchanged between hosts but also to the one returned to the user

ksqldb-streams/src/main/java/io/confluent/ksql/execution/streams/materialization/KsqlMaterialization.java
```diff
@@ -128,148 +131,157 @@ public MaterializedWindowedTable windowed() {
...
-      }
+    public KsqlMaterializedQueryResult<Row> get(
+        final GenericKey key,
+        final int partition,
+        final Optional<ConsistencyOffsetVector> consistencyVector
+    ) {
+      final Optional<Position> position = consistencyVector
+          .map(offsetVector -> Position.fromMap(offsetVector.getOffsetVector()));
+      final KsMaterializedQueryResult<Row> result = table.get(key, partition, position);
+      updateConsistencyVector(consistencyVector, result.getPosition());
```
vpapavas, line 160: Update the consistency token with the position we get from streams after executing the query

ksqldb-streams/src/main/java/io/confluent/ksql/execution/streams/materialization/StreamsMaterialization.java
```diff
@@ -0,0 +1,48 @@
...
+ * specific language governing permissions and limitations under the License.
+ */
+
+package io.confluent.ksql.execution.streams.materialization;
+
+import io.confluent.ksql.model.WindowType;
+import io.confluent.ksql.schema.ksql.LogicalSchema;
+import java.util.Optional;
+
+public interface StreamsMaterialization {
```
vpapavas, line 22: I made the distinction between Streams and Ksql materialization. The former uses the `Position` whereas the latter uses the `ConsistencyVector`

ksqldb-streams/src/main/java/io/confluent/ksql/execution/streams/materialization/ks/KsLocator.java
```diff
@@ -140,6 +141,11 @@
     return partitionLocations.build();
   }
 
+  @SuppressFBWarnings(value = "EI_EXPOSE_REP")
```
vpapavas, line 144: Added this to throw more meaningful exceptions which are also helpful for debugging when there are multiple hosts

ksqldb-rest-client/src/main/java/io/confluent/ksql/rest/client/KsqlRestClient.java
```diff
@@ -373,4 +359,18 @@ private static URI parseUri(final String serverAddress) {
           "The supplied serverAddress is invalid: " + serverAddress, e);
     }
   }
+
+  private Map<String, Object> setConsistencyVector(
+      final Map<String, ?> requestProperties
+  ) {
+    final Map<String, Object> requestPropertiesToSend = new HashMap<>(requestProperties);
+    if (ConsistencyOffsetVector.isConsistencyVectorEnabled(requestPropertiesToSend)) {
```
patrickstuedi, line 367: This method is solely operating on the overrides map and the KsqlConfig, should we move it to Utils?

ksqldb-streams/src/main/java/io/confluent/ksql/execution/streams/materialization/KsqlMaterialization.java
patrickstuedi: Is the "Optional<ConsistencyOffsetVector> consistencyVector" also used an output param (input anyway)? If so why, don't we pass the consistency vector as part of the iterator?

ksqldb-streams/src/main/java/io/confluent/ksql/execution/streams/materialization/StreamsMaterialization.java
patrickstuedi: not suggesting any changes, but this in my mind an indication  that eventually we should drop ConsistencyVector and use Position(Bound). Keeping these different Materialization classes seems just to blow up the code unnecessarily. Or is there any other reason why we need these different materializtion tables for streams and ksql?

patrickstuedi (commented):
Thanks @vpapavas ! Added some comments. One high level comment is that this PR is really huge which makes it a bit difficult to review. In the future would be good to break stuff like this down into multiple PRs.

ksqldb-streams/src/main/java/io/confluent/ksql/execution/streams/materialization/KsqlMaterialization.java
vpapavas: I don't understand the question :(

ksqldb-streams/src/main/java/io/confluent/ksql/execution/streams/materialization/StreamsMaterialization.java
vpapavas: The materialization classes do different things. The streams materialization does the streams query, the ksql materialization applies the transformers (having, merge udaf) on the value returned from streams

ksqldb-rest-client/src/main/java/io/confluent/ksql/rest/client/KsqlRestClient.java
vpapavas: It's only used in this class, I don't think this warrants moving it to a Utils class, which btw, I don't see exists in this package

vpapavas:
Thank you @patrickstuedi for the review! I know the PR is huge, in hindsight I should have split it into two. The main comments are about the use of Optionals. I don't see anything wrong with it. If I were to add default methods that don't use them, I would end up replicating a lot of code which is error-prone. Also, the optionals are mainly for the consistency vector which once the feature is stable, will always be there and won't be an Optional anymore

ksqldb-rest-app/src/test/java/io/confluent/ksql/rest/integration/PullQueryConsistencyFunctionalTest.java
```diff
@@ -294,57 +305,242 @@ public void cleanUp() {
...
+
+    waitForRemoteServerToChangeStatus(
+        clusterFormation.router.getApp(),
+        clusterFormation.router.getHost(),
+        HighAvailabilityTestUtil.lagsReported(clusterFormation.standBy.getHost(), Optional.of(5L), 5),
+        USER_CREDS);
+
+    // Cut off standby from Kafka to simulate lag
+    clusterFormation.standBy.getShutoffs().setKafkaPauseOffset(0);
+    Thread.sleep(2000);
```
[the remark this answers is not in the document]
vpapavas: This is the technique we follow in PullQueryRouting test. On top of my head, I am guessing this is needed to ensure the cutoff actually happened since the cutoff essentially just pauses the consumer

ksqldb-engine/src/test/java/io/confluent/ksql/materialization/ks/KsMaterializationFunctionalTest.java
```diff
@@ -318,7 +318,7 @@ public void shouldQueryMaterializedTableForTumblingWindowed() {
 
       final List<WindowedRow> resultAtWindowStart =
           withRetry(() -> Lists.newArrayList(
-              table.get(key, PARTITION, Range.singleton(w.start()), Range.all()).getRowIterator()));
+              table.get(key, PARTITION, Range.singleton(w.start()), Range.all(), Optional.empty()).getRowIterator()));
```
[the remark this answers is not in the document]
vpapavas: Addressed this

ksqldb-rest-app/src/test/java/io/confluent/ksql/rest/integration/RestApiTest.java
```diff
@@ -1218,4 +1339,12 @@ private static void assertExpectedScalablePushQueries(
       return true;
     }, is(true));
   }
+
+  private static void verifyConsistencyVector(
+      final String consistencyText, final ConsistencyOffsetVector consistencyOffsetVector) {
+    String serializedCV = consistencyText.split(":\"")[1];
+    serializedCV = serializedCV.substring(0, serializedCV.length()-4);
```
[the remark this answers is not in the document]
vpapavas: Added a comment

patrickstuedi (approved):
LGTM, thanks @vpapavas

ksqldb-streams/src/main/java/io/confluent/ksql/execution/streams/materialization/KsqlMaterialization.java
```diff
@@ -128,148 +131,157 @@ public MaterializedWindowedTable windowed() {
...
-
-      final Iterator<Row> iterator = Streams.stream(result.getRowIterator())
-          .map(row -> filterAndTransform(row.key(), getIntermediateRow(row), row.rowTime())
-              .map(v -> row.withValue(v, schema())))
-          .filter(Optional::isPresent)
-          .map(Optional::get)
-          .iterator();
+    public KsqlMaterializedQueryResult<Row> get(
+        final int partition,
+        final Optional<ConsistencyOffsetVector> consistencyVector
```
agavra, line 178: I'm wondering (this can be done in a separate PR) if instead of using optional everywhere if it was possible to use a no-op consistency vector. tends to make the code a little easier to read and encapsulation a little clearer but definitely not a big deal

then we could make methods like `updateConsistencyVector` part of the no-op class' implementation of `updateFromPosition` (it just does nothing)

ksqldb-streams/src/main/java/io/confluent/ksql/execution/streams/materialization/KsqlMaterialization.java
```diff
@@ -128,148 +131,157 @@ public MaterializedWindowedTable windowed() {
...
+    ) {
+      final Optional<Position> position = consistencyVector
+          .map(offsetVector -> Position.fromMap(offsetVector.getOffsetVector()));
+      final KsMaterializedQueryResult<Row> result = table.get(partition, position);
```
agavra, line 182: to make sure i'm reviewing the pr correctly, this (and the same part below) is the "meat" of it, right? passing in `position` to the underlying table?

(can be done in a different PR) do we need ot have ConsistencyVector now that we have Position?

ksqldb-streams/src/main/java/io/confluent/ksql/execution/streams/materialization/ks/KsqlMaterializedQueryResult.java
```diff
@@ -0,0 +1,41 @@
...
+
+package io.confluent.ksql.execution.streams.materialization.ks;
+
+import edu.umd.cs.findbugs.annotations.SuppressFBWarnings;
+import io.confluent.ksql.execution.streams.materialization.TableRow;
+import java.util.Iterator;
+import java.util.Objects;
+
+@SuppressFBWarnings(value = "EI_EXPOSE_REP")
+public final class KsqlMaterializedQueryResult<T extends TableRow> {
```
agavra, line 24: why do we need a class for this? also can `rowIterator` be `private final`?

agavra (commented):
I saw that @patrickstuedi reviewed and approved so I'm only partially reviewing. Mostly LGTM

ksqldb-streams/src/main/java/io/confluent/ksql/execution/streams/materialization/KsqlMaterialization.java
vpapavas: Yes, this is definitely something we were thinking about. Patrick mentioned it as well. It will make the code cleaner for sure

ksqldb-streams/src/main/java/io/confluent/ksql/execution/streams/materialization/KsqlMaterialization.java
vpapavas: We decided to keep both for now among other reasons so that we can make changes to the ConsistencyVector faster than what would be possible with streams.

ksqldb-engine/src/main/java/io/confluent/ksql/physical/pull/HARouting.java
```diff
@@ -361,24 +378,23 @@ private static void forwardTo(
...
       // If we get some kind of unknown error, we assume it's network or other error from the
       // KsqlClient and try standbys
+      final String exceptionMessage = ksqlException == null ? e.getMessage() :
+          ksqlException.getMessage();
       throw new StandbyFallbackException(String.format(
-          "Forwarding pull query request [%s, %s] failed with error %s ",
-          statement.getSessionConfig().getOverrides(), requestProperties,
-          e.getMessage()), e);
+          "Forwarding pull query request failed with error %s ", exceptionMessage), e);
```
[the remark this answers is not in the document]
vpapavas: The exception messages was really really long and I don't think they provided any useful information

ksqldb-rest-app/src/main/java/io/confluent/ksql/api/server/QuerySubscriber.java
```diff
@@ -93,9 +94,19 @@ private void checkMakeRequest() {
...
+    for (Throwable s: t.getSuppressed()) {
+      if (s instanceof NotUpToBoundException) {
+        stringBuilder.append(" Failed to get value from materialized table, reason: "
```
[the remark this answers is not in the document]
vpapavas: The reason for this is that the message of the exception contains details about the position (topic, partition, offset). It is my understanding that this is the kind of detail we want to hide from users for security reasons, cc @vvcephei

ksqldb-rest-app/src/main/java/io/confluent/ksql/api/server/QuerySubscriber.java
```diff
@@ -93,9 +94,19 @@ private void checkMakeRequest() {
...
+                                 + "NOT_UP_TO_BOUND");
+      } else {
+        stringBuilder.append(s.getMessage());
+      }
+    }
+    final KsqlErrorMessage errorResponse = new KsqlErrorMessage(
+        ERROR_CODE_SERVER_ERROR, stringBuilder.toString());
+    log.error("Error in processing query {}", stringBuilder);
```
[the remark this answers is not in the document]
vpapavas: Hmm, I am conflicted about this. I thought it would be helpful for the user to know what nodes we tried and why they failed so that they can "debug" or at least understand why routing failed. I feel we get a lot of questions/complaints about pull queries failing on standbys and the users could not understand why.

PR #8674, opened 2022-01-28
fix: use topology-specific IQ APIs

jnh5y:
This is a rebase of https://github.com/confluentinc/ksql/pull/8579 to resolve merge conflicts and get a green build!


wcarlson5:
@jnh5y its hard to tell from the output you give, but I don't see that you have updated `KsStateStoreTest`. This issue was that the type we were checking was specify but the test class had use a parent type so they didn't match

guozhangwang (approved):

PR #8677, opened 2022-01-28
chore: Using the native Git client on CI servers.

jnh5y:


Jenkinsfile
```diff
@@ -11,5 +11,6 @@ dockerfile {
     nanoVersion = true
     maxBuildsToKeep = 99
     maxDaysToKeep = 90
+    extraBuildArgs = "-Dmaven.gitcommitid.nativegit=true"
```
jnh5y, line 14:
```suggestion
    extraBuildArgs = "-Dmaven.gitcommitid.nativegit=true -T4"
```

PR #8681, opened 2022-01-28
chore: WIP: try build with -T3

jnh5y:
Trying to speed up the CI build!


spena:
How much time did it save?

jnh5y:
This saved something like 10-15 minutes.  Using the native git client is a separate savings.  I'm not quite done / sure what to suggest yet.  I'd like to keep tinkering with it.

I think the *big* thing to do would be to think about the functional test suite.  That'll likely require more thought!

PR #8682, opened 2022-01-28
feat: limit set to query level configs with shared runtimes enabled

wcarlson5:
### Description 
Limits the set cmd to only be able to set query level config when the shared runtimes are enabled. It directs all other config set attempts to use `ALTER SYSTEM` instead.

### Testing done 
unit testing

### Reviewer checklist

### DIFF
--- a/ksqldb-engine/src/main/java/io/confluent/ksql/properties/PropertyOverrider.java
+++ b/ksqldb-engine/src/main/java/io/confluent/ksql/properties/PropertyOverrider.java
@@ -24,7 +24,6 @@ import io.confluent.ksql.statement.ConfiguredStatement;
 import io.confluent.ksql.util.KsqlConfig;
 import io.confluent.ksql.util.KsqlException;
 import io.confluent.ksql.util.KsqlStatementException;
-
 import java.util.Arrays;
 import java.util.HashSet;
 import java.util.Map;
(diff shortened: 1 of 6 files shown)


wcarlson5:
these test failures are the fk join tests with recent issues and are not related to this PR

ksqldb-engine/src/main/java/io/confluent/ksql/properties/PropertyOverrider.java
```diff
@@ -21,11 +21,26 @@
...
 
+  private static final Set<String> QUERY_LEVEL_CONFIGS = new HashSet<>(Arrays.asList(
+      "auto.offset.reset",
+      "buffered.records.per.partition",
+      "cache.max.bytes.buffering",
+      "default.deserialization.exception.handler",
+      "default.timestamp.extractor",
+      "max.task.idle.ms",
+      "task.timeout.ms",
+      "topology.optimization"));
```
ableegoldman, line 42: Do we not already have a list of query-level configs maintained somewhere within the ksql repo? Or is that only hard-coded into the front-end?

IIRC from @tolgadur's work on the new `list properties` fields, we do have something like this that explicitly marks the query-level configs -- we should make sure to reuse that, or consolidate the two in some other way, so we don't have to worry about maintaining  multiple copies and keeping them all up to date

ksqldb-engine/src/main/java/io/confluent/ksql/properties/PropertyOverrider.java
```diff
@@ -60,6 +75,18 @@ private static void throwIfInvalidPropertyValues(
               setProperty.getPropertyName(),
               setProperty.getPropertyValue()
           ));
+      if (statement
+          .getSessionConfig()
+          .getConfig(true)
+          .getBoolean(KsqlConfig.KSQL_SHARED_RUNTIME_ENABLED)
+          && !QUERY_LEVEL_CONFIGS.contains(setProperty.getPropertyName())) {
+        throw new KsqlException(String.format("%s is not a settable property at the query"
+                + " level with %s on. Please use ALTER SYSTEM to set %s for the cluster.",
```
ableegoldman, line 84: `KsqlConfig.KSQL_SHARED_RUNTIME_ENABLED` is technically an internal config that users shouldn't be exposed to or really even aware of -- should rephrase to something more along the lines of 


```suggestion
        When shared runtimes are enabled, the config %s can only be set for the entire cluster and all queries currently running in it, and not configurable for individual queries for example by using `SET`.  Please use...
```

or something

ksqldb-engine/src/main/java/io/confluent/ksql/properties/PropertyOverrider.java
wcarlson5: ah I didn't see that I will reuse it instead

wcarlson5:
Im going to change how I approach this so I am just going to open a clean PR

PR #8685, opened 2022-01-30
fix: fix create connector config handling for name config

vcrfxia:
### Description 

This PR contains a number of miscellaneous fixes/changes to ksqlDB's connector management functionality:
* fixes a bug where the `name` connector config parameter was automatically populated when validating the config for CREATE CONNECTOR statements, but not when sending the actual create request
* no longer automatically sets `key.converter` connector config parameter to the string converter -- this should've been removed a while back when ksqlDB expanded key type support to include types other than strings
* makes the connector request timeout configurable as the current hard-coded default (5 seconds) is insufficient for  certain environments where some requests (particularly, CREATE CONNECTOR requests) may take longer if more validations are performed
* updates `DefaultConnectClient#delete()` to accept 200 status codes in addition to 204s, as 200s are returned from Confluent Cloud (while 204s are still returned from other Confluent Connect deployments).

BREAKING CHANGE: When creating connectors through ksqlDB, ksqlDB no longer automatically sets the `key.converter` config to be the StringConverter, as ksqlDB has supported key types other than strings for many releases now.

### Testing done 

Unit + manual.

### Reviewer checklist


ksqldb-engine/src/main/java/io/confluent/ksql/connect/supported/Connectors.java
```diff
@@ -58,15 +57,9 @@
 
   public static Map<String, String> resolve(final Map<String, String> configs) {
     final SupportedConnector connector = CONNECTORS.get(configs.get(CONNECTOR_CLASS));
-    final Map<String, String> resolvedConfigs = new HashMap<>(
+    return new HashMap<>(
         connector == null ? configs : connector.resolveConfigs(configs)
     );
-
-    // at the moment, KSQL only supports String keys - when we
-    // support non-string keys we should remove this line
```
vcrfxia, line 66: @agavra I've removed this line on the basis of this comment, as a minor breaking change. Let me know if this isn't what you had in mind for deprecation of this convenience feature. Thanks!

ksqldb-engine/src/main/java/io/confluent/ksql/services/DefaultConnectClient.java
```diff
@@ -76,7 +78,7 @@
   private static final String STATUS = "/status";
   private static final String TOPICS = "/topics";
   private static final String VALIDATE_CONNECTOR = CONNECTOR_PLUGINS + "/%s/config/validate";
-  private static final int DEFAULT_TIMEOUT_MS = 5_000;
+  private static final int DEFAULT_TIMEOUT_MS = 30_000;
```
vcrfxia, line 81: I considered making this value configurable, if we think defaulting to 30 seconds in order to accommodate certain high latency environments is too extreme. OTOH, we already have so many server configs, maybe it doesn't make sense to add another. I could go either way on this one, might be leaning towards adding a config.

nateab (approved):
LGTM!

ksqldb-engine/src/main/java/io/confluent/ksql/services/DefaultConnectClient.java
vcrfxia: Update: made the value configurable and set the default back to 5 ms. Users with in high latency environments can increase their timeouts via the new config.

PR #8692, opened 2022-02-01
build: use pinned cp-base-new image

nateab:
For this PR in order to simplify the release pipeline we just use the `cp-base-new` image on confluent dockerhub pinned to some tag, right now pinned to 7.0.1 and we will update this as needed.

Previously `cp-base-new` beta tagged images were created along with beta packaging. And we would use these beta tagged images as our base, but Andrew Egolhofer (confluent packaging and release manager) has pointed out several times that we don't actually need these beta tagged images, amd that we can just use the `cp-base-new` image on confluent dockerhub. 

1) https://confluent.slack.com/archives/D013VSG3YQP/p1643676941852899 
2) https://confluent.slack.com/archives/C01KH8SH939/p1638227636063900?thread_ts=1638216050.063300&cid=C01KH8SH939

I did this manually for the 0.23.1 release and it worked well since we werent able to get the tooling to create the stabilization tagged images in the same convoluted manner as the beta tagged images.


vcrfxia:
Is this reproducible for hotfixes? As in, if we cut a hotfix a few weeks after the main release, will the hotfix pick up the same base image as the original release?

swist (approved):

nateab:
That is a valid point Victoria, Tom and I worked on this offline together and have a different solution that will hopefully work instead

nateab:
@vcrfxia now that we are pinning the base image to a specific version like 7.0.1 (which we may have to update once in a while), hotfixes should be reproducible

agavra (approved):

swist (approved):

swist:
green light to just hit merge 👍

PR #8694, opened 2022-02-01
feat: cleanup transient query resources

cprasad1:
### Description 
Primarily introduces `TransientQueryCleanupListener` and `TransientQueryCleanupService`.
- `TransientQueryCleanupService` periodically searched for leaked topics and state directories, and cleans them. 
It achieves that by scanning all transient query topics + state directories and checking if the query associated with them has been terminated. 

This PR also introduces 3 configs:
- `ksql.transient.query.cleanup.service.enable`: Enables transient query cleanup service. Set to `true` by default.
- `ksql.transient.query.cleanup.service.initial.delay.seconds`: The time to delay the first execution of the transient query cleanup service in seconds. Set to 600 seconds by default.
- `ksql.transient.query.cleanup.service.period.seconds`: The period between successive executions of the transient query cleanup service. Set to 600 seconds by default.

### Testing done 
- Manual Testing
- Unit Tests
- Integration tests in `TransientQueryResourceCleanerIntTest`

### Reviewer checklist


ksqldb-rest-app/src/test/java/io/confluent/ksql/rest/integration/TransientQueryResourceCleanerIntTest.java
```diff
@@ -0,0 +1,351 @@
...
+        assertTrue(transientTopics.get(1).contains(transientQueryId));
+
+        // terminate the transient query
+        RestIntegrationTestUtil.makeKsqlRequest(
+                REST_APP_0,
+                "terminate " + transientQueryId + ";"
+        );
+
+        // Give some time for the natural onClose() cleanup to take effect
+        Thread.sleep(1000);
```
vvcephei, line 186: There's no guarantee this will be long enough. We should either hook into the lifecycle of the query or at least poll until it's been cleaned up before proceeding with the test.

Experience shows that sooner or later, this is going to start taking slightly longer than a second to clean up in CI, and this will become a flaky test.

vvcephei (commented):
Thanks, @cprasad1 !

I left some inline comments. One high level question: it looks like this is only going to clean up leaked topics for the currently running host, right? Specifically, if we restart the server, and there are topics or files from before the restart, is it correct that they would not be cleaned up?

cprasad1:
@vvcephei 
>if we restart the server, and there are topics or files from before the restart, is it correct that they would not be cleaned up?
They won't get cleaned up with this mechanism, but we have the `LocalCommands` file for cleaning those leaked resources up. It's a separate question if we want to rely on the `LocalCommands` file

ksqldb-engine/src/main/java/io/confluent/ksql/engine/TransientQueryCleanupService.java
```diff
@@ -0,0 +1,276 @@
...
+    public Boolean call() {
+      try {
+        LOG.info("Deleting topics for transient query: {}", appId);
+        serviceContext.getTopicClient().deleteInternalTopics(appId);
+      } catch (final Exception e) {
+        LOG.warn("Failed to cleanup topics for transient query: {}", appId, e);
+        return false;
+      }
+
+      final Set<String> orphanedTopics = serviceContext
```
AlanConfluent, line 216: Maybe it's not a big deal in practice, but each time a query is completed, it will do this call to Kafka to list all of the topics.  Also, deleteInternalTopics lists topics as well.  If you have a bunch of queued up Tasks, they each will do the identical call, at all once.  In practice we don't do many push queries, but you can imagine that you could queue up 300 of these in a ten minute period.

It might make sense to use the batch delete and only call listTopics once.

Anyway, we currently use this approach for `QueryCleanupService`, so it's probably fine, but maybe something to keep in mind.

ksqldb-engine/src/main/java/io/confluent/ksql/internal/TransientQueryCleanupListener.java
```diff
@@ -0,0 +1,89 @@
...
+  @Override
+  public void onClose(
+          final QueryMetadata query
+  ) {
+    if (query instanceof TransientQueryMetadata) {
+      final String applicationId = query.getQueryApplicationId();
+      if (query.hasEverBeenStarted()) {
+        log.info("Cleaning up after query {}", applicationId);
+
+        final TransientQueryTopicCleanupTask topicTask = new TransientQueryTopicCleanupTask(
```
AlanConfluent, line 72: This is also done by `QueryCleanupService` in a similar way.  I'm just wondering if we need to do both attempts.  I guess these have the benefit of trying 5 times, but it also does the `findPossiblyLeakedTranientTopics` which also attempts to clean things up based on the current state of what's actually there, which is much more robust in a sense since we don't know if failures appear as failures.  I just wondering, if maybe we need all three types of attempts.  It might not hurt but it's a bit intense.

ksqldb-engine/src/main/java/io/confluent/ksql/engine/TransientQueryCleanupService.java
```diff
@@ -0,0 +1,276 @@
...
+  private static final Logger LOG = LoggerFactory.getLogger(TransientQueryCleanupService.class);
+  private static final Pattern TRANSIENT_PATTERN =
+          Pattern.compile("(?i).*transient_.*_[0-9]\\d*_[0-9]\\d*");
+  private static final int NUM_OF_RETRIES_TO_DELETE_RESOURCE = 5;
+
+  private final BlockingQueue<Callable<Boolean>> cleanupTasks;
+  private final Retryer<Boolean> retryer;
+  private final String stateDir;
+  private final ServiceContext serviceContext;
+  private final Set<TransientQueryMetadata> allTransientQueriesEverCreated;
```
AlanConfluent, line 57: We probably shouldn't keep a reference to the metadata directly since we'll never GC any of these objects or the streams topologies they point to, so it's essentially a memory leak.  Doing that for a set of Strings is maybe tolerable in practice, but we probably should't do this.

ksqldb-engine/src/main/java/io/confluent/ksql/engine/TransientQueryCleanupService.java
```diff
@@ -0,0 +1,276 @@
...
+                appId);
+        return false;
+      } else {
+        return true;
+      }
+    }
+  }
+
+  private boolean isTerminatedQuery(final String queryId) {
+    return !allTransientQueriesEverCreated
```
AlanConfluent, line 264: We can always grab the metadata from the query registry using the id to avoid the memory issues.  If it's not there, it's safe to assume that the query has been stopped.

ksqldb-engine/src/main/java/io/confluent/ksql/engine/TransientQueryCleanupService.java
```diff
@@ -0,0 +1,276 @@
...
+    LOG.info("Cleaning up {} leaked topics.", numLeakedTopics);
+    LOG.info("Cleaning up {} leaked state directories.", numLeakedStateDirs);
+    try {
+      while (!cleanupTasks.isEmpty()) {
+        final Callable<Boolean> task = cleanupTasks.take();
+
+        try {
+          retryer.call(task);
+        } catch (RetryException | ExecutionException e) {
+          cleanupTasks.add(task);
```
AlanConfluent, line 115: If you fail for a given task, it doesn't seem that there's any backoff, so it'll just hammer Kafka 5 times in a row and then be done.  If you used a PriorityQueue and had time to run, you could peek at the top task and sleep until it was ready.  Then you could make each task adjustable and have X seconds between attempts.

Another simpler approach is that at each call to `runOneIteration`, you can go through everything once, but if you get to a task that failed in this round, you declare yourself done and wait until the next iteration.

ksqldb-engine/src/main/java/io/confluent/ksql/engine/TransientQueryCleanupService.java
```diff
@@ -0,0 +1,276 @@
...
+      }
+    } catch (final InterruptedException e) {
+      // gracefully exit if this method was interrupted and reset
+      // the interrupt flag
+      Thread.currentThread().interrupt();
+    }
+  }
+
+  public Scheduler scheduler() {
+    return Scheduler.newFixedRateSchedule(initialDelay, intervalPeriod, TimeUnit.SECONDS);
```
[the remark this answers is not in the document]
cprasad1: from the docs of `Scheduler.newFixedRateSchedule`: "If any execution of this task takes longer than its period, then subsequent executions may start late, but will not concurrently execute." https://docs.oracle.com/javase/7/docs/api/java/util/concurrent/ScheduledExecutorService.html

It already guards against that, so it should be fine.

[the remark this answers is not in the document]
vvcephei: Great! Thanks for checking.

ksqldb-engine/src/main/java/io/confluent/ksql/engine/TransientQueryCleanupService.java
cprasad1: changed. Upon failure, it doesn't retry but instead, queues the task back again

ksqldb-engine/src/main/java/io/confluent/ksql/engine/TransientQueryCleanupService.java
```diff
@@ -0,0 +1,276 @@
...
+            .count();
+
+    LOG.info("Cleaning up {} leaked topics.", numLeakedTopics);
+    LOG.info("Cleaning up {} leaked state directories.", numLeakedStateDirs);
+    try {
+      while (!cleanupTasks.isEmpty()) {
+        final Callable<Boolean> task = cleanupTasks.take();
+
+        try {
+          retryer.call(task);
```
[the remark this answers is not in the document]
cprasad1: changed. Upon failure, it doesn't retry but instead, queues the task back again.

I think if a deleteion task fails at a moment, then it is less probably to fail after ~10 minutes as compared to immediately. Not the most scientific reasoning but a good enough heuristic

ksqldb-engine/src/main/java/io/confluent/ksql/engine/TransientQueryCleanupService.java
cprasad1: I think if a deleteion task fails at a moment, then it is less probably to fail after ~10 minutes as compared to immediately. Not the most scientific reasoning but a good enough heuristic

ksqldb-engine/src/main/java/io/confluent/ksql/engine/TransientQueryCleanupService.java
cprasad1: good idea. Changed to query the `QueryRegistry` to check the status of the query (running or not) instead of maintaining a set of all the queries ever created.

ksqldb-engine/src/main/java/io/confluent/ksql/engine/TransientQueryCleanupService.java
cprasad1: good idea. Changed to query the QueryRegistry to check the status of the query (running or not) instead of maintaining a set of all the queries ever created. This will avoid the memory leak

ksqldb-engine/src/main/java/io/confluent/ksql/internal/TransientQueryCleanupListener.java
cprasad1: Let's keep the more intense version. Since the exact cause of the leaks is unknown, it doesn't hurt to be more intense here

ksqldb-engine/src/main/java/io/confluent/ksql/engine/TransientQueryCleanupService.java
```diff
@@ -0,0 +1,276 @@
...
+        } catch (RetryException | ExecutionException e) {
```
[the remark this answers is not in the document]
cprasad1: removed

ksqldb-rest-app/src/test/java/io/confluent/ksql/rest/integration/TransientQueryResourceCleanerIntTest.java
cprasad1: chnaged to check for precondition

cprasad1:
@vvcephei @AlanConfluent ready for review again. I have addressed the comments and also made a few changes:
1) The cleanup service now checks the `QueryRegistry` directly to determine if a query has be terminated or not instead of maintaining an in memory set of all the Transient queries ever created to avoid memory leak.
2) Added queueing up of `LocalComands` app IDs upon server start to the cleanup service.
3) Upon failure of a cleanup task it doesn't retry but instead queues the task back again.
4) Added logging test to check that the appropriate message has been logged to the integration test
5) Integration test now checks for precondition before proceeding instead of `Thread.sleep`

ksqldb-engine/src/main/java/io/confluent/ksql/engine/TransientQueryCleanupService.java
```diff
@@ -0,0 +1,312 @@
...
+    LOG.info("Cleaning up {} leaked topics.", numLeakedTopics);
+    cleanupLeakedTopics();
+
+    LOG.info("Cleaning up {} leaked state directories.", numLeakedStateDirs);
+    cleanupLeakedStateDirs();
+  }
+
+  @Override
+  public Scheduler scheduler() {
+    return Scheduler.newFixedRateSchedule(initialDelay, intervalPeriod, TimeUnit.SECONDS);
```
vvcephei, line 119: Sorry, one more quick question about this: What's the behavior if an unhandled exception kills this thread? Does the scheduler just create a new one the next time around? In that case, do we have any concerns about tasks left behind in our task lists from the prior, incomplete run?

ksqldb-engine/src/main/java/io/confluent/ksql/engine/TransientQueryCleanupService.java
```diff
@@ -0,0 +1,312 @@
...
+        addTopicCleanupTask(new TransientQueryTopicCleanupTask(serviceContext, id));
+        addStateCleanupTask(new TransientQueryStateCleanupTask(id, stateDir));
+      });
+      isLocalCommandsProcessed = true;
+    }
+
+    LOG.info("Starting cleanup for leaked resources.");
+
+    topicsCleanupTasks.addAll(findPossiblyLeakedTransientTopics());
+    stateDirsCleanupTasks.addAll(findPossiblyLeakedStateDirs());
```
vvcephei, line 105: So we enqueue all possible tasks here, and later, if we fail on a task we also re-enqueue it. So, on the next run, the same task will be in the queue twice. I don't think that's a problem, but on the other hand, since we have these two lines here, it doesn't seem like we strictly need those queues to persist between runs. We could also just freshly discover all potentially leaked resources from scratch every time, and if anything fails, oh well, we'll discover it again next time.

ksqldb-engine/src/main/java/io/confluent/ksql/engine/TransientQueryCleanupService.java
```diff
@@ -0,0 +1,312 @@
...
+          }
+        } catch (Exception e) {
+          topicsCleanupTasks.add(task);
+        }
+      }
+    } catch (final InterruptedException e) {
+      // gracefully exit if this method was interrupted and reset
+      // the interrupt flag
+      Thread.currentThread().interrupt();
+    }
```
vvcephei, line 140: It does seem like a bummer that we don't just collect all the topic names and send a single delete-topics request (Alan also pointed this out below).

I'm fine if you don't want to refactor it; just calling this out.

vvcephei: Also, even though the internal KSQL API might be synchonous, the actual Admin client call is async. So you could just fire off the async request and immediately move on to cleaning up the state directories. Then, after the state directory deletion is done, you can wait on the Future from the delete-topics so that you can actually log whether it succeeded or failed, and also to prevent stacking up async calls on the brokers.

ksqldb-engine/src/main/java/io/confluent/ksql/engine/TransientQueryCleanupService.java
```diff
@@ -0,0 +1,312 @@
...
+    return leakedQueries;
+  }
+
+  private boolean isCorrespondingQueryRunning(final String topic) {
+    return this.queryRegistry
+            .getAllLiveQueries()
+            .stream()
+            .map(qm -> qm.getQueryId().toString())
+            .anyMatch(topic::contains);
+  }
```
vvcephei, line 241: Race-condition gut-check here: is there any possibility that we could discover a resource for cleanup _before_ it becomes "running" according to this check and wind up deleting a resource out from under a recently-started query?

ksqldb-engine/src/main/java/io/confluent/ksql/engine/TransientQueryCleanupService.java
```diff
@@ -0,0 +1,312 @@
...
+        final Matcher topicPrefixMatcher = LEAKED_TOPIC_PREFIX_PATTERN.matcher(topic);
+        if (topicPrefixMatcher.find()) {
+          final String leakedTopicPrefix = topicPrefixMatcher.group();
+          leakedQueries.add(new TransientQueryTopicCleanupTask(
+                  serviceContext,
+                  leakedTopicPrefix
+          ));
+        }
+      }
+    }
```
vvcephei, line 230: Quick check here: is this only going to match topics for transient queries created by _this_ node, or will we also see topics created by transient queries on other nodes? In the latter case, it seems like we'd delete the topic for running queries on other hosts, right?

ksqldb-engine/src/main/java/io/confluent/ksql/engine/TransientQueryCleanupService.java
```diff
@@ -0,0 +1,312 @@
...
+
+  private boolean isCorrespondingQueryRunning(final String topic) {
+    return this.queryRegistry
+            .getAllLiveQueries()
+            .stream()
+            .map(qm -> qm.getQueryId().toString())
+            .anyMatch(topic::contains);
+  }
+
+  public static class TransientQueryTopicCleanupTask implements Callable<Boolean> {
```
vvcephei, line 243: More on the nitpick side of things, but I think these Callables were necessary when you were using the Retryer. Now, it's just a silly layer of indirection to what could instead be a simple method call.

ksqldb-engine/src/main/java/io/confluent/ksql/internal/TransientQueryCleanupListener.java
```diff
@@ -0,0 +1,81 @@
...
+                applicationId
+        );
+
+        final TransientQueryStateCleanupTask stateTask = new TransientQueryStateCleanupTask(
+                applicationId,
+                stateDir
+        );
+
+        cleanupService.addTopicCleanupTask(topicTask);
+        cleanupService.addStateCleanupTask(stateTask);
```
vvcephei, line 75: Just belaboring a point here: following on my earlier point about idempotently discovering all potentially leaked resources in every iteration of runOnce, do we really need this listener at all? It seems like in the next run after closing the query, we'd already discover any potentially leaked resources AND see that the query isn't running, and therefore clean up anyway.

vvcephei (commented):
Thanks @cprasad1 ! 

I think this is really close. I'd be ok if you choose not to accept any of my suggestions, but I did have a couple of bug-check follow-ups you should do before merging.

ksqldb-engine/src/main/java/io/confluent/ksql/engine/TransientQueryCleanupService.java
```diff
@@ -0,0 +1,312 @@
...
+  }
+
+  @Override
+  public Scheduler scheduler() {
+    return Scheduler.newFixedRateSchedule(initialDelay, intervalPeriod, TimeUnit.SECONDS);
+  }
+
+  public void cleanupLeakedTopics() {
+    try {
+      while (!topicsCleanupTasks.isEmpty()) {
```
AlanConfluent, line 124: This won't exist the while loop until topicsCleanupTasks is empty, but what happens if you keep getting exceptions when running them? It looks like you're still enqueuing them again to immediately rerun.  That was my main comment last time.  I think you need two lists so that you let everything run once and anything that fails goes to the next round.

ksqldb-engine/src/main/java/io/confluent/ksql/engine/KsqlEngine.java
```diff
@@ -186,6 +195,11 @@ public KsqlEngine(
     this.metricCollectors = metricCollectors;
 
     cleanupService.startAsync();
+    if (ksqlConfig.getBoolean(KsqlConfig.KSQL_TRANSIENT_QUERY_CLEANUP_SERVICE_ENABLE)) {
+      this.transientQueryCleanupService
+              .setQueryRegistry(this.primaryContext.getQueryRegistry());
+      this.transientQueryCleanupService.startAsync();
+    }
```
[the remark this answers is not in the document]
cprasad1: good point. To stay safe, I am putting the instantiation, termination of the cleanup service and the registration of the cleanup service behind the feature flag. Committed changes.

ksqldb-engine/src/main/java/io/confluent/ksql/engine/KsqlEngine.java
```diff
@@ -556,6 +570,7 @@ public void close(final boolean closeQueries) {
 
   @Override
   public void close() {
+    transientQueryCleanupService.stopAsync();
```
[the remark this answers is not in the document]
cprasad1: good point. To stay safe, I am putting the instantiation, termination of the cleanup service and the registration of the cleanup service behind the feature flag. Committed changes.

ksqldb-engine/src/main/java/io/confluent/ksql/engine/TransientQueryCleanupService.java
cprasad1: good catch. Any exception escaping the run method halts all further work, without notice. from the docs: ```Run one iteration of the scheduled task. If any invocation of this method throws an exception, the service will transition to the Service.State.FAILED state and this method will no longer be called.```
added a try-catch block and logging the failure to safeguard against that.

ksqldb-engine/src/main/java/io/confluent/ksql/engine/TransientQueryCleanupService.java
cprasad1: >We could also just freshly discover all potentially leaked resources from scratch every time, and if anything fails, oh well, we'll discover it again next time.

I agree, the retry logic is unnecessarily complex. changed the logic to `just freshly discover all potentially leaked resources from scratch every time, and if anything fails, oh well, we'll discover it again next time.`

ksqldb-engine/src/main/java/io/confluent/ksql/engine/TransientQueryCleanupService.java
cprasad1: i agree. Refactored and changed send a single delete-topics batch request.

ksqldb-engine/src/main/java/io/confluent/ksql/engine/TransientQueryCleanupService.java
cprasad1: i agree they are now redundant and unnecessary. removed the callable interface implementations.

ksqldb-engine/src/main/java/io/confluent/ksql/internal/TransientQueryCleanupListener.java
cprasad1: done

ksqldb-engine/src/main/java/io/confluent/ksql/engine/TransientQueryCleanupService.java
cprasad1: there might be. To be safe, I am now recording the queries that have transitioned from rebalancing or running in the listener and then only deleting resources for the queries that have once transitioned from rebalancing or running.

ksqldb-engine/src/main/java/io/confluent/ksql/engine/TransientQueryCleanupService.java
cprasad1: updated to search for all leaked resources on every. If any deletion fasls then we just move on hoping that they would get discovered again in the next scheduled run

ksqldb-rest-app/src/test/java/io/confluent/ksql/rest/integration/TransientQueryResourceCleanerIntTest.java
```diff
@@ -0,0 +1,399 @@
...
+import static org.hamcrest.MatcherAssert.assertThat;
+import static org.hamcrest.Matchers.is;
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertFalse;
+import static org.junit.Assert.assertNotNull;
+import static org.junit.Assert.assertTrue;
+
+@Category({IntegrationTest.class})
+@RunWith(MockitoJUnitRunner.class)
+public class TransientQueryResourceCleanerIntTest {
```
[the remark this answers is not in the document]
cprasad1: added unit tests to very that the leaked resources are appropriated discovered on each run

cprasad1:
@vvcephei @AlanConfluent I changed the logic to `just freshly discover all potentially leaked resources from scratch every time, and if anything fails, oh well, we'll discover it again next time.` and added some unit tests.

ksqldb-engine/src/main/java/io/confluent/ksql/engine/KsqlEngine.java
```diff
@@ -186,6 +199,11 @@ public KsqlEngine(
     this.metricCollectors = metricCollectors;
 
     cleanupService.startAsync();
+    if (getTransientQueryCleanupServiceEnabled(ksqlConfig)) {
+      this.transientQueryCleanupService
```
AlanConfluent, line 203: It would be nice if you could set this in the constructor, but I see that there's a bit of a circular dependency.  Maybe just put a comment in the setter about how it's being used.

ksqldb-engine/src/main/java/io/confluent/ksql/engine/TransientQueryCleanupService.java
```diff
@@ -0,0 +1,210 @@
...
+    return Scheduler.newFixedRateSchedule(initialDelay, intervalPeriod, TimeUnit.SECONDS);
+  }
+
+  public void setQueryRegistry(final QueryRegistry queryRegistry) {
+    this.queryRegistry = queryRegistry;
+  }
+
+  @SuppressFBWarnings(value = "EI_EXPOSE_REP2")
+  public void setLocalCommandsQueryAppIds(final Set<String> ids) {
+    this.localCommandsQueryAppIds = Optional.of(ids);
```
AlanConfluent, line 122: Maybe leave a comment about how this is is set by the startup thread and therefore doesn't need synchronization since it's set before this service is started.

AlanConfluent (approved):
I definitely like this simplified version.  It's much easier to reason about.

ksqldb-engine/src/main/java/io/confluent/ksql/engine/KsqlEngine.java
cprasad1: will take care of it in the follow-up

ksqldb-engine/src/main/java/io/confluent/ksql/engine/TransientQueryCleanupService.java
cprasad1: will take care of it in the follow-up

PR #8695, opened 2022-02-01
feat: add shared-runtimes-flag-to-immutable-list

tolgadur:
### Description 
Users were able to override the shared runtimes fflag at the query level which might result in unwanted side effects. 

### Testing done 
Manual testing 

### Reviewer checklist

### DIFF
--- a/ksqldb-common/src/main/java/io/confluent/ksql/config/ImmutableProperties.java
+++ b/ksqldb-common/src/main/java/io/confluent/ksql/config/ImmutableProperties.java
@@ -35,6 +35,7 @@ public final class ImmutableProperties {
       .add(KsqlConfig.KSQL_READONLY_TOPICS_CONFIG)
       .add(KsqlConfig.KSQL_SOURCE_TABLE_MATERIALIZATION_ENABLED)
       .add(KsqlConfig.KSQL_HEADERS_COLUMNS_ENABLED)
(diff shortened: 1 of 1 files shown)


ksqldb-common/src/main/java/io/confluent/ksql/config/ImmutableProperties.java
```diff
@@ -35,6 +35,7 @@
       .add(KsqlConfig.KSQL_READONLY_TOPICS_CONFIG)
       .add(KsqlConfig.KSQL_SOURCE_TABLE_MATERIALIZATION_ENABLED)
       .add(KsqlConfig.KSQL_HEADERS_COLUMNS_ENABLED)
+      .add(KsqlConfig.KSQL_SHARED_RUNTIME_ENABLED)
```
wcarlson5, line 38: Will this affect setting via LD?

tolgadur: My understanding was this only affects the query level properties but I am not 100% sure. @rodesai do you know perhaps?

guozhangwang: My understanding is that LD is not via fflag but via external overrides e.g. incoming requests. If yes then they would not be affected.

PR #8704, opened 2022-02-02
fix: recreating old queries in the old runtime will still use the old runtimes

wcarlson5:
When a query using the old runtimes is recreated with the shared runtimes flag on then it should still use the shared runtimes.  This PR makes it so that the query will stay in the appropriate runtime

### Description 

### Testing done 

### Reviewer checklist

### DIFF
--- a/ksqldb-engine/src/main/java/io/confluent/ksql/query/QueryRegistryImpl.java
+++ b/ksqldb-engine/src/main/java/io/confluent/ksql/query/QueryRegistryImpl.java
@@ -292,7 +292,7 @@ public class QueryRegistryImpl implements QueryRegistry {
       );
     } else {
       query = queryBuilder.buildPersistentQueryInDedicatedRuntime(
-          oldQuery != null
+          oldQuery != null && sharedRuntimeId.isPresent()
               ? ksqlConfig.cloneWithPropertyOverwrite(oldQuery.getOverriddenProperties())
               : ksqlConfig,
           persistentQueryType,
--- a/ksqldb-engine/src/main/java/io/confluent/ksql/query/QueryRegistryImpl.java
+++ b/ksqldb-engine/src/main/java/io/confluent/ksql/query/QueryRegistryImpl.java
@@ -268,7 +268,9 @@ public class QueryRegistryImpl implements QueryRegistry {
 
     final PersistentQueryMetadata query;
 
-    if (sharedRuntimeId.isPresent()) {
+    final PersistentQueryMetadata oldQuery = persistentQueries.get(queryId);
+
+    if (sharedRuntimeId.isPresent() && (oldQuery == null || oldQuery instanceof BinPackedPersistentQueryMetadataImpl)) {
       if (sandbox) {
         streams.addAll(sourceStreams.stream()
             .map(SandboxedSharedKafkaStreamsRuntimeImpl::new)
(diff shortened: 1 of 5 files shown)


ksqldb-engine/src/main/java/io/confluent/ksql/query/QueryRegistryImpl.java
```diff
@@ -290,7 +292,9 @@ public PersistentQueryMetadata createOrReplacePersistentQuery(
       );
     } else {
       query = queryBuilder.buildPersistentQueryInDedicatedRuntime(
-          ksqlConfig,
+          oldQuery != null && sharedRuntimeId.isPresent()
```
jnh5y, line 295: nit: I might compute this config outside of the function call.  

As a separate question, is the logic around keeping and/or overriding the properties tested below?

ksqldb-engine/src/test/java/io/confluent/ksql/query/QueryRegistryImplTest.java
```diff
@@ -483,6 +499,18 @@ public void shouldCallSandboxOnClose() {
...
+  @Test
+  public void shouldReplaceQueryfromOldRuntimeUsingOldRuntime() {
+    sharedRuntimes = false;
+    QueryMetadata query = givenCreate(registry, "q1", "source",
+        Optional.of("sink1"), CREATE_AS);
+    assertThat("does not use old runtime", query instanceof PersistentQueryMetadataImpl);
+    sharedRuntimes = true;
+    query = givenCreate(registry, "q1", "source",
+        Optional.of("sink1"), CREATE_AS);
+    assertThat("does not use old runtime", query instanceof PersistentQueryMetadataImpl);
```
jnh5y, line 511: Is it worth adding to this test case what happens when adding a new query?  

E.g., once sharing is enabled, does the old query go to the old runtime because all queries would go there?

jnh5y:
@wcarlson5 I reviewed this PR briefly to follow along with the project.  It may be worth having someone else check it out as well.  That said, if you want me to approve it after we discuss things, I can do that as well!

ksqldb-engine/src/main/java/io/confluent/ksql/query/QueryRegistryImpl.java
wcarlson5: looks like we are still figuring out how we want to handle the configs. There has been more discussion since I made this change. I am just going to remove this logic and only have the location control in this PR and handle the config changes in a follow up once we are set on what we want to do

ksqldb-engine/src/test/java/io/confluent/ksql/query/QueryRegistryImplTest.java
wcarlson5: It shouldn't be necessary but I don't see any harm in adding it.

wcarlson5:
@jnh5y Thanks for the review! If you feel comfortable approving it that would be great. If not I can get someone else to give it a pass as well. If you have more comments let hear them :)

ksqldb-engine/src/test/java/io/confluent/ksql/query/QueryRegistryImplTest.java
jnh5y: Thanks!

jnh5y (approved):

jnh5y:
> @jnh5y Thanks for the review! If you feel comfortable approving it that would be great. If not I can get someone else to give it a pass as well. If you have more comments let's hear em :)

Walker, I've approved it!  Here's hoping we get a green master build soon and get things merged!

wcarlson5:
looks like it failed io.confluent.ksql.rest.integration.PullQueryRoutingFunctionalTest.shouldQueryActiveWhenActiveAliveStandbyDeadQueryIssuedToRouter

That is a an unrelated test. I will try rebasing

PR #8705, opened 2022-02-02
chore: Port over buffering for pull queries to /query-stream

AlanConfluent:
### Description 
Ports over the buffering behavior to `/query-stream` for pull queries that exists in `PullQueryStreamWriter`.  This ensures that each chunk is a good size rather than outputting one row per chunk.

### Testing done 
Unit and existing integration tests.

### Reviewer checklist


ksqldb-rest-app/src/main/java/io/confluent/ksql/api/server/JsonStreamedRowResponseWriter.java
```diff
@@ -170,12 +199,50 @@ private void writeLastRow(final boolean isLast) {
...
+      response.write(buff);
+    }
+  }
+
+  private void maybeFlushBuffer() {
+    if (writerState.length() > 0) {
+      if (writerState.length() >= FLUSH_SIZE_BYTES
+          || clock.millis() - writerState.getLastFlushMs() >= MAX_FLUSH_MS
+      ) {
+        response.write(writerState.getStringToFlush());
```
agavra, line 222: shouldn't we cancel/reset the timer if there's one set here?

agavra (commented):
quick pass

agavra (approved):
Thanks @AlanConfluent

PR #8711, opened 2022-02-03
feat: allow for only Gen 1 create or replace queries with server level configs via rest api, bars all gen 2 queries

wcarlson5:
This forces any queries coming into the shared runtimes to not have overridden the server level props. If the query will not be going into the new shared runtimes (i.e. is a `create or replace` of an old query) it will not be held to those standards.

### Testing done 

### Reviewer checklist

### DIFF
--- a/ksqldb-engine/src/main/java/io/confluent/ksql/query/QueryRegistryImpl.java
+++ b/ksqldb-engine/src/main/java/io/confluent/ksql/query/QueryRegistryImpl.java
@@ -28,6 +28,7 @@ import io.confluent.ksql.metastore.MetaStore;
 import io.confluent.ksql.metastore.model.DataSource;
 import io.confluent.ksql.metrics.MetricCollectors;
 import io.confluent.ksql.name.SourceName;
+import io.confluent.ksql.rest.entity.PropertiesList;
 import io.confluent.ksql.schema.ksql.LogicalSchema;
 import io.confluent.ksql.serde.WindowInfo;
 import io.confluent.ksql.services.ServiceContext;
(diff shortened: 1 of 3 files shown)


wcarlson5:
@rodesai This is the check we talked about that should allow for server level configs to be set for old queries during a create or replace. But it will make sure that any queries going into the new runtimes will fail with a useful message

wcarlson5:
When testing this PR I found an issue with recovering replaced gen 1 queries that I fixed here https://github.com/confluentinc/ksql/pull/8713

I will hold off merging until that is merged

guozhangwang (approved):
Hi @wcarlson5 the title and the description and the content of the PR seems to be indicating a bit different scopes: from the title and the content of the PR, we are only checking against the `create or replace` query stmt to not have any server level config overrides; whereas in your PR description you mentioned `This forces any queries coming into the shared runtimes..` could you modify it to be consistent and hence easy to understand fo other readers?

Otherwise, LGTM!

ksqldb-engine/src/main/java/io/confluent/ksql/query/QueryRegistryImpl.java
```diff
@@ -272,6 +273,7 @@ public PersistentQueryMetadata createOrReplacePersistentQuery(
 
     if (sharedRuntimeId.isPresent() && (oldQuery == null
         || oldQuery instanceof BinPackedPersistentQueryMetadataImpl)) {
+      throwOnNonQueryLevelConfigs(config.getOverrides());
```
rodesai, line 276: Sorry for the late feedback here. We should only do validation like this in the validation phase of executing a statement. Otherwise we might change the validation later on in a way that causes us to fail statements we've already accepted. The pattern for doing that is to do this type of validation in the sandbox instance only - can you make that change in a follow-up?

wcarlson5: https://github.com/confluentinc/ksql/pull/8722 sure

Those configs will be ignored but it won't mean there is a record in the cmd topic that will cause it to fail.

PR #8713, opened 2022-02-03
fix: when a isolated query is restarted then it should stay isolated

wcarlson5:
When we restart a query with `create or replace` stmt we would like to use the previous runtime that query was executed on. Today if we `create or replace` an existing query that's running on the old, non-shared runtime, it will still come up in a non shared runtime. However it's plan written to the cmd topic will indicate that it should use the new shared runtime.

To fix this, when executing the `create or replace` stmt we would only create the new query in the new runtime only if 1) a shared runtime id is assigned, 2) The queries config has the shared runtimes FF enabled and 3) the old query either does not exist or it has already been running on the shared runtime.

### Testing done 
Test to make sure the cmd gen'd is for a gen 1 query in the case it is being replaced. And that is doesn't happen to a gen 2 query.

### Reviewer checklist

### DIFF
--- a/ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/computation/ValidatedCommandFactory.java
+++ b/ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/computation/ValidatedCommandFactory.java
@@ -18,6 +18,7 @@ package io.confluent.ksql.rest.server.computation;
 import static org.apache.kafka.streams.StreamsConfig.PROCESSING_GUARANTEE_CONFIG;
 
 import com.fasterxml.jackson.core.JsonProcessingException;
+import com.google.common.collect.ImmutableMap;
 import io.confluent.ksql.KsqlExecutionContext;
 import io.confluent.ksql.config.ConfigItem;
 import io.confluent.ksql.config.KsqlConfigResolver;
@@ -37,6 +38,7 @@ import io.confluent.ksql.util.KsqlConstants;
 import io.confluent.ksql.util.KsqlServerException;
 import io.confluent.ksql.util.KsqlStatementException;
 import io.confluent.ksql.util.PersistentQueryMetadata;
+import io.confluent.ksql.util.PersistentQueryMetadataImpl;
 import io.confluent.ksql.util.QueryMetadata;
 import java.util.Collection;
 import java.util.Objects;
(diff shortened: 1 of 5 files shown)


jnh5y (approved):
Looks reasonable to me.

ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/computation/ValidatedCommandFactory.java
```diff
@@ -212,11 +214,22 @@ private static Command createForPlannedQuery(
...
 
-    final ConfiguredKsqlPlan configuredPlan = ConfiguredKsqlPlan
+    ConfiguredKsqlPlan configuredPlan = ConfiguredKsqlPlan
         .of(plan, statement.getSessionConfig());
 
-    context.execute(serviceContext, configuredPlan);
-
+    final KsqlExecutionContext.ExecuteResult result = context
+        .execute(serviceContext, configuredPlan);
+    if (result.getQuery().isPresent()
```
guozhangwang, line 222: Could we just simplify this as 

```
if (result.getQuery().get() instanceof PersistentQueryMetadataImpl) // always override the config to false
```
?

guozhangwang (approved):
Thanks for the PR @wcarlson5 , LGTM!

Before you merge, could you consider rephrasing the PR description so that a broader audience can understand better the purpose? Sorry for being nitpick here, but here's some quick thoughts:

1) Some terms are not widely recognized by other readers, such as `gen 1` v.s. `gen 2`.
2) Some sentences are a bit ill-presented and hard to gasp -- I'm honestly not a native speaker really, but e.g. "We would still bring it up as a gen 1 query but it would be written in the cmd topic as a gen 2 query." may sounds self-conflicting to people not familiar with the ongoing work.

How about rephrasing it to sth. like this:

```
When we restart a query with `create or replace` stmt we would like to use the previous runtime that query was executed on. However, today if we `create or replace` an existing query that's running on the old, non-shared runtime, the cmd topic may be written for this query to be run on the new shared runtime disregarding the config.

To fix this, when executing the `create or replace` stmt we would only create the new query in the new runtime only if 1) a shared runtime exist (with a known id), 2) we enabled shared runtime via the config, 3) the old query either does not exist or it has already been running on the shared runtime.
```

ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/computation/ValidatedCommandFactory.java
wcarlson5: The get without the is present check leaves us open to an no such element exception. I would rather just have the check than have to set up error handling logic. 

As for always setting the flag to be false. We can conceivably do this, but I don't really want to add an unrelated override to queries that are not already using the FF. I am just not sure about the consequences and figured it would be safer to just do the check. If you have a strong opinion on this I have no problem modifying it :)

wcarlson5:
Had to rebase

wcarlson5:
https://jenkins.confluent.io/job/confluentinc-pr/job/ksql/job/PR-8713/6/ 
The build passed I don't know why GitHub is not picking that up

ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/computation/ValidatedCommandFactory.java
guozhangwang: No strong opinions, was just wondering the motivation for being very specific. All good now.

Issue #8413, opened 2021-11-24
Scale/Precision error with decimal literal between -1 and 1 (0 left of decimal point)


mikebin:
**Describe the bug**
A scale/precision error occurs when using a decimal literal between -1 and 1 in a ksqlDB query - in other words, when the value left of the decimal point is 0. For example, comparing a field to the literal `0.005` in a query will result in the following error:
```
DECIMAL precision must be >= scale: DECIMAL(1,3)
```

Version: 0.22.0 and CP 7.0.0

```sql
create stream s (val double) with (kafka_topic='s', value_format='json', partitions=1);
insert into s (val) values (123.456);

select * from s where val > 0.005 emit changes;
```

**Expected behavior**
No errors, query result returned.

**Actual behaviour**
Error message:
```
DECIMAL precision must be >= scale: DECIMAL(1,3)
```

**Additional context**
Decimals < -1 or > 1 have no issues. For example, the following works:
```
select * from s where val > 1.005 emit changes;
```

This also works:
```
select * from s where val > (1.005-1) emit changes;
```

There was a fix a while back for a similar issue: https://github.com/confluentinc/ksql/pull/5531

agavra:
Natea was able to confirm, we'll pick it up in a fix-it-week

maheshba:
I am taking a look at this with some help from @nae701. The root cause seems to be that BigDecimal's precision is not defined exactly the same way as SQL Decimal's precision.

TL;DR: can we treat "0.12" and ".12" the same way and return precision=3 for both instead of precision=2 for the latter? (this requires changing an existing unit test")

For SQL Decimal, precision is just the total number of digits; and scale is the number of digits to the right of the period. So scale has to be smaller or equal to precision; and the scale cannot be negative. So for "0.005", we expect to get (precision=4, scale=3). For ".005", we expect to get (precision=3, scale=3).

BigDecimal has a broader definition of precision and scale. It has the same definition for scale if the scale is zero or positive and for precision if the decimal is <= -1 or >= 1. But if the decimal is in [-1, 1], precision can be smaller than scale. For example, for both "0.005" and ".005", it returns (precision=1, scale=3).

[...]

PR #8720, opened 2022-02-04
fix: bug preventing decimals in (-1, 1) from being inserted / queried

maheshba:
### Description 
Fix for bug in https://github.com/confluentinc/ksql/issues/8413 where ksqldb doesn't allow non-zero decimals between (-0.1, 0.1) to be inserted or queried. The root cause is that we use BigDecimal to parse the string, and it has a different behavior for precision than the SQL standard for (-1, 1): "0.005" has precision 1 in BigDecimal, whereas SQL expects it to have precision 4 (total number of digits). For anything in (-0.1, 0.1) this causes the code to throw since scale is greater than precision.

Something to note: the fix introduces a quirk in how decimals starting with a period are processed: "0.005" and ".005" are both treated as having precision 4, whereas SQL expects precision 3 for the latter. To fix this we'd have to stop using BigDecimal as the internal representation (since it parses both strings identically) or extend it somehow. This doesn't change ksql behavior for the worse since right now both strings would fail.

### Testing done 
Added a unit test and a functional test; locally deployed and verified the case described in https://github.com/confluentinc/ksql/issues/8413.

### Reviewer checklist

### DIFF
--- a/ksqldb-common/src/main/java/io/confluent/ksql/util/DecimalUtil.java
+++ b/ksqldb-common/src/main/java/io/confluent/ksql/util/DecimalUtil.java
@@ -274,10 +274,24 @@ public final class DecimalUtil {
         ? value.setScale(0, BigDecimal.ROUND_UNNECESSARY)
         : value;
 
-    /* When a BigDecimal has value 0, the built-in method precision() always returns 1. To account
-    for this edge case, we just take the scale and add one and use that for the precision instead.
-    */
-    if (decimal.compareTo(BigDecimal.ZERO) == 0) {
+    /* We can't use BigDecimal.precision() directly for all cases, since it defines
+     * precision differently from SQL Decimal.
+     * In particular, if the decimal is between -0.1 and 0.1, BigDecimal precision can be
+     * lower than scale, which is disallowed in SQL Decimal. For example, 0.005 in
+     * BigDecimal has a precision,scale of 1,3; whereas we expect 4,3.
+     * If the decimal is in (-1,1) but outside (-0.1,0.1), the code doesn't throw, but
+     * gives lower precision than expected (e.g., 0.8 has precision 1 instead of 2).
+     * To account for this edge case, we just take the scale and add one and use that
+     * for the precision instead. This works since BigDecimal defines scale as the
+     * number of digits to the right of the period; which is one lower than the precision for
+     * anything in the range (-1, 1).
+     * This covers the case where BigDecimal has a value of 0.
+     * Note: This solution differs from the SQL definition in that it returns (4, 3) for
+     * both "0.005" and ".005", whereas SQL expects (3, 3) for the latter. This is unavoidable
+     * if we use BigDecimal as an intermediate representation, since the two strings are parsed
+     * identically by it to have precision 1.
+     */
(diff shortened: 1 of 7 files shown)


agavra (commented):
nice! thanks for the fix @maheshba - I'll save the full review for @nateab but I had a spare cycle so figured I'd look at the code

ksqldb-functional-tests/src/test/resources/rest-query-validation-tests/insert-values.json
```diff
@@ -357,6 +357,18 @@
         {"topic": "test_topic", "key": 1, "value": {"V0": 0.00}}
       ]
     },
+    {
+      "name": "should insert decimals between zero and one",
```
agavra, line 361: nit: rest query translation tests take much longer to run than the standard QTT tests. instead it may make sense to just add a QTT test where we use a decimal literal that's `1 > x > -1`

maheshba: Thank you! Will do that. I'm still figuring out how to run the feature tests selectively locally (this test is breaking anyway right now) -- will fix and update!

agavra: Here are my configurations (the first runs all QTT tests, the second runs just a specific file). Also FYI there exists a `decimal.json` QTT test.

image
image

I think similar things work for RQTT

jnh5y: @agavra Would it be ok for me to steal these images and toss them in a debugging.md file?:)

maheshba: @agavra Thank you, this is super helpful! I arrived at the same IntelliJ dialog after a bunch of messing around, but your setup is a lot cleaner!

agavra: @jnh5y go for it :)

maheshba: Interestingly the bug doesn't trigger on the query translation tests, only on the rest query translation tests. Leaving the rest-based test in here while I figure out exactly why the query translation tests don't trigger it...

ksqldb-common/src/main/java/io/confluent/ksql/util/DecimalUtil.java
```diff
@@ -274,10 +274,22 @@ public static SqlType fromValue(final BigDecimal value) {
         ? value.setScale(0, BigDecimal.ROUND_UNNECESSARY)
         : value;
 
-    /* When a BigDecimal has value 0, the built-in method precision() always returns 1. To account
-    for this edge case, we just take the scale and add one and use that for the precision instead.
-    */
-    if (decimal.compareTo(BigDecimal.ZERO) == 0) {
+    /* We can't use BigDecimal.precision() directly for all cases, since it defines
```
nateab, line 277: Great documentation and explanation! If we ever decide to strictly follow the SQL standard for precision, it will be clear where to start from

nateab (approved):
LGTM! Thanks for the fix, and congrats on your first ksqldb change!!  🎉

PR #8721, opened 2022-02-04
chore: update netty to 4.1.72.Final to account for CVE

patrickstuedi:
Updating netty to 4.1.72 and also netty-tcnative-boringssl-static to 2.0.46

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -123,7 +123,7 @@
         <wiremock.version>2.24.0</wiremock.version>
         <clearspring-analytics.version>2.9.5</clearspring-analytics.version>
         <icu.version>67.1</icu.version>
-        <vertx.version>3.9.10</vertx.version>
+        <vertx.version>3.9.12</vertx.version>
         <reactive-streams.version>1.0.3</reactive-streams.version>
         <skip.docker.build>true</skip.docker.build>
         <skip.docker.test>true</skip.docker.test>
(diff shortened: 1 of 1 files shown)


pom.xml
```diff
@@ -135,8 +135,8 @@
 
         <netty-tcnative-version>2.0.39.Final</netty-tcnative-version>
         <!-- We normally get this from common, but Vertx is built against this -->
-        <netty.version>4.1.69.Final</netty.version>
-        <netty-codec-http2-version>4.1.69.Final</netty-codec-http2-version>
+        <netty.version>4.1.72.Final</netty.version>
+        <netty-codec-http2-version>4.1.72.Final</netty-codec-http2-version>
```
jnh5y, line 139: Netty version 4.1.73 appears to be available and recommended: https://netty.io/downloads.html

I imagine there's some coordination among repos, so if .72 is what's needed, that's fine!

jnh5y (approved):

pom.xml
agavra: yup - we've had history where a vertx/netty mismatch causes very tough to debug problems so we've just been playing it safe and keeping netty version to whatever vertx uses

jnh5y: +1.  Yeah, I've learned that to avoid mismatching Netty versions the hard way.:)

mjsax (approved):

guozhangwang (approved):

PR #8722, opened 2022-02-04
fix: only check configs on validation

wcarlson5:
### Description 
To avoid having a broken query in the cmd topic we should only do the check for query level configs during validation

### Testing done 

### Reviewer checklist


rodesai (approved):
LGTM

PR #8724, opened 2022-02-07
docs: fix line continuation chars in curl examples

vcrfxia:
### Description 

Minor docs fix: these two curl examples are missing line continuation characters so they don't work when copy-pasted.

### Testing done 

Docs-only change.

### Reviewer checklist


JimGalasyn (approved):
Awesome, thank you!

Gerrrr (approved):

Issue #8545, opened 2021-12-18
Several functions don't do null checks on inputs


jzaralim:
**Describe the bug**
There are a few functions (`AS_MAP`, `ARRAY_SORT` and `URL_ENCODE_PARAM` for example) that don't do null checks on inputs. We should do null checks because these errors are left uncaught and end up crashing queries. An example of one that has been fixed recently is https://github.com/confluentinc/ksql/pull/8400

jzaralim:
#8364 is a general fix for this

jnh5y:
@jzaralim `ARRAY_SORT` seems to have a null check.  (I'm looking for others and have confirmed the other two are problematic.). I just wanted to check that I'm not missing something with ARRAY_SORT.

jzaralim:
@jnh5y https://github.com/confluentinc/ksql/blob/6b65cd4da78e38832b6a9ec232595c3a8b8d1443/ksqldb-engine/src/main/java/io/confluent/ksql/function/udf/array/ArraySort.java#L51

The direction parameter doesn't have a null check. If it's null then `direction.toUpperCase()` will throw an NPE.

jnh5y:
I believe https://github.com/confluentinc/ksql/pull/8726 closes this issue at the minute.  Feel free to reopen if I've missed anything!

PR #8726, opened 2022-02-07
fix: Add null handling to functions

jnh5y:
Address https://github.com/confluentinc/ksql/issues/8545 by
adding null to each function which does not have them.

### Testing done 
Added unit test coverage for showing any change in behavior. 

### Reviewer checklist

### DIFF
--- a/ksqldb-engine/src/main/java/io/confluent/ksql/function/udf/json/JsonArrayContains.java
+++ b/ksqldb-engine/src/main/java/io/confluent/ksql/function/udf/json/JsonArrayContains.java
@@ -88,7 +88,8 @@ public class JsonArrayContains {
 
         parser.skipChildren();
         if (TOKEN_COMPAT.getOrDefault(token, foo -> false).test(val)) {
-          if (token == VALUE_NULL || Objects.equals(parser.readValueAs(val.getClass()), val)) {
+          if (token == VALUE_NULL
+              || (val != null && Objects.equals(parser.readValueAs(val.getClass()), val))) {
             return true;
           }
         }
--- a/ksqldb-engine/src/main/java/io/confluent/ksql/function/udf/json/JsonConcat.java
+++ b/ksqldb-engine/src/main/java/io/confluent/ksql/function/udf/json/JsonConcat.java
@@ -46,6 +46,9 @@ import java.util.List;
 public class JsonConcat {
   @Udf
   public String concat(@UdfParameter final String... jsonStrings) {
+    if (jsonStrings == null) {
+      return null;
+    }
     final List<JsonNode> nodes = new ArrayList<>(jsonStrings.length);
     boolean allObjects = true;
     for (final String jsonString : jsonStrings) {
--- a/ksqldb-engine/src/main/java/io/confluent/ksql/function/udf/math/Greatest.java
+++ b/ksqldb-engine/src/main/java/io/confluent/ksql/function/udf/math/Greatest.java
@@ -44,7 +44,7 @@ public class Greatest {
   @Udf
   public Integer greatest(@UdfParameter final Integer val, @UdfParameter final Integer... vals) {
 
-    return Stream.concat(Stream.of(val), Arrays.stream(vals))
+    return (vals == null) ? null : Stream.concat(Stream.of(val), Arrays.stream(vals))
         .filter(Objects::nonNull)
         .max(Integer::compareTo)
         .orElse(null);
@@ -53,7 +53,7 @@ public class Greatest {
   @Udf
   public Long greatest(@UdfParameter final Long val, @UdfParameter final Long... vals) {
 
-    return Stream.concat(Stream.of(val), Arrays.stream(vals))
+    return (vals == null) ? null : Stream.concat(Stream.of(val), Arrays.stream(vals))
         .filter(Objects::nonNull)
         .max(Long::compareTo)
         .orElse(null);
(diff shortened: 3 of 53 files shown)


jnh5y:
@colinhicks would you mind taking a spin through the changes in the unit tests to see if there are any notable changes in what you'd expect to happen with nulls?

ksqldb-engine/src/main/java/io/confluent/ksql/function/udf/json/JsonConcat.java
```diff
@@ -46,6 +46,9 @@
 public class JsonConcat {
   @Udf
   public String concat(@UdfParameter final String... jsonStrings) {
+    if (jsonStrings == null) {
+      return null;
+    }
```
colinhicks, line 51: Can the variadic parameter itself actually be null? I was trying to understand how the function could be invoked in this way without reflection.

ksqldb-engine/src/main/java/io/confluent/ksql/function/udf/math/Greatest.java
```diff
@@ -44,7 +44,7 @@
   @Udf
   public Integer greatest(@UdfParameter final Integer val, @UdfParameter final Integer... vals) {
 
-    return Stream.concat(Stream.of(val), Arrays.stream(vals))
+    return (vals == null) ? null : Stream.concat(Stream.of(val), Arrays.stream(vals))
```
colinhicks, line 47: Here and below, similar to JsonConcat, can `vals` actually be null?

colinhicks, line 47: Here and below, similar to JsonConcat, can `vals` actually be null?

ksqldb-engine/src/main/java/io/confluent/ksql/function/udf/math/Least.java
```diff
@@ -43,7 +43,7 @@
   @Udf
   public Integer least(@UdfParameter final Integer val, @UdfParameter final Integer... vals) {
 
-    return Stream.concat(Stream.of(val), Arrays.stream(vals))
+    return (vals == null) ? null : Stream.concat(Stream.of(val), Arrays.stream(vals))
```
colinhicks, line 46: Here and below, similar to JsonConcat and Greatest, can `vals` actually be null?

colinhicks (approved):
LGTM. Thanks for doing this! The behavior based on the tests is in line with what I'd expect as a user.

I was curious about null checking variadic parameters (which doesn't seem like it would have an effect on the result in any case), but I will leave the full review of the implementation to others.

ksqldb-engine/src/main/java/io/confluent/ksql/function/udf/json/JsonConcat.java
jnh5y: Good question.  Ksqldb does call these methods with reflection.  I think this questions boils down to "Can one construct a null array of type T?". 

I figured out how to do with it Strings.  Here's an example: `nullif('[a]', '[a]')`.

With a query like `select concat(nullif('[a]', '[a]')) from STREAM`, it returns empty instead of null.

I haven't figured out how to generate null lists/arrays for other types quite ready.  One really silly way to do would be to write a UDF which just returns a null list/array for whatever type.  

That aside, I'm hoping that the null checks are sane to keep rather than trying to reason through which functions ought not have null checks.  (Of course, if we want to dive into it, we can totally do that!:)

ksqldb-engine/src/main/java/io/confluent/ksql/function/udf/math/Greatest.java
jnh5y: Yeah, the only thing left to figure out is how to get a null list of a numeric type.  I'm having a little trouble doing that quickly; that may just be a limitation of my SQL creativity.:)

ksqldb-engine/src/main/java/io/confluent/ksql/function/udf/json/JsonConcat.java
Gerrrr: I think that it makes sense to check for `null` here and below even if we can't catch an NPE via SQL right now. Given that the null check is cheap and ksqlDB is an actively developing project, IMO we should better be a bit more defensive with this kind of checks.

ksqldb-engine/src/main/java/io/confluent/ksql/function/udf/math/Greatest.java
```diff
@@ -44,7 +44,7 @@
   @Udf
   public Integer greatest(@UdfParameter final Integer val, @UdfParameter final Integer... vals) {
 
-    return Stream.concat(Stream.of(val), Arrays.stream(vals))
```
Gerrrr, line 47: nit: I generally find explicit `if` statements more readable rather than ternary operators.

Gerrrr (approved):
LGTM!

ksqldb-engine/src/main/java/io/confluent/ksql/function/udf/json/JsonArrayContains.java
```diff
@@ -88,7 +88,8 @@
 
         parser.skipChildren();
         if (TOKEN_COMPAT.getOrDefault(token, foo -> false).test(val)) {
-          if (token == VALUE_NULL || Objects.equals(parser.readValueAs(val.getClass()), val)) {
+          if (token == VALUE_NULL
```
Gerrrr, line 91: This _is_ a non-trivial catch!

Issue #8725, opened 2022-02-07
Fix bugs in sample standard deviation UDAF


jnh5y:
The sample standard deviation function has a few bugs.

First, the return value does not seem to have the square root applied to it.  E.g. the standard deviation of {1,2,3,4,5} is square root of 2.5 not 2.5.  See here:  (https://github.com/confluentinc/ksql/blob/master/ksqldb-engine/src/test/java/io/confluent/ksql/function/udaf/stddev/StandardDeviationSampUdafTest.java#L48) 

Second, these lines perform integer rather than double arithmetic: https://github.com/confluentinc/ksql/blob/master/ksqldb-engine/src/main/java/io/confluent/ksql/function/udaf/stddev/StandardDeviationSampUdaf.java#L87-L88

jnh5y:
Duplicate of https://github.com/confluentinc/ksql/issues/8345.

jnh5y:
https://github.com/confluentinc/ksql/pull/8728 addresses this issue.

PR #8728, opened 2022-02-07
fix: Fix bugs in sample standard deviation UDAF

jnh5y:
* The square root was not being taken.
Previously, the square of the standard deviation was being returned.

* Additionally, Integers and Longs using integeral arithmetic and hence computing incorrect results.

Fixes https://github.com/confluentinc/ksql/issues/8725

### Testing done 
Updated unit tests.

### Reviewer checklist

### DIFF
--- a/ksqldb-engine/src/main/java/io/confluent/ksql/function/udaf/stddev/StandardDeviationSampUdaf.java
+++ b/ksqldb-engine/src/main/java/io/confluent/ksql/function/udaf/stddev/StandardDeviationSampUdaf.java
@@ -66,9 +66,8 @@ public final class StandardDeviationSampUdaf {
         (agg, newValue) ->
             Double.valueOf(newValue * (agg.getInt64(COUNT) + 1) - (agg.getInt64(SUM) + newValue)),
         (agg1, agg2) ->
-            Double.valueOf(
-                agg1.getInt64(SUM) / agg1.getInt64(COUNT)
-                    - agg2.getInt64(SUM) / agg2.getInt64(COUNT)),
+            agg1.getInt64(SUM).doubleValue() / agg1.getInt64(COUNT).doubleValue()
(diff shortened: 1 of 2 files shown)


agavra (changes requested):
Unfortunately this is backwards incompatible, we'll need to make it so that anyone who originally used this UDF gets the old behavior. I'll try to dig up a way to make this happen, but it's a bit of a pain.

jnh5y:
> Unfortunately this is backwards incompatible, we'll need to make it so that anyone who originally used this UDF gets the old behavior. I'll try to dig up a way to make this happen, but it's a bit of a pain.

@agavra makes sense.  Lemme know what you think.  Seems like we ought to find some way to warn users using the original UDF.

jnh5y:
@agavra as an option, we could keep the current UDAF and add a new one with the name `STDDEV_SAMPLE`.

As part of that, the description could be updated to explain that `STDDEV_SAMP` is deprecated due to the bug.

agavra:
@jnh5y i'm supportive of that, though we may want to run it by @colinhicks

colinhicks:
> @jnh5y i'm supportive of that, though we may want to run it by @colinhicks

I'm also +1 for adding the new function and deprecating the problematic one.

jnh5y:
@agavra @colinhicks @JimGalasyn ok, I've updated this PR.  Can you all check it out at your leisure?

JimGalasyn (approved):

agavra (approved):
LGTM based on test coverage, I didn't review the math or the "real" return values to make sure they work. It'd be nice to run the QTT tests in a postgres/mysql database to make sure we return identical things if you have the spare time

jnh5y:
@agavra Ah, I checked the results with an online standard deviation calculator.  The biggest differences were down to precision.  The changes in the answer boil down to taking the square root.

PR #8730, opened 2022-02-07
docs: remove beta tag tokens

nateab:
Similar reasoning to https://github.com/confluentinc/ksql/pull/8729, but essentially we no longer use beta tags for our dependency, so we don't need these tokens to try find the correct beta packaging url. Intsead we just use https://ksqldb-maven.s3.amazonaws.com/maven/ for everything


JimGalasyn (approved):
Yay!

agavra (approved):

PR #8740, opened 2022-02-09
John fix truncated spq

vvcephei:
Previously, stream pull queries would hang on topics that had
been truncated (by explicit truncation or by all the data falling
out of retention).

Now, the queries will return an empty result as expected.

### Testing done 
* Existing test suite
* Added integration test for truncated stream pull queries

### Reviewer checklist

### DIFF
--- a/ksqldb-api-client/src/test/java/io/confluent/ksql/api/client/integration/ClientIntegrationTest.java
+++ b/ksqldb-api-client/src/test/java/io/confluent/ksql/api/client/integration/ClientIntegrationTest.java
@@ -41,6 +41,7 @@ import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableListMultimap;
 import com.google.common.collect.ImmutableMap;
 import com.google.common.collect.Multimap;
+import com.google.common.collect.Sets;
 import io.confluent.common.utils.IntegrationTest;
 import io.confluent.ksql.GenericKey;
 import io.confluent.ksql.GenericRow;
@@ -107,12 +108,21 @@ import java.util.Collections;
 import java.util.LinkedList;
 import java.util.List;
 import java.util.Map;
+import java.util.Map.Entry;
 import java.util.Optional;
 import java.util.UUID;
 import java.util.concurrent.ExecutionException;
 import java.util.concurrent.TimeUnit;
 import java.util.function.Supplier;
+import java.util.stream.Collectors;
 import kafka.zookeeper.ZooKeeperClientException;
+import org.apache.kafka.clients.admin.Admin;
+import org.apache.kafka.clients.admin.DescribeTopicsResult;
+import org.apache.kafka.clients.admin.ListOffsetsResult.ListOffsetsResultInfo;
+import org.apache.kafka.clients.admin.OffsetSpec;
+import org.apache.kafka.clients.admin.RecordsToDelete;
+import org.apache.kafka.clients.admin.TopicDescription;
+import org.apache.kafka.common.TopicPartition;
 import org.apache.kafka.common.header.Header;
 import org.apache.kafka.common.header.internals.RecordHeader;
 import org.apache.kafka.connect.data.Struct;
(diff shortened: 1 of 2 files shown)


ksqldb-api-client/src/test/java/io/confluent/ksql/api/client/integration/ClientIntegrationTest.java
```diff
@@ -442,6 +510,36 @@ public void shouldStreamPullQueryOnEmptyStreamSync() throws Exception {
...
+
+    // Then
+    assertThat(streamedQueryResult.columnNames(), is(TEST_COLUMN_NAMES));
+    assertThat(streamedQueryResult.columnTypes(), is(TEST_COLUMN_TYPES));
+    assertThat(streamedQueryResult.queryID(), is(notNullValue()));
+
+    final List<Row> results = new LinkedList<>();
+    Row row;
+    while (true) {
+      row = streamedQueryResult.poll();
```
guozhangwang, line 530: Just wondering in normal scenarios, could it be possible that the first few polls return empty as the operator is not yet ready to fetch the first record? In other words do we guarantee that the first poll will always return some data as long as they are available?

ksqldb-engine/src/main/java/io/confluent/ksql/engine/KsqlEngine.java
```diff
@@ -463,7 +466,44 @@ private TopicDescription getTopicDescription(final Admin admin, final String sou
           // special case where we expect no work at all on the partition, so we don't even
           // need to check the committed offset (if we did, we'd potentially wait forever,
           // since Streams won't commit anything for an empty topic).
-          .filter(e -> e.getValue().offset() > 0L)
+          .filter(e -> e.getValue().offset() > 0L && e.getValue().offset() > startOffsetsForStreamPullQuery.get(e.getKey()))
```
guozhangwang: Could you augment the above comment as well?

ksqldb-engine/src/main/java/io/confluent/ksql/engine/KsqlEngine.java
```diff
@@ -454,6 +454,9 @@ private TopicDescription getTopicDescription(final Admin admin, final String sou
         new ListOffsetsOptions(IsolationLevel.READ_UNCOMMITTED)
     );
 
+    final ImmutableMap<TopicPartition, Long> startOffsetsForStreamPullQuery =
```
guozhangwang, line 457: Sigh.. The `ListOffsetsPartition` in the request can support fetching for multiple specs actually, but it seems they are not exposed via admin APIs. But I guess paying one more round trip is not too much of a deal breaker for stream pull queries.

guozhangwang (approved):
Only minor comments, please feel free to merge afterwards.

ksqldb-api-client/src/test/java/io/confluent/ksql/api/client/integration/ClientIntegrationTest.java
vvcephei: I'm not sure I follow. When you start the query, it resets its offsets to "earliest" and checks for the "end offsets". Then it reads from the start point to the end point. IIUC, those "start" and "end" offset admin client checks will hit the leaders of the partitions, so if there was an acknowledged send to the topic, then a subsequent query would see the sent record.

ksqldb-engine/src/main/java/io/confluent/ksql/engine/KsqlEngine.java
vvcephei: Ah, I thought it still applied, but I can see the value of providing more detail.

ksqldb-api-client/src/test/java/io/confluent/ksql/api/client/integration/ClientIntegrationTest.java
guozhangwang: What I was asking is in the `client.streamQuery` call are we guaranteed that the first `poll()` would always return the data, since if the fetching is async then maybe it does not guarantee --- similar to `consumer.poll()` which may not guarantee that the first `poll()` would always return data since the fetching IO is async. In the case, here if we beak immediately after the first `poll` returns no rows we may not be covering exactly what we want to test.

I cannot honestly remember how `streamQuery` is executed behind the scene, so if that's not an issue please ignore :P

PR #8742, opened 2022-02-09
feat: create command topic with command topic configs

lct45:
### Description 
We want the ability to put the ksql command topic on a separate kafka tenant from the main kafka tenant. This involves passing in configs with the `ksql.server.command` prefix that point to a different kafka cluster, generally by providing the path to a new config file.

Previously, we didn't include the `ksql.server.command` configs when creating the kafka topic client so there wasn't a way to specify different configs for the command topic. Now, we create a separate `kafkaTopicClient` that's specifically for the command topic and includes any of the `ksql.server.command` configs that might have been passed in. We use this when creating the command store and command runner, and when checking through the health check that the command topic is properly created.

We create 
### Testing done 
Updated existing unit tests, tested locally + in cloud development

### Reviewer checklist


ksqldb-engine/src/main/java/io/confluent/ksql/services/KafkaTopicClientImpl.java
```diff
@@ -168,10 +169,15 @@ public boolean isTopicExists(final String topic) {
     LOG.trace("Checking for existence of topic '{}'", topic);
     try {
       ExecutorUtil.executeWithRetries(
-          () -> adminClient.get().describeTopics(
+          () -> {
+            final DescribeTopicsResult result = adminClient.get().describeTopics(
+                ImmutableList.of(topic),
+                new DescribeTopicsOptions().includeAuthorizedOperations(true));
+            return adminClient.get().describeTopics(
```
wcarlson5, line 176: Do you just want to return `result` here? it seems duplicated

ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/healthcheck/HealthCheckAgent.java
```diff
@@ -58,22 +59,23 @@
...
+  private final Admin adminClient;
 
   public HealthCheckAgent(
       final SimpleKsqlClient ksqlClient,
       final KsqlRestConfig restConfig,
       final ServiceContext serviceContext,
       final KsqlConfig ksqlConfig,
-      final CommandRunner commandRunner
+      final CommandRunner commandRunner,
+      final Admin adminClient
```
wcarlson5, line 72: Are we making a `HealthCheckAgent` for the internal and external tenents?

ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/KsqlRestApplication.java
```diff
@@ -413,7 +420,8 @@ private void checkPreconditions() {
     for (final KsqlServerPrecondition precondition : preconditions) {
       final Optional<KsqlErrorMessage> error = precondition.checkPrecondition(
           restConfig,
-          serviceContext
+          serviceContext,
+          internalTopicClient
```
wcarlson5, line 424: Nice! I like that you distinguish internal vs external here

wcarlson5 (commented):
Pretty much good. Just a couple of nits

ksqldb-engine/src/main/java/io/confluent/ksql/services/KafkaTopicClientImpl.java
lct45: Mm yeah actually I think we can just go back to returning `adminClient.get()...`, this was leftover from the local POC work but I don't think we need it. Thanks for the catch

ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/healthcheck/HealthCheckAgent.java
lct45: Hmm that is an interesting point. The current implementation of `HealthCheckAgent` just checks to see what kind of permissions we have on the command topic and then assumes health based on that, but maybe we should also check to see if we can describe the processing log, since they're on different tenants

ksqldb-engine/src/main/java/io/confluent/ksql/services/KafkaTopicClientImpl.java
```diff
@@ -48,6 +48,7 @@
 import org.apache.kafka.clients.admin.CreateTopicsOptions;
 import org.apache.kafka.clients.admin.DeleteTopicsResult;
 import org.apache.kafka.clients.admin.DescribeTopicsOptions;
+import org.apache.kafka.clients.admin.DescribeTopicsResult;
```
jnh5y, line 51: nit: Is this import used?

jnh5y (approved):
Looks good to me.  I don't think anything is blocking.

ksqldb-engine/src/main/java/io/confluent/ksql/services/KafkaTopicClientImpl.java
lct45: it's not, good catch!

ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/healthcheck/HealthCheckAgent.java
rodesai: The intent behind this check is to assess whether ksql can communicate with its backing Kafka for the purpose of serving the ksqlDB API. So IMO checking that we can access the internal tenant here is ok.

As discussed, we should file a follow-up ticket to remove the checks for authorization and not-found exceptions once we've finished migrating every cluster to internal tenants.

rodesai (approved):
LGTM

ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/restore/KsqlRestoreCommandTopic.java
```diff
@@ -287,13 +288,14 @@ public static void main(final String[] args) throws Exception {
...
-        ServiceContextFactory.create(serverConfig,
-            () -> /* no ksql client */ null).getTopicClient(),
-        () -> transactionalProducer(serverConfig)
-    );
+    final Map<String, Object> adminClientConfigs =
+      new HashMap<>(serverConfig.getKsqlAdminClientConfigProps());
+    adminClientConfigs.putAll(serverConfig.originalsWithPrefix("ksql.server.command.consumer."));
+    final Admin admin = new DefaultKafkaClientSupplier().getAdmin(adminClientConfigs);
+    this.serverConfig = serverConfig;
+    commandTopicName = ReservedInternalTopics.commandTopic(serverConfig);
```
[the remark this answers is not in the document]
lct45: Reverted this commit to get this PR in while I do some more testing on this change, but will change this in the new PR

ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/healthcheck/HealthCheckAgent.java
lct45: Filed a follow-up ticket

PR #8749, opened 2022-02-11
docs: add javadoc artifacts for API client

nateab:
Description
Adds API docs to the release branch so that they can be served on https://docs.ksqldb.io/.

This PR should NOT be cherry-picked to other branches. We'd like to keep the generated API docs only on docs branches, as the source of truth for these are the Javadocs on the interfaces.

Testing done
Docs-only change.


JimGalasyn (approved):
Sweet!

nateab:
merging since this is a docs only change and builds are expected to fail for the docs branch

PR #8755, opened 2022-02-12
fix: various fixes for tracking shared runtime queries in validation

wcarlson5:
### Description 
During a refactor we missed a couple of issue. 

1. removing queries for the map when stoping the queries.
2. adding the topology to the sandbox to properly validate
3. registering the query outside for validation runtimes.

### Testing done 

### Reviewer checklist

### DIFF
--- a/ksqldb-engine/src/main/java/io/confluent/ksql/util/SharedKafkaStreamsRuntimeImpl.java
+++ b/ksqldb-engine/src/main/java/io/confluent/ksql/util/SharedKafkaStreamsRuntimeImpl.java
@@ -176,7 +176,7 @@ public class SharedKafkaStreamsRuntimeImpl extends SharedKafkaStreamsRuntime {
   public void stop(final QueryId queryId, final boolean resetOffsets) {
     log.info("Attempting to stop query: {} in runtime {} with resetOffsets={}",
              queryId, getApplicationId(), resetOffsets);
-    if (collocatedQueries.containsKey(queryId)) {
+    if (kafkaStreams.getTopologyByName(queryId.toString()).isPresent()) {
       if (kafkaStreams.state().isRunningOrRebalancing()) {
         try {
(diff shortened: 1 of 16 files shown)


ksqldb-engine/src/main/java/io/confluent/ksql/util/SharedKafkaStreamsRuntimeImpl.java
```diff
@@ -176,7 +176,7 @@ public TimeBoundedQueue getNewQueryErrorQueue() {
   public void stop(final QueryId queryId, final boolean resetOffsets) {
     log.info("Attempting to stop query: {} in runtime {} with resetOffsets={}",
              queryId, getApplicationId(), resetOffsets);
-    if (collocatedQueries.containsKey(queryId)) {
+    if (kafkaStreams.getTopologyByName(queryId.toString()).isPresent()) {
```
guozhangwang, line 179: Why we cannot use the old condition as well, and why not keep the `remove` call below inside the if condition?

wcarlson5: If the query was registered but not started this causes issues when removing form streams. This was just the simplest way I could get it to work

guozhangwang: Got it. My understanding is that:

1) If the query is registered, it would be in the `collocatedQueries`.
2) If the query is started, it would be added to `kafkaStreams`.
3) The query must always be registered before started, i.e. 2) always happens after 1).

So upon `stop`, the query can either be registered but not started, or started. Is that right? My question was that could it ever be not registered at all (i.e. not in `collocatedQueries`)? If not, then this fix looks elegant.

guozhangwang (approved):
Made a pass on the latest non-testing part, lgtm.

PR #8756, opened 2022-02-14
WIP: feat: Enable earliest_by_offset to take general types.

jnh5y:
### Description 

### Testing done 

### Reviewer checklist

### DIFF
--- a/ksqldb-engine/src/main/java/io/confluent/ksql/function/udaf/offset/EarliestByOffset.java
+++ b/ksqldb-engine/src/main/java/io/confluent/ksql/function/udaf/offset/EarliestByOffset.java
@@ -45,7 +45,7 @@ import org.apache.kafka.connect.data.Schema;
 import org.apache.kafka.connect.data.Struct;
 
 @UdafDescription(
-    name = "EARLIEST_BY_OFFSET",
+    name = "EARLIEST_BY_OFFSET_OLD",
     description = EarliestByOffset.DESCRIPTION,
     author = KsqlConstants.CONFLUENT_AUTHOR
 )
--- a/ksqldb-engine/src/main/java/io/confluent/ksql/function/udaf/offset/EarliestByOffsetFactory.java
+++ b/ksqldb-engine/src/main/java/io/confluent/ksql/function/udaf/offset/EarliestByOffsetFactory.java
@@ -1,0 +1,90 @@
+/*
+ * Copyright 2022 Confluent Inc.
+ *
+ * Licensed under the Confluent Community License (the "License"); you may not use
+ * this file except in compliance with the License.  You may obtain a copy of the
+ * License at
+ *
+ * http://www.confluent.io/confluent-community-license
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OF ANY KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations under the License.
+ */
+
+package io.confluent.ksql.function.udaf.offset;
+
+import com.google.common.collect.ImmutableList;
+import io.confluent.ksql.function.AggregateFunctionFactory;
+import io.confluent.ksql.function.AggregateFunctionInitArguments;
+import io.confluent.ksql.function.KsqlAggregateFunction;
+import io.confluent.ksql.function.types.ParamType;
+import io.confluent.ksql.schema.ksql.SchemaConverters;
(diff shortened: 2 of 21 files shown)


ksqldb-engine/src/main/java/io/confluent/ksql/function/udaf/offset/EarliestByOffset.java
```diff
@@ -45,7 +45,7 @@
 import org.apache.kafka.connect.data.Struct;
 
 @UdafDescription(
-    name = "EARLIEST_BY_OFFSET",
+    name = "EARLIEST_BY_OFFSET_OLD",
```
jnh5y, line 48: This change is made in order to show that the new, generalized UDAF is being used.

ksqldb-engine/src/main/java/io/confluent/ksql/function/udaf/offset/EarliestByOffsetFactory.java
```diff
@@ -0,0 +1,98 @@
...
+        final Object arg = args.get(0);
+        if (arg instanceof Boolean) {
+          ignoreNulls = (Boolean) arg;
+        } else if (arg instanceof Integer) {
+          n = (Integer) arg;
+        }
+      } else if (args.size() > 2) {
+        // JNH: This needs to be updated.
+        // udaf.json specifies the error a bit too much.
+        throw new KsqlException("Function 'EARLIEST_BY_OFFSET' does not accept "
```
jnh5y, line 66: I am still trying to work through how errors can / should be handled when resolving a UDAF.

This appears to be different for each UDAF.

ksqldb-engine/src/main/java/io/confluent/ksql/function/udaf/offset/EarliestByOffsetKudaf.java
```diff
@@ -0,0 +1,98 @@
...
+
+  public EarliestByOffsetKudaf(
+      final AggregateFunctionInitArguments initArgs,
+      final SqlType aggregateType,
+      final SqlType returnType,
+      final boolean ignoreNulls,
+      final Schema internalStructSchema) {
+    super(EarliestByOffsetFactory.FUNCTION_NAME,
+        initArgs.udafIndex(),
+        () -> null,
```
spena, line 46: For the sake of readability, should we use constant like `Object INITIALIZED_VALUE = null` and use it here? When I see nulls, I always look at the function to see this is the value we want. I do see this is just initialization.

ksqldb-engine/src/main/java/io/confluent/ksql/function/udaf/offset/EarliestByOffsetKudaf.java
```diff
@@ -0,0 +1,98 @@
...
+        aggregateType,
+        returnType,
+        Collections.singletonList(
+            new ParameterInfo("key", ParamTypes.LONG, "", false)), "description");
```
spena, line 50: What is this ParameterInfo? Btw, description needs a good message.

ksqldb-engine/src/main/java/io/confluent/ksql/function/udaf/offset/EarliestByOffsetFactory.java
```diff
@@ -0,0 +1,98 @@
...
+      final List<SqlArgument> argTypeList,
+      final AggregateFunctionInitArguments initArgs
+  ) {
+    // This parses the various initArgs that could be used.
+    // If there are two arguments, then they should be (Boolean, Integer)
+    // A single parameter could be passed in to handle either setting "n" or "ignoreNulls".
+    boolean ignoreNulls = true;
+    int n = 1;
+
+    if (initArgs != null) {
```
spena, line 50: Doesn't argTypeList contain the input parameters?

jnh5y: Good question.  It may contain the data that's been pulled out into `initArgs`.  If so, maybe some of this mess can be simplified...

Either way, the `argTypeLIst` is used down line 71 to get the type of the input column.

ksqldb-engine/src/main/java/io/confluent/ksql/function/udaf/offset/EarliestByOffsetKudaf.java
jnh5y: I'm feeling cheap and want to suggest using a comment instead of having a separate object.  Would that be ok?

PR #8761, opened 2022-02-15
fix: reinstate the old KsqlRestClient.create overload

swist:
### Description 
Commit a33b1e68426ae61fb8bf45c267f198d3eb3d6ff3
has changed the `create` method signature without
deprecating the old one first, causing the drift benchmarks
to fail with:
```
java.lang.NoSuchMethodError: io.confluent.ksql.rest.client.KsqlRestClient.create(
	java/lang/String;
	java/util/Map;
	java/util/Map;
	java/util/Optional;
)
io/confluent/ksql/rest/client/KsqlRestClient;
```
We're just adding back the missing signagure

### Testing done 
No test changes, just added a basic sanity check that we don't remove the signature later down the line

### Reviewer checklist

### DIFF
--- a/ksqldb-rest-client/src/test/java/io/confluent/ksql/rest/client/KsqlRestClientTest.java
+++ b/ksqldb-rest-client/src/test/java/io/confluent/ksql/rest/client/KsqlRestClientTest.java
@@ -19,6 +19,7 @@ import static io.confluent.ksql.rest.client.KsqlRestClient.CCLOUD_CONNECT_PASSWO
 import static io.confluent.ksql.rest.client.KsqlRestClient.CCLOUD_CONNECT_USERNAME_HEADER;
 import static org.hamcrest.MatcherAssert.assertThat;
 import static org.hamcrest.Matchers.containsString;
+import static org.hamcrest.Matchers.instanceOf;
 import static org.hamcrest.Matchers.is;
 import static org.junit.Assert.assertThrows;
 import static org.mockito.ArgumentMatchers.any;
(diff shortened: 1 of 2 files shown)


cadonna (approved):
LGTM!

Thanks @swist !

ksqldb-rest-client/src/test/java/io/confluent/ksql/rest/client/KsqlRestClientTest.java
```diff
@@ -19,9 +19,12 @@
 import static io.confluent.ksql.rest.client.KsqlRestClient.CCLOUD_CONNECT_USERNAME_HEADER;
 import static org.hamcrest.MatcherAssert.assertThat;
 import static org.hamcrest.Matchers.containsString;
+import static org.hamcrest.Matchers.instanceOf;
 import static org.hamcrest.Matchers.is;
+import static org.hamcrest.Matchers.not;
```
cadonna, line 24: I guess this is an unused import, isn't it?

ksqldb-rest-client/src/test/java/io/confluent/ksql/rest/client/KsqlRestClientTest.java
```diff
@@ -19,9 +19,12 @@
...
 import static org.junit.Assert.assertThrows;
 import static org.mockito.ArgumentMatchers.any;
+import static org.mockito.ArgumentMatchers.isNotNull;
```
cadonna, line 27: Same here.

vcrfxia (commented):
KsqlRestClient is not a public API. Why can't we just update the benchmarks repo to pass the additional `Optional.empty()` param? I can make the change if you'd like.

swist:
@vcrfxia I've looked at updating the benchmarks repo - the call doesn't live in the benchmarks repo itself, rather in the `drift` framework one, we would have to do some additional maven-fu to make sure that picks the correct version - `drift` uses `confluent.version` ref while drift-benchmark uses a custom range to resolve the nano versions. The most reliable way is to bring back the missing signature like this PR does

PR #8764, opened 2022-02-15
fix: change conflicting metrics name that leads to infinite loop

tolgadur:
[KCI-1455](https://confluentinc.atlassian.net/browse/KCI-1455)

### Description 
The metric name `max_used_task_storage_bytes` was conflicting with the `TASK_STORAGE_USED_BYTES` in `getMaxTaskUsage(metricRegistry)` and was thus leading to an infinite loop. 

### Testing done 
unit test

### Reviewer checklist


lct45:
Just checked this in CPD and I'm not seeing the error in the logs anymore

PR #8765, opened 2022-02-15
coerce property values to correct type

rodesai:
This patch changes Command to coerce overwritten properties to the right
types. Previously, if a user provided a property that deserializes to a
different type by default (e.g. a property that is declared as Short will
deserialize to Integer by default), the command would fail to execute due
to a type mismatch.


ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/computation/Command.java
```diff
@@ -117,7 +118,7 @@ public String getStatement() {
       justification = "overwriteProperties is unmodifiableMap()"
   )
   public Map<String, Object> getOverwriteProperties() {
-    return overwriteProperties;
+    return PropertiesUtil.coerceTypes(overwriteProperties, true);
```
guozhangwang, line 121: Just to clarify this is the actual and only fix of this PR right? All others seems to be cleanups with no logical changes.

ksqldb-common/src/main/java/io/confluent/ksql/properties/PropertiesUtil.java
```diff
@@ -159,4 +163,47 @@ private static void throwOnBlackListedProperties(final Map<String, ?> properties
...
+    try {
+      final String stringValue = value == null
+          ? null
+          : value instanceof List
+              ? listToString((List<?>) value)
+              : String.valueOf(value);
+
+      return PROPERTY_PARSER.parse(key, stringValue);
+    } catch (final Exception e) {
+      throw new KsqlException("Failed to set '" + key + "' to '" + value + "'", e);
```
guozhangwang, line 200: This exception message is a bit unclear to me, should that be "failed to coerce type of value {} for key {}"?

ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/computation/Command.java
guozhangwang: Also why we would like to ignore PropertyNotFoundException in command, while not in the ksql request below?

rodesai: The statement should already have been validated at this point, and we've committed to executing it. We should just deserialize and pass it along to the engine.

ksqldb-common/src/main/java/io/confluent/ksql/properties/PropertiesUtil.java
rodesai: Makes sense - it was just copypasta

ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/computation/Command.java
guozhangwang: Ah you're right!

guozhangwang (approved):

guozhangwang:
Took another look at the latest commit, still LGTM.

rodesai:
I'm merging this. The build passed here: https://jenkins.confluent.io/job/confluentinc-pr/job/ksql/view/change-requests/job/PR-8765/7/

PR #8769, opened 2022-02-15
fix: always close validation runtimes even on an exception

wcarlson5:
### Description 
If the validation fails due to an exception it can leave the runtimes lying around and this can cause issues for future validations until the cleaner service can get to them. To prevent that we should just clean them up even if there is an exception. 

### Testing done 

### Reviewer checklist

### DIFF
--- a/ksqldb-engine/src/main/java/io/confluent/ksql/engine/SandboxedExecutionContext.java
+++ b/ksqldb-engine/src/main/java/io/confluent/ksql/engine/SandboxedExecutionContext.java
@@ -162,19 +162,23 @@ final class SandboxedExecutionContext implements KsqlExecutionContext {
       final ServiceContext serviceContext,
       final ConfiguredKsqlPlan ksqlPlan
   ) {
-    final ExecuteResult result = EngineExecutor.create(
-        engineContext,
-        serviceContext,
-        ksqlPlan.getConfig()
-    ).execute(ksqlPlan.getPlan());
+    try {
+      final ExecuteResult result = EngineExecutor.create(
+          engineContext,
+          serviceContext,
+          ksqlPlan.getConfig()
+      ).execute(ksqlPlan.getPlan());
 
-    // Having a streams running in a sandboxed environment is not necessary
-    if (!getKsqlConfig().getBoolean(KsqlConfig.KSQL_SHARED_RUNTIME_ENABLED)) {
-      result.getQuery().map(QueryMetadata::getKafkaStreams).ifPresent(streams -> streams.close());
-    } else {
-      engineContext.getQueryRegistry().closeRuntimes();
+      // Having a streams running in a sandboxed environment is not necessary
+      if (!getKsqlConfig().getBoolean(KsqlConfig.KSQL_SHARED_RUNTIME_ENABLED)) {
+        result.getQuery().map(QueryMetadata::getKafkaStreams).ifPresent(streams -> streams.close());
+      }
+      return result;
+    } finally {
+      if (getKsqlConfig().getBoolean(KsqlConfig.KSQL_SHARED_RUNTIME_ENABLED)) {
+        engineContext.getQueryRegistry().closeRuntimes();
+      }
     }
-    return result;
   }
 
   @Override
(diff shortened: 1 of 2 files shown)


ksqldb-engine/src/main/java/io/confluent/ksql/util/BinPackedPersistentQueryMetadataImpl.java
```diff
@@ -382,7 +382,7 @@ public void onStateChange(final State newState, final State oldState) {
 
   @Override
   public void close() {
-    sharedKafkaStreamsRuntime.stop(queryId, true);
```
wcarlson5, line 385: Missed this in my previous PR. https://github.com/confluentinc/ksql/pull/8755

lct45: Catching up here - how do you know for sure that this isn't a `createOrReplace`?

ksqldb-engine/src/main/java/io/confluent/ksql/engine/SandboxedExecutionContext.java
```diff
@@ -162,19 +162,23 @@ public ExecuteResult execute(
       final ServiceContext serviceContext,
       final ConfiguredKsqlPlan ksqlPlan
   ) {
-    final ExecuteResult result = EngineExecutor.create(
-        engineContext,
-        serviceContext,
-        ksqlPlan.getConfig()
-    ).execute(ksqlPlan.getPlan());
+    try {
```
lct45, line 165: Where was the error being thrown before?

ksqldb-engine/src/main/java/io/confluent/ksql/engine/SandboxedExecutionContext.java
```diff
@@ -162,19 +162,23 @@ public ExecuteResult execute(
...
-    } else {
-      engineContext.getQueryRegistry().closeRuntimes();
+      // Having a streams running in a sandboxed environment is not necessary
+      if (!getKsqlConfig().getBoolean(KsqlConfig.KSQL_SHARED_RUNTIME_ENABLED)) {
+        result.getQuery().map(QueryMetadata::getKafkaStreams).ifPresent(streams -> streams.close());
+      }
+      return result;
+    } finally {
+      if (getKsqlConfig().getBoolean(KsqlConfig.KSQL_SHARED_RUNTIME_ENABLED)) {
+        engineContext.getQueryRegistry().closeRuntimes();
```
lct45, line 179: At this point we really can close all the runtimes, right? We don't need them for anything? And is there any risk that we'll miss the `ifPresent(streams -> streams.close())` above and would want to do that cleaning here too?

ksqldb-engine/src/main/java/io/confluent/ksql/util/BinPackedPersistentQueryMetadataImpl.java
wcarlson5: A create of replace uses the `stop(boolean)`, you can see that [here](https://github.com/confluentinc/ksql/blob/b8df77716226efcc93aa932e0ce8c0f58381cb0e/ksqldb-engine/src/main/java/io/confluent/ksql/query/QueryRegistryImpl.java#L446)

ksqldb-engine/src/main/java/io/confluent/ksql/engine/SandboxedExecutionContext.java
wcarlson5: This is the validation context. SO if the query fails due to an error it would show up here. We want to close the validation runtimes no matter what

ksqldb-engine/src/main/java/io/confluent/ksql/engine/SandboxedExecutionContext.java
wcarlson5: This is validation so yes we won't use them anymore. And the old engine didn't really care but now we need the cleaned up

lct45 (approved):
lgtm 🚂

wcarlson5:
The build passed https://jenkins.confluent.io/job/confluentinc-pr/job/ksql/job/PR-8769/

Idk why but sometimes it not showing up on github for me...

PR #8775, opened 2022-02-16
fix: pass DCN_NULLPOINTER_EXCEPTION spotbugs error

swist:
### Description 
We have changed our SpotBugs version to 4.5.3 which was causing:

```
[ERROR] Medium: Do not catch NullPointerException like in io.confluent.ksql.datagen.DataGen$Arguments$Builder.build() [io.confluent.ksql.datagen.DataGen$Arguments$Builder] At DataGen.java:[line 440] DCN_NULLPOINTER_EXCEPTION
[INFO]
```

This change is SpotBugs compliant way of doing the same thing
- explicit null checks, rather than delegating to a method
that checks for null and throws NPE

### Testing done 
Verified that spotbugs is happy with the new code

### Reviewer checklist

### DIFF
--- a/ksqldb-examples/src/main/java/io/confluent/ksql/datagen/DataGen.java
+++ b/ksqldb-examples/src/main/java/io/confluent/ksql/datagen/DataGen.java
@@ -29,7 +29,6 @@ import java.io.IOException;
 import java.io.InputStream;
 import java.util.Arrays;
 import java.util.Map;
-import java.util.Objects;
 import java.util.Optional;
 import java.util.Properties;
 import java.util.Random;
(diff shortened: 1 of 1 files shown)


ksqldb-examples/src/main/java/io/confluent/ksql/datagen/DataGen.java
```diff
@@ -364,14 +363,20 @@ Arguments build() {
...
+          throw new ArgumentParseException("Message key format not provided");
+        }
+        if (valueFormat == null) {
+          throw new ArgumentParseException("Message value format not provided");
+        }
+        if (topicName == null) {
+          throw new ArgumentParseException("Kafka topic name not provided");
+        }
+        if (keyName == null) {
+          throw new ArgumentParseException("Name of key column not provided");
```
cadonna, line 379: I would prefer to deduplicate code by having a method like
```
void requireArgumentsNotNull(final Object argument, final String errorMessage) {
    if (argument == null) {
        throw new ArgumentParseException(errorMessage);
    }
}
```

But I am also fine with this.

cadonna (approved):
LGTM with a minor comment

ksqldb-examples/src/main/java/io/confluent/ksql/datagen/DataGen.java
swist: I'll land this and follow up separately

PR #8777, opened 2022-02-16
fix: spotbugs warnings with Java 11

xvrl:
* ignore all spotbugs warnings in tests that only occur with Java 11
* remove duplicate commons-compress dependency declaration causing maven build warnings

PR #8780, opened 2022-02-16
chore: inherit mockito version from common

xvrl:
* common is on a newer version of mockito already
* overriding the version also prevents using the mockito-bom in
  https://github.com/confluentinc/common/pull/433 since 3.x doesn't
  publish a bom

### Description 

### Testing done 

### Reviewer checklist

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -145,7 +145,6 @@
         <apache.io.version>2.7</apache.io.version>
         <io.confluent.ksql.version>7.2.0-0</io.confluent.ksql.version>
         <io.confluent.schema-registry.version>${confluent.version.range}</io.confluent.schema-registry.version>
-        <mockito.version>3.8.0</mockito.version>
         <netty-tcnative-version>2.0.46.Final</netty-tcnative-version>
         <!-- We normally get this from common, but Vertx is built against this -->
         <!-- Note: `netty` depends on `tcnative` and if we bump `netty`


andrewegel (approved):

jnh5y:
As a comment/question, it seems like this could make the ksql build more fragile based on changes to the common pom.xml.  I'm new to building this ecosystem.; looks like there's some CI to avoid those issues?
FWIW, it looks like the GitHub link to CI is broken: https://github.com/confluentinc/ksql/pull/8780
This is one time where I'd really like to see that this change (which would bump mockito from 3.8 to 3.12.4 and the common change which would further move it from 3.12.4 to 4.3.1 does not build the ksqldb build.

xvrl:
@jnh5y it shouldn't make the build more fragile. All changes in common trigger a downstream build for ksql to make sure the change does not break downstream repos. Unfortunately ksql did not pass builds with Java 11 until #8777 was merged, so the checks weren't working for KSQL.

A previous run of this PR already succeeded yesterday with the current common pom, so this means moving to 3.12.4 will not break things (unfortunately jenkins no longer has the build history for this PR)

https://github.com/confluentinc/ksql/pull/8780 will fail until we merge this PR because mockito 3.x does not provide a mockito-bom. The bom is only available with 4.3.x, and once we remove the override in this repo, we can rerun the build for 
https://github.com/confluentinc/ksql/pull/8780 and ensure that nothing breaks in ksql before we merge it.

jnh5y:
Ok, @xvrl sounds good.  I'm concerned about the missing CI history, but I'll trust you on this one.

I could imagine the change from 3.12.4 to 4.3.1 requiring some code changes, so I would hope for a completely green build before that change is merged into common.

xvrl:
@jnh5y the CI issues seem to affect all ksql PRs, seems like a separate problem

xvrl:
> so I would hope for a completely green build before that change is merged into common

yes, that's the goal, it might require fixing use of deprecated mockito methods, but given that ksql has pretty strict rules around compiler warnings, I don't anticipate a huge amount of work.

PR #8782, opened 2022-02-16
build: add staging s3 url for dependency jars

nateab:
### Description 
As part of the internal release stabilization process, we produce non-public artifacts, and this change allows us to access them by using the staging repo.


AlanConfluent (approved):

nateab:
merging since this worked for running the aws benchmarks

PR #8783, opened 2022-02-16
chore: Upgrading ANTLR to version 4.9.2.

jnh5y:


pgaref (approved):

jnh5y:
I was wrong... `ce-kafka` depends on antlr here: https://github.com/confluentinc/ce-kafka/blob/master/build.gradle#L4091

I'm going to merge this PR as-is.  @vcrfxia if you have an idea about how to coordinate things better, let me know and I'll work on that as a follow-on.

vcrfxia:
Ah got'cha. That's tough since ksql doesn't depend on ce-kafka -- we build against confluentinc/kafka instead. We can take this discussion offline. I'm curious how we discovered this version incompatibility and what breaks when the versions are mismatched. Hopefully that'll inform how we can set up alerting so we at least know when the version is bumped in ce-kafka and react before a release is cut so it doesn't slip in the future.

PR #8785, opened 2022-02-16
feat: add java client support for push query v2 ALOS through retry method

nateab:
### Description 
We want to abstract away from the http endpoints, and we expect users to be using our official java client. Adding this native client support will make push query v2 ALOS more robust to network failures since we can add retries to the client that pick up where they left off using continuation tokens.
### Testing done 
Unit and integration testing.

### Reviewer checklist

### DIFF
--- a/ksqldb-api-client/src/main/java/io/confluent/ksql/api/client/StreamedQueryResult.java
+++ b/ksqldb-api-client/src/main/java/io/confluent/ksql/api/client/StreamedQueryResult.java
@@ -17,6 +17,7 @@ package io.confluent.ksql.api.client;
 
 import java.time.Duration;
 import java.util.List;
+import java.util.concurrent.atomic.AtomicReference;
 import org.reactivestreams.Publisher;
 
 /**
@@ -89,5 +90,9 @@ public interface StreamedQueryResult extends Publisher<Row> {
    * @return whether the {@code StreamedQueryResult} is failed.
    */
   boolean isFailed();
+
+  boolean hasContinuationToken();
+
+  AtomicReference<String> getContinuationToken();
 
 }
--- a/ksqldb-api-client/src/main/java/io/confluent/ksql/api/client/StreamedQueryResult.java
+++ b/ksqldb-api-client/src/main/java/io/confluent/ksql/api/client/StreamedQueryResult.java
@@ -15,11 +15,10 @@
 
 package io.confluent.ksql.api.client;
 
-import org.reactivestreams.Publisher;
-
 import java.time.Duration;
 import java.util.List;
 import java.util.Optional;
+import org.reactivestreams.Publisher;
 
 /**
  * The result of a query (push or pull), streamed one row at time. Records may be consumed by either
--- a/ksqldb-api-client/src/main/java/io/confluent/ksql/api/client/StreamedQueryResult.java
+++ b/ksqldb-api-client/src/main/java/io/confluent/ksql/api/client/StreamedQueryResult.java
@@ -15,10 +15,11 @@
 
 package io.confluent.ksql.api.client;
 
+import org.reactivestreams.Publisher;
+
 import java.time.Duration;
 import java.util.List;
-import java.util.concurrent.atomic.AtomicReference;
-import org.reactivestreams.Publisher;
+import java.util.Optional;
 
 /**
  * The result of a query (push or pull), streamed one row at time. Records may be consumed by either
@@ -93,6 +94,6 @@ public interface StreamedQueryResult extends Publisher<Row> {
 
   boolean hasContinuationToken();
 
-  AtomicReference<String> getContinuationToken();
+  Optional<String> getContinuationToken();
 
 }
--- a/ksqldb-api-client/src/main/java/io/confluent/ksql/api/client/StreamedQueryResult.java
+++ b/ksqldb-api-client/src/main/java/io/confluent/ksql/api/client/StreamedQueryResult.java
@@ -18,6 +18,8 @@ package io.confluent.ksql.api.client;
 import java.time.Duration;
 import java.util.List;
 import java.util.Optional;
+import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.ExecutionException;
 import org.reactivestreams.Publisher;
 
 /**
@@ -94,5 +96,8 @@ public interface StreamedQueryResult extends Publisher<Row> {
   boolean hasContinuationToken();
 
   Optional<String> getContinuationToken();
+
+  CompletableFuture<StreamedQueryResult> retry(int maxRetries)
+      throws InterruptedException, ExecutionException;
 
 }
--- a/ksqldb-api-client/src/main/java/io/confluent/ksql/api/client/StreamedQueryResult.java
+++ b/ksqldb-api-client/src/main/java/io/confluent/ksql/api/client/StreamedQueryResult.java
@@ -99,5 +99,4 @@ public interface StreamedQueryResult extends Publisher<Row> {
 
   CompletableFuture<StreamedQueryResult> retry(int maxRetries)
       throws InterruptedException, ExecutionException;
-
 }
--- a/ksqldb-api-client/src/main/java/io/confluent/ksql/api/client/StreamedQueryResult.java
+++ b/ksqldb-api-client/src/main/java/io/confluent/ksql/api/client/StreamedQueryResult.java
@@ -17,9 +17,7 @@ package io.confluent.ksql.api.client;
 
 import java.time.Duration;
 import java.util.List;
-import java.util.Optional;
 import java.util.concurrent.CompletableFuture;
-import java.util.concurrent.atomic.AtomicReference;
 import org.reactivestreams.Publisher;
 
 /**
@@ -93,9 +91,16 @@ public interface StreamedQueryResult extends Publisher<Row> {
    */
   boolean isFailed();
 
-  boolean hasContinuationToken();
-
-  Optional<AtomicReference<String>> getContinuationToken();
-
-  CompletableFuture<StreamedQueryResult> retry();
+  /**
+   * Returns a new {@code StreamedQueryResult} which starts returning results from the last saved
+   * continuation token.
+   *
+   * <p>This method will only work for push queries V2 when continuation tokens have also been
+   * enabled. It will not work for regular push queries or pull queries. This method will throw an
+   * error if there is no continuation token saved.
+   *
+   * @return a new {@code StreamedQueryResult} that starts returning results from the last saved
+   * continuation token.
+   */
+  CompletableFuture<StreamedQueryResult> continueFromLastContinuationToken();
 }
--- a/ksqldb-api-client/src/main/java/io/confluent/ksql/api/client/StreamedQueryResult.java
+++ b/ksqldb-api-client/src/main/java/io/confluent/ksql/api/client/StreamedQueryResult.java
@@ -99,8 +99,7 @@ public interface StreamedQueryResult extends Publisher<Row> {
    * enabled. It will not work for regular push queries or pull queries. This method will throw an
    * error if there is no continuation token saved.
    *
-   * @return a new {@code StreamedQueryResult} that starts returning results from the last saved
-   * continuation token.
+   * @return a new {@code StreamedQueryResult} that continues from last continuation token.
    */
   CompletableFuture<StreamedQueryResult> continueFromLastContinuationToken();
 }
--- a/ksqldb-api-client/src/main/java/io/confluent/ksql/api/client/StreamedQueryResult.java
+++ b/ksqldb-api-client/src/main/java/io/confluent/ksql/api/client/StreamedQueryResult.java
@@ -15,8 +15,6 @@
 
 package io.confluent.ksql.api.client;
 
-import com.google.common.annotations.VisibleForTesting;
-import io.vertx.core.Context;
 import java.time.Duration;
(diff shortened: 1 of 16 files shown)


ksqldb-api-client/src/main/java/io/confluent/ksql/api/client/impl/StreamQueryResponseHandler.java
```diff
@@ -88,6 +106,10 @@ protected void handleRow(final Buffer buff) {
         LOG.info("Response contains consistency vector " + jsonObject);
         serializedConsistencyVector.set((String) ((JsonObject) json).getMap().get(
             "consistencyToken"));
+      } else if (jsonObject.getMap() != null && jsonObject.getMap().containsKey("continuationToken")) {
+        LOG.info("Response contains continuation token " + jsonObject);
+        continuationToken.set((String) ((JsonObject) json).getMap().get(
```
nateab, line 111: We save the continuation token here from the query response

ksqldb-api-client/src/main/java/io/confluent/ksql/api/client/impl/ClientImpl.java
```diff
@@ -134,10 +138,16 @@ private ClientImpl(final ClientOptions clientOptions, final Vertx vertx,
           KsqlRequestConfig.KSQL_REQUEST_QUERY_PULL_CONSISTENCY_OFFSET_VECTOR,
           serializedConsistencyVector.get());
     }
+    if (PushOffsetVector.isContinuationTokenEnabled(properties)) {
+      requestProperties.put(
```
nateab, line 142: we supply the saved continuation token here when we do a retry

ksqldb-api-client/src/main/java/io/confluent/ksql/api/client/impl/StreamedQueryResultImpl.java
```diff
@@ -127,6 +144,40 @@ private void handleErrorWhilePolling(final Throwable t) {
...
+    return !this.continuationToken.isPresent();
+  }
+
+  @Override
+  public Optional<String> getContinuationToken() {
+    return this.continuationToken;
+  }
+
+  @Override
+  public CompletableFuture<StreamedQueryResult> retry(
```
nateab, line 158: the retry mechanism

ksqldb-api-client/src/test/java/io/confluent/ksql/api/client/integration/ClientIntegrationTest.java
```diff
@@ -611,6 +610,27 @@ public void shouldStreamPushQueryWithLimitSync() throws Exception {
     assertThat(streamedQueryResult.isComplete(), is(true));
   }
 
+  @Test
+  public void shouldRetryStreamPushQueryAfterError() throws Exception {
+    // When
+    final StreamedQueryResult streamedQueryResult = client.streamQuery(PUSH_QUERY).get();
```
nateab, line 616: the test i am working on. Some things I need to add are the appropriate configs to use spqs (latest offset, enabled push queryv2 etc). I also need to produce rows to the topic after I start the query like we do in spq functional tests. I had some of this code done, but I messed up something on git so i lost it ;/ but will add it back soon in another commit

ksqldb-api-client/src/test/java/io/confluent/ksql/api/client/integration/ConsistencyOffsetVectorFunctionalTest.java
```diff
@@ -196,6 +212,25 @@ public void shouldRoundTripCVWhenPullQueryOnTableAsync() throws Exception {
     verifyConsistencyVector(serializedCV);
   }
 
+  @Test
+  public void shouldRetry() throws Exception {
+    final Map<String, Object> properties = new HashMap<>();
+    properties.put("ksql.query.push.v2.enabled", true);
```
nateab, line 218: ignore this test, I think I should do my testing in the other testing class

ksqldb-api-client/src/main/java/io/confluent/ksql/api/client/impl/StreamedQueryResultImpl.java
```diff
@@ -127,6 +144,42 @@ private void handleErrorWhilePolling(final Throwable t) {
...
+      final int maxRetries) throws InterruptedException, ExecutionException {
+    long waitMs = 100;
+
+    requestProperties.put("request.ksql.query.push.start.offsets", continuationToken.get());
```
nateab, line 162: this is redundant since we do this in streamQuery already, and it also uses the old config name which is wrong

ksqldb-api-client/src/test/java/io/confluent/ksql/api/client/integration/ConsistencyOffsetVectorFunctionalTest.java
```diff
@@ -116,6 +129,9 @@
       .withProperty(KSQL_DEFAULT_KEY_FORMAT_CONFIG, "JSON")
       .withProperty(KSQL_QUERY_PULL_CONSISTENCY_OFFSET_VECTOR_ENABLED, true)
       .withProperty(KsqlConfig.KSQL_HEADERS_COLUMNS_ENABLED, true)
+      .withProperty(KSQL_QUERY_PUSH_V2_ENABLED, true)
```
nateab, line 132: i need to add these properties to the client integration test

ksqldb-api-client/src/main/java/io/confluent/ksql/api/client/impl/StreamQueryResponseHandler.java
nateab: this should be another if statement, not an else if, since we may want to save the consistency token and the continuation token

ksqldb-api-client/src/main/java/io/confluent/ksql/api/client/StreamedQueryResult.java
```diff
@@ -90,4 +93,9 @@
    */
   boolean isFailed();
 
+  boolean hasContinuationToken();
```
AlanConfluent, line 96: `hasContinuationToken()` and `getContinuationToken()`only seem to be used in tests.  It would probably be best to avoid adding these to the public API and just caste the class to the impl version to expose them.

ksqldb-api-client/src/main/java/io/confluent/ksql/api/client/StreamedQueryResult.java
```diff
@@ -90,4 +93,9 @@
...
+
+  Optional<AtomicReference<String>> getContinuationToken();
```
AlanConfluent, line 98: We could also just expose a single getter method like this, and assume the optional will be empty if there's no token.  At the moment, it's always there, from what I recall.

ksqldb-api-client/src/main/java/io/confluent/ksql/api/client/StreamedQueryResult.java
```diff
@@ -90,4 +93,9 @@
...
+
+  CompletableFuture<StreamedQueryResult> retry();
```
AlanConfluent, line 100: We should definitely add comments for this method to make clear when it should be used.  If there's no continuation token or the feature is disabled, normal push queries are used or it's a pull query, it will just issue the request as normal, so it shouldn't require more introspection in practice.  Not sure if it makes sense to call it `continue()` or maybe `retryOrContinue()`.

ksqldb-api-client/src/main/java/io/confluent/ksql/api/client/impl/StreamQueryResponseHandler.java
AlanConfluent: I agree, that makes sense.  Not sure in what situation we'd have both (not currently a thing), but maybe it could be.

ksqldb-api-client/src/main/java/io/confluent/ksql/api/client/impl/StreamedQueryResultImpl.java
```diff
@@ -127,6 +144,24 @@ private void handleErrorWhilePolling(final Throwable t) {
...
+  }
+
+  @Override
+  public Optional<AtomicReference<String>> getContinuationToken() {
+    return this.continuationToken;
+  }
+
+  @Override
+  public CompletableFuture<StreamedQueryResult> retry() {
+    if (!this.hasContinuationToken()) {
```
AlanConfluent, line 159: If we're going to throw an error with no continuation token, maybe it makes sense to call this `continue()`.

ksqldb-api-client/src/main/java/io/confluent/ksql/api/client/impl/ClientImpl.java
AlanConfluent: We also need to set `KSQL_QUERY_PUSH_V2_CONTINUATION_TOKENS_ENABLED` and `KSQL_QUERY_PUSH_V2_ENABLED` for the feature to work. We should probably set `KSQL_QUERY_PUSH_V2_CONTINUATION_TOKENS_ENABLED` since it's a bit internal, if `PushOffsetVector.isContinuationTokenEnabled(properties)` is true (regardless of if we have a continuation token yet).  Currently, you just set it in the test

ksqldb-common/src/main/java/io/confluent/ksql/util/PushOffsetVector.java
```diff
@@ -164,4 +164,15 @@ public boolean equals(final Object o) {
     final PushOffsetVector that = (PushOffsetVector) o;
     return Objects.equals(offsets.get(), that.offsets.get());
   }
+
+  public static boolean isContinuationTokenEnabled(final Map<String, Object> requestProperties) {
+    final Object continuationTokenEnabled
```
AlanConfluent, line 169: Do we also want to check `KSQL_QUERY_PUSH_V2_ENABLED` as well since this is required to enable V2 to begin with?

ksqldb-api-client/src/main/java/io/confluent/ksql/api/client/StreamedQueryResult.java
AlanConfluent: Also, this method can probably just return an Optional<String>, hiding the fact that it's held by an AtomicReference.

ksqldb-api-client/src/main/java/io/confluent/ksql/api/client/StreamedQueryResult.java
nateab: I guess I was thinking of the original example like below we had where the user checks if there is a continuation token and then retries it, but I think it makes to sense to check that internally and hide the tokens logic from the user if possible.
```
  if (result.hasError() && result.hasContinuationToken()) {
    // Explicit retry or could store and try again later
    result.retry()
  }```

Comment (nateab) [reply ksqldb-api-client/src/main/java/io/confluent/ksql/api/client/StreamedQueryResult.java:96]: I removed them from the public api and cast to the impl version in the tests

File: ksqldb-api-client/src/main/java/io/confluent/ksql/api/client/StreamedQueryResult.java (hunk 9 above)
Comment (nateab) [reply ksqldb-api-client/src/main/java/io/confluent/ksql/api/client/StreamedQueryResult.java:98]: i just went ahead and tested removing the optional all together now and it worked. This is also only now exposed in the impl class anyway

File: ksqldb-api-client/src/main/java/io/confluent/ksql/api/client/StreamedQueryResult.java (hunk 10 above)
Comment (nateab) [reply ksqldb-api-client/src/main/java/io/confluent/ksql/api/client/StreamedQueryResult.java:100]: added javadoc

File: ksqldb-api-client/src/main/java/io/confluent/ksql/api/client/impl/StreamedQueryResultImpl.java (hunk 11 above)
Comment (nateab) [reply ksqldb-api-client/src/main/java/io/confluent/ksql/api/client/impl/StreamedQueryResultImpl.java:159]: `continue` is a reserved keyword, so i renamed it to `continueFromLastContinuationToken`

File: ksqldb-common/src/main/java/io/confluent/ksql/util/PushOffsetVector.java (hunk 12 above)
Comment (nateab) [reply ksqldb-common/src/main/java/io/confluent/ksql/util/PushOffsetVector.java:169]: added check

File: ksqldb-api-client/src/test/java/io/confluent/ksql/api/client/integration/ClientIntegrationTest.java [hunk 13]
```diff
@@ -611,6 +633,61 @@ public void shouldStreamPushQueryWithLimitSync() throws Exception {
...
+    final String oldContinuationToken = ((StreamedQueryResultImpl) oldStreamedQueryResult).getContinuationToken().get();
+
+    shouldReceiveStreamRows(oldStreamedQueryResult, false, TEST_NUM_ROWS );
+
+    // Continuation token should get updated
+    assertThatEventually(() -> oldContinuationToken.equals(((StreamedQueryResultImpl) oldStreamedQueryResult).getContinuationToken().get()), is(false));
+
+    // Mimic a network fault by abruptly stopping and starting the server
+    try {
+      REST_APP.stop();
```
AlanConfluent, line 673: At the moment, this instance of the app is shared between all of the tests, since it's initialized with `@ClassRule`.  Maybe it makes sense to pull this test out to another class so that a failure here isn't going to cause failures with the other tests.

AlanConfluent (approved):

PR #8789, opened 2022-02-17
fix: check for the queryId in the Task ID for registering saturation metric

wcarlson5:
### Description 
When registering the saturation metric we used to parse the Thread name to get the query ID. When we use shared runtimes the query ID is not in the Thread name. Instead that is in the Task ID, so we will just try to parse that to see if we can find the Query ID  first then fall back to the old approach. Then we don't have to worry about passing through a config. 

### Testing done 
Just add a unit test with no Query ID in the thread name put it is in the Task ID

### Reviewer checklist

### DIFF
--- a/ksqldb-engine/src/main/java/io/confluent/ksql/internal/StorageUtilizationMetricsReporter.java
+++ b/ksqldb-engine/src/main/java/io/confluent/ksql/internal/StorageUtilizationMetricsReporter.java
@@ -268,7 +268,7 @@ public class StorageUtilizationMetricsReporter implements MetricsReporter {
 
   private String getQueryId(final KafkaMetric metric) {
     final String taskName = metric.metricName().tags().getOrDefault("task-id", "");
-    final Pattern namedTopologyPattern = Pattern.compile("(.*?)__(\\d*_\\d*)");
+    final Pattern namedTopologyPattern = Pattern.compile("(.*?)__\\d*_\\d*");
     final Matcher namedTopologyMatcher = namedTopologyPattern.matcher(taskName);
     if (namedTopologyMatcher.find()) {
       return namedTopologyMatcher.group(1);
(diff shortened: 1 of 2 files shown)


ksqldb-engine/src/test/java/io/confluent/ksql/internal/StorageUtilizationMetricsReporterTest.java
```diff
@@ -103,8 +109,8 @@ public void shouldAddNodeMetricsOnConfigure() throws IOException {
     assertThat((Long) storageTotalValue, greaterThan(0L));
     assertThat((Long) storageUsedValue, greaterThan(0L));
     assertThat((Double) pctUsedValue, greaterThan(0.0));
+    assertThat((Integer) numStatefulTasksValue, greaterThan(0));
     assertThat((BigInteger) maxTaskUsageValue, greaterThanOrEqualTo(BigInteger.ZERO));
-    assertEquals(numStatefulTasksValue, 7);
```
wcarlson5, line 107: This test was relying on other test. When run by itself it would fail. I just adapted it to not do so

ksqldb-engine/src/main/java/io/confluent/ksql/internal/StorageUtilizationMetricsReporter.java
```diff
@@ -267,6 +267,13 @@ public static synchronized BigInteger getMaxTaskUsage(final Metrics metricRegist
   }
 
   private String getQueryId(final KafkaMetric metric) {
+    final String taskName = metric.metricName().tags().getOrDefault("task-id", "");
+    final Pattern namedTopologyPattern = Pattern.compile("(.*?)__(\\d*_\\d*)");
```
guozhangwang, line 271: I'm not an expert on regex, just checking for the last two digits should that be with or without the parentheses?

guozhangwang (approved):
Just a question for my own educations :) otherwise LGTM

ableegoldman (approved):
LGTM!

wcarlson5:
The same thing happened on my last PR but the build passed here https://jenkins.confluent.io/job/confluentinc-pr/job/ksql/job/PR-8789/

Issue #5437, opened 2020-05-20
earliest/latest_by_offset don't work with structs, maps, and arrays


MichaelDrogalis:
**Describe the bug**

`earliest/latest_by_offset` are critical functions for converting changelogs to streams, but they don't work with all the data types. Namely, they don't support structs, maps or arrays. These functions are one of the only mechanisms for directly during a stream into a table, so the lack of coverage can really hurt.

**To Reproduce**

ksqlDB 0.9.0:

```sql
ksql> CREATE STREAM s1 (
>    k VARCHAR,
>    b STRUCT<
>        e VARCHAR,
>        f INT
>    >,                                                                                                                          
>    c MAP<VARCHAR, INT>,
>    d ARRAY<INT>
>) WITH (
>    kafka_topic = 's1',
>    partitions = 1,
>    value_format = 'avro',
>    key = 'k'
>);

 Message        
----------------
 Stream created 
----------------
ksql> CREATE TABLE t1 AS
>    SELECT k,
>           LATEST_BY_OFFSET(b),
>           LATEST_BY_OFFSET(c),
>           LATEST_BY_OFFSET(d)                                                                                                  
>    FROM s1
>    GROUP BY k
>    EMIT CHANGES;
Function 'LATEST_BY_OFFSET' does not accept parameters (STRUCT<E STRING, F INTEGER>).
Valid alternatives are:
LATEST_BY_OFFSET(BOOLEAN val)
LATEST_BY_OFFSET(DOUBLE val)
LATEST_BY_OFFSET(INT val)
LATEST_BY_OFFSET(BIGINT val)
LATEST_BY_OFFSET(VARCHAR val)
For detailed information on a function run: DESCRIBE FUNCTION <Function-Name>;
```

Both earliest + latest by offset should support structs, maps, and arrays.

purplefox:
I don't think this is a bug. I'm not sure there's an easy way of supporting this using the current UDAF framework.

MichaelDrogalis:
@purplefox Maybe bug was the wrong label. But you can see how you're basically toast if you can't use this with all your data types, right?

purplefox:
Maybe @agavra can chime in here - he is the expert on the UDAF type system. Perhaps there is some esoteric way of doing this.

big-andy-coates:
And decimals too!

wlaforest:
Until this is supported we should indicate when a type isn't supported.  Currently I get an uncaught exception.

[...]

PR #8801, opened 2022-02-23
fix: Make earliest_by_offset general.

jnh5y:
### Description 
This PR allows for `earliest_by_offset` to take arbitrary types.

Once this is approved and merged, I will put a similar PR for `latest_by_offset`.

### Testing done 
Added unit tests.

### Reviewer checklist


jnh5y:
This PR is part of addressing https://github.com/confluentinc/ksql/issues/5437

ksqldb-engine/src/main/java/io/confluent/ksql/function/udaf/offset/EarliestByOffsetFactory.java
```diff
@@ -0,0 +1,100 @@
...
+
+  public EarliestByOffsetFactory() {
+    super(NAME);
+  }
+
+  @SuppressWarnings("checkstyle:CyclomaticComplexity")
+  @Override
+  public KsqlAggregateFunction createAggregateFunction(
+      final List<SqlArgument> argTypeList,
+      final AggregateFunctionInitArguments initArgs
```
spena, line 44: What's the difference between these two arguments? I'm getting confused about them.

jnh5y: Yeah, they are confusing....

We need an example like `earliest_by_offset(column, n)` to see the difference.  

Here `column` would be passed the `argTypeList`.  In the case of UDAFs there is a current restriction/assumption that UDAFS may only act on one column.  So `argTypeList` will usually be a singleton.  (It won't have more elements.  I cannot remember how `CountKudaf` works precisely.  Maybe that is empty...)

On the other hand, `initArgs` contains everything else one would need to know in order to make the UDAF go.  In this specific example, the value `n` would be in an array of arguments.  Additionally there's configuration Map which can hold whatever one likes which is passed to Udafs which implement `Configurable`.  

As an example of a `Configurable` UDAF, `collect_list` and `collect_set` have limits which are configured at the system level to limit the size of the collection maintained in memory.

One option I considered is adding the `argTypeList` in the configuration map.  This would work, but it seems a little brittle....

jnh5y:
Closing in favor of https://github.com/confluentinc/ksql/pull/8878

PR #8804, opened 2022-02-24
docs: klip-61: bootstrapped TABLE subscriptions

nateab:
klip 61 discussion

### DIFF
--- a/design-proposals/klip-61-bootstrapped-table-subscriptions.md
+++ b/design-proposals/klip-61-bootstrapped-table-subscriptions.md
@@ -26,7 +26,7 @@ to achieve this.
 ## What is in scope
 
 - Queries should be supported on `TABLES` with `EMIT ALL` at the end.
-- Bootstrapped subscriptions should retrieve all past data and subscribe to any further changes.
+- Bootstrapped subscriptions should retrieve all current rows in the `TABLE` and subscribe to any further changes.
 - Bootstrapped subscriptions should not terminate until the user quits them.
 - Bootstrapped subscriptions will have feature parity with push queries v2, such as filters and projections.
 
@@ -63,6 +63,10 @@ For a bootstrapped `TABLE` subscription, we will peek at the latest offset in th
 console. We will then save this offset and start the pull query, which can now take as long as needed since we have already
 saved the latest offset. Once the pull query finishes and returns the current state, we can now begin the subscription 
 and supply the saved offset as the starting subscription token to the push query in order to receive any future changes.
+
+One limitation of this design is that there is a race condition between getting the latest offsets in the ksqlDB output 
+topic and executing the pull query, since the pull query can possibly see new data. This can result in "time travel" 
+once the subscription starts since we may return older results that the pull query has already returned until we catch up.
 
 ## Test plan
 
--- a/design-proposals/klip-61-bootstrapped-table-subscriptions.md
+++ b/design-proposals/klip-61-bootstrapped-table-subscriptions.md
@@ -1,0 +1,96 @@
+# KLIP 61 - Bootstrapped TABLE subscriptions
+
+**Author**: Natea Eshetu Beshada @nateab |
+**Release Target**: 0.27.0 |
+**Status**: _In Discussion_ |
+**Discussion**: _link to the design discussion PR_
+
+**tl;dr:** We want users to be able to very easily with a single query get all past data from a table and 
+simultaneously subscribe to any future changes to that table by specifying 'EMIT ALL'.
+
+## Motivation and background
+
+ksqlDB is a streaming database that allows users to work with data in motion (stream processing) and data at rest (databases).
+Thus, we have push queries which let you subscribe to a table as it changes in real-time, and also pull 
+queries which retrieves the current state of the table, similar to a traditional database query. To further cement ksqlDB's
+positioning at the intersection of stream processing and traditional databases, it is necessary to develop features that 
+seamlessly combine the two. Adding bootstrapped table subscriptions to ksqlDB achieves this purpose by allowing users to query their 
+data at rest up until now, while also subscribing to any later changes from data that is in motion. 
+
+Users can try to achieve this functionality by simply combining a table pull query followed by a table push query, but this 
+could result in gaps in the returned data since there are no guarantees that the push query will not miss any data after the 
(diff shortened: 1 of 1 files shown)


design-proposals/klip-61-bootstrapped-table-subscriptions.md
```diff
@@ -0,0 +1,85 @@
...
+Users can try to achieve this functionality by simply combining a `TABLE` pull query followed by a `TABLE` push query, but this 
+could result in gaps in the returned data since there are no guarantees that the push query will not miss any data after the 
+pull query is executed. Implementing this feature, we will keep track off the offsets under the hood for the user so that 
+we can guarantee there is no missing data. It will also be more clean and intuitive for users to have to only use one query
+to achieve this.
+
+## What is in scope
+
+- Queries should be supported on `TABLES` with `EMIT ALL` at the end.
+- Bootstrapped subscriptions should retrieve all past data and subscribe to any further changes.
```
mjsax, line 29: > all past data

Sound like historic updates? Should we say all "current rows in the TABLE" instead?

design-proposals/klip-61-bootstrapped-table-subscriptions.md
```diff
@@ -0,0 +1,85 @@
...
+  [ AND and_condition ]
+  EMIT ALL;
+```
+
+## Design
+
+For a bootstrapped `TABLE` subscription, we will peek at the latest offset in the ksqlDB output topic using the kafka admin
+console. We will then save this offset and start the pull query, which can now take as long as needed since we have already
+saved the latest offset. Once the pull query finishes and returns the current state, we can now begin the subscription 
+and supply the saved offset as the starting subscription token to the push query in order to receive any future changes.
```
mjsax, line 65: Should we point out that there is still a race condition between getting the output topic end-offsets and running the pull query, that may result in "time travel" when the subscription starts (ie, the subscription might contain a few updates that happened _before_ the pull query(>

mjsax (commented):
Thanks for the KLIP. Overall LGTM. Very well written!

design-proposals/klip-61-bootstrapped-table-subscriptions.md
nateab: Yeah I agree this can be clearer, will update to say "current rows in the TABLE"

design-proposals/klip-61-bootstrapped-table-subscriptions.md
nateab: Yes this is a valid point, we should bring up this limitation to make it clear, and also makes it clear what future work can be done

design-proposals/klip-61-bootstrapped-table-subscriptions.md
```diff
@@ -0,0 +1,89 @@
...
+- Queries should be supported on `TABLES` with `EMIT ALL` at the end.
+- Bootstrapped subscriptions should retrieve all current rows in the `TABLE` and subscribe to any further changes.
+- Bootstrapped subscriptions should not terminate until the user quits them.
+- Bootstrapped subscriptions will have feature parity with push queries v2, such as filters and projections.
+
+## What is not in scope
+
+- Bootstrapped subscriptions will not be supported on `STREAMS`. Instead, we will throw an informative error that tells the
+user to use `SET 'auto.offset.reset'='earliest';` with a push query in order to get all past data in the `STREAM` while 
+simultaneously subscribing to any future updates. This is because unlike `TABLES` which have a current state that a pull query retrieves,
```
AlanConfluent, line 37: Maybe even "materialized current state".  It presumably also won't work for tables that are not materialized.

design-proposals/klip-61-bootstrapped-table-subscriptions.md
```diff
@@ -0,0 +1,89 @@
...
+## Design
+
+For a bootstrapped `TABLE` subscription, we will peek at the latest offset in the ksqlDB output topic using the kafka admin
+console. We will then save this offset and start the pull query, which can now take as long as needed since we have already
+saved the latest offset. Once the pull query finishes and returns the current state, we can now begin the subscription 
+and supply the saved offset as the starting subscription token to the push query in order to receive any future changes.
+
+One limitation of this design is that there is a race condition between getting the latest offsets in the ksqlDB output 
+topic and executing the pull query, since the pull query can possibly see new data. This can result in "time travel" 
+once the subscription starts since we may return older results that the pull query has already returned until we catch up.
```
AlanConfluent, line 69: We can mention that future work is possible to address this such as getting the output offsets from the materialized data, or from a cache which maps upstream offsets to output offsets.  This is an extension of the work that our team is doing on consistency.  We don't need to go into detail here, but maybe stating that it's possible is reasonable.

design-proposals/klip-61-bootstrapped-table-subscriptions.md
```diff
@@ -0,0 +1,89 @@
...
+
+## Test plan
+
+- Unit testing.
+- Integration testing in `ksqldb-rest-app` to ensure that modules are working as expected when combined.
+- Functional testing through the rest query validation tests (RQTT) to ensure that we are returning the correct results
```
AlanConfluent, line 75: We might need to extend this functionality a bit.  Currently, there's a flag which effectively says, produce the input either before or after the query is run.  We want a combination if you're really going to test it all.

AlanConfluent (approved):

design-proposals/klip-61-bootstrapped-table-subscriptions.md
nateab: good catch, added

design-proposals/klip-61-bootstrapped-table-subscriptions.md
nateab: Yeah that will make the current limitation seem more reasonable, added the points you suggested

design-proposals/klip-61-bootstrapped-table-subscriptions.md
nateab: added that we will need to update the RQTT framework in order to properly test this

nateab:
Thanks for the speedy review @mjsax! I've addressed your comments. Thanks for the approval @AlanConfluent !

mjsax (approved):

PR #8807, opened 2022-02-24
chore: Remove netty from functional tests packaging.

jnh5y:
### Description 
This removes netty from the functional tests packaging.


swist (approved):
makes sense to me

PR #8808, opened 2022-02-24
chore: Switching from shade to assembly plugin for ksqldb-benchmarks.

jnh5y:
### Description 
Switches plugins used to build the benchmarks.jar.

This will allow developers to skip all assembly work with `mvn -Dassembly.skipAssembly <goals>`


vcrfxia (approved):
LGTM assuming you've verified that the benchmarks still run using the same commands as before (else we need docs updates). Thanks!

PR #8809, opened 2022-02-24
feat: add rate-limiting to ksql command topic

lct45:
### Description 

### Testing done 

### Reviewer checklist

### DIFF
--- a/ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/computation/CommandRunner.java
+++ b/ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/computation/CommandRunner.java
@@ -16,6 +16,7 @@
 package io.confluent.ksql.rest.server.computation;
 
 import com.google.common.annotations.VisibleForTesting;
+import com.google.common.util.concurrent.RateLimiter;
 import edu.umd.cs.findbugs.annotations.SuppressFBWarnings;
 import io.confluent.ksql.rest.Errors;
 import io.confluent.ksql.rest.entity.ClusterTerminateRequest;
@@ -90,6 +91,7 @@ public class CommandRunner implements Closeable {
   private final Supplier<Boolean> commandTopicExists;
   private boolean commandTopicDeleted;
   private Status state = new Status(CommandRunnerStatus.RUNNING, CommandRunnerDegradedReason.NONE);
+  private RateLimiter rateLimiter;
 
   public enum CommandRunnerStatus {
     RUNNING,
@@ -218,6 +220,7 @@ public class CommandRunner implements Closeable {
         Objects.requireNonNull(commandTopicExists, "commandTopicExists");
     this.incompatibleCommandDetected = false;
     this.commandTopicDeleted = false;
+    this.rateLimiter = RateLimiter.create(500.0);
   }
 
   /**
@@ -346,6 +349,7 @@ public class CommandRunner implements Closeable {
       if (closed) {
         LOG.info("Execution aborted as system is closing down");
       } else {
+        rateLimiter.acquire();
         statementExecutor.handleStatement(queuedCommand);
         LOG.info("Executed statement: " + commandStatement);
       }


ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/KsqlRestApplication.java
```diff
@@ -932,7 +932,8 @@ static KsqlRestApplication buildApplication(
         errorHandler,
         internalTopicClient,
         commandTopicName,
-        metricCollectors.getMetrics()
+        metricCollectors.getMetrics(),
+        ksqlConfig.getDouble("ksql.command.topic.rate.limit")
```
wcarlson5, line 936:
```suggestion
        ksqlConfig.getDouble(KsqlConfig.KSQL_COMMAND_TOPIC_RATE_LIMIT_CONFIG)
```

ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/computation/CommandRunner.java
```diff
@@ -346,6 +354,11 @@ private void executeStatement(final QueuedCommand queuedCommand) {
       if (closed) {
         LOG.info("Execution aborted as system is closing down");
       } else {
+        if (!rateLimiter.tryAcquire()) {
+          throw new KsqlRestException(
```
wcarlson5, line 358: How will this error get handled?

And can you a unit test to make sure its gets thrown and handled?

wcarlson5 (commented):
Couple things. It would be nice to get a test or two nothing crazy. Just so if someone changes related code without knowing about this and break it we will catch it

ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/computation/CommandRunner.java
lct45: Yeah writing a unit test rn - it will return an error back up to the user

ksqldb-common/src/main/java/io/confluent/ksql/util/KsqlConfig.java
```diff
@@ -663,6 +663,12 @@
   private static final String KSQL_ENDPOINT_MIGRATE_QUERY_DOC
       = "Migrates the /query endpoint to use the same handler as /query-stream.";
 
+  public static final String KSQL_COMMAND_TOPIC_RATE_LIMIT_CONFIG
+      = "ksql.command.topic.rate.limit";
+  public static final double KSQL_COMMAND_TOPIC_RATE_LIMIT_CONFIG_DEFAULT = 10.0;
```
wcarlson5, line 668: Nit: Do we need a double or can we use an int?

wcarlson5 (approved):
LGTM

ksqldb-common/src/main/java/io/confluent/ksql/util/KsqlConfig.java
lct45: the library takes a double as `permitsPerSecond` https://guava.dev/releases/19.0/api/docs/index.html?com/google/common/util/concurrent/RateLimiter.html

jnh5y:
Is this going to cause users scripting commands to have a rough time?

lct45:
@jnh5y correct me if my understanding is wrong, but users who are scripting commands would also be running their own ksql, right? From my understanding Confluent Cloud users can't write a script to create streams / tables unless it issues a curl request, but it seems like that would take enough time to prevent us getting over 10 requests in a second.

If a user is running their own ksql they can configure this through the properties file

jnh5y:
Were I running my own ksql, I think having to set this would be somewhat surprising.

For ccloud, I suppose it depends on how someone was scripting things;).

stevenpyzhang (commented):
I thought we wanted to rate limit the number of records written to the command topic? This PR looks like it only rate limits how fast we read the records from the command topic, someone could still write a ton of records to the command topic.

stevenpyzhang (approved):

lct45:
error on the cli:
```
ksql> create stream c1 as select * from riderlocations; 
Too many writes to the command topic within a 1 second timeframe
```

wcarlson5:
@lct45 what happens if you try a batch like this all at once?

> Create stream users (A string, B string key) with (kafka_topic='users', VALUE_FORMAT='json', PARTITIONS=1);

create or replace table users_agg0 as select B, count(*) from USERS group by B;
create table users_agg1 as select B, count(*) from USERS group by B;
create table users_agg2 as select B, count(*) from USERS group by B;
create table users_agg3 as select B, count(*) from USERS group by B;
create table users_agg4 as select B, count(*) from USERS group by B;
create table users_agg5 as select B, count(*) from USERS group by B;
create table users_agg6 as select B, count(*) from USERS group by B;
create table users_agg7 as select B, count(*) from USERS group by B;
create table users_agg8 as select B, count(*) from USERS group by B;
create table users_agg9 as select B, count(*) from USERS group by B;
create table users_agg10 as select B, count(*) from USERS group by B;
create table users_agg11 as select B, count(*) from USERS group by B;
create table users_agg12 as select B, count(*) from USERS group by B;
create table users_agg13 as select B, count(*) from USERS group by B;
create table users_agg14 as select B, count(*) from USERS group by B;
create table users_agg15 as select B, count(*) from USERS group by B;
create table users_agg16 as select B, count(*) from USERS group by B;
create table users_agg17 as select B, count(*) from USERS group by B;
create table users_agg18 as select B, count(*) from USERS group by B;

lct45:
@wcarlson5 it executed all the statements with no errors (I did it as one block from the CLI)

ksqldb-rest-app/src/test/java/io/confluent/ksql/rest/server/computation/DistributingExecutorTest.java
```diff
@@ -496,4 +500,33 @@ public void shouldNotEnqueueRedundantIfNotExists() {
     assertThat("Should be present", response.getEntity().isPresent());
     assertThat(((WarningEntity) response.getEntity().get()).getMessage(), containsString(""));
   }
+  
+  @Test
+  public void shouldThrowIfRateLimitHit() {
+    // Given:
+    final DistributingExecutor rateLimitedDistributor = new DistributingExecutor(
+      new KsqlConfig(ImmutableMap.of("ksql.command.topic.rate.limit", 1.0)),
```
stevenpyzhang, line 508: Nit: use the const in KsqlConfig

ksqldb-common/src/main/java/io/confluent/ksql/util/KsqlConfig.java
stevenpyzhang: Can we set the default much higher (like 1000)? I think 10 is too low for a default and many users may inadvertently hit this.

stevenpyzhang (approved):

ksqldb-common/src/main/java/io/confluent/ksql/util/KsqlConfig.java
stevenpyzhang: Also, could you move the config to KsqlRestConfig and follow a similar naming pattern to `KSQL_COMMAND_RUNNER_BLOCKED_THRESHHOLD_ERROR_MS` and `DISTRIBUTED_COMMAND_RESPONSE_TIMEOUT_MS_CONFIG` with `ksql.server.command` prefix? I think we should keep the command topic/command runner configs together and named consistently

lct45: sure, moved it. I upped the limit for now, @rodesai WDYT about doing double max value versus 10? seems ok to me

rodesai: Yes we should definitely disable rate-limiting by default by setting this to MAX_INT. We can enable in cloud as needed.

ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/computation/DistributingExecutor.java
```diff
@@ -196,6 +203,13 @@ public StatementExecutorResponse execute(
           statement.getStatementText()), e);
     }
 
+    if (!rateLimiter.tryAcquire()) {
+      throw new KsqlRestException(
+        Errors.tooManyRequests(
+          "Too many writes to the command topic within a 1 second timeframe"
```
rodesai, line 209: nit: change this to "DDL/DML rate is crossing the configured rate limit of <rate-limit> statements/second"

ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/computation/DistributingExecutor.java
```diff
@@ -196,6 +203,13 @@ public StatementExecutorResponse execute(
           statement.getStatementText()), e);
     }
 
+    if (!rateLimiter.tryAcquire()) {
```
rodesai, line 206: I think it would make sense to tryAcquire with a delay of a second here. This way we don't fail scripts that serially try to enqueue a large number of commands.

ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/computation/DistributingExecutor.java
lct45: updated

ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/computation/DistributingExecutor.java
lct45: @rodesai just added -> will you confirm this is what you were thinking?

rodesai (approved):
LGTM

lct45:
https://jenkins.confluent.io/job/confluentinc-pr/job/ksql/job/PR-8809/16/ had a green build

PR #8810, opened 2022-02-24
[WIP] fix: generalize earliest_by_offset

jnh5y:
### Description 

### Testing done 

### Reviewer checklist

### DIFF
--- a/ksqldb-engine/src/main/java/io/confluent/ksql/function/InternalFunctionRegistry.java
+++ b/ksqldb-engine/src/main/java/io/confluent/ksql/function/InternalFunctionRegistry.java
@@ -245,6 +245,7 @@ public class InternalFunctionRegistry implements MutableFunctionRegistry {
     }
 
     private void addUdafFunctions() {
+      // functionRegistry.addAggregateFunctionFactory(new EarliestByOffsetFactory());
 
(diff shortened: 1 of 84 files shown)


spena:
I think we can make the schema provider work in UDAF with this approach. We could make it a non-polymorphic way if add 3 more methods for the struct, arrays, and maps ,and let the new implemented methods return the dynamic type.

It  would probably be easier to see how it is implemented if you implement CollectSet?

jnh5y:
> I think we can make the schema provider work in UDAF with this approach. We could make it a non-polymorphic way if add 3 more methods for the struct, arrays, and maps ,and let the new implemented methods return the dynamic type.
> 
> It would probably be easier to see how it is implemented if you implement CollectSet?

Why would we need methods for each of struct, array, and map?

Separately, apologies that this PR has two implementations of Earliest_By_Offset.  The new (to this branch) approach is here: https://github.com/confluentinc/ksql/pull/8810/files#diff-270184f9941f91affac2de764551cb15da9a66f8cfd314c796dd635b1a13d00bR390-R466

jnh5y:
And yes, I'll give implementing `collect_list` a try with this approach.  That'll give us something concrete to talk about for that.

PR #8811, opened 2022-02-24
fix: add copious amounts of logging

wcarlson5:
### Description 
I added a lot of logging to fill in gaps I found when running tests. 

### Testing done 

### Reviewer checklist


ksqldb-engine/src/main/java/io/confluent/ksql/util/SandboxedSharedKafkaStreamsRuntimeImpl.java
```diff
@@ -71,9 +72,36 @@ public void register(
...
-    if (!kafkaStreams.getTopologyByName(
-        binpackedPersistentQueryMetadata.getQueryId().toString()).isPresent()) {
-      kafkaStreams.addNamedTopology(binpackedPersistentQueryMetadata.getTopologyCopy(this));
+    try {
+      if (!kafkaStreams.getTopologyByName(
+          binpackedPersistentQueryMetadata.getQueryId().toString()).isPresent()) {
+        log.debug("Registering query {} for validation", queryId);
+        try {
+          kafkaStreams.addNamedTopology(binpackedPersistentQueryMetadata.getTopologyCopy(this))
+              .all()
```
wcarlson5, line 81: I added blocking calls to the validation. I don't know if this is necessary or will even work. I might remove them, running tests now

ksqldb-engine/src/main/java/io/confluent/ksql/util/SandboxedSharedKafkaStreamsRuntimeImpl.java
```diff
@@ -71,9 +72,36 @@ public void register(
...
+              .get();
+        } catch (final Exception e) {
+          log.error(String.format("Validation for query %s failed due to:", queryId), e);
+          throw e;
+        }
+      } else {
+        log.info("Validation for a create or replace runtime");
+        try {
```
wcarlson5, line 89: I also made a small change here. When verifying a COR statement we can remove the old topology and try the new one instead of just skipping that part of validation

ksqldb-engine/src/main/java/io/confluent/ksql/engine/RuntimeAssignor.java
```diff
@@ -79,11 +80,15 @@ public String getRuntimeAndMaybeAddRuntime(final QueryId queryId,
     }
     runtimesToSources.get(runtime).addAll(sources);
     idToRuntime.put(queryId, runtime);
+    log.info("Assigning query {} to runtime {}", queryId, runtime);
     return runtime;
   }
 
   public void dropQuery(final PersistentQueryMetadata queryMetadata) {
     if (queryMetadata instanceof BinPackedPersistentQueryMetadataImpl) {
+      log.info("Unassigning query {} form runtime {}",
```
alapidas, line 89: typo

ksqldb-engine/src/main/java/io/confluent/ksql/engine/RuntimeAssignor.java
```diff
@@ -79,11 +80,15 @@ public String getRuntimeAndMaybeAddRuntime(final QueryId queryId,
...
+          queryMetadata.getQueryId(),
+          idToRuntime.getOrDefault(queryMetadata.getQueryId(), "no runtime"));
```
guozhangwang, line 91: Hmm... could it ever happen that `idToRuntime` is empty after we have done the registration? Maybe we can just do `log(Unassign..)` if it is not empty, and if it is empty meaning the query has not be registered yet we log a different like `log(Dropping un-registered query...)`.

wcarlson5: This should not be possible and would indicate a bug. I like the idea of `log(Dropping un-registered query...)` I'll try that instead

ksqldb-engine/src/main/java/io/confluent/ksql/util/SandboxedSharedKafkaStreamsRuntimeImpl.java
wcarlson5: Seems this is necessary in-order to get the error in the first place.

ksqldb-engine/src/main/java/io/confluent/ksql/util/SandboxedSharedKafkaStreamsRuntimeImpl.java
```diff
@@ -71,9 +72,36 @@ public void register(
...
+          log.error(String.format("Validation for query %s failed due to:", queryId), e);
+          throw e;
+        }
+      }
+    } catch (ExecutionException | InterruptedException e) {
+      log.error(String.format(
+          "Failed to validate query %s within the allotted timeout due to",
+          queryId), e);
     }
     log.debug("mapping {}", collocatedQueries);
```
ableegoldman, line 106: can you update/improve this log line? It's a bit too terse, and note that logging the entire `collocatedQueries` map will log the entire metadata object which is huge. Let's have it match what we do in the real runtime class.

Also I think we can bump this up to info.

ksqldb-engine/src/main/java/io/confluent/ksql/util/SandboxedSharedKafkaStreamsRuntimeImpl.java
ableegoldman: Nice, only suggestion is a slight grammatical improvement to the logging, something like 
```
        log.info("Validating CREATE OR REPLACE for query {} in runtime {}", queryId, getApplicationId());
```

ksqldb-engine/src/main/java/io/confluent/ksql/util/SandboxedSharedKafkaStreamsRuntimeImpl.java
```diff
@@ -71,9 +72,36 @@ public void register(
       final QueryId queryId
   ) {
     collocatedQueries.put(queryId, binpackedPersistentQueryMetadata);
-    if (!kafkaStreams.getTopologyByName(
-        binpackedPersistentQueryMetadata.getQueryId().toString()).isPresent()) {
-      kafkaStreams.addNamedTopology(binpackedPersistentQueryMetadata.getTopologyCopy(this));
+    try {
+      if (!kafkaStreams.getTopologyByName(
+          binpackedPersistentQueryMetadata.getQueryId().toString()).isPresent()) {
```
ableegoldman, line 77:
```suggestion
      if (!kafkaStreams.getTopologyByName(queryId.toString()).isPresent()) {
```

ksqldb-engine/src/main/java/io/confluent/ksql/util/SandboxedSharedKafkaStreamsRuntimeImpl.java
```diff
@@ -71,9 +72,36 @@ public void register(
...
+              .all()
+              .get();
+        } catch (final Throwable e) {
+          log.error(String.format("Validation for query %s failed due to:", queryId), e);
+          throw e;
+        }
+      }
+    } catch (ExecutionException | InterruptedException e) {
+      log.error(String.format(
+          "Failed to validate query %s within the allotted timeout due to",
```
ableegoldman, line 103:
```suggestion
          "Failed to validate query %s due to",
```
There's no timeout for query validation (technically there's no timeout for adding or removing either, which is where you got this from I'm guessing -- should probably remove any references to a timeout in the real runtime class for now)

ksqldb-engine/src/main/java/io/confluent/ksql/util/SharedKafkaStreamsRuntimeImpl.java
```diff
@@ -203,6 +203,7 @@ public void stop(final QueryId queryId, final boolean isCreateOrReplace) {
       // we don't want to lose it from this runtime
```
ableegoldman, line 203: Github won't let me comment but in the catch block on line 192 can we also do the unwrapping there as well? ie something like
```
final Throwable t = e.getCause == null ? e : e.getCause();
throw t;
```

Same for any other instances of ExecutionException, I forgot but when you `get()` on a KafkaFuture it always wraps the real root cause in this ExecutionException so we should be unwrapping it to get the real root cause

ableegoldman (commented):
Thanks for the improvements (plus the fix), looking good but had a few suggestions to hopefully keep this as clean and easy to follow as possible

ksqldb-engine/src/main/java/io/confluent/ksql/util/SandboxedSharedKafkaStreamsRuntimeImpl.java
```diff
@@ -71,9 +72,36 @@ public void register(
       final QueryId queryId
   ) {
     collocatedQueries.put(queryId, binpackedPersistentQueryMetadata);
```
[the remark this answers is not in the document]
wcarlson5: I don't think we need to, they should be tracked together tightly. Besides the issue would be more about the topology is in the colocated queries and not in streams and the other way around

ableegoldman (approved):
LGTM, once you get the build fixed + catch/unwrap the ExecutionException anytime we call `all().get()`

wcarlson5:
There is a little issue so I am just going to merge the logging and file a ticket to fix the other issue.

wcarlson5:
https://confluentinc.atlassian.net/browse/KCI-1491

PR #8813, opened 2022-02-25
Enable consistency mode by default

patrickstuedi:
Turning on consistency mode by default. Still requires client to set 'ksql.query.pull.consistency.token.enabled', otherwise consistency mode will be off.


vpapavas (approved):
LGTM!

vvcephei (approved):
Thanks, @patrickstuedi !

It looks like there are some failing tests, and the title of the PR doesn't conform to the semantic commit requirement.

PR #8814, opened 2022-02-25
chore: cleanup consistency token

vpapavas:
### Description 
Removed the RW locks from the code and used ConcurrentHashMap like we do in Stream's `Position`. Ported some methods over as well and removed not used methods that had errors like `dominate`

### Testing done 
Updated the unit test

### Reviewer checklist


ksqldb-common/src/main/java/io/confluent/ksql/util/ConsistencyOffsetVector.java
```diff
@@ -51,141 +51,95 @@
...
   // Topic -> Partition -> Offset
-  private Map<String, Map<Integer, Long>> offsetVector;
-  private ReadWriteLock rwLock;
-
-  public ConsistencyOffsetVector() {
-    this.version = 0;
-    this.offsetVector = new HashMap<>();
-    this.rwLock = new ReentrantReadWriteLock();
-  }
+  private final ConcurrentHashMap<String, ConcurrentHashMap<Integer, Long>> offsetVector;
```
patrickstuedi, line 54: Instead of copying all that from Position, why don't we just keep an private member of type Position and map all the operations?

patrickstuedi: Do we need to keep the OffsetVector interface?

ksqldb-common/src/main/java/io/confluent/ksql/util/ConsistencyOffsetVector.java
```diff
@@ -222,16 +176,8 @@ public boolean equals(final Object o) {
...
-
-    try {
-      rwLock.readLock().lock();
-      other.rwLock.readLock().lock();
-      return this.version == other.version && offsetVector.equals(other.offsetVector);
-    } finally {
-      rwLock.readLock().unlock();
-      other.rwLock.readLock().unlock();
-    }
+    final ConsistencyOffsetVector vector1 = (ConsistencyOffsetVector) o;
```
patrickstuedi, line 179: Given we have decided to keep the client exposed type ConsistencyOffsetVector separate from the internal Position type, shouldn't we then make sure we don't expose Position in any of those methods? For instance, we can either make this logic above a non-public translation method, or we do it outside this class.

ksqldb-common/src/main/java/io/confluent/ksql/util/ConsistencyOffsetVector.java
```diff
@@ -271,4 +205,35 @@ public static boolean isConsistencyVectorEnabled(final Map<String, Object> reque
...
+  @JsonIgnore
+  @Override
+  public List<Long> getDenseRepresentation() {
+    throw new UnsupportedOperationException("Unsupported");
+  }
+
+  @JsonIgnore
+  @Override
+  public boolean dominates(final OffsetVector other) {
+    throw new UnsupportedOperationException("Unsupported");
```
patrickstuedi, line 223: I guess that's an artifact of the OffsetVector interface we are implementing. Are we using OffsetVector anywhere?

patrickstuedi (commented):
Thanks @vpapavas , added a few comments.

ksqldb-common/src/main/java/io/confluent/ksql/util/ConsistencyOffsetVector.java
vpapavas: I think we discussed this offline, unless you mean something else. `Position` does not have a public constructor so I cannot use the serialize/deserialize we currently have.

ksqldb-common/src/main/java/io/confluent/ksql/util/ConsistencyOffsetVector.java
vpapavas: I am not sure I follow. Where is the Position exposed to the client? The code you are pointing is the `equals`. What is wrong here?

ksqldb-common/src/main/java/io/confluent/ksql/util/ConsistencyOffsetVector.java
vpapavas: The interface per se, no we are not using it currently anywhere. But both the `PushOffsetVector` and the `ConsistencyOffsetVector` are implementing it

ksqldb-common/src/main/java/io/confluent/ksql/util/ConsistencyOffsetVector.java
vvcephei: Thanks, all, I'm not sure whether we need the OffsetVector interface or not (personally, I think not), but it's not a public API, so it's not crucially important to hammer that out.

Whether we keep a hashmap internally or a Position seems to be also up in the air; there are pros and cons on both sides. FWIW, I'm happier when there's less coupling in my life, so I'm more inclined toward the approach in this PR.

ksqldb-common/src/main/java/io/confluent/ksql/util/ConsistencyOffsetVector.java
vvcephei: I think @patrickstuedi was referring to `public void updateFromPosition(Position position)` above. GH doesn't let you comment on arbitrary lines in a PR, so sometimes you have to just comment on a random line.

Philosophically, I agree. If we want to achieve good decoupling, the way to do it is not reference Position or ConsistencyOffsetVector at all from each other, but have a separate utility method to translate between them.

On the other hand, ConsistencyOffsetVector is not a public API, and it is itself a utility (it's in the `util` package), so that method isn't too much of a sin.

In the interest of forward progress, I'd be happy to accept this PR as-is, but I'd be equally happy to see Patrick's suggestion adopted.

ksqldb-common/src/main/java/io/confluent/ksql/util/ConsistencyOffsetVector.java
vvcephei: Yeah, I think a good further cleanup would be to drop that interface. It seems to be causing more harm than good, and it also doesn't seem like we have discovered any desire to actually unify the continuation tokens of push queries with the consistency tokens of pull queries.

Since this is a private API, though, it seems better to just merge this PR and have another follow-up to drop the interface.

vvcephei (approved):

patrickstuedi (approved):
@vpapavas I'm fine with merging the PR as is, my previous comments were more like proposals, feel free to pick them in a follow-up or not at all.

ksqldb-common/src/main/java/io/confluent/ksql/util/ConsistencyOffsetVector.java
vpapavas: Thanks both, I will keep the map for now

PR #8819, opened 2022-02-26
fix: use blocking call and throw exception on failure

wcarlson5:
### Description 
In the validation runtime we should use the blocking call so that if the future gets completed exceptionally we can fail validation. Also for COR queries we are going to validate stoping and and re-adding the topology.

Also this improves the failure message.

##

### Reviewer checklist

### DIFF
--- a/ksqldb-engine/src/main/java/io/confluent/ksql/util/SandboxedSharedKafkaStreamsRuntimeImpl.java
+++ b/ksqldb-engine/src/main/java/io/confluent/ksql/util/SandboxedSharedKafkaStreamsRuntimeImpl.java
@@ -21,8 +21,8 @@ import io.confluent.ksql.util.QueryMetadataImpl.TimeBoundedQueue;
 import java.time.Duration;
 import java.util.Map;
 import java.util.concurrent.ConcurrentHashMap;
-import java.util.stream.Collectors;
 import java.util.concurrent.ExecutionException;
+import java.util.stream.Collectors;
 import org.apache.kafka.streams.StreamsConfig;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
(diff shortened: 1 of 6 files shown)


wcarlson5:
This can't be merged until blocking calls in created state make some sort of sense. @ableegoldman I think had some ideas on how to do that

ksqldb-engine/src/main/java/io/confluent/ksql/util/SandboxedSharedKafkaStreamsRuntimeImpl.java
```diff
@@ -74,15 +75,30 @@ public void register(
     log.info("Registering query {} for validation for runtime {}", queryId, getApplicationId());
     collocatedQueries.put(queryId, binpackedPersistentQueryMetadata);
     try {
-      if (!kafkaStreams.getTopologyByName(queryId.toString()).isPresent()) {
-        kafkaStreams.addNamedTopology(binpackedPersistentQueryMetadata.getTopologyCopy(this));
+      if (!kafkaStreams.getTopologyByName(
+          binpackedPersistentQueryMetadata.getQueryId().toString()).isPresent()) {
```
guozhangwang, line 79: I think the reason we passed in the second param `queryId` is that, except for the non-testing caller, `start()` which just pass in itself and the `queryId` field, some testing code used a mock and hence cannot get `binpackedPersistentQueryMetadata.getQueryId()` from the mock, right? If not, then we do not need this second param in `register`, but if yes, then it seems we should not replace `queryId` to `binpackedPersistentQueryMetadata.getQueryId()`.

ksqldb-engine/src/main/java/io/confluent/ksql/util/SandboxedSharedKafkaStreamsRuntimeImpl.java
```diff
@@ -74,15 +75,30 @@ public void register(
...
+        kafkaStreams.addNamedTopology(binpackedPersistentQueryMetadata.getTopologyCopy(this))
+            .all()
+            .get();
+      } else {
+        kafkaStreams.removeNamedTopology(queryId.toString(), false).all().get();
```
guozhangwang, line 84: nit: we can refactor this as

```
if (kafkaStreams.getTopologyByName..isPresent()) {
   // log.info("Replacing an existing query with id {}");
   kafkaStreams.removeNamedTopology..get();
}

kafkaStreams.addNamedTopology..get();
```

ksqldb-engine/src/main/java/io/confluent/ksql/util/SandboxedSharedKafkaStreamsRuntimeImpl.java
```diff
@@ -74,15 +75,30 @@ public void register(
...
+        kafkaStreams.addNamedTopology(binpackedPersistentQueryMetadata.getTopologyCopy(this))
+            .all()
+            .get();
       }
-    } catch (final Throwable e) {
+    } catch (final ExecutionException e) {
+      final Throwable t = e.getCause() == null ? e : e.getCause();
```
guozhangwang, line 90: Another nit: I think we can use one `catch` still with the last parameter as

```
(e instanceof ExecutionException && e.getCause() != null) ? e.getCause() : e
```

ksqldb-engine/src/main/java/io/confluent/ksql/util/SandboxedSharedKafkaStreamsRuntimeImpl.java
wcarlson5: I don't remember why we took the query ID as a parameter. I will try removing it

ableegoldman (approved):
Nice, LGTM

PR #8822, opened 2022-02-28
chore: Split the CI build into two parallel jobs.

jnh5y:
### Description 
The goal is to speed up the Jenkins CI job.

### Testing done
Manually checking out the CI job to make sure that it is running things sanely and is faster.

With this change, CI runs in around 70 minutes instead of over two hours.


Jenkinsfile
```diff
@@ -1,16 +1,32 @@
...
-    extraDeployArgs = '-Ddocker.skip=true'
-    dockerPush = false
-    dockerScan = false
-    dockerImageClean = false
-    downStreamRepos = ["confluent-security-plugins", "confluent-cloud-plugins"]
-    nanoVersion = true
-    maxBuildsToKeep = 99
-    maxDaysToKeep = 90
-    extraBuildArgs = "-Dmaven.gitcommitid.nativegit=true"
+parallel unitTests: { // Skip the integration tests.  The unit tests take longer the the ITs, so we push the spotbugs and Maven site goals to the other job.
```
suhas-satish, line 3: typo: longer than. After this change, the integration tests will take longer than the unit tests right, so should we revise the comment?

suhas-satish (approved):

PR #8823, opened 2022-02-28
chore: Fixing QueryAnonymizerTest so that it runs its enclosed class.

jnh5y:
### Description 
Run the inner class's tests.

### Testing done 
I found this will poking at the build generally.  To verify the fix, I ran the test manually in IntelliJ.  Without the change, the enclosed class's tests do not run.

### DIFF
--- a/ksqldb-functional-tests/src/test/java/io/confluent/ksql/test/QueryAnonymizerTest.java
+++ b/ksqldb-functional-tests/src/test/java/io/confluent/ksql/test/QueryAnonymizerTest.java
@@ -13,6 +13,7 @@
 package io.confluent.ksql.test;
 
 import static java.nio.charset.StandardCharsets.UTF_8;
+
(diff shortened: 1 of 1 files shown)


ksqldb-functional-tests/src/test/java/io/confluent/ksql/test/QueryAnonymizerTest.java
```diff
@@ -13,32 +13,34 @@
...
 import java.util.stream.Stream;
+import org.approvaltests.Approvals;
 import org.junit.Assert;
 import org.junit.Before;
 import org.junit.Test;
+import org.junit.experimental.runners.Enclosed;
 import org.junit.runner.RunWith;
 import org.junit.runners.Parameterized;
 
+@RunWith(Enclosed.class)
```
suhas-satish, line 43: thanks for the fix, as per the docs , abstract inner classes are still ignored. But it should work for this case.

PR #8830, opened 2022-03-02
fix: Set the sslFactory properly for the SR REST client.

jnh5y:
### Description 

### Testing done 
Manual testing is in progress.

### Reviewer checklist

### DIFF
--- a/ksqldb-engine/src/main/java/io/confluent/ksql/schema/registry/KsqlSchemaRegistryClientFactory.java
+++ b/ksqldb-engine/src/main/java/io/confluent/ksql/schema/registry/KsqlSchemaRegistryClientFactory.java
@@ -115,11 +115,8 @@ public class KsqlSchemaRegistryClientFactory {
     }
   
     final RestService restService = serviceSupplier.get();
-    if (sslContext != null) {
-      restService.setSslSocketFactory(sslContext.getSocketFactory());
-    }
-
-    return schemaRegistryClientFactory.create(
+    // This call sets a default sslSocketFactory.
+    final SchemaRegistryClient client = schemaRegistryClientFactory.create(
         restService,
         1000,
         ImmutableList.of(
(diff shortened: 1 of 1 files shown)


ksqldb-engine/src/main/java/io/confluent/ksql/schema/registry/KsqlSchemaRegistryClientFactory.java
```diff
@@ -115,17 +115,18 @@ public SchemaRegistryClient get() {
...
-    return schemaRegistryClientFactory.create(
+    final SchemaRegistryClient client = schemaRegistryClientFactory.create(
         restService,
         1000,
         ImmutableList.of(
             new AvroSchemaProvider(), new ProtobufSchemaProvider(), new JsonSchemaProvider()),
         schemaRegistryClientConfigs,
         httpHeaders
     );
+    // Let's reset the sslContext...
```
suhas-satish, line 126: just for my understanding, what happens if this sslContext is not reset?

jnh5y: The code above `schemaRegistryClientFactory.create(...)` calls https://github.com/confluentinc/schema-registry/blob/master/client/src/main/java/io/confluent/kafka/schemaregistry/client/CachedSchemaRegistryClient.java#L181-L232

The last lines of that method call `restService.setSslSocketFactory(...)`...  effectively, that undoes the code that was setting the factory before we called the constructor.

spena (approved):
Looks good. Can you just make a comment in the code the reason to reset the SSL factory? In case we look at it in the future.

jnh5y:
> Looks good. Can you just make a comment in the code the reason to reset the SSL factory? In case we look at it in the future.

Yes, will do.

Issue #8816, opened 2022-02-25
pull query deadlock in multi-server configuration


agavra:
**Describe the bug**

![image](https://user-images.githubusercontent.com/3172405/155774638-d31d69ee-c827-4171-8ba5-4c8aa055cf11.png)

**To Reproduce**

Reproduced on 0.23.3

1. spin up two servers locally with different ports and streams dirs and set the ksql.worker.pool.size=1. make sure one of them is attached with a debugger
2. put a breakpoint somewhere on the query execution path for Server A (this simulates a "long running" query)
3. create a queryable table that has multiple partitions
4. issue two table scans to server A (`curl -H 'Content-Type: json' -X POST -d '{"ksql":"SELECT * FROM bar;"}' http://localhost:8089/query`)
5. issue a table scan to server B
6. resume the breakpoint from server A
7. the servers become bricked

**Expected behavior**

When we resume the breakpoint from server A progress is made

**Actual behaviour**

The servers become entirely deadlocked

**Additional context**

nateab:
Good news, I was able to reproduce this on 0.23.3 but NOT on 0.24.0 which is promising at least

agavra:
@patrickstuedi will time box some investigation here, otherwise we'll pull it into the HA rewrite that we're planning on doing in a bit.

anupsdtiwari:
@agavra @nateab 

Guys - I believe this issue is still there even in 0.25.1.

I have also raise a ticket for this https://github.com/confluentinc/ksql/issues/9161

Till this issue gets fixed, do we have any workaround ? 
Because we are getting this issue everytime we go RPS > 500 in 3 node cluster.

RPS = 500 is working because we have changed "ksql.query.pull.thread.pool.size=2000" and "ksql.query.pull.router.thread.pool.size=2000" but even changing to higher values are not working for RPS > 500.

Also due to this issue it seems that we can't use ksqlDB in production.

agavra:
@anupsdtiwari are you hitting the deadlock? if so do you mind posting thread dumps from your three node clusters here? that will help debug it

anupsdtiwari:
@agavra 

One more observation - I can see that changing both thread.pool.size directly impacts pull query performance.
i.e. if i set it 1000 then i can see ksql is able to serve 4500 rps under 100ms and if i increase it to 4000 then performance gets degraded.

[...]

PR #8833, opened 2022-03-03
chore: avoid deadlocking in HARouting

patrickstuedi:
Attempt to mitigate this problem: https://github.com/confluentinc/ksql/issues/8816


ksqldb-engine/src/main/java/io/confluent/ksql/physical/pull/HARouting.java
```diff
@@ -200,7 +202,7 @@ private void executeRounds(
     while (processedPartitions < totalPartitions) {
       final Future<PartitionFetchResult> future = completionService.take();
       try {
-        final PartitionFetchResult fetchResult = future.get();
+        final PartitionFetchResult fetchResult = future.get(10, TimeUnit.SECONDS);
```
agavra, line 205: should we make this configurable per-query in case people have really large table scans?

ksqldb-engine/src/main/java/io/confluent/ksql/physical/pull/HARouting.java
```diff
@@ -219,6 +221,16 @@ private void executeRounds(
           Preconditions.checkState(fetchResult.getResult() == RoutingResult.SUCCESS);
           processedPartitions++;
         }
+      } catch (TimeoutException timeoutException) {
+        final MaterializationException exception =
+            new MaterializationException(
+            "Query timeout exception");
```
agavra, line 227: would it help to specify which node timed out so they can debug issues with connectivity etc...?

ksqldb-engine/src/main/java/io/confluent/ksql/physical/pull/HARouting.java
```diff
@@ -219,6 +221,16 @@ private void executeRounds(
...
+        for (Entry<Integer, List<Exception>> entry: exceptionsPerPartition.entrySet()) {
+          for (Exception excp: entry.getValue()) {
+            exception.addSuppressed(excp);
+          }
+        }
+        throw exception;
```
agavra, line 233: is this something we could easily build a test for? (I guess re my first comment about configuration would make it easier to test)

agavra (commented):
Thanks Patrick! Did you try reproducing the issue with/without this fix?

guozhangwang (commented):
A meta comment: I think for such live lock situations a configured, fixed value may not be effective mitigating if both sides are using the same timeout / backoff values. Maybe a randomized timeout (or backoff) would be more effective to avoid consecutive live locks.

ksqldb-engine/src/main/java/io/confluent/ksql/physical/pull/HARouting.java
patrickstuedi: Yes we should, thanks for pointing out

ksqldb-engine/src/main/java/io/confluent/ksql/physical/pull/HARouting.java
```diff
@@ -200,7 +203,11 @@ private void executeRounds(
     while (processedPartitions < totalPartitions) {
       final Future<PartitionFetchResult> future = completionService.take();
       try {
-        final PartitionFetchResult fetchResult = future.get();
+        final long fetchTimeout = ThreadLocalRandom.current().nextLong(
+            KsqlConfig.KSQL_QUERY_PULL_FETCH_TIMEOUT_DEFAULT - 1,
```
agavra, line 207: nit: we should make sure this isn't negative, right?

patrickstuedi: Right, I was wondering if we have something like a verification check for the config variables. For instance, instead of checking for a negative value each time at this position one could require the timeout config value to be > 0 when parsing the config. I didn't spot anything like this though.

PR #8835, opened 2022-03-03
fix: turn on shared runtime flag

wcarlson5:
### Description 
I am turning on the feature flag for shared runtimes by fault, I also found a couple tests that need to updated/ looked at.

### Testing done 

### Reviewer checklist

### DIFF
--- a/ksqldb-cli/src/test/java/io/confluent/ksql/cli/CliTest.java
+++ b/ksqldb-cli/src/test/java/io/confluent/ksql/cli/CliTest.java
@@ -241,6 +241,17 @@ public class CliTest {
     assertThat(rowCaptor.getRows(), matcher);
   }
 
+  private void assertCommandFail(
+      final String command,
+      final Matcher<Iterable<? extends Iterable<? extends String>>> matcher
+  ) {
+    rowCaptor.resetTestResult();
+    assertThrows(
+        IllegalArgumentException.class,
+        () -> localCli.handleLine(command)
+    );
+  }
+
   private static void run(final String command, final Cli localCli) {
     try {
       localCli.handleLine(command);
@@ -442,27 +453,10 @@ public class CliTest {
   }
 
   @Test
-  public void testDisableVariableSubstitution() {
-    // Given:
-    assertRunCommand(
-        "set '" + KsqlConfig.KSQL_VARIABLE_SUBSTITUTION_ENABLE + "' = 'false';", is(EMPTY_RESULT));
-    assertRunCommand("define topicName = '" + DELIMITED_TOPIC + "';", is(EMPTY_RESULT));
-
-    // When:
-    run("PRINT ${topicName} FROM BEGINNING INTERVAL 1 LIMIT 2;", localCli);
-
-    // Then:
-    assertThatEventually(() -> terminal.getOutputString(),
-        containsString("Failed to Describe Kafka Topic(s): [${topicName}]"));
-    assertThatEventually(() -> terminal.getOutputString(),
-        containsString("Caused by: The request attempted to perform an operation on an invalid topic."));
-  }
-
-  @Test
   public void testVariableSubstitution() {
     // Given:
     assertRunCommand(
-        "set '" + KsqlConfig.KSQL_VARIABLE_SUBSTITUTION_ENABLE + "' = 'true';", is(EMPTY_RESULT));
+        "ALTER SYSTEM '" + KsqlConfig.KSQL_VARIABLE_SUBSTITUTION_ENABLE + "' = 'true';", is(EMPTY_RESULT));
     assertRunCommand("define topicName = '" + DELIMITED_TOPIC + "';", is(EMPTY_RESULT));
 
     // When:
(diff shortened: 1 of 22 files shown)


wcarlson5:
@ableegoldman we should look into these tests I marked. It seems that a few need to updated

cc @guozhangwang

ksqldb-engine/src/test/java/io/confluent/ksql/util/SandboxedSharedKafkaStreamsRuntimeImplTest.java
```diff
@@ -71,7 +71,6 @@ public void setUp() throws Exception {
     when(queryId.toString()).thenReturn("query 1");
 
     when(binPackedPersistentQueryMetadata.getTopologyCopy(any())).thenReturn(topology);
-    when(binPackedPersistentQueryMetadata.getQueryId()).thenReturn(queryId);
```
guozhangwang, line 74: This is no needed in this test?

guozhangwang (commented):
Just to clarify, the ignored tests are all that will fail if we turn it on?

wcarlson5:
@guozhangwang yes, these are the tests that will fail

ksqldb-engine/src/test/java/io/confluent/ksql/integration/WindowingIntTest.java
```diff
@@ -125,6 +126,7 @@ public void before() {
   }
 
   @Test
+  @Ignore
```
wcarlson5, line 129: It looks like we are not cleaning up repartition topics, we should be doing this in `QueryCleanupService` but it doesn't seem to be picking them up.

ksqldb-engine/src/test/java/io/confluent/ksql/integration/SecureIntegrationTest.java
```diff
@@ -307,6 +309,7 @@ public void shouldClassifyMissingSourceTopicExceptionAsUserErrorSharedRuntimes()
   }
 
   @Test
+  @Ignore
```
wcarlson5, line 312: This and the one below do get classified as user errrors and do happen. But as the error occurs in streams and happens after the topology gets added the error is not thrown up to the statement response. The error is in the error queue for the query though

ksqldb-engine/src/test/java/io/confluent/ksql/integration/DependentStatementsIntegrationTest.java
```diff
@@ -131,6 +132,7 @@ public void shouldRegisterAvroSchemaInSandboxViaCSAS() {
   }
 
   @Test
+  @Ignore
```
wcarlson5, line 135: This is a new test (in the last month) it is being caught by out check for non query level configs. Some how KEY_SCHEMA_ID and VALUE_SCHEMA_ID are making their way into overridden properties. 

more details: [here](https://confluent.slack.com/archives/C01SDQ5Q8HL/p1646337759343309)

ksqldb-engine/src/test/java/io/confluent/ksql/integration/SecureIntegrationTest.java
```diff
@@ -175,6 +176,7 @@ public void after() {
...
```
wcarlson5, line 179: This is failing to create the topic and might be an issue. I have no idea why though

wcarlson5: It seems that query is getting created and not until the `after` the query is getting added and started. However when its being closed we hit 
```[2022-03-03 16:09:18,143] INFO Principal = User:ANONYMOUS is Denied Operation = Delete from host = 192.168.1.97 on resource = Group:LITERAL:_confluent-ksql-default_query_0 for request = OffsetDelete with resourceRefCount = 1 (kafka.authorizer.logger:614)
```

wcarlson5: @rodesai Do you have an idea of what is causing this?

rodesai: It looks like the test is trying to delete offsets for the group as part of running the query. Seems like it could be a legitimate bug to me. Have you tried setting a breakpoint in `deleteConsumerGroupOffsets` and seeing what's calling that?

wcarlson5: We delete offsets when closing a query. We need to inorder to make that runtime usable again. What other permissions do we need for that other than the SSL the test is giving us?

rodesai: not quite sure - but you need to configure ACLs explicitly. Can you try changing the line that grants describe/read acls on groups to:
```
givenAllowAcl(ALL_USERS,
                  resource(GROUP, ResourcePattern.WILDCARD_RESOURCE),
                  ops(DESCRIBE, READ, DELETE));
```

wcarlson5: Yeah that fixed the test. Do we need to make sure that permission is given or is this just a test issue?

ksqldb-engine/src/test/java/io/confluent/ksql/integration/DependentStatementsIntegrationTest.java
wcarlson5: Test is fixed in this PR https://github.com/confluentinc/ksql/pull/8869, will get picked to 0.25

wcarlson5:
All the tests are passing now. We should merge this as soon as 0.25 gets its own branches created

wcarlson5:
All the tests are passing now. We should merge this as soon as 0.25 gets its own branches created

rodesai (approved):
LGTM

ksqldb-cli/src/test/java/io/confluent/ksql/cli/CliTest.java
```diff
@@ -441,28 +452,11 @@ public void shouldPrintTopicWithDelimitedValue() {
         containsString(", key: ITEM_1, value: home cinema"));
   }
 
-  @Test
-  public void testDisableVariableSubstitution() {
-    // Given:
-    assertRunCommand(
```
wcarlson5, line 447: No longer an editable cmd

Issue #7211, opened 2021-03-11
Cannot INSERT VALUES into data source. ksqlDB generated schema would overwrite existing key schema.


rmoff:
ksqlDB 0.15

Hitting a problem with a complex key, see below.

```sql
-- Create a stream with a complex key
CREATE STREAM FOO_08 (COL1 VARCHAR KEY, COL2 INT KEY, COL3 VARCHAR, COL4 VARCHAR)
  WITH (KAFKA_TOPIC='FOO_08', VALUE_FORMAT='AVRO', KEY_FORMAT='AVRO', PARTITIONS=1);

-- Insert a row - this works fine
INSERT INTO FOO_08 (COL1, COL2, COL3, COL4) VALUES ('mykey_val_A',1,'NEVER','GONNA');

-- Show the topic
PRINT FOO_08 FROM BEGINNING LIMIT 1;

-- Insert another row; the fails :-(
INSERT INTO FOO_08 (COL1, COL2, COL3, COL4) VALUES ('mykey_val_A',2,'GIVE','YOU');

-- Try to insert the original row value again - also fails
INSERT INTO FOO_08 (COL1, COL2, COL3, COL4) VALUES ('mykey_val_A',1,'NEVER','GONNA');
```

Console output: 

```
ksql> -- Create a stream with a complex key
>CREATE STREAM FOO_08 (COL1 VARCHAR KEY, COL2 INT KEY, COL3 VARCHAR, COL4 VARCHAR)
>  WITH (KAFKA_TOPIC='FOO_08', VALUE_FORMAT='AVRO', KEY_FORMAT='AVRO', PARTITIONS=1);
>

 Message
----------------
 Stream created
----------------
ksql> -- Insert a row - this works fine
>INSERT INTO FOO_08 (COL1, COL2, COL3, COL4) VALUES ('mykey_val_A',1,'NEVER','GONNA');
>
ksql> -- Show the topic
>PRINT FOO_08 FROM BEGINNING LIMIT 1;
>
Key format: AVRO or HOPPING(KAFKA_STRING) or TUMBLING(KAFKA_STRING) or KAFKA_STRING
Value format: AVRO or KAFKA_STRING
rowtime: 2021/03/11 21:44:01.275 Z, key: {"COL1": "mykey_val_A", "COL2": 1}, value: {"COL3": "NEVER", "COL4": "GONNA"}, partition: 0
Topic printing ceased
ksql> -- Insert another row; the fails :-(
[...]

PR #8839, opened 2022-03-03
chore: ignore connect name when comparing keyschema

lihaosky:
### Description 
fixes https://github.com/confluentinc/ksql/issues/7211 

Ignore comparing `connect.name` when inserting. I tried to remove https://github.com/confluentinc/ksql/blob/master/ksqldb-serde/src/main/java/io/confluent/ksql/serde/avro/KsqlAvroSerdeFactory.java#L171 and it breaks ~hundreds~ (thousand) of history QTT tests...

### Testing done 
Unit and QTT

### Reviewer checklist

### DIFF
--- a/ksqldb-functional-tests/src/test/resources/query-validation-tests/multi-col-keys.json
+++ b/ksqldb-functional-tests/src/test/resources/query-validation-tests/multi-col-keys.json
@@ -345,26 +345,6 @@
         "type": "io.confluent.ksql.util.KsqlStatementException",
         "message": "Invalid join condition: stream-table joins require to join on the table's primary key. Got S.K = T.K."
       }
-    },
-    {
-      "name": "multi-key insert - AVRO",
-      "statements": [
-        "CREATE STREAM INPUT (COL1 VARCHAR KEY, COL2 INT KEY, COL3 VARCHAR) WITH (kafka_topic='input_topic', VALUE_FORMAT='AVRO', KEY_FORMAT='AVRO', PARTITIONS=1);",
-        "CREATE STREAM OUTPUT AS SELECT * FROM INPUT;"
-      ],
-      "inputs": [
-        {"topic": "input_topic", "key": {"COL1": "1", "COL2": 1}, "value": {"COL3": "1"}},
-        {"topic": "input_topic", "key": {"COL1": "2", "COL2": 2}, "value": {"COL3": "2"}}
-      ],
-      "outputs": [
-        {"topic": "OUTPUT", "key": {"COL1": "1", "COL2": 1}, "value": {"COL3": "1"}},
-        {"topic": "OUTPUT", "key": {"COL1": "2", "COL2": 2}, "value": {"COL3": "2"}}
-      ],
-      "post": {
-        "sources": [
-          {"name": "INPUT", "type": "stream", "schema": "COL1 VARCHAR KEY, COL2 INT KEY, COL3 VARCHAR"}
-        ]
-      }
     }
   ]
 }
--- a/ksqldb-functional-tests/src/test/resources/query-validation-tests/multi-col-keys.json
+++ b/ksqldb-functional-tests/src/test/resources/query-validation-tests/multi-col-keys.json
@@ -345,6 +345,26 @@
         "type": "io.confluent.ksql.util.KsqlStatementException",
         "message": "Invalid join condition: stream-table joins require to join on the table's primary key. Got S.K = T.K."
       }
+    },
+    {
+      "name": "multi-key insert - AVRO",
+      "statements": [
+        "CREATE STREAM INPUT (COL1 VARCHAR KEY, COL2 INT KEY, COL3 VARCHAR) WITH (kafka_topic='input_topic', VALUE_FORMAT='AVRO', KEY_FORMAT='AVRO', PARTITIONS=1);",
+        "CREATE STREAM OUTPUT AS SELECT * FROM INPUT;"
+      ],
+      "inputs": [
+        {"topic": "input_topic", "key": {"COL1": "1", "COL2": 1}, "value": {"COL3": "1"}},
+        {"topic": "input_topic", "key": {"COL1": "2", "COL2": 2}, "value": {"COL3": "2"}}
+      ],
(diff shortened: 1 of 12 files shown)


jnh5y:
Would it make sense to add one or more QTTs to cover examples in one or both of those tickets?  (I don't quite grok this issue, so someone else probably should actually give an approval.)

ksqldb-functional-tests/src/test/resources/query-validation-tests/multi-col-keys.json
```diff
@@ -345,6 +345,26 @@
         "type": "io.confluent.ksql.util.KsqlStatementException",
         "message": "Invalid join condition: stream-table joins require to join on the table's primary key. Got S.K = T.K."
       }
+    },
+    {
+      "name": "multi-key insert - AVRO",
```
agavra, line 350: what does this test that isn't already tested? I don't see any `INSERT` in this test?

ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/execution/InsertValuesExecutor.java
```diff
@@ -348,6 +352,21 @@ private static void ensureKeySchemasMatch(
     }
 
     if (latest.isPresent() && !latest.get().getSchema().equals(schema.canonicalString())) {
+      // Hack: skip comparing connect name. See https://github.com/confluentinc/ksql/issues/7211
+      // Avro schema are registered in source creation time as well data insertion time.
+      // CONNECT_META_DATA_CONFIG is configured to false in Avro Serializer, but it's true in
+      // AvroSchemaTranslator. It needs to be true to make ConnectSchema map type work. But
+      // enabling it breaks lots of history QTT test which implies backward compatibility issues.
```
agavra, line 359: we should verify that this actually causes compatibility issues before committing a "hack". if it isn't actually a compatibility issue we should just rewrite the old files.

ksqldb-serde/src/main/java/io/confluent/ksql/serde/protobuf/ProtobufSchemaTranslator.java
```diff
@@ -31,6 +32,10 @@
   private final ProtobufData protobufData =
       new ProtobufData(new ProtobufDataConfig(ImmutableMap.of()));
 
+  @Override
+  public void configure(final Map<String, ?> configs) {
```
agavra, line 36: should we go ahead and recreate the probotuf data with the configs?

agavra (commented):
just a quick pass, you probably should get an engineer from KSE to look at this as well :)

ksqldb-functional-tests/src/test/resources/query-validation-tests/multi-col-keys.json
lihaosky: I was under the wrong impression input was doing insertion. Does QTT support insertion? Or I should use YATT?

ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/execution/InsertValuesExecutor.java
lihaosky: If `connect` meta is enabled during serialization, output schema will have connect meta. QTT test failures are from historical ones like: https://github.com/confluentinc/ksql/blob/master/ksqldb-functional-tests%2Fsrc%2Ftest%2Fresources%2Fhistorical_plans%2Fdecimal_-_mod%2F5.5.0_1581572087725%2Fspec.json#L186-L209

New schema for this case will be:
```
{
    "name":"TEST2",
    "keyFormat":{
        "format":"KAFKA"
    },
    "valueFormat":{
        "format":"AVRO"
    },
    "partitions":4,
    "valueSchema":{
        "type":"record",
        "name":"KsqlDataSourceSchema",
        "namespace":"io.confluent.ksql.avro_schemas",
        "fields":[
            {
                "name":"RESULT",
                "type":[
                    "null",
                    {
                        "type":"bytes",
                        "scale":2,
                        "precision":4,
                        "connect.version":1,
                        "connect.parameters":{
                            "scale":"2",
                            "connect.decimal.precision":"4"
                        },
                        "connect.name":"org.apache.kafka.connect.data.Decimal",
                        "logicalType":"decimal"
                    }
                ],
                "default":null
            }
        ],
        "connect.name":"io.confluent.ksql.avro_schemas.KsqlDataSourceSchema"
    }
}
```

This one seems fine but there are ~1000 QTT failures and I'm not sure if they are all ok... What's the best way to double check and if we are going to rewrite old files, how to do that?

lihaosky (commented):
Hi @vcrfxia , can you take a look? Thanks!

ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/execution/InsertValuesExecutor.java
vcrfxia: Is the only change you're expecting to see the addition of the `connect.name` field in Avro schemas? What about the other metadata mentioned in the [docs for `connect.meta.data`](https://docs.confluent.io/platform/current/schema-registry/connect.html) such as doc, version, type, etc? Are those unchanged because ksql doesn't use them?

ksqldb-functional-tests/src/test/resources/query-validation-tests/multi-col-keys.json
vcrfxia: QTT does not test INSERT VALUES. RQTT does, though: https://github.com/confluentinc/ksql/blob/master/ksqldb-functional-tests/src/test/resources/rest-query-validation-tests/insert-values.json

lihaosky:
Discussed with @vcrfxia offline. Fixing the root cause is not trivial as it's unclear whether it's backward compatible and we'd ideally unify the schema generation logic during translation and serialization. We can use this hack to fix https://github.com/confluentinc/ksql/issues/7211 and I'll create another issue tracking the general fix.

PR #8845, opened 2022-03-04
feat: Add support for Stream-Stream and Table-Table right join

### LINKED ISSUE #8478: Support verbose join syntax
SQL supports inner/left/right/outer joins including verbose and short syntax. We should consider to support the full verbose syntax, instead of just short syntax.

We should support
- `a [INNER] JOIN b ON`
- `a LEFT [OUTER] JOIN b ON`
- `a RIGHT [OUTER] JOIN b ON` (we don't support `RIGHT` join at all atm)
- `a FULL [OUTER] JOIN b ON`

Gerrrr:
### Description 

This patch adds support right join for Stream-Stream and Table-Table. Both operations are essentially syntactical sugar - internally they perform a left join changing the order of operands.

### Testing done 

* Integration tests
* Unit tests

### Reviewer checklist

### DIFF
--- a/ksqldb-engine/src/main/java/io/confluent/ksql/planner/plan/JoinNode.java
+++ b/ksqldb-engine/src/main/java/io/confluent/ksql/planner/plan/JoinNode.java
@@ -69,7 +69,7 @@ import org.apache.kafka.clients.consumer.ConsumerConfig;
 public class JoinNode extends PlanNode implements JoiningNode {
 
   public enum JoinType {
-    INNER, LEFT, OUTER;
+    INNER, LEFT, RIGHT, OUTER;
 
     @Override
     public String toString() {
@@ -78,6 +78,8 @@ public class JoinNode extends PlanNode implements JoiningNode {
           return "[INNER] JOIN";
         case LEFT:
           return "LEFT [OUTER] JOIN";
+        case RIGHT:
+          return "RIGHT [OUTER] JOIN";
         case OUTER:
           return "[FULL] OUTER JOIN";
         default:
@@ -433,6 +435,15 @@ public class JoinNode extends PlanNode implements JoiningNode {
               JoiningNode.getValueFormatForSource(joinNode.right).getFormatInfo(),
               contextStacker
           );
+        case RIGHT:
+          return leftStream.rightJoin(
+              rightStream,
+              joinNode.getKeyColumnName(),
+              joinNode.withinExpression.get(),
+              JoiningNode.getValueFormatForSource(joinNode.left).getFormatInfo(),
+              JoiningNode.getValueFormatForSource(joinNode.right).getFormatInfo(),
+              contextStacker
+          );
         case OUTER:
           return leftStream.outerJoin(
               rightStream,
@@ -540,6 +551,16 @@ public class JoinNode extends PlanNode implements JoiningNode {
             );
           } else {
             return leftTable.leftJoin(
+                rightTable,
+                joinNode.getKeyColumnName(),
+                contextStacker
+            );
+          }
+        case RIGHT:
+          if (joinKey.isForeignKey()) {
+            throw new KsqlException("RIGHT OUTER JOIN on a foreign key is not supported");
+          } else {
+            return leftTable.rightJoin(
                 rightTable,
                 joinNode.getKeyColumnName(),
                 contextStacker
--- a/ksqldb-engine/src/test/java/io/confluent/ksql/engine/rewrite/DataSourceExtractorTest.java
+++ b/ksqldb-engine/src/test/java/io/confluent/ksql/engine/rewrite/DataSourceExtractorTest.java
@@ -149,7 +149,7 @@ public class DataSourceExtractorTest {
   }
 
   @Test
-  public void shouldThrowIfRightJoinSourceDoesNotExist() {
+  public void shouldThrowIfInnerJoinSourceDoesNotExist() {
     // Given:
     final AstNode stmt = givenQuery("SELECT * FROM TEST1 JOIN UNKNOWN"
         + " ON test1.col1 = UNKNOWN.col1;");
--- a/ksqldb-functional-tests/src/test/resources/query-validation-tests/joins.json
+++ b/ksqldb-functional-tests/src/test/resources/query-validation-tests/joins.json
@@ -509,7 +509,7 @@
       "statements": [
         "CREATE STREAM S_LEFT (ID BIGINT KEY, NAME varchar) WITH (kafka_topic='left_topic', value_format='KAFKA');",
         "CREATE STREAM S_RIGHT (ID BIGINT KEY, NAME varchar) WITH (kafka_topic='right_topic', value_format='KAFKA');",
-        "CREATE STREAM OUTPUT WITH(value_format='delimited') as SELECT * FROM s_left join s_right WITHIN 1 second ON s_left.id = s_right.id;"
+        "CREATE STREAM OUTPUT WITH(value_format='delimited') as SELECT * FROM s_left LEFT JOIN s_right WITHIN 1 second ON s_left.id = s_right.id;"
       ],
       "expectedException": {
         "type": "io.confluent.ksql.util.KsqlStatementException",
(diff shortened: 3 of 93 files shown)


ksqldb-functional-tests/src/test/resources/query-validation-tests/joins.json
```diff
@@ -648,7 +648,205 @@
...
+        {"topic": "right_topic", "key": 0, "value": {"F1": "a", "F2": 10}, "timestamp": 15000},
+        {"topic": "right_topic", "key": 100, "value": {"F1": "newblah", "F2": 150}, "timestamp": 16000},
+        {"topic": "left_topic", "key": 90, "value": {"NAME": "ninety", "VALUE": 90}, "timestamp": 17000},
+        {"topic": "left_topic", "key": 0, "value": {"NAME": "bar", "VALUE": 99}, "timestamp": 30000}
+      ],
+      "outputs": [
+        {"topic": "OUTPUT", "key": 0, "value": {"NAME": "zero", "VALUE": 0, "F1": "blah", "F2": 50}, "timestamp": 10000},
+        {"topic": "OUTPUT", "key": 0, "value": {"NAME": "foo", "VALUE": 100, "F1": "blah", "F2": 50}, "timestamp": 13000},
+        {"topic": "OUTPUT", "key": 0, "value": {"NAME": "foo", "VALUE": 100, "F1": "a", "F2": 10}, "timestamp": 15000},
+        {"topic": "OUTPUT", "key": 100, "value": {"NAME": "", "VALUE": 0, "F1": "newblah", "F2": 150}, "timestamp": 16000}
```
mjsax, line 706: Should both `NAME` and `VALUE` not be `null` ?

ksqldb-functional-tests/src/test/resources/query-validation-tests/joins.json
```diff
@@ -648,7 +648,205 @@
...
+        {"topic": "right_topic", "key": "foo", "value": {"ID": 0, "F1": "a", "F2": 10}, "timestamp": 15000},
+        {"topic": "right_topic", "key": "foo", "value": {"ID": 100, "F1": "newblah", "F2": 150}, "timestamp": 16000},
+        {"topic": "left_topic", "key": "foo", "value": {"ID": 90, "NAME": "ninety", "VALUE": 90}, "timestamp": 17000},
+        {"topic": "left_topic", "key": "foo", "value": {"ID": 0, "NAME": "bar", "VALUE": 99}, "timestamp": 30000}
+      ],
+      "outputs": [
+        {"topic": "OUTPUT", "key": 0, "value": {"T_K": "foo", "NAME": "zero", "VALUE": 0, "F1": "blah", "F2": 50}, "timestamp": 10000},
+        {"topic": "OUTPUT", "key": 0, "value": {"T_K": "foo", "NAME": "foo", "VALUE": 100, "F1": "blah", "F2": 50}, "timestamp": 13000},
+        {"topic": "OUTPUT", "key": 0, "value": {"T_K": "foo", "NAME": "foo", "VALUE": 100, "F1": "a", "F2": 10}, "timestamp": 15000},
+        {"topic": "OUTPUT", "key": 100, "value": {"T_K": "", "NAME": "", "VALUE": 0, "F1": "newblah", "F2": 150}, "timestamp": 16000}
```
mjsax, line 781: As above. Why `""` and `0` instead of `null`?

ksqldb-functional-tests/src/test/resources/query-validation-tests/joins.json
```diff
@@ -648,7 +648,205 @@
...
+        {"topic": "right_topic", "value": {"ID": 0, "F1": "a", "F2": 10}, "timestamp": 15000},
+        {"topic": "right_topic", "value": {"ID": 100, "F1": "newblah", "F2": 150}, "timestamp": 16000},
+        {"topic": "left_topic", "value": {"ID": 90, "NAME": "ninety", "VALUE": 90}, "timestamp": 17000},
+        {"topic": "left_topic", "value": {"ID": 0, "NAME": "bar", "VALUE": 99}, "timestamp": 30000}
+      ],
+      "outputs": [
+        {"topic": "OUTPUT", "key": 0, "value": {"NAME": "zero", "VALUE": 0, "F1": "blah", "F2": 50}, "timestamp": 10000},
+        {"topic": "OUTPUT", "key": 0, "value": {"NAME": "foo", "VALUE": 100, "F1": "blah", "F2": 50}, "timestamp": 13000},
+        {"topic": "OUTPUT", "key": 0, "value": {"NAME": "foo", "VALUE": 100, "F1": "a", "F2": 10}, "timestamp": 15000},
+        {"topic": "OUTPUT", "key": 100, "value": {"NAME": "", "VALUE": 0, "F1": "newblah", "F2": 150}, "timestamp": 16000}
```
mjsax, line 841: Same here (is this some protobuf thing?)

ksqldb-functional-tests/src/test/resources/query-validation-tests/joins.json
```diff
@@ -1070,6 +1268,69 @@
...
+        {"topic": "left_topic", "key": 0, "value": {"NAME": "foo", "VALUE": 100}, "timestamp": 13000},
+        {"topic": "right_topic", "key": 0, "value": {"F1": "a", "F2": 10}, "timestamp": 15000},
+        {"topic": "left_topic", "key": 0, "value": {"NAME": "bar", "VALUE": 99}, "timestamp": 16000},
+        {"topic": "left_topic", "key": 90, "value": {"NAME": "ninety", "VALUE": 90}, "timestamp": 17000}
+      ],
+      "outputs": [
+        {"topic": "OUTPUT", "key": 0, "value": {"NAME": "zero", "VALUE": 0, "F1": "blah", "F2": 50}, "timestamp": 10000},
+        {"topic": "OUTPUT", "key": 0, "value": {"NAME": "foo", "VALUE": 100, "F1": "blah", "F2": 50}, "timestamp": 13000},
+        {"topic": "OUTPUT", "key": 0, "value": {"NAME": "foo", "VALUE": 100, "F1": "a", "F2": 10}, "timestamp": 15000},
+        {"topic": "OUTPUT", "key": 0, "value": {"NAME": "bar", "VALUE": 99, "F1": "a", "F2": 10}, "timestamp": 16000}
```
mjsax, line 1292: The result does not contain any "right" result. Might be good to add one -- similar below?

ksqldb-parser/src/test/java/io/confluent/ksql/parser/SqlFormatterTest.java
```diff
@@ -525,6 +525,34 @@ public void shouldFormatLeftJoinWithWithin() {
...
+        Optional.empty(),
+        rightAlias,
+        JoinedSource.Type.RIGHT,
+        criteria,
+        Optional.of(new WithinExpression(10, TimeUnit.SECONDS)))));
+
+    final String expected = "`left` L\nRIGHT OUTER JOIN `right` R WITHIN 10 SECONDS ON "
+        + "(('left.col0' = 'right.col0'))";
+    assertEquals(expected, SqlFormatter.formatSql(join));
+  }
```
mjsax, line 554: Side question: do we have any test for `GRACE` clause? We actually want to make `GRACE` mandatory. \cc @spena

mjsax (commented):
Should we also extend the QQT stream-stream join tests that use `GRACE`?

We are also missing tests for `table RIGHT JOIN stream` ?

Adding test for `SELECT *` might also be good -- as we flip the order it would be good to understand what default order the result schema is, and if it's ok as-is or if we need to write more code to actually get a user intuitive schema.

ksqldb-engine/src/main/java/io/confluent/ksql/planner/plan/JoinNode.java
```diff
@@ -545,6 +556,16 @@ private StreamToTableJoiner(
                 contextStacker
             );
           }
+        case RIGHT:
+          if (joinKey.isForeignKey()) {
+            throw new KsqlException("RIGHT OUTER JOIN on a foreign key is not supported");
```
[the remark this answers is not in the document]
Gerrrr: I could not find a way to implement FK right join given the existing primitives. As the scope of the issue is to provide syntactical sugar in KSQL rather than to extend KS joins, I decided to throw an exception here.

The issue with supporting FK right join is that KS implementation of FK join is not really symmetric. Let's look at the foreign key left join:

https://github.com/confluentinc/ksql/blob/a2efcc53f05de45daf0724ff7c4e1467f2548224/ksqldb-streams/src/main/java/io/confluent/ksql/execution/streams/ForeignKeyTableTableJoinBuilder.java#L110-L114

Several observations:
* The types are not symmetric. Table-Table join has a type `TableTableJoin<K>` while Table-Table foreign key join has a type `ForeignKeyTableTableJoin<KLeftT, KRightT>`.
* KS extracts the FK from the RHS of the expression. It is not possible to extract FK from the primary table.
* On the KS side, the join behavior is controlled by `SubscriptionWrapper#Instruction`:

```java
    public enum Instruction {
        //Send nothing. Do not propagate.
        DELETE_KEY_NO_PROPAGATE((byte) 0x00),

        //Send (k, null)
        DELETE_KEY_AND_PROPAGATE((byte) 0x01),

        //(changing foreign key, but FK+Val may not exist)
        //Send (k, fk-val) OR
        //Send (k, null) if fk-val does not exist
        PROPAGATE_NULL_IF_NO_FK_VAL_AVAILABLE((byte) 0x02),

        //(first time ever sending key)
        //Send (k, fk-val) only if fk-val exists.
        PROPAGATE_ONLY_IF_FK_VAL_AVAILABLE((byte) 0x03);
```

Given that, I do not see a way to implement the right FK join with what we have at hand. I can work on a supporting right FK join, but it seems to dramatically extend the scope of the issue. Thinking about a proper fix, in my opinion, it should look as follows:

* Introduce a new instruction that sends `(k, fk-val)` or `(null, fk-val)`
* Extend `SubscriptionJoinForeignProcessorSupplier` to support the new instruction
* Introduce new KTable method(s) for right FK join. While at it we can also introduce a "usual" right join. This step will require a KIP.

ksqldb-functional-tests/src/test/resources/query-validation-tests/joins.json
Gerrrr: It appears to be a protobuf thing - see other tests that existed before this PR:

https://github.com/confluentinc/ksql/blob/8253a9050c9e9ab62ca985b014a72387f4ec3cfb/ksqldb-functional-tests/src/test/resources/query-validation-tests/joins.json#L553-L583

ksqldb-functional-tests/src/test/resources/query-validation-tests/joins.json
Gerrrr: See the response to the comment above.

ksqldb-engine/src/main/java/io/confluent/ksql/planner/plan/JoinNode.java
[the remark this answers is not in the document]
mjsax: > As the scope of the issue is to provide syntactical sugar in KSQL rather than to extend KS joins, I decided to throw an exception here.

Well, adding `dimensionTable RIGHT JOIN factTable` is still syntactic sugar as it computes the same result as `factTable LEFT JOIN dimensionTable` -- so it is not extending joins at all.

In the end, we just need to "flip" left and right table, and convert `RIGHT JOIN` to `LEFT JOIN`. The rest should fall into place automatically?

If it too cumbersome given the current code base, we can still fix it in the logical planner rewrite I started, so no big deal.

ksqldb-engine/src/main/java/io/confluent/ksql/planner/plan/JoinNode.java
```diff
@@ -545,6 +556,16 @@ private StreamToTableJoiner(
...
+          } else {
+            return leftTable.rightJoin(
```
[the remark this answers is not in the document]
Gerrrr: I added a separate `rightJoin` call in the Stream-Stream case above in 5003f8e148d16bf346b56fbb647338f32a347307.

ksqldb-engine/src/test/java/io/confluent/ksql/engine/rewrite/DataSourceExtractorTest.java
```diff
@@ -134,7 +134,7 @@ public void shouldHandleAliasedJoinDataSources() {
   @Test
   public void shouldThrowIfLeftJoinSourceDoesNotExist() {
     // Given:
-    final AstNode stmt = givenQuery("SELECT * FROM UNKNOWN JOIN TEST2"
+    final AstNode stmt = givenQuery("SELECT * FROM UNKNOWN LEFT JOIN TEST2"
```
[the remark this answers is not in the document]
Gerrrr: Reverted in 8e474f5e6685db5c9f389139e5fc89192e98083f.

ksqldb-functional-tests/src/test/resources/query-validation-tests/joins.json
```diff
@@ -648,7 +648,205 @@
         ]
       }
     },
-     {
+    {
+      "name": "stream stream right join",
+      "format": ["AVRO", "JSON"],
+      "statements": [
+        "CREATE STREAM TEST (ID BIGINT KEY, NAME varchar, VALUE bigint) WITH (kafka_topic='left_topic', value_format='{FORMAT}');",
```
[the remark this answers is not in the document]
Gerrrr: 06165a29dfa8bcf940891e07ce540d3f1319f703

ksqldb-functional-tests/src/test/resources/query-validation-tests/joins.json
Gerrrr: https://github.com/confluentinc/ksql/pull/8845/commits/efcbe222b8125997bd23c9826fad9be93b0a4be6

ksqldb-parser/src/test/java/io/confluent/ksql/parser/SqlFormatterTest.java
Gerrrr: https://github.com/confluentinc/ksql/pull/8845/commits/319bdd7a345856968b61670dbb2d6a4e877f3040

Gerrrr:
@mjsax Are there any outstanding issues with this PR or I can merge it?

mjsax (approved):
This one dropped on the floor. Sorry about this.

LGTM.

Gerrrr:
Green build - https://jenkins.confluent.io/job/Confluent%20Public%20Repo%20PR%20builder/job/ksql/view/change-requests/job/PR-8845/3/.

colinhicks:
@Gerrrr hey, did we update the documentation for this?

Gerrrr:
You are totally right @colinhicks! I Created a follow-up PR - https://github.com/confluentinc/ksql/pull/9002.

ksqldb-engine/src/main/java/io/confluent/ksql/planner/plan/JoinNode.java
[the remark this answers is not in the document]
colinhicks: Is there an issue tracking support for right joins against foreign keys?

PR #8849, opened 2022-03-04
fix: Ensures response end handler is invoked just once

AlanConfluent:
### Description 
On http2, when a client connection is closed, the response end handler is called twice, so this prevents that.  The metrics were getting double counted as a result. 

### Testing done 
Wrote unit test and tested manually.

### Reviewer checklist

### DIFF
--- a/ksqldb-rest-app/src/main/java/io/confluent/ksql/api/server/QueryStreamHandler.java
+++ b/ksqldb-rest-app/src/main/java/io/confluent/ksql/api/server/QueryStreamHandler.java
@@ -38,6 +38,7 @@ import java.time.Clock;
 import java.util.Map;
 import java.util.Objects;
 import java.util.Optional;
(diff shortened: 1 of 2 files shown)


ksqldb-rest-app/src/main/java/io/confluent/ksql/api/server/QueryStreamHandler.java
```diff
@@ -219,6 +233,10 @@ private void handleQueryPublisher(
 
       // When response is complete, publisher should be closed and query unregistered
       routingContext.response().endHandler(v -> {
+        if (endedResponse.getAndSet(true)) {
```
agavra, line 236: super nit: let's extract this into a `close(Closeable item)` method that gets and sets, closes the underlying item and reports the metrics?

agavra (approved):

PR #8857, opened 2022-03-07
fix: clean up mistaken topic names

wcarlson5:
### Description 
To make sure the repartition topics get cleaned up we now check both possible names

### Testing done 
I made sure these tests get run with the flag on always

### Reviewer checklist

### DIFF
--- a/ksqldb-engine/src/main/java/io/confluent/ksql/engine/QueryCleanupService.java
+++ b/ksqldb-engine/src/main/java/io/confluent/ksql/engine/QueryCleanupService.java
@@ -94,6 +94,7 @@ public class QueryCleanupService extends AbstractExecutionThreadService {
   public static class QueryCleanupTask implements Runnable {
     private final String appId;
     private final String queryTopicPrefix;
+    private final String altQueryTopicPrefix;
     private final Optional<String> topologyName;
     private final String pathName;
     private final boolean isTransient;
@@ -114,6 +115,11 @@ public class QueryCleanupService extends AbstractExecutionThreadService {
           .map(s -> QueryApplicationId.buildInternalTopicPrefix(
               serviceId,
               persistentQueryPrefix) + s)
+          .orElse(appId);
+      altQueryTopicPrefix = queryId
+          .map(s -> QueryApplicationId.buildInternalTopicPrefix(
+              serviceId,
+              persistentQueryPrefix.split("_")[0] + "-") + s)
           .orElse(appId);
       //generate the prefix depending on if using named topologies
       this.isTransient = isTransient;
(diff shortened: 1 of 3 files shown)


ksqldb-engine/src/test/java/io/confluent/ksql/integration/WindowingSharedRuntimeIntTest.java
```diff
@@ -0,0 +1,306 @@
...
+import static io.confluent.ksql.test.util.AssertEventually.assertThatEventually;
+import static io.confluent.ksql.test.util.ConsumerTestUtil.hasUniqueRecords;
+import static io.confluent.ksql.test.util.MapMatchers.mapHasItems;
+import static org.hamcrest.MatcherAssert.assertThat;
+import static org.hamcrest.Matchers.containsInAnyOrder;
+import static org.hamcrest.Matchers.hasSize;
+import static org.hamcrest.Matchers.is;
+
+@Category({IntegrationTest.class})
+public class WindowingSharedRuntimeIntTest {
```
ableegoldman, line 67: Can we just parametrize the other test vs adding an entire new test? Is it too complicated or something? Just wondering if that might be easier -- also keep in mind, adding new integration tests is really what ultimately impacts the test run time in a significant way, since it often has a nontrivial startup time to bring up the cluster/test harness/etc.

Not sure what the actual overhead for this specific test is, just wanted to point that out in case it's possible to avoid adding minutes to the time it takes to run all the tests (which is already wayyy too long lol)

ksqldb-engine/src/main/java/io/confluent/ksql/engine/QueryCleanupService.java
```diff
@@ -94,6 +94,7 @@ public void addCleanupTask(final QueryCleanupTask task) {
   public static class QueryCleanupTask implements Runnable {
     private final String appId;
     private final String queryTopicPrefix;
+    private final String altQueryTopicPrefix;
```
ableegoldman, line 97: nit: can you leave a quick comment here explaining why we need to check and possibly delete this "alt prefix"?

ableegoldman (approved):
Few minor things but overall LGTM, I'll leave it up to you to address those and feel free to merge afterwards

ableegoldman:
Looks like you have a build error though, FYI

ksqldb-engine/src/test/java/io/confluent/ksql/integration/WindowingSharedRuntimeIntTest.java
wcarlson5: I tired. It doesn't work with the IntegrationTestHarness. When we turn it on by default we can remove the old test classes. The new test classes are quite a bit faster tho!

PR #8860, opened 2022-03-07
chore: add feature flag for new logical planner

mjsax:
Add feature flag for new persistent query planner.

### DIFF
--- a/ksqldb-common/src/main/java/io/confluent/ksql/util/KsqlConfig.java
+++ b/ksqldb-common/src/main/java/io/confluent/ksql/util/KsqlConfig.java
@@ -576,6 +576,11 @@ public class KsqlConfig extends AbstractConfig {
           + "Default is false. If false, persistent queries will use separate "
           + " runtimes, if true, new queries may share streams instances.";
 
+  public static final String KSQL_NEW_QUERY_PLANNER_ENABLED =
+      "ksql.new.query.planner.enabled";
+  private static final Boolean KSQL_NEW_QUERY_PLANNER_ENABLED_DEFAULT = false;
+  private static final String KSQL_NEW_QUERY_PLANNER_ENABLED_DOC =
+      "Feature flag that enables the new planner for persistent queries. Default is false.";
 
   public static final String KSQL_SHARED_RUNTIMES_COUNT = "ksql.shared.runtimes.count";
   public static final Integer KSQL_SHARED_RUNTIMES_COUNT_DEFAULT = 8;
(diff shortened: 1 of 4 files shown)


ksqldb-rest-model/src/main/java/io/confluent/ksql/rest/entity/PropertiesList.java
```diff
@@ -189,7 +190,8 @@
       STATE_DIR_CONFIG,
       TASK_TIMEOUT_MS_CONFIG,
       WINDOW_SIZE_MS_CONFIG,
-      UPGRADE_FROM_CONFIG
+      UPGRADE_FROM_CONFIG,
+      KSQL_NEW_QUERY_PLANNER_ENABLED
```
mjsax, line 194: Was looking into other feature flag PRs, that also add it here, but not sure why?

guozhangwang: Is this for LD?

guozhangwang (approved):

ksqldb-common/src/main/java/io/confluent/ksql/util/KsqlConfig.java
```diff
@@ -576,6 +576,11 @@
           + "Default is false. If false, persistent queries will use separate "
           + " runtimes, if true, new queries may share streams instances.";
 
+  public static final String KSQL_NEW_QUERY_PLANNER_ENABLED =
+      "ksql.new.query.planner.enabled";
```
spena, line 580: Shouldn't we have a good config name instead of `new.query.planner`? Just wondering what if we add an new improved planner after this new planner was added :P?

ksqldb-rest-model/src/main/java/io/confluent/ksql/rest/entity/PropertiesList.java
spena: I think this is related to the `show properties` statement.This class is used in `ListPropertiesExecutor`, so all properties listed here are displayed by that statement. 

Should this flag be hidden from clients? If so, then we don't need to add it in this list.

jzaralim: These are properties that are alterable after starting ksql

jzaralim (approved):

ksqldb-common/src/main/java/io/confluent/ksql/util/KsqlConfig.java
mjsax: Happy to use a different name if you have a suggestion. But the config should be temporary only anyway.

ksqldb-rest-model/src/main/java/io/confluent/ksql/rest/entity/PropertiesList.java
mjsax: After some follow up discussion, we should not add the config here, as we don't want users to change it via `ALTER SYSTEM` -- we also added it to `ImmutableProperties` to disallow changing it via `SET` and thus should not "allow list" is here.

Will fix in a follow up PR.

PR #8862, opened 2022-03-08
feat: add stubs for new query planner

mjsax:
Add stubs for new logical and physical planners. It's still a little rough, but a start. The goal of this PR is just to support `CREATE STREAM AS SELECT * FROM inputStream;`

Btw: this might only be an intermediate step with regard to the physical plan -- we translated the new LogicalPlan directly into `ExecutionSteps` right now, but we might want to add a PhysicalPlan in between -- IMHO, `ExecutionSteps` are not a "physical plan" but already one layer deeper (more like a runtime/execution plan). Still thinking about it...

### DIFF
--- a/ksqldb-engine/src/main/java/io/confluent/ksql/engine/EngineExecutor.java
+++ b/ksqldb-engine/src/main/java/io/confluent/ksql/engine/EngineExecutor.java
@@ -86,14 +86,19 @@ import io.confluent.ksql.planner.plan.KsqlBareOutputNode;
 import io.confluent.ksql.planner.plan.KsqlStructuredDataOutputNode;
 import io.confluent.ksql.planner.plan.OutputNode;
 import io.confluent.ksql.planner.plan.PlanNode;
+import io.confluent.ksql.planner.plan.PlanNodeId;
+import io.confluent.ksql.planner.plan.VerifiableNode;
 import io.confluent.ksql.query.PullQueryQueue;
 import io.confluent.ksql.query.QueryId;
 import io.confluent.ksql.query.QueryRegistry;
 import io.confluent.ksql.query.TransientQueryQueue;
 import io.confluent.ksql.schema.ksql.LogicalSchema;
+import io.confluent.ksql.schema.ksql.LogicalSchema.Builder;
 import io.confluent.ksql.schema.utils.FormatOptions;
+import io.confluent.ksql.serde.FormatInfo;
 import io.confluent.ksql.serde.KeyFormat;
 import io.confluent.ksql.serde.RefinementInfo;
+import io.confluent.ksql.serde.SerdeFeatures;
 import io.confluent.ksql.serde.ValueFormat;
 import io.confluent.ksql.services.ServiceContext;
 import io.confluent.ksql.statement.ConfiguredStatement;
@@ -453,15 +458,13 @@ final class EngineExecutor {
     }
   }
 
-
-  @SuppressWarnings("OptionalGetWithoutIsPresent") // Known to be non-empty
   TransientQueryMetadata executeTransientQuery(
       final ConfiguredStatement<Query> statement,
       final boolean excludeTombstones
   ) {
     final ExecutorPlans plans = planQuery(statement, statement.getStatement(),
         Optional.empty(), Optional.empty(), engineContext.getMetaStore());
-    final KsqlBareOutputNode outputNode = (KsqlBareOutputNode) plans.logicalPlan.getNode().get();
+    final KsqlBareOutputNode outputNode = (KsqlBareOutputNode) plans.outputNode;
     engineContext.createQueryValidator().validateQuery(
         config,
         plans.physicalPlan,
@@ -486,7 +489,6 @@ final class EngineExecutor {
     );
   }
 
-  @SuppressWarnings("OptionalGetWithoutIsPresent") // Known to be non-empty
   TransientQueryMetadata executeStreamPullQuery(
       final ConfiguredStatement<Query> statement,
       final boolean excludeTombstones,
@@ -494,7 +496,7 @@ final class EngineExecutor {
   ) {
     final ExecutorPlans plans = planQuery(statement, statement.getStatement(),
         Optional.empty(), Optional.empty(), engineContext.getMetaStore());
-    final KsqlBareOutputNode outputNode = (KsqlBareOutputNode) plans.logicalPlan.getNode().get();
+    final KsqlBareOutputNode outputNode = (KsqlBareOutputNode) plans.outputNode;
     engineContext.createQueryValidator().validateQuery(
         config,
         plans.physicalPlan,
@@ -592,7 +594,7 @@ final class EngineExecutor {
     );
 
     final KsqlBareOutputNode outputNode =
-        (KsqlBareOutputNode) plans.logicalPlan.getNode().get();
+        (KsqlBareOutputNode) plans.outputNode;
 
     final QueryPlan queryPlan = new QueryPlan(
         getSourceNames(outputNode),
@@ -621,8 +623,6 @@ final class EngineExecutor {
         .getBoolean(KsqlConfig.KSQL_SOURCE_TABLE_MATERIALIZATION_ENABLED);
   }
 
-  // Known to be non-empty
-  @SuppressWarnings("OptionalGetWithoutIsPresent")
   KsqlPlan plan(final ConfiguredStatement<?> statement) {
     try {
       throwOnNonExecutableStatement(statement);
@@ -664,7 +664,7 @@ final class EngineExecutor {
       );
 
       final KsqlStructuredDataOutputNode outputNode =
-          (KsqlStructuredDataOutputNode) plans.logicalPlan.getNode().get();
+          (KsqlStructuredDataOutputNode) plans.outputNode;
 
       final Optional<DdlCommand> ddlCommand = maybeCreateSinkDdl(
           statement,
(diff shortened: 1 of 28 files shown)


ksqldb-engine/src/main/java/io/confluent/ksql/logicalplanner/LogicalPlan.java
```diff
@@ -0,0 +1,39 @@
...
+ */
+
+package io.confluent.ksql.logicalplanner;
+
+import com.google.common.collect.ImmutableSet;
+import io.confluent.ksql.logicalplanner.nodes.Node;
+import io.confluent.ksql.name.SourceName;
+import java.util.Objects;
+
+public class LogicalPlan {
```
jnh5y, line 23: Just so I know the strategy, this class will be extended as the new query planner adds handling for limits, joins, etc, right?

ksqldb-engine/src/main/java/io/confluent/ksql/logicalplanner/LogicalPlanner.java
```diff
@@ -0,0 +1,67 @@
...
+
+    final SourceName sourceName = inputStreamOrTable.getAlias();
+    final DataSource sourceStreamOrTable = metaStore.getSource(sourceName);
+
+    final SourceNode<?> sourceNode;
+    if (sourceStreamOrTable.getDataSourceType() == DataSourceType.KSTREAM) {
+      sourceNode = new StreamSourceNode(sourceName, sourceStreamOrTable.getSchema());
+    } else if (sourceStreamOrTable.getDataSourceType() == DataSourceType.KTABLE) {
+      throw new UnsupportedOperationException("Input TABLES are not supported by the new planner");
+      //sourceNode = new TableSourceNode(sourceName, sourceStreamOrTable.getSchema());
```
jnh5y, line 55: Are you leaving comments like this to show future strategies?

ksqldb-engine/src/main/java/io/confluent/ksql/logicalplanner/nodes/Node.java
```diff
@@ -0,0 +1,27 @@
...
+package io.confluent.ksql.logicalplanner.nodes;
+
+import com.google.common.collect.ImmutableList;
+import io.confluent.ksql.schema.ksql.LogicalColumn;
+
+public interface Node<AcceptsT extends Node<?>> {
+
+  ImmutableList<LogicalColumn> getOutputSchema();
+
+  <ReturnsT> ReturnsT accept(NodeVisiter<AcceptsT, ReturnsT> visitor);
```
jnh5y, line 25: Neat.  I haven't seen the naming types this clearly before.  I kinda like it.:)

ksqldb-engine/src/main/java/io/confluent/ksql/logicalplanner/nodes/SourceNode.java
```diff
@@ -0,0 +1,56 @@
...
+import com.google.common.collect.ImmutableList;
+import io.confluent.ksql.name.SourceName;
+import io.confluent.ksql.schema.ksql.LogicalColumn;
+import io.confluent.ksql.schema.ksql.LogicalSchema;
+import java.util.Objects;
+
+public abstract class SourceNode<SourceT extends SourceNode<?>> implements Node<SourceT> {
+  final SourceName sourceName;
+  final LogicalSchema simpleSchema; // schema without system columns
+  final ImmutableList<LogicalColumn> outputSchema; // to-do: should we include system columns?
```
jnh5y, line 27: What are some examples of system columns?  Would those include headers and/or timestamps?  If so, I imagine the answer is "yes".

ksqldb-engine/src/main/java/io/confluent/ksql/physical/PhysicalPlan.java
```diff
@@ -25,7 +25,7 @@
   private final QueryId queryId;
   private final ExecutionStep<?> physicalPlan;
 
-  PhysicalPlan(
+  public PhysicalPlan(
```
jnh5y, line 28: Oooo, this is probably the biggest decision being made here.  

Are there any options besides making this public?

ksqldb-engine/src/main/java/io/confluent/ksql/physicalplanner/PhysicalPlanner.java
```diff
@@ -0,0 +1,63 @@
...
+  public static PhysicalPlan buildPhysicalPlan(
+      final MetaStore metaStore,
+      final LogicalPlan logicalPlan
+  ) {
+    final LogicalToPhysicalPlanTranslator translator =
+        new LogicalToPhysicalPlanTranslator(metaStore);
+
+    final ExecutionStep<?> root = translator.process(logicalPlan.getRoot());
+
+    final Formats formats = Formats.of(
```
jnh5y, line 46: What are these formats being used for?

ksqldb-engine/src/main/java/io/confluent/ksql/physicalplanner/PhysicalPlanner.java
```diff
@@ -0,0 +1,63 @@
...
+import io.confluent.ksql.logicalplanner.LogicalPlan;
+import io.confluent.ksql.metastore.MetaStore;
+import io.confluent.ksql.physical.PhysicalPlan;
+import io.confluent.ksql.query.QueryId;
+import io.confluent.ksql.serde.FormatInfo;
+import io.confluent.ksql.serde.SerdeFeatures;
+import java.util.Collections;
+import java.util.Optional;
+
+public final class PhysicalPlanner {
```
jnh5y, line 33: Maybe some Java Docs for the new classes would help explain strategy.  Is it too early to add those?

ksqldb-rest-model/src/main/java/io/confluent/ksql/rest/entity/PropertiesList.java
```diff
@@ -119,6 +118,11 @@
       FAIL_ON_DESERIALIZATION_ERROR_CONFIG
   );
 
+  /**
+   * List os properties that can be changes via `ALTER SYSTEM` command.
+   * We use this "allow list" for security reasons.
+   * (Independent of LD.)
```
jnh5y, line 124: What's `LD` mean?

ksqldb-functional-tests/src/test/resources/query-validation-tests/select.json
```diff
@@ -3,6 +3,27 @@
     "Tests covering general SELECT clause, a.k.a projection, handling"
   ],
   "tests": [
+    {
+      "name": "new query planner test",
```
jnh5y, line 7: Two things:
1.  This test does not validate that the new query planner is being used, right?  (Sanity check ask; I don't think we should validate that the new query planner being used in functional tests.  Doing that in a unit test may be amusing, but I am not worried about it.)
2.  Does a query plan need to be generated for this?   (Assuming we check it in.  I'd be fine with not having it.)

ksqldb-engine/src/main/java/io/confluent/ksql/engine/EngineExecutor.java
```diff
@@ -762,7 +851,59 @@ private ExecutorPlans planQuery(
           queryId,
           oldPlanInfo
       );
-      return new ExecutorPlans(logicalPlan, physicalPlan);
+
+      return new ExecutorPlans(logicalPlan.getNode().get(), physicalPlan);
+    }
+  }
+
+  private static class StubbedOutputNode extends KsqlStructuredDataOutputNode {
```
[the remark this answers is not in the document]
mjsax: Still might need further refactoring. The problem is, that we need to pass some "logical plan" (old format) into `ExecutorPlans` (so we just use a mock for now, as we don't really generate the old logical plan that we can pass in).

Not sure to what extend we can modify `ExecutorPlans` later to avoid this -- it's really just to make the code work for now; not a permanent solution.

ksqldb-engine/src/main/java/io/confluent/ksql/logicalplanner/LogicalPlan.java
mjsax: Right. It's the "container" class of the new logical plan. The plan itself with be a graph of `Nodes` so we will add `JoinNode`, `LimitNode` etc

ksqldb-engine/src/main/java/io/confluent/ksql/logicalplanner/LogicalPlanner.java
mjsax: Yes. Did not try to extend the scope to add input `TABLE` support yet. It's the intended code when we add `TABLES`.

ksqldb-engine/src/main/java/io/confluent/ksql/logicalplanner/nodes/SourceNode.java
mjsax: Yes, it's `ROWTIME`, `ROWOFFSET`, `ROWPARTITION` etc.

It's not clear for me atm, how the old code exactly treat them -- they are actually added at some point in the old code -- still need to figure out / understand the details.

My gut feeling is also that the answer might be "yes".

ksqldb-engine/src/main/java/io/confluent/ksql/physical/PhysicalPlan.java
mjsax: What is the concern?

Btw: I might actually want to rename this, as the `PhysicalPlan` is actually an "execution plan" (ie, one level below the physical plan), and I might want to add an actual (new) "physical plan" layer. It's a little be TBD to figure out the details.

ksqldb-engine/src/main/java/io/confluent/ksql/physicalplanner/PhysicalPlanner.java
mjsax: This is hard coded atm and needs to be fixed in follow up PR.

It's just to make the new QTT work that is using those formats. Need to add logic to actually get it from the query. This might require to add a new "physical plan" layer (did not want to extend the scope of this PR further to avoid a 5000 LOC PR).

ksqldb-functional-tests/src/test/resources/query-validation-tests/select.json
mjsax: 1. Right. I agree that we don't need (and maybe not even can?) verify it. But I am not worried about it either.
2. Not sure if I understand the question? Can you elaborate?

jnh5y: Ah, I feel like the `PlannedTestGeneratorTest` fails if I don't run this function: https://github.com/confluentinc/ksql/blob/master/ksqldb-functional-tests/src/test/java/io/confluent/ksql/test/planned/PlannedTestGeneratorTest.java#L28-L40 to generate the historical plans for a given test.

(I think) Basically, if you keep this test, then you may need to generate a plan for it (otherwise CI will fail on that).

jnh5y (approved):
NB: The CI link is busted, so check https://jenkins.confluent.io/job/confluentinc-pr/job/ksql/job/PR-8862/ before merging.

It is presently failing with:

```00:14:53.352  [ERROR] /home/jenkins/workspace/confluentinc-pr_ksql_PR-8862/ksqldb-engine/src/main/java/io/confluent/ksql/engine/EngineExecutor.java:859:3: Class StubbedOutputNode should be declared as final. [FinalClass]
00:14:53.352  [ERROR] /home/jenkins/workspace/confluentinc-pr_ksql_PR-8862/ksqldb-engine/src/main/java/io/confluent/ksql/engine/EngineExecutor.java:886: Line is longer than 100 characters (found 105). [LineLength]
00:14:53.352  [ERROR] /home/jenkins/workspace/confluentinc-pr_ksql_PR-8862/ksqldb-engine/src/main/java/io/confluent/ksql/engine/EngineExecutor.java:888:5: Redundant 'public' modifier. [RedundantModifier]
00:14:53.352  [ERROR] /home/jenkins/workspace/confluentinc-pr_ksql_PR-8862/ksqldb-engine/src/main/java/io/confluent/ksql/engine/EngineExecutor.java:899:36: Parameter sinkName should be final. [FinalParameters]```

ksqldb-functional-tests/src/test/resources/query-validation-tests/select.json
mjsax: Ah. Yes. I did not intent to keep this test -- it's just for "showcasing" on the PR.

ksqldb-engine/src/main/java/io/confluent/ksql/physical/PhysicalPlan.java
jnh5y: I was just trying to figure out if we want this as part of the public API or we consider it internal.

ksqldb-rest-model/src/main/java/io/confluent/ksql/rest/entity/PropertiesList.java
suhas-satish: LaunchDarkly (the tool that manages our feature flags which can be turned on/off at runtime in ccloud)

PR #8868, opened 2022-03-08
feat: add support for simple column name projection to new query planner

mjsax:
Extent projections from only supporting `*` to support simple input column names.

### DIFF
--- a/ksqldb-engine/src/main/java/io/confluent/ksql/engine/EngineExecutor.java
+++ b/ksqldb-engine/src/main/java/io/confluent/ksql/engine/EngineExecutor.java
@@ -46,7 +46,6 @@ import io.confluent.ksql.metastore.model.KsqlTable;
 import io.confluent.ksql.name.SourceName;
 import io.confluent.ksql.parser.OutputRefinement;
 import io.confluent.ksql.parser.tree.AliasedRelation;
-import io.confluent.ksql.parser.tree.AllColumns;
 import io.confluent.ksql.parser.tree.CreateAsSelect;
 import io.confluent.ksql.parser.tree.CreateStream;
 import io.confluent.ksql.parser.tree.CreateStreamAsSelect;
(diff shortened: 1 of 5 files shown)


ksqldb-engine/src/main/java/io/confluent/ksql/engine/EngineExecutor.java
```diff
@@ -716,41 +716,36 @@ private void throwIfUnsupported(final Query query) {
...
     final Relation fromClause = query.getFrom();
     if (fromClause instanceof Join) {
       throw new UnsupportedOperationException("New query planner does not support joins."
-          + "Set " + KsqlConfig.KSQL_NEW_QUERY_PLANNER_ENABLED + "=false.");
+          + " Set " + KsqlConfig.KSQL_NEW_QUERY_PLANNER_ENABLED + "=false.");
     }
     if (fromClause instanceof JoinedSource) {
       throw new IllegalStateException(); // top level node should always be Join
     }
-    if (query.getSelect().getSelectItems().size() > 2
```
jnh5y: Since this PR is for just the "simple column", does it make sense to continue checking any restrictions here?

ksqldb-engine/src/main/java/io/confluent/ksql/logicalplanner/nodes/SelectNode.java
```diff
@@ -31,15 +43,41 @@ public SelectNode(
...
+                throw new UnknownColumnException("SELECT", (ColumnReferenceExp) e);
+
+              }
+              if (matchingInputColumns.size() > 1) {
+                throw new KsqlException("Ambiguous column " + columnName + " in SELECT clause");
+              }
+
+              outputSchemaBuilder.add(Iterators.getOnlyElement(matchingInputColumns.iterator()));
+            } else {
+              throw new UnsupportedOperationException(
```
jnh5y, line 72: The check here makes sense.  (Ignore the above comment.)

jnh5y (approved):
LGTM.

ksqldb-engine/src/main/java/io/confluent/ksql/logicalplanner/nodes/SelectNode.java
```diff
@@ -31,15 +43,41 @@ public SelectNode(
...
+              final ColumnName columnName = ((ColumnReferenceExp) e).getColumnName();
+              final ImmutableList<LogicalColumn> matchingInputColumns = inputSchema.stream()
+                  .filter(logicalColumn -> logicalColumn.name().equals(columnName))
+                  .collect(ImmutableList.toImmutableList());
+              if (matchingInputColumns.isEmpty()) {
+                throw new UnknownColumnException("SELECT", (ColumnReferenceExp) e);
+
+              }
+              if (matchingInputColumns.size() > 1) {
+                throw new KsqlException("Ambiguous column " + columnName + " in SELECT clause");
```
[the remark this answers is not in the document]
mjsax: It depends. For a "transient query" it does not really matter (but atm the planner is only for persistent queries). For a `CREATE AS` is does matter though because a schema cannot contain overlapping column names (and the `SELECT` defines the schema of the result STREAM/TABLE): it would be required to rename one column using `AS newName` (will as support for `AS` later -- I would assume there is some test already but will double check when working on `AS`).

To be fair, atm ksqlDB has some "auto renaming" feature, but I think it's actually not a good thing to auto-rename and we should rather throw an error and force users to expliclity pick column names IMHO.

Thoughts?

ksqldb-engine/src/main/java/io/confluent/ksql/engine/EngineExecutor.java
mjsax: I think we cannot do a good check here any longer, without writing a log of code. Better to push checks down into the "select node" class.

ksqldb-engine/src/main/java/io/confluent/ksql/logicalplanner/nodes/SelectNode.java
```diff
@@ -31,15 +39,41 @@ public SelectNode(
...
+    final Builder<LogicalColumn> outputSchemaBuilder = ImmutableList.builder();
 
-    outputSchema = selectClause.getSelectItems().stream().flatMap(
-        s -> input.getOutputSchema().stream()
-    ).collect(ImmutableList.toImmutableList());
+    selectClause.getSelectItems().forEach(
+        s -> {
+          if (s instanceof AllColumns) {
+            outputSchemaBuilder.addAll(inputSchema);
+          } else {
```
lihaosky, line 49: Is it legit to have both `*` and `SingleColumn`?

ksqldb-engine/src/main/java/io/confluent/ksql/logicalplanner/nodes/SelectNode.java
```diff
@@ -31,15 +39,41 @@ public SelectNode(
...
+                  .collect(ImmutableList.toImmutableList());
+              if (matchingInputColumns.isEmpty()) {
+                throw new UnknownColumnException("SELECT", (ColumnReferenceExp) e);
+
+              }
+              if (matchingInputColumns.size() > 1) {
+                throw new KsqlException("Ambiguous column " + columnName + " in SELECT clause");
+              }
+
+              outputSchemaBuilder.add(Iterators.getOnlyElement(matchingInputColumns.iterator()));
```
lihaosky, line 66: This is called `outputSchemaBuilder` but seems the `LogicalName` is from input. Why not use output alias as output logicalname if alias exists?

mjsax: Will add alias support only later.

lihaosky (approved):
LGTM!

PR #8869, opened 2022-03-09
fix: do not include schema id in session config

lihaosky:
### Description 
Remove schema id from SessionConfig since SessionConfig keys are validated later

### Testing done 
Unit 

### Reviewer checklist


wcarlson5 (approved):
LGTM, I pulled and tested with the new runtimes. 

Though I don't know much about this feature. If you feel confident that this is a fine change then you are good, but you might also want to get someone who is more familiar with the feature to give it a look

Issue #7697, opened 2021-06-18
Add support for compound types to COLLECT_SET and COLLECT_LIST


derekjn:
`COLLECT_SET` and `COLLECT_LIST` currently only support scalar types. We should expand the supported types to allow users to pass compound types (`Array`, `Map` and `Struct`) to these aggregates. Given that compound types may now be used as row keys, we should have the associated equality operators to make this possible.

agavra:
depends on #6638

jnh5y:
https://github.com/confluentinc/ksql/pull/8877 should handle this.

Issue #6638, opened 2020-11-18
Add schemaProvider functionality to UDAFs


big-andy-coates:
Both UDFs and UDTFs support calling a method to obtain the return type of a function by setting a `schemaProvider`.  This can be very useful when working with generic or Struct parameters.  However, UDAFs have no such capability and this is stopping people writing UDAFs that work with more generic / structured typed.

UDAFs support input, aggregate and output schemas. Likely, we'd need a schema provider method for the last two. Maybe the first?

This work should probably be done at the same time as [extending UDAFs to support multiple parameters](https://github.com/confluentinc/ksql/issues/5747)

big-andy-coates:
Needed to fixc https://github.com/confluentinc/ksql/issues/5437

jnh5y:
https://github.com/confluentinc/ksql/pull/8871 and https://github.com/confluentinc/ksql/pull/8896 should cover this issue.

PR #8871, opened 2022-03-09
feat: Extend Udaf interface to allow for polymorphic UDAFs.

jnh5y:
Various UDAFs can operate on input of any type.  This presents a challenge since ksqlDB needs to compute
the aggregate and return types of a function at runtime.

Here the `Udaf` interface is extended in order to allow for such functions to be written.

This work addresses https://github.com/confluentinc/ksql/issues/6638.

As motivation for the change, see https://github.com/confluentinc/ksql/issues/7697 and https://github.com/confluentinc/ksql/issues/5437.

### Testing done 
This change has been validated with other work-in-progress branches.  This PR only has the interface changes to focus discussion.

### Reviewer checklist
- Depending on discussion, docs may be updated with a separate PR.

### DIFF
--- a/ksqldb-engine/src/main/java/io/confluent/ksql/function/UdafFactoryInvoker.java
+++ b/ksqldb-engine/src/main/java/io/confluent/ksql/function/UdafFactoryInvoker.java
@@ -94,20 +94,10 @@ class UdafFactoryInvoker implements FunctionSignature {
         ((Configurable) udaf).configure(initArgs.config());
       }
 
-      final SqlType aggregateSqlType;
-      final SqlType returnSqlType;
-
-      if (udaf.getAggregateSqlType().isPresent()) {
-        aggregateSqlType = (SqlType) udaf.getAggregateSqlType().get();
-      } else {
-        aggregateSqlType = SchemaConverters.functionToSqlConverter().toSqlType(aggregateArgType);
-      }
-
-      if (udaf.getReturnSqlType().isPresent()) {
-        returnSqlType = (SqlType) udaf.getReturnSqlType().get();
-      } else {
-        returnSqlType = SchemaConverters.functionToSqlConverter().toSqlType(aggregateReturnType);
-      }
+      final SqlType aggregateSqlType = (SqlType) udaf.getAggregateSqlType()
+          .orElse(SchemaConverters.functionToSqlConverter().toSqlType(aggregateArgType));
+      final SqlType returnSqlType = (SqlType) udaf.getReturnSqlType()
+          .orElse(SchemaConverters.functionToSqlConverter().toSqlType(aggregateReturnType));
 
       final KsqlAggregateFunction function;
       if (TableUdaf.class.isAssignableFrom(method.getReturnType())) {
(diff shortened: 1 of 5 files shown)


ksqldb-udf/src/main/java/io/confluent/ksql/function/udaf/Udaf.java
```diff
@@ -63,4 +67,33 @@
...
+   */
+  default void initializeTypeArguments(List<SqlArgument> argTypeList) { }
+
+  /**
+   * Most UDAFs advertise their aggregate type statically via the Java type signature or
+   * annotations.
+   * For polymorphic UDAFs, implement this method to return the aggregate SQL Type.
+   * @return The aggregate SQL Type of type `A` returned by `aggregate` and `merge`.
+   */
+  default SqlType getAggregateSqlType() {
```
spena, line 86: Make this and the getReturnSqlType to return Optional<SqlType>. The ksql code is trying to use Optional more often to prevent NPE later these methods are called in the future and null is not handled.

ksqldb-engine/src/test/java/io/confluent/ksql/function/UdfLoaderTest.java
```diff
@@ -826,7 +826,7 @@ public void shouldInvokeUdafWithMethodWithNoArgs() throws Exception {
         "",
         "",
         "");
-    assertThat(creator.createFunction(AggregateFunctionInitArguments.EMPTY_ARGS),
+    assertThat(creator.createFunction(AggregateFunctionInitArguments.EMPTY_ARGS, Collections.EMPTY_LIST),
```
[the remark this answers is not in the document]
jnh5y: Since this is a widening of the API, none of the existing tests depend on what is being passed into that function.

The function here is not public, so I'm unsure how much I'm worried with testing it. 

Also, I'm not sure what test would show things really well. 

That said, I did put together a rough example test which shows that the parameters are used.

ksqldb-udf/src/main/java/io/confluent/ksql/function/udaf/Udaf.java
jnh5y: Done and done!

ksqldb-udf/src/main/java/io/confluent/ksql/function/udaf/Udaf.java
```diff
@@ -63,4 +68,33 @@
    * @return new value of current record
    */
   O map(A agg);
+
+  /**
+   * Some UDAFs can operate on any type.  In that case, the UDAF needs to be aware of the column
+   * type being used.  This method is called once when the UDAF is being created.
+   * Implementors may need to hold on to argument types or compute some other state to be re-used
+   * in methods like `aggregate` or `merge`.
```
spena, line 76:
```suggestion
   * in methods like {@code aggregate} or {@code merge}.
```

ksqldb-engine/src/main/java/io/confluent/ksql/function/UdafFactoryInvoker.java
```diff
@@ -78,24 +81,42 @@
...
       }
 
+      final SqlType aggregateSqlType;
+      final SqlType returnSqlType;
+
+      if (udaf.getAggregateSqlType().isPresent()) {
+        aggregateSqlType = (SqlType) udaf.getAggregateSqlType().get();
+      } else {
+        aggregateSqlType = SchemaConverters.functionToSqlConverter().toSqlType(aggregateArgType);
+      }
```
spena, line 104: (optional) A trick using optional values:

```suggestion
aggregateSqlType = udaf.getAggregateSqlType()
   .orElse(SchemaConverters.functionToSqlConverter().toSqlType(aggregateArgType));    
```

spena (approved):
LGTM. I left a few minor comments.

PR #8872, opened 2022-03-09
fix: update rate limiting test so it's not flaky

lct45:
### Description 
We recently saw a failure for this test where an exception wasn't thrown, indicating that we didn't hit the rate limit. I wasn't able to reproduce locally so it must be something specific with the CI. I added more chances to hit the exception so hopefully, this won't be flaky anymore.

### Reviewer checklist


jnh5y (approved):
LGTM!

PR #8876, opened 2022-03-10
fix: resolve schema registry issue for pull query 

### LINKED ISSUE #7174: Schema Registry 409 when issuing pull query against table with multi-column keys
**Describe the bug**

When issuing the follwoing pull query against a table with a multi-column key
```sql
SELECT * FROM  REEFERS_AND_VESSELS_CLOSE_FOR_GT_2HR 
WHERE fishing_vessel_mmsi = '273219900' and reefer_mmsi='273219900';
```

ksqlDB returns the following error:

```
io.confluent.ksql.util.KsqlStatementException: Error serializing message to topic: _confluent-ksql-pksqlc-k8m02query_CTAS_REEFERS_AND_VESSELS_CLOSE_FOR_GT_2HR_23-Aggregate-GroupBy-repartition. Failed to serialize Avro data from topic _confluent-ksql-pksqlc-k8m02query_CTAS_REEFERS_AND_VESSELS_CLOSE_FOR_GT_2HR_23-Aggregate-GroupBy-repartition
```

When digging into the stack trace, it looks like we're trying to register a different schema during the pull query step than we do during the persistent query processing:

```
Caused by: org.apache.kafka.common.errors.SerializationException: Error registering Avro schema for subject _confluent-ksql-pksqlc-k8m02query_CTAS_REEFERS_AND_VESSELS_CLOSE_FOR_GT_2HR_23-Aggregate-GroupBy-repartition-key and schema: {"type":"record","name":"ReefersAndVesselsCloseForGt2hrKey","namespace":"io.confluent.ksql.avro_schemas","fields":[{"name":"FISHING_VESSEL_MMSI","type":["null","string"],"default":null},{"name":"REEFER_MMSI","type":["null","string"],"default":null}]}
Caused by: io.confluent.kafka.schemaregistry.client.rest.exceptions.RestClientException: Schema being registered is incompatible with an earlier schema for subject "_confluent-ksql-pksqlc-k8m02query_CTAS_REEFERS_AND_VESSELS_CLOSE_FOR_GT_2HR_23-Aggregate-GroupBy-repartition-key"; error code: 409; error code: 409
	at io.confluent.kafka.schemaregistry.client.rest.RestService.sendHttpRequest(RestService.java:295)
	at io.confluent.kafka.schemaregistry.client.rest.RestService.httpRequest(RestService.java:355)
	at io.confluent.kafka.schemaregistry.client.rest.RestService.registerSchema(RestService.java:498)
	at io.confluent.kafka.schemaregistry.client.rest.RestService.registerSchema(RestService.java:489)
	at io.confluent.kafka.schemaregistry.client.rest.RestService.registerSchema(RestService.java:462)
	at io.confluent.kafka.schemaregistry.client.CachedSchemaRegistryClient.registerAndGetId(CachedSchemaRegistryClient.java:214)
	at io.confluent.kafka.schemaregistry.client.CachedSchemaRegistryClient.register(CachedSchemaRegistryClient.java:276)
	at io.confluent.kafka.schemaregistry.client.CachedSchemaRegistryClient.register(CachedSchemaRegistryClient.java:252)
	at io.confluent.kafka.serializers.AbstractKafkaAvroSerializer.serializeImpl(AbstractKafkaAvroSerializer.java:77)
	at io.confluent.connect.avro.AvroConverter$Serializer.serialize(AvroConverter.java:143)
	at io.confluent.connect.avro.AvroConverter.fromConnectData(AvroConverter.java:84)
	at io.confluent.ksql.serde.connect.KsqlConnectSerializer.serialize(KsqlConnectSerializer.java:53)
	at io.confluent.ksql.serde.tls.ThreadLocalSerializer.serialize(ThreadLocalSerializer.java:37)
	at io.confluent.ksql.serde.connect.ConnectFormat$ListToStructSerializer.serialize(ConnectFormat.java:173)
	at io.confluent.ksql.serde.connect.ConnectFormat$ListToStructSerializer.serialize(ConnectFormat.java:136)
	at io.confluent.ksql.serde.GenericSerializer.serialize(GenericSerializer.java:62)
	at io.confluent.ksql.logging.processing.LoggingSerializer.serialize(LoggingSerializer.java:47)
	at org.apache.kafka.streams.processor.internals.DefaultStreamPartitioner.partition(DefaultStreamPartitioner.java:38)
	at org.apache.kafka.streams.processor.internals.StreamsMetadataState.getKeyQueryMetadataForKey(StreamsMetadataState.java:370)
	at org.apache.kafka.streams.processor.internals.StreamsMetadataState.getKeyQueryMetadataForKey(StreamsMetadataState.java:237)
	at org.apache.kafka.streams.processor.internals.StreamsMetadataState.getKeyQueryMetadataForKey(StreamsMetadataState.java:195)
	at org.apache.kafka.streams.KafkaStreams.queryMetadataForKey(KafkaStreams.java:1288)
	at io.confluent.ksql.execution.streams.materialization.ks.KsLocator.locate(KsLocator.java:88)
	at io.confluent.ksql.physical.pull.HARouting.handlePullQuery(HARouting.java:106)
	at io.confluent.ksql.engine.EngineExecutor.executePullQuery(EngineExecutor.java:171)
	at io.confluent.ksql.engine.KsqlEngine.executePullQuery(KsqlEngine.java:283)
	at io.confluent.ksql.rest.server.resources.streaming.PullQueryPublisher.lambda$subscribe$1(PullQueryPublisher.java:93)
	at io.confluent.ksql.rest.server.resources.streaming.PullQueryPublisher$PullQuerySubscription.request(PullQueryPublisher.java:140)
	at io.confluent.ksql.rest.server.resources.streaming.WebSocketSubscriber.onSubscribe(WebSocketSubscriber.java:47)
	at io.confluent.ksql.rest.server.resources.streaming.PullQueryPublisher.subscribe(PullQueryPublisher.java:109)
	at io.confluent.ksql.rest.server.resources.streaming.WSQueryEndpoint.startPullQueryPublisher(WSQueryEndpoint.java:385)
	at io.confluent.ksql.rest.server.resources.streaming.WSQueryEndpoint.handleQuery(WSQueryEndpoint.java:297)
	at io.confluent.ksql.rest.server.resources.streaming.WSQueryEndpoint.executeStreamQuery(WSQueryEndpoint.java:216)
	at io.confluent.ksql.rest.server.KsqlServerEndpoints.lambda$executeWebsocketStream$20(KsqlServerEndpoints.java:293)
	at io.confluent.ksql.rest.server.KsqlServerEndpoints.lambda$executeOnWorker$21(KsqlServerEndpoints.java:304)
	at io.vertx.core.impl.ContextImpl.lambda$executeBlocking$2(ContextImpl.java:313)
	at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1128)
	at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:628)
	at io.netty.util.concurrent.FastThreadLocalRunnable.run(FastThreadLocalRunnable.java:30)
	at java.base/java.lang.Thread.run(Thread.java:834)
```

This is extra concerning because we should not be registering anything for a pull query!

**To Reproduce**

I have not reproduced this locally, but is the table definition for the one that caused the issue:

```sql
CREATE TABLE REEFERS_AND_VESSELS_CLOSE_FOR_GT_2HR 
  WITH (KAFKA_TOPIC='REEFERS_AND_VESSELS_CLOSE_FOR_GT_2HR_V00') AS
SELECT   FISHING_VESSEL_MMSI,
         REEFER_MMSI,
         STRUCT("lat":=LATEST_BY_OFFSET(FISHING_VESSEL_LAT),"lon":=LATEST_BY_OFFSET(FISHING_VESSEL_LON)) AS LATEST_FISHING_VESSEL_LOCATION,
         STRUCT("lat":=LATEST_BY_OFFSET(REEFER_LAT),        "lon":=LATEST_BY_OFFSET(REEFER_LON))         AS LATEST_REEFER_LOCATION,
         MIN(DISTANCE_KM)                                                                                AS CLOSEST_DISTANCE_KM,
         MAX(DISTANCE_KM)                                                                                AS FURTHEST_DISTANCE_KM,
         MIN(FISHING_VESSEL_TS)                                                                          AS FIRST_TS,
         MAX(FISHING_VESSEL_TS)                                                                          AS LAST_TS,
         (MAX(FISHING_VESSEL_TS) - MIN(FISHING_VESSEL_TS)) / 1000                                        AS DIFF_SEC
FROM     REEFERS_AND_VESSELS_WITHIN_1KM WINDOW SESSION (10 MINUTES)
WHERE    IN_RANGE_AND_SPEED = 1
GROUP BY FISHING_VESSEL_MMSI,
         REEFER_MMSI
HAVING   (MAX(ROWTIME) - MIN(ROWTIME)) / 1000 > 7200;
```

The source for this was a stream that's a result of a stream-stream join and some filtering. 

**Expected behavior**

We can issue a pull query. 

**Actual behaviour**

See above

**Additional context**

The problem occurs on ccloud master at time of writing (which I believe is 0.15)

hli21:
### Description 
Fix #7174, Schema Registry 409 when issuing pull query against table with multi-column keys. 
Detail analysis of the problem:
https://github.com/confluentinc/ksql/issues/7174#issuecomment-1055944920

For a lookup query, pull query engine sends the lookup keys to kafka stream, which finds the partitions for the keys with the schema registry of the corresponding persistent query. The stream engine looks for the schema registry with two inputs: subject-key and schema. The subject-key and schema do not match and lead to the error.

### Testing done 
RestQueryTranslationTest and unit tests.

### Reviewer checklist

### DIFF
--- a/ksqldb-engine/src/main/java/io/confluent/ksql/query/MaterializationProviderBuilderFactory.java
+++ b/ksqldb-engine/src/main/java/io/confluent/ksql/query/MaterializationProviderBuilderFactory.java
@@ -15,7 +15,6 @@
 
 package io.confluent.ksql.query;
 
-import static io.confluent.ksql.serde.SerdeFeature.UNWRAP_SINGLES;
 import static java.util.Objects.requireNonNull;
 
 import io.confluent.ksql.GenericKey;
@@ -26,10 +25,8 @@ import io.confluent.ksql.execution.streams.materialization.ks.KsMaterialization;
 import io.confluent.ksql.execution.streams.materialization.ks.KsMaterializationFactory;
 import io.confluent.ksql.logging.processing.NoopProcessingLogContext;
 import io.confluent.ksql.schema.ksql.PhysicalSchema;
-import io.confluent.ksql.schema.ksql.types.SqlBaseType;
 import io.confluent.ksql.serde.GenericKeySerDe;
 import io.confluent.ksql.serde.KeyFormat;
-import io.confluent.ksql.serde.SerdeFeatures;
 import io.confluent.ksql.services.ServiceContext;
 import io.confluent.ksql.util.KsqlConfig;
 import java.util.Map;
--- a/ksqldb-engine/src/main/java/io/confluent/ksql/query/MaterializationProviderBuilderFactory.java
+++ b/ksqldb-engine/src/main/java/io/confluent/ksql/query/MaterializationProviderBuilderFactory.java
@@ -62,13 +62,13 @@ public final class MaterializationProviderBuilderFactory {
   }
 
   public MaterializationProviderBuilder materializationProviderBuilder(
-          final MaterializationInfo materializationInfo,
-          final PhysicalSchema querySchema,
-          final KeyFormat keyFormat,
-          final Map<String, Object> streamsProperties,
-          final String applicationId,
-          final String queryId,
-          final KeyFormat pullQueryKeyFormat) {
+      final MaterializationInfo materializationInfo,
+      final PhysicalSchema querySchema,
+      final KeyFormat keyFormat,
+      final Map<String, Object> streamsProperties,
+      final String applicationId,
+      final String queryId
+  ) {
     return (kafkaStreams, topology) -> buildMaterializationProvider(
         kafkaStreams,
         topology,
@@ -77,23 +77,22 @@ public final class MaterializationProviderBuilderFactory {
         keyFormat,
         streamsProperties,
         applicationId,
-        queryId,
-        pullQueryKeyFormat
+        queryId
     );
   }
 
   private Optional<MaterializationProvider> buildMaterializationProvider(
-          final KafkaStreams kafkaStreams,
-          final Topology topology,
-          final MaterializationInfo materializationInfo,
-          final PhysicalSchema schema,
-          final KeyFormat keyFormat,
-          final Map<String, Object> streamsProperties,
-          final String applicationId,
-          final String queryId,
-          final KeyFormat pullQueryKeyFormat) {
+      final KafkaStreams kafkaStreams,
+      final Topology topology,
+      final MaterializationInfo materializationInfo,
+      final PhysicalSchema schema,
+      final KeyFormat keyFormat,
+      final Map<String, Object> streamsProperties,
+      final String applicationId,
+      final String queryId
+  ) {
     final Serializer<GenericKey> keySerializer = new GenericKeySerDe().create(
-        pullQueryKeyFormat.getFormatInfo(),
+        keyFormat.getFormatInfo(),
         schema.keySchema(),
         ksqlConfig,
         serviceContext.getSchemaRegistryClientFactory(),
--- a/ksqldb-engine/src/main/java/io/confluent/ksql/query/MaterializationProviderBuilderFactory.java
+++ b/ksqldb-engine/src/main/java/io/confluent/ksql/query/MaterializationProviderBuilderFactory.java
@@ -85,23 +85,6 @@ public final class MaterializationProviderBuilderFactory {
     );
   }
 
-  private boolean isQuerySupported(final KeyFormat pullQueryKeyFormat,
-                                   final PhysicalSchema schema) {
-    final boolean isFormat = (pullQueryKeyFormat.getFormatInfo().getFormat()
-            .equalsIgnoreCase("KAFKA")
-            || pullQueryKeyFormat.getFormatInfo().getFormat().equalsIgnoreCase("DELIMITED"));
-    boolean notSupported = (isFormat && ((schema.keySchema().columns().size() != 1)
-            || schema.keySchema().columns().get(0).type().baseType() == SqlBaseType.STRUCT
-            || schema.keySchema().columns().get(0).type().baseType() == SqlBaseType.DECIMAL));
-    final SerdeFeatures features = pullQueryKeyFormat.getFeatures();
-    final boolean notSupportedNoneFormat = pullQueryKeyFormat.getFormatInfo().getFormat()
-            .equalsIgnoreCase("NONE") && (features.enabled(UNWRAP_SINGLES)
-            || schema.keySchema().columns().size() > 0);
-    notSupported = notSupported || notSupportedNoneFormat;
-
-    return !notSupported;
-  }
-
   private Optional<MaterializationProvider> buildMaterializationProvider(
           final KafkaStreams kafkaStreams,
           final Topology topology,
--- a/ksqldb-engine/src/main/java/io/confluent/ksql/query/MaterializationProviderBuilderFactory.java
+++ b/ksqldb-engine/src/main/java/io/confluent/ksql/query/MaterializationProviderBuilderFactory.java
@@ -93,7 +93,7 @@ public final class MaterializationProviderBuilderFactory {
           final String queryId,
           final KeyFormat pullQueryKeyFormat) {
     final Serializer<GenericKey> keySerializer = new GenericKeySerDe().create(
-        keyFormat.getFormatInfo(),
+        pullQueryKeyFormat.getFormatInfo(),
         schema.keySchema(),
         ksqlConfig,
         serviceContext.getSchemaRegistryClientFactory(),
@@ -101,17 +101,6 @@ public final class MaterializationProviderBuilderFactory {
         NoopProcessingLogContext.INSTANCE,
         Optional.empty()
     ).serializer();
-
-    final Serializer<GenericKey> pullQueryKeySerializer = new GenericKeySerDe().create(
-            pullQueryKeyFormat.getFormatInfo(),
-            schema.keySchema(),
-            ksqlConfig,
-            serviceContext.getSchemaRegistryClientFactory(),
-            "",
-            NoopProcessingLogContext.INSTANCE,
-            Optional.empty()
-    ).serializer();
-    ksMaterializationFactory.setPullQueryKeySerializer(pullQueryKeySerializer);
 
     final Optional<KsMaterialization> ksMaterialization = ksMaterializationFactory
         .create(
--- a/ksqldb-engine/src/main/java/io/confluent/ksql/query/MaterializationProviderBuilderFactory.java
+++ b/ksqldb-engine/src/main/java/io/confluent/ksql/query/MaterializationProviderBuilderFactory.java
@@ -46,6 +46,11 @@ public final class MaterializationProviderBuilderFactory {
   private final ServiceContext serviceContext;
   private final KsMaterializationFactory ksMaterializationFactory;
   private final KsqlMaterializationFactory ksqlMaterializationFactory;
+  private KeyFormat pullQueryKeyFormat;
+
+  public void setPullQueryKeyFormat(final KeyFormat pullQueryKeyFormat) {
+    this.pullQueryKeyFormat = pullQueryKeyFormat;
+  }
 
   public MaterializationProviderBuilderFactory(
       final KsqlConfig ksqlConfig,
@@ -100,6 +105,19 @@ public final class MaterializationProviderBuilderFactory {
         NoopProcessingLogContext.INSTANCE,
         Optional.empty()
     ).serializer();
+
+    if (!ksqlConfig.getBoolean(KsqlConfig.KSQL_SHARED_RUNTIME_ENABLED)) {
+      final Serializer<GenericKey> pullQueryKeySerializer = new GenericKeySerDe().create(
+              pullQueryKeyFormat.getFormatInfo(),
+              schema.keySchema(),
+              ksqlConfig,
+              serviceContext.getSchemaRegistryClientFactory(),
+              "",
+              NoopProcessingLogContext.INSTANCE,
+              Optional.empty()
+      ).serializer();
+      ksMaterializationFactory.setPullQueryKeySerializer(pullQueryKeySerializer);
+    }
 
     final Optional<KsMaterialization> ksMaterialization = ksMaterializationFactory
         .create(
--- a/ksqldb-engine/src/main/java/io/confluent/ksql/query/MaterializationProviderBuilderFactory.java
+++ b/ksqldb-engine/src/main/java/io/confluent/ksql/query/MaterializationProviderBuilderFactory.java
@@ -122,18 +122,16 @@ public final class MaterializationProviderBuilderFactory {
         Optional.empty()
     ).serializer();
 
-    if (isQuerySupported(pullQueryKeyFormat, schema)) {
-      final Serializer<GenericKey> pullQueryKeySerializer = new GenericKeySerDe().create(
-              pullQueryKeyFormat.getFormatInfo(),
-              schema.keySchema(),
-              ksqlConfig,
-              serviceContext.getSchemaRegistryClientFactory(),
-              "",
-              NoopProcessingLogContext.INSTANCE,
-              Optional.empty()
-      ).serializer();
-      ksMaterializationFactory.setPullQueryKeySerializer(pullQueryKeySerializer);
-    }
+    final Serializer<GenericKey> pullQueryKeySerializer = new GenericKeySerDe().create(
+            pullQueryKeyFormat.getFormatInfo(),
+            schema.keySchema(),
+            ksqlConfig,
+            serviceContext.getSchemaRegistryClientFactory(),
+            "",
+            NoopProcessingLogContext.INSTANCE,
+            Optional.empty()
+    ).serializer();
+    ksMaterializationFactory.setPullQueryKeySerializer(pullQueryKeySerializer);
 
     final Optional<KsMaterialization> ksMaterialization = ksMaterializationFactory
         .create(
--- a/ksqldb-engine/src/main/java/io/confluent/ksql/query/MaterializationProviderBuilderFactory.java
+++ b/ksqldb-engine/src/main/java/io/confluent/ksql/query/MaterializationProviderBuilderFactory.java
@@ -15,6 +15,7 @@
 
 package io.confluent.ksql.query;
 
+import static io.confluent.ksql.serde.SerdeFeature.UNWRAP_SINGLES;
 import static java.util.Objects.requireNonNull;
 
 import io.confluent.ksql.GenericKey;
@@ -25,8 +26,10 @@ import io.confluent.ksql.execution.streams.materialization.ks.KsMaterialization;
 import io.confluent.ksql.execution.streams.materialization.ks.KsMaterializationFactory;
 import io.confluent.ksql.logging.processing.NoopProcessingLogContext;
 import io.confluent.ksql.schema.ksql.PhysicalSchema;
+import io.confluent.ksql.schema.ksql.types.SqlBaseType;
 import io.confluent.ksql.serde.GenericKeySerDe;
 import io.confluent.ksql.serde.KeyFormat;
+import io.confluent.ksql.serde.SerdeFeatures;
 import io.confluent.ksql.services.ServiceContext;
 import io.confluent.ksql.util.KsqlConfig;
 import java.util.Map;
@@ -46,11 +49,6 @@ public final class MaterializationProviderBuilderFactory {
   private final ServiceContext serviceContext;
   private final KsMaterializationFactory ksMaterializationFactory;
   private final KsqlMaterializationFactory ksqlMaterializationFactory;
-  private KeyFormat pullQueryKeyFormat;
-
-  public void setPullQueryKeyFormat(final KeyFormat pullQueryKeyFormat) {
-    this.pullQueryKeyFormat = pullQueryKeyFormat;
-  }
 
   public MaterializationProviderBuilderFactory(
       final KsqlConfig ksqlConfig,
@@ -67,13 +65,13 @@ public final class MaterializationProviderBuilderFactory {
   }
 
   public MaterializationProviderBuilder materializationProviderBuilder(
-      final MaterializationInfo materializationInfo,
-      final PhysicalSchema querySchema,
-      final KeyFormat keyFormat,
-      final Map<String, Object> streamsProperties,
-      final String applicationId,
-      final String queryId
-  ) {
+          final MaterializationInfo materializationInfo,
+          final PhysicalSchema querySchema,
+          final KeyFormat keyFormat,
+          final Map<String, Object> streamsProperties,
+          final String applicationId,
+          final String queryId,
+          final KeyFormat pullQueryKeyFormat) {
     return (kafkaStreams, topology) -> buildMaterializationProvider(
         kafkaStreams,
         topology,
@@ -82,20 +80,38 @@ public final class MaterializationProviderBuilderFactory {
         keyFormat,
         streamsProperties,
         applicationId,
-        queryId
+        queryId,
+        pullQueryKeyFormat
     );
   }
 
+  private boolean isQuerySupported(final KeyFormat pullQueryKeyFormat,
+                                   final PhysicalSchema schema) {
+    final boolean isFormat = (pullQueryKeyFormat.getFormatInfo().getFormat()
+            .equalsIgnoreCase("KAFKA")
+            || pullQueryKeyFormat.getFormatInfo().getFormat().equalsIgnoreCase("DELIMITED"));
+    boolean notSupported = (isFormat && ((schema.keySchema().columns().size() != 1)
+            || schema.keySchema().columns().get(0).type().baseType() == SqlBaseType.STRUCT
+            || schema.keySchema().columns().get(0).type().baseType() == SqlBaseType.DECIMAL));
+    final SerdeFeatures features = pullQueryKeyFormat.getFeatures();
+    final boolean notSupportedNoneFormat = pullQueryKeyFormat.getFormatInfo().getFormat()
+            .equalsIgnoreCase("NONE") && (features.enabled(UNWRAP_SINGLES)
+            || schema.keySchema().columns().size() > 0);
+    notSupported = notSupported || notSupportedNoneFormat;
+
+    return !notSupported;
+  }
+
   private Optional<MaterializationProvider> buildMaterializationProvider(
-      final KafkaStreams kafkaStreams,
-      final Topology topology,
-      final MaterializationInfo materializationInfo,
-      final PhysicalSchema schema,
-      final KeyFormat keyFormat,
-      final Map<String, Object> streamsProperties,
-      final String applicationId,
-      final String queryId
-  ) {
+          final KafkaStreams kafkaStreams,
+          final Topology topology,
+          final MaterializationInfo materializationInfo,
+          final PhysicalSchema schema,
+          final KeyFormat keyFormat,
+          final Map<String, Object> streamsProperties,
+          final String applicationId,
+          final String queryId,
+          final KeyFormat pullQueryKeyFormat) {
     final Serializer<GenericKey> keySerializer = new GenericKeySerDe().create(
         keyFormat.getFormatInfo(),
         schema.keySchema(),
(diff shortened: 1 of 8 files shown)


ksqldb-engine/src/main/java/io/confluent/ksql/query/MaterializationProviderBuilderFactory.java
```diff
@@ -101,6 +106,19 @@ public MaterializationProviderBuilder materializationProviderBuilder(
         Optional.empty()
     ).serializer();
 
+    if (!ksqlConfig.getBoolean(KsqlConfig.KSQL_SHARED_RUNTIME_ENABLED)) {
```
agavra, line 109: what does this have to do with whether or not the shared runtime is enabled?

ksqldb-engine/src/main/java/io/confluent/ksql/query/QueryBuilder.java
```diff
@@ -290,20 +290,33 @@ PersistentQueryMetadata buildPersistentQueryInDedicatedRuntime(
...
-
+        pullQueryKeyFormat = dataSource.getKsqlTopic().getKeyFormat();
+        materializationProviderBuilderFactory.setPullQueryKeyFormat(pullQueryKeyFormat);
         break;
       default:
         logicalSchema = sinkDataSource.get().getSchema();
+        if (sinkDataSource.get().getKsqlTopic().getKeyFormat()
+                .getFormatInfo().getFormat().equalsIgnoreCase("JSON")
+                || sinkDataSource.get().getKsqlTopic().getValueFormat().getFormatInfo()
+                .getFormat().equalsIgnoreCase("JSON")
```
agavra, line 312: why is the behavior different for topics that have JSON key / value? What does the value format have to do with it? Can we add tests that show this (e.g. avro key with JSON value)?

ksqldb-functional-tests/src/test/resources/rest-query-validation-tests/pull-queries-against-ctas-materialized.json
```diff
@@ -794,6 +794,330 @@
         ]}
       ]
     },
+    {
+      "name": "windowed - select star with HAVING filter - AVRO",
```
agavra, line 798: are these related to this PR? I thought the problem was only with multi-column keys. I guess is it possible to window by a multi-column key?

ksqldb-functional-tests/src/test/resources/rest-query-validation-tests/pull-queries-with-predicates.json
```diff
@@ -224,6 +224,292 @@
...
+        {"admin": {"@type": "currentStatus"}},
+        {"admin": {"@type": "currentStatus"}},
+        {"query": [
+          {"header":{"schema":"`ID` INTEGER KEY, `V0` STRING KEY, `V1` STRING"}},
+          {"row":{"columns":[10, "v10a", "v10b"]}}
+        ]}
+      ]
+    },
+    {
+      "name": "schema registry table-test three partitions protobuf",
```
agavra, line 493: nit: is there any reason we need to test 1, 2 & 3 partitions? Could we test just `2`?

ksqldb-functional-tests/src/test/resources/rest-query-validation-tests/pull-queries-with-predicates.json
```diff
@@ -224,6 +224,292 @@
...
+          {"header":{"schema":"`ID` INTEGER KEY, `V0` STRING KEY, `V1` STRING"}},
+          {"row":{"columns":[10, "v10a", "v10b"]}}
+        ]}
+      ]
+    },
+    {
+      "name": "schema registry table-test one partition avro",
+      "statements": [
+        "CREATE TABLE INPUT (ID INT PRIMARY KEY, V0 STRING PRIMARY KEY, V1 STRING) WITH (kafka_topic='test_topic', format='AVRO', PARTITIONS=1);",
+        "CREATE TABLE MATVIEW AS SELECT * FROM INPUT;",
```
agavra, line 254: can we add a test for what happens when we create a table with a different format than the original?

e.g.
```
CREATE TABLE MATVIEW WITH (FORMAT='PROTOBUF') AS SELECT * FROM INPUT;
```

agavra (commented):
Almost LGTM, I'm not sure about why we're treating JSON differently and why we're treating the shared runtime differently

ksqldb-engine/src/main/java/io/confluent/ksql/query/QueryBuilder.java
```diff
@@ -290,20 +290,33 @@ PersistentQueryMetadata buildPersistentQueryInDedicatedRuntime(
     final KeyFormat keyFormat;
     final ValueFormat valueFormat;
     final KsqlTopic ksqlTopic;
+    final KeyFormat pullQueryKeyFormat;
 
     switch (persistentQueryType) {
       // CREATE_SOURCE does not have a sink, so the schema is obtained from the query source
       case CREATE_SOURCE:
-        final DataSource dataSource = Iterables.getOnlyElement(sources);
+        DataSource dataSource = Iterables.getOnlyElement(sources);
```
AlanConfluent, line 298: Do we want to make a similar change in `buildPersistentQueryInSharedRuntime`?

ksqldb-engine/src/main/java/io/confluent/ksql/query/QueryBuilder.java
hli21: For JSON format, it uses JsonSerde from sink and does not interact with schema registry. This keeps the existing logic.

ksqldb-engine/src/main/java/io/confluent/ksql/query/QueryBuilder.java
```diff
@@ -290,20 +290,33 @@ PersistentQueryMetadata buildPersistentQueryInDedicatedRuntime(
...
+                || sinkDataSource.get().getKsqlTopic().getValueFormat().getFormatInfo()
+                .getFormat().equalsIgnoreCase("JSON")
+                || sources.size() != 1) {
+          pullQueryKeyFormat = sinkDataSource.get().getKsqlTopic().getKeyFormat();
+        } else {
+          dataSource = Iterables.getOnlyElement(sources);
+          pullQueryKeyFormat = dataSource.getKsqlTopic().getKeyFormat();
+        }
+        materializationProviderBuilderFactory.setPullQueryKeyFormat(pullQueryKeyFormat);
         keyFormat = sinkDataSource.get().getKsqlTopic().getKeyFormat();
```
AlanConfluent, line 320: The main thing that `keyFormat` is used for is the materialization info for pull queries.  Do we need to have a separate reference to `pullQueryKeyFormat` or should we just change that reference to be the right thing?

ksqldb-engine/src/main/java/io/confluent/ksql/query/QueryBuilder.java
```diff
@@ -290,20 +290,33 @@ PersistentQueryMetadata buildPersistentQueryInDedicatedRuntime(
...
         logicalSchema = sinkDataSource.get().getSchema();
+        if (sinkDataSource.get().getKsqlTopic().getKeyFormat()
+                .getFormatInfo().getFormat().equalsIgnoreCase("JSON")
+                || sinkDataSource.get().getKsqlTopic().getValueFormat().getFormatInfo()
+                .getFormat().equalsIgnoreCase("JSON")
+                || sources.size() != 1) {
+          pullQueryKeyFormat = sinkDataSource.get().getKsqlTopic().getKeyFormat();
+        } else {
+          dataSource = Iterables.getOnlyElement(sources);
+          pullQueryKeyFormat = dataSource.getKsqlTopic().getKeyFormat();
```
AlanConfluent, line 317: I'm a bit confused about in what scenarios we need to use the source key vs the sink key.  I would think that the format would depend on what's actually going to be materialized, which usually resembles the sink.  In what situations does it resemble the source?

AlanConfluent (commented):
I remember that we had talked about potentially doing a scatter gather for a certain type of compound key as a "workaround" to avoid doing a pure keybased lookup which would require the serde and cause the error.

This appears better because it seems that we're just trying to use the right serde, which is the source the problem.  I'm a little confused though why some queries require using the source key format rather than the sink.  It might be good to explain that a bit in the description of the PR.

ksqldb-engine/src/main/java/io/confluent/ksql/query/QueryBuilder.java
agavra: @hli21 but it wouldn't be a problem to treat JSON the same way, right? I think this special casing is pretty confusing.

ksqldb-engine/src/main/java/io/confluent/ksql/query/MaterializationProviderBuilderFactory.java
hli21: Add same logic to the shared runtime code path.

ksqldb-functional-tests/src/test/resources/rest-query-validation-tests/pull-queries-against-ctas-materialized.json
hli21: We do not have test cases for AVRO and protobuf formats. Add test cases for those formats and add a multi-column key test.

ksqldb-functional-tests/src/test/resources/rest-query-validation-tests/pull-queries-with-predicates.json
hli21: The purpose is to test different partition numbers.

ksqldb-engine/src/main/java/io/confluent/ksql/query/QueryBuilder.java
hli21: Remove this check.

ksqldb-engine/src/main/java/io/confluent/ksql/query/MaterializationProviderBuilderFactory.java
```diff
@@ -46,6 +46,11 @@
   private final ServiceContext serviceContext;
   private final KsMaterializationFactory ksMaterializationFactory;
   private final KsqlMaterializationFactory ksqlMaterializationFactory;
+  private KeyFormat pullQueryKeyFormat;
```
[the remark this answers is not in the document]
hli21: Remove this. Use function parameter.

ksqldb-engine/src/main/java/io/confluent/ksql/query/QueryBuilder.java
```diff
@@ -290,20 +290,33 @@ PersistentQueryMetadata buildPersistentQueryInDedicatedRuntime(
...
+        }
+        materializationProviderBuilderFactory.setPullQueryKeyFormat(pullQueryKeyFormat);
```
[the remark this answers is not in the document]
hli21: Done.

ksqldb-streams/src/main/java/io/confluent/ksql/execution/streams/materialization/ks/KsLocator.java
```diff
@@ -269,7 +272,7 @@ protected KeyQueryMetadata getKeyQueryMetadata(final KsqlKey key) {
       return ((KafkaStreamsNamedTopologyWrapper) kafkaStreams)
           .queryMetadataForKey(storeName, key.getKey(), keySerializer, queryId);
     }
-    return kafkaStreams.queryMetadataForKey(storeName, key.getKey(), keySerializer);
+    return kafkaStreams.queryMetadataForKey(storeName, key.getKey(), pullQueryKeySerializer);
```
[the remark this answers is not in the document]
hli21: Use pullQueryKeySerializer for shared runtime too.

ksqldb-engine/src/main/java/io/confluent/ksql/query/QueryBuilder.java
hli21: Done

ksqldb-engine/src/main/java/io/confluent/ksql/query/QueryBuilder.java
hli21: Change the current logic of keyFormat breaks some existing code paths. Many QueryTranslationTest cases fail.

ksqldb-engine/src/main/java/io/confluent/ksql/query/QueryBuilder.java
hli21: For materialized view of joins, sink will be used.

ksqldb-engine/src/main/java/io/confluent/ksql/query/QueryBuilder.java
```diff
@@ -235,6 +238,23 @@ TransientQueryMetadata buildTransientQuery(
     return Optional.empty();
   }
 
+  private boolean isQuerySupported(final KeyFormat pullQueryKeyFormat,
+                                   final LogicalSchema schema) {
+    final boolean isFormat = (pullQueryKeyFormat.getFormatInfo().getFormat()
+            .equalsIgnoreCase("KAFKA")
+            || pullQueryKeyFormat.getFormatInfo().getFormat().equalsIgnoreCase("DELIMITED"));
+    boolean notSupported = (isFormat && ((schema.key().size() != 1)
+            || schema.key().get(0).type().baseType() == SqlBaseType.STRUCT
```
AlanConfluent, line 247: Do you need to check the other keys beyond the 0th entry?

AlanConfluent (commented):
Can you say a little more in the description the role of the format.  It could also make it a little easier to understand if you restate the solution there.  When I was originally discussing this with you, I thought that the wrong physical schema was being registered by streams, swapping the sink for the source.  It seems that it's just the format, which makes a bit more sense (I thought how could this even work otherwise!).  Also, I'm assuming this is a workaround since this is how it's just currently done in Streams and is easier than making the change there.  Is that correct?

ksqldb-engine/src/main/java/io/confluent/ksql/query/MaterializationProviderBuilderFactory.java
```diff
@@ -101,6 +102,17 @@ public MaterializationProviderBuilder materializationProviderBuilder(
...
+    final Serializer<GenericKey> pullQueryKeySerializer = new GenericKeySerDe().create(
+            pullQueryKeyFormat.getFormatInfo(),
+            schema.keySchema(),
+            ksqlConfig,
+            serviceContext.getSchemaRegistryClientFactory(),
+            "",
+            NoopProcessingLogContext.INSTANCE,
+            Optional.empty()
+    ).serializer();
+    ksMaterializationFactory.setPullQueryKeySerializer(pullQueryKeySerializer);
```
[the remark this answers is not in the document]
hli21: Added description. The setter is not needed.

ksqldb-functional-tests/src/test/resources/rest-query-validation-tests/pull-queries-with-predicates.json
```diff
@@ -224,6 +224,204 @@
...
+        {"query": [
+          {"header":{"schema":"`ID` INTEGER KEY, `V0` STRING KEY, `CNT` BIGINT"}},
+          {"row":{"columns":[10, "v10a", 1]}}
+        ]}
+      ]
+    },
+    {
+      "name": "schema registry stream-test two partitions protobuf",
+      "statements": [
+        "CREATE STREAM INPUT (ID INT KEY, V0 STRING KEY, V1 STRING) WITH (kafka_topic='test_topic', format='PROTOBUF', PARTITIONS=2);",
```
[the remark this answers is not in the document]
hli21: Yes. The "create table" case also triggers this.

ksqldb-streams/src/main/java/io/confluent/ksql/execution/streams/materialization/ks/KsLocator.java
```diff
@@ -267,9 +270,9 @@
   protected KeyQueryMetadata getKeyQueryMetadata(final KsqlKey key) {
     if (sharedRuntimesEnabled && kafkaStreams instanceof KafkaStreamsNamedTopologyWrapper) {
       return ((KafkaStreamsNamedTopologyWrapper) kafkaStreams)
-          .queryMetadataForKey(storeName, key.getKey(), keySerializer, queryId);
+          .queryMetadataForKey(storeName, key.getKey(), pullQueryKeySerializer, queryId);
```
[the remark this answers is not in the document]
hli21: Only use keySerializer now.

ksqldb-engine/src/main/java/io/confluent/ksql/query/QueryBuilder.java
```diff
@@ -290,20 +310,30 @@ PersistentQueryMetadata buildPersistentQueryInDedicatedRuntime(
...
         logicalSchema = dataSource.getSchema();
         keyFormat = dataSource.getKsqlTopic().getKeyFormat();
         valueFormat = dataSource.getKsqlTopic().getValueFormat();
         ksqlTopic = dataSource.getKsqlTopic();
-
+        pullQueryKeyFormat = dataSource.getKsqlTopic().getKeyFormat();
         break;
       default:
         logicalSchema = sinkDataSource.get().getSchema();
+        if (sources.size() != 1) {
```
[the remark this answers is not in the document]
hli21: The isQuerySupported function needs source key format as a parameter.

ksqldb-engine/src/main/java/io/confluent/ksql/query/QueryBuilder.java
```diff
@@ -235,6 +238,23 @@ TransientQueryMetadata buildTransientQuery(
     return Optional.empty();
   }
 
+  private boolean isQuerySupported(final KeyFormat pullQueryKeyFormat,
```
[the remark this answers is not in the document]
hli21: Rename the function and add comment.

ksqldb-engine/src/main/java/io/confluent/ksql/query/QueryBuilder.java
```diff
@@ -235,6 +238,23 @@ TransientQueryMetadata buildTransientQuery(
...
+                                   final LogicalSchema schema) {
+    final boolean isFormat = (pullQueryKeyFormat.getFormatInfo().getFormat()
```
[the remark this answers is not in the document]
hli21: Done

ksqldb-engine/src/main/java/io/confluent/ksql/query/QueryBuilder.java
hli21: Only need to check the 0th entry.

ksqldb-engine/src/main/java/io/confluent/ksql/query/MaterializationProviderBuilderFactory.java
```diff
@@ -77,22 +77,23 @@ public MaterializationProviderBuilder materializationProviderBuilder(
...
-      final KeyFormat keyFormat,
-      final Map<String, Object> streamsProperties,
-      final String applicationId,
-      final String queryId
-  ) {
+          final KafkaStreams kafkaStreams,
+          final Topology topology,
+          final MaterializationInfo materializationInfo,
+          final PhysicalSchema schema,
+          final KeyFormat keyFormat,
```
AlanConfluent, line 90: I don't think you need a separate `keyFormat` for the `keyFormat.getWindowInfo()`, do you -- can you just use the same value as `pullQueryKeyFormat`?  When I was saying you could remove `pullQueryKeyFormat` earlier, I thought you could remove it all the way to the top where you're deciding whether or not a special case is needed.  I'll defer to you on this, since you've tested it, but it's confusing when `keyFormat` is used vs `pullQueryKeyFormat`.  They're all for pull queries! (That's the only functionality that uses this materialization provider)

ksqldb-engine/src/main/java/io/confluent/ksql/query/QueryBuilder.java
```diff
@@ -290,20 +311,30 @@ PersistentQueryMetadata buildPersistentQueryInDedicatedRuntime(
     final KeyFormat keyFormat;
     final ValueFormat valueFormat;
     final KsqlTopic ksqlTopic;
+    KeyFormat pullQueryKeyFormat;
```
AlanConfluent, line 314: I still think you can just over write `keyFormat` directly as opposed to introducing `pullQueryKeyFormat` but maybe there's a reason they need to be separate?  It wasn't obvious from looking at the code.

ksqldb-streams/src/test/java/io/confluent/ksql/execution/streams/materialization/ks/KsMaterializationFactoryTest.java
```diff
@@ -68,7 +68,8 @@
   private Topology topology;
   @Mock
   private Serializer<GenericKey> keySerializer;
-
+  @Mock
+  private Serializer<GenericKey> pullQueryKeySerializer;
```
AlanConfluent, line 72: Uneeded?

AlanConfluent (approved):

ksqldb-engine/src/main/java/io/confluent/ksql/query/QueryBuilder.java
```diff
@@ -235,6 +238,24 @@ TransientQueryMetadata buildTransientQuery(
     return Optional.empty();
   }
 
+  //This covers the queries that source-key format does not support.
+  private boolean isUnsupportedFormat(final KeyFormat pullQueryKeyFormat,
```
agavra, line 242: I'm still really confused about this. As much as I "understand" the fix, we should be using the same serializer for looking up the key as we do when serializing it into the changelog topic. That should be true independently of whether or not the format uses schema registry or not.

Why is it different if the key is a struct or a decimal? Why is that different if the format is `KAFKA`?

I recall you saying that some QTTs failed without this method, we should figure out why that's the case because I'm worried we may be introducing a different bug. This shouldn't affect QTTs at all, right? It should only affect the pull query path.

At a minimum, we should add a detailed comment into the code explaining this (since it really is a "hack")

agavra (changes requested):
Sorry to block this PR, but I really don't understand what's going on with the `isUnsupportedFormat` method - I think it would be difficult in the future for someone else to understand what's going on as well. At a minimum, we should have a super detailed explanation in the PR description of how and why this fixes the problem and ideally have that in code as comments as well.

ksqldb-streams/src/main/java/io/confluent/ksql/execution/streams/materialization/ks/KsLocator.java
```diff
@@ -119,9 +119,15 @@
     // Depending on whether this is a key-based lookup, determine which metadata method to use.
     // If we don't have keys, find the metadata for all partitions since we'll run the query for
     // all partitions of the state store rather than a particular one.
-    final List<PartitionMetadata> metadata = keys.isEmpty() || isRangeScan
-        ? getMetadataForAllPartitions(filterPartitions, keySet)
-        : getMetadataForKeys(keys, filterPartitions);
+    //For issue #7174. Temporarily turn off metadata finding for a partition with keys
+    //if there are more than one key.
+    final List<PartitionMetadata> metadata;
+    if (keys.size() == 1 && keys.get(0).getKey().size() == 1 && !isRangeScan) {
```
AlanConfluent, line 125: I thought the issue happens if we have a multi-column key, not if they're searching for multiple instances of a key.  If `keys.size() > 1`, this is like `KEY_COL IN (1, 2, 3, 4, 5);`, which I would think shouldn't trigger the bug.

You can always look at the schema or even grab one of the keys to see if there's more than one column to determine if it's multi-column.

This change would technically still work, I'm just trying not to entirely get rid of the precise forwarding optimization where we can still keep it.

ksqldb-streams/src/test/java/io/confluent/ksql/execution/streams/materialization/ks/KsLocatorTest.java
```diff
@@ -461,6 +462,7 @@ public void shouldReturnOneStandByWhenActiveAndOtherStandByDown() {
     assertThat(nodeList.stream().findFirst().get(), is(standByNode2));
   }
 
+  @Ignore
```
AlanConfluent, line 465: Why are you disabling this test?  Might be good to comment on that here.

ksqldb-engine/src/main/java/io/confluent/ksql/query/QueryBuilder.java
hli21: Added comments and link of analysis to the description of PR.

Taymindis:
Group By Struct data type will cause this issue again.

PR #8877, opened 2022-03-11
feat: Generalize the UDAFs collect_list and collect_set

jnh5y:
Addresses https://github.com/confluentinc/ksql/issues/7697

### Description 
The goal is to allow the UDAFs collect_list and collect_set to operate on arbitrary input types.  

### Testing done 
QTTs have been added to show the functions working generally.

### Reviewer checklist


ksqldb-engine/src/main/java/io/confluent/ksql/function/udaf/array/CollectListUdaf.java
```diff
@@ -47,54 +51,51 @@ private CollectListUdaf() {
...
+  public static <T> TableUdaf<T, List<T>, List<T>> createCollectListT() {
+    return new Collect<>();
+  }
+
   public static TableUdaf<Long, List<Long>, List<Long>> createCollectListLong() {
     return new Collect<>();
   }
 
-  @UdafFactory(description = "collect values of an Integer field into a single Array")
   public static TableUdaf<Integer, List<Integer>, List<Integer>> createCollectListInt() {
```
jnh5y, line 62: These functions are unneeded now.  That said, they are public.  What do folks think about removing them?

agavra: yeah we can remove them, nobody should be calling these directly

ksqldb-functional-tests/src/test/resources/query-validation-tests/collect-set.json
```diff
@@ -8,6 +8,86 @@
     "a join or another aggregate."
   ],
   "tests": [
+    {
+      "name": "collect_set with arrays, structs, and maps",
```
agavra, line 12: can we add tests for both of these for decimal type as well (and nested decimal). We should make sure that the set membership check works. 

Also what if we have a struct with a `null` value in one of the fields (we should make sure the set membership check works there as well)? What about a `null` at the top level?

agavra (approved):
Nice! You can definitely remove all the `@UdafFactory` methods. Nobody should be calling them publicly and it'll clean up a lot of code. Love to see it.

Issue #8368, opened 2021-11-16
`latest_by_offset` does not support `DECIMAL`


krisajenkins:
**Describe the bug**
There is no `latest_by_offset` implementation for the `DECIMAL` type. This is a problem if, for example, you want the latest price of something.

**To Reproduce**
```sql
CREATE OR REPLACE STREAM stocks (
  stock_id VARCHAR KEY,
  description VARCHAR,
  price DECIMAL(10,2)
) WITH (
  KAFKA_TOPIC = 'stocks',
  VALUE_FORMAT = 'avro',
  PARTITIONS = 3
);

INSERT INTO stocks ( stock_id, description, price ) VALUES ( 'CFLT', 'Confluent Inc.', 80.02 );
INSERT INTO stocks ( stock_id, description, price ) VALUES ( 'CFLT', 'Confluent Inc.', 81.02 );
INSERT INTO stocks ( stock_id, description, price ) VALUES ( 'NASDAQ', 'NASDAQ 100', 4003.24 );
INSERT INTO stocks ( stock_id, description, price ) VALUES ( 'CFLT', 'Confluent Inc.', 92.02 );
INSERT INTO stocks ( stock_id, description, price ) VALUES ( 'CFLT', 'Confluent Inc.', 83.02 );
INSERT INTO stocks ( stock_id, description, price ) VALUES ( 'NASDAQ', 'NASDAQ 100', 4028.19 );
INSERT INTO stocks ( stock_id, description, price ) VALUES ( 'CFLT', 'Confluent Inc.', 81.02 );

SET 'auto.offset.reset' = 'earliest';
SET 'ksql.query.pull.table.scan.enabled' = 'true';

CREATE OR REPLACE TABLE current_stock_prices AS
  SELECT
    stock_id,
    latest_by_offset( price ) AS price
  FROM stocks
  GROUP BY stock_id;

SELECT *
FROM current_stock_prices;

```

**Expected behavior**
I was expecting a table of two prices, `CFLT=81.02` and `NASDAQ=4028.19`.

**Actual behaviour**
```

 Message        
----------------
 Stream created 
----------------
Successfully changed local property 'auto.offset.reset' to 'earliest'. Use the UNSET command to revert your change.
Successfully changed local property 'ksql.query.pull.table.scan.enabled' to 'true'. Use the UNSET command to revert your change.
Could not determine output schema for query due to error: Function 'LATEST_BY_OFFSET' does not accept parameters (DECIMAL(10, 2)).
[...]

PR #8878, opened 2022-03-11
feat: Generalize the UDAFs earliest_by_offset and latest_by_offset

jnh5y:
Addresses: https://github.com/confluentinc/ksql/issues/5437 and https://github.com/confluentinc/ksql/issues/8368

### Description 
The goal is to allow the UDAFs earliest_by_offset and latest_by_offset to operate on arbitrary input types.

### Testing done 
QTTs have been added to show the functions working generally.

### Reviewer checklist


ksqldb-engine/src/test/java/io/confluent/ksql/function/udaf/offset/EarliestByOffsetTest.java
```diff
@@ -86,7 +89,8 @@ public void shouldThrowExceptionForInvalidN() {
   @Test
   public void shouldComputeEarliestInteger() {
     // Given:
-    final Udaf<Integer, Struct, Integer> udaf = EarliestByOffset.earliestInteger();
+    final Udaf<Integer, Struct, Integer> udaf = EarliestByOffset.earliest();
+    udaf.initializeTypeArguments(Collections.singletonList(SqlArgument.of(SqlTypes.INTEGER)));
```
jnh5y, line 93: The biggest question I have is if we want to keep these unit tests.  

The previous methods like `earliestInteger` have been removed.  Since `earliest_by_offset` is now configured with the input SQL type, this test is really quite silly looking.  That said, with the modification, it does show that all the various details work.

Should these be left or removed?  Thoughts?

jnh5y:
I've put this up as a draft PR to ask a question about how the unit tests should be managed.  Once that's settled, I'll sort out `latest_by_offset` and look for an approval.

ksqldb-engine/src/test/java/io/confluent/ksql/function/UdfLoaderTest.java
```diff
@@ -1000,7 +1000,7 @@ public void shouldThrowIfMissingAggregateTypeSchema() throws Exception {
     assertThat(e.getMessage(), containsString("Must specify 'aggregateSchema' for STRUCT parameter in @UdafFactory."));
   }
 
-  @Test
+  //@Test
```
jnh5y, line 1003: Ok, another thing to discuss... functions either need an annotated type or compute aggregate/return types.  

I'll think about these more next week.  Shout if you have an options about how to sort this out.

ksqldb-engine/src/main/java/io/confluent/ksql/function/UdafFactoryInvoker.java
```diff
@@ -70,10 +72,12 @@
     if (!Modifier.isStatic(method.getModifiers())) {
       throw new KsqlException("UDAF factory methods must be static " + method);
     }
-    final UdafTypes types = new UdafTypes(method, functionName, typeParser);
+    this.types = new UdafTypes(method, functionName, typeParser);
     this.functionName = Objects.requireNonNull(functionName);
-    this.aggregateArgType = Objects.requireNonNull(types.getAggregateSchema(aggregateSchema));
-    this.aggregateReturnType = Objects.requireNonNull(types.getOutputSchema(outputSchema));
+    this.aggregateSchema = aggregateSchema;
+    this.outputSchema = outputSchema;
```
agavra, line 78: nit: we should keep the `Objects.requireNonNull` (unless it can be null?) also remove the commented out lines :)

ksqldb-engine/src/main/java/io/confluent/ksql/function/udaf/offset/KudafByOffsetUtils.java
```diff
@@ -20,12 +20,14 @@
...
 
 import java.util.Comparator;
+import java.util.concurrent.atomic.AtomicLong;
 import org.apache.kafka.connect.data.Schema;
 import org.apache.kafka.connect.data.SchemaBuilder;
 import org.apache.kafka.connect.data.Struct;
 
 final class KudafByOffsetUtils {
 
+  static final AtomicLong sequence = new AtomicLong();
```
agavra, line 30: is this used?

ksqldb-engine/src/test/java/io/confluent/ksql/function/UdfLoaderTest.java
agavra: 😅  😆  😈  I know I promised I'd drop it, but if we went back to having a wrapper on the three methods you could check if calling that returns the default implementation or not

ksqldb-engine/src/test/java/io/confluent/ksql/function/udaf/offset/EarliestByOffsetTest.java
agavra: I'm +1 to removing it - I don't think it adds much value, especially if we have the QTTs

ksqldb-engine/src/test/java/io/confluent/ksql/function/UdfLoaderTest.java
jnh5y: This test / behavior is changing since one can either have types given with an annotation or via the new methods.  

If we had a wrapper, we'd still be in the same situation between annotations and the wrapper.  🤷 

I'll see what I come up when I look again.

ksqldb-engine/src/main/java/io/confluent/ksql/function/UdafFactoryInvoker.java
jnh5y: This value will be null if the annotation is unused.

ksqldb-engine/src/test/java/io/confluent/ksql/function/UdfLoaderTest.java
jnh5y: Ok, I've sorted this out by adding a little to the error to indicate choice in implementation.

agavra (approved):

Alter11:
I still have the same issue on Confluent 7.1.2 on premises version, the LATEST_BY_OFFSET can not be used with Array type, is this a regression on Confluent 7.1.2? I also tried v7.1.1 and v7.0.0

PR #8888, opened 2022-03-14
fix: Reduce query creation time with shared runtimes

ableegoldman:
Benchmarks showed that creating 100 queries took up to 5 hours. This is likely because we currently wait on the future returned by `#addNamedTopology`, which blocks until all threads in the runtimes have ack'ed the new topology. Since this requires exiting the processing loop to check for topology updates before polling, and this depends on the configured `max.poll.interval.ms` for which we increased the default value with shared runtimes, waiting on this future can take a long time and will only increase as more queries are added.

Since we only really care about the topology being synced to the runtime's TopologyMetadata, and not whether the threads on a particular client have received it yet, we should just remove the `.all().get()` in `SharedKafkaStreamsRuntime#start(query)`

### DIFF
--- a/ksqldb-engine/src/main/java/io/confluent/ksql/util/SharedKafkaStreamsRuntimeImpl.java
+++ b/ksqldb-engine/src/main/java/io/confluent/ksql/util/SharedKafkaStreamsRuntimeImpl.java
@@ -220,15 +220,7 @@ public class SharedKafkaStreamsRuntimeImpl extends SharedKafkaStreamsRuntime {
     log.info("Attempting to start query {} in runtime {}", queryId, getApplicationId());
     if (collocatedQueries.containsKey(queryId) && !collocatedQueries.get(queryId).everStarted) {
       if (!kafkaStreams.getTopologyByName(queryId.toString()).isPresent()) {
-        try {
           kafkaStreams.addNamedTopology(collocatedQueries.get(queryId).getTopology());
-        } catch (final Exception e) {
-          throw new IllegalStateException(String.format(
-                "Encountered an error when trying to start query %s in runtime: %s",
-                queryId,
-                getApplicationId()),
-              e);
-        }
       } else {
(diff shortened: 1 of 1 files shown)


ksqldb-engine/src/main/java/io/confluent/ksql/util/SharedKafkaStreamsRuntimeImpl.java
```diff
@@ -221,16 +221,13 @@ public void start(final QueryId queryId) {
     if (collocatedQueries.containsKey(queryId) && !collocatedQueries.get(queryId).everStarted) {
       if (!kafkaStreams.getTopologyByName(queryId.toString()).isPresent()) {
         try {
-          kafkaStreams.addNamedTopology(collocatedQueries.get(queryId).getTopology())
-              .all()
-              .get();
-        } catch (ExecutionException | InterruptedException e) {
-          final Throwable t = e.getCause() == null ? e : e.getCause();
+          kafkaStreams.addNamedTopology(collocatedQueries.get(queryId).getTopology());
+        } catch (final Exception e) {
```
wcarlson5, line 225: I don't think we want to wrap all exception here

guozhangwang: What kind of exceptions can be thrown here? It seems inside `addNamedTopology` we do not throw directly but always via the future object.

ableegoldman:
Jenkins is having some kind of issues right now and failing on some build step consistently, however you can see in [the logs for this build](https://jenkins.confluent.io/job/Confluent%20Public%20Repo%20PR%20builder/job/ksql/job/PR-8888/4/console) that the job did indeed finish running all the tests and reported it passing some time before the step it keeps tripping on:
```
18:31:07  	Passed: 13455, Failed: 0, Skipped: 5
```

ableegoldman:
Merged to master cc @wcarlson5

PR #8889, opened 2022-03-14
fix: decrease default number of shared runtimes

ableegoldman:
Since we bumped up the default number of StreamThreads per runtime to 4, we should decrease the default number of runtimes accordingly. 

The optimal number should ultimately be derived from the benchmarks, and we have a LaunchDarkly flag so we can modify the default further as more results come out, but for now we should still set the default in the code based on our initial observations.

If we start with just two runtimes, then with 4 threads each we will end up with 8 StreamThreads by default, which follows our rule of thumb for configuring the num.stream.threads (aka 2x # of cores on each node)


wcarlson5 (approved):
LGTM

PR #8892, opened 2022-03-15
fix: CLI should return non-zero error code on failure

### LINKED ISSUE #7051: RUN SCRIPT does not return a true exit code
**Describe the bug**

When using `RUN SCRIPT` or `--file` an exit code does not reflect that outcome.
A return code should return `0` or other int. 

Today we always return `0`, regardless to the result. 

**To Reproduce**
1. The version of KSQL. CLI v6.1.0, Server v6.1.0 located at http://localhost:8088
2. Sample source data.  

q.sql
```
CREATE STREAM stream1 (profileId VARCHAR, latitude DOUBLE, longitude DOUBLE)
  WITH (kafka_topic='topic1', value_format='json', partitions=1);
```

Bash file: 

```
#!/usr/bin/env bash
ksql --file q.sql 2> /dev/null

if [ $? -eq 0 ]
then
  echo "Successfully created queries"
  exit 0
else
  echo "Could not create queries" >&2
  exit 1
fi
```

Outcome: 

```
% ./sql.sh

 Message        
----------------
 Stream created 
----------------
Successfully created queries
% ./sql.sh
Cannot add stream 'STREAM1': A stream with the same name already exists
Successfully created queries

```

3. Any SQL statements you ran

```
CREATE STREAM stream1 (profileId VARCHAR, latitude DOUBLE, longitude DOUBLE)
  WITH (kafka_topic='topic1', value_format='json', partitions=1);
```
**Expected behavior**
Return exit code the reflects the outcome of the script

**Actual behaviour**
Returned exit code is always 0.

mjsax:
### Description 
If there is an error executing a script or single command, the CLI should return error code -1 instead of 0 (no error).

Closes #7051 

### Testing done 
Extended existing unit tests.

### Reviewer checklist

### DIFF
--- a/ksqldb-cli/src/main/java/io/confluent/ksql/Ksql.java
+++ b/ksqldb-cli/src/main/java/io/confluent/ksql/Ksql.java
@@ -69,14 +69,17 @@ public final class Ksql {
       options.setPassword(readPassword());
     }
 
+    int error_code = 0;
     try {
-      new Ksql(options, System.getProperties(), KsqlRestClient::create, Cli::build).run();
+      error_code = new Ksql(options, System.getProperties(), KsqlRestClient::create, Cli::build).run();
     } catch (final Exception e) {
       final String msg = ErrorMessageUtil.buildErrorMessage(e);
       LOGGER.error(msg);
       System.err.println(msg);
       System.exit(-1);
     }
+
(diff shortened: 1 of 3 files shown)


lihaosky (approved):
Nice!

PR #8898, opened 2022-03-16
fix: make sure topologies are done added before removing

wcarlson5:
This commit https://github.com/confluentinc/ksql/commit/f3c58508a379543b0e545a9cb03af1e9166760c3 is  causing drop queries to block.

The behavior in streams is:
All threads are waiting for a non empty topology
A topology gets added.
Thread 1 updates, others do not yet
topology is removed Thread 1 removes and unsubscribes and goes back to waiting for a non empty topology
All other threads have not moved and so the remove future can never complete

So before all removing a topology we make sure all the adds are compelete

### Reviewer checklist


ksqldb-engine/src/main/java/io/confluent/ksql/util/SharedKafkaStreamsRuntimeImpl.java
```diff
@@ -42,6 +46,7 @@
   private final long shutdownTimeout;
   private final QueryErrorClassifier errorClassifier;
   private final int maxQueryErrorsQueueSize;
+  private final List<KafkaFuture<Void>>  topolgogiesToAdd;
```
jnh5y, line 49: nit: Extra space:)

Issue #8861, opened 2022-03-08
JSON deserializer error with nested maps


coughman:
**Describe the bug**
For data in stream with nested map attributes (ie. map of maps) and its value format is JSON_SR, select query will fail with serialization exception. Note that inserts still works.

**To Reproduce**
Steps to reproduce the behavior, include:
1. The version of KSQL: 0.24.0

# create the following stream with nested map (attribute m)

create stream if not exists nestedmap_json_sr 
(
    id int,
    m Map<String, Map<String, int>>
)
with (
    kafka_topic = 'nestedmap_json_sr',
    partitions = 1,
    value_format = 'JSON_SR'
);

# insert some data

insert into nestedmap_json_sr (id, m) values (1, MAP('name' := MAP('nested1' := 1)));
insert into nestedmap_json_sr (id, m) values (2, MAP('name' := MAP('nested1' := 2)));
insert into nestedmap_json_sr (id, m) values (3, MAP('name' := MAP('nested1' := 3)));

# now run the select query and no data would be emitted
select * from nestedmap_json_sr emit changes;

**Expected behavior**
The expected behavior is data previously inserted will be emitted.

**Actual behavior**

No results is emitted from ksqldb CLI

Error messages from ksqldb server:

ksqldb-server  | [2022-03-08 04:00:13,258] ERROR {"type":0,"deserializationError":{"target":"value","errorMessage":"Failed to deserialize value from topic: nestedmap_json_sr. Can't convert type. sourceType: ArrayNode, requiredType: MAP<VARCHAR, MAP<VARCHAR, INT>>, path: $.M","recordB64":null,"cause":["Can't convert type. sourceType: ArrayNode, requiredType: MAP<VARCHAR, MAP<VARCHAR, INT>>, path: $.M","Can't convert type. sourceType: ArrayNode, requiredType: MAP<VARCHAR, MAP<VARCHAR, INT>>, path: .M","Can't convert type. sourceType: ArrayNode, requiredType: MAP<VARCHAR, MAP<VARCHAR, INT>>"],"topic":"nestedmap_json_sr"},"recordProcessingError":null,"productionError":null,"serializationError":null,"kafkaStreamsThreadError":null} (processing.transient_NESTEDMAP_JSON_SR_8793810493621047377.KsqlTopic.Source.deserializer:44)
[...]

Issue #6049, opened 2020-08-19
JSON_SR format stores maps as list of entries


big-andy-coates:
**Describe the bug**

The new(ish) `JSON_SR` format is storing `MAP` types as a list of entries, rather than a JSON Object.  The existing `JSON` format stores a `MAP` correctly as a JSON Object. The existing `Avro` format stores `MAP` correctly as an Avro Map.

**To Reproduce**
Enable these QTT test cases to `json_sr.json`.

```json
 {
      "name": "deserialize anonymous map - value",
      "statements": [
        "CREATE STREAM INPUT (foo MAP<STRING, INT>) WITH (WRAP_SINGLE_VALUE=false, kafka_topic='input_topic', value_format='JSON_SR');",
        "CREATE STREAM OUTPUT AS SELECT * FROM INPUT;"
      ],
      "inputs": [
        {"topic": "input_topic", "value": {"a": 1, "b": 2, "c": 3}},
        {"topic": "input_topic", "value": {"a": 1, "b": 2, "c": null}},
        {"topic": "input_topic", "value": null}
      ],
      "outputs": [
        {"topic": "OUTPUT", "value": {"FOO": {"a": 1, "b": 2, "c": 3}}},
        {"topic": "OUTPUT", "value": {"FOO": {"a": 1, "b": 2, "c": null}}},
        {"topic": "OUTPUT", "value": null}
      ],
      "post": {
        "sources": [
          {"name": "INPUT", "type": "stream", "schema": "FOO MAP<STRING, INT>"}
        ]
      }
    },
    {
      "name": "deserialize nested map - value",
      "statements": [
        "CREATE STREAM INPUT (foo MAP<STRING, INT>) WITH (WRAP_SINGLE_VALUE=true, kafka_topic='input_topic', value_format='JSON_SR');",
        "CREATE STREAM OUTPUT AS SELECT * FROM INPUT;"
      ],
      "inputs": [
        {"topic": "input_topic", "value": {"FOO": {"a": 1, "b": 2, "c": 3}}},
        {"topic": "input_topic", "value": {"FOO": {"a": 1, "b": 2, "c": null}}},
        {"topic": "input_topic", "value": {"FOO": null}},
        {"topic": "input_topic", "value": null}
      ],
      "outputs": [
        {"topic": "OUTPUT", "value": {"FOO": {"a": 1, "b": 2, "c": 3}}},
[...]

PR #8899, opened 2022-03-16
test: Adding test to show GitHub issue #8861

jnh5y:
### Description 
Demonstrates the test case / error from https://github.com/confluentinc/ksql/issues/8861

### DIFF
--- a/ksqldb-functional-tests/src/test/resources/query-validation-tests/json_sr.json
+++ b/ksqldb-functional-tests/src/test/resources/query-validation-tests/json_sr.json
@@ -146,6 +146,32 @@
       }
     },
     {
+      "enabled": false,
+      "comment": "disabled until fix in for https://github.com/confluentinc/ksql/issues/8861",
+      "name": "deserialize map nested in map - value",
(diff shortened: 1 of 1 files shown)


lihaosky:
Thanks! Are we going to fix https://github.com/confluentinc/ksql/issues/8861 soon to enable this test? Curious why we add a disabled test?

jnh5y:
> Thanks! Are we going to fix #8861 soon to enable this test? Curious why we add a disabled test?

I'm following the pattern where the test right next to it shows https://github.com/confluentinc/ksql/issues/6049.

I don't know if we are going to address those issues are not.  I'm just trying to triage/understand the issue and since it is related to an existing one, figured I'd toss the test in.

@colinhicks @suhas-satish any thoughts on priorities for the underlying issue?

lihaosky (approved):
I see. If we need to put the test before fixing the issue, I think we can comment or put description in those issues that we already have tests added, so whoever fixing those issues can enable the tests or avoid adding duplicate ones.

PR #8905, opened 2022-03-17
chore: rename package `physical` to `execution`

mjsax:
In order to get be separation of logical and physical query planning, we want to introduce a "proper" physical plan layer. To prepare for this change, we should rename the current package (that is not really a physical plan, but one layer below).


AlanConfluent:
Just curious to read up on this work more to understand how it will interact with other components that current exist.  Do you have a writeup on the new physical layer?

mjsax:
I don't have any writeup... :( 

Basically, we just split the current "LogicalPlan" (that is not really a logical plan, but a mix of logical and physical stuff) into two layers: a proper logical plan and a proper physical plan.

The current PhysicalPlan is not really a "physical plan" either, but it's already an "execution plan".

We currently have:
```
AST -> LogicalPlan -> PhysicalPlan
```
The new planning will be:
```
AST -> NewLogicalPlan -> NewPhysicalPlan -> ExecutionPlan(aka renamed PhysicalPlan)
```

Thus, nothing should really change for other components. We want to produce the same JSON (`ExecutionPlan`) into the command topic as before.

jnh5y (approved):

PR #8802, opened 2022-02-23
fix: update restore command topic tool to work with command topic configs

lct45:
### Description 
We recently updated the ksql command topic to make sure that it's using the command topic configs that can be passed in through the properties file, https://github.com/confluentinc/ksql/pull/8742. With these changes, we needed to update the restore command topic tool to make sure it's using the same configs as would be used when we create the command topic. This ensures that if the tool is used on a ksql that has its command topic on a separate kafka from the rest of its topics, the command topic will be properly recreated on the separate kafka.

### Testing done 
Added an integration test, the same tests as we currently do for the restore tool but separates the command topic onto a different kafka broker

### Reviewer checklist
- [ ] Ensure docs are updated if necessary. (eg. if a user visible feature is being added or changed).
- [ ] Ensure relevant issues are linked (description should include text like "Fixes #<issue number>")

PR #8908, opened 2022-03-18
chore: update print guide to remove command topic example

lct45:
### Description 
We made updates recently in https://github.com/confluentinc/ksql/pull/8802 to allow users more control over where their command topic lives. Since the command topic doesn't have to live on the same kafka that the rest of the topics do, it doesn't make sense to use the command topic as the basice example for what printing a topic would look like. This updates the print topic guide to use a user topic.

### Testing done 
Did the example in CLI

### Reviewer checklist


JimGalasyn (approved):
LGTM!

Issue #8917, opened 2022-03-22
Null struct deferencing inside a function call does not work as expected


jnh5y:
**Describe the bug**
In the Confluent Community Slack, Miguel reported that a statement like 
`SELECT IFNULL(requirements->education->degree, 'n/a') as required_edu_degree FROM 'my-stream' emit changes;`
does not produce the expected results.
(https://confluentcommunity.slack.com/archives/C6UJNMY67/p1647949968220039)

**To Reproduce**
Deference a nested struct.

The issue is that the failed null dereferencing in a function call prevents the function evaluation.  This issue is similar to https://github.com/confluentinc/ksql/issues/8324 and the fix is similar to https://github.com/confluentinc/ksql/commit/2a60269cadddc8a11522d103d800ae2a6769d690

jnh5y:
In light of the other issue, the interesting question is what other expressions should be guarded in a similar way.

alexwoolford:
I'm excited to see that you created a PR for this, @jnh5y .

A friend of mine ran into this issue:

```
ksql> CREATE STREAM ORDER_LINE_DEBEZIUM WITH (KAFKA_TOPIC='order_line_debezium', VALUE_FORMAT='AVRO');

ksql> SELECT
 BEFORE -> ID AS BEFORE_ID,
 AFTER->ID AS AFTER_ID,
 COALESCE(BEFORE->ID, AFTER->ID) AS ID
FROM ORDER_LINE_DEBEZIUM EMIT CHANGES;
+-------------------+-------------------+-------------------+
|BEFORE_ID          |AFTER_ID           |ID                 |
+-------------------+-------------------+-------------------+
|null               |13363              |null               |
|null               |13364              |null               |
|null               |13365              |null               |
```

Note that the `COALESCE` didn't work.

Without your PR, they'd need to write an SMT or Kafka Streams job. That's not ideal. They want a nice, neat ksqlDB statement that's simple to maintain and understand.

Any idea when your PR will be merged and made available in Confluent Cloud?

jnh5y:
https://github.com/confluentinc/ksql/pull/8918 should address this issue.

PR #8918, opened 2022-03-22
fix: Guard null struct dereferencing inside function calls

jnh5y:
Addresses https://github.com/confluentinc/ksql/issues/8917


jzaralim (approved):
lgtm. doesn't have to be in this PR, but we should check subscript expressions as well

jnh5y:
> lgtm. doesn't have to be in this PR, but we should check subscript expressions as well

Good call.  I just took a quick look to see what else may need this kind of fix...  I wonder if Lambda functions are complex enough that they could hit nulls? 

If not, the subscript expressions might be it?

jnh5y:
I chatted with @colinhicks about this and I'm going to see if it is easy to add a configuration to let folks continue with the existing processing.

The idea is that folks could be looking for that handling in the event that they have built processes using it.  It is somewhat unlikely, but the configuration would let users revert to the old way.

PR #8920, opened 2022-03-23
feat: add physical plan stubs for new query planner

mjsax:
The stub for the new physical plan allows basic tracking of topic names
and data formats.

We need to discuss the behavior change for `formats.json` -- cf. newly added tests.

### DIFF
--- a/ksqldb-engine/src/main/java/io/confluent/ksql/execution/ExecutionPlanner.java
+++ b/ksqldb-engine/src/main/java/io/confluent/ksql/execution/ExecutionPlanner.java
@@ -1,0 +1,100 @@
+/*
+ * Copyright 2022 Confluent Inc.
+ *
+ * Licensed under the Confluent Community License (the "License"); you may not use
+ * this file except in compliance with the License.  You may obtain a copy of the
+ * License at
+ *
+ * http://www.confluent.io/confluent-community-license
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OF ANY KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations under the License.
+ */
+
+package io.confluent.ksql.execution;
+
+import io.confluent.ksql.execution.context.QueryContext.Stacker;
+import io.confluent.ksql.execution.plan.ExecutionStep;
+import io.confluent.ksql.execution.plan.Formats;
+import io.confluent.ksql.execution.plan.KStreamHolder;
+import io.confluent.ksql.execution.plan.StreamSink;
+import io.confluent.ksql.execution.streams.ExecutionStepFactory;
+import io.confluent.ksql.metastore.MetaStore;
+import io.confluent.ksql.parser.properties.with.CreateSourceAsProperties;
+import io.confluent.ksql.parser.tree.Sink;
+import io.confluent.ksql.physicalplanner.PhysicalPlan;
+import io.confluent.ksql.physicalplanner.nodes.Node;
+import io.confluent.ksql.query.QueryId;
+import io.confluent.ksql.serde.FormatInfo;
+import io.confluent.ksql.serde.SerdeFeatures;
+import java.util.Optional;
+
+/**
+ * The {@code ExecutionPlanner} takes a {@link PhysicalPlan} and uses the visitor pattern
+ * to translate the logical plan into an execution plan.
+ *
+ * <p>A execution plan, in contrast to a physical plan, consist of concrete
+ * {@link ExecutionStep ExecutionSteps} that describe an executable Kafka Streams
+ * {@link org.apache.kafka.streams.Topology}.
+ */
+public final class ExecutionPlanner {
+
+  private ExecutionPlanner() {}
+
+  public static ExecutionPlan buildPlan(
+      final MetaStore metaStore,
+      final PhysicalPlan physicalPlan,
+      final Sink sink
+  ) {
+    final PhysicalToExecutionPlanTranslator translator =
+        new PhysicalToExecutionPlanTranslator(metaStore);
+
+    final Node<?> planRoot = physicalPlan.getRoot();
+    final ExecutionStep<?> root = translator.process(planRoot);
+
+    final FormatInfo keyFormatInfo;
+    final FormatInfo valueFormatInfo;
+    final SerdeFeatures keyFeatures;
+    final SerdeFeatures valueFeatures;
+
+    final CreateSourceAsProperties properties = sink.getProperties();
+    final Optional<String> keyFormatName = properties.getKeyFormat();
+    if (keyFormatName.isPresent()) {
+      keyFormatInfo = FormatInfo.of(keyFormatName.get());
+      keyFeatures = SerdeFeatures.of(); // to-do
+    } else {
+      final Formats formats = planRoot.getFormats();
+      keyFormatInfo = formats.getKeyFormat();
+      keyFeatures = formats.getKeyFeatures();
+    }
+
+    final Optional<String> valueFormatName = properties.getValueFormat();
+    if (valueFormatName.isPresent()) {
+      valueFormatInfo = FormatInfo.of(valueFormatName.get());
+      valueFeatures = SerdeFeatures.of(); // to-do
+    } else {
+      final Formats formats = planRoot.getFormats();
+      valueFormatInfo = formats.getValueFormat();
+      valueFeatures = formats.getValueFeatures();
+    }
+
+    final Formats outputFormat = Formats.of(
+        keyFormatInfo,
+        valueFormatInfo,
+        keyFeatures,
+        valueFeatures
+    );
+
+    final StreamSink streamSink = ExecutionStepFactory.streamSink(
+        new Stacker().push("OUTPUT"),
+        outputFormat,
+        (ExecutionStep<KStreamHolder<Object>>) root,
+        sink.getName().text(),
+        Optional.empty()// timestampColumn
+    );
+
+    return new ExecutionPlan(new QueryId("query-id"), streamSink);
+  }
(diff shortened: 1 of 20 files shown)


ksqldb-engine/src/main/java/io/confluent/ksql/execution/ExecutionPlanner.java
```diff
@@ -0,0 +1,100 @@
...
+    final FormatInfo keyFormatInfo;
+    final FormatInfo valueFormatInfo;
+    final SerdeFeatures keyFeatures;
+    final SerdeFeatures valueFeatures;
+
+    final CreateSourceAsProperties properties = sink.getProperties();
+    final Optional<String> keyFormatName = properties.getKeyFormat();
+    if (keyFormatName.isPresent()) {
+      keyFormatInfo = FormatInfo.of(keyFormatName.get());
+      keyFeatures = SerdeFeatures.of(); // to-do
```
mjsax, line 66: not sure if empty properties are the right thing here... (same for value below)

ksqldb-engine/src/main/java/io/confluent/ksql/physicalplanner/LogicalToPhysicalPlanTranslator.java
```diff
@@ -16,88 +16,82 @@
...
 import io.confluent.ksql.name.ColumnName;
-import io.confluent.ksql.schema.ksql.Column;
+import io.confluent.ksql.name.SourceName;
 import io.confluent.ksql.schema.ksql.LogicalColumn;
-import io.confluent.ksql.schema.ksql.LogicalSchema;
 import java.util.Objects;
-import java.util.Optional;
 
-public class LogicalToPhysicalPlanTranslator implements NodeVisiter<Node<?>, ExecutionStep<?>> {
+public class LogicalToPhysicalPlanTranslator
```
mjsax, line 29: Updates this to translate into the new `PhysicalPlan` (previously, this translated directly into `ExectuionStep` aka "execution plan" aka "old PhysialPlan").

ksqldb-functional-tests/src/test/resources/query-validation-tests/formats.json
```diff
@@ -271,6 +274,28 @@
...
+        "CREATE STREAM OUTPUT WITH (key_format='JSON') AS SELECT * FROM INPUT;"
+      ],
+      "properties": {
+        "ksql.new.query.planner.enabled": "true"
+      },
+      "inputs": [
+        {"topic": "input_topic", "key": "hello", "value": {"FOO": 10}}
+      ],
+      "outputs": [
+        {"topic": "OUTPUT", "key": {"K": "hello"}, "value": {"FOO": 10}}
```
mjsax, line 290: The new planner does not unwrap the key compared to the old planner. Need to discuss this. (same for other test below).

ksqldb-functional-tests/src/test/resources/query-validation-tests/select.json
```diff
@@ -3,6 +3,27 @@
     "Tests covering general SELECT clause, a.k.a projection, handling"
   ],
   "tests": [
+    {
```
mjsax, line 6: Need to revert this before merging.

ksqldb-functional-tests/src/test/resources/query-validation-tests/formats.json
```diff
@@ -291,11 +316,14 @@
...
     },
     {
-      "name": "override formats - create stream as select",
+      "name": "override formats - create stream as select (old planner)",
       "statements": [
         "CREATE STREAM INPUT (K STRING KEY, foo INT) WITH (kafka_topic='input_topic', key_format='KAFKA', value_format='AVRO');",
         "CREATE STREAM OUTPUT WITH (format='JSON') AS SELECT * FROM INPUT;"
       ],
+      "properties": {
+        "ksql.new.query.planner.enabled": "false"
```
mjsax, line 325: We added this config to immutable properties what actually disallows to set for QTT. Not sure if there is some workaround? QTT should be some kind of super-user?

ksqldb-engine/src/main/java/io/confluent/ksql/execution/PhysicalToExecutionPlanTranslator.java
```diff
@@ -0,0 +1,91 @@
...
+  PhysicalToExecutionPlanTranslator(final MetaStore metaStore) {
+    this.metaStore = Objects.requireNonNull(metaStore, "metaStore");
+  }
+
+  @Override
+  public ExecutionStep<?> process(final Node<?> node) {
+    if (node instanceof StreamSourceNode) {
+      return processStreamSourceNode((StreamSourceNode) node);
+    //} else if (node instanceof TableSourceNode) {
+    //  return processTableSourceNode((TableSourceNode) node);
```
jnh5y, line 49: Do you mean to leave these comments in?

ksqldb-engine/src/main/java/io/confluent/ksql/execution/PhysicalToExecutionPlanTranslator.java
```diff
@@ -0,0 +1,91 @@
...
+        streamSourceNode.getFormats(),
+        Optional.empty(),
+        1 // to-do
+    );
+
+  }
+
+  //private ExecutionStep<?> processTableSourceNode(final TableSourceNode tableSourceNode) {
+  //  throw new UnsupportedOperationException("not implemented yet");
+  //}
```
jnh5y, line 71: Ditto?

ksqldb-engine/src/main/java/io/confluent/ksql/physicalplanner/PhysicalPlan.java
```diff
@@ -0,0 +1,32 @@
+/*
+ * Copyright 2019 Confluent Inc.
```
jnh5y, line 2: 2022?

ksqldb-engine/src/main/java/io/confluent/ksql/physicalplanner/PhysicalPlanner.java
```diff
@@ -15,56 +15,30 @@
...
-  public static ExecutionPlan buildPhysicalPlan(
+  public static PhysicalPlan buildPlan(
       final MetaStore metaStore,
       final LogicalPlan logicalPlan
   ) {
     final LogicalToPhysicalPlanTranslator translator =
         new LogicalToPhysicalPlanTranslator(metaStore);
 
-    final ExecutionStep<?> root = translator.process(logicalPlan.getRoot());
+    final Node<?> root = translator.process(logicalPlan.getRoot());
```
jnh5y, line 40: What can the type of the Node be?  Just curious?

ksqldb-engine/src/main/java/io/confluent/ksql/physicalplanner/nodes/StreamSourceNode.java
```diff
@@ -0,0 +1,91 @@
...
+  }
+
+  private static ImmutableList<ColumnName> toColumNames(final List<Column> columns) {
+    return columns.stream()
+        .map(Column::name)
+        .collect(ImmutableList.toImmutableList());
+  }
+
+  @Override
+  public ImmutableList<ColumnName> keyColumnNames() {
```
jnh5y, line 59: Do we want to insist on `ImmutableList` instead of `List`?  I can see it either way.

ksqldb-functional-tests/src/test/resources/query-validation-tests/formats.json
jnh5y: What is behind this change?  And is it intentional or just part of the work-in-progress?

jnh5y (approved):

ksqldb-functional-tests/src/test/resources/query-validation-tests/formats.json
mjsax: Both. I wrote the code and the test failed. This is the WIP part. However, we discussed this topic in the past, and I believe the new behavior is actually better. So the question is if we want to keep the new behavior, or if we need to follow up and teach the new planner to apply the old behavior.

ksqldb-engine/src/main/java/io/confluent/ksql/execution/PhysicalToExecutionPlanTranslator.java
mjsax: Yes, it one of those WIP stubs comments that will be turn into code soon.

ksqldb-engine/src/main/java/io/confluent/ksql/physicalplanner/PhysicalPlan.java
mjsax: Good catch!

ksqldb-engine/src/main/java/io/confluent/ksql/physicalplanner/PhysicalPlanner.java
mjsax: It can be multiple things depending on the query (eg, a final projection, or a HAVING filter) but we don't really care what the type is anyway at this point in the code.

ksqldb-engine/src/main/java/io/confluent/ksql/physicalplanner/nodes/StreamSourceNode.java
mjsax: Personally, I (strong) prefer `ImmutableList` because stronger typing is always good. If we would return `List` we would need to make a deep-copy IMHO and really return a mutable-list.

ksqldb-functional-tests/src/test/resources/query-validation-tests/formats.json
jnh5y: What have folks like @colinhicks said about this change?

PR #8923, opened 2022-03-23
fix: shared runtimes calculate cache size for validation properly

wcarlson5:
### Description 
This will allow for more than 107 quereis to be made without hitting cache errors using shared runtimes. When using shared runtimes the cache can be shared among queries. When we go to see if the new query could put us over the total limit we cannot just add the query cache sizes together. Instead the queries must be grouped by runtime. Then we can see what the total usage is as the sum of the runtimes. We will assume worst case for the new query, that it will make a new runtime. This is not always the case but will prevent the runtime from being created and not having enough space. We can improve this but it would be a much more pervasive change and I think this is a strict improvement and we should get this in first regardless.

### Testing done 
added a couple of unit tests.

### Reviewer checklist

### DIFF
--- a/ksqldb-engine/src/main/java/io/confluent/ksql/query/KafkaStreamsQueryValidator.java
+++ b/ksqldb-engine/src/main/java/io/confluent/ksql/query/KafkaStreamsQueryValidator.java
@@ -26,7 +26,6 @@ import java.util.Collection;
 import java.util.HashMap;
 import java.util.HashSet;
 import java.util.Map;
-import java.util.Objects;
 import java.util.Set;
 import java.util.stream.Collectors;
 import org.apache.kafka.streams.StreamsConfig;
--- a/ksqldb-engine/src/main/java/io/confluent/ksql/query/KafkaStreamsQueryValidator.java
+++ b/ksqldb-engine/src/main/java/io/confluent/ksql/query/KafkaStreamsQueryValidator.java
@@ -15,8 +15,8 @@
 package io.confluent.ksql.query;
 
 import io.confluent.ksql.config.SessionConfig;
-import io.confluent.ksql.util.BinPackedPersistentQueryMetadataImpl;
 import io.confluent.ksql.execution.ExecutionPlan;
+import io.confluent.ksql.util.BinPackedPersistentQueryMetadataImpl;
 import io.confluent.ksql.util.KsqlConfig;
 import io.confluent.ksql.util.KsqlException;
 import io.confluent.ksql.util.PersistentQueryMetadata;
@@ -29,7 +29,6 @@ import java.util.Map;
 import java.util.Set;
 import java.util.stream.Collectors;
 import org.apache.kafka.streams.StreamsConfig;
-import scala.Int;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
--- a/ksqldb-engine/src/main/java/io/confluent/ksql/query/KafkaStreamsQueryValidator.java
+++ b/ksqldb-engine/src/main/java/io/confluent/ksql/query/KafkaStreamsQueryValidator.java
@@ -91,23 +91,25 @@ public class KafkaStreamsQueryValidator implements QueryValidator {
               .getLong(StreamsConfig.CACHE_MAX_BYTES_BUFFERING_CONFIG))
           .sum();
       final Set<String> runtimes = new HashSet<>();
-      Long maxCache = -1L;
-      for (final QueryMetadata queryMetadata: running) {
+      long cacheSizeBytesPerRuntime = -1L;
+      for (final QueryMetadata queryMetadata : running) {
         if (queryMetadata instanceof BinPackedPersistentQueryMetadataImpl) {
-          if (maxCache == -1L) {
-            maxCache = (Long) queryMetadata.getStreamsProperties()
+          if (cacheSizeBytesPerRuntime == -1L) {
+            cacheSizeBytesPerRuntime = (long) queryMetadata.getStreamsProperties()
                 .get(StreamsConfig.CACHE_MAX_BYTES_BUFFERING_CONFIG);
           }
           if (!runtimes.contains(queryMetadata.getQueryApplicationId())) {
             runtimes.add(queryMetadata.getQueryApplicationId());
-            usedByRunning += (Long) queryMetadata.getStreamsProperties()
+            usedByRunning += (long) queryMetadata.getStreamsProperties()
                 .get(StreamsConfig.CACHE_MAX_BYTES_BUFFERING_CONFIG);
           } else {
-            if (!Objects.equals(maxCache, (Long) queryMetadata.getStreamsProperties()
-                .get(StreamsConfig.CACHE_MAX_BYTES_BUFFERING_CONFIG))) {
+            if (cacheSizeBytesPerRuntime == (long) queryMetadata.getStreamsProperties()
+                .get(StreamsConfig.CACHE_MAX_BYTES_BUFFERING_CONFIG)) {
               LOG.warn("Inconsistent "
                   + StreamsConfig.CACHE_MAX_BYTES_BUFFERING_CONFIG
-                  + " in shared runtimes");
+                  + " in shared runtimes {} and {}", cacheSizeBytesPerRuntime,
+                  queryMetadata.getStreamsProperties()
+                      .get(StreamsConfig.CACHE_MAX_BYTES_BUFFERING_CONFIG));
             }
           }
         }
--- a/ksqldb-engine/src/main/java/io/confluent/ksql/query/KafkaStreamsQueryValidator.java
+++ b/ksqldb-engine/src/main/java/io/confluent/ksql/query/KafkaStreamsQueryValidator.java
@@ -15,6 +15,7 @@
 package io.confluent.ksql.query;
 
 import io.confluent.ksql.config.SessionConfig;
+import io.confluent.ksql.util.BinPackedPersistentQueryMetadataImpl;
 import io.confluent.ksql.execution.ExecutionPlan;
 import io.confluent.ksql.util.KsqlConfig;
 import io.confluent.ksql.util.KsqlException;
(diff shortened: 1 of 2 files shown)


ksqldb-engine/src/test/java/io/confluent/ksql/query/KafkaStreamsQueryValidatorTest.java
```diff
@@ -138,6 +149,38 @@ public void shouldIgnoreGlobalLimitSetInOverrides() {
     );
   }
 
+  @Test
```
guozhangwang, line 152: Do we have a test for the mix of BinPackedPersistentQueryMetadataImpl and non-binpacked metadata impl to test the summing behavior above, that when a new binpacked query is added we may violate the limit?

guozhangwang (commented):
Just a question about the tests otherwise LGTM.

ksqldb-engine/src/test/java/io/confluent/ksql/query/KafkaStreamsQueryValidatorTest.java
wcarlson5: This is already testing that actually. We have 2 old queries and 2 new qureies that are in the same runtime

ksqldb-engine/src/main/java/io/confluent/ksql/query/KafkaStreamsQueryValidator.java
```diff
@@ -69,10 +70,32 @@ private void validateCacheBytesUsage(
...
+                  r.put(
+                      e.getQueryApplicationId(),
+                      (Long) e.getStreamsProperties()
+                          .get(StreamsConfig.CACHE_MAX_BYTES_BUFFERING_CONFIG)),
+              (r, l) -> l.putAll(r))
+          .values()
+          .stream()
+          .mapToLong(t -> t)
+          .sum();
+    }
```
ableegoldman, line 98: Can we just use a single `for` loop here? This is kind of hard to follow and the double iteration is unnecessary (idk how smart the java compiler is, maybe it would optimize this but maybe not)

ableegoldman: Should we also maybe validate that every runtime has the same value? Maybe just log a warning if there's a discrepancy don't throw an error or anything, for one thing it's not that severe and for a second I wonder if there might be a race condition if the user is trying to modify the system configs at the same time

ksqldb-engine/src/main/java/io/confluent/ksql/query/KafkaStreamsQueryValidator.java
```diff
@@ -69,10 +77,41 @@ private void validateCacheBytesUsage(
...
+          if (maxCache == -1L) {
+            maxCache = (Long) queryMetadata.getStreamsProperties()
+                .get(StreamsConfig.CACHE_MAX_BYTES_BUFFERING_CONFIG);
+          }
+          if (!runtimes.contains(queryMetadata.getQueryApplicationId())) {
+            runtimes.add(queryMetadata.getQueryApplicationId());
+            usedByRunning += (Long) queryMetadata.getStreamsProperties()
+                .get(StreamsConfig.CACHE_MAX_BYTES_BUFFERING_CONFIG);
+          } else {
+            if (!Objects.equals(maxCache, (Long) queryMetadata.getStreamsProperties()
```
ableegoldman, line 105: Why use `Objects.equals` here?

ksqldb-engine/src/main/java/io/confluent/ksql/query/KafkaStreamsQueryValidator.java
```diff
@@ -69,10 +77,41 @@ private void validateCacheBytesUsage(
...
+              .getLong(StreamsConfig.CACHE_MAX_BYTES_BUFFERING_CONFIG))
+          .sum();
+    } else {
+      usedByRunning = running.stream()
+          .filter(t -> !(t instanceof BinPackedPersistentQueryMetadataImpl))
+          .mapToLong(r -> new StreamsConfig(r.getStreamsProperties())
+              .getLong(StreamsConfig.CACHE_MAX_BYTES_BUFFERING_CONFIG))
+          .sum();
+      final Set<String> runtimes = new HashSet<>();
+      Long maxCache = -1L;
```
ableegoldman, line 93: Why does this need to be a `Long` rather than a `long` / why all the casting? It kind of makes it look like there's something much more complicated going on when it's actually quite straightforward logic. Makes me feel like I'm missing something... 🤔

ableegoldman: also, super nit: `maxCache` --> `cacheSizeBytesPerRuntime` or similar

ksqldb-engine/src/main/java/io/confluent/ksql/query/KafkaStreamsQueryValidator.java
```diff
@@ -69,10 +77,41 @@ private void validateCacheBytesUsage(
...
+                .get(StreamsConfig.CACHE_MAX_BYTES_BUFFERING_CONFIG);
+          }
+          if (!runtimes.contains(queryMetadata.getQueryApplicationId())) {
+            runtimes.add(queryMetadata.getQueryApplicationId());
+            usedByRunning += (Long) queryMetadata.getStreamsProperties()
+                .get(StreamsConfig.CACHE_MAX_BYTES_BUFFERING_CONFIG);
+          } else {
+            if (!Objects.equals(maxCache, (Long) queryMetadata.getStreamsProperties()
+                .get(StreamsConfig.CACHE_MAX_BYTES_BUFFERING_CONFIG))) {
+              LOG.warn("Inconsistent "
```
ableegoldman, line 107: nit: can we log what the two inconsistent values are? And maybe the application id of the runtime that isn't matching?

ksqldb-engine/src/main/java/io/confluent/ksql/query/KafkaStreamsQueryValidator.java
```diff
@@ -69,10 +77,41 @@ private void validateCacheBytesUsage(
...
+      for (final QueryMetadata queryMetadata: running) {
+        if (queryMetadata instanceof BinPackedPersistentQueryMetadataImpl) {
+          if (maxCache == -1L) {
+            maxCache = (Long) queryMetadata.getStreamsProperties()
+                .get(StreamsConfig.CACHE_MAX_BYTES_BUFFERING_CONFIG);
+          }
+          if (!runtimes.contains(queryMetadata.getQueryApplicationId())) {
+            runtimes.add(queryMetadata.getQueryApplicationId());
+            usedByRunning += (Long) queryMetadata.getStreamsProperties()
+                .get(StreamsConfig.CACHE_MAX_BYTES_BUFFERING_CONFIG);
```
ableegoldman, line 103: nit: might be good to add some debug logging in here, just in case something wonky ever happens so we can verify what's going on inside this loop. It's probably sufficient to just log each time we find a new runtime and increase the total size rather than logging for each query, but I'll leave that up to you

ableegoldman (commented):
Left a few comments, sorry if it feels like I'm being extremely nit-picky here but I just want to make sure I'm not missing something. 

On another note, I wonder if we should synchronize this? I assume the query validation can be run in parallel with other actions, ie we could be dropping or adding queries or even adding/removing runtimes while we're looping over this...or do we synchronize higher up the callstack? @rodesai

ksqldb-engine/src/main/java/io/confluent/ksql/query/KafkaStreamsQueryValidator.java
wcarlson5: That Long stuff because I was using a map and he casting because the config map is just an object

rodesai:
> or do we synchronize higher up the callstack? @rodesai

We shouldn't need to synchronize here. The only thread that touches the main copy of this class is the command-runner thread. The API threads all create sandbox copies and work on those.

ksqldb-engine/src/main/java/io/confluent/ksql/query/KafkaStreamsQueryValidator.java
```diff
@@ -69,10 +76,43 @@ private void validateCacheBytesUsage(
...
+        if (queryMetadata instanceof BinPackedPersistentQueryMetadataImpl) {
+          if (cacheSizeBytesPerRuntime == -1L) {
+            cacheSizeBytesPerRuntime = (long) queryMetadata.getStreamsProperties()
+                .get(StreamsConfig.CACHE_MAX_BYTES_BUFFERING_CONFIG);
+          }
+          if (!runtimes.contains(queryMetadata.getQueryApplicationId())) {
+            runtimes.add(queryMetadata.getQueryApplicationId());
+            usedByRunning += (long) queryMetadata.getStreamsProperties()
+                .get(StreamsConfig.CACHE_MAX_BYTES_BUFFERING_CONFIG);
+          } else {
```
ableegoldman, line 103: nit: technically we should also check/verify that the cache size is the same across all runtimes, not just within the runtime. Since `ALTER SYSTEM` applies to all runtimes* there should never be a discrepancy in the server level cache size config

*one exception to this would be if there are gen 1 queries with different cache sizes running on an upgraded gen 2 cluster. Maybe we can distinguish these by the application id since only gen 1 queries will contain the query id in the app id? Or is there a better way to differentiate these -- eg do gen 1 queries still use `BinPackedPersistentQueryMetadata`  or are they just regular `PersistentQueryMetadata`?

ableegoldman (approved):
Pretty much looks good, one suggestion about performing the cache size consistency check for all queries but I'll leave it up to you to decide whether this seems necessary/useful

wcarlson5:
The last run was green ([9](https://jenkins.confluent.io/job/Confluent%20Public%20Repo%20PR%20builder/job/ksql/job/PR-8923/9/)) but for some reason its not showing up on the PR. I am just going to merge it as is

Issue #8757, opened 2022-02-14
Support DATE and TIMESTAMP types with MIN/MAX aggregate functions


mikebin:
It would be useful to be able to use `MIN`/`MAX` aggregate functions directly with `DATE` and `TIMESTAMP` types. Currently have to convert to numeric values with `UNIX_DATE` and `UNIX_TIMESTAMP`.

pgaref:
Addresses by https://github.com/confluentinc/ksql/pull/8924

Issue #8798, opened 2022-02-23
Support Reduce as aggregation function in CTAS


kehao95:
**Is your feature request related to a problem? Please describe.**
I was disappointed that I can't find a way to use some Scalar functions to create TABLE. For example I have a column in stream which is an ARRAY. I wish I can use something like [ARRAY_CONCAT](https://docs.ksqldb.io/en/latest/developer-guide/ksqldb-reference/scalar-functions/#array_concat) to do an aggregation to get a result ARRAY.

---

####  in the syntax of `reduce(<function>, column)`
where `<function>` should be a binary function like `fn(a,b)-> c` and `a,b,c,column` should have same type.
And we can calculate the result following this logic

```mermaid 
graph TD
note("reduce(fn,col)")
start(init statestore=null)-->in
in[/stream input: col=x/] --> init{is statestore null?}
init--Y-->p1(y=x)
init--N-->p2("y=fn(statestore,x)")
p1-->set
p2-->set
set(set statestore=y)-->out[/output y/]
out --loop--> in

set-.-|set|statestore[(statestore)]-.-|check|init
```

With this implementation all we need to track is the previous step result. But it gives us ability to convert most scala function in to an aggregation. 

### examples
```
create stream as 
(
id string,
col type
) with (..);

create table as select(
  id,
  reduce(fun,col)
)group by id
emit changes;


```

#### Example 1:  reduce add, equal to sum
Stream data(col) | statestore | `reduce(add,col)`
-- | -- | --
1 | null | 1
2 | 1 | add(1,2) = 3
3 | 3 | add(3,3) = 6
4 | 6 | add(6,4) = 10

#### Example 2: reduce ARRAY_UNION
Stream(col) | statestore | `reduce(ARRAY_UNION,col)`
-- | -- | --
[1,2,3] | null | [1,2,3]
[] | [1,2,3] | ARRAY_UNION([1,2,3],[]) = [1,2,3]
[1,4] | [1,2,3] | ARRAY_UNION([1,4],[1,2,3]) = [1,2,3,4]
[4,5,6] | [1,2,3,4] | ARRAY_UNION([4,5,6],[1,2,3,4])=[1,2,3,4,5,6]
 
## Usages
This feature will have a very wide of usage once implemented. Actually a great number of the existing aggregation can be achieved by this from.

Here are some examples, 

SUM | reduce(add, _) |  
-- | -- | --
MAX | reduce(greater, _) |  
MIN | reduce(lesser, _) |  
[...]

PR #8924, opened 2022-03-24
feat: Support MIN/MAX udafs for Time/TS/Date types

pgaref:
### Description 
Natively support MIN/MAX aggregations for DATE, TIME, and TIMESTAMP types.
All existing types implement Comparable interface so a base BaseComparableKudaf replaced BaseNumberKudaf and appropriate 2 Min/Max Kudafs were implemented (getting rid of 12+ type specific implementations).

Addresses #8757

### Testing done 
Added unit Tests for every type - Min/Max combination
Extended max-group-by min-group-by functional tests

### Reviewer checklist


ksqldb-functional-tests/src/test/resources/query-validation-tests/max-group-by.json
```diff
@@ -135,6 +135,108 @@
         {"topic": "S2", "key": 0,"value": {"VALUE": "20.99"}},
         {"topic": "S2", "key": 0,"value": {"VALUE": "20.99"}}
       ]
+    },
+    {
+      "name": "max date group by valid after 7.2.0",
```
jnh5y, line 140: We probably should add QTTs for min as well.

pgaref: Yeah, pushing an addendum as we speak

ksqldb-engine/src/main/java/io/confluent/ksql/function/udaf/BaseDateKudaf.java
```diff
@@ -0,0 +1,92 @@
...
+import java.util.function.BiFunction;
+import java.util.function.Function;
+import org.apache.kafka.streams.kstream.Merger;
+
+public class BaseDateKudaf<T extends Date> extends BaseAggregateFunction<T, T, T> {
+
+  private final BiFunction<T, T, T> aggregatePrimitive;
+
+  private final BiFunction<T, T, T> maxAggregatePrimitive = (a, b) -> {
+    if (a.after(b)) {
```
jnh5y, line 34: I think there's a possible refactoring which would let us get rid of a ton of classes here.

All of the types involved are `Comparable`.  That ordering ought to be sufficient to calculate mins/maxs.  (E.g., we don't need to use `after`.)

One option would be to replace `BaseNumberKudaf` with a `ComparableKudaf`, create concrete Min and Max versions, and then in the `M*AggFunctionFactory`, call the constructor for the concrete classes with the parameters to make it go.

The second option to consider would be trying to switch these comparable functions to the UdafFactory method approach (set `collect_list` and `latest_by_offset` for examples).

ksqldb-functional-tests/src/test/resources/query-validation-tests/max-group-by.json
jnh5y: Awesome; great work to get the unit test and QTTs!

ksqldb-engine/src/main/java/io/confluent/ksql/function/udaf/BaseDateKudaf.java
pgaref: Great ideas! Let me have another go to make this cleaner ;)

jnh5y:
Also, you'll want to add the new types to `AggregateFunctionFactory.java` line 44 or so.  That's used by the `DESCRIBE FUNCTION` command.

jnh5y:
> Also, you'll want to add the new types to `AggregateFunctionFactory.java` line 44 or so. That's used by the `DESCRIBE FUNCTION` command.

Actually, I misspoke, you'll want to update `supportedArgs()` for each of the `Min|MaxAggFunctionFactory` classes. ;)

ksqldb-engine/src/main/java/io/confluent/ksql/function/udaf/max/TimeMaxKudaf.java
```diff
@@ -0,0 +1,35 @@
...
+public class TimeMaxKudaf extends BaseComparableKudaf<Time> {
+
+  TimeMaxKudaf(
+      final String functionName,
+      final Integer argIndexInValue
+  ) {
+    super(functionName,
+        argIndexInValue,
+        SqlTypes.TIME,
+        (first, second) -> first.compareTo(second) > 0 ? first : second,
```
jnh5y, line 31: This function ought to work for all the implementations.  I think we can push it into the `BaseComparableKudaf`.

Further, I'd suggest making concrete `Min` and `Max` versions of `BaseComparableKudaf` and deleting the 14 or so {Time, Long, Double, ...}{Max|Min}Kudaf classes.

pgaref: Yup, thats a good idea! Was playing safe keeping the existing classes intact

jnh5y (approved):

ksqldb-engine/src/main/java/io/confluent/ksql/function/udaf/BaseComparableKudaf.java
```diff
@@ -26,11 +26,11 @@
 import java.util.function.Function;
 import org.apache.kafka.streams.kstream.Merger;
 
-public abstract class BaseNumberKudaf<T extends Number> extends BaseAggregateFunction<T, T, T> {
+public abstract class BaseComparableKudaf<T> extends BaseAggregateFunction<T, T, T> {
 
   private final BiFunction<T, T, T> aggregatePrimitive;
 
-  public BaseNumberKudaf(
+  public BaseComparableKudaf(
```
spena, line 33: should line 47 change to represent this UDAF accepts numbers and time/date parameters? It says "number" only.

If this "number" is still used, which means the class is for numbers and times, then should the class be renamed to something that reflects the intention of it? i.e. BaseNumberAndTimeTypesKudaf? Just an idea.

ksqldb-engine/src/main/java/io/confluent/ksql/function/udaf/max/MaxComparableKudaf.java
```diff
@@ -15,22 +15,23 @@
...
 package io.confluent.ksql.function.udaf.max;
 
-import io.confluent.ksql.function.udaf.BaseNumberKudaf;
+import io.confluent.ksql.function.udaf.BaseComparableKudaf;
 import io.confluent.ksql.schema.ksql.types.SqlType;
-import java.math.BigDecimal;
+import io.confluent.ksql.schema.utils.FormatOptions;
 
-public class DecimalMaxKudaf extends BaseNumberKudaf<BigDecimal> {
+public class MaxComparableKudaf<T extends Comparable<? super T>> extends BaseComparableKudaf<T> {
```
spena, line 22: Same comment from BaseComparableKudaf. Should it reflect the intention of the class instead? Such as MaxNumberAndTimeTypesKudaf? Same comment for the Min. What do you think?

pgaref: Hey @spena thanks for chipping in!
[Line 47 
](https://github.com/confluentinc/ksql/pull/8924/files#diff-1c7a573f0ba0b7096f72b8355fcd3f4ba8fbe669aa2f5ed7df3c608df5b34945L47) should be changed indeed but I would go for something generic like "comparable".
With the latest refactoring the only thing we actually need to support Min/Max aggregation Kudaf is a comparable Type -- currently Number and Date/Time -- but this could change in the future right?

jnh5y: Ooo.... you two are nerd sniping me this morning...  

If we are going to be precise about the names here, I think we can safely name this `MaxKudaf`.  The name of the other class is completely separately.  I think its signature just gives one what you'd want for a Fold/Reduce.  (E.g., see this issue: https://github.com/confluentinc/ksql/issues/8798)

Ok, so how can I justify dropping `Comparable` from the name? 

As a brief argument, (I think) it is easy to see that a Java `Comparable` type has a total ordering, and as such, it makes sense to talk about `Max` and `Min` values.  

Going the other way, if one has a `Max` (or a `Min`) function, then one could use it pairwise to write an implementation of `Comparable`.  (It'd look something like: {If `a==b`, return 0; else, if Max(a,b)==a, return 1, else return -1} (Up to a sign:)).)

At that point, the only thing which'd go wrong is someone having a type which is "comparable" which does not implement `Comparable`.  We'd have to have user defined types and then folks would have to work around that by adding a `Comparable` wrapper or they'd have to implement a UDAF that handles their "comparable" interface.  That'd be pretty rare...  

So where does that leave us?  I'm kinda ok with the name as-is (it doesn't hurt anything) or a smaller name (`MaxKudaf`).

pgaref: Thanks for the input here!
Introduced type `T` bounds and opted for simpler Min/Max Kudaf names in https://github.com/confluentinc/ksql/commit/2cb7546fc07f9285e3089fb916ad52b020332c92

jnh5y: Awesome.  @spena any more thoughts?

ksqldb-engine/src/main/java/io/confluent/ksql/function/udaf/BaseComparableKudaf.java
```diff
@@ -44,7 +45,7 @@ public BaseNumberKudaf(
         outputSchema,
         Collections.singletonList(
             new ParameterInfo(
-                "number",
+                "comparable",
```
spena, line 48: This word still makes some noise. 

This is what `describe function max;` displays when `number` is used:
```
ksql> describe function max;

Name        : MAX
Author      : Confluent
Type        : AGGREGATE
Jar         : internal
Variations  : 

	Variation   : MAX(number INT)
	Returns     : INT
	Description : Computes the maximum integer value for a key.
	number      : the value to aggregate

	Variation   : MAX(number BIGINT)
	Returns     : BIGINT
	Description : Computes the maximum long value for a key.
	number      : the value to aggregate
        
       ...
```

It now displays `comparable`:
```
ksql> describe function max;

Name        : MAX
Author      : Confluent
Type        : AGGREGATE
Jar         : internal
Variations  : 

	Variation   : MAX(comparable INT)
	Returns     : INT
	Description : Computes the maximum INTEGER value for a key.
	comparable  : the value to aggregate

	Variation   : MAX(comparable BIGINT)
	Returns     : BIGINT
	Description : Computes the maximum BIGINT value for a key.
	comparable  : the value to aggregate

       ...
```

The `comparable` should be something else that reflects what the parameter is used in those functions, i.e. `number`, `date`, `time`, etc.

jnh5y: Would "value" be a suitable replacement for "comparable"?  Worst case, things could be reworked so that each of the constructors could pass in a different word there.  E.g., "number" for Int,BigInt,etc, and date/time for those new types.

spena (approved):
LGTM

Issue #8528, opened 2021-12-16
Foreign key join with Schema Registry enabled key format on the right-side table may miss results


vcrfxia:
**Describe the bug**

When performing a foreign key join, if the table on the right side (the primary key table) uses a Schema-Registry-enabled key format (e.g., Avro or Protobuf) with a custom schema, i.e., not a schema produced by ksqlDB, then the foreign key join may miss results. This can happen because when the foreign key is extracted from the left table, it is serialized using ksql's schema, which may have a different Schema Registry schema ID than the custom schema used for the right table's key. If this is the case, then the serialized bytes will not match even if the values are logically the same, causing missed join results.

**To Reproduce**

On any version of ksqlDB since foreign key joins were introduced, issue an FK join with custom Avro schema for the right table's key, and data to match on the left side.

**Expected behavior**

Join results are present.

**Actual behaviour**

Join results may be missing.

**Additional context**

We don't have the analogous problem with SR-enabled key formats for non-FK-joins since ksql has logic to detect the join key format and automatically repartition to ensure matching schema IDs if an SR-enabled key format is detected. We also have test coverage to ensure this.

For FK joins, we don’t have any logic to determine key formats since we don’t perform any copartitioning checks for FK joins, since FK join sources don’t need to be copartitioned. When introducing the feature, we missed the edge case where the right key format is SR-enabled and a non-ksql schema is being used. We have [FK join test coverage](https://github.com/confluentinc/ksql/blob/97fa117ec33d42e145518568cf4a441f0bd8cb33/ksqldb-functional-tests/src/test/resources/query-validation-tests/fk-join.json#L639) for Avro keys on the left source only, which is why this edge case was missed. 

To fix this bug, ksql should handle the case where an SR-enabled key format is used on the RHS of an FK join by automatically repartitioning to use the ksql schema, as we do for non-FK-joins.

mjsax:
Kafka Streams by default uses the right table key-serializer to write into the subscription topic, and the left table key-serializer to write into the response topic. This ensure co-partitioning of the subscription data with the right table, and co-partitioning of the response data with the left table.

If AVRO is used, we need to ensure that the schema-id use is the same (ie, when the right table key schema is re-registered for the subscription topic it might use the same schema-id as the right table key). Similar for the response topic.

If the schema-id would change, co-partitioning might break.

Gerrrr:
As a status update, I found 2 ways to make make foreign key joins miss entries.

[...]

PR #8926, opened 2022-03-24
fix: Repartition RHS of a FK join if it uses SR schema

Gerrrr:
### Description 

This PR addresses the original issue identified in https://github.com/confluentinc/ksql/issues/8528. The compound primary key bug is addressed in https://issues.apache.org/jira/browse/KAFKA-13769.

#### Root Cause Analysis

FK join can miss results in 2 cases. First, if the RHS of the join uses a custom SR schema, it might differ from the schema produced by ksqlDB. Consequently, keys in the source topic can be partitioned differently from the keys in, say, Kafka Streams `ForeignJoinSubscriptionSendProcessorSupplier`. 

The second issue happens the table on LHS uses a compound primary key. Such a key is internally represented as a `STRUCT`, where the table column names are the `STRUCT` keys. ksqlDB uses `ConnectFormat` serializer that turns `STRUCT` into a JSON string. For example, a row in a table with a primary key with 2 columns `A` and `B` can look like `{"A": 1, "B" 2}`. FK-join pre-processes both sides of the FK-join by prefixing column names with the corresponding table names to avoid duplicates in the result schema. This column renaming affects the serialization of the `STRUCT`. Consider the example above. With the renamed columns, the same row will be serialized as `{"TABLENAME_A": 1, "TABLENAME_B" 2}`. The difference in serialization in the source and the FK processing nodes leads to miss-partitioning.

## Solution

The first issue is solved in this PR by repartitioning the data source on the RHS if the schema is powered by SR. This way, the FK join always deals with the schema created by ksqlDB. 

The second issue is solved in Kafka Streams by explicitly specifying the target partition in `FK response-sink`. See https://github.com/apache/kafka/pull/11945 for more details.

### Testing done 

QTT test

### Reviewer checklist


Gerrrr:
@mjsax Can you please review?

ksqldb-engine/src/main/java/io/confluent/ksql/planner/LogicalPlanner.java
```diff
@@ -1092,6 +1106,48 @@ private LogicalSchema buildRepartitionedSchema(
     );
   }
 
+  /**
+   * Given a source node and its side, return {@code true} if the source should be repartitioned
+   * for the foreign key join.
+   *
+   * <p>The right hand side of the FK join must be repartitioned if it uses schema format baked by
+   * Schema Registry and the primary key is a struct. In that case, there is a possibility that
```
mjsax, line 1114: > and the primary key is a struct

The PK should only be a struct for multi-column keys? Hence, such a a join should not be supported? Seems I am missing something? -- For what case we this actually happen?

ksqldb-engine/src/main/java/io/confluent/ksql/planner/LogicalPlanner.java
```diff
@@ -1092,6 +1106,48 @@ private LogicalSchema buildRepartitionedSchema(
...
+    final boolean isJsonSR = dataSourceNode.getDataSource()
+        .getKsqlTopic()
+        .getKeyFormat()
+        .getFormat()
+        .equals(JsonSchemaFormat.NAME);
+    final boolean isSREnabledFormat = isAvro || isProtobuf || isJsonSR;
+    final boolean hasStructInPrimaryKey = dataSourceNode.getSchema()
+        .key()
+        .stream()
+        .anyMatch(col -> col.type() instanceof SqlStruct);
```
mjsax, line 1147: `SqlStruct` is really `STRUCT`, right? IIRC we only support Joins on primitive types though?

ksqldb-functional-tests/src/test/resources/query-validation-tests/fk-join.json
```diff
@@ -595,6 +595,56 @@
         }
       }
     },
+    {
+      "name": "Should repartition RHS if the key format comes from Schema Registry",
+      "statements": [
+        "CREATE TABLE left_table (l_id BIGINT PRIMARY KEY, name VARCHAR, foreign_key STRUCT<val BIGINT>) WITH (kafka_topic='left_topic', format='AVRO', partitions=128);",
+        "CREATE TABLE right_table (r_id STRUCT<val BIGINT> PRIMARY KEY, f1 INTEGER) WITH (kafka_topic='right_topic', format='AVRO', partitions=128);",
```
mjsax, line 602: I am surprise that this actually works? I was expecting `ON foreign_key = r_id` to fail as `STRUCT` should not be supported?

ksqldb-engine/src/main/java/io/confluent/ksql/planner/LogicalPlanner.java
Gerrrr: This can happen if we define the PK type as struct explicitly, see the added qtt test:

https://github.com/confluentinc/ksql/blob/aeae363c4c2d265241aa30583e121be01a81d5c0/ksqldb-functional-tests/src/test/resources/query-validation-tests/fk-join.json#L599-L604

ksqldb-engine/src/main/java/io/confluent/ksql/planner/LogicalPlanner.java
Gerrrr: Yeah, `SqlStruct` is the implementation of `SqlType` that maps to connect's `Struct`. See the converter:

https://github.com/confluentinc/ksql/blob/aeae363c4c2d265241aa30583e121be01a81d5c0/ksqldb-engine-common/src/main/java/io/confluent/ksql/schema/ksql/DefaultSqlValueCoercer.java#L245-L276

I didn't find any restrictions on FK or regular joins that prevent using non-primitive types.

mjsax (approved):

mjsax:
Seems Jenkins did not pass: https://jenkins.confluent.io/job/Confluent%20Public%20Repo%20PR%20builder/job/ksql/job/PR-8926/ -- can you take a look?

Gerrrr:
I forgot to generate qtt tests; added them in [8fc5deb](https://github.com/confluentinc/ksql/pull/8926/commits/8fc5debd678e35970a1ce882dffd2f3ab6bc55fc). [The last CI run](https://jenkins.confluent.io/job/Confluent%20Public%20Repo%20PR%20builder/job/ksql/job/PR-8926/4/) is green, so I will merge the PR.

PR #8927, opened 2022-03-24
chore: better error message on ClusterAuthorizationException during INSERT VALUES

vcrfxia:
### Description 

AK recently enabled idempotent producers by default: https://issues.apache.org/jira/browse/KAFKA-13598. This changes the exception thrown when a producer does not have write permissions from TopicAuthorizationException (for the non-idempotent producer case) to ClusterAuthorizationException when the idempotent producer fails to initialize a transaction (because either write permissions are missing or, for broker versions no later than 2.8, because IDEMPOTENT_WRITE permissions on the cluster are missing). 

This PR updates ksql to account for the ClusterAuthorizationException case to throw a better error message when INSERT VALUES fails, similar to the special handling that exists today for TopicAuthorizationException.

### Testing done 

Added unit test.

### Reviewer checklist

### DIFF
--- a/ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/execution/InsertValuesExecutor.java
+++ b/ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/execution/InsertValuesExecutor.java
@@ -55,6 +55,7 @@ import io.confluent.ksql.util.KsqlException;
 import io.confluent.ksql.util.KsqlStatementException;
 import io.confluent.ksql.util.ReservedInternalTopics;
 import java.time.Duration;
+import java.util.Collections;
 import java.util.List;
 import java.util.Map;
 import java.util.Objects;
@@ -70,6 +71,7 @@ import org.apache.kafka.clients.producer.Producer;
 import org.apache.kafka.clients.producer.ProducerRecord;
 import org.apache.kafka.clients.producer.RecordMetadata;
 import org.apache.kafka.common.acl.AclOperation;
+import org.apache.kafka.common.errors.ClusterAuthorizationException;
 import org.apache.kafka.common.errors.TopicAuthorizationException;
 import org.apache.kafka.common.serialization.Serde;
 import org.slf4j.Logger;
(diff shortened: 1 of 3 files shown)


ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/execution/InsertValuesExecutor.java
```diff
@@ -167,6 +169,24 @@ public void execute(
...
+      // distinguish this type of failure from other permissions exceptions
+      // such as the ones thrown above when TopicAuthorizationException is caught.
+      final Exception rootCause = new KsqlTopicAuthorizationException(
+          AclOperation.WRITE,
+          Collections.singletonList(dataSource.getKafkaTopicName()),
+          // Ideally we would forward e.getMessage() instead of the hard-coded
+          // message below, but until this error message is improved on the Kafka
+          // side, e.getMessage() is not helpful. (Today it is just "Cluster
+          // authorization failed.")
+          "The producer is not authorized to do idempotent sends"
```
spena, line 187: Should we include how to disable idempotent? or should we include a server config that disables idempotent for producers so users don't need to do it every time they run INSERT VALUES?

vcrfxia: I'm not convinced it's necessary. The only cases in which a user would hit this is if one of the following is true:
1. They do not have write permissions to the topic.
2. They have write permissions, but they are using a broker version 2.8 or older and do not have IDEMPOTENT_WRITE permissions.
3. They are using a broker version 0.11 or older.

If it's the first, then disabling idempotent writes isn't going to help. The third should be extremely rare at this point, and we don't want to be in the business of supporting that. So that just leaves the second as a reason in which disabling the idempotent write would help. If you think it's helpful, we can add additional context here such as "The producer is not authorized to do idempotent sends. Check that you have write permissions to the specified topic, and disable idempotent sends by setting `enable.idempotent=false` if necessary."

vcrfxia: Updated the error message to the example in my comment above. Lmk what you think!

spena (approved):
+1

PR #8928, opened 2022-03-24
feat: limit the number of runtimes that can be created

wcarlson5:
### Description 
When a query that is issued that would cause a runtime to exceed the limit that query will fail.

### Testing done 
add a unit test in the runtime assignor

### Reviewer checklist


guozhangwang (approved):

wcarlson5:
I don't think we are going to do this anymore

PR #8932, opened 2022-03-25
fix: only make shared runtime that is chosen

wcarlson5:
### Description 
only create relevant runtimes for validation. Because we are now selecting runtimes in the assignor I think we don't need to make every runtime anymore. This should really speed up validation, especially when we have a topic that is queried a lot

### Testing done 

### Reviewer checklist


wcarlson5:
@rodesai We talked about making all the runtimes for validation. Now I think we only need to make the one that is being used now that the runtime assignor makes the decision and puts it in the cmd topic. Is that correct?

ableegoldman (approved):
Ah nice, glad we remembered to do this

Issue #5265, opened 2020-05-04
KSQL creates streams with wrong Protobuf schema if the Protobuf definition has multiple messages 


lang1991:
**Describe the bug**
When we create a Kafka stream in ksql-cli from a Protobuf definition in the schema registry, the schema definition of the first message in the ".proto" is used, instead of the one that actually matches the content of the stream. 

Given the following Protobuf definition:
```
syntax = "proto2";

package me.kafkaplayground.service;

option java_package = "com.me.protos.kafkaplayground.service";
option java_generic_services = true;

message LogLine {
  optional string ip = 1;
  optional uint64 timestamp = 2;
  optional string url = 3;
  optional string referer = 4;
  optional string useragent = 5;
  optional User user = 6;
}

message User {
  optional string first_name = 1;
  optional string last_name = 2;
  optional int32 age = 3;
}

message DummyC {
  enum StateC {
    STATE_C_DO_NOT_USE = 0;
    ACTIVE = 1;
    INACTIVE = 2;
  }

  optional string c_token = 1;

  optional string u_token = 2;

  optional StateC state = 3;

  repeated Event events = 4;

  oneof p_data {
    BcData bb = 5;
  }

  message Event {
    optional int64 id = 1;

    oneof product_data {
      BceData b = 5;
    }

    message BceData {
      optional CState c_state = 1;
    }
  }

  message BcData {
    oneof ch_id {
      string u_token = 1;
    }

    optional string bc_token = 2;
  }
}

enum CState {
  ACTIVE = 1;
  DISABLED = 2;
  HAHA = 3;
  HEHE = 4;
  XIXI = 5;
  HUHU = 6;
  XDXD = 7;
  QQ = 8;
  EE = 9;
  A = 10;
  B = 11;
  C = 12;
  D = 13;
  E = 14;
  F = 15;
}
```
Assuming we are sending instances of DummyC through a Kafka topic "dummyC", and we create a Kafka stream in ksql-cli with the following command:
`create stream dummyC with (KAFKA_TOPIC='dummyC', VALUE_FORMAT='PROTOBUF');`
The schema of stream "dummyC" will be "Logline" in this case, because it is the first "message" defined in the ".proto" file. If we swap the order to make "DummyC" the first message in this ".proto", the stream will be created with the correct schema.

**To Reproduce**
Steps to reproduce the behavior, include:
1. The version of KSQL. **5.5.0-post**
2. Sample source data. Please refer to the Protobuf definition above
[...]

Issue #6091, opened 2020-08-25
New schema version registered on INSERT VALUES


mikebin:
**Describe the bug**
When using Avro with Schema Registry in ksqlDB, a schema is registered for the `CREATE STREAM` statement (if it doesn't already exist). Then when inserting records with `INSERT VALUES`, a second version of the schema (with one minor metadata difference - no `connect.name` attribute) is subsequently registered. The second version seems unnecessary.

**To Reproduce**
Version: 0.11.0

Create a new Avro-based stream:
```sql
create stream test (a varchar ) with ( kafka_topic='test', value_avro_schema_full_name='test.name', partitions='1', value_format='avro');
```

View the registered schema:
```
curl -s http://localhost:8081/subjects/test-value/versions/1 | jq -r '.schema' | jq
{
  "type": "record",
  "name": "name",
  "namespace": "test",
  "fields": [
    {
      "name": "A",
      "type": [
        "null",
        "string"
      ],
      "default": null
    }
  ],
  "connect.name": "test.name"
}
```

Insert a record:
```sql
insert into test (a) values ('hi');
```

View the new schema version:
```
curl -s http://localhost:8081/subjects/test-value/versions/2 | jq -r '.schema' | jq
{
  "type": "record",
  "name": "name",
  "namespace": "test",
  "fields": [
    {
      "name": "A",
      "type": [
        "null",
        "string"
      ],
      "default": null
    }
  ]
}
```

Note that the only difference in the second schema version is no `connect.name` metadata attribute.

**Expected behavior**
Only one schema version is created when the stream is defined

**Actual behaviour**
See above

agavra:
**tl;dr** This is a real bug, but I'm going down downgrade it from `p0` because it's mostly cosmetic.

---

To add some background on the issue - we are eagerly registering schemas when we create streams/tables that don't have any to make sure that the compatibility will be maintained going forward. The default behavior for connect-based serializers is to register the schema when data is produced.

Unfortunately, it seems that the serializers are using a different mechanism for building the schema (namely, we don't construct the schema using the basic `AvroConverter` and instead we use the `AvroSchemas` class to generate schemas).

TBH, I'm not entirely sure why we don't just use the vanilla converter (as we do for both JSON_SR and PROTOBUF formats). The javadoc mentions:

```
/**
 * Translates KSQL data and schemas to Avro equivalents.
 *
 * <p>Responsible for converting the KSQL schema to a version ready for connect to convert to an
 * avro schema.
[...]

PR #8933, opened 2022-03-25
fix: Create stream fails when multiple Protobuf schema definitions exist

spena:
### Description 

Fixes https://github.com/confluentinc/ksql/issues/5265

Ignore 1st commit https://github.com/confluentinc/ksql/pull/8933/commits/58e470336cc3a583872cf33f46557a1fd65acb41 which is being reviewed here -> https://github.com/confluentinc/ksql/pull/8984

This PR supports the `KEY_SCHEMA_FULL_NAME` and `VALUE_SCHEMA_FULL_NAME` properties in the `WITH` clause when creating a stream. These properties allows users to specify the full name of a protobuf schema instead of using a default name (current default is `ConnectDefault1`).

When registering a new schema, if the properties are given, then the schema will be registered in SR with the new name, i.e.
```
ksql> create stream t1(id int key, c1 double, c2 int) with (kafka_topic='t1', format='protobuf', key_schema_full_name='ProtobufKey1', value_schema_full_name='ProtobufValue1', partitions=1);

 Message        
----------------
 Stream created 
----------------

ksql> describe t1;

Name                 : T1
 Field | Type                   
--------------------------------
 ID    | INTEGER          (key) 
 C1    | DOUBLE                 
 C2    | INTEGER                
--------------------------------

$ curl -X GET http://localhost:8081/subjects/t1-key/versions/1/schema
syntax = "proto3";

message ProtobufKey1 {
  int32 ID = 1;
}

$ curl -X GET http://localhost:8081/subjects/t1-value/versions/1/schema
syntax = "proto3";

message ProtobufValue1 {
  double C1 = 1;
  int32 C2 = 2;
}
```

The KEY/VALUE properties can also be used to specify the name of the Protobuf message schema if multiple schema definitions are available in SR. i.e.
```
$ curl -X GET http://localhost:8081/subjects/s1-key/versions/2/schema
syntax = "proto3";

message ProtobufKey1 {
  int32 ID = 1;
}
message ProtobufKey2 {
  string ID = 1;
}

$ curl -X GET http://localhost:8081/subjects/s1-value/versions/2/schema
syntax = "proto3";

message ProtobufValue1 {
  double C1 = 1;
  int32 C2 = 2;
}
message ProtobufValue2 {
  int32 C1 = 1;
}

ksql> create stream s1 with (kafka_topic='s1', format='protobuf', key_schema_full_name='ProtobufKey2', value_schema_full_name='ProtobufValue2', partitions=1);

 Message        
----------------
 Stream created 
----------------

ksql> describe s1;

Name                 : S1
 Field | Type                   
--------------------------------
 ID    | VARCHAR(STRING)  (key) 
 C1    | INTEGER                
--------------------------------

ksql> drop stream s1;

ksql> create stream s1 with (kafka_topic='s1', format='protobuf', key_schema_full_name='ProtobufKey1', value_schema_full_name='ProtobufValue1', partitions=1);

 Message        
----------------
 Stream created 
----------------

ksql> describe s1;

Name                 : S1
 Field | Type                   
--------------------------------
 ID    | INTEGER          (key) 
 C1    | DOUBLE                 
 C2    | INTEGER                
--------------------------------
```

The KEY/VALUE properties are already supported for AVRO. This PR assumes those properties were meant for other formats too, but when Protobuf was implemented then these props weren't fixed.

### Testing done 
Manually. See the above examples.
I have some limitations with QTT that does not allow me to test the KEY/VALUE properties. 
Reason:
```
I can't make the multiple protobuf schema work because QTT first prepares all input/output topics and schema before executing statements, which sets up Protobuf translators without the properties specified in the CREATE statements
```

### Reviewer checklist

### DIFF
--- a/ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/execution/InsertValuesExecutor.java
+++ b/ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/execution/InsertValuesExecutor.java
@@ -501,6 +501,15 @@ public class InsertValuesExecutor {
       }
     }
 
+    // The SCHEMA_ID is used by the serializer to fetch the right SR schema by id
+    if (supportedProperties.contains(ConnectProperties.SCHEMA_ID)) {
+      if (!formatInfo.getProperties().containsKey(ConnectProperties.SCHEMA_ID)) {
+        SchemaRegistryUtil.getSchemaId(srClient, topicName, isKey)
+            .ifPresent(schemaId ->
+                propertiesBuilder.put(ConnectProperties.SCHEMA_ID, String.valueOf(schemaId)));
+      }
+    }
+
     return FormatInfo.of(formatInfo.getFormat(), propertiesBuilder.build());
   }
 
--- a/ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/execution/InsertValuesExecutor.java
+++ b/ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/execution/InsertValuesExecutor.java
@@ -16,6 +16,7 @@
 package io.confluent.ksql.rest.server.execution;
 
 import com.google.common.annotations.VisibleForTesting;
+import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableMap;
 import io.confluent.connect.avro.AvroDataConfig;
 import io.confluent.kafka.schemaregistry.ParsedSchema;
@@ -529,9 +530,21 @@ public class InsertValuesExecutor {
     // are available.
     if (supportedProperties.contains(ConnectProperties.SCHEMA_ID)) {
       if (!formatInfo.getProperties().containsKey(ConnectProperties.SCHEMA_ID)) {
-        SchemaRegistryUtil.getLatestSchemaId(srClient, topicName, isKey)
-            .ifPresent(schemaId ->
-                propertiesBuilder.put(ConnectProperties.SCHEMA_ID, String.valueOf(schemaId)));
+        // Inject the SCHEMA_ID only if the format to serialize is Protobuf and if the SR schema
+        // has multiple schema definitions. This ensures that the serializer does not attempt to
+        // register a new schema during the object serialization.
+        if (format instanceof ProtobufFormat) {
+          final List<String> schemaNames =
+              SchemaRegistryUtil.getLatestParsedSchema(srClient, topicName, isKey)
+                  .map(format::schemaFullNames)
+                  .orElse(ImmutableList.of());
+
+          if (schemaNames.size() > 1) {
+            SchemaRegistryUtil.getLatestSchemaId(srClient, topicName, isKey)
+                .ifPresent(schemaId ->
+                    propertiesBuilder.put(ConnectProperties.SCHEMA_ID, String.valueOf(schemaId)));
+          }
+        }
       }
     }
 
--- a/ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/execution/InsertValuesExecutor.java
+++ b/ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/execution/InsertValuesExecutor.java
@@ -19,7 +19,6 @@ import com.google.common.annotations.VisibleForTesting;
 import com.google.common.collect.ImmutableMap;
 import io.confluent.connect.avro.AvroDataConfig;
 import io.confluent.kafka.schemaregistry.ParsedSchema;
-import io.confluent.kafka.schemaregistry.client.SchemaMetadata;
 import io.confluent.kafka.schemaregistry.client.SchemaRegistryClient;
 import io.confluent.kafka.schemaregistry.client.rest.exceptions.RestClientException;
 import io.confluent.ksql.GenericKey;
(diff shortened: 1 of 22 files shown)


pgaref:
Thanks for this @spena! A bit more general question: multiple schema definitions can exist and users can specify what they want to pick -- however I noticed that multiple schema versions can also exist? Are we assuming users always pick the latest?

spena:
@pgaref Yes, multiple schema versions can exist. There is another property to select the schema version or `SCHEMA_ID`. This property and the one for this PR may be combined, i.e.
```
create stream s1 with (schema_id='5', value_schema_full_name='Proto1');
```
Notice that I am not using a version in `schema_id`. Each new schema registered in SR has a schema ID, so we support ID only.

colinhicks:
@spena does the latest commit for insert values also cover this bug? https://github.com/confluentinc/ksql/issues/6091

ksqldb-serde/src/main/java/io/confluent/ksql/serde/protobuf/ProtobufFormat.java
```diff
@@ -80,4 +85,18 @@ protected ConnectSchemaTranslator getConnectSchemaTranslator(
     return new ProtobufSerdeFactory(new ProtobufProperties(formatProps))
         .createSerde(connectSchema, config, srFactory, targetType, isKey);
   }
+
+  @Override
+  public List<String> getSchemaDefinitions(final ParsedSchema schema) {
```
lihaosky, line 90: Can you add some test for this?

ksqldb-serde/src/main/java/io/confluent/ksql/serde/protobuf/ProtobufSchemaTranslator.java
```diff
@@ -64,13 +71,24 @@ public String name() {
...
+
+  private ProtobufSchema withSchemaFullName(final ProtobufSchema origSchema) {
+    return fullNameSchema.map(origSchema::copy).orElse(origSchema);
+  }
+
+  private Schema injectSchemaFullName(final Schema origSchema) {
+    return fullNameSchema
+        .map(fullName -> SchemaFullNameAppender.appendSchemaFullName(origSchema, fullName))
+        .orElse(origSchema);
   }
```
lihaosky, line 93: Can you also add some tests for these?

lihaosky (commented):
Aweseome! LGTM overall. Some small nits.

ksqldb-serde/src/main/java/io/confluent/ksql/serde/Format.java
```diff
@@ -166,4 +167,11 @@ default void validateProperties(final Map<String, String> properties) {
    * @return true if the given sql type is supported
    */
   boolean supportsKeyType(SqlType type);
+
+  /**
+   * Returns a list of schema names found in the {@code ParsedSchema}.
+   */
+  default List<String> getSchemaDefinitions(final ParsedSchema schema) {
```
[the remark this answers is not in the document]
spena: True. I just added the method to `ConnectFormat` and return a single name as default there. I still don't know how Json and Avro have multiple definitions, so they will return one single name. For others formats, such as delimited, it still returns Zero (the default from Format.java). I also changed the method name to `schemaFullNames`

ksqldb-serde/src/main/java/io/confluent/ksql/serde/protobuf/ProtobufFormat.java
spena: Added tests

ksqldb-serde/src/main/java/io/confluent/ksql/serde/protobuf/ProtobufSchemaTranslator.java
spena: Added tests

ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/execution/InsertValuesExecutor.java
```diff
@@ -435,6 +462,57 @@ private byte[] serializeValue(
...
+    }
+
+    // The SCHEMA_ID is used by the serializer to fetch the right SR schema by id
+    if (supportedProperties.contains(ConnectProperties.SCHEMA_ID)) {
+      if (!formatInfo.getProperties().containsKey(ConnectProperties.SCHEMA_ID)) {
+        SchemaRegistryUtil.getSchemaId(srClient, topicName, isKey)
+            .ifPresent(schemaId ->
+                propertiesBuilder.put(ConnectProperties.SCHEMA_ID, String.valueOf(schemaId)));
+      }
+    }
```
lihaosky, line 511: Why do we need this? This seems to change existing logic of data serialization drastically

spena: Because we need the Serde factory to get the Schema and extract the specified schema definition from SR. This happens in `ProtobufSerdeFactory.createSerde()` which calls `SerdeUtils.getAndTranslateSchema()` when the schemaId is present. 

Also, by setting the SCHEMA_ID, the ProtobufSerdeFactory.getConverter() disables the AUTO_REGISTER_SCHEMAS
to prevent failing because multiple schema definitions on SR.

The same happens for Avro and JsonSR factories. I could refactor the interfaces and classes to accept the physical schema extracted, which is what I need. But seems schemaId does the same, so I reuse it.

What other things does the schemaId do that it's for concern?

lihaosky (approved):
Discussed offline. We will only inject schema id during insertion time when it's protobuf and has multiple messages. 

One caveat is that if schema in SR has multiple messages, `CREATE or replace` with more columns and then `insertion` will always fail because old schema in SR has less columns and we don't do autoregistration to SR. 

Approve the PR to unblock but @spena will merge after fixing the schema id injection logic and do more testing :)

PR #8934, opened 2022-03-25
fix: preserve old schema behavior for protobuf wrapped primitives

vcrfxia:
### Description 

This PR is stacked on https://github.com/confluentinc/ksql/pull/8931 -- only the commits starting with `fix: preserve old schema behavior for protobuf wrapped primitives` need to be reviewed.

Prior to ksqlDB 0.24.0 and CP 7.1.0 (i.e., ksqlDB 0.23.1 and CP 7.0.1 are known good versions), the protobuf converter used by ksql translated protobuf wrapped types such as `google.protobuf.StringValue` into a struct with a single, primitive-type field named `value`. However, https://github.com/confluentinc/schema-registry/pull/2102 changed the default behavior of the protobuf converter used by ksql to instead translate such types into raw primitive types instead. This resulted in backwards incompatibility for existing ksql queries with such schemas. 

This patch fixes the backwards incompatibility by adding a new format property (`unwrapPrimitives`), specific to the protobuf format, which will configure the converter's behavior. This property is set for new queries (and new queries only) by a new statement injector (`SourcePropertyInjector`). 

This solution works for queries issued on any ksql version besides ksqlDB 0.24.0 and CP 7.1.0 -- these two versions are known to misbehave with protobuf wrapped types and should be skipped by such users.

This patch needs to be backported to 7.0.x and 7.1.x, since the relevant schema-registry change is on those branches as well.

### Testing done 

QTT with old and new behavior + unit tests + manual.

### Reviewer checklist

### DIFF
--- a/ksqldb-engine/src/main/java/io/confluent/ksql/statement/SourcePropertyInjector.java
+++ b/ksqldb-engine/src/main/java/io/confluent/ksql/statement/SourcePropertyInjector.java
@@ -1,0 +1,123 @@
+/*
+ * Copyright 2022 Confluent Inc.
+ *
+ * Licensed under the Confluent Community License; you may not use this file
+ * except in compliance with the License.  You may obtain a copy of the License at
+ *
+ * http://www.confluent.io/confluent-community-license
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OF ANY KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations under the License.
+ */
+
+package io.confluent.ksql.statement;
+
+import io.confluent.ksql.parser.KsqlParser.PreparedStatement;
+import io.confluent.ksql.parser.SqlFormatter;
+import io.confluent.ksql.parser.properties.with.CreateSourceAsProperties;
+import io.confluent.ksql.parser.properties.with.CreateSourceProperties;
+import io.confluent.ksql.parser.tree.CreateAsSelect;
+import io.confluent.ksql.parser.tree.CreateSource;
+import io.confluent.ksql.parser.tree.Statement;
+import io.confluent.ksql.util.ErrorMessageUtil;
+import io.confluent.ksql.util.KsqlException;
+import io.confluent.ksql.util.KsqlStatementException;
+
+/**
+ * An injector which injects information into {@code CreateSourceProperties}
+ * or {@code CreateSourceAsProperties}.
+ *
+ * <p>The only information injected currently is that new queries using the
+ * protobuf format should use the new default converter behavior to unwrap
+ * wrapped primitives. This injector causes new queries to use the new behavior
+ * while old (existing) queries will continue to use the old behavior.
+ *
+ * <p>If a statement that is not {@code CreateAsSelect} or {@code CreateSource}
+ * is passed in, this results in a no-op that returns the incoming statement.</p>
+ */
+public class SourcePropertyInjector implements Injector {
+
+  @SuppressWarnings("unchecked")
+  public <T extends Statement> ConfiguredStatement<T> inject(
+      final ConfiguredStatement<T> statement
+  ) {
+    if (!(statement.getStatement() instanceof CreateSource)
+        && !(statement.getStatement() instanceof CreateAsSelect)) {
+      return statement;
+    }
+
+    try {
+      if (statement.getStatement() instanceof CreateAsSelect) {
+        return (ConfiguredStatement<T>) injectForCreateAsSelect(
+            (ConfiguredStatement<? extends CreateAsSelect>) statement);
+      } else {
+        return (ConfiguredStatement<T>) injectForCreateSource(
+            (ConfiguredStatement<? extends CreateSource>) statement);
+      }
+    } catch (final KsqlStatementException e) {
+      throw e;
+    } catch (final KsqlException e) {
+      throw new KsqlStatementException(
+          ErrorMessageUtil.buildErrorMessage(e),
+          statement.getStatementText(),
+          e.getCause());
+    }
+  }
+
+  private ConfiguredStatement<? extends CreateSource> injectForCreateSource(
+      final ConfiguredStatement<? extends CreateSource> original
+  ) {
+    final CreateSource statement = original.getStatement();
+    final CreateSourceProperties properties = statement.getProperties();
+
+    final CreateSourceProperties injectedProps = properties.withUnwrapProtobufPrimitives(true);
+
+    return buildConfiguredStatement(original, injectedProps);
+  }
+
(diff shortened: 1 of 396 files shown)


ksqldb-engine/src/main/java/io/confluent/ksql/statement/SourcePropertyInjector.java
```diff
@@ -0,0 +1,123 @@
...
+    }
+  }
+
+  private ConfiguredStatement<? extends CreateSource> injectForCreateSource(
+      final ConfiguredStatement<? extends CreateSource> original
+  ) {
+    final CreateSource statement = original.getStatement();
+    final CreateSourceProperties properties = statement.getProperties();
+
+    final CreateSourceProperties injectedProps = properties.withUnwrapProtobufPrimitives(true);
```
lihaosky, line 75: I thought we need use KsqlConfig setting to avoid making changes in QTT to distinguish old/new QTT tests? But I didn't see where you made QTT changes to distinguish those. Did you manage to make QTT work by just adding the injector?

lihaosky (commented):
Thanks @vcrfxia ! LGTM overall.

ksqldb-engine/src/test/java/io/confluent/ksql/statement/SourcePropertyInjectorTest.java
```diff
@@ -0,0 +1,109 @@
...
+ * Licensed under the Confluent Community License (the "License"); you may not use
+ * this file except in compliance with the License.  You may obtain a copy of the
+ * License at
+ *
+ * http://www.confluent.io/confluent-community-license
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OF ANY KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations under the License.
```
vcrfxia, line 13: I had to add this since otherwise findbugs was complaining about the `Mockito.verify()` lines in the two tests. I have no idea why this was happening -- AFAICT our other Mockito tests don't hit this issue when using `verify()`.

ksqldb-serde/src/test/java/io/confluent/ksql/serde/protobuf/ProtobufSchemaTranslatorTest.java
```diff
@@ -0,0 +1,121 @@
...
+  public void shouldWrapPrimitives() {
+    // Given:
+    givenWrapPrimitives();
+
+    // When:
+    final Schema schema = schemaTranslator.toConnectSchema(SCHEMA_WITH_WRAPPED_PRIMITIVES);
+
+    // Then:
+    assertThat(schema.field("c1").schema().type(), is(Type.STRUCT));
+    assertThat(schema.field("c1").schema().field("value").schema().type(), is(Type.BOOLEAN));
```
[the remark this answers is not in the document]
vcrfxia: This is what happens for old queries, when the new protobuf format property is not set, yes.

ksqldb-engine/src/main/java/io/confluent/ksql/statement/SourcePropertyInjector.java
vcrfxia: That's correct. Turns out the place in QTT that I was worried about being shared between new test execution and historic test execution was not shared after all. That was my bad. (Good news is I didn't have to add the config anymore. Bad news is that means there are two completely different codepaths in QTT for executing new code versus old code...)

ksqldb-engine/src/test/java/io/confluent/ksql/statement/SourcePropertyInjectorTest.java
```diff
@@ -0,0 +1,109 @@
...
+  private SourcePropertyInjector injector;
+
+  @Before
+  public void setUp() {
+    when(csStatement.getStatement()).thenReturn(createSource);
+    when(csStatement.getSessionConfig()).thenReturn(sessionConfig);
+    when(createSource.getProperties()).thenReturn(originalCSProps);
+    when(createSource.getElements()).thenReturn(tableElements);
+    when(originalCSProps.withUnwrapProtobufPrimitives(true)).thenReturn(csPropsWithUnwrapping);
+    when(createSource.copyWith(tableElements, csPropsWithUnwrapping)).thenReturn(csWithUnwrapping);
```
lihaosky, line 76: Maybe because nothing is mocked for `csWithUnwrapping`? You don't need to mock `csWithUnwrapping`?

ksqldb-engine/src/test/java/io/confluent/ksql/statement/SourcePropertyInjectorTest.java
```diff
@@ -0,0 +1,109 @@
...
+
+    when(csasStatement.getStatement()).thenReturn(createAsSelect);
+    when(csasStatement.getSessionConfig()).thenReturn(sessionConfig);
+    when(createAsSelect.getProperties()).thenReturn(originalCsasProps);
+    when(originalCsasProps.withUnwrapProtobufPrimitives(true)).thenReturn(csasPropsWithUnwrapping);
+    when(createAsSelect.copyWith(csasPropsWithUnwrapping)).thenReturn(csasWithUnwrapping);
```
lihaosky, line 82: ditto

lihaosky (approved):
Thanks for taking effort to cleanup the code and make it work in QTT. LGTM and good to go after tests pass and figuring out `RV_RETURN_VALUE_IGNORED_INFERRED`.

clap

ksqldb-engine/src/test/java/io/confluent/ksql/statement/SourcePropertyInjectorTest.java
vcrfxia: Hm, I don't think that's the problem. It's true that I don't need to mock `csWithUnwrapping` but `createSource.copyWith()` needs to be mocked and needs to return something, so I think this is a valid usage of a mock (otherwise I'd have to create a new `CreateSource` -- I guess it could be a stub but that should be the same thing from the perspective of findbugs).

Findbugs was specifically pointing to the lines `verify(originalCSProps).withUnwrapProtobufPrimitives(true);` and `verify(originalCsasProps).withUnwrapProtobufPrimitives(true);` as the offending lines, which is weird to me since it's valid Mockito usage to not care about the return value from `verify()`.

I spent some more time looking at this and couldn't figure it out so I ended up just removing the offending lines. The test coverage is still the same since the return value of `csWithUnwrapping` cannot be reached without going through the line I wanted to verify. (I only wanted to include the extra `verify()` check to make extra explicit for readers which method in the chain of method calls is the relevant one for the test.)

vcrfxia:
PR builder passed the ksql build but then failed on "downstream validation" on a repo that this PR doesn't impact -- the build failure for that repo is a version resolution error that is unrelated to this PR. Merging.

PR #8935, opened 2022-03-26
feat: clean up extra runtimes so the assignment is more focused

wcarlson5:
### Description 
When we remove the last query from a runtime we clean it up. However we don't clean up the assignor so a new query can be assigned again. This make the assignment spread out over all the runtimes ever created instead of as few as possible.

### Testing done 
added a unit test

### Reviewer checklist

### DIFF
--- a/ksqldb-engine/src/main/java/io/confluent/ksql/engine/RuntimeAssignor.java
+++ b/ksqldb-engine/src/main/java/io/confluent/ksql/engine/RuntimeAssignor.java
@@ -102,6 +102,8 @@ public class RuntimeAssignor {
       if (runtimesToSources.get(queryMetadata.getQueryApplicationId()).isEmpty()
           && runtimesToSources.size() > numDefaultRuntimes) {
         runtimesToSources.remove(queryMetadata.getQueryApplicationId());
+        log.info("Removing runtime {} form selection of possible runtimes",
+            queryMetadata.getQueryApplicationId());
       }
     }
   }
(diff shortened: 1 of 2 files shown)


ksqldb-engine/src/main/java/io/confluent/ksql/engine/RuntimeAssignor.java
```diff
@@ -97,6 +99,10 @@ public void dropQuery(final PersistentQueryMetadata queryMetadata) {
       runtimesToSources.get(queryMetadata.getQueryApplicationId())
           .removeAll(queryMetadata.getSourceNames());
       idToRuntime.remove(queryMetadata.getQueryId());
+      if (runtimesToSources.get(queryMetadata.getQueryApplicationId()).isEmpty()
+          && runtimesToSources.size() >= runtimes) {
```
ableegoldman, line 103:
```suggestion
          && runtimesToSources.size() > runtimes) {
```

ksqldb-engine/src/main/java/io/confluent/ksql/engine/RuntimeAssignor.java
```diff
@@ -37,18 +37,20 @@
   private final Map<String, Collection<SourceName>> runtimesToSources;
   private final Map<QueryId, String> idToRuntime;
   private final Logger log = LoggerFactory.getLogger(RuntimeAssignor.class);
+  private final int runtimes;
```
ableegoldman, line 40: nit: rename to `numDefaultRuntimes` or similar

ableegoldman (commented):
Needs to be `>` instead of `>=`, I think? Also the variable name is too generic right now, could you just give it a more meaningful name when you go to fix the condition

Otherwise looks good, nice find

ableegoldman (approved):
LGTM

PR #8936, opened 2022-03-26
fix: clean up state when restarting runtimes to avoid errors

wcarlson5:
### Description 
When running `Alter System` with many queries running we were encountering errors like 
```Stack trace: org.apache.kafka.streams.errors.StreamsException: Unable to initialize state, this can happen if multiple instances of Kafka Streams are running in the same state directory (current state directory is [/mnt/data/data/ksql-state/_confluent-ksql-pksqlc-8xo8rquery_0]
	at org.apache.kafka.streams.processor.internals.StateDirectory.initializeProcessId(StateDirectory.java:191)
	at org.apache.kafka.streams.KafkaStreams.<init>(KafkaStreams.java:891)
	at org.apache.kafka.streams.KafkaStreams.<init>(KafkaStreams.java:874)
	at org.apache.kafka.streams.processor.internals.namedtopology.KafkaStreamsNamedTopologyWrapper.<init>(KafkaStreamsNamedTopologyWrapper.java:87)```

the clean up of state from the old runtime is not ideal but for now that might be necessary
 
### Testing done 

### Reviewer checklist


ableegoldman (approved):
LGTM...weird that we didn't catch this until now? I guess maybe when Tolga tested the `ALTER SYSTEM` feature it was without any running queries?

PR #8894, opened 2022-03-15
feat: allow STREAMS with no key

mjsax:
### Description 
Previously, ksqlDB requires a CSAS statement to include key columns in
the SELECT. There is no semantic reason for this restriction and this
PR removes the key-column-requirement for CSAS statements.

### Testing done 
_Describe the testing strategy. Unit and integration tests are expected for any behavior changes._

### Reviewer checklist
- [ ] Ensure docs are updated if necessary. (eg. if a user visible feature is being added or changed).
- [ ] Ensure relevant issues are linked (description should include text like "Fixes #<issue number>")

PR #8949, opened 2022-03-29
feat: allow STREAMS with no key

mjsax:
Second try for #8894


ksqldb-rest-app/src/test/resources/ksql-plan-schema/schema.json
```diff
@@ -483,6 +483,12 @@
             "type" : "string"
           }
         },
+        "selectedKeys" : {
+          "type" : "array",
+          "items" : {
+            "type" : "string"
+          }
+        },
```
mjsax, line 491: Needed to re-generate the schema because we added the new optional field `selectedKeys` -- should be a backward compatible change so it seem ok.

jnh5y (approved):

mjsax:
Rebased this PR to get the QTT changes and added new QTTs for this new feature.

ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/restore/KsqlRestoreCommandTopic.java
```diff
@@ -400,7 +400,7 @@ private static void maybeCleanUpQuery(final byte[] command, final KsqlConfig ksq
     boolean sharedRuntimeQuery = false;
     String queryId = "";
     final JSONObject jsonObject = new JSONObject(new String(command, StandardCharsets.UTF_8));
-    if (hasKey(jsonObject, "plan")) {
+    if (hasKey(jsonObject, "plan") && !jsonObject.isNull("plan")) {
```
mjsax, line 403: @vcrfxia @jnh5y -- build was failing due to the change in `PlanJsonMapper.java`: 

https://github.com/confluentinc/ksql/pull/8949/files#diff-df73b0e9550f5c1f929e6d5758135b5947bbbf44af70ac27bac08603091d0806L47-R49

The test indeed called System.exit that is why the Jenkins output was hard to read... The plan can now be `null` so we need to guard for this condition.

PR #8950, opened 2022-03-30
test: extend QTT to support parameterized configs

mjsax:
To run QTTs with old and new optimizer, we want to re-run test. The easiest way is to allow QTT to re-run tests based on the feature flag.

### DIFF
--- a/ksqldb-common/src/main/java/io/confluent/ksql/config/ImmutableProperties.java
+++ b/ksqldb-common/src/main/java/io/confluent/ksql/config/ImmutableProperties.java
@@ -36,7 +36,7 @@ public final class ImmutableProperties {
       .add(KsqlConfig.KSQL_READONLY_TOPICS_CONFIG)
       .add(KsqlConfig.KSQL_SOURCE_TABLE_MATERIALIZATION_ENABLED)
       .add(KsqlConfig.KSQL_HEADERS_COLUMNS_ENABLED)
-      .add(KsqlConfig.KSQL_NEW_QUERY_PLANNER_ENABLED)
+      //.add(KsqlConfig.KSQL_NEW_QUERY_PLANNER_ENABLED) -- to-do: protect in release
       .addAll(KsqlConfig.SSL_CONFIG_NAMES)
       .build();
 
--- a/ksqldb-functional-tests/src/main/java/io/confluent/ksql/test/tools/TestCaseBuilder.java
+++ b/ksqldb-functional-tests/src/main/java/io/confluent/ksql/test/tools/TestCaseBuilder.java
@@ -62,17 +62,23 @@ public final class TestCaseBuilder {
           ? Collections.singletonList(Optional.empty())
           : test.formats().stream().map(Optional::of).collect(Collectors.toList());
 
-      final List<Optional<String>> configs = test.config().isEmpty()
-          ? Collections.singletonList(Optional.empty())
-          : test.config().stream().map(Optional::of).collect(Collectors.toList());
+      final List<Optional<String>> configs;
+      final Entry<String, Object> overwrite;
 
-      final Entry<String, Object> overwrite = Iterables.getOnlyElement(
-          test.properties().entrySet().stream()
-              .filter(e -> {
-                final Object v = e.getValue();
-                return v instanceof String && "{CONFIG}".equalsIgnoreCase((String) v);
-              })
-              .collect(Collectors.toList()));
+      if (test.config().isEmpty()) {
+        configs = Collections.singletonList(Optional.empty());
+        overwrite = null;
+      } else {
+        configs= test.config().stream().map(Optional::of).collect(Collectors.toList());
+
+        overwrite = Iterables.getOnlyElement(
+            test.properties().entrySet().stream()
+                .filter(e -> {
+                  final Object v = e.getValue();
+                  return v instanceof String && "{CONFIG}".equalsIgnoreCase((String) v);
+                })
+                .collect(Collectors.toList()));
+      }
 
       for (final Optional<String> format : formats) {
         for (final Optional<String> config : configs) {
--- a/ksqldb-functional-tests/src/main/java/io/confluent/ksql/test/tools/TestCaseBuilder.java
+++ b/ksqldb-functional-tests/src/main/java/io/confluent/ksql/test/tools/TestCaseBuilder.java
@@ -16,6 +16,7 @@
 package io.confluent.ksql.test.tools;
 
 import com.google.common.collect.ImmutableList;
+import com.google.common.collect.Iterables;
 import io.confluent.ksql.function.InternalFunctionRegistry;
 import io.confluent.ksql.test.model.ExpectedExceptionNode;
 import io.confluent.ksql.test.model.PostConditionsNode;
@@ -26,12 +27,15 @@ import io.confluent.ksql.test.model.TopicNode;
 import io.confluent.ksql.test.tools.conditions.PostConditions;
 import io.confluent.ksql.util.KsqlConfig;
 import java.nio.file.Path;
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.LinkedList;
 import java.util.List;
 import java.util.Map;
+import java.util.Map.Entry;
 import java.util.Optional;
 import java.util.function.Function;
 import java.util.stream.Collectors;
-import java.util.stream.Stream;
 import org.hamcrest.Matcher;
 
 /**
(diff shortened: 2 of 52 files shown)


ksqldb-functional-tests/src/main/java/io/confluent/ksql/test/tools/TestCaseBuilder.java
```diff
@@ -52,15 +56,49 @@ private TestCaseBuilder() {
...
+
+      for (final Optional<String> format : formats) {
+        for (final Optional<String> config : configs) {
+          final TestLocation location = testLocator.apply(test.name());
+
+          final Map<String, Object> updatedProperties = new HashMap<>(test.properties());
+          final Optional<String> cfg;
+          if (config.isPresent()) {
+
+            final Entry<String, Object> overwrite = Iterables.getOnlyElement(
```
pgaref, line 77: Looks like the `overwrite key` can be outside the loop and `updatedProperties` actually updated on every conf iteration

ksqldb-common/src/main/java/io/confluent/ksql/config/ImmutableProperties.java
```diff
@@ -36,7 +36,7 @@
       .add(KsqlConfig.KSQL_READONLY_TOPICS_CONFIG)
       .add(KsqlConfig.KSQL_SOURCE_TABLE_MATERIALIZATION_ENABLED)
       .add(KsqlConfig.KSQL_HEADERS_COLUMNS_ENABLED)
-      .add(KsqlConfig.KSQL_NEW_QUERY_PLANNER_ENABLED)
+      //.add(KsqlConfig.KSQL_NEW_QUERY_PLANNER_ENABLED) -- to-do: protect in release
```
pgaref, line 39: Whats the plan protecting the conf here?

ksqldb-functional-tests/src/main/java/io/confluent/ksql/test/tools/TestCaseBuilderUtil.java
```diff
@@ -72,15 +72,20 @@ private TestCaseBuilderUtil() {
...
     final String pf = explicitFormat
         .map(f -> " - " + f)
         .orElse("");
 
-    return prefix + testName + pf;
+    final String pc = config
+        .map(f -> " - " + f)
+        .orElse("");
+
+    return prefix + testName + pf + pc;
```
jnh5y, line 88: Does this impact historical names of tests?  (E.g., is there any backwards compatibility that come up with the names in any way?)

FWIW, if the build passes, I'm totally happy.;)

jnh5y (approved):
LGTM!

As a note, none of the tests show that the config can vary while the format varies.  The code makes sense, so that ought to work just fine.

ksqldb-common/src/main/java/io/confluent/ksql/config/ImmutableProperties.java
mjsax: As long as the new planner is not completed, we want to disable it and don't allow users to change the config to enable it. -- However, for QTT, we need to allow to change the config so we need to lift the immutability protection for now (I guess the problem is, that we will also need to remove the corresponding test in the release branch when we re-enable the protection..)

Not sure if there is a better way? -- Ideally, we would change QTT that it is not bound to this protection, but not sure how complex this would be and/or if it's worth it?

ksqldb-functional-tests/src/main/java/io/confluent/ksql/test/tools/TestCaseBuilder.java
mjsax: Good catch. `overwrite` can be outside.

And yes, `updatedProperties` should be updated in each iteration, as we would set a different config each time.

ksqldb-functional-tests/src/main/java/io/confluent/ksql/test/tools/TestCaseBuilderUtil.java
mjsax: Not 100% sure... I hope @vcrfxia can chime in?

My understanding is, that it should be ok, as we run all historic tests no matter their name.

ksqldb-common/src/main/java/io/confluent/ksql/config/ImmutableProperties.java
pgaref: > I guess the problem is, that we will also need to remove the corresponding test in the release branch when we re-enable the protection
I am just assuming that we will always have to validate old-plans and make sure we are compatible. So even though the property should not be changed at runtime we need a way to test both.

An idea would be to trigger a QTT run for KSQL_NEW_QUERY_PLANNER_ENABLED=true tests and one run for false ?

ksqldb-functional-tests/src/main/java/io/confluent/ksql/test/tools/TestCaseBuilder.java
```diff
@@ -52,15 +56,54 @@ private TestCaseBuilder() {
...
+      final Entry<String, Object> overwrite;
+
+      if (test.config().isEmpty()) {
+        configs = Collections.singletonList(Optional.empty());
+        overwrite = null;
+      } else {
+        configs = test.config().stream().map(Optional::of).collect(Collectors.toList());
+
+        overwrite = Iterables.getOnlyElement(
+            test.properties().entrySet().stream()
```
vcrfxia, line 75: This assumes that the "config" field can only be used to configure properties, right? I don't love that this addition to the QTT framework feels very targeted / non-extensible, but if it's only meant as a short-term addition (with plans to be removed once it's served it's purpose), then this might be OK.

ksqldb-functional-tests/src/main/java/io/confluent/ksql/test/tools/TestCaseBuilderUtil.java
vcrfxia: It doesn't matter for the historic ones, yeah. It might matter for the latest ones but if that's the case then PlannedTestsUpToDateTest would fail the build and we'd be alerted to it.

ksqldb-functional-tests/src/test/resources/query-validation-tests/formats.json
```diff
@@ -2,6 +2,10 @@
   "tests": [
     {
       "name": "invalid value format - create source",
+      "config": ["true", "false"],
```
vcrfxia, line 5: Is this new "config" field meant for general purpose use? If so, we should add it to the docs: https://github.com/confluentinc/ksql/blob/master/ksqldb-functional-tests/README.md

vcrfxia (commented):
Hey @mjsax -- I'm fine with this as a short-term solution but I have concerns about general extensibility. Is the intention that others can use the new "config" field for their use cases too, or is this a temporary approach you're planning to back out after the query planner rewrite is done?

It feels weird to me that the new "config" field only allows configuring a single property; to generalize this we'd want test writers to be able to name configs and substitute specified values for specific config names. We'd also want to document the new "config" field and its usage, so others will know how to use it without needing to read the code.

If we make these changes, then I'm fine with this as a long-term framework improvement as well.

ksqldb-functional-tests/src/main/java/io/confluent/ksql/test/tools/TestCaseBuilderUtil.java
mjsax: `PlannedTestsUpToDateTest` did fail and I needed to generate new plans for the newly named tests.

ksqldb-functional-tests/src/main/java/io/confluent/ksql/test/tools/TestCaseBuilder.java
mjsax: > This assumes that the "config" field can only be used to configure properties, right?

Yes.

> I don't love that this addition to the QTT framework feels very targeted / non-extensible, but if it's only meant as a short-term addition (with plans to be removed once it's served it's purpose), then this might be OK.

Yeah, it's more or less a short-term hack. We can either remove it again, or make it more generic if we want. Given that we did discuss to maybe replace QTT, I did not want to go "all in" and make it super generic.

ksqldb-functional-tests/src/main/java/io/confluent/ksql/test/tools/TestCaseBuilderUtil.java
mjsax: Haha -- would have been smart to add the newly generated plan to the commit and lot have the only locally :D

jnh5y: Ah-ha!  🤔   Are you really trying to rename all the unit tests to juice your GitHub additions/subtractions?

"Look, I did 10 million lines of code this year...."

mjsax: Busted! 😂

PR #8964, opened 2022-03-31
chore: disable downstream validation for PR builds

vcrfxia:
### Description 

### Testing done 

### Reviewer checklist

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -2,6 +2,7 @@
 
 common {
     slackChannel = '#ksqldb-quality-oncall'
+    timeoutHours = 4
     upstreamProjects = 'confluentinc/schema-registry'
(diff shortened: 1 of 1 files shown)


nateab (approved):

Jenkinsfile
```diff
@@ -2,12 +2,14 @@
 
 common {
     slackChannel = '#ksqldb-quality-oncall'
+    timeoutHours = 4
```
vcrfxia, line 5: Setting `downStreamValidate = false` drops the default build timeout (from common) from 3 hours to 1 hour which is too short. We should manually set it back to either 3 hours or 4 hours. Going with 4 hours here since we've seen some timeouts at 3 hours lately.

PR #8969, opened 2022-04-01
test: avoid calling System.exit in tests

mjsax:
If we call System.exit in tests, it can crash the Jenkins job and it's
hard to reason about the root cause. This PR introduces an new `System`
interface that we can use to mock calls to `System.exit` in tests.


pgaref (approved):
Just wondering if we can somehow enforce this -- LGTM otherwise

vvcephei:
Hey @mjsax , it looks like shadowing System has some other consequences: https://github.com/confluentinc/ksql/pull/8972/commits/65bb6c83d2e67d216e2f658828767e52b46545b5

PR #8973, opened 2022-04-04
fix: wait longer while waiting for expected spq

nateab:
this fixed another flakey spq integration test so going to try to add it here as well in case it helps.

### DIFF
--- a/ksqldb-rest-app/src/test/java/io/confluent/ksql/api/integration/ScalablePushBandwidthThrottleIntegrationTest.java
+++ b/ksqldb-rest-app/src/test/java/io/confluent/ksql/api/integration/ScalablePushBandwidthThrottleIntegrationTest.java
@@ -273,7 +273,7 @@ public class ScalablePushBandwidthThrottleIntegrationTest {
   private void assertExpectedScalablePushQueries(
       final int expectedScalablePushQueries
   ) {
-    assertThatEventually(() -> {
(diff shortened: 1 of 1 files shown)


ksqldb-rest-app/src/test/java/io/confluent/ksql/api/integration/ScalablePushBandwidthThrottleIntegrationTest.java
```diff
@@ -280,6 +280,11 @@ private void assertExpectedScalablePushQueries(
             || !metadata.getScalablePushRegistry().get().latestHasAssignment()) {
           return false;
         }
+        try {
```
agavra, line 283: assertThatEventually has a backoff built into it. You can configure it with any of these options:
```
  public static <T> T assertThatEventually(
      final Supplier<String> message,
      final Supplier<? extends T> actualSupplier,
      final Matcher<? super T> expected,
      final long timeout,
      final TimeUnit unit,
      final long initialPausePeriodMs,
      final long maxPausePeriodMs,
      final ExceptionHandling exceptionHandling
  ) {
```

nateab: added a longer period pause

PR #8975, opened 2022-04-04
chore: upgrade vertx to 4.2

agavra:
### Description 

Upgrading to Vert.X 4.2. A lot of it was trial and error trying to figure it out, but the guide is here: https://vert-x3.github.io/vertx-4-migration-guide/index.html

I'll comment more inline about why certain things changed

### Testing done 

Ran all the tests - I'll test basic authentication E2E as well after publishing the PR

### Reviewer checklist


ksqldb-engine/src/main/java/io/confluent/ksql/execution/scalablepush/PushRouting.java
```diff
@@ -291,8 +291,8 @@ public void preparePushQuery(
       final String thisHostName
   ) {
     if (node.isLocal()) {
-      LOG.info("Query {} id {} executed locally at host {} at timestamp {}.",
-          statement.getStatementText(), pushPhysicalPlanManager.getQueryId(), node.location(),
+      LOG.info("Query with id {} executed locally at host {} at timestamp {}.",
```
agavra, line 294: this was logging queries at INFO level which is a security violation - it also was logging ridiculously long strings of `aaaaaaaaaa` in the tests and clobbering my terminal so I was happy to get rid of it

ksqldb-rest-app/src/main/java/io/confluent/ksql/api/auth/JaasAuthProvider.java
```diff
@@ -44,40 +42,34 @@
...
   ) {
     this.server = Objects.requireNonNull(server, "server");
-    this.config = Objects.requireNonNull(config, "config");
-    this.loginContextSupplier =
-        Objects.requireNonNull(loginContextSupplier, "loginContextSupplier");
-    final List<String> authRoles = config.getList(KsqlRestConfig.AUTHENTICATION_ROLES_CONFIG);
-    this.allowedRoles = authRoles.stream()
-        .filter(role -> !"*".equals(role)) // remove "*"
-        .map(role -> "**".equals(role) ? "*" : role) // Change "**" to "*"
-        .collect(Collectors.toList());
```
agavra, line 73: this was moved to a dedicated class because authorization and authentication were split into two in VertX 4 (see https://vert-x3.github.io/vertx-4-migration-guide/index.html#authentication-and-authorization_{context})

ksqldb-rest-app/src/main/java/io/confluent/ksql/api/server/AuthHandlers.java
```diff
@@ -22,27 +22,31 @@
...
-import io.vertx.ext.web.handler.AuthHandler;
 import io.vertx.ext.web.handler.BasicAuthHandler;
 import java.net.URI;
 import java.util.Optional;
 
 public final class AuthHandlers {
 
+  private static final String PROVIDER_KEY = "provider";
+  private enum Provider {
+    SYSTEM, JAAS, PLUGIN
```
agavra, line 41: we select one of these methods of AuthN/Z - we use this enum to communicate across threads which ones we chose (it uses the context local data)

this allows us to login using JAAS separately from authorization, but still use the correct authZ mechanism

ksqldb-rest-app/src/test/java/io/confluent/ksql/api/CorsTest.java
```diff
@@ -251,7 +251,7 @@ private void shouldAcceptPreflightRequest(final String origin, final String allo
         MultiMap.caseInsensitiveMultiMap().add("origin", origin)
             .add(ACCESS_CONTROL_REQUEST_METHOD_HEADER, "POST"));
 
-    assertThat(response.statusCode(), is(200));
+    assertThat(response.statusCode(), is(204));
```
agavra, line 254: I couldn't track down what made this change, but it seems like a refinement on 200 so figured it's OK

ksqldb-rest-app/src/test/java/io/confluent/ksql/api/integration/ScalablePushBandwidthThrottleIntegrationTest.java
```diff
@@ -183,7 +184,13 @@ public void scalablePushBandwidthThrottleTestHTTP2()
     // header, row
     assertThatEventually(() ->  subscriber.getRows().size(), is (2));
     subscriber.close();
-    publisher.close();
+
+    // http2 connections are reused, meaning if we close the request
```
agavra, line 188: this used to work because `close` was synchronous

pom.xml
```diff
@@ -145,15 +145,15 @@
         <apache.io.version>2.7</apache.io.version>
         <io.confluent.ksql.version>7.3.0-0</io.confluent.ksql.version>
         <io.confluent.schema-registry.version>${confluent.version.range}</io.confluent.schema-registry.version>
-        <netty-tcnative-version>2.0.46.Final</netty-tcnative-version>
+        <netty-tcnative-version>2.0.51.Final</netty-tcnative-version>
```
agavra, line 148: we'll need to bump these downstream as well, but this is the netty version that works with this version of vertx

ksqldb-api-client/src/main/java/io/confluent/ksql/api/client/impl/QueryResponseHandler.java
```diff
@@ -15,8 +15,7 @@
 
 package io.confluent.ksql.api.client.impl;
 
-import com.fasterxml.jackson.databind.ObjectMapper;
-import io.confluent.ksql.api.client.util.JsonMapper;
+import io.confluent.ksql.rest.ApiJsonMapper;
```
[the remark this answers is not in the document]
jnh5y: I'm guessing that the answer is "yes", but do we still need the Jackson databind dependency?

ksqldb-api-client/src/main/java/io/confluent/ksql/api/client/util/JsonMapper.java
```diff
@@ -1,46 +0,0 @@
...
-
-package io.confluent.ksql.api.client.util;
-
-import com.fasterxml.jackson.core.JsonGenerator;
-import com.fasterxml.jackson.databind.DeserializationFeature;
-import com.fasterxml.jackson.databind.ObjectMapper;
-import com.fasterxml.jackson.databind.node.JsonNodeFactory;
-import com.fasterxml.jackson.datatype.jdk8.Jdk8Module;
-
-public final class JsonMapper {
```
jnh5y, line 24: Nice clean up!

ksqldb-api-client/src/test/java/io/confluent/ksql/api/client/integration/PushV2ClientContinueIntegrationTest.java
```diff
@@ -380,11 +383,31 @@ private static void addObjectToKsqlArray(final KsqlArray array, final Object val
...
   }
 
+  /**
+   * VertX 4 changed to serialize using Base64 without padding but our server
+   * still encodes the data with padding - this uses the same serializer that
+   * the server uses to serialize the data into a string
+   *
+   * @see <a href=https://github.com/eclipse-vertx/vert.x/pull/3197>vertx#3197</a>
+   */
+  private static String serializeVertX3CompatibleByte(final byte[] bytes) {
```
jnh5y, line 400: Can we refactor this to a common place?  (Not a blocking question.)

jnh5y: Does `-Dvertx.json.base64=legacy` help us at any point?  (I'm blindly cribbing from the docs...)

ksqldb-common/src/main/java/io/confluent/ksql/reactive/BasePublisher.java
```diff
@@ -65,8 +66,9 @@ public void subscribe(final Subscriber<? super T> subscriber) {
     }
   }
 
-  public void close() {
+  public Future<Void> close() {
     ctx.runOnContext(v -> doClose());
+    return Future.succeededFuture();
```
jnh5y, line 71: Did the signature for `Publisher` change?  

Can line 70 fail in any interesting way?

ksqldb-common/src/main/java/io/confluent/ksql/util/VertxUtils.java
```diff
@@ -40,7 +41,20 @@ public static void checkContext(final Context context) {
...
 
   public static boolean isEventLoopAndSameContext(final Context context) {
-    return Context.isOnEventLoopThread() && context == Vertx.currentContext();
+    return Context.isOnEventLoopThread()
+        && (context == Vertx.currentContext()
+        || checkDuplicateContext(Vertx.currentContext(), context)
+        || checkDuplicateContext(context, Vertx.currentContext()));
+  }
+
+  private static boolean checkDuplicateContext(final Context context, final Context other) {
```
jnh5y, line 50: *blinks*

I kinda want a method like this to check that the two inputs are duplicate regardless of the order.  Not sure if the refactoring is worth it; your call.

ksqldb-engine/src/main/java/io/confluent/ksql/execution/scalablepush/PushRouting.java
jnh5y: 😆

ksqldb-rest-app/src/main/java/io/confluent/ksql/api/auth/AuthenticationPluginHandler.java
```diff
@@ -107,6 +108,19 @@ public void handle(final RoutingContext routingContext) {
       this.principal = new DefaultKsqlPrincipal(principal);
     }
 
+    @Override
+    public JsonObject attributes() {
```
jnh5y, line 112: What are the new methods used for?

ksqldb-rest-app/src/main/java/io/confluent/ksql/api/auth/JaasAuthProvider.java
```diff
@@ -107,104 +100,80 @@ private void getUser(
...
-    @Override
-    public DefaultKsqlPrincipal getPrincipal() {
-      return principal;
-    }
+    final Authorizations authorizations = new AuthorizationsImpl();
+    user.getSubject()
+        .getPrincipals()
+        .forEach(p -> authorizations.add("default", RoleBasedAuthorization.create(p.getName())));
+
+    promisedUser.complete(new ApiUser() {
```
jnh5y, line 136: Any particular reason to use an anonymous class here?  I'm kinda ambivalent, but figured I'd ask since we had a concrete one previously.

jnh5y (approved):
Nothing looks terribly worrying to me.  Almog, if you think any of the points are worth ironing out, let's chat through it.  Otherwise, I think it looks good to go!

ksqldb-api-client/src/main/java/io/confluent/ksql/api/client/impl/QueryResponseHandler.java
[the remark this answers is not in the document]
agavra: I believe we do still need it - but I'm not 100% sure. I know that some parts of the code rely on the databind mapper (the `vertx.Buffer.toJson` for example defaults to Databind if the library is present)

https://github.com/eclipse-vertx/vert.x/blob/b284c07b4578b5cd5a6e39b153422c0148459226/src/main/java/io/vertx/core/json/jackson/JacksonFactory.java#L25-L34

I was of two minds of whether to just implement our own `JsonFactory` based on our `ApiJsonMapper` and plug into their spi but decided against it (I had actually implemented it) because vertx is a bit inconsistent about using it unfortunately  (see: https://github.com/eclipse-vertx/vert.x/blob/b284c07b4578b5cd5a6e39b153422c0148459226/src/main/java/io/vertx/core/json/impl/JsonUtil.java#L37-L50 where they don't care about what you've specified, this caused me much grief in the migration haha)

ksqldb-api-client/src/test/java/io/confluent/ksql/api/client/integration/ClientIntegrationTest.java
```diff
@@ -1357,11 +1361,33 @@ private static void addObjectToKsqlArray(final KsqlArray array, final Object val
...
       array.add(SqlTimeTypes.formatTime((Time) value));
+    } else if (value instanceof byte[]) {
+      array.add(serializeVertX3CompatibleByte((byte[]) value));
     } else {
       array.add(value);
     }
   }
 
+  /**
+   * VertX 4 changed to serialize using Base64 without padding but our server
```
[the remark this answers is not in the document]
agavra: I gave that a lot of thought but I don't think there should actually be an issue here (at least, the VertX folk seem to agree). The only reason this is a problem is because we're doing a schema-less `String` comparison on what really should be a `byte[]` type. Any Base64 decoder will decode both values identically (`AQ==` is the same as `AQ`), meaning that if you're using the values as a byte[] (which you should be) your client will be fine.

ksqldb-api-client/src/test/java/io/confluent/ksql/api/client/integration/PushV2ClientContinueIntegrationTest.java
agavra: > Can we refactor this to a common place? (Not a blocking question.)

yeah, but we can do that in a different PR 😆  this one was already getting a bit big

> Does `-Dvertx.json.base64=legacy` help us at any point? (I'm blindly cribbing from the docs...)

I tried using that but it's a really crappy workaround that they provide TBH, for one there's no way to do that programmatically because the choice is done in the static class loader so there's no guarantee that the class loader will run my `System.setProperty` call before their static class is loaded.

Then if we want to add it in the runtime configurations, there's no way to do that by default in IntelliJ (it won't respect the maven `pom` runtime parameters because it uses it's own runtime configuration) so even if I could get the test to pass in maven (which it did) you have to specify it manually when you run this test (which works, but is a bad experience)

There's also no guarantees that people will be running whatever scripts we choose when running their clients, so ultimately I think relying on that behavior isn't that straightforward.

ksqldb-common/src/main/java/io/confluent/ksql/reactive/BasePublisher.java
agavra: `Publisher` is super simple:
```
public interface Publisher<T> {
  void subscribe(Subscriber<? super T> var1);
}
```

The reason I made this change is because `HttpConnection#close` became asynchronous. 

That caused some of our tests to fail (specifically the `scalablePushBandwidthThrottleTestHTTP2`) because it assumed that the connection was closed before doing some other stuff. This allowed me to attach a `CountdownLatch` to the close callback.

agavra: > Can line 70 fail in any interesting way?

Good question. It can but we ignore it - so it's trivially "successful". This is all it does:
```
  protected void sendComplete() {
    try {
      sentComplete = true;
      // If a publisher immediately completes, we allow it not to yet have a subscriber, which will
      // get a complete message upon subscription.
      if (subscriber != null) {
        subscriber.onComplete();
      }
    } catch (Exception ex) {
      logError("Exception encountered in onComplete", ex);
    }
  }
```

ksqldb-rest-app/src/main/java/io/confluent/ksql/api/auth/AuthenticationPluginHandler.java
agavra: we bastardize the vertx authentication  😅   these methods are to help support the json ser/de of the principals, but we don't use that

ksqldb-rest-app/src/main/java/io/confluent/ksql/api/auth/JaasAuthProvider.java
agavra: not really 🤷  I think this is an artifact of some refactoring I did but I don't remember what

ksqldb-rest-app/src/main/java/io/confluent/ksql/api/auth/AuthenticationPluginHandler.java
agavra: image

We don't use that method at all because we have an SPI for creating principals and not a json API

ksqldb-api-client/src/main/java/io/confluent/ksql/api/client/impl/QueryResponseHandler.java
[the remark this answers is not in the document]
jnh5y: Sounds good.

ksqldb-api-client/src/test/java/io/confluent/ksql/api/client/integration/PushV2ClientContinueIntegrationTest.java
jnh5y: Yeah, feel free to push that change to another PR (or drop it; can't be DRY every day).

And thanks for trying that System property!  It does sound a little finicky.

agavra:
Thanks for the review @jnh5y! I'll go ahead and merge this and do some of the minor refactoring in a follow-up PR so I don't get any merge conflicts or have to wait for another green build 😆

ksqldb-rest-app/src/main/java/io/confluent/ksql/api/server/AuthHandlers.java
```diff
@@ -52,51 +57,73 @@ static void setupAuthHandlers(final Server server, final Router router,
...
+      jaas.ifPresent(h -> router.route().handler(wrappedHandler(h, Provider.JAAS)));
+      router.route().handler(wrappedHandler(
+          new RoleBasedAuthZHandler(
+              server.getConfig().getList(KsqlRestConfig.AUTHENTICATION_ROLES_CONFIG)),
+          Provider.JAAS));
 
+      // set up using PLUGIN
+      pluginHandler.ifPresent(h -> router.route().handler(wrappedHandler(h, Provider.PLUGIN)));
+
+      router.route().handler(AuthHandlers::pauseHandler);
```
vcrfxia, line 73: I thought order was important here -- that the pauseHandler had to come before either the JAAS or plugin handler. Did this still work even after the order was flipped?

ksqldb-rest-app/src/main/java/io/confluent/ksql/api/server/AuthHandlers.java
```diff
@@ -52,51 +57,73 @@ static void setupAuthHandlers(final Server server, final Router router,
...
-      if (authHeader != null && authHeader.toLowerCase().startsWith("basic ")) {
-        jaasAuthHandler.get().handle(routingContext);
-        return;
+  private static Handler<RoutingContext> wrappedHandler(
+      final Handler<RoutingContext> handler,
+      final Provider provider
+  ) {
+    return rc -> {
+      if (rc.data().get(PROVIDER_KEY) != provider) {
+        rc.next();
```
vcrfxia, line 89: nit: can we add a code comment about why this is OK? Took me a number of read-throughs to understand how `selectHandler()` ensures that a valid handler will always be selected, so we'll never end up in a situation where all the handlers are skipped and no auth is performed.

ksqldb-rest-app/src/main/java/io/confluent/ksql/api/server/AuthHandlers.java
```diff
@@ -113,15 +140,10 @@ private static Optional<AuthHandler> getJaasAuthHandler(final Server server) {
...
 
-  private static AuthHandler basicAuthHandler(final Server server) {
-    final AuthProvider authProvider = new JaasAuthProvider(server, server.getConfig());
+  private static BasicAuthHandler basicAuthHandler(final Server server) {
     final String realm = server.getConfig().getString(KsqlRestConfig.AUTHENTICATION_REALM_CONFIG);
-    final AuthHandler basicAuthHandler = BasicAuthHandler.create(authProvider, realm);
-    // It doesn't matter what we set here as we actually do the authorisation at the
-    // authentication stage and cache the result, but we must add an authority or
-    // no authorisation will be done
-    basicAuthHandler.addAuthority("ksql");
```
vcrfxia, line 123: Note for posterity: the reason it's OK to remove this is because OrHandlerImplementation (called indirectly from RoleBasedAuthZHandler) will properly fail if no roles are provided.

ksqldb-rest-app/src/test/java/io/confluent/ksql/api/CorsTest.java
vcrfxia: We're going to need a system test update when this change gets deployed: https://github.com/confluentinc/cc-system-tests-ksqldb/blob/953ba5df208ccd7cc2c3e4011465b81ecc9e69a7/test/ksql/ksql_test.go#L749

vcrfxia (commented):
Hey @agavra , sorry I'm late with this review -- left a few notes inline. Thanks for the change!

ksqldb-rest-app/src/main/java/io/confluent/ksql/api/server/AuthHandlers.java
agavra: ordering is important here, which is why I needed to move it around. The new default implementation of the `BasicAuthHandler.create` that we plug into now calls pause/resume for you (meaning that it was resumed before it got to the `KsqlAuthorizationProviderHandler`).

ksqldb-rest-app/src/main/java/io/confluent/ksql/api/server/AuthHandlers.java
agavra: 👍

ksqldb-rest-app/src/main/java/io/confluent/ksql/api/server/AuthHandlers.java
vcrfxia: But it's not just the `KsqlAuthorizationProviderHandler` that the context needs to be paused for, right? I think we also want it paused for the `pluginHandler` and the `RoleBasedAuthZHandler` too. 

From the docs, the request needs to be paused before any async calls are made: https://vertx.io/docs/vertx-web/java/#_request_body_handling

The `pluginHandler` is a user's custom auth handler implementation, so we can't control whether async calls are being made or not.

`RoleBasedAuthZHandler` calls out to `AuthorizationHandler.create(...)` which I believe makes it [here](
https://github.com/vert-x3/vertx-web/blob/2f9905b36ff3cbc1f9cb6275bcc3f7a192370a0e/vertx-web/src/main/java/io/vertx/ext/web/handler/impl/AuthorizationHandlerImpl.java#L107), and this looks to be an async call.

vcrfxia: It might be safer to just pause for all handlers _except_ the one handler which has its own pause/resume logic built-in. That way we don't have to reason about which ones need to be paused for and which don't.

agavra: makes sense, I'll make that change. For some reason I thought I followed that code path and it worked but now I can't do that again so either there were refactors I made that don't work or it never worked from the beginning.

Thanks for the review!

PR #8983, opened 2022-04-05
Revert "feat: Allow to plug-in custom error handling for Connect serv…

vcrfxia:
…er errors (#8480)"

This reverts commit c4b4d67ed993b9bab0b5cfe4b1d705bfd62060e7.

### Description 

... in preparation for removing ErrorEntity, in order to clean up error handling for ksql's connect integration. See https://github.com/confluentinc/ksql/pull/8998 for more.

We thought we had a use case for this feature at the time that it was implemented but we found a different way around it and the feature is currently unused (and was intentionally never documented for public consumption).

### Testing done 

Build passes.

### Reviewer checklist


pgaref (approved):
LGTM -- just wondering is we want to keep any of the unrelated changes (can file separately of course)
* ConnectExecutor#execute is an instance method now
* CustomExecutors is a class now rather than an enum of static methods

Gerrrr (approved):

vcrfxia:
> just wondering if we want to keep any of the unrelated changes (can file separately of course)

I think there's a good chance that we'll want to reintroduce these changes in the future (whenever we next want to make a custom executor configurable) but that we don't to introduce this change preemptively. In other words, we can leave it out for now and reintroduce it when it becomes needed again. I don't think it's inherently better with the additional changes, besides that it allows custom executors to be configured.

vcrfxia:
PR build passed and the only failed check is the semantic title check, which I'll override manually when merging. Merging now!

Issue #8952, opened 2022-03-30
Need better error messaging for "Failed to serialize Avro data from topic ..."


ybyzek:
**Is your feature request related to a problem? Please describe.**

Trying to do an `INSERT INTO` on an Avro formatted stream, three people spent a long time troubleshooting the following error message:

```
Failed to insert values into 'AIRCRAFT'. Could not serialize value: [ xxx ]. Error serializing message to topic: xxx. Failed to serialize Avro data from topic xxx :
```

We went down a lot of paths to figure out what was wrong, and it turns out the issue was (as diagnosed by @jzaralim , thank you!)

> Whats happening is, the name of the registered schema is `io.confluent.demo.aircraft.avro.pojo.AircraftState` , but when ksqlDB does an insert, it tries to register a schema. And the schema will always be named `KsqlDataSourceSchema`

This is because the stream had already some data in it produced by a Java app, and then ksqlDB was doing `INSERT INTO`.  

**Describe the solution you'd like**
Given the amount of time spent troubleshooting, it would be an enhancement to provide better messaging that indicated what the problem was (schema name conflict) and what the solution was: adding `KEY_FORMAT='KAFKA', VALUE_SCHEMA_ID=10001` to the `CREATE STREAM` command).

**Describe alternatives you've considered**

**Additional context**

spena:
Reopene issue 'cause the fix only works for Protobuf

suhas-satish:
@aliehsaeedii , can you check if this is a duplicate of another P0 that you're already working on and link it here ?

stebl:
> Reopene issue 'cause the fix only works for Protobuf
Is this still true?  I found this issue while debugging what looks to be a similar root cause.  

https://github.com/confluentinc/ksql/pull/9637 looks like it only addressed the error message.

PR #8984, opened 2022-04-06
fix: INSERT VALUES fail when SR schema has a non-default name

spena:
### Description 
Fixes https://github.com/confluentinc/ksql/issues/8952

This PR allows INSERT VALUES statements to insert data on a stream that was created with a SR schema that uses a non-default name. 

[Issue]
When ksql infers a stream schema from SR, it ignores the SR schema name and uses a default name provided by the Connect format, such as ConnectDefault1 for Protobuf  or KsqlDataSourceSchema for Avro. However, when inserting values to the stream, ksql expects the SR schema to use the default names mentioned. If other names are used, then the INSERT fails.

```
# protobuf schema
syntax = "proto2";
package io.confluent.examples.proto;

message product {
  optional uint32 PRODUCT_ID = 1;
  optional string PRODUCT_NAME = 2;
  optional double PRODUCT_PRICE = 3;
}

ksql> create stream products with (kafka_topic='products', value_format='protobuf');

 Message        
----------------
 Stream created 
----------------

ksql> insert into products(product_id, product_name, product_price) values(5, 'p5', 5.99);
Failed to insert values into 'PRODUCTS'. Could not serialize value: [ 5 | 'p5' | 5.99 ]. Error serializing message to topic: products. Failed to serialize Protobuf data from topic products
 :
```

The reason of the failure is that ksql and Connect API look up the schema (with the default name) in SR for compatibility checks before serializing the value. 

[fix]
The fix starts in the `InsertValuesExecutor`. When a INSERT  VALUES statement is executed, it fetches the schema from SR to extract the full schema name, and then uses the format property `fullSchemaName` to serialize the value to insert. Connect later checks with SR if the schema use for serialization exists in SR and if they are compatible.

### Testing done 
Unit tests and manual tests

### Reviewer checklist


ksqldb-rest-app/src/test/java/io/confluent/ksql/rest/server/execution/InsertValuesExecutorTest.java
```diff
@@ -1065,6 +1072,60 @@ public void shouldThrowOnSchemaInferenceMismatchForKey() throws Exception {
...
+        )
+    );
+
+    // When:
+    executor.execute(statement, mock(SessionProperties.class), engine, serviceContext);
+
+    // Then:
+    verify(keySerdeFactory).create(
+        FormatInfo.of(FormatFactory.AVRO.name(), ImmutableMap.of(
+            AvroProperties.FULL_SCHEMA_NAME,"io.confluent.ksql.avro_schemas.KsqlDataSourceSchema"
```
lihaosky, line 1106: nit: is `io.confluent.ksql.avro_schemas.KsqlDataSourceSchema` the default name? If so, can you change it to something which is not default?

ksqldb-rest-app/src/test/java/io/confluent/ksql/rest/server/execution/InsertValuesExecutorTest.java
```diff
@@ -1065,6 +1072,60 @@ public void shouldThrowOnSchemaInferenceMismatchForKey() throws Exception {
...
+        new KsqlConfig(ImmutableMap.of()),
+        srClientFactory,
+        "",
+        NoopProcessingLogContext.INSTANCE,
+        Optional.empty()
+    );
+
+    verify(valueSerdeFactory).create(
+        FormatInfo.of(FormatFactory.AVRO.name(), ImmutableMap.of(
+            AvroProperties.FULL_SCHEMA_NAME,"io.confluent.ksql.avro_schemas.KsqlDataSourceSchema"
```
lihaosky, line 1118: ditto

ksqldb-serde/src/main/java/io/confluent/ksql/serde/SchemaFullNameAppender.java
```diff
@@ -65,14 +79,14 @@ public String name() {
     }
   }
 
-  private static String avroCompatibleFieldName(final Field field) {
+  private static String compatibleFieldName(final Field field) {
     // Currently the only incompatible field names expected are fully qualified
     // column identifiers. Once quoted identifier support is introduced we will
     // need to implement something more generic here.
     return field.name().replace(".", "_");
```
lihaosky, line 86: I'm not sure if this replacement is specific for avro format? If protobuf supports ".", we don't need to replace? In general, I'm not sure how many of the conversions are also needed for protobuf and concerned if we convert too much and break existing protobuf.

Also for the `MAP_KEY_NAME` and and `MAP_VALUE_NAME`, not sure if protobuf needs them

ksqldb-serde/src/main/java/io/confluent/ksql/serde/protobuf/ProtobufDataTranslator.java
```diff
@@ -0,0 +1,120 @@
...
+public class ProtobufDataTranslator implements DataTranslator {
+  private final DataTranslator innerTranslator;
+  private final Schema ksqlSchema;
+  private final Schema protoCompatibleSchema;
+
+  ProtobufDataTranslator(final Schema schema, final String schemaFullName) {
+    this.ksqlSchema = Objects.requireNonNull(schema, "schema");
+
+    this.protoCompatibleSchema = SchemaFullNameAppender.appendSchemaFullName(
+        schema, schemaFullName
```
lihaosky, line 43: Seems we only care about the schema name here, not sure if we need to call `appendSchemaFullName` which builds other changes.

ksqldb-serde/src/main/java/io/confluent/ksql/serde/protobuf/ProtobufDataTranslator.java
```diff
@@ -0,0 +1,120 @@
...
+  }
+
+  @Override
+  public Object toKsqlRow(final Schema connectSchema, final Object connectObject) {
+    final Object protoCompatibleRow = innerTranslator.toKsqlRow(connectSchema, connectObject);
+    if (protoCompatibleRow == null) {
+      return null;
+    }
+
+    return replaceSchema(ksqlSchema, protoCompatibleRow);
```
lihaosky, line 60: These were for avro schema. 

> Responsible for converting the KSQL schema to a version ready for connect to convert to an
avro schema.
This includes ensuring field names are valid Avro field names and that nested types do not
have name clashes.

Do we need this for protobuf? I'm concerned we are changing too much which is not necessary.

lihaosky (commented):
Thanks @spena ! I suggest to have others taking a look as well. Maybe @vcrfxia or @agavra . I couldn't think of other approaches to resolve this, maybe others can chime in if this is the best solution

ksqldb-serde/src/main/java/io/confluent/ksql/serde/protobuf/ProtobufDataTranslator.java
spena: Replacing the schema to add the full schema name is the main thing needed to make the INSERT to work. The Connect serializer checks the row schema and try to match it with SR. If it has a different name, then it fails.

I debugged this when the `toConnectRow` is called, though. I haven't found a test that calls the `toKsqlRow` to verify it. Any ideas?

ksqldb-serde/src/main/java/io/confluent/ksql/serde/protobuf/ProtobufDataTranslator.java
spena: The `appendSchemaFullName` just adds the new name to the passed schema, right? It has other changes to replace `.` for `_`, which Avro and Protobuf don't support. I just found the bug in Protobuf https://github.com/confluentinc/ksql/issues/9005

Without this full name added to the schema, the INSERT fails too.

ksqldb-serde/src/main/java/io/confluent/ksql/serde/SchemaFullNameAppender.java
spena: It is also for Protobuf. See the bug https://github.com/confluentinc/ksql/issues/9005

ksqldb-rest-app/src/test/java/io/confluent/ksql/rest/server/execution/InsertValuesExecutorTest.java
spena: Done

ksqldb-rest-app/src/test/java/io/confluent/ksql/rest/server/execution/InsertValuesExecutorTest.java
spena: Done

ksqldb-serde/src/main/java/io/confluent/ksql/serde/avro/AvroSchemaTranslator.java
```diff
@@ -57,8 +58,8 @@ public Schema toConnectSchema(final ParsedSchema schema) {
 
   @Override
   public ParsedSchema fromConnectSchema(final Schema schema) {
-    final Schema avroCompatibleSchema = AvroSchemas
-        .getAvroCompatibleConnectSchema(schema, formatProps.getFullSchemaName());
+    final Schema avroCompatibleSchema = SchemaFullNameAppender
```
vcrfxia, line 61: Do we not need this change for JSON_SR as well? Does JSON_SR not have the same schema name requirements?

ksqldb-rest-app/src/test/java/io/confluent/ksql/rest/server/execution/InsertValuesExecutorTest.java
```diff
@@ -1105,7 +1176,40 @@ public void shouldBuildCorrectSerde() {
...
+    );
+
+    // Then:
+    assertThat(e.getMessage(), containsString(
+        "Authorization denied to Read on Schema Registry subject: ["
+            + KsqlConstants.getSRSubject(TOPIC_NAME, true)));
+  }
+
+  @Test
+  public void shouldThrowWhenNotAuthorizedToReadKeySchemaFromSR() throws Exception {
```
vcrfxia, line 1211: How is this test different from `shouldThrowWhenNotAuthorizedToReadKeySchemaToSR()` above?

vcrfxia (commented):
Thanks @spena -- some questions inline!

ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/execution/InsertValuesExecutor.java
```diff
@@ -356,14 +367,23 @@ private static void ensureKeySchemasMatch(
       return;
     }
 
+    final SchemaRegistryClient schemaRegistryClient = serviceContext.getSchemaRegistryClient();
+
+    final FormatInfo formatInfo = addSerializerMissingFormatFields(
```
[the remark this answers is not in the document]
spena: Prior to this PR, insets only worked with SR schema using a default name. If a non-default name was used, then INSERT fail.

The problem is that  when calling the Connect API to serialize the row, the API looks at the row schema and checks with SR that the schema exists. If the row schema has a different name than SR, then the API fails. If the row schema does not have a name, then the API uses a default one. The fix is to add the full name to the schema before calling the API.

I couldn't make the API to use the SR schema to serialize the row. The API validates the SR schema matches the row KSQL intents to serialize.

ksqldb-serde/src/main/java/io/confluent/ksql/serde/SchemaFullNameAppender.java
spena: There are docs for:
Avro https://avro.apache.org/docs/current/spec.html#names
```
Names

Record, enums and fixed are named types. Each has a fullname that is composed of two parts; a name and a namespace. Equality of names is defined on the fullname.

The name portion of a fullname, record field names, and enum symbols must:

    start with [A-Za-z_]
    subsequently contain only [A-Za-z0-9_]

```
- Proto https://developers.google.com/protocol-buffers/docs/reference/proto3-spec
```
Letters and digits  
  letter = "A" … "Z" | "a" … "z"
  decimalDigit = "0" … "9"

Identifiers
  ident = letter { letter | decimalDigit | "_" }
  fieldName = ident
```

It shouldn't be a problem. There's still a bug, though, for Protobuf  before and after this PR where CREATE STREAM does  not accept `.` column names (https://github.com/confluentinc/ksql/issues/9005). I am unsure if we should allow that yet, but this PR does not fix it.

For the other logic, such as Maps. This is the part I agree I don't know if we need it. I will figure out if I can get rid of it for Protobuf.

spena:
@lihaosky @vcrfxia I decided to create a `ProtobufSchemas` class instead. I removed the old `SchemaFullNameAppender` and return the old `AvroSchemas` back. The new `ProtobufSchemas` just adds the schema name to the parent protobuf schema. It does not have all the logic that `AvroSchemas` has. I didn't find an incompatible use case, but just in case, I will move slowly with these Proto changes.

ksqldb-serde/src/main/java/io/confluent/ksql/serde/protobuf/ProtobufSchemas.java
```diff
@@ -0,0 +1,76 @@
...
+        builder = SchemaBuilder.array(schema.valueSchema());
+        break;
+      case MAP:
+        builder = SchemaBuilder.map(
+            schema.keySchema(),
+            schema.valueSchema()
+        );
+        break;
+      default:
+        builder = new SchemaBuilder(schema.type());
```
lihaosky, line 46: Curious in what situation the root schema is not a `struct`? Should we just deal with `struct`?

ksqldb-serde/src/main/java/io/confluent/ksql/serde/protobuf/ProtobufDataTranslator.java
```diff
@@ -0,0 +1,115 @@
...
+        final Map<Object, Object> ksqlMap = new HashMap<>();
+        ((Map<Object, Object>) object).forEach(
+            (key, value) -> ksqlMap.put(
+                replaceSchema(schema.keySchema(), key),
+                replaceSchema(schema.valueSchema(), value)
+            ));
+        return ksqlMap;
+
+      case STRUCT:
+        return convertStruct((Struct) object, schema);
```
lihaosky, line 104: Seems the purpose for now is just setting the struct name? Can we just do that? I think for `avro`, it's doing more stuff because the `compatibleSchema` is more complex.

ksqldb-serde/src/main/java/io/confluent/ksql/serde/avro/AvroSRSchemaDataTranslator.java
```diff
@@ -1,69 +0,0 @@
...
-/**
- * Translates KSQL data and schemas to Avro equivalents.
- *
- * <p>Responsible for converting the KSQL schema to a version ready for connect to convert to an
- * avro schema.
- *
- * <p>This includes ensuring field names are valid Avro field names and that nested types do not
- * have name clashes.
- */
-public class AvroSRSchemaDataTranslator extends ConnectSRSchemaDataTranslator {
```
lihaosky, line 34: Now I remember the reason I added just this for Avro might be `protobuf` doesn't support optional fields and `json_sr`'s field is always optional. So it's different I think. Can we avoid this refactor and just fix the schema name problem?

lihaosky (commented):
High level question is that can we set `VALUE_SCHEMA_FULL_NAME` or `KEY_SCHEMA_FULL_NAME` during schema inference time so that we don't need to check SR and set it.

ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/execution/InsertValuesExecutor.java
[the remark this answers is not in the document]
vcrfxia: IIUC, you're saying that the Connect APIs validate that the schema object we pass matches what's in SR and fails if it doesn't. If that's the case, then we have a problem beyond just figuring out the base schema name, right? The logic in AvroSchemas right now has logic for creating additional schema names (for maps and structs) in a predetermined way from the base schema name provided. If the user has a schema that doesn't fit this pattern, then even if we fetch the base schema name from SR, we won't have a fully matching schema afterwards and the insertion will still fail.

I think the only foolproof way to get around this is to fetch the entire schema from SR and use that for serialization (possibly converting the fetched schema into a Connect schema first, if that's required by the Connect APIs). If we don't want to fetch on every insertion (as @lihaosky pointed out in another comment), then we'll have to store the schema somehow. Either way, this seems like it might be a larger conversation with product implications -- we could take the approach in this PR and solve the issue for (hopefully) most use cases, but we won't have a complete solution unless we do something more drastic.

I wonder why the Connect APIs we use are so rigid. Does it make sense for us to consider relaxing requirements from their API (with a contribution ourselves) as another option around these challenges?

ksqldb-engine/src/main/java/io/confluent/ksql/schema/registry/SchemaRegistryUtil.java
```diff
@@ -106,6 +106,39 @@ public static boolean subjectExists(
...
+          throw new KsqlSchemaAuthorizationException(
+              deniedOperation,
+              subject
+          );
+        }
+      }
+
+      throw new KsqlException(
+          "Could not get schema for subject " + subject + " and id " + schemaId, e);
+    }
```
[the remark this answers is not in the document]
spena: Done

ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/execution/InsertValuesExecutor.java
```diff
@@ -435,6 +462,48 @@ private byte[] serializeValue(
...
+    final ImmutableMap.Builder propertiesBuilder = ImmutableMap.builder();
+    propertiesBuilder.putAll(formatInfo.getProperties());
+
+    // The FULL_SCHEMA_NAME allows the serializer to choose the schema definition
+    if (supportedProperties.contains(ConnectProperties.FULL_SCHEMA_NAME)) {
+      if (!formatInfo.getProperties().containsKey(ConnectProperties.FULL_SCHEMA_NAME)) {
+        SchemaRegistryUtil.getParsedSchema(srClient, topicName, isKey).map(ParsedSchema::name)
+            .ifPresent(schemaName ->
+                propertiesBuilder.put(ConnectProperties.FULL_SCHEMA_NAME, schemaName));
+      }
```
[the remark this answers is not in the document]
spena: I made an improvement on the latest commit. I obtained the key/value schema full name during creation and keep it in the metastore to avoid hurting the performance. This woks only for new streams and tables, though. When users upgrade their ksql, the older streams will check for SR to get the key/value full name. It's good user experience to not make users re-create their streams. Btw, ksql uses a cached SR client, so consecutive INSERT calls won't get bad performance. 

@colinhicks any thoughts?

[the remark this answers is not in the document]
colinhicks: I think it's fine for pre-existing objects to have reduced insert performance. We don't promote INSERT ... VALUES as an optimized write path.

ksqldb-serde/src/main/java/io/confluent/ksql/serde/protobuf/ProtobufSchemas.java
spena: Done

ksqldb-serde/src/main/java/io/confluent/ksql/serde/avro/AvroSchemas.java
```diff
@@ -24,12 +24,12 @@
 import org.apache.kafka.connect.data.Schema;
 import org.apache.kafka.connect.data.SchemaBuilder;
 
-final class AvroSchemas {
+public final class AvroSchemas {
```
lihaosky, line 27: Revert this change?

lihaosky (approved):
LGTM! Thanks @spena ! We can revert the changes in `DefaultSchemaInjector` and create an issue to track it.

ksqldb-serde/src/main/java/io/confluent/ksql/serde/avro/AvroSchemas.java
spena: Done

PR #8985, opened 2022-04-06
chore: resolve configs before checking if they are query level

wcarlson5:
### Description 
resolve the configs before checking if they are at the query level. This way we can have the ksql.streams prefix for streams configs
### Testing done 

### Reviewer checklist


tolgadur (approved):
Maybe a short test to verify that it works as expected? Otherwise lgtm

PR #8986, opened 2022-04-06
feat: allow aggregations without group bys

### LINKED ISSUE #430: Cannot create aggregate without a GROUP BY
KSQL insists on having a `GROUP BY` column, even if logically it makes sense without. 

For example: how many tweets were there in each hourly period? 

```
ksql> select count(*) from twitter window tumbling (size 1 hour);
Aggregate query needs GROUP BY clause.
ksql>
```

nateab:
### Description 
Helps to partially fix #430 by allowing users to create aggregations without group by clauses in push queries .

### Testing done 

### Reviewer checklist

### DIFF
--- a/ksqldb-engine/src/main/java/io/confluent/ksql/analyzer/Analyzer.java
+++ b/ksqldb-engine/src/main/java/io/confluent/ksql/analyzer/Analyzer.java
@@ -30,6 +30,7 @@ import io.confluent.ksql.execution.expression.tree.ComparisonExpression;
 import io.confluent.ksql.execution.expression.tree.ComparisonExpression.Type;
 import io.confluent.ksql.execution.expression.tree.Expression;
 import io.confluent.ksql.execution.expression.tree.FunctionCall;
+import io.confluent.ksql.execution.expression.tree.IntegerLiteral;
 import io.confluent.ksql.execution.expression.tree.LogicalBinaryExpression;
 import io.confluent.ksql.execution.expression.tree.SearchedCaseExpression;
 import io.confluent.ksql.execution.expression.tree.TraversalExpressionVisitor;
@@ -43,6 +44,7 @@ import io.confluent.ksql.name.ColumnName;
 import io.confluent.ksql.name.FunctionName;
 import io.confluent.ksql.name.SourceName;
 import io.confluent.ksql.parser.DefaultTraversalVisitor;
+import io.confluent.ksql.parser.NodeLocation;
 import io.confluent.ksql.parser.properties.with.CreateSourceAsProperties;
 import io.confluent.ksql.parser.tree.AliasedRelation;
 import io.confluent.ksql.parser.tree.AllColumns;
@@ -157,7 +159,9 @@ class Analyzer {
       this.analysis = new Analysis(
               query.getRefinement(),
               rowpartitionRowoffsetEnabled,
-              pullLimitClauseEnabled);
+              pullLimitClauseEnabled,
+              query
+          );
 
       this.persistent = persistent;
     }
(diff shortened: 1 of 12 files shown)


ksqldb-engine/src/main/java/io/confluent/ksql/analyzer/Analyzer.java
```diff
@@ -668,8 +671,11 @@ private void throwOnUdafs(final Expression expression) {
         public Void visitFunctionCall(final FunctionCall functionCall, final Void context) {
           final FunctionName functionName = functionCall.getName();
           if (metaStore.isAggregate(functionName)) {
-            throw new KsqlException("Use of aggregate function "
-                + functionName.text() + " requires a GROUP BY clause.");
+            analysis.addAggregateFunction(functionCall);
```
agavra, line 674: we should probably change the method name, right?

ksqldb-engine/src/main/java/io/confluent/ksql/analyzer/Analyzer.java
```diff
@@ -668,8 +671,11 @@ private void throwOnUdafs(final Expression expression) {
...
+            // Since this is a dummy group by, we don't actually need a correct node location
+            final NodeLocation nodeLocation = new NodeLocation(0,0);
```
agavra, line 676: can we just use `Optional.empty()` here?

ksqldb-functional-tests/src/test/resources/rest-query-validation-tests/push-queries.json
```diff
@@ -653,6 +653,26 @@
         ]}
       ]
     },
+    {
+      "name": "aggregate without GROUP BY expression",
```
agavra, line 657: can we add tests for (1) what happens if you try to `CREATE TABLE` without a group by (basically confirm a good error message), (2) what happens if you select a non-grouping column (e.g. `SELECT COUNT(*), ID`) and (3) that multiple aggregate functions work?

ksqldb-engine/src/main/java/io/confluent/ksql/analyzer/Analyzer.java
```diff
@@ -668,8 +671,11 @@ private void throwOnUdafs(final Expression expression) {
...
+            analysis.setGroupBy(new GroupBy(Optional.of(nodeLocation),
+                ImmutableList.of(new IntegerLiteral(1))));
```
agavra, line 678: thoughts on using just a single byte to save some space?

agavra (commented):
Thanks Natea - this looks way simpler than I thought it would be! that's pretty elegant. I think we should try to make sure the error message is super clear if you try to `CREATE TABLE` without a GROUP BY but with an aggregate function (e.g. it should say "not supported without GROUP BY" instead of "must select key").

ksqldb-engine/src/main/java/io/confluent/ksql/analyzer/Analyzer.java
nateab: yes good catch!

ksqldb-engine/src/main/java/io/confluent/ksql/analyzer/Analyzer.java
nateab: added

ksqldb-engine/src/main/java/io/confluent/ksql/analyzer/Analyzer.java
nateab: switched to BytesLiteral

ksqldb-functional-tests/src/test/resources/rest-query-validation-tests/push-queries.json
nateab: I will add tests covering cases 1 and 3.

For 2, this is what currently happens 
```
ksql> select avg_stars,count(*) from RATINGS_TABLE EMIT CHANGES;
Non-aggregate SELECT expression(s) not part of GROUP BY: AVG_STARS
Either add the column(s) to the GROUP BY or remove them from the SELECT.
```
Is that reasonable?

nateab: discussed offline, confirmed this was reasonable

hli21: How about a stream pull query case, like "SELECT count(column) AS count FROM test"?

nateab: I have manually tested the stream pull query case, but I can also add an explicit test case for it too

nateab: Actually pull queries dont support group by clauses at all, unless you mean add a test to make sure we get the expected error? This is the current error
```
Pull queries don't support GROUP BY clauses. See https://cnfl.io/queries for more info.
Statement: select count(*), sum(user_id) from RATINGS_STREAM;
```

agavra (approved):
Nice! Thanks @nateab

PR #8859, opened 2022-03-07
chore: Update references to ConnectorPluginInfo to PluginInfo

jnh5y:
Addresses https://github.com/confluentinc/ksql/issues/8841 / https://github.com/confluentinc/ksql/issues/8863

### Description 
This follows up on a change in Apache Kafka here: https://github.com/apache/kafka/commit/029a14b530dcaab980faee4fa18c925b95db3122

### Testing done 
Unit test updates should be sufficient.

### Reviewer checklist
- [ ] Ensure docs are updated if necessary. (eg. if a user visible feature is being added or changed).
- [ ] Ensure relevant issues are linked (description should include text like "Fixes #<issue number>")

PR #8990, opened 2022-04-07
refactor: migrate away from Kafka Connect internal library

jzaralim:
### Description 
We use the `org.apache.kafka.connect.runtime` library pretty often, which is an internal API and should not be used because if there is a change in there, then our builds will fail. There are some usages of this library that still remain, but this handles the following classes:

*  `org.apache.kafka.connect.runtime.rest.entities.ConnectorInfo ` -> `io.confluent.ksql.rest.entity.ConnectorInfo `
* `org.apache.kafka.connect.runtime.rest.entities.ConnectorStateInfo ` -> `io.confluent.ksql.rest.entity.SimpleConnectorStateInfo `
* `org.apache.kafka.connect.runtime.rest.entities.ConnectorType `-> `io.confluent.ksql.rest.entity.ConnectorType `
* `org.apache.kafka.connect.runtime.rest.entities.PluginInfo ` -> `io.confluent.ksql.rest.entity.SimpleConnectorPluginInfo `
* `org.apache.kafka.connect.runtime.rest.entities.ConfigInfos ` -> `io.confluent.ksql.rest.entity.SimpleConfigInfos `

Most of these were copy pasted from the Connect library with fields that we don't use deleted. The copying is necessary because Connect's public APIs are limited.

### Testing done 
Integration tests still pass

### Reviewer checklist


ksqldb-rest-model/src/main/java/io/confluent/ksql/rest/entity/SimpleConnectorPluginInfo.java
```diff
@@ -33,7 +32,7 @@
 
   @JsonCreator
   public SimpleConnectorPluginInfo(
-      @JsonProperty("className") final String className,
+      @JsonProperty("class") final String className,
```
vcrfxia, line 35: This change is backwards incompatible, right? This class is returned from the public API response for `DESCRIBE CONNECTOR PLUGINS;`. I don't think we can make a change like this without bumping the API version.

What's the motivation for this change?

ksqldb-engine/src/main/java/io/confluent/ksql/services/ConnectJsonMapper.java
```diff
@@ -40,14 +41,15 @@
...
-    module.addSerializer(PluginInfo.class, new PluginInfoJsonSerializer());
+    module.addSerializer(SimpleConnectorPluginInfo.class, new PluginInfoJsonSerializer());
 
     return new ObjectMapper()
         .disable(JsonGenerator.Feature.AUTO_CLOSE_TARGET)
         .registerModule(new Jdk8Module())
         .registerModule(new StructSerializationModule())
         .registerModule(new KsqlTypesSerializationModule())
         .registerModule(module)
+        .registerModule(new GuavaModule())
```
vcrfxia, line 52: Why is this module being added? It looks like the two public-facing classes (SimpleConnectorInfo and SimpleConnectorPluginInfo) aren't using guava's immutable types, and that the new classes that do use guava's immutable types are internal-only, but if that's the case then I'm not seeing why the new types that do use guava'a immutable types are being serialized/deserialized by this mapper. What am I missing?

ksqldb-engine/src/main/java/io/confluent/ksql/services/DefaultConnectClient.java
```diff
@@ -156,15 +156,15 @@ public DefaultConnectClient(
...
+      final ConnectResponse<SimpleConfigInfos> connectResponse = withRetries(() -> Request
           .put(resolveUri(String.format(VALIDATE_CONNECTOR, plugin)))
           .setHeaders(requestHeaders)
           .responseTimeout(Timeout.ofMilliseconds(requestTimeoutMs))
           .connectTimeout(Timeout.ofMilliseconds(requestTimeoutMs))
           .bodyString(MAPPER.writeValueAsString(config), ContentType.APPLICATION_JSON)
           .execute(httpClient)
           .handleResponse(
-              createHandler(HttpStatus.SC_OK, new TypeReference<ConfigInfos>() {},
+              createHandler(HttpStatus.SC_OK, new TypeReference<SimpleConfigInfos>() {},
```
vcrfxia, line 167: What's the purpose in switching to our own snapshot of these Connect APIs? If Connect evolves them in a backwards incompatible way, then this client code will start to break anyway. I think I'm not understanding what types of changes in Connect are breaking our builds / what this PR is trying to guard against. Do you have examples?

vcrfxia (commented):
Hey @jzaralim -- left some questions inline!

Also sorry about the merge conflicts introduced by my other PR (https://github.com/confluentinc/ksql/pull/8998). If you're busy I can take a stab at resolving them; I think they should be pretty straightforward.

ksqldb-engine/src/main/java/io/confluent/ksql/services/DefaultConnectClient.java
jzaralim: An example is https://github.com/confluentinc/ksql/pull/8859 - Connect changed the name of one of their classes, which caused ksql to hit compilation errors. We're trying to make us depend only on Connect's changes to their public REST API. The unfortunate thing is that they don't have an API client so anyone who tries to make a Connect client will have to implement POJOs on their own. It's impractical, but IMO if this was a library owned by anyone else, we would not be using internal libraries to begin with.

vcrfxia: Interesting... a rename is pretty much the only example I can think of right now that's non-backwards compatible if we depend on the internal object but is backwards compatible if we take a snapshot. I'm not convinced that these changes happen frequently enough to warrant the changes in this PR but I also don't think the changes in this PR are a dis-improvement compared to what we have today, so if you want to continue with the changes here I can be onboard. Two additional thoughts, though:
* I don't think it's a bad thing that we have to respond when Connect changes their APIs as they did in the example you shared. It's a decently significant change and includes other changes that we might want to respond to besides the ones that broke our build. For example, the change [added new plugin types](https://github.com/apache/kafka/commit/029a14b530dcaab980faee4fa18c925b95db3122#diff-463771dd8339b513a558ad19adccd51f310632ad429f1a2fc337ad3504a4978eR35), which would be nice to support in ksql (when listing connector plugins through ksql). Also, we missed [this change](https://github.com/apache/kafka/commit/029a14b530dcaab980faee4fa18c925b95db3122#diff-d75a4fd905c695af2ad23f2d2db7748a6ee32317c30bcc26e6ffb2106a7de2eaR36) in the public API in our patch -- ksql is still representing plugin types as ConnectorType even though Connect switched to using PluginType instead. I don't think it's a bad thing for us to get eyes on PRs containing Connect API changes (as long as they don't happen so frequently that it just becomes a time sink).
* Our integration test coverage for connect integration is lacking. In particular, we don't have integration test coverage for successful `CREATE CONNECTOR` responses or `LIST CONNECTOR PLUGINS`. This is bad because with the changes in this PR, we'll no longer be alerted when Connect makes API changes. If a backwards incompatible change breaks our snapshot for these statements, we wouldn't even know. This test gap needs to be fixed separate from this PR, but it'd be nice if the improved test coverage landed first or at least shortly after.

suhas-satish: +1 on improving test coverage to catch breaking issues on connect side in our automated tests

jzaralim: If our snapshots break without us knowing, then that is a dis-improvement. I'll close this PR, since we definitely should add the extra test coverage before this. If it does get added we can revisit.

PR #8998, opened 2022-04-07
fix: remove ErrorEntity and throw on connector error instead

vcrfxia:
### Description 

This PR is stacked on https://github.com/confluentinc/ksql/pull/8983. Only the commits starting with `chore: remove ErrorEntity and throw on connector error instead` need to be reviewed.

Currently, when connector requests encounter errors they return a 200 response code with an ErrorEntity object in the response. This is problematic because:
* the 200 response code is confusing, given that the request did not succeed
* in the case when multiple ksql statements are issued as a single request, the 200 response code from a failed connector statement causes subsequent statements to be executed even though they shouldn't be.

This PR updates the connect executors to return an error response (resulting in a non-200 response code) when requests fail.

### Testing done 

Unit + integration + manual.

### Reviewer checklist


ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/validation/CustomValidators.java
```diff
@@ -100,8 +98,7 @@
   SHOW_COLUMNS(ShowColumns.class, ListSourceExecutor::columns),
   EXPLAIN(Explain.class, ExplainExecutor::execute),
   DESCRIBE_FUNCTION(DescribeFunction.class, DescribeFunctionExecutor::execute),
-  DESCRIBE_CONNECTOR(DescribeConnector.class,
-      new DescribeConnectorExecutor(new DefaultConnectServerErrors())::execute),
+  DESCRIBE_CONNECTOR(DescribeConnector.class, StatementValidator.NO_VALIDATION),
```
vcrfxia, line 101: I've removed this validation because it was a no-op -- validation only fails if an error is thrown, and no error was ever thrown from `DescribeConnectorExecutor::validate` since it returned a 200 response with an error entity on error, instead of throwing.

It doesn't make sense to validate this request because there's nothing to validate (validating it amounts to issuing the request to Connect, which is what execution does). The same logic applies to DropConnector and ListConnectors, both of which are already not being validated.

See https://github.com/confluentinc/ksql/pull/8999 for more on validating connector statements.

ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/execution/DescribeConnectorExecutor.java
```diff
@@ -75,22 +78,26 @@ public StatementExecutorResponse execute(
...
-      return StatementExecutorResponse.handled(Optional.of(
-          new ErrorEntity(
-              configuredStatement.getStatementText(),
-              statusResponse.error().get())
-      ));
+      throw new KsqlRestException(EndpointResponse.create()
+          .status(statusResponse.httpCode())
+          .entity(new KsqlErrorMessage(
+              Errors.toErrorCode(statusResponse.httpCode()),
+              "Failed to query connector status: " + statusResponse.error().get()))
```
pgaref, line 85: nit: move errorMsg to a final var for consistency and readability -- applies to all KsqlRestExceptions

ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/execution/ConnectExecutor.java
```diff
@@ -54,8 +59,7 @@ public static StatementExecutorResponse execute(
     final List<String> errors = validate(createConnector, client);
     if (!errors.isEmpty()) {
       final String errorMessage = "Validation error: " + String.join("\n", errors);
-      return StatementExecutorResponse.handled(Optional.of(new ErrorEntity(
-          statement.getStatementText(), errorMessage)));
+      throw new KsqlException(errorMessage);
```
pgaref, line 62: Assuming client connection failed here thus KsqlException instead of KsqlRestException?

ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/validation/CustomValidators.java
pgaref: Good catch, ``DescribeConnectorExecutor::execute`` never throws an exception indeed.

I guess we should do the same for LIST_VARIABLES and DESCRIBE_FUNCTION -- can do it in a follow-up

Gerrrr (approved):
This PR looks good!

ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/execution/ConnectExecutor.java
vcrfxia: Not quite -- KsqlException means it's a user error, which is necessarily the case if there's a validation error but not necessarily the case if other errors are encountered.

The final exception type returned from the engine is always KsqlRestException. KsqlException is automatically wrapped in KsqlRestException here: https://github.com/confluentinc/ksql/blob/7b3fb6b078a02c48ab4fe976ca836b8efe514a59/ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/resources/KsqlResource.java#L328-L341

For the non-validation connector errors, we don't know whether it's a user error or not without inspecting the error itself, which is why this PR simply returns a KsqlRestException directly rather than trying to figure out what inner error to throw in those cases.

ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/execution/DescribeConnectorExecutor.java
vcrfxia: Hm interesting. I think most places in the codebase inline the error message directly rather than having a separate line for `final String errorMsg = ...;` but I do see a few places where we opt for the latter and are inconsistent as a result. For relatively short error messages such as this one, I don't personally think it's harder to read when the error message is inlined but I don't feel too strongly. I'll make the change.

PR #8999, opened 2022-04-07
fix: move create connector validation to validate phase

vcrfxia:
### Description 

This PR is stacked on https://github.com/confluentinc/ksql/pull/8998. Only commits starting with `chore: move create connector validation to validate phase` need to be reviewed.

Currently, a `CREATE CONNECTOR` request has three steps performed during the execution phase, and no validation. The execution steps are, in order:
1. use the Connect client to validate configs
2. check for existing connectors, if `IF NOT EXISTS` is present in the statement
3. use the Connect client to create the connector

The fact that there's no ksql validation for `CREATE CONNECTOR` statements (validation is only performed during execution) isn't great because it means that, in the case where a `CREATE CONNECTOR` statement is issued along with other ksql statements in the same request, the entire group of statements may pass validation and ksql will begin to execute them in order, only to fail when executing `CREATE CONNECTOR`. Instead, it's preferable to validate configs with Connect during ksql engine's validate phase to fail fast.

With the changes in this PR, the new order is the following steps during the validate phase:
1. use the Connect client to validate configs

And then the following two steps during the execute phase:
1. check for existing connectors, if `IF NOT EXISTS` is present in the statement
2. use the Connect client to create the connector

Side note: ideally the `IF NOT EXISTS` handling would happen before the config validation so that we wouldn't require valid configs if the entire `CREATE CONNECTOR` statement is going to be a no-op anyway but doing so means we'd have to add the check to list connectors into the validate phase, which means it'd be repeated during the execute phase. I don't think this minor improvement is worth the additional request to Connect, so I've left it out.

### Testing done 

Unit + integration + manual.

### Reviewer checklist

### DIFF
--- a/ksqldb-engine/src/main/java/io/confluent/ksql/services/SandboxConnectClient.java
+++ b/ksqldb-engine/src/main/java/io/confluent/ksql/services/SandboxConnectClient.java
@@ -17,11 +17,8 @@ package io.confluent.ksql.services;
 
 import static io.confluent.ksql.util.LimitedProxyBuilder.methodParams;
 
-import com.google.common.collect.ImmutableList;
-import io.confluent.ksql.services.ConnectClient.ConnectResponse;
 import io.confluent.ksql.util.LimitedProxyBuilder;
 import java.util.Map;
-import org.apache.hc.core5.http.HttpStatus;
 
 /**
  * Supplies {@link ConnectClient}s to use that do not make any
@@ -36,8 +33,6 @@ final class SandboxConnectClient {
   public static ConnectClient createProxy(final ConnectClient delegate) {
     return LimitedProxyBuilder.forClass(ConnectClient.class)
         .forward("validate", methodParams(String.class, Map.class), delegate)
-        .swallow("connectors", methodParams(),
-            ConnectResponse.success(ImmutableList.of(), HttpStatus.SC_OK))
         .build();
   }
 }
--- a/ksqldb-engine/src/main/java/io/confluent/ksql/services/SandboxConnectClient.java
+++ b/ksqldb-engine/src/main/java/io/confluent/ksql/services/SandboxConnectClient.java
@@ -18,10 +18,13 @@ package io.confluent.ksql.services;
 import static io.confluent.ksql.util.LimitedProxyBuilder.methodParams;
 
 import com.google.common.collect.ImmutableList;
+import com.google.common.collect.ImmutableMap;
 import io.confluent.ksql.services.ConnectClient.ConnectResponse;
 import io.confluent.ksql.util.LimitedProxyBuilder;
 import java.util.Map;
 import org.apache.hc.core5.http.HttpStatus;
+import org.apache.kafka.connect.runtime.rest.entities.ConnectorInfo;
+import org.apache.kafka.connect.runtime.rest.entities.ConnectorType;
 
 /**
  * Supplies {@link ConnectClient}s to use that do not make any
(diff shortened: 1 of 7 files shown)


ksqldb-engine/src/main/java/io/confluent/ksql/services/SandboxConnectClient.java
```diff
@@ -33,18 +36,16 @@ private SandboxConnectClient() {
...
-        .swallow("describe", methodParams(String.class),
-            ConnectResponse.failure("sandbox", HttpStatus.SC_INTERNAL_SERVER_ERROR))
+        .forward("validate", methodParams(String.class, Map.class), delegate)
         .swallow("connectors", methodParams(),
             ConnectResponse.success(ImmutableList.of(), HttpStatus.SC_OK))
-        .swallow("status", methodParams(String.class),
-            ConnectResponse.failure("sandbox", HttpStatus.SC_INTERNAL_SERVER_ERROR))
-        .swallow("delete", methodParams(String.class),
-            ConnectResponse.success("sandbox", HttpStatus.SC_NO_CONTENT))
+        .swallow("create", methodParams(String.class, Map.class),
```
pgaref: Not sure what is the reason behind changing the existing ``create`` behavior?

Gerrrr (commented):
It took me a bit to understand why we can skip `IF NOT EXISTS` check in the validation step, so I would like to make sure I got it correctly.  

`CREATE CONNECTOR IF NOT EXISTS` results in a no-op and a warning if the connector is already present. Hence, there is no reason to check for the existing connector if `IF NOT EXISTS` is specified - it won't break the execution flow. Is this right?

On the other hand, if the connector exists and the `CREATE CONNECTOR` statement does not include `IF NOT EXISTS`, the statement will fail during execution. If this is correct, I'd argue that the validation step should check for the existing connector if the `CREATE CONNECTOR` statement does not contain `IF NOT EXISTS`. WDYT?

ksqldb-engine/src/main/java/io/confluent/ksql/services/SandboxConnectClient.java
vcrfxia: No, that's not enough. The datum type needs to match what's expected here, else this will throw: https://github.com/confluentinc/ksql/blob/7b3fb6b078a02c48ab4fe976ca836b8efe514a59/ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/execution/ConnectExecutor.java#L75

vcrfxia: That said, in the latest version of this PR the create() command is not part of the validation at all so I've simply removed this line from the sandbox client.

vcrfxia:
Hey @Gerrrr , thanks for the questions (and the review)!

> `CREATE CONNECTOR IF NOT EXISTS` results in a no-op and a warning if the connector is already present. Hence, there is no reason to check for the existing connector if `IF NOT EXISTS` is specified - it won't break the execution flow. Is this right?

Ideally, if `IF NOT EXISTS` is present then we would also check if the connector exists already during validation and bypass the config validation check if so, since it shouldn't really matter to a user whether their configs are valid if the entire statement is going to be a no-op anyway. However, if `IF NOT EXISTS` is present then we definitely need to check for existing connectors during the execute phase, so also checking during the validate phase means we'd make two list connector calls to Connect, which makes the minor improvement in user experience seem more questionable. You're right that there's no correctness issue here, just a matter of enhanced user experience. 

> On the other hand, if the connector exists and the `CREATE CONNECTOR` statement does not include `IF NOT EXISTS`, the statement will fail during execution. If this is correct, I'd argue that the validation step should check for the existing connector if the `CREATE CONNECTOR` statement does not contain `IF NOT EXISTS`. WDYT?

Your statement is correct, and I agree with your conclusion that ideally we'd check for an existing connector of the same name during the validate phase for `CREATE CONNECTOR` statements that do not contain `IF NOT EXISTS` in order to fail fast if this is the case. My one hesitation is around the permissions required for this. It's reasonable for a user to expect that a `CREATE CONNECTOR` statement requires CREATE permissions on connectors, and that `CREATE CONNECTOR IF NOT EXISTS` additionally requires LIST permissions on connectors. Should a user also expect that a `CREATE CONNECTOR` statement without `IF NOT EXISTS` also requires LIST permissions on connectors? Maybe the answer's yes, in which case we might as well add this, but if not then I think we should leave it out.

ksqldb-engine/src/main/java/io/confluent/ksql/services/SandboxedServiceContext.java
```diff
@@ -47,7 +47,8 @@ public static SandboxedServiceContext create(final ServiceContext serviceContext
         .createProxy(serviceContext.getTopicClient(), serviceContext::getAdminClient);
     final SchemaRegistryClient schemaRegistryClient =
         SandboxedSchemaRegistryClient.createProxy(serviceContext.getSchemaRegistryClient());
-    final ConnectClient connectClient = SandboxConnectClient.createProxy();
+    final ConnectClient connectClient =
+        SandboxConnectClient.createProxy(serviceContext.getConnectClient()); // TODO: does this load custom configs?
```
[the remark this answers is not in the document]
vcrfxia: Heh, yes. The earlier version of this PR indeed started causing the connect client to be created (and for custom credentials to be loaded) on each sandbox service context invocation, even if the connect client is not needed, which is bad. I've pushed a new commit to fix this, and added a unit test as well.

Gerrrr (approved):

vcrfxia:
> Should a user also expect that a `CREATE CONNECTOR` statement without `IF NOT EXISTS` also requires LIST permissions on connectors? Maybe the answer's yes, in which case we might as well add this, but if not then I think we should leave it out.

Discussed offline -- we believe the answer is yes, users trying to create connectors should be expected to also have permissions to view connectors. I'll add the additional validation as a follow-up PR.

PR #9014, opened 2022-04-11
fix: fail validation on create connector if connector already exists

vcrfxia:
### Description 

Quick follow-up to https://github.com/confluentinc/ksql/pull/8999 (specifically, [this discussion](https://github.com/confluentinc/ksql/pull/8999#issuecomment-1095309050)): for `CREATE CONNECTOR` statements that do not include `IF NOT EXISTS` in the statement, these statements will fail if a connector with the same name already exists. Currently, this failure happens during ksql's execute phase. Ideally, the failure would occur during ksql's validation phase, in order to fail fast so that no statements (in a multi-statement request) are executed. This PR adds a check for this case during validation accordingly.

### Testing done 

Unit + integration.

### Reviewer checklist


Gerrrr (approved):
Thank you for fixing this issue!

PR #9015, opened 2022-04-11
chore: improve registry not found error message

### LINKED ISSUE #8428: Improve scalable push registry cannot be found error message and docs
Customer tried to use scalable push queries in 0.22, but encountered this error message:

`Scalable push registry cannot be found`

This happened even with scalable push queries enabled, and can be reproduced as follows:
```
SET 'auto.offset.reset' = 'latest';
SET 'ksql.query.push.v2.enabled' = 'true';
create stream s (val string) with (kafka_topic='s', value_format='json', partitions=1);
create table agg as select val, count(*) as total from s group by val;
select * from agg emit changes;
```

The problem is that the customer also needs to set another config, `ksql.query.push.v2.registry.installed` to true. 
We should include this suggestion in the error message, and also make it clearer in the docs

nateab:
### Description 
Fixes #8428


agavra (approved):
nice, thanks!

Issue #8978, opened 2022-04-05
Avro headers not read unless other fields are specified in CREATE STREAM


fgong-confluent:
**Describe the bug**
With Avro topics, if specifying no fields in a CREATE STREAM statement except for header(s), ksql skips reading the header. If other fields are specified, the headers are read.

**To Reproduce**
ksqlDB 0.24.0+
1. Produce a record with headers:
```
try (KafkaProducer<String, Payment> producer = new KafkaProducer<String, Payment>(props)) {

    for (long i = 0; i < 1; i++) {
        final String orderId = "id" + Long.toString(i);
        final Payment payment = new Payment(orderId, 1000.00d, 100.0d);
        final ProducerRecord<String, Payment> record = new ProducerRecord<String, Payment>(TOPIC, payment.getId().toString(), payment);
        record.headers().add(new RecordHeader("TimestampUTC", "TimestampUTC".getBytes()));
        record.headers().add(new RecordHeader("TEST", "TimestampUTC".getBytes()));
        producer.send(record);
        Thread.sleep(1000L);
    }
    producer.flush();
```
Schema:
```
{
 "namespace": "io.confluent.examples.clients.basicavro",
 "type": "record",
 "name": "Payment",
 "fields": [
     {"name": "id", "type": "string"},
     {"name": "amount", "type": "double"},
     {"name": "amount2", "type": "double"}
 ]
```

2. Try to create stream without defining other fields:

```
ksql> create stream HEADERS_TEST13(KEY_VALUE0 VARCHAR KEY,
>headers ARRAY<STRUCT<key STRING, value BYTES>> HEADERS
>) with (kafka_topic='transactions1',partitions=1,value_format='avro');
```
```
create stream HEADERS_TEST11(
>TEST BYTES HEADER('TEST')
>) with (kafka_topic='transactions1',partitions=1,value_format='avro');
```

**Expected behavior**
`SELECT * FROM HEADERS_TEST13` shows the header fields as well

**Actual behaviour**
```
select * from HEADERS_TEST13 emit changes;
+--------------------------------------------------+--------------------------------------------------+--------------------------------------------------+--------------------------------------------------+
|KEY_VALUE0                                        |ID                                                |AMOUNT                                            |AMOUNT2                                           |
+--------------------------------------------------+--------------------------------------------------+--------------------------------------------------+--------------------------------------------------+
|id0                                               |id0                                               |1000.0                                            |100.0                                             |
```
```
[...]

PR #9023, opened 2022-04-12
fix: include header columns when injecting schemas

jzaralim:
### Description 
Fixes https://github.com/confluentinc/ksql/issues/8978

When injecting key and value schemas to a source that has header columns, the header columns get lost, causing the schema to be incorrect.

### Testing done 
Unit+QTT tests

### Reviewer checklist


nbonavia:
Looking forward for this to be released.

spena (approved):
LGTM

PR #9026, opened 2022-04-13
feat: support checking preconditions before starting core app

rodesai:
### Description 
This patch changes up the precondition checker a bit so that it
runs before the main KsqlRestApplication. This way, preconditions
can look out for critical config properties to arrive, before they
are handled by the actual application.

Detailed changes:
    - Run the precondition checker as a separate Executable implementation
      before running KsqlRestApplication. The precondition checker waits for
      any preconditions to pass, and while this is ongoing it runs a simple
      server that responds to all requests (other than the health probes)
      with a 503.
    - Change the precondition interface to accept a properties loader so
      that precondition checkers can wait on the arrival of dynamic properties.
    - Support multiple properties files as args (which are then overlaid)
    
### Testing done 
Added unit tests for the new precondition checker
Tested manually by updating a config file at runtime with a precondition plugin that looks
    for such a change.

### Reviewer checklist


ksqldb-rest-app/src/main/java/io/confluent/ksql/api/server/Server.java
```diff
@@ -311,130 +307,24 @@ private static HttpServerOptions createHttpServerOptions(final KsqlRestConfig ks
           : ksqlRestConfig.getClientAuth();
 
       final String alias = ksqlRestConfig.getString(ksConfigName);
-      setTlsOptions(ksqlRestConfig, options, alias, clientAuth);
+      ApiServerUtils.setTlsOptions(ksqlRestConfig, options, alias, clientAuth);
     }
     return options;
   }
 
-  private static void setTlsOptions(
```
rodesai, line 319: All this code was moved over into a common utils class (ApiServerUtils)

ksqldb-rest-app/src/test/java/io/confluent/ksql/rest/server/KsqlRestApplicationTest.java
```diff
@@ -354,51 +347,6 @@ public void shouldSendCreateStreamRequestBeforeSettingReady() {
     inOrder.verify(serverState).setReady();
   }
 
-  @Test
```
rodesai, line 357: As we no longer check preconditions from the rest application, we no longer need these tests.

ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/PreconditionChecker.java
```diff
@@ -0,0 +1,220 @@
...
+        Integer.MAX_VALUE,
+        1000,
+        30000,
+        this::checkPreconditions,
+        terminatedFuture::isDone,
+        predicates
+    );
+  }
+
+  private void checkPreconditions() {
```
stevenpyzhang, line 160: Could you add a comment on how this function continuously runs and determines if PreconditionServer will continue running or not? And if it passes, we'll terminate the pre condition server and move onto the main KSQL server.

stevenpyzhang (commented):
Overall makes sense to me. I think we could use some more comments for clarity on the new precondition checker executable and server.

Also, is it not possible to rework PreconditionFunctionalTest.java for this new model? We should probably have some integration test for this.

stevenpyzhang (approved):
lgtm

PR #9035, opened 2022-04-14
feat: add metric that's emitted when processing log emits an error

stevenpyzhang:
### Description 
Every processing logger now emits metrics (both push and pull). Going to add clean up of the push query processing logger metrics in a future pr.

I ended up passing the Metrics and the value of the custom metrics tags config through ProcessingLogContext and ProcessingLoggerFactory, that way, the only thing the factory get logger function needs is the logger name, and maybe also some additional metrics tags.

The processing logger factory now creates the metric and the sensor. To prevent duplicate metrics from being created, we're synchronizing the the method that creates the logger.

### Testing done 
Unit test and integration test added

Created these streams
```
create stream tests(f BIGINT)  with (KAFKA_TOPIC='hello', VALUE_FORMAT='json', partitions=1);
create stream test_addition_20 as select f+10 from tests;
```

JMX beans: 
```
io.confluent.ksql.metrics:type=processing-diagnostic-metrics,query-id=CSAS_TEST_ADDITION_20_7,loggerid=CSAS_TEST_ADDITION_20_7
io.confluent.ksql.metrics:type=processing-diagnostic-metrics,query-id=CSAS_TEST_ADDITION_20_7,logger-id=CSAS_TEST_ADDITION_20_7.KsqlTopic.Source
io.confluent.ksql.metrics:type=processing-diagnostic-metrics,query-id=CSAS_TEST_ADDITION_20_7,logger-id=CSAS_TEST_ADDITION_20_7.KsqlTopic.Source.deserializer
io.confluent.ksql.metrics:type=processing-diagnostic-metrics,query-id=CSAS_TEST_ADDITION_20_7,logger-id=CSAS_TEST_ADDITION_20_7.KsqlTopic.Source.serializer
io.confluent.ksql.metrics:type=processing-diagnostic-metrics,query-id=CSAS_TEST_ADDITION_20_7,logger-id=CSAS_TEST_ADDITION_20_7.Project
io.confluent.ksql.metrics:type=processing-diagnostic-metrics,query-id=CSAS_TEST_ADDITION_20_7,logger-id=CSAS_TEST_ADDITION_20_7.TEST_ADDITION_20.deserializer
io.confluent.ksql.metrics:type=processing-diagnostic-metrics,query-id=CSAS_TEST_ADDITION_20_7,logger-id=CSAS_TEST_ADDITION_20_7.TEST_ADDITION_20.serializer
io.confluent.ksql.metrics:type=processing-diagnostic-metrics,query-id=CSAS_TEST_ADDITION_20_7,logger-id=CSAS_TEST_ADDITION_20_7.ksql.logger.thread.exception.uncaught
```

Used local console producer to produce these records
```
>{"f":null}
>{"f":"hello"}   
>{"f":5
```
Verified that Project metric had 1 value and source deserializer had 2 value

### Reviewer checklist


ksqldb-common/src/main/java/io/confluent/ksql/logging/processing/ProcessingLoggerFactory.java
```diff
@@ -25,6 +27,16 @@ public interface ProcessingLoggerFactory {
...
 
+  /**
+   * Get a processing logger for writing record processing log messages. This processing logger
+   * also emits metrics
+   * @param name The name of the logger to get.
+   * @param metrics The metrics that are emitted.
+   * @param customMetricsTags A map of metrics tags.
+   * @return The logger with the given name that emits metrics.
+   */
+  ProcessingLogger getLoggerWithMetrics(String name, Metrics metrics, Map<String, String> customMetricsTags);
```
rodesai, line 38: Can we pass the `Metrics` object into the logger factory constructor? This way you don't need to pass it into every call that creates a logger. And, you can also use metrics from logs returned by `getLogger` - they just wouldn't have all the tags.

ksqldb-common/src/main/java/io/confluent/ksql/logging/processing/ProcessingLoggerImplWithMetrics.java
```diff
@@ -0,0 +1,61 @@
...
+ */
+
+package io.confluent.ksql.logging.processing;
+
+import java.util.Map;
+import org.apache.kafka.common.metrics.Metrics;
+import org.apache.kafka.common.metrics.Sensor;
+import org.apache.kafka.common.metrics.stats.CumulativeSum;
+
+public class ProcessingLoggerImplWithMetrics extends ProcessingLoggerImpl {
```
rodesai, line 23: Let's model this as a composition relationship rather than inheritance:
```
interface MeteredProcessinLogger implements ProcessingLogger {
    ProcessingLogger inner;

    @Override
    public void error(final ErrorMessage msg) {
        totalProcessingErrors.record(0.5)
        inner.error(msg);
    }
}
```

ksqldb-common/src/main/java/io/confluent/ksql/logging/processing/ProcessingLoggerWithMetricsInstantiator.java
```diff
@@ -0,0 +1,83 @@
...
+
+import static java.util.Objects.requireNonNull;
+
+import edu.umd.cs.findbugs.annotations.SuppressFBWarnings;
+import io.confluent.common.logging.StructuredLogger;
+import java.util.Map;
+import java.util.Objects;
+import org.apache.kafka.common.metrics.Metrics;
+
+public class ProcessingLoggerWithMetricsInstantiator {
```
rodesai, line 26: what's this class for? It seems to be just used in the implementation of `getLoggerWithMetrics`

stevenpyzhang: I used it as a wrapper class for everything needed for a `ProcessingLoggerImplWithMetrics`. It's so I can use a function type for `loggerWithMetricsFactory` in `ProcessingLoggerFactoryImpl`, otherwise I'd need need some special QuadFunction interface

stevenpyzhang: Ok I think, I can get away with a TriFunction after looking at some of your other suggestions, I can get rid of this class.

ksqldb-common/src/main/java/io/confluent/ksql/logging/processing/MeteredProcessingLogger.java
```diff
@@ -0,0 +1,80 @@
...
+import java.util.Map;
+import java.util.Objects;
+import org.apache.kafka.common.MetricName;
+import org.apache.kafka.common.metrics.KafkaMetric;
+import org.apache.kafka.common.metrics.MetricConfig;
+import org.apache.kafka.common.metrics.Metrics;
+import org.apache.kafka.common.metrics.stats.CumulativeSum;
+
+public class MeteredProcessingLogger implements ProcessingLogger {
+  public static final String PROCESSING_LOG_ERROR_METRIC_NAME = "processing-log-errors-sum";
```
rodesai, line 28: nit: let's just name this `processing-error-total`

ksqldb-common/src/main/java/io/confluent/ksql/logging/processing/ProcessingLoggerFactory.java
stevenpyzhang: done

ksqldb-common/src/main/java/io/confluent/ksql/logging/processing/ProcessingLoggerImplWithMetrics.java
stevenpyzhang: done

ksqldb-common/src/main/java/io/confluent/ksql/util/MetricsTagsUtil.java
```diff
@@ -0,0 +1,34 @@
...
+
+  private MetricsTagsUtil() {}
+
+  public static Map<String, String> getCustomMetricsTagsForQuery(
+      final String id,
+      final KsqlConfig config
+  ) {
+    final Map<String, String> customMetricsTags =
+        new HashMap<>(config.getStringAsMap(KsqlConfig.KSQL_CUSTOM_METRICS_TAGS));
+    customMetricsTags.put("query-id", id);
```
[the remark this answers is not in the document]
stevenpyzhang: Sure I can change it, but I looked around and saw that in https://github.com/confluentinc/ksql/pull/7955/files we used `query-id` as the tag in the saturation metric.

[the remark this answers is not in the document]
rodesai: Ah my bad. I forgot the KsqlProvider converts - to _ this automatically.

ksqldb-common/src/main/java/io/confluent/ksql/logging/processing/MeteredProcessingLogger.java
```diff
@@ -0,0 +1,80 @@
...
+        PROCESSING_LOG_METRICS_GROUP_NAME,
+        PROCESSING_LOG_METRIC_DESCRIPTION,
+        metricsTags);
+
+    final KafkaMetric metric = metrics.metric(errorMetric);
+
+    // If metric doesn't exist, add the metric. If it already exists, grab the CumulativeSum
+    // measurable so that it can be used in this particular processing logger instance.
+    if (metric == null) {
+      metrics.addMetric(errorMetric, sum);
```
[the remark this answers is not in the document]
guozhangwang: +1. The metrics registry assumes the caller to synchronize the access itself.

I think we can also just synchronize the whole block starting at line 69. I've also filed an AK ticket to improve the pattern: https://issues.apache.org/jira/browse/KAFKA-13846

ksqldb-common/src/main/java/io/confluent/ksql/logging/processing/MeteredProcessingLogger.java
```diff
@@ -0,0 +1,80 @@
...
+import org.apache.kafka.common.metrics.stats.CumulativeSum;
+
+public class MeteredProcessingLogger implements ProcessingLogger {
+  public static final String PROCESSING_LOG_ERROR_METRIC_NAME = "processing-error-total";
+  public static final String PROCESSING_LOG_METRICS_GROUP_NAME = "processing-log-metrics";
+  public static final String PROCESSING_LOG_METRIC_DESCRIPTION =
+      "The total number of errors emitted by the processing log.";
+
+  private final ProcessingLogger logger;
+  private final CumulativeSum sumOfErrors;
```
guozhangwang, line 34: Just checking do we really want to have a cumulative sum (it is ever growing) or a windowed sum?

stevenpyzhang: I think we want cumulative sum here because the actual metric we're trying to get is a delta metric, and delta metrics are automatically calculated for cumulative sum ones

stevenpyzhang:
I'm actually working on an alternative way of doing this. My thought now is to create a sensor in QueryBuilder that gets stored in the QueryMetadataImpl so that it can be cleaned up later on. We'd then pass this sensor everytime we get a processing logger from ProcessingLoggerFactory. The factory method would take a sensor object, so we'll have a single sensor shared amongst all the MeteredProcessingLoggers that records the metric.

ksqldb-engine/src/main/java/io/confluent/ksql/engine/EngineContext.java
```diff
@@ -132,7 +132,7 @@ synchronized EngineContext createSandbox(final ServiceContext serviceContext) {
     this.runtimeAssignor.rebuildAssignment(queryRegistry.getPersistentQueries().values());
     return new EngineContext(
         SandboxedServiceContext.create(serviceContext),
-        processingLogContext,
+        ProcessingLogContext.create(),
```
stevenpyzhang, line 135: Need to use the empty create or otherwise we're registering metrics when sandbox executing queries

ksqldb-common/src/main/java/io/confluent/ksql/logging/processing/ProcessingLoggerFactoryImpl.java
```diff
@@ -18,36 +18,73 @@
...
+        customMetricsTags
+    );
+  }
+
   @Override
   public Collection<String> getLoggers() {
     return innerFactory.getLoggers();
   }
+
+  interface TriFunction<A,B,C,R> {
```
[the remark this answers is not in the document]
stevenpyzhang: this is gone now

ksqldb-common/src/main/java/io/confluent/ksql/logging/processing/ProcessingLoggerFactoryImpl.java
```diff
@@ -18,36 +18,123 @@
...
     this.loggerFactory = loggerFactory;
+    this.loggerWithMetricsFactory = loggerWithMetricsFactory;
+    this.metricsTags = metricsTags;
+    this.processingLoggers = new ConcurrentHashMap<>();
   }
 
   @Override
-  public ProcessingLogger getLogger(final String name) {
-    return loggerFactory.apply(config, innerFactory.getLogger(name));
+  public ProcessingLogger getLoggerWithMetrics(
```
rodesai, line 83: nit: I don't think these names need the `withMetrics` suffix

ksqldb-common/src/main/java/io/confluent/ksql/logging/processing/ProcessingLoggerFactoryImpl.java
```diff
@@ -18,36 +18,123 @@
...
+  ) {
+    return getLoggerWithMetrics(name, "");
+  }
+
+  @Override
+  public ProcessingLogger getLoggerWithMetrics(
+      final String name,
+      final String queryId
+  ) {
+    if (processingLoggers.containsKey(name)) {
```
rodesai, line 94: Use computeIfAbsent here, or synchronize this method. There is still a race condition here where 2 threads can wind up creating a logger for a given name if they hit line 94 concurrently.

ksqldb-common/src/main/java/io/confluent/ksql/util/MetricsTagsUtil.java
```diff
@@ -0,0 +1,47 @@
...
+    if (queryId.equals("")) {
+      return tags;
+    }
+    final Map<String, String> newMetricsTags =
+        new HashMap<>(tags);
+    newMetricsTags.put("query-id", queryId);
+    return newMetricsTags;
+  }
+
+  public static Map<String, String> getMetricsTagsWithLoggerId(
```
rodesai, line 38: can you factor out this bit into a common method:
```
final Map<String, String> newMetricsTags =
        new HashMap<>(tags);
    newMetricsTags.put(tagName, tagValue);
    return newMetricsTags;
```

ksqldb-common/src/main/java/io/confluent/ksql/util/MetricsTagsUtil.java
```diff
@@ -0,0 +1,47 @@
...
+      final String queryId,
+      final Map<String, String> tags
+  ) {
+    if (queryId.equals("")) {
+      return tags;
+    }
+    final Map<String, String> newMetricsTags =
+        new HashMap<>(tags);
+    newMetricsTags.put("query-id", queryId);
+    return newMetricsTags;
```
rodesai, line 35: return an immutable map here

ksqldb-common/src/main/java/io/confluent/ksql/logging/processing/ProcessingLoggerFactory.java
```diff
@@ -19,11 +19,21 @@
...
+  ProcessingLogger getLoggerWithMetrics(String name);
+
+  /**
+   * Get a processing logger for writing record processing log messages. This processing logger
+   * also emits metrics
+   * @param name The name of the logger to get.
+   * @param queryId The queryId to tag the metrics with
+   * @return The logger with the given name that emits metrics.
+   */
+  ProcessingLogger getLoggerWithMetrics(String name, String queryId);
```
rodesai, line 36: Lets leave the queryId out of this interface and make it more generic, like:

```
ProcessingLogger getLoggerWithMetrics(String name, Map<String, String> additionalMetricTags);
```

rodesai (commented):
Took a quick pass through. Structure looks good overall. Left a few comments inline.

ksqldb-common/src/main/java/io/confluent/ksql/logging/processing/ProcessingLoggerFactoryImpl.java
stevenpyzhang: Should we just change the interface for the factory from `public interface ProcessingLoggerFactory` to `public interface MeteredProcessingLoggerFactory`? Then we could drop the with metrics suffix. I think somewhere in the interface, it should indicate that these loggers are emitting metrics.

ksqldb-common/src/main/java/io/confluent/ksql/logging/processing/ProcessingLoggerFactory.java
stevenpyzhang: changed it to take a map

ksqldb-common/src/main/java/io/confluent/ksql/util/MetricsTagsUtil.java
stevenpyzhang: added a common method

ksqldb-common/src/main/java/io/confluent/ksql/logging/processing/ProcessingLoggerFactoryImpl.java
stevenpyzhang: I decided to go with adding the synchronized keyword

rodesai (approved):
LGTM once inline comments addressed

ksqldb-common/src/main/java/io/confluent/ksql/logging/processing/MeteredProcessingLoggerFactoryImpl.java
```diff
@@ -0,0 +1,142 @@
...
+          loggerWithMetricsFactory,
+      final Map<String, String> metricsTags
+  ) {
+    this.config = config;
+    this.innerFactory = innerFactory;
+    this.metrics = metrics;
+    this.loggerFactory = loggerFactory;
+    this.loggerWithMetricsFactory = loggerWithMetricsFactory;
+    this.metricsTags = metricsTags;
+    this.processingLoggers = new ConcurrentHashMap<>();
```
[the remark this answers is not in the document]
stevenpyzhang: ok changed it to regular hashmap

ksqldb-common/src/main/java/io/confluent/ksql/logging/processing/MeteredProcessingLoggerFactoryImpl.java
```diff
@@ -0,0 +1,142 @@
...
+        getProcessLogger(name),
+        errorSensor
+    );
+    processingLoggers.put(name, meteredProcessingLogger);
+    return meteredProcessingLogger;
+  }
+
+  @Override
+  public Collection<String> getLoggers() {
+    return innerFactory.getLoggers();
```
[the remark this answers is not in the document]
stevenpyzhang: ok changed it to the map from class instead of the inner factory function and also synchronized

PR #9036, opened 2022-04-14
fix: re fetch streams for each materializationProviderBuilder

wcarlson5:
### Description 
We need to re fetch the streams instance when providing the Materialization incase it was updated when changing a config.

### Testing done 
pull queries after an `ALTER SYSTEM` do not return an error with gen 2 queries

### Reviewer checklist

### DIFF
--- a/ksqldb-engine/src/main/java/io/confluent/ksql/execution/pull/HARouting.java
+++ b/ksqldb-engine/src/main/java/io/confluent/ksql/execution/pull/HARouting.java
@@ -132,7 +132,14 @@ public final class HARouting implements AutoCloseable {
         .collect(Collectors.toMap(
             KsqlPartitionLocation::getPartition,
             loc -> loc.getNodes().stream().map(KsqlNode::getHost).collect(Collectors.toList())));
-
+    if (allLocations.isEmpty()) {
+      final MaterializationException materializationException = new MaterializationException(
+          "Unable to execute pull query as the locations are not found."
+              + " ALTER SYSTEM may not have finished."
+              + " Please wait a few minutes and try again.");
+      LOG.debug(materializationException.getMessage());
+      throw materializationException;
+    }
     if (!emptyPartitions.isEmpty()) {
       final MaterializationException materializationException = new MaterializationException(
           "Unable to execute pull query. "
--- a/ksqldb-engine/src/main/java/io/confluent/ksql/execution/pull/HARouting.java
+++ b/ksqldb-engine/src/main/java/io/confluent/ksql/execution/pull/HARouting.java
@@ -132,14 +132,7 @@ public final class HARouting implements AutoCloseable {
         .collect(Collectors.toMap(
             KsqlPartitionLocation::getPartition,
             loc -> loc.getNodes().stream().map(KsqlNode::getHost).collect(Collectors.toList())));
-    if (allLocations.isEmpty()) {
-      final MaterializationException materializationException = new MaterializationException(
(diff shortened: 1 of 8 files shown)


ksqldb-engine/src/main/java/io/confluent/ksql/util/BinPackedPersistentQueryMetadataImpl.java
```diff
@@ -217,7 +209,11 @@ public ProcessingLogger getProcessingLogger() {
   public Optional<Materialization> getMaterialization(
       final QueryId queryId,
       final QueryContext.Stacker contextStacker) {
-    return materializationProvider.map(builder -> builder.build(queryId, contextStacker));
+    return materializationProviderBuilder
+            .flatMap(builder -> builder.apply(
+                    this.sharedKafkaStreamsRuntime.getKafkaStreams(),
+                    topology
+            )).map(builder -> builder.build(queryId, contextStacker));
```
jnh5y, line 216: Is it possible to have a unit test for this issue?  (Asking naively)

ableegoldman (approved):
To the extent I understand the materialization provider code, this LGTM

ksqldb-engine/src/main/java/io/confluent/ksql/util/BinPackedPersistentQueryMetadataImpl.java
wcarlson5: Yes just finished adding one :)

wcarlson5:
@agavra Thanks for the help looking into the issue. I dug a bit deeper and it doesn't seem to be the KS state but the locator is not finding the partitions (as I think you suggested).

For now I just added a check to see if it was finding no positions with a possible explanation. Can you give it a quick look and see if you have any suggestions?

ksqldb-engine/src/main/java/io/confluent/ksql/util/SharedKafkaStreamsRuntimeImpl.java
```diff
@@ -247,6 +248,8 @@ public void start(final QueryId queryId) {
 
   @Override
   public void overrideStreamsProperties(final Map<String, Object> newProperties) {
+    newProperties.put(StreamsConfig.APPLICATION_SERVER_CONFIG,
+        streamsProperties.get(StreamsConfig.APPLICATION_SERVER_CONFIG));
```
agavra, line 252: this could benefit from a comment, what could cause `newProperties` to have a different `APPLITCATION_SERVER_CONFIG` value and can we just prevent that from happening in first place? this seems like it silently overwrites something that the user wanted to do and can cause some confusion

ksqldb-engine/src/main/java/io/confluent/ksql/execution/pull/HARouting.java
```diff
@@ -132,7 +132,14 @@ public CompletableFuture<Void> handlePullQuery(
...
             KsqlPartitionLocation::getPartition,
             loc -> loc.getNodes().stream().map(KsqlNode::getHost).collect(Collectors.toList())));
-
+    if (allLocations.isEmpty()) {
+      final MaterializationException materializationException = new MaterializationException(
+          "Unable to execute pull query as the locations are not found."
+              + " ALTER SYSTEM may not have finished."
+              + " Please wait a few minutes and try again.");
+      LOG.debug(materializationException.getMessage());
+      throw materializationException;
```
agavra, line 141: I think this is also missing a test, which would be nice to have :)

PR #9038, opened 2022-04-15
fix: INSERT fails when serializing Proto/Avro nested Structs

spena:
### Description 
The INSERT/VALUES statement was failing to insert `struct()` values after the PRs (https://github.com/confluentinc/ksql/commit/75da9fd1e7dc9eae47203f3e40f167de7e3a6526)  that added support for multiple schema definitions. 

```
ksql> create stream s1(c1 struct<id int>) with (kafka_topic='s1', value_format='protobuf', partitions=1);

 Message        
----------------
 Stream created 
----------------

ksql> insert into s1(c1) values(struct(id := 1));
Failed to insert values into 'S1'. Could not serialize value: [ Struct{ID=1} ]. Error serializing message to topic: s1. Failed to access Protobuf data from topic s1 : Schema being registered is incompatible with an earlier schema for subject "s1-value"; error code: 409

```

The problem is caused when converting the ksql object to a compatible SR object. The conversion was not adding all the nested struct names which caused the SR serializer to not find a compatible SR schema. 

For instance, the following SR schema is registered:
```
syntax = "proto3";

message ConnectDefault1 {
  ConnectDefault2 C1 = 1;

  message ConnectDefault2 {
    int32 ID = 1;
  }
}
```

When a KSQL object is converted to a compatible SR protobuf object, the object ended with only the first `ConnectDefault1` schema embedded in it. See the `[ Struct{ID=1} ]` in the error message. This object should have had the `ConnectDefault2` name embedded.

This was not an issue before because KSQL was not converting to Protobuf compatible schema (it was not adding the schema names to the object). The SR API detects no schema names and start using default names, which start with `ConnectDefault`, so it used to work fine before.

This fix adds all schema names to the nested structs.

### Testing done 
Unit tests
Several manual tests (will add YATT tests in a follow-up change)

### Reviewer checklist


ksqldb-serde/src/main/java/io/confluent/ksql/serde/protobuf/ProtobufSerdeFactory.java
```diff
@@ -67,20 +81,22 @@ <T> Serde<T> createSerde(
       final boolean isKey) {
     validate(schema);
 
-    final Optional<Schema> physicalSchema = properties.getSchemaId().isPresent() ? Optional.of(
-        SerdeUtils.getAndTranslateSchema(srFactory, properties.getSchemaId()
-            .get(), new ProtobufSchemaTranslator(properties))) : Optional.empty();
+    final Optional<Schema> physicalSchema = getSRSchema(srFactory);
+
+    final ConnectSchema compatibleSchema = getSRSchema(srFactory)
```
lihaosky, line 86: Why do we need to call `getSRSchema` twice here?

ksqldb-serde/src/main/java/io/confluent/ksql/serde/protobuf/ProtobufSerdeFactory.java
```diff
@@ -145,18 +158,12 @@ private <T> KsqlConnectDeserializer<T> createDeserializer(
...
-                : new ConnectSRSchemaDataTranslator(value))
-        .orElseGet(() -> {
-          if (fullSchemaName.isPresent()) {
-            return new ProtobufDataTranslator(schema, fullSchemaName.get());
-          } else {
-            return new ConnectDataTranslator(schema);
-          }
-        });
+    // During deserialization, we use the original schema to translate to KSQL data.
+    if (isDeserializer || !physicalSchema.isPresent()) {
```
lihaosky, line 162: physicalSchema will always exist?

ksqldb-serde/src/main/java/io/confluent/ksql/serde/protobuf/ProtobufSerdeFactory.java
```diff
@@ -145,18 +158,12 @@ private <T> KsqlConnectDeserializer<T> createDeserializer(
 
   private DataTranslator getDataTranslator(final Schema schema,
       final Optional<Schema> physicalSchema, final boolean isDeserializer) {
-    // If physical schema exists, we use physical schema to translate to connect data. During
-    // deserialization, if physical schema exists, we use original schema to translate to ksql data.
-    return physicalSchema.<DataTranslator>map(
-            value -> isDeserializer ? new ConnectDataTranslator(schema)
-                : new ConnectSRSchemaDataTranslator(value))
```
lihaosky, line 152: Why not call `ConnectSRSchemaDataTranslator` for `schemaId` case? We don't want to translate `physcialSchema` obtained from `schemaId` to compatible one and use that to do serialization. We want to use raw `physcialSchema` to do serialization if `schemaId` is provided.

ksqldb-serde/src/main/java/io/confluent/ksql/serde/protobuf/ProtobufSchemas.java
```diff
@@ -23,12 +25,119 @@ public final class ProtobufSchemas {
...
+    // Set name to match it with targetSchema
+    if (targetSchema.name() != null) {
+      schemaBuilder.name(targetSchema.name());
+    }
+
+    final Iterator<Field> targetIt = targetSchema.fields().iterator();
+
+    for (final Field f : schema.fields()) {
+      final Schema adjustedSchema;
+      if (targetIt.hasNext()) {
```
lihaosky, line 83: Can you check if `target` has same field name as `f`?

ksqldb-serde/src/main/java/io/confluent/ksql/serde/protobuf/ProtobufSerdeFactory.java
```diff
@@ -67,20 +81,22 @@ <T> Serde<T> createSerde(
       final boolean isKey) {
     validate(schema);
 
-    final Optional<Schema> physicalSchema = properties.getSchemaId().isPresent() ? Optional.of(
-        SerdeUtils.getAndTranslateSchema(srFactory, properties.getSchemaId()
-            .get(), new ProtobufSchemaTranslator(properties))) : Optional.empty();
+    final Optional<Schema> physicalSchema = getSRSchema(srFactory);
+
+    final ConnectSchema compatibleSchema = getSRSchema(srFactory)
+        .map(s -> (ConnectSchema) ProtobufSchemas.schemaCompatibleWithSR(schema, s))
```
lihaosky, line 87: With this change, why do we still need https://github.com/confluentinc/ksql/blob/master/ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/execution/InsertValuesExecutor.java#L519-L526?

unrelated: I don't know why github doesn't allow commenting on lines which is not changed in PR...

lihaosky (commented):
Ping @vcrfxia to also take a look if I couldn't finish it today.

ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/execution/InsertValuesExecutor.java
```diff
@@ -546,6 +546,16 @@ private static FormatInfo addSerializerMissingFormatFields(
           }
         }
       }
+
+      // If SCHEMA_ID was not added in the previous step, then use the SUBJECT_NAME
+      // Using a SUBJECT_NAME will help the serializer to identify which schema to fetch from SR
+      // but without using the restrictions we have with SCHEMA_ID
+      if (supportedProperties.contains(ConnectProperties.SUBJECT_NAME)) {
+        if (!formatInfo.getProperties().containsKey(ConnectProperties.SCHEMA_ID)) {
+          propertiesBuilder.put(ConnectProperties.SUBJECT_NAME,
```
lihaosky, line 555: Is `subject_name`'s purpose to distinguish if protobuf serde is called from `InsertValuesExecutor`? Not sure if there's other ways...

ksqldb-serde/src/main/java/io/confluent/ksql/serde/avro/AvroProperties.java
```diff
@@ -34,7 +34,8 @@ public class AvroProperties extends ConnectProperties {
 
   static final ImmutableSet<String> SUPPORTED_PROPERTIES = ImmutableSet.of(
       FULL_SCHEMA_NAME,
-      SCHEMA_ID
+      SCHEMA_ID,
+      SUBJECT_NAME
```
lihaosky, line 38: Why do we need it for avro? I don't see it's used in avro serde

ksqldb-serde/src/main/java/io/confluent/ksql/serde/protobuf/ProtobufSerdeFactory.java
spena: My mistake, I removed it.

ksqldb-serde/src/main/java/io/confluent/ksql/serde/protobuf/ProtobufSerdeFactory.java
spena: Yes, but I noticed there are places this does not exist yet. I will fix this after the regression issue  is fixed.

ksqldb-serde/src/main/java/io/confluent/ksql/serde/protobuf/ProtobufSerdeFactory.java
spena: The `ConnectSRSchemaDataTranslator` does not work either. I just made a fix and created a `ProtobufSRSchemaDataTranslator` for protobuf only. Once I understand all these formats, I will put everything into one class (in a follow-up PR)

ksqldb-serde/src/main/java/io/confluent/ksql/serde/protobuf/ProtobufSerdeFactory.java
spena: We don't need it anymore. But I am getting other code refactors I'd like to keep away from this so I can fix the regression.

ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/execution/InsertValuesExecutor.java
spena: Yes. That's the only reason I added this. The other way was to pass the `topicName` when creating the Serde, but that causes too many refactor classes and test classes. I ended up just passing this info on the FormatInfo.

ksqldb-serde/src/main/java/io/confluent/ksql/serde/avro/AvroProperties.java
spena: Removed. I might use it later, but at least the regression issued is fixed.

ksqldb-functional-tests/src/test/resources/query-validation-tests/insert-into.json
```diff
@@ -137,6 +137,42 @@
         {"topic": "sink", "key": "0", "value": {"A": 789, "B": "turtle"}, "timestamp": 0}
       ]
     },
+    {
+      "name": "convert formats: JSON to PROTOBUF",
```
vcrfxia, line 141: Do these tests fail prior to the changes in this PR? I don't see them using nested structs so I'm not sure how they're related to the bugfix in this PR.

ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/execution/InsertValuesExecutor.java
```diff
@@ -499,56 +495,35 @@ private byte[] serializeValue(
...
 
-    // The FULL_SCHEMA_NAME allows the serializer to choose the schema definition
-    if (supportedProperties.contains(ConnectProperties.FULL_SCHEMA_NAME)) {
-      if (!formatInfo.getProperties().containsKey(ConnectProperties.FULL_SCHEMA_NAME)) {
-        SchemaRegistryUtil.getLatestParsedSchema(srClient, topicName, isKey).map(ParsedSchema::name)
-            .ifPresent(schemaName ->
-                propertiesBuilder.put(ConnectProperties.FULL_SCHEMA_NAME, schemaName));
-      }
-    }
+        propertiesBuilder.put(ConnectProperties.SUBJECT_NAME,
```
vcrfxia, line 519: Why do we need to pass the subject name here? It looks like downstream it's used in order to get a schema ID -- why can't we just get the schema ID from the subject name here and then pass the schema ID, in order to avoid introducing a new property?

ksqldb-serde/src/main/java/io/confluent/ksql/serde/protobuf/ProtobufSchemaTranslator.java
```diff
@@ -34,25 +34,46 @@
...
+        // This flag is needed so that the schema translation in toConnectRow() adds the
+        // package name information to the row schema
+        ProtobufDataConfig.ENHANCED_PROTOBUF_SCHEMA_SUPPORT_CONFIG, "true"
     );
+    this.fullNameSchema = ignoreDefaultName(properties.getFullSchemaName());
     this.updatedConfigs = baseConfigs;
     this.protobufData = new ProtobufData(new ProtobufDataConfig(baseConfigs));
   }
 
+  private Optional<String> ignoreDefaultName(final String fullNameSchema) {
```
vcrfxia, line 60: Where does this schema name get populated? I'm wondering if we can avoid adding it, rather than adding it and then ignoring it if it's the default.

ksqldb-serde/src/main/java/io/confluent/ksql/serde/protobuf/ProtobufSchemas.java
```diff
@@ -23,12 +23,21 @@ public final class ProtobufSchemas {
   private ProtobufSchemas() {
   }
 
+  /**
+   * Adds the {@code schemaName} to the provided {@code schema}. The name is only added to the
+   * parent schema name to allow Protobuf to extract the name later. Support for nested schema
```
vcrfxia, line 28: What does it mean that nested schema extractions are not supported? (What is a nested schema extraction?)

ksqldb-serde/src/main/java/io/confluent/ksql/serde/protobuf/ProtobufSchemas.java
```diff
@@ -23,12 +23,21 @@ public final class ProtobufSchemas {
...
+   * extractions are not supported.
+   */
   static Schema schemaWithName(final Schema schema, final String schemaName) {
-    if (schemaName == null || schema.type() != Schema.Type.STRUCT) {
```
vcrfxia, line 27: Why is this being removed?

ksqldb-serde/src/main/java/io/confluent/ksql/serde/protobuf/ProtobufSerdeFactory.java
```diff
@@ -67,15 +115,29 @@ <T> Serde<T> createSerde(
...
+    // If a SR schema exists (physical schema), then fetch it and use it for the serializer.
+    // The schema ID is fetched along the physical schema to be used in case multiple schema
+    // definitions are found, which will define how Connect serializer will work (see
+    // the getConverter() class in this file)
+    final Optional<SchemaAndId> schemaAndId = getSchemaFromSR(srFactory);
+    final Optional<Schema> physicalSchema = schemaAndId.map(SchemaAndId::toConnectSchema);
+
+    // The schemaId may be used when creating the ProtobufConverter for the serializer.
+    // It should be used only if SCHEMA_ID was specified manually or if the SR schema has
+    // multiple schema definitions.
```
vcrfxia, line 127: What's the reason for using schema ID if the schema has multiple schema definitions? Is it because each schema definition within the schema has its own schema ID?

vcrfxia (commented):
Hey @spena -- thanks for the additional context around these changes. I'm having trouble wrapping my head around the fix, still. Left some questions inline. I'm worried that this part of the code is only going to become harder and harder to reason about as more functionality is added and we need to work around the Connect serdes in more and more ways, and that we're likely to continue finding more bugs/regressions related to this since the relevant test coverage is hard to add.

If this is urgent, feel free to merge as long as you feel good about these changes and test coverage. I saw in your discussions with Hao that you have plans to refactor this code in a number of ways after this PR.

ksqldb-functional-tests/src/test/resources/query-validation-tests/insert-into.json
spena: I just added these tests to have some INSERT INTO coverage with protobuf. I was running INSERT INTO manually, so wanted to automated them.

ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/execution/InsertValuesExecutor.java
spena: It was like you suggests in a another PR, but Hao recommended to not inject the SCHEMA_ID if it was not explicitly specified in the WITH properties. The reason is that in the ProtobufSerdeFactory, the SCHEMA_ID is used to configure the converter (See https://github.com/confluentinc/ksql/blob/master/ksqldb-serde/src/main/java/io/confluent/ksql/serde/protobuf/ProtobufSerdeFactory.java#L179). The serialization then takes a different path internally (in the SR API) when that config is set, which causes some corner cases to fail during INSERT VALUES. 

For instance, the 'auto.register.schemas' is disabled when SCHEMA_ID is present. If we set this ID on any case, then the auto.register.schemas will be false for any case, which will not register a missing schema in SR during INSERT VALUES. This is an issue I raised, but we haven't discussed if we want to keep it or not. So I ended up passing a SUBJECT_NAME which will keep the current behavior. It was that, or a boolean value in the Format class to say `do not set the ID in the converter config`. I don't like this SUBJECT_NAME approach, but it was the change with less refactor code. The other option was to pass the topic name on the createSerde() so it creates the subject name in the serde, but that refactor causes too many files to change.

ksqldb-serde/src/main/java/io/confluent/ksql/serde/protobuf/ProtobufSchemaTranslator.java
spena: It's like in 5 places. Two main sources are what this PR does when fetching the SR schema and create the schema translator based on the SR schema name. The other main one is when a user specifies the full schema name explicitly in the WITH properties. There are other places I don't know yet when they are called. So, to make it less prone to errors or miss other places, I just ignore it in the translator. It's just a property that the translator will ignore.

ksqldb-serde/src/main/java/io/confluent/ksql/serde/protobuf/ProtobufSchemas.java
spena: I should have added a better comment :). But multiple schemas definitions may be available in SR, like:
```
syntax = "proto3";

message Key1 {
  int32 ID = 1;
}

message Key2 {
  string ID = 1;
}

message Key3 {
  double ID = 1;
}
```

If a schema name `Key2` is specified, then the SR API will extract the following schema from the above:
```
syntax = "proto3";

message Key2 {
  string ID = 1;
}
```

The comment mentions this is only allowed for a parent schema name like the above examples. A nested schema name (not supported) would be `Key2.InternalKey`, like below:
```
syntax = "proto3";

message Key2 {
  string ID = 1;
  InternalKey IK = 2;

  message InternalKey { ... }
}
```

ksqldb-serde/src/main/java/io/confluent/ksql/serde/protobuf/ProtobufSchemas.java
spena: Just added back.

ksqldb-serde/src/main/java/io/confluent/ksql/serde/protobuf/ProtobufSerdeFactory.java
spena: This is only to activate the config used in the converter. The config I mentioned here (https://github.com/confluentinc/ksql/pull/9038#discussion_r853032771). I need to disable the auto.register.schemas and also make the SR API take the path when the schema id is present. 

I need a better comment.

ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/execution/InsertValuesExecutor.java
vcrfxia: > If we set this ID on any case, then the auto.register.schemas will be false for any case, which will not register a missing schema in SR during INSERT VALUES. This is an issue I raised, but we haven't discussed if we want to keep it or not.

My vote is to remove the behavior (i.e., I think we should disable auto-registering of schemas) if it means our code can be a lot cleaner, which it sounds like it does.

This SUBJECT_NAME approach doesn't feel right to me since subject name doesn't seem like it should be considered a format property (I don't think schema ID really is, either, but we already had this conversation when schema ID was first introduced), and it's confusing that subject name is only used as an internal format property (users can't set it) whereas most of the other properties can be set by users.

Is the strategy for cleaning up this code (assuming we get rid of auto registering schemas) clear to you? I'm fine with this SUBJECT_NAME approach as a short-term solution if so. I really don't think we should keep this on a medium-term, though. This code is really hard to reason about.

Can we document (in code comments) that the SUBJECT_NAME property is internal-only, and is safe to remove in the future because it's only ever added during INSERT VALUES which means it's never written to the command topic? If it's protobuf-only (looks like it is), we can also move it just to the ProtobufProperties class instead of ConnectProperties so that it's even less likely someone will accidentally wire it into somewhere we don't want it.

vcrfxia (approved):
Hey @spena -- thanks for the detailed explanations to my questions and the additional test coverage. As mentioned inline, I'm not a fan of the approach in this PR but I understand there's pressure to get this merged ASAP in order to avoid introducing a regression in the upcoming release. I'm fine with the PR being merged as a short-term stop gap but it'd be great to have concrete thoughts on how to refactor this code to make it easier to reason about. If we don't have a github issue for this refactor already, it'd be good to start one and record thoughts there. My vote is to prioritize the cleanup since these one-off fixes and regressions are only going to become more common until we do. WDYT?

PR #9040, opened 2022-04-15
fix: use the engine's KsqlConfig to build queries

rodesai:
This patch changes InteractiveStatementExecutor and QueryExecutor
to use the engine's KsqlConfig reference to build queries. With the
introduction of ALTER SYSTEM, the engine is now the source of truth
for the server's configuration - and so component's that need an
up-to-date config (anything that builds a continuous query) should
use the engine's reference.

This patch also moves the implementation of configure to the engine
from the executors. We need to have the listener property set in
the engine's config so that IQ still works correctly.

Finally, update usages of the engine to use the KsqlExecutionContext
interface rathe than the engine directly (so that we don't give
the executors access to `configure`)

### Description 

### Testing done 

### Reviewer checklist


wcarlson5 (approved):
LGTM with a green build

PR #9044, opened 2022-04-20
Update Jenkinsfile

tolgadur:
### Description 
This PR increases the timeout duration. Currently it is set to 4 hours but our build took longer and this was leading to weird error messages such as `This job is using over 50GB of disk space. Please reduce!`.

### Reviewer checklist

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -2,7 +2,7 @@
 
 common {
     slackChannel = '#ksqldb-quality-oncall'
-    timeoutHours = 4
+    timeoutHours = 6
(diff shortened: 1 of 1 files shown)


nateab (approved):

vcrfxia:
Hey @tolgadur , I'm not sure that increasing the build timeout here makes sense. Normal builds shouldn't take more than 4 hours -- if a build hits the timeout, in the past it's been because a test was hung and the build wasn't going to pass with more time anyway. Unless that's changed, I think we should keep the lower timeout in order to allow failed builds to fail sooner.

PR #9050, opened 2022-04-22
chore: bump jackson-databind version

lct45:
### Description 
There's a bug in jackson-databind before 2.13.0, see details [here](https://nvd.nist.gov/vuln/detail/CVE-2020-36518), so we're bumping the version
### Reviewer checklist


ksqldb-rest-model/pom.xml
```diff
@@ -50,6 +50,7 @@
         <dependency>
             <groupId>com.fasterxml.jackson.core</groupId>
             <artifactId>jackson-databind</artifactId>
+            <version>2.13.2.2</version>
```
lct45, line 53: I considered using `${jackson.version}` here which we use in a few other poms like the `ksqldb-api-client` but those are for the group `com.fasterxml.jackson.datatype` so I wasn't sure if we wanted these two groups to always have to have the same version, although its possible that implicitly this happened before I added the version field

agavra: can we introduce a property in the parent pom for this version? (e.g. `jackson.databind.version`)

lct45: @agavra is that the proper way to do it instead of using the variable from `common`? I can do that

agavra: defining a new variable gives us the flexibility to have different versions from `jackson.version`. Idk if that's necessary or not

lct45:
@agavra for any branches before master that need to be fixes I can just cherry-pick this commit and push, is that what you've done in the past?

ksqldb-rest-model/pom.xml
lct45: Cool, thanks! Just updated swapped to a variable in parent pom

agavra:
@lct45 - generally I target the earliest branch and then pint merge forward, but cherry-picking to an earlier branch and then pint merging works as well

agavra (approved):

lct45:
With the version bump in common, we don't need to make this change anymore https://github.com/confluentinc/common/pull/446

PR #9052, opened 2022-04-22
fix: sum udf change exception type & classify as USER error

bvarghese1:
### Description 
Throws KsqlFunctionException instead of ArithmeticException when there is an overflow for a
DECIMAL field using a SUM function. Additionally, it classifies KSqlFunctionException as a USER
error.

### Testing done 
Added unit and integration tests

### Reviewer checklist

### DIFF
--- a/ksqldb-common/src/main/java/io/confluent/ksql/util/DecimalUtil.java
+++ b/ksqldb-common/src/main/java/io/confluent/ksql/util/DecimalUtil.java
@@ -17,6 +17,7 @@ package io.confluent.ksql.util;
 
 import static io.confluent.ksql.schema.ksql.types.SqlDecimal.validateParameters;
 
+import io.confluent.ksql.function.KsqlFunctionException;
 import io.confluent.ksql.schema.ksql.types.SqlBaseType;
 import io.confluent.ksql.schema.ksql.types.SqlDecimal;
 import io.confluent.ksql.schema.ksql.types.SqlType;
@@ -258,7 +259,7 @@ public final class DecimalUtil {
     final int digits = precision - scale;
     final BigDecimal maxValue = BigDecimal.valueOf(Math.pow(10, digits));
     if (maxValue.compareTo(value.abs()) < 1) {
-      throw new ArithmeticException(
+      throw new KsqlFunctionException(
           String.format("Numeric field overflow: A field with precision %d and scale %d "
               + "must round to an absolute value less than 10^%d. Got %s",
               precision,
--- a/ksqldb-engine/src/main/java/io/confluent/ksql/query/KSqlFunctionClassifier.java
+++ b/ksqldb-engine/src/main/java/io/confluent/ksql/query/KSqlFunctionClassifier.java
@@ -1,0 +1,58 @@
+/*
+ * Copyright 2022 Confluent Inc.
+ *
+ * Licensed under the Confluent Community License (the "License"; you may not use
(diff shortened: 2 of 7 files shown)


ksqldb-common/src/main/java/io/confluent/ksql/util/DecimalUtil.java
```diff
@@ -258,7 +259,7 @@ private static void ensureMax(final BigDecimal value, final int precision, final
     final int digits = precision - scale;
     final BigDecimal maxValue = BigDecimal.valueOf(Math.pow(10, digits));
     if (maxValue.compareTo(value.abs()) < 1) {
-      throw new ArithmeticException(
```
spena, line 261: This ensureMax is an utility method that I think is called in different places other than functions, not just functions. In those cases, then KsqlFunctionException won't be a valid exception.

ksqldb-engine/src/main/java/io/confluent/ksql/query/KSqlFunctionClassifier.java
```diff
@@ -0,0 +1,58 @@
...
+  public KSqlFunctionClassifier(final String queryId) {
+    this.queryId = Objects.requireNonNull(queryId, "queryId");
+  }
+
+  @Override
+  public Type classify(final Throwable e) {
+    Type type = Type.UNKNOWN;
+    if (e instanceof KsqlFunctionException
+        || e instanceof StreamsException
+        && ExceptionUtils.getRootCause(e) instanceof KsqlFunctionException) {
```
jnh5y, line 44: I always get confused about the order of operations with `||` and `&&`.  I'd be tempted to add parens to make it completely clear.

bvarghese1: Sure, that makes sense. Will add parens to make it clear.

ksqldb-engine/src/main/java/io/confluent/ksql/query/KSqlFunctionClassifier.java
```diff
@@ -0,0 +1,58 @@
...
+
+  public KSqlFunctionClassifier(final String queryId) {
+    this.queryId = Objects.requireNonNull(queryId, "queryId");
+  }
+
+  @Override
+  public Type classify(final Throwable e) {
+    Type type = Type.UNKNOWN;
+    if (e instanceof KsqlFunctionException
+        || e instanceof StreamsException
```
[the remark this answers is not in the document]
bvarghese1: ```StreamsException``` is not being classified as a USER error. I will add parens to make that if condition easier to read. I have also added a test to validate the same.

ksqldb-engine/src/main/java/io/confluent/ksql/query/KSqlFunctionClassifier.java
bvarghese1: Sure, that makes sense. Will add parens to make it clear.

ksqldb-common/src/main/java/io/confluent/ksql/util/DecimalUtil.java
bvarghese1: @spena Thanks for pointing this out. After discussing with @jnh5y we have decided to revert this and make the change locally in the sum udaf.

PR #9055, opened 2022-04-24
chore: fix surefire caused build failures

stevenpyzhang:
### Description 

### Testing done 

### Reviewer checklist

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -156,7 +156,7 @@
         <netty-codec-http2-version>4.1.72.Final</netty-codec-http2-version>
         <jersey-common>2.34</jersey-common>
 
-        <maven.surefire.version>3.0.0-M5</maven.surefire.version>
+        <maven-surefire-plugin.version>3.0.0-M5</maven-surefire-plugin.version>
     </properties>
 
     <dependencyManagement>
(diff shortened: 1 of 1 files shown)


pom.xml
```diff
@@ -155,6 +155,8 @@
         <netty.version>4.1.72.Final</netty.version>
         <netty-codec-http2-version>4.1.72.Final</netty-codec-http2-version>
         <jersey-common>2.34</jersey-common>
+
+        <maven.surefire.version>3.0.0-M5</maven.surefire.version>
```
jnh5y, line 159: Any reason not to use 3.0.0-M6? 

Or is an issue caused by that newer version?

jnh5y (approved):
My question about using M6 is non-blocking.  :shipit: when ready.

pom.xml
stevenpyzhang: I'm waiting to see if the jenkins build passes with this pin, the confluent commons got updated to M6 which is when build errors started causing. As to why the new version causes build errors, I haven't dug into that yet.

mjsax:
I cherry-picked this fix to `7.2.x` branch that is also affected.

Issue #8707, opened 2022-02-02
SUM function does not check for limits


jzaralim:
**Describe the bug**
The `SUM` function does not check that the result of the aggregation will fit into the data type. For example, if you call `SUM(col)` on a `DECIMAL(4,2)` column, then you eventually run into this if you're adding `30.33 + 22.34 + 40.99 + 20`

This applies for all the data types.

**To Reproduce**
1. Create a stream with a DECIMAL column
2. Create a table from that stream calls `SUM` on the decimal column
3. Keep adding positive numbers to the stream until `SUM` column in the table is too big to fit into the decimal

**Expected behavior**
Any of the following:
* Log the error as `KsqlFunctionException`. 
* Return null
* Skip adding that entry if it will cause overflow

**Actual behaviour**
We get a serialization error with

```
java.lang.ArithmeticException: Numeric field overflow: A field with precision AAA and scale BBB must round to an absolute value less than CCC. Got DDD
	at io.confluent.ksql.util.DecimalUtil.ensureMax(DecimalUtil.java:262)
	at io.confluent.ksql.util.DecimalUtil.ensureFit(DecimalUtil.java:142)
	at io.confluent.ksql.util.DecimalUtil.ensureFit(DecimalUtil.java:121)
	at io.confluent.ksql.serde.avro.AvroDataTranslator.replaceSchema(AvroDataTranslator.java:124)
	at io.confluent.ksql.serde.avro.AvroDataTranslator.convertStruct(AvroDataTranslator.java:92)
	at io.confluent.ksql.serde.avro.AvroDataTranslator.replaceSchema(AvroDataTranslator.java:121)
	at io.confluent.ksql.serde.avro.AvroDataTranslator.toConnectRow(AvroDataTranslator.java:77)
	at io.confluent.ksql.serde.connect.KsqlConnectSerializer.serialize(KsqlConnectSerializer.java:52)
	... 26 more
```

**Additional context**

guozhangwang:
I saw this issue happening as well but with a slightly different error message on changelogs:

```
org.apache.kafka.common.errors.SerializationException: Error serializing message to topic: XYZ-changelog. Numeric field overflow: A field with precision 11 and scale 2 must round to an absolute value less than 10^9. Got 1000004556.0
```

suhas-satish:
@jnh5y , do you think this is a good fix-it-week item?

jnh5y:
[...]

PR #9058, opened 2022-04-25
Fix sum udaf decimal overflow

bvarghese1:
### Description 

Throw KsqlFunctionException when a Decimal field overflows while using the sum udaf
Addresses https://github.com/confluentinc/ksql/issues/8707

### Testing done 
Added unit and integration(QTT) tests

### Reviewer checklist

### DIFF
--- a/ksqldb-engine/src/main/java/io/confluent/ksql/query/QueryBuilder.java
+++ b/ksqldb-engine/src/main/java/io/confluent/ksql/query/QueryBuilder.java
@@ -642,7 +642,7 @@ final class QueryBuilder {
   ) {
     final QueryErrorClassifier userErrorClassifiers = new MissingTopicClassifier(applicationId)
         .and(new AuthorizationClassifier(applicationId))
-        .and(new KSqlFunctionClassifier(applicationId))
+        .and(new KsqlFunctionClassifier(applicationId))
         .and(new MissingSubjectClassifier(applicationId))
         .and(new SchemaAuthorizationClassifier(applicationId));
     return buildConfiguredClassifiers(ksqlConfig, applicationId)
--- a/ksqldb-engine/src/main/java/io/confluent/ksql/query/QueryBuilder.java
+++ b/ksqldb-engine/src/main/java/io/confluent/ksql/query/QueryBuilder.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2019 Confluent Inc.
+ * Copyright 2022 Confluent Inc.
  *
  * Licensed under the Confluent Community License (the "License"); you may not use
  * this file except in compliance with the License.  You may obtain a copy of the
@@ -641,7 +641,8 @@ final class QueryBuilder {
       final String applicationId
   ) {
     final QueryErrorClassifier userErrorClassifiers = new MissingTopicClassifier(applicationId)
-        .and(new AuthorizationClassifier(applicationId));
+        .and(new AuthorizationClassifier(applicationId))
+        .and(new KSqlFunctionClassifier(applicationId));
     return buildConfiguredClassifiers(ksqlConfig, applicationId)
         .map(userErrorClassifiers::and)
         .orElse(userErrorClassifiers);
--- a/ksqldb-engine/src/main/java/io/confluent/ksql/function/udaf/sum/DecimalSumKudaf.java
+++ b/ksqldb-engine/src/main/java/io/confluent/ksql/function/udaf/sum/DecimalSumKudaf.java
@@ -65,7 +65,7 @@ public class DecimalSumKudaf
       return aggregateValue;
     }
 
-    BigDecimal value = aggregateValue.add(currentValue, context);
+    final BigDecimal value = aggregateValue.add(currentValue, context);
 
     if (maxValue.compareTo(value.abs()) < 1) {
       throw new KsqlFunctionException(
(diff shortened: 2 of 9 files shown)


ksqldb-engine/src/main/java/io/confluent/ksql/query/QueryBuilder.java
```diff
@@ -641,7 +641,8 @@ private QueryErrorClassifier getConfiguredQueryErrorClassifier(
       final String applicationId
   ) {
     final QueryErrorClassifier userErrorClassifiers = new MissingTopicClassifier(applicationId)
-        .and(new AuthorizationClassifier(applicationId));
+        .and(new AuthorizationClassifier(applicationId))
+        .and(new KsqlFunctionClassifier(applicationId));
```
jnh5y, line 645: I want to make sure that we get buy-in on the new classifier; I definitely like the idea!

jnh5y (approved):

spena:
@jnh5y @bvarghese1 I started look at this PR and got wonder why to throw an overflow on the SUM function? A function should be allowed to change the returned schema. 

For instance, 

A query should display the correct sum even if it is higher than `4,1`. The returned value will have a different schema, which may be `10,2` for example, but that should be right. 

There are some rules about the resulted schema in different arithmetic operations:

```
Operation        Result precision                                      Result scale *
e1 + e2             max(s1, s2) + max(p1-s1, p2-s2) + 1        max(s1, s2)
e1 - e2             max(s1, s2) + max(p1-s1, p2-s2) + 1         max(s1, s2)
e1 * e2             p1 + p2 + 1                                                 s1 + s2
e1 / e2             p1 - s1 + s2 + max(6, s1 + p2 + 1)             max(6, s1 + p2 + 1)
e1 % e2           min(p1-s1, p2 -s2) + max( s1,s2 )              max(s1, s2)
```


Here's an example of Mysql:
Other DBs allow the above too. For instance, Mysql:
```
mysql> create table t1(p decimal(2,1));
mysql> insert into t1(p) values(9.99);
mysql> insert into t1(p) values(9.99);
mysql> select sum(p) from t1;
+--------+
| sum(p) |
+--------+
|   19.8  |
+--------+
```

The difference we'll have between other DBs, though, is when using CTAS statements. Other DBs create the resulted table with the schema returned by SUM. For instance:
```
mysql> create table t2 as select sum(p) from t1;
mysql> create table t2 as select sum(p) from t1;
mysql> describe t2;
+--------+---------------+------+-----+---------+-------+
| Field  | Type          | Null | Key | Default | Extra |
+--------+---------------+------+-----+---------+-------+
| sum(p) | decimal(24,1) | YES  |     | NULL    |       |
+--------+---------------+------+-----+---------+-------+
```

Whereas in ksql, the above CTAS does not know what the resulted schema would be until the rows that change the schema are processed. 

I think the option would be to change the Sink table schema dynamically when a new row in in the source table results in a higher decimal schema. Decimal evolution is allowed when precision is higher. Scale can stay the same as that's what the user chooses to do.

@colinhicks Any thoughts?

colinhicks:
Thanks for raising this, @spena, There was some discussion in the corresponding issue: https://github.com/confluentinc/ksql/issues/8707, but it makes sense to give this further thought.

In the CTAS case you're saying there's essentially no way to choose the scale/precision of the output column because the scale of the input is used? And it's not possible to use CAST against the SUM()?

What should happen if I specify the output schema with `value_schema_id`?

spena:
Right, CTAS does not know what the scale and precision would be. Even if you choose a SUM, Ksql uses the schema of the source column.

See:
```
ksql> CREATE TABLE s3 as SELECT ID, sum(value) AS SUM FROM test group by id;
ksql> describe s3;

Name                 : S3
 Field | Type                           
----------------------------------------
 ID    | BIGINT           (primary key) 
 SUM   | DECIMAL(2, 1)
```

Users could use casting if they want to change the sink schema, however this PR still fails on the SUM because it uses the source decimal type to check for overflow instead of the sink.
```
ksql> CREATE TABLE S2 as SELECT ID, CAST(sum(value) AS DECIMAL(4,1)) AS SUM FROM test group by id;
ksql> describe s2;

Name                 : S2
 Field | Type                           
----------------------------------------
 ID    | BIGINT           (primary key) 
 SUM   | DECIMAL(4, 1)                  
----------------------------------------
```

I'm not sure if changing the schema dynamically on the sink is good, though. It's a quick idea it came to my mind, but it needs more thinking. It is the same use case as this user who wanted to pass through all Protobuf fields from source -> sink even if Sink didn't specify them initially. It's part of schema and data evolution.

bvarghese1:
> > Users could use casting if they want to change the sink schema, however this PR still fails on the SUM because it uses the source decimal type to check for overflow instead of the sink.
> 
> @bvarghese1 do you have a sense of how feasible it would be to check the output schema decimal type for overflow instead?
> 
> To me the CAST approach seems pretty reasonable because it's explicit. If we can adjust the overflow check I think it's a good option to consider.

@colinhicks Let me investigate this.

spena:
@bvarghese1 Tip: I think you'll find easier to check during serialization if the conversion is allowed. The sum function will correctly do the sum, then the cast might change the schema (truncate it or something), then serializing to the Sink that has less precision/scale is the one that would fail.

bvarghese1:
As @spena pointed out, the following CTAS doesn't preserve the sink type and uses the source type
```
CREATE TABLE S2 AS SELECT ID, CAST(sum(VALUE) AS DECIMAL(6,2)) AS SUM FROM DEC_TEST GROUP BY ID;
```
but, the following does:
```
CREATE TABLE S2 AS SELECT ID, sum(CAST(VALUE AS DECIMAL(6,2))) AS SUM FROM DEC_TEST GROUP BY ID;
```

bvarghese1:
As dicussed with @spena and @colinhicks, we have decided to use the following syntax as workaround
```
CREATE TABLE S2 AS SELECT ID, sum(CAST(VALUE AS DECIMAL(6,2))) AS SUM FROM DEC_TEST GROUP BY ID;
```
The above syntax supports both upcasting and down casting and honors the users input in both cases. The sum udf will check for the up/down casted value in this scenario and raise a KSqlFunctionException.

PR #9072, opened 2022-04-27
fix: classify SR missing subject and access rights query errors as USER errors

spena:
Added a couple of classifiers that classify missing SR subjects and authorization errors from SR as USER errors. The classifiers just log the errors similar to the  `MissingTopicClassifier` class.

### Description 
Added unit tests

### Testing done 

### Reviewer checklist

### DIFF
--- a/ksqldb-engine/src/test/java/io/confluent/ksql/query/MissingSubjectClassifierTest.java
+++ b/ksqldb-engine/src/test/java/io/confluent/ksql/query/MissingSubjectClassifierTest.java
@@ -1,0 +1,61 @@
+/*
+ * Copyright 2022 Confluent Inc.
+ *
+ * Licensed under the Confluent Community License (the "License"; you may not use
+ * this file except in compliance with the License. You may obtain a copy of the
+ * License at
+ *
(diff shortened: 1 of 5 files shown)


ksqldb-engine/src/test/java/io/confluent/ksql/query/MissingSubjectClassifierTest.java
```diff
@@ -0,0 +1,61 @@
...
+
+    // When:
+    final QueryError.Type type = new MissingSubjectClassifier("").classify(e);
+
+    // Then:
+    assertThat(type, is(QueryError.Type.UNKNOWN));
+  }
+
+  @Test
+  public void shouldClassifyNoMissingSubjectAsUnknownException() {
```
vcrfxia, line 51:
```suggestion
  public void shouldClassifyOtherExceptionAsUnknownException() {
```
^ suggested name change because it took me a while to figure out how this test was different from the one above it, since the two look very similar.

vcrfxia (approved):
Thanks @spena -- LGTM with a couple minor suggestions inline!

PR #9083, opened 2022-04-28
test: add timeout to future for TlsTest to avoid blocking

mjsax:
We observes some test failures:
```
org.junit.runners.model.TestTimedOutException: test timed out after 90 seconds
	at io.confluent.ksql.api.TlsTest.lambda$shouldReloadCert$0(TlsTest.java:145)
	at io.confluent.ksql.api.TlsTest.shouldReloadCert(TlsTest.java:137)
```

It seems we are blocking in the `get()` call on the future. Adding a timeout should help as it would trigger a re-try to send the request, avoiding that the test blocks and eventually times out via the `@Rule public final Timeout timeout = Timeout.seconds(90);` annotation.


bvarghese1 (approved):

spena (commented):
That was too simple. Thanks for the PR.

PR #9086, opened 2022-05-01
feat: assert not exists topic

jzaralim:
### Description 
Adds `NOT EXISTS` to `ASSERT TOPIC` which asserts that a topic doesn't exist

### Testing done 
Manual, unit+integration tests

### Reviewer checklist

### DIFF
--- a/ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/execution/AssertTopicExecutor.java
+++ b/ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/execution/AssertTopicExecutor.java
@@ -34,10 +34,13 @@ import java.util.List;
 import java.util.Map;
 import java.util.Optional;
 import org.apache.kafka.common.TopicPartitionInfo;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 public final class AssertTopicExecutor {
 
   private static final int RETRY_MS = 100;
+  private static final Logger LOG = LoggerFactory.getLogger(AssertTopicExecutor.class);
 
   private AssertTopicExecutor() {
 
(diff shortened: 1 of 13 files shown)


ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/execution/AssertTopicExecutor.java
```diff
@@ -59,52 +62,75 @@ public static StatementExecutorResponse execute(
...
-        }
-      });
-      if (configErrors.size() > 0) {
-        throw new KsqlException(String.join("\n", configErrors));
+    if (!assertExists) {
+      if (config.size() > 0) {
+        LOG.warn("Will skip topic config check for topic non-existence assertion.");
+      }
+      if (topicExists) {
+        throw new KsqlException("Topic " + topic + " exists");
```
jnh5y, line 98: I'm 95% sure, but just to make sense, but this logic is instead the call which is retried, so if the topic is being deleted, the assertion can pass, etc, etc, right?  (Still waiting for my coffee to kick in.)

jnh5y (approved):
LGTM.

ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/execution/AssertTopicExecutor.java
jzaralim: Yup, if this fails then the entire `assertTopic` method is retried until the timeout

jzaralim:
> LGTM. (Once this is in, looks like we'd just need a doc PR.)

We also need to update the Java client and the migration tool, and we need to do all this again for `ASSERT SCHEMA` :)

PR #9095, opened 2022-05-04
docs: KLIP-63: PAUSE and RESUME for persistent queries

jnh5y:


design-proposals/klip-63-pause-resume-persistent-queries.md
```diff
@@ -0,0 +1,62 @@
...
+**Author**: Jim Hughes (jnh5y) | 
+**Release Target**: Unknown | 
+**Status**: In Discussion | 
+**Discussion**: 
+
+**tl;dr:** Adds the ability to pause the processing of a persistent query.
+           
+## Motivation and background
+
+Users may wish to pause a query for a number of reasons and then resume processing.  Presently, they would have to terminate a query and recreate it.  This approach means that any state would have to be restored; the goal of this work is to add the capability to pause processing temporarily while allowing Kafka consumers and state stores to be retained.
```
suhas-satish, line 12: will be good to enumerate a few reasons users might want to pause a persistent query

design-proposals/klip-63-pause-resume-persistent-queries.md
```diff
@@ -0,0 +1,62 @@
...
+
+## What is in scope
+
+Pausing and resuming persistent queries and the necessary language modifications to support this are in scope.
+
+## What is not in scope
+
+We are not considering the ability to pause and resume transient queries.  Additionally, modifying the consumers' state is out of scope.  (This would allow for processing to skip records or reprocess data.)
```
suhas-satish, line 20: Even if we were to modify consumer's state, how would we do it? Were you thinking of something like Flink's savepoint which is a user-controlled checkpoint (which doesnt have an analogous equivalent in kafka streams)

design-proposals/klip-63-pause-resume-persistent-queries.md
```diff
@@ -0,0 +1,62 @@
...
+
+For `PAUSE ALL` and `RESUME ALL`, all persistent queries will be paused or resumed.
+
+## Test plan
+
+Integration tests will show that 
+- Pausing a query moves it from the RUNNING state to the PAUSED state.
+- Queries in the PAUSED state do not process data.
+- Paused queries can be resumed.  Once resumed, they will be in the running state and process data.
+- Pausing a query does not impact other queries managed by the system.  (This will with and without shared runtimes enabled.)
```
suhas-satish, line 44: "This will with" - are you missing a word between will and with here?

design-proposals/klip-63-pause-resume-persistent-queries.md
```diff
@@ -0,0 +1,62 @@
...
+resume (respectively) the topology associated with the query.
+
+For `PAUSE ALL` and `RESUME ALL`, all persistent queries will be paused or resumed.
+
+## Test plan
+
+Integration tests will show that 
+- Pausing a query moves it from the RUNNING state to the PAUSED state.
+- Queries in the PAUSED state do not process data.
+- Paused queries can be resumed.  Once resumed, they will be in the running state and process data.
```
suhas-satish, line 43: should we add a test case for Once resumed, they will not reprocess previously processed data / offsets?

design-proposals/klip-63-pause-resume-persistent-queries.md
```diff
@@ -0,0 +1,62 @@
...
+- Pausing a query does not impact other queries managed by the system.  (This will with and without shared runtimes enabled.)
+- Queries can be paused when multiple ksqlDB servers are in use.  (This demonstrates that the PAUSE messages use the ksqlDB command topic.)
```
suhas-satish, line 45: does this imply that the command topic is not used when there is a single ksqlDB server (as against multiple ksqlDB servers) ?

design-proposals/klip-63-pause-resume-persistent-queries.md
```diff
@@ -0,0 +1,62 @@
...
+Add PAUSE <queryID>, PAUSE ALL, RESUME <queryID>, RESUME ALL.
+
+## Design
+
+For `PAUSE <queryID>` and `RESUME <queryID>`, appropriate methods on the `KafkaStreams` client will be called to pause or 
+resume (respectively) the topology associated with the query.
+
+For `PAUSE ALL` and `RESUME ALL`, all persistent queries will be paused or resumed.
+
+## Test plan
```
suhas-satish, line 38: should we add a test case for "when 2 queries share an underlying resource (input or output topic or partition or state store or ksql server node)", pausing 1 does not inadvertently affect another ?

design-proposals/klip-63-pause-resume-persistent-queries.md
jnh5y: You've stated the problem with resuming from a different place (and hence why it is out of scope for this work!).  As far as I know, we don't have a clear way to do this in Streams.

In addition to savepoint, I've seen folks suggest managing the offsets for each partition.

design-proposals/klip-63-pause-resume-persistent-queries.md
jnh5y: Yes... that is reasonable thing to test.  

I'm going to update this line to be overly specific.;)

design-proposals/klip-63-pause-resume-persistent-queries.md
jnh5y: The command topic will always be used for PAUSE/RESUME commands.  

I can see the implication...  I'm adding some words to the design section to explain the use of the command topic.

design-proposals/klip-63-pause-resume-persistent-queries.md
jnh5y: I'm updating one of the lines to read:

"- Pausing a query does not impact other queries managed by the system.  Concretely, when multiple queris share a topic, pausing one of the queries will not stop the processing for the other queries.  This will be demonstrated with and without shared runtimes enabled."

I don't think we can share state stores at this point.  If I'm wrong about that or if sharing state stores becomes possible, then pausing one of the queries sharing a state store becomes interesting.

design-proposals/klip-63-pause-resume-persistent-queries.md
jnh5y: Sure, I updated some words here.  Feel free to suggest other use cases!

suhas-satish: new text looks good. In addition to those, is improved responsiveness on the UI also a use case at all for streaming designer? If there are multiple persistent queries running concurrently, does pausing a few ensure others may make faster progress?

design-proposals/klip-63-pause-resume-persistent-queries.md
suhas-satish: In the new phrasing " the commands will be considered for compaction" , for my own understanding, compaction is triggered by rocksDB and not kafka streams (as I understand it).  What is the sequence of steps during a server restart wrt rocksDB that will trigger this compaction?

suhas-satish (approved):
I'm approving this. But feel free to wait on others in the team if there are gaps in my review.

design-proposals/klip-63-pause-resume-persistent-queries.md
jnh5y: I don't know if pausing queries will help with the UI for streaming designer; that may depend on implementation details of that product.  I was trying to keep the motivation general.

In some situations, I could imagine pausing a portion of the queries would allow the remainder to run faster.  I was trying to suggest that with the new text "... or to manage computational resources."

Should I be more concrete?

Also, just as a note, paused queries will still use memory and CPU for the Kafka consumers.  Given that, I don't think folks will be using PAUSE and RESUME to address lag in a data pipeline.  (It'd like be better to scale the cluster or modify the flow to achieve the desired latency.)

design-proposals/klip-63-pause-resume-persistent-queries.md
jnh5y: The runtimes are KafkaStreams instances.  Originally, each persistent query got its own instance.  Each instance creates a certain (configurable) number of StreamThreads (which map to real threads, I believe).  

This would mean that a given large number of queries, the JVM would slow down due to the large number of threads computing for CPUs.  

Shared runtimes address this by packing multiple queries into a given KafkaStreams instance.  

With the original setup, pausing in AK Streams would look like skipping all processing.  With multiple queries in a runtime, AK should skip over the topologies which are paused.

Presently, queries/topologies in a shared runtime do not share input topics not state stores.

jnh5y: Ah!  Compaction is overloaded.  ksqlDB has a "command compactor" which is used on server start to replay the state of the system.

If a group of commands would interact with each other, those are considered together.  For instance, a pair of create stream and drop stream commands may cancel each other; when that's the case, there's no need to create the resources only to remove it later.

With PAUSE and RESUME, those should cancel each other out in a similar manner.  The "interesting" corner case is what should happen to PAUSED queries on server restart.  I'm suggesting that they would be paused after the system comes up.  (To be fair, I haven't worked out all the details in my prototype, but I think that it is doable.)

suhas-satish: thank you very much for the explanations. They helped improve my naive understanding.

design-proposals/klip-63-pause-resume-persistent-queries.md
```diff
@@ -0,0 +1,66 @@
...
+**Status**: In Discussion | 
+**Discussion**: 
+
+**tl;dr:** Adds the ability to pause the processing of a persistent query.
+           
+## Motivation and background
+
+Users may wish to pause a query for a number of reasons and then resume processing.  As examples, users may wish to modify data pipelines built from persistent queries or to manage computational resources.
+
+Presently to manage processing, they would have to terminate a query and recreate it.  This approach means that any state would have to be restored; the goal of this work is to add the capability to pause processing temporarily while allowing Kafka consumers and state stores to be retained.
```
pgaref, line 14: How is this affecting data producers? Is the assumption here that we will be able to hold/buffer data upstream as long as the query is paused? What happens if the query is paused indefinitely? Is there going to be a threshold?

design-proposals/klip-63-pause-resume-persistent-queries.md
```diff
@@ -0,0 +1,66 @@
...
+
+Pausing and resuming persistent queries and the necessary language modifications to support this are in scope.
+
+## What is not in scope
+
+We are not considering the ability to pause and resume transient queries.  Additionally, modifying the consumers' state is out of scope.  (This would allow for processing to skip records or reprocess data.)
+
+## Value/Return
+
+End users will be able to manage query processing once this feature is implemented.  For example, this will allow for better control while designing data pipelines.  Pausing an upstream query will allow users to update a downstream query or other processing and iterate through design without processing all the data in a stream.
```
pgaref, line 26: Assuming that users pause and update a query that introduces breaking changes to downstream consumers/queries -- will we be able to detect that? How are we going to handle such cases?

design-proposals/klip-63-pause-resume-persistent-queries.md
```diff
@@ -0,0 +1,66 @@
...
+
+## Public APIS
+
+SQL Grammar changes:
+Add PAUSE <queryID>, PAUSE ALL, RESUME <queryID>, RESUME ALL.
```
pgaref, line 31: Isn't pause/resume ALL too wide/risky?

jnh5y: ksqlDB supports TERMINATE ALL which does more damage: https://docs.ksqldb.io/en/latest/developer-guide/ksqldb-reference/terminate/

I'm fine with or without the ALL keyword.  I prototyped it, but it does not have to stay.

design-proposals/klip-63-pause-resume-persistent-queries.md
jnh5y: I'd suggest that situation is tantamount to any time that a Kafka producer starts writing a new data format to a topic.  In general, I don't think there's much defense for that.

The place where there are guardrails are topics with schemas registered.  I don't think ksqlDB will let a persistent query be created which writes incorrect output to an existing topic (which has a topic managed).

design-proposals/klip-63-pause-resume-persistent-queries.md
jnh5y: (This is a great point for us to discuss as a group!)

design-proposals/klip-63-pause-resume-persistent-queries.md
jnh5y: I believe the output from a KafkaStreams topology would be sent out to a Kafka producer during the execution of the topology. 

Pausing a query will not interrupt a single execution of the topology, so the output from a topology should be written out to the Kafka producers a-ok.  

If a query is paused indefinitely, the query is paused.  At the moment, there is no threshold or timeout.

design-proposals/klip-63-pause-resume-persistent-queries.md
```diff
@@ -0,0 +1,66 @@
...
+**Author**: Jim Hughes (jnh5y) | 
+**Release Target**: Unknown | 
+**Status**: In Discussion | 
+**Discussion**: 
+
+**tl;dr:** Adds the ability to pause the processing of a persistent query.
+           
+## Motivation and background
+
+Users may wish to pause a query for a number of reasons and then resume processing.  As examples, users may wish to modify data pipelines built from persistent queries or to manage computational resources.
```
lct45, line 12: Just curious - since the size of the ksql cluster is maintained, users wouldn't see any cost savings from this, would they? Or would they see some since the queries wouldn't be processing?

jnh5y: Presently, they would not see costs savings from pausing individual queries.

I was meaning "manage computational resources" to suggest that some queries may process a little faster.  

If one was counting pennies, maybe one could find some loose change in savings for how long data was in Kafka?  Overall, this is not a feature aimed at reducing costs.

lct45: That makes sense, I was just making sure my understanding was correct 😄  Maybe worth making sure that the customer knows that their bill will stay the same

jnh5y: Hmm...  I don't think we need to provide billing details in this KLIP.  (Shout if I'm wrong!)

design-proposals/klip-63-pause-resume-persistent-queries.md
mjsax: If a query is pause for too long, and the input topic is subject to data purging via retention, it would be an issue semantically, but it should not crash the query.

The consumer would try to fetch from a non-existing offset when resumed, and the broker would return an error resulting in an exception. Kafka Streams would catch this exception and would reset the offset based on `auto.offset.reset` configuration.

mjsax: > This approach means that any state would have to be restored

This might be miss-leading -- the state is lost and the changelog topic are deleted if you terminate a query. Thus, you cannot _restore_ state -- if (but only if) you still have all your input data, you can only _recompute_ it though.

design-proposals/klip-63-pause-resume-persistent-queries.md
mjsax: Personally, I think `ALL` is fine -- I could imagine a use case for which you want to pause all queries at the same time because they form some processing pipeline and the "final result" is consumed by some downstream application (ie, external to ksqlDB) and you want to make changes downstream.

mjsax (approved):

design-proposals/klip-63-pause-resume-persistent-queries.md
jnh5y: To the first comment, is that situation worth calling out in the KLIP?

For the second, nice call, I am replacing `restored` with `recomputed`.

mjsax: I think it's not necessary to discuss on the KLIP. Would go down the rabbit whole too much -- just wanted to explain the situation (in case anybody is interested in the technical details).

jnh5y:
As a side note, should the KLIP define what happens to IQs against paused tables?

mjsax:
Not sure if we need to mention anything? In the end, the table state can still be queried -- it will just stop to advance, right?

jnh5y:
> Not sure if we need to mention anything? In the end, the table state can still be queried -- it will just stop to advance, right?

I was trying to think through what would happen with queries which are started in a paused state.  I'm slightly unsure about some of the fine details there...  I'll report back if things seem untenable.

mjsax:
From my understanding, we don't block restoring of state stores (at least from what we discussed on the KIP). It's an interesting question if we would want to block state restoration, too?

jnh5y:
> From my understanding, we don't block restoring of state stores (at least from what we discussed on the KIP). It's an interesting question if we would want to block state restoration, too?

I believe that the same code in AK is used to do state restoration and following the changelog topics.

vvcephei:
Hey @mjsax and @jnh5y , It seems like the reasoning from the KIP for standbys also applies to restoration. Namely, the simplest behavior for pausing is if "pause means pause" and just everything stops happening when you pause it.

Of course, unlike standbys, restoration is finite, so if the app stays paused long enough, the restoration will complete, and then the app will be fully paused.

Then again, that feels like premature optimization, and the downside is that we may find ourselves chasing down and explaining why the restoration didn't get paused when the app did. Eg, how we're constantly debugging rebalances only to realize they are probing rebalances and everything is fine. It seems better if the operational profile is simple, even if it is a little dumb.

jnh5y:
One consequence of pausing restoration is that the KafkaStreams instance will not transition to running.  I wonder what issues that'll cause.  

Perhaps this is indicating that my goal of being able to start topologies/queries in a paused state is a tad too complex?

PR #9096, opened 2022-05-06
feat: automatically build confluent cloud image on every master merge

swist:
### Description 
This should enable the tooling to automatically trigger build
in cc-docker-ksql whenever we hit a green master commit in this
repository
### Testing done 

### Reviewer checklist

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -8,7 +8,7 @@ common {
     dockerPush = false
     dockerScan = false
     dockerImageClean = false
-    downStreamRepos = ["confluent-security-plugins", "confluent-cloud-plugins"]
+    downStreamRepos = ["confluent-security-plugins", "confluent-cloud-plugins", "cc-docker-ksql"]
     downStreamValidate = false
     nanoVersion = true
     maxBuildsToKeep = 99


jnh5y (approved):

Jenkinsfile
```diff
@@ -8,7 +8,7 @@ common {
     dockerPush = false
     dockerScan = false
     dockerImageClean = false
-    downStreamRepos = ["confluent-security-plugins", "confluent-cloud-plugins"]
+    downStreamRepos = ["confluent-security-plugins", "confluent-cloud-plugins", "cc-docker-ksql"]
```
jnh5y, line 11: This runs separate from the Jenkins CI for checking PRs, right?

swist: Ish, there is a bit of magic involved. When you configure a `downstream` repo, our build pipeline goes through the list and checks if the repos in the list have a branch matching the one you're trying to build "here". If there is one, it will trigger a build on that branch if and only if the build "here" is green

jnh5y: Ah, so the 99% use case will not 'cause CI to build things.  One would have to be coordinating branches or get really unlucky for the downstream Docker to build...

PR #9097, opened 2022-05-06
chore: improve variable substitution for assert statements

jzaralim:
### Description 
1. Allows users to specify topic names and schema subjects using either strings or identifier names. This is important because topics and schemas are allowed to use the `-` character in names, but ksqlDB does not allow that in identifiers.
2. Allows variable substitution for quoted names. 

### Testing done 
Unit tests, manual testing. Integration testing will be included with java client changes.

### Reviewer checklist

### DIFF
--- a/ksqldb-parser/src/main/antlr4/io/confluent/ksql/parser/SqlBase.g4
+++ b/ksqldb-parser/src/main/antlr4/io/confluent/ksql/parser/SqlBase.g4
@@ -59,7 +59,7 @@ statement
     | DESCRIBE STREAMS EXTENDED?                                            #describeStreams
     | DESCRIBE FUNCTION identifier                                          #describeFunction
     | DESCRIBE CONNECTOR identifier                                         #describeConnector
-    | PRINT (identifier| STRING) printClause                                #printTopic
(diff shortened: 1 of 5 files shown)


ksqldb-parser/src/main/antlr4/io/confluent/ksql/parser/SqlBase.g4
```diff
@@ -106,6 +106,11 @@ runScript
     : RUN SCRIPT STRING
     ;
 
+resourceName
+    : identifier
+    | STRING
```
jnh5y, line 111: This one deserves some careful thought...  

Does this let in anything we wouldn't want?

jnh5y (commented):
Let's check on the grammar on Monday.

ksqldb-parser/src/main/antlr4/io/confluent/ksql/parser/SqlBase.g4
jzaralim: For topic names, no because the original `PRINT TOPIC` defined topic names like this anyway. Schema registry subject names are pretty similar to toptic names, so I don't think there are any problems there either.

PR #9103, opened 2022-05-10
[DRAFT] Pull proto draft

cprasad1:
### Description 

### Testing done 

### Reviewer checklist


ksqldb-rest-app/src/main/java/io/confluent/ksql/api/server/ProtobufQueryStreamResponseWriter.java
```diff
@@ -0,0 +1,166 @@
...
+      for (Field field : connectSchema.fields()) {
+        ksqlRecord.put(
+                field,
+                keyValue.value().get(i));
+        i += 1;
+      }
+      final byte[] protoMessage = serializer.serialize("", ksqlRecord);
+      final Optional<DataRowProtobuf> protoRow =
+              StreamedRow.pullRowProtobuf(protoMessage).getRowProtobuf();
+      response.write(",\n").write(serializeObject(protoRow));
```
agavra, line 104: is there any reason we're adding newlines to everything? the other formats seem to just rely on `,` in between rows and then users can pipe it into `jq` or something if they want to

I'm asking because it seems like we may be able to reuse a lot of the `JsonQueryStreamResponseWriter` code if we standardized

agavra (commented):
Did a quick skim to give you some initial feedback, I'll review the tests in more depth tomorrow!

Gerrrr:
The new content type behaves differently compared to `application/vnd.ksql.v1+json` for empty responses. We should definitely fix that before merging.

Consider a new table:

```
ksql> CREATE SOURCE TABLE t1 (K INT PRIMARY KEY, VS ARRAY<INT>) WITH (KAFKA_TOPIC='t1', value_format='JSON', partitions=1);
```

Let's query it with `Accept: application/vnd.ksql.v1+protobuf`:

```
curl --header "Accept: application/vnd.ksql.v1+protobuf" \
     -v --request POST \
     -d '{"ksql": "SELECT K, VS FROM t1 WHERE K=2 LIMIT 1;"}' \
      http://localhost:8088/query
Note: Unnecessary use of -X or --request, POST is already inferred.
*   Trying ::1...
* TCP_NODELAY set
* Connected to localhost (::1) port 8088 (#0)
> POST /query HTTP/1.1
> Host: localhost:8088
> User-Agent: curl/7.64.1
> Accept: application/vnd.ksql.v1+protobuf
> Content-Length: 51
> Content-Type: application/x-www-form-urlencoded
>
* upload completely sent off: 51 out of 51 bytes
< HTTP/1.1 406 Not Acceptable
< content-length: 0
<
* Connection #0 to host localhost left intact
* Closing connection 0
```

We've got `406 Not Acceptable` and an empty response. The same request with `Accept: application/vnd.ksql.v1+json` returns 200 and a header in the body:

```
curl --header "Accept: application/vnd.ksql.v1+json" \
     -v --request POST \
     -d '{"ksql": "SELECT K, VS FROM t1 WHERE K=2 LIMIT 1;"}' \
      http://localhost:8088/query
Note: Unnecessary use of -X or --request, POST is already inferred.
*   Trying ::1...
* TCP_NODELAY set
* Connected to localhost (::1) port 8088 (#0)
> POST /query HTTP/1.1
> Host: localhost:8088
> User-Agent: curl/7.64.1
> Accept: application/vnd.ksql.v1+json
> Content-Length: 51
> Content-Type: application/x-www-form-urlencoded
>
* upload completely sent off: 51 out of 51 bytes
< HTTP/1.1 200 OK
< Transfer-Encoding: chunked
<
* Connection #0 to host localhost left intact
[{"header":{"queryId":"query_1652258828691","schema":"`K` INTEGER KEY, `VS` ARRAY<INTEGER>"}}]* Closing connection 0
```

ksqldb-rest-client/src/main/java/io/confluent/ksql/rest/client/KsqlTarget.java
```diff
@@ -221,6 +222,21 @@ public RestResponse<List<StreamedRow>> postQueryRequest(
         KsqlTarget::toRows);
   }
 
+  public RestResponse<List<StreamedRow>> postQueryStreamRequestProto(
+          final String ksql,
+          final Map<String, ?> requestProperties
+  ) {
+    final Map<String, Object> requestPropertiesObject = requestProperties.entrySet().stream()
```
Gerrrr, line 229: Can we make `requestProperties` of type `Map<String, Object>`  instead of creating a full copy?

ksqldb-rest-client/src/main/java/io/confluent/ksql/rest/client/KsqlTargetUtil.java
```diff
@@ -69,6 +70,41 @@ public static List<StreamedRow> toRows(final Buffer buff) {
     return rows;
   }
 
+  public static List<StreamedRow> toRowsFromProto(final Buffer buff) {
+    final List<StreamedRow> rows = new ArrayList<>();
+    boolean headerRead = false;
+    int begin = 0;
+
+    for (int i = 0; i <= buff.length(); i++) {
```
Gerrrr, line 78: This method deserves a detailed javadoc.

cprasad1:
@Gerrrr the new proto format is only acceptable by the `/query-stream` endpoint and not the `/query` endpoint, so we expect to get a `406` (on all pull queries with the new format and not just on empty tables) while querying the `/query` endpoint.

ksqldb-functional-tests/src/test/java/io/confluent/ksql/test/rest/RestTestExecutor.java
```diff
@@ -993,6 +1028,97 @@ public void verify(
...
+        final Map<String, Object> expected = (Map<String, Object>) expectedRows.get(i);
+
+        if (actual.containsKey(HEADER_PROTOBUF)
+                && ((HashMap<String, Object>) actual.get(HEADER_PROTOBUF)).containsKey(SCHEMA)) {
+
+          schema = new ProtobufSchema((String) ((Map<?, ?>)actual.get(HEADER_PROTOBUF)).get(SCHEMA));
+          final String actualSchema = (String) ((Map<?, ?>)actual.get(HEADER_PROTOBUF)).get(SCHEMA);
+          final String expectedSchema = (String) ((Map<?, ?>)expected.get(HEADER_PROTOBUF)).get(SCHEMA);
+
+          assert(actualSchema.equals(expectedSchema));
```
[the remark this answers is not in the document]
cprasad1: done

ksqldb-rest-model/src/main/java/io/confluent/ksql/rest/entity/StreamedRow.java
```diff
@@ -197,32 +250,45 @@ public static StreamedRow consistencyToken(final ConsistencyToken consistencyTok
   @JsonCreator
   private StreamedRow(
       @JsonProperty("header") final Optional<Header> header,
+      @JsonProperty("headerProtobuf") final Optional<HeaderProtobuf> headerProtobuf,
```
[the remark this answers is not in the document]
cprasad1: Done. I have refactored the `Header` to include an `Optional<LogicalSchema> schema` or an `Optional<String> protoSchema` depending on the situation

ksqldb-rest-app/src/main/java/io/confluent/ksql/api/server/ProtobufQueryStreamResponseWriter.java
cprasad1: Done. `JsonQueryStreamResponseWriter` now takes care of both json and protobuf format depending on the media type

ksqldb-functional-tests/src/test/java/io/confluent/ksql/test/rest/RestTestExecutor.java
```diff
@@ -175,13 +181,27 @@ void buildAndExecuteQuery(final RestTestCase testCase) {
...
+
+
       if (!queryResults.isEmpty()) {
         failIfExpectingError(testCase);
       }
 
+      List<RqttResponse> protoResponses = ImmutableList.of();
+
+      if (!testCase.getTestLocation().toString().contains("scalable-push-queries")) {
+        protoResponses = statements.queries.stream()
```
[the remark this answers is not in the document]
cprasad1: done. added a configuration `"testPullWithProtoFormat"` to parameterize the rqtt. When true, it runs and validates the test with the protobuf format

ksqldb-functional-tests/src/test/java/io/confluent/ksql/test/rest/RestTestExecutor.java
```diff
@@ -993,6 +1028,97 @@ public void verify(
...
+                      .collect(Collectors.joining(INDENT, INDENT, ""))
+                      + System.lineSeparator(),
+              rows,
+              hasSize(expectedRows.size())
+      );
+
+      ProtobufSchema schema = null;
+      ProtobufNoSRConverter.Deserializer deserializer = new ProtobufNoSRConverter.Deserializer();
+      final JsonMapper mapper = new JsonMapper();
+      for (int i = 0; i != rows.size(); ++i) {
```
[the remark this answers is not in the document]
cprasad1: sorry, did not understand this. Can you elaborate?

Gerrrr:
Thank you for elaborating @cprasad1!

ksqldb-rest-client/src/main/java/io/confluent/ksql/rest/client/KsqlRestClient.java
```diff
@@ -308,6 +308,14 @@ public RestResponse<List<StreamedRow>> makeQueryRequest(
         ksql, requestPropertiesToSend, Optional.ofNullable(commandSeqNum));
   }
 
+  public RestResponse<List<StreamedRow>> makeQueryStreamRequestProto(
```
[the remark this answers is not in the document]
cprasad1: All the methods exposed in `KsqlRestClient` and `KsqlTarget` have hard-coded either the endpoint as `/query` or the response format as json or json delimited. We don't have a method that is generic enough to query any combination of endpoints x formats. Ideally, we would want to have an api like that. I tried refactoring all the querying logic into a few generic methods, but it was a bit too much of refactoring given how lots of these parameters are hardcoded and have custom logic wrt sending the query requests (/query and /query-stream have different request frmats), the different content types have different handling and how async handling logic differs from the sync handling logic.

An ideal api would take in (endpoint, content-type, sync/async) as its input, but we don't have that, therefore the need for this

ksqldb-rest-client/src/main/java/io/confluent/ksql/rest/client/KsqlTargetUtil.java
```diff
@@ -69,6 +70,41 @@ public static List<StreamedRow> toRows(final Buffer buff) {
...
+        if (begin != i) { // Ignore random newlines - the server can send these
+          final Buffer sliced = buff.slice(begin, i);
+          final Buffer tidied = toJsonMsg(sliced, true);
+          if (tidied.length() > 0) {
+            if (!headerRead) {
+              final StreamedRow row = deserialize(tidied, StreamedRow.class);
+              rows.add(row);
+              headerRead = true;
+            } else {
+              try {
```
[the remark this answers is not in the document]
cprasad1: done

ksqldb-rest-client/src/main/java/io/confluent/ksql/rest/client/KsqlTargetUtil.java
cprasad1: done

ksqldb-functional-tests/src/test/java/io/confluent/ksql/test/rest/RestTestExecutor.java
```diff
@@ -993,6 +1028,97 @@ public void verify(
...
+        } else if (actual.containsKey("finalMessage")) {
+          assert(expected.equals(actual));
+        } else {
+          JSONObject row = new JSONObject(actual);
+
+          byte[] bytes = new byte[0];
+          try {
+            bytes = mapper.readTree(row.toString()).get("rowProtobuf").get("row").binaryValue();
+          } catch (IOException e) {
+            e.printStackTrace();
```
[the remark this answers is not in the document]
cprasad1: yep, failing the test upon a json deserialization exception.

ksqldb-rest-client/src/main/java/io/confluent/ksql/rest/client/KsqlTarget.java
cprasad1: done

ksqldb-functional-tests/src/test/java/io/confluent/ksql/test/rest/RestTestExecutor.java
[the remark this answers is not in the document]
cprasad1: If you meant asserting that the header comes before the rows, then I have added an assertion of i==0 down below.

cprasad1:
@agavra @Gerrrr Thank you for the helpful comments! :)
I hope that I have addressed all of them. This PR is ready for review again

ksqldb-rest-app/src/main/java/io/confluent/ksql/api/server/JsonQueryStreamResponseWriter.java
```diff
@@ -76,7 +112,23 @@ public QueryStreamResponseWriter writeRow(
     if (keyValue.value() == null) {
       LOG.warn("Dropped tombstone. Not currently supported");
     } else {
-      writeBuffer(ServerUtils.serializeObject(keyValue.value().values()));
+      if (mediaType.equals(KsqlMediaType.KSQL_V1_PROTOBUF.mediaType())) {
```
Gerrrr, line 115: Do I understand correctly that you merged `JsonQueryStreamResponseWriter` and `ProtobufQueryStreamResponseWriter` since the last review review round and the motivation was to reduce the code duplication as @agavra asked?

There are some differences between `ProtobufQueryStreamResponseWriter` and ` JsonQueryStreamResponseWriter` that were not ported, specifically in methods `writeContinuationToken`, `writeError`, `writeConsistencyToken`, `writeCompletionMessage`, `writeLimitMessage`. Why are they no longer necessary?

ksqldb-rest-client/src/main/java/io/confluent/ksql/rest/client/KsqlTargetUtil.java
```diff
@@ -69,6 +71,63 @@ public static List<StreamedRow> toRows(final Buffer buff) {
...
+            .stream(buff
+                    .toString()
+                    .replace("},{\"row\":", "},\n\t\n{\"row\":")
+                    .split("\n\t\n"))
+            .collect(Collectors.toList())
+            .stream()
+            .map(row -> toJsonMsg(Buffer.buffer(row), true))
+            .collect(Collectors.toList());
+
+    for (Buffer tidied: buffRows) {
```
Gerrrr, line 98: Thank you for adding the javadoc and elaborate comments! It helped with understanding this method a lot.

We can further reduce complexity of this method, specifically nesting level and one unnecessary condition, by extracting reading of the header out of the main loop:

```java
    // deserialize the first row (header) into a StreamedRow and add it to
    // the list to return.
    rows.add(deserialize(buffRows.get(0), StreamedRow.class));
    for (int i = 1; i < buffRows.size(); i++) {
      final Buffer tidied = buffRows.get(i);
      try {
        // try to deserialize every subsequent buffer row into a StreamedRow
        // and add it to the list to return. These can be continuation tokens,
        // consistency tokens, error messages, final messages or limit messages.
        // We expect it to fail for the rowProtobuf objects which is handled below.
        rows.add(deserialize(tidied, StreamedRow.class));
      } catch (Exception e) {
        // if it is none of the above, then deserialize it to a
        // StreamedRow.DataRowProtobuf object.
        // We can't directly deserialize into a StreamedRow like we did above
        // because DataRowProtobuf's serialization is {"row":"CAESBlVTRVJfMA=="}
        // whereas the serialization of a StreamedRow containing a DataRowProtobuf
        // looks like {"rowProtobuf":{"row":"CAESBlVTRVJfMA=="}}. The Rest API
        // produces {"row":"CAESBlVTRVJfMA=="}, so we have to do a bit of extra conversion
        // here.
        final DataRowProtobuf protoBytes =
                deserialize(tidied, DataRowProtobuf.class);
        final byte[] bytes = protoBytes.getRow();
        rows.add(StreamedRow.pullRowProtobuf(bytes));
      }
    }
    return rows;
```

WDYT?

ksqldb-rest-client/src/main/java/io/confluent/ksql/rest/client/KsqlTargetUtil.java
```diff
@@ -69,6 +71,63 @@ public static List<StreamedRow> toRows(final Buffer buff) {
...
+
+    // split the buffer across the commas delimiting the individual rows
+    // Since, there would be commas in the header we are replacing
+    // },{"row": with },\n\t\n{"row": and then splitting on \n\t\n to avoid
+    // splitting on commas that might be more that 1 level deep
+    final List<Buffer> buffRows = Arrays
+            .stream(buff
+                    .toString()
+                    .replace("},{\"row\":", "},\n\t\n{\"row\":")
+                    .split("\n\t\n"))
```
agavra, line 92: this seems pretty brittle - I guess we "know" that the contents of `row` are protobuf encoded base64 bytes, right? Is a comma even possible to show up in base64? if it isn't base64, then why can't `\n\t\n` show up as well?

can we use a "real" json parser here to extract each row?

ksqldb-rest-client/src/main/java/io/confluent/ksql/rest/client/KsqlTargetUtil.java
```diff
@@ -69,6 +71,63 @@ public static List<StreamedRow> toRows(final Buffer buff) {
...
+        } catch (Exception e) {
+          // if it is none of the above, then deserialize it to a
+          // StreamedRow.DataRowProtobuf object.
+          // We can't directly deserialize into a StreamedRow like we did above
+          // because DataRowProtobuf's serialization is {"row":"CAESBlVTRVJfMA=="}
+          // whereas the serialization of a StreamedRow containing a DataRowProtobuf
+          // looks like {"rowProtobuf":{"row":"CAESBlVTRVJfMA=="}}. The Rest API
+          // produces {"row":"CAESBlVTRVJfMA=="}, so we have to do a bit of extra conversion
+          // here.
+          final DataRowProtobuf protoBytes =
```
agavra, line 121: You mention that the Rest API produced the StreamdRow directly - what produces `DataRowProtobuf` and can we just standardize on one format? (i.e. why can't they all produce just `{"row":"CAESBlVTRVJfMA=="}`?

ksqldb-rest-model/src/main/java/io/confluent/ksql/rest/entity/StreamedRow.java
```diff
@@ -325,25 +377,32 @@ public QueryId getQueryId() {
...
+        @JsonProperty(value = "protobufSchema") final Optional<String> protoSchema
     ) {
-      return new Header(queryId, columnsSchema);
+      return new Header(queryId, columnsSchema, protoSchema);
     }
 
     private Header(
         final QueryId queryId,
-        final LogicalSchema columnsSchema
+        final Optional<LogicalSchema> columnsSchema,
```
agavra, line 400: nit: as discussed offline, I think we should just keep this as required

ksqldb-rest-app/src/main/java/io/confluent/ksql/api/server/JsonQueryStreamResponseWriter.java
```diff
@@ -54,16 +72,34 @@ public class JsonQueryStreamResponseWriter implements QueryStreamResponseWriter
...
-    buff.appendBuffer(ServerUtils.serializeObject(metaData));
+    if (mediaType.equals(KsqlMediaType.KSQL_V1_PROTOBUF.mediaType())) {
+      final LogicalSchema schema = metaData.schema;
+      final String queryId = metaData.queryId;
+
+      connectSchema = ConnectSchemas.columnsToConnectSchema(schema.columns());
+      serializer = new ProtobufNoSRSerdeFactory(ImmutableMap.of())
+              .createSerializer(connectSchema, Struct.class, false);
+
+      final StreamedRow header = StreamedRow.headerProtobuf(
```
agavra, line 96: just checked here, we have `schema` anyway so it should be pretty cheap to just pipe it in right here

ksqldb-functional-tests/src/test/java/io/confluent/ksql/test/rest/RestTestExecutor.java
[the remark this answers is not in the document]
agavra: that's almost what I meant. We should fail if:
- there is no header row
- there is no final row
- the header row doesn't happen at the beginning
- the final row doesn't happen at the end

your new change just addresses the third point

ksqldb-functional-tests/src/test/resources/rest-query-validation-tests/pull-queries-protobuf.json
```diff
@@ -0,0 +1,288 @@
...
+      "testPullWithProtoFormat": true,
+      "format": ["JSON"],
+      "statements": [
+        "CREATE STREAM riderLocations (profileId VARCHAR, latitude DOUBLE, longitude DOUBLE) WITH (kafka_topic='test_topic', value_format='{FORMAT}');",
+        "SELECT * FROM riderLocations LIMIT 2;",
+        "SELECT * FROM riderLocations LIMIT 0;",
+        "SELECT * FROM riderLocations LIMIT 5;"
+      ],
+      "topics": [
+        {"name": "test_topic", "partitions": 1} // to get a stable ordering
```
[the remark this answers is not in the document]
agavra: I believe we use a Jackson add on that supports comments

agavra (commented):
This is getting really close, only two things I consider "blocking" are the questions I asked on `KsqlTargetUtil`

ksqldb-cli/src/main/java/io/confluent/ksql/cli/console/Console.java
```diff
@@ -406,7 +406,7 @@ private void printRowHeader(final Header header) {
         writer().println(
             TabularRow.createHeader(
                 getWidth(),
-                header.getSchema().columns(),
+                header.getSchema().get().columns(),
```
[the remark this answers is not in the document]
cprasad1: making the schema non-optional, and always including it

ksqldb-rest-client/src/main/java/io/confluent/ksql/rest/client/KsqlTargetUtil.java
cprasad1: using a real json parser now

ksqldb-rest-client/src/main/java/io/confluent/ksql/rest/client/KsqlTargetUtil.java
cprasad1: looks great. incorporated the suggestion

Gerrrr: Thank you!

ksqldb-rest-app/src/main/java/io/confluent/ksql/api/server/JsonQueryStreamResponseWriter.java
Gerrrr: FWIW, I think that it would be better to have format-specific implementations in different classes rather than having conditions in every method. W could reduce code duplication by:
* moving default implementations to the interface or abstract class
* moving format-specfic serialization to a separate class or method.

For completness, we should not necessarily block this PR on it; it is fine with me to just create a follow-up issue.

ksqldb-rest-client/src/main/java/io/confluent/ksql/rest/client/KsqlTargetUtil.java
cprasad1: >You mention that the Rest API produced the StreamdRow directly
not 100%. The REST API for proto produces `StreamedRow` for everything except the protobuf data rows. For the protobuf data rows, the `JsonQueryStreamResponseWriter` would produce the `DataRowProtobuf` which is an optional member of `StreamedRow` and gets serialized to `{"row":"CAESBlVTRVJfMA=="}`. 

Why is there this weird conditionality for protobuf data rows? If we just serialize a `StreamedRow` that contains a `DataRowProtobuf`, then it would look something like `{"rowProtobuf":{"row":"CAESBlVTRVJfMA=="}` which is not the public api that we want to abide by (unless we change our minds and loosen the restrictions on the api).

We can achieve the behavior where serializing a `StreamedRow` that contains protobuf row data yields something closer to `{"row":"CAESBlVTRVJfMA=="}`, but we can never get it to be the same because `row` is already taken up by the regular JSON rows. I hope this makes sense

ksqldb-rest-app/src/test/java/io/confluent/ksql/rest/integration/RestApiTest.java
```diff
@@ -916,6 +919,64 @@ public void shouldExecutePullQueryOverHttp2QueryStream() {
...
+        return Integer.MAX_VALUE;
+      }
+    }, is(6));
+
+    assertTrue(resp[0].body().toString().contains(expectedResponsePart1));
+    assertTrue(resp[0].body().toString().contains(expectedResponsePart2));
+  }
+
+  @Test
+  public void shouldNotExecutePullQueryOverHttp2QueryProto() {
```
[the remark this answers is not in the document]
cprasad1: This PR would've become harder to manage. I can make a follow-up and enable the same format on `/query` as well. For the purposes of this PR, I would vote for shipping this without touching `/query` to unblock the release.

cprasad1:
addressed blocking comments on `KsqlTargetUtil` and made logical schema non optional in the streamed row header

ksqldb-rest-app/src/main/java/io/confluent/ksql/api/server/JsonQueryStreamResponseWriter.java
cprasad1: addressed

ksqldb-rest-app/src/main/java/io/confluent/ksql/api/server/JsonQueryStreamResponseWriter.java
```diff
@@ -54,16 +72,34 @@ public class JsonQueryStreamResponseWriter implements QueryStreamResponseWriter
...
+  public JsonQueryStreamResponseWriter(final HttpServerResponse response, final String mediaType) {
     this.response = Objects.requireNonNull(response);
+    this.mediaType = Objects.requireNonNull(mediaType);
   }
 
   @Override
   public QueryStreamResponseWriter writeMetadata(final QueryResponseMetadata metaData) {
     final Buffer buff = Buffer.buffer().appendByte((byte) '[');
-    buff.appendBuffer(ServerUtils.serializeObject(metaData));
+    if (mediaType.equals(KsqlMediaType.KSQL_V1_PROTOBUF.mediaType())) {
```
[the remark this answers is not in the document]
cprasad1: addressed this. with an enum for the serialization

cprasad1:
I moved most of the serialization logic to `JsonStreamedRowResponseWriter` and refactored it to serialize depending on the response format (proto or json). I also made the proto byte[] and optional field in StreamedRow.DataRow. This also gets rid of the ugly pro parsing logic in `KsqlTargetUtil` since we can now leverage the already existing parser logic there. Ready for review again @agavra @Gerrrr

ksqldb-rest-app/src/main/java/io/confluent/ksql/api/server/JsonStreamedRowResponseWriter.java
```diff
@@ -178,6 +186,72 @@ public void end() {
     response.end();
   }
 
+  public enum RowFormat {
```
agavra, line 189: nice :)

agavra (approved):
Thanks @cprasad1! This looks good to me, we can address the minor follow-ups (if any are leftover) in a future PR

PR #9107, opened 2022-05-11
feat: add support for assert statements to migration tool

jzaralim:
### Description 
What the title says.

Parses `ASSERT` commands in migrations and runs them with the Java client. If the assertion fails, then the migration will be halted.

### Testing done 
Manual, unit, integration

### Reviewer checklist


ksqldb-tools/src/test/java/io/confluent/ksql/tools/migrations/MigrationsTest.java
```diff
@@ -266,6 +268,8 @@ private void shouldApplyMigrations() throws Exception {
         configFilePath,
         migrationsDir,
         "CREATE STREAM ${streamName} (A STRING) WITH (KAFKA_TOPIC='FOO', PARTITIONS=1, VALUE_FORMAT='JSON');\n" +
+            "ASSERT TOPIC FOO WITH (PARTITIONS=1) TIMEOUT 5 SECONDS;\n" +
+            "ASSERT NOT EXISTS SCHEMA ID 3 TIMEOUT 5 SECONDS;\n" +
```
jnh5y, line 272: What would happen to a migration script if an assert failed?

Is that case worth testing?  (Admittedly, it may be a pain.)

PR #9108, opened 2022-05-12
chore: cleanup compiler warnings

mjsax:
Fix type inference, avoid using deprecated methods, and suppress warnings we cannot avoid.

There are still a few type warnings that I did not suppress on purpose, because it's newer (unfinished) code (part of planner rewrite) and we should try to clean it up properly instead of suppressing those warnings.

There is also still the usage of deprecated `StreamsConfig.CACHE_MAX_BYTES_BUFFERING_CONFIG` -- I was not sure how to address it. Any proposals?

### DIFF
--- a/ksqldb-engine/src/main/java/io/confluent/ksql/utilization/PersistentQuerySaturationMetrics.java
+++ b/ksqldb-engine/src/main/java/io/confluent/ksql/utilization/PersistentQuerySaturationMetrics.java
@@ -55,8 +55,8 @@ public class PersistentQuerySaturationMetrics implements Runnable {
   private static final String STREAMS_THREAD_METRICS_GROUP = "stream-thread-metrics";
   private static final String THREAD_ID = "thread-id";
   private static final String QUERY_ID = "query-id";
-  private static Map<String, String> customTags = new HashMap<>();
 
+  private Map<String, String> customTags;
   private final Map<String, KafkaStreamsSaturation> perKafkaStreamsStats = new HashMap<>();
   private final KsqlExecutionContext engine;
   private final Supplier<Instant> time;
@@ -88,8 +88,7 @@ public class PersistentQuerySaturationMetrics implements Runnable {
     this.reporter = Objects.requireNonNull(reporter, "reporter");
     this.window = Objects.requireNonNull(window, "window");
     this.sampleMargin = Objects.requireNonNull(sampleMargin, "sampleMargin");
-    PersistentQuerySaturationMetrics.customTags =
-        Objects.requireNonNull(customTags, "customTags");
+    this.customTags = Objects.requireNonNull(customTags, "customTags");
   }
 
   @Override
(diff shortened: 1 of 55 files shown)


jnh5y:
Relative to the deprecation of `CACHE_MAX_BYTES_BUFFERING_CONFIG`; it has been replaced by `STATESTORE_CACHE_MAX_BYTES_CONFIG` with this commit: https://github.com/apache/kafka/commit/0924fd3f9f75c446310ed1e97b44bbc3f33c6c31 which implemented KIP-770: https://cwiki.apache.org/confluence/pages/viewpage.action?pageId=186878390.

I checked the code and the latter wins out if both happen to be set.  I think we can safely switch to the new config.

ksqldb-engine/src/main/java/io/confluent/ksql/utilization/PersistentQuerySaturationMetrics.java
```diff
@@ -88,7 +88,8 @@ public PersistentQuerySaturationMetrics(
     this.reporter = Objects.requireNonNull(reporter, "reporter");
     this.window = Objects.requireNonNull(window, "window");
     this.sampleMargin = Objects.requireNonNull(sampleMargin, "sampleMargin");
-    this.customTags = Objects.requireNonNull(customTags, "customTags");
+    PersistentQuerySaturationMetrics.customTags =
```
jnh5y, line 91: This is correct to fix the warning.  The fact that shared static state is updated gives me pause.

Is it worth thinking about this?  I'm unclear on how this class is used.

jnh5y (approved):
Looks good to me; I had a few questions which are not blocking. :shipit:

ksqldb-engine/src/main/java/io/confluent/ksql/utilization/PersistentQuerySaturationMetrics.java
mjsax: Yeah -- my understanding is, that it's actually a singleton (so it "ok", meaning nothing breaks; not "ok" from a code layout POV) -- I did not want to do more refactoring to convert it into a real singleton. Might be best to refactor in a follow up PR?

\cc @rodesai @lct45 who worked on this code.

mjsax:
> Is it a goal to turn a check so that warnings are errors in the future?

Personally, I would hope though...

> I think we can safely switch to the new config.

Not sure if it's that easy. The question is, how do we handle the case when a user has the old-config in some config file, or use it via `set` command? Do we accept it and forward to the deprecates KS config? Or do we translate into the new KS config internally? How can we tell uses about the deprecation? For `set`, for CLI interaction we could just reject the old config and point to the new one -- however, `set` could also be used in some scripts and I think we should not break those?

jnh5y:
> > I think we can safely switch to the new config.
> 
> Not sure if it's that easy. The question is, how do we handle the case when a user has the old-config in some config file, or use it via `set` command? Do we accept it and forward to the deprecates KS config? Or do we translate into the new KS config internally? How can we tell uses about the deprecation? For `set`, for CLI interaction we could just reject the old config and point to the new one -- however, `set` could also be used in some scripts and I think we should not break those?

I like the idea of translating the configuration from old to new.

Given that complexity, it seems likely that the configuration should be sorted in another PR.

ksqldb-engine/src/main/java/io/confluent/ksql/utilization/PersistentQuerySaturationMetrics.java
lct45: Thanks @mjsax for the crash course on a singleton. @rodesai may be more equipped to answer this fully but yeah this is essentially a singleton without proper singleton implementation. We create this one in the `KsqlRestApplication` as a single reporter for these metrics and as far as I know, we wouldn't need a second instance

mjsax: I talked to Rohan and he proposed to rewrite the code not make the field non-static. I'll just update this PR accordingly.

mjsax:
> Given that complexity, it seems likely that the configuration should be sorted in another PR.

Cool. I filed a ticket: https://github.com/confluentinc/ksql/issues/9139

PR #9120, opened 2022-05-13
feat: Support all wildcard (*) on struct reference syntax

spena:
### Description 
This PR supports the following struct reference `select struct_col->*` which selects all fields of `struct_col`. Each field will be converted to a column.

i.e.
```
CREATE STREAM test(A INT, B STRUCT<F1 INTEGER, F2 INTEGER>) WITH(kafka_topic='test', value_format='json', partitinos=1);

ksql> select b->* from test;
+---------------------------------------------------------------------------+---------------------------------------------------------------------------+
|F1                                                                         |F2                                                                         |
+---------------------------------------------------------------------------+---------------------------------------------------------------------------+
|10                                                                         |11                                                                         |
Query Completed
```

### Testing done 
Several. I will add QTT tests.

### Reviewer checklist


config/ksql-server.properties
```diff
@@ -78,4 +78,4 @@ compression.type=snappy
 #------ Schema Registry -------
 
 # Uncomment and complete the following to enable KSQL's integration to the Confluent Schema Registry:
-# ksql.schema.registry.url=http://localhost:8081
+ksql.schema.registry.url=http://localhost:8081
```
jnh5y, line 81: This will have to be backed out to merge, right?

ksqldb-engine/src/main/java/io/confluent/ksql/planner/plan/SelectionUtil.java
```diff
@@ -208,4 +216,99 @@ private static Stream<SelectExpression> resolveAllColumns(
...
+    //
+    // 1. DereferenceExpression. Contains struct columns with -> references. i.e. `col0->l1`
+    // 2. ColumnReferenceExp. Contains column references only. i.e. `col0`
+    //
+    // This resolver goes back to the column reference, so it has access to the column schema. It
+    // then goes forward each struct field, specified in the DereferenceExpression, to get
+    // access to the nested fields.
+
+    // This is the column part of the parent schema, which is used to access the column type
+    if (expression instanceof ColumnReferenceExp) {
```
jnh5y, line 272: Ok, so are qualified and unqualified reference expressions handled ok in here?

ksqldb-engine/src/main/java/io/confluent/ksql/planner/plan/SelectionUtil.java
```diff
@@ -208,4 +216,99 @@ private static Stream<SelectExpression> resolveAllColumns(
...
+        } else {
+          throw new IllegalArgumentException("Field " + fieldName + " is not a STRUCT type.");
+        }
+      } else {
+        throw new IllegalArgumentException("Field " + fieldName + " was not found on STRUCT.");
+      }
+    }
+
+    throw new IllegalArgumentException(
+        "Unsupported struct column expression: " + expression.getClass().getName());
```
jnh5y, line 312: What about subscript expressions?  That'd cover resolving maps and arrays, right?

jnh5y, line 312: What if a function returned a struct?  Could one use ->* to expand it?  E.g. should `FunctionCall` be handled?

jnh5y, line 312: The only other `Expression` I'm mildly amused by is the `Case` Expressions...  that is probably a bit of stretch though?

ksqldb-parser/src/main/antlr4/io/confluent/ksql/parser/SqlBase.g4
```diff
@@ -223,9 +223,10 @@ values
     ;
 
 selectItem
-    : expression (AS? identifier)?  #selectSingle
-    | identifier '.' ASTERISK       #selectAll
-    | ASTERISK                      #selectAll
+    : expression (AS? identifier)?                       #selectSingle
+    | base=primaryExpression STRUCT_FIELD_REF ASTERISK   #selectStructAll
```
jnh5y, line 227: Ah.... for when we try and catch all the cases... 

https://github.com/confluentinc/ksql/blob/master/ksqldb-parser/src/main/antlr4/io/confluent/ksql/parser/SqlBase.g4#L312-L328

ksqldb-parser/src/main/java/io/confluent/ksql/parser/tree/StructAll.java
```diff
@@ -0,0 +1,72 @@
...
+public class StructAll extends SelectItem {
+  private final Expression baseStruct;
+
+  public StructAll(final Expression identifier) {
+    this(Optional.empty(), identifier);
+  }
+
+  public StructAll(final Optional<NodeLocation> location, final Expression baseStruct) {
+    super(location);
+    this.baseStruct = requireNonNull(baseStruct, "identifier");
```
jnh5y, line 36: Hmmm... where do we validate that the `baseStruct` is a Struct?

ksqldb-parser/src/test/java/io/confluent/ksql/parser/AstBuilderTest.java
```diff
@@ -206,6 +208,41 @@ public void shouldExtractAsAliasedJoinDataSources() {
...
+    final Query result = (Query) builder.buildStatement(stmt);
+
+    // Then:
+    assertThat(result.getSelect(), is(new Select(ImmutableList.of(
+        new StructAll(column("NESTED_ORDER_COL"))
+    ))));
+  }
+
+  @Test
+  public void shouldHandleSelectStructAllOnNestedStruct() {
```
jnh5y, line 227: Ok, so this is the class where we can run through the various cases we want to make sure and cover.  (Like arrays, maps, functions, whatever...)

jnh5y (commented):
Looking good so far.  Started some discussions.

config/ksql-server.properties
spena: I missed it. Just updated the PR.

ksqldb-engine/src/main/java/io/confluent/ksql/planner/plan/SelectionUtil.java
```diff
@@ -208,4 +216,99 @@ private static Stream<SelectExpression> resolveAllColumns(
...
+    );
+
+    // The selectExpression produced above should return one single expression
+    final Expression expression = Streams.findLast(selectExpression).get().getExpression();
+
+    // Find all fields of the most nested struct referenced in the expression.
+    // i.e. `col0->l1->l2->*` should resolve all fields of `col0->l1->f2`
+    final SqlStruct mostNestedStructFields = resolveStructFields(parentSchema, expression);
+
+    // Now convert all fields to a SelectExpression
```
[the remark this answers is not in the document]
spena: Aha, that's good case. All columns and fields must be back quoated in order to allow special strings, 
i.e.
```
ksql> select `a 1`->`f 1` from input;
+--------------------------------------------------------------------------------------------------------------------------------------------------------+
|f 1                                                                                                                                                     |
+--------------------------------------------------------------------------------------------------------------------------------------------------------+
|1                                                                                                                                                       |
Query Completed
```

But I just found the wildcard does not detect that.
```
ksql> select `a 1`->* from input;
The server encountered an internal error when processing the query. Please consult the server logs for more information.
```

I'll fix it. Thanks for the catch.

ksqldb-engine/src/main/java/io/confluent/ksql/planner/plan/SelectionUtil.java
spena: Yes, they're handled ok. Both expressions inherit from ColumnReferenceExp

ksqldb-engine/src/main/java/io/confluent/ksql/planner/plan/SelectionUtil.java
spena: There's a different code path that detects invalid references to non-struct columns. By the time the expression got into this code, the validation was done. The error message is not useful, though. Try to run `select array->f` in ksql without this PR and you get a server error message. I'll report a bug to fix it in a different issue.

ksqldb-engine/src/main/java/io/confluent/ksql/planner/plan/SelectionUtil.java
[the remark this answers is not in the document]
spena: Just updated the PR and added a QTT test

ksqldb-parser/src/main/antlr4/io/confluent/ksql/parser/SqlBase.g4
jnh5y, line 227: Separate question... how is whitespace handled?

Should `primaryExpress -> *` work?

jnh5y:
Since this is trying to handle the case where there are multiple types on a given topic, what happens if one wanted to join across two streams each with multiple types?  Could the `->*` operator be used from both sides in a join sensibly?

jnh5y (approved):

ksqldb-parser/src/main/antlr4/io/confluent/ksql/parser/SqlBase.g4
spena: ANTlr handles that for us. I think it ignores spaces. I tested the ` -> ` and it works.

ksqldb-functional-tests/src/test/resources/query-validation-tests/select-struct-all.json
```diff
@@ -0,0 +1,120 @@
...
+      "outputs": [
+        {"topic":  "OUTPUT", "key": null, "value": {"A": 1, "B": 2}}
+      ]
+    },
+    {
+      "name": "joins and struct all on projection",
+      "statements": [
+        "CREATE STREAM INPUT_1 (k1 INT KEY, v1 STRUCT<a INT, b INT>) WITH (kafka_topic='test_topic_1', format='JSON');",
+        "CREATE STREAM INPUT_2 (k1 INT KEY, v1 STRUCT<a INT, b INT>) WITH (kafka_topic='test_topic_2', format='JSON');",
+        "CREATE STREAM OUTPUT AS SELECT I1.k1, I1.v1->*, I2.v1->a AS I2_A, I2.v1->b AS I2_B FROM INPUT_1 I1 JOIN INPUT_2 I2 WITHIN (1 SECOND, 100 SECONDS) ON I1.k1 = I2.k1;"
```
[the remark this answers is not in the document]
spena: It's those kind of errors that must be fixed. The error is only shown in the Server logs, but the CLI shows just a server error has occurred. This same thing happens with `SELECT *, * FROM...` even before this PR, so it's not a regression. I will open new GH issues later to improve some error messages I saw.

PR #9130, opened 2022-05-18
fix: use JsonSchemaConverter to support JSON anyOf types

spena:
### Description 
More details later

### Testing done 

### Reviewer checklist


ksqldb-serde/src/main/java/io/confluent/ksql/serde/connect/ConnectDataTranslator.java
```diff
@@ -232,6 +233,12 @@ private Object toKsqlValue(
       case STRING:
         // use String.valueOf to convert various int types and Boolean to string
         return String.valueOf(convertedValue);
+      case BYTES:
```
jnh5y, line 236: Ok, this looks reasonable, but it also looks like it could change things.  

What all does this interact with?

ksqldb-serde/src/main/java/io/confluent/ksql/serde/json/JsonSchemaTranslator.java
```diff
@@ -15,19 +15,28 @@
...
+import org.apache.kafka.connect.json.DecimalFormat;
+import org.apache.kafka.connect.json.JsonConverterConfig;
 
 /**
  * Translates between Connect and JSON Schema Registry schema types.
  */
-class JsonSchemaTranslator implements ConnectSchemaTranslator {
-
-  private final JsonSchemaData jsonData = new JsonSchemaData();
+public class JsonSchemaTranslator implements ConnectSchemaTranslator {
```
jnh5y, line 32: Is there any discussion we need to have about making this class public?

ksqldb-serde/src/main/java/io/confluent/ksql/serde/json/KsqlJsonSerdeFactory.java
```diff
@@ -44,20 +45,20 @@
 
 @SuppressWarnings("checkstyle:ClassDataAbstractionCoupling")
 @Immutable
-class KsqlJsonSerdeFactory implements SerdeFactory {
+public class KsqlJsonSerdeFactory implements SerdeFactory {
```
jnh5y, line 48: Ditto.  I imagine we need to make this class public for things to work?

ksqldb-serde/src/test/java/io/confluent/ksql/serde/json/KsqlJsonDeserializerTest.java
```diff
@@ -131,22 +130,16 @@ public class KsqlJsonDeserializerTest {
           .addSerializer(java.sql.Date.class, new EpochDaySerializer())
       );
 
-  @Parameters(name = "{0}")
-  public static Collection<Object[]> data() {
-    return Arrays.asList(new Object[][]{{"Plain JSON", false}, {"Magic byte prefixed", true}});
```
jnh5y, line 136: Ok, so do we still test both of these options/parameters?

If not, why?

ksqldb-serde/src/test/java/io/confluent/ksql/serde/json/KsqlJsonDeserializerTest.java
```diff
@@ -1115,11 +1126,7 @@ private byte[] serializeJson(final Object expected) {
   }
 
   private byte[] addMagic(final byte[] json) {
-    if (useSchemas) {
```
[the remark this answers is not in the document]
spena: The majority of these tests do not apply anymore. The `JsonSchemaConverter` does validations with the SR schema before serializing or deserializing. These tests are making sure those validations are working for plain JSON. It's hard to make tests work for JSON_SR, for instance, I cannot force a Connect schema with a JSON-SR schema to have a field that does not exist. The Connect schema object fails. So, I decided to remove any `useSchemas` from this class and write my own on `KsqlJsonSchemaDeserializerTest`.

ksqldb-serde/src/test/java/io/confluent/ksql/serde/json/KsqlJsonDeserializerTest.java
spena: Not anymore. These test class was testing JSON and JSON_SR. The new JsonSchemaConverter does not use like 95% of these tests. I created a new `KsqlJsonSchemaDeserializerTest` class for that. I will add tests on it the more I understand JSON_SR.

ksqldb-serde/src/main/java/io/confluent/ksql/serde/json/JsonSchemaTranslator.java
spena: I need this class for the QTT tests. Proto and Avro schema translators are also public for QTT.

ksqldb-serde/src/main/java/io/confluent/ksql/serde/json/KsqlJsonSerdeFactory.java
spena: I updated it.

ksqldb-serde/src/main/java/io/confluent/ksql/serde/connect/ConnectDataTranslator.java
spena: I thought the same. I am not sure why this part didn't wrap the bytes to a ByteBuffer before if KSQL uses ByteBuffer for the BYTES column type. So far, only the `AvroDataTranslator` was failing on this test, but I updated it. I also run some QTT tests and run all formats (avro, json, json-sr, proto, delimited) manually with simple, array, map and struct bytes and none of them failed. So I don't know if this code was just missed when supporting the BYTES type.

ksqldb-functional-tests/src/test/resources/historical_plans/json_sr_-_ARRAY_-_key_-_inference/6.2.0_1607115574018/spec.json
```diff
@@ -87,7 +87,7 @@
       "valueSchema" : {
         "type" : "object",
         "properties" : {
-          "COL1" : {
```
spena, line 90: This plan and the others related to `key - inference` do not apply to JsonSchemaConverter anymore. The QTT were failing when comparing the JSON-SR schemas. However, I tried running the same test in Prod and things work fine. 

I created a JSON-SR schema with a `COL1 INTEGER` field:
Schema
```
{
  "properties": {
    "COL1": {
      "description": "The integer type is used for integral numbers.",
      "type": "integer"
    }
  },
  "type": "object"
}
```

I then created a stream with a `FOO INTEGER` using the above schema, then insert/select from the stream which it worked:
```
ksql> create stream json(foo INT) with (kafka_topic='json', value_format='JSON_SR');

 Message        
----------------
 Stream created 
----------------

ksql> insert into json(foo) values(1);

ksql> select * from json;
+----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+
|FOO                                                                                                                                                                                     |
+----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+
|1                                                                                                                                                                                       |
Query Completed
```

I noticed the SR schema changed now to after inserting the values:
```
{
  "properties": {
    "FOO": {
      "connect.index": 0,
      "oneOf": [
        {
          "type": "null"
        },
        {
          "connect.type": "int32",
          "type": "integer"
        }
      ]
    }
  },
  "type": "object"
}
```

Perhaps the above changes (renaming the field) is compatible with JSON_SR only? I tried with Protobuf/Avro schemas and they do not allow the QTT test case to pass. 

I modified the QTT plans to make QTT to work because Proto/Avro dot now allow such test and JSON_SR allowed me to do it in Production. I don't think there will be an incompatibility problem.

ksqldb-functional-tests/src/test/resources/historical_plans/json_sr_-_deserialize_anonymous_array_-_value/6.1.0_1597874026054/spec.json
```diff
@@ -120,24 +114,18 @@
...
-                    "oneOf" : [ {
-                      "type" : "null"
-                    }, {
-                      "type" : "string"
-                    } ]
-                  }
+                  "type" : "string"
                 } ]
               }
-            }
```
spena, line 140: This is another incompatibility issue with QTT. 

The JSON_SR schema has an ARRAY field wrapped into another `FOO` object which should be `STRUCT<FOO ARRAY<STRING>>`. When creating a JSON-SR stream with `WRAP_SINGLE_VALUE=false`, the internal array is unwrapped and used in the stream as `FOO ARRAY<STRING>` only.  

However, JSON_SR with JsonSchemaConverter does not work like that. JSON_SR supports anonymous fields by registering a schema without a field name. This makes SR API to understands the record is anonymous and does not have a field. When attempting to deserialize an anonymous record using the SR schema with the wrapped array, then the deserialization fails because the schema is not the one used during serialization. However, before this PR, JSON_SR deserialization was done using the JSON mapper (no JsonSchemaConverter) which ignores the SR schema. This deserialization was using the KSQL schema to deserialize and everything looked good. Running INSERT VALUES fails, though, even before the PR 'cause serialization uses the JsonSchemaConverter.

I found the above incompatibility was only on QTT, though. If I run the tests in Production, when I serialize a value with wrapped fields, the JsonSchemaConverter adds the SR schema ID in the record for future reference when deserializing it. When deserializing it with JsonSchemaConverter, the API will look at the right schema ID and deserialize it using that schema, then pass it to KSQL which will unwrap the field. 

Seems the test case is compatible in Prod but not in QTT. This is another type behavior that QTT fails to validate corrrectly. I think we should fix QTT to use the Prod serializers instead of its own ones.

ksqldb-functional-tests/src/main/java/io/confluent/ksql/test/serde/ConnectSerdeSupplier.java
```diff
@@ -87,14 +88,30 @@ public byte[] serialize(final String topic, final Object spec) {
...
+        id = srClient.getLatestSchemaMetadata(subject).getId();
         schema = (T) srClient.getSchemaBySubjectAndId(subject, id);
       } catch (Exception e) {
         throw new KsqlException(e);
       }
 
+      // Do not let the serializer to auto register a new schema on JSON_SR formats.
+      // QTT tests with JSON_SR started failing on historic plans after supporting the
+      // JsonSchemaConverter because the converter was auto registering a schema with extra
+      // properties. These new props are not required, but was causing noise on the historic plans.
```
lihaosky, line 103: Prod code when you insert still registers new schema for json_sr format, why we don't have problem there?

ksqldb-functional-tests/src/main/java/io/confluent/ksql/test/serde/json/ValueSpecJsonSchemaSerdeSupplier.java
```diff
@@ -0,0 +1,102 @@
...
+      final Object deserialized = internalDeserializer.deserialize(topic, bytes);
+      if (!(deserialized instanceof Map)) {
+        return deserialized;
+      }
+
+      final Map<String, Object> deserializedMap = (Map<String, Object>) deserialized;
+
+      // This makes JSON_SR output compatible with old QTT tests that used to deserialize
+      // JSON_SR with a JSON object mapper instead of the ConnectSerdeSupplier.
+      // The ConnectSerdeSupplier is adding the missing fields with null values even if they're
```
lihaosky, line 90: QQ: why the field is serialized to map if it's not in schema? I might miss something...

ksqldb-functional-tests/src/test/java/io/confluent/ksql/test/serde/json/ValueSpecJsonSerdeSupplierTest.java
```diff
@@ -72,32 +68,4 @@ public void shouldDeserializeDecimalsWithoutStrippingTrailingZeros_Plain() {
...
-    final byte[] bytes = new byte[jsonBytes.length + JsonSerdeUtils.SIZE_OF_SR_PREFIX];
-    System.arraycopy(jsonBytes, 0, bytes, JsonSerdeUtils.SIZE_OF_SR_PREFIX, jsonBytes.length);
-
-    // When:
-    final Object result = deserializer.deserialize("t", bytes);
-
-    // Then:
-    assertThat(result, is(new BigDecimal("10.0")));
-  }
 }
```
lihaosky, line 71: Should they be move to somewhere else instead of deleted?

ksqldb-serde/src/main/java/io/confluent/ksql/serde/connect/ConnectDataTranslator.java
lihaosky: is this related to `JsonSchemaConverter`? If not, I guess we can separate this out to a different PR. The reason is `JsonSchemaConverter` changes are already complex and probably we don't want to mix with other fixes if not necessary.

lihaosky (commented):
Thanks @spena

ksqldb-functional-tests/src/test/resources/historical_plans/json_sr_-_ARRAY_-_key_-_inference/6.2.0_1607115574018/spec.json
lihaosky: Does this after this change, insertion on existing json_sr won't work because of incompatible schema?

ksqldb-serde/src/main/java/io/confluent/ksql/serde/json/KsqlJsonSerdeFactory.java
```diff
@@ -163,6 +168,9 @@ private static Converter getSchemaConverter(
     }
     config.put(JsonConverterConfig.DECIMAL_FORMAT_CONFIG, DecimalFormat.NUMERIC.name());
 
+    // Makes naming of unions consistent between all SR formats (i.e. connect_union_field_0)
+    config.put(JsonSchemaDataConfig.GENERALIZED_SUM_TYPE_SUPPORT_CONFIG, true);
```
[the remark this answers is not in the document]
lihaosky: Does https://github.com/confluentinc/ksql/blob/master/ksqldb-serde/src/main/java/io/confluent/ksql/serde/json/JsonSchemaTranslator.java#L48 also need this config? I think this is called when we first register the schema in SR for `ct/cs`.

ksqldb-serde/src/main/java/io/confluent/ksql/serde/json/KsqlJsonDeserializer.java
```diff
@@ -117,17 +126,18 @@ public T deserialize(final String topic, final byte[] bytes) {
...
-      // don't use the JsonSchemaConverter to read this data because
-      // we require that the MAPPER enables USE_BIG_DECIMAL_FOR_FLOATS,
-      // which is not currently available in the standard converters
-      final JsonNode value = isJsonSchema
-          ? JsonSerdeUtils.readJsonSR(bytes, MAPPER, JsonNode.class)
-          : MAPPER.readTree(bytes);
+      final Object coerced;
+      if (isJsonSchema) {
+        final SchemaAndValue schemaAndValue = converter.toConnectData(topic, bytes);
+        coerced = translator.toKsqlRow(schemaAndValue.schema(), schemaAndValue.value());
```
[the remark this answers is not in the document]
spena: Done

ksqldb-serde/src/main/java/io/confluent/ksql/serde/json/KsqlJsonSerdeFactory.java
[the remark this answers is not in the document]
spena: I needed for the `oneOf` support on JSON_SR. I removed for this PR to reduce some complexity. I will add more cases when testing multiple json schema events.

ksqldb-functional-tests/src/test/java/io/confluent/ksql/test/serde/json/ValueSpecJsonSerdeSupplierTest.java
spena: I moved it to `ValueSpecJsonSchemaSerdeSupplierTest`

ksqldb-serde/src/test/java/io/confluent/ksql/serde/json/KsqlJsonDeserializerTest.java
```diff
@@ -904,65 +896,6 @@ public void shouldThrowIfCanNotCoerceMapValue() {
         "Can't convert type. sourceType: BooleanNode, requiredType: INTEGER"))));
   }
 
-  @Test
-  public void shouldThrowOnMapSchemaWithNonStringKeys() {
```
spena, line 915: These tests are available already in `KsqlJsonSerdeFactoryTest.java`

ksqldb-functional-tests/src/main/java/io/confluent/ksql/test/serde/json/ValueSpecJsonSchemaSerdeSupplier.java
spena: This is what the ConnectSerdeSupplier is returning. I got lost on this part for some time. After debugging I found out that QTT creates a Map with the columns and its deserialized values. This is the part that is special for QTT.

ksqldb-functional-tests/src/main/java/io/confluent/ksql/test/serde/ConnectSerdeSupplier.java
spena: It's caused by the historical plans with wrap singles = false. For instance, in 6.1, the SR schema used by the historical plan is `FOO STRUCT<ARRAY<STRING>>`; but when the wrap_single_value=false is used, then KSQL will attempt to serialize the schema `ARRAY<STRING>` which the SR API will try to register. SR will fail because the schema is not compatible with the latest schema that has a STRUCT.

All this is caused by QTT. I can explain you in more detail how QTT was working before offline. For QTT to work, I had to prevent QTT to register the schema as it was already registered before.

ksqldb-serde/src/main/java/io/confluent/ksql/serde/connect/ConnectDataTranslator.java
lihaosky: Again, if this is not related to this PR, we can separate it out to another PR

ksqldb-serde/src/test/java/io/confluent/ksql/serde/json/KsqlJsonSchemaDeserializerTest.java
```diff
@@ -0,0 +1,149 @@
...
+        Struct.class,
+        false
+    );
+
+    serializer = serde.serializer();
+    deserializer = serde.deserializer();
+  }
+
+  @Test
+  public void shouldDeserializeJsonObjectCorrectly() {
```
lihaosky, line 134: Can you add a test for `validateSchema` failure?

spena: The `validateSchema` test was already in `KsqlJsonSerdeFactoryTest`. I added the fflag test too.

spena:
@lihaosky I addressed all the comments in the latest 2 commits. The latest one has a bunch of QTT plans updated so I could remove the code from your comments. Th removal only included `connect.type`. 


> Is it possible to update historic plans to include the extra fields instead of adding the special logic? Even though, I'm not sure why we don't use production code to do the serialization, but I guess we should better mimic what prod is doing
I just did. The latest commit has several historical plans updated. The changes were just the `connect.type` property that it was something added and sometimes removed. Besides that, everything else worked fine. This `connect.type` (when it is added or not) is still a mystery.

> (Regarding the BYTES type) Again, if this is not related to this PR, we can separate it out to another PR
It is needed. The old JSON deserializer was working different. After using the JsonSchemaConverter, the BYTES types were failing due to this.

> I think the old post schema is wrong. If wrap_single_valuer=false, then FOO shouldn't be in the schema. I guess the reason why FOO is in SR in QTT previously is because the schema is registered by QTT manually? Now the ksqldb-functional-tests/src/main/java/io/confluent/ksql/test/serde/json/ValueSpecJsonSchemaSerdeSupplier.java you created will register the correct schema again during insert?
I think so. I'm still not sure how the 6.1 plan was there. I talked with Robert and he thinks there was a bug in 6.1 in ksql that it is now fixed.

ksqldb-functional-tests/src/test/resources/historical_plans/elements_-_validate_schema_id_without_elements_OK_-_JSON_SR_SCHEMA/7.1.0_1636006774347/spec.json
```diff
@@ -44,7 +44,9 @@
         "type" : "object",
         "properties" : {
           "c1" : {
-            "type" : "integer"
+            "type" : "integer",
+            "connect.index": 0,
+            "connect.type": "int64"
```
lihaosky, line 49: Is this compatible with schema without connect meta. A good test might be:
1. Create stream with json_sr format in SR without this fix.
2. Insert into the stream.
3. With this fix, insert into stream again. Since insertion will try to register schema with connect metadata now, we can see if it's compatible with old schema.

I'm ok if you verify this manually.

ksqldb-serde/src/main/java/io/confluent/ksql/serde/json/KsqlJsonSerdeFactory.java
```diff
@@ -124,14 +133,33 @@ private <T> Serializer<T> createSerializer(
...
+    if (useSchemaRegistryFormat
+        && ksqlConfig.getBoolean(KsqlConfig.KSQL_JSON_SR_CONVERTER_DESERIALIZER_ENABLED)) {
+      return new KsqlConnectDeserializer<>(
+          converter,
+          dataTranslator,
+          targetType
+      );
+    } else {
+      if (useSchemaRegistryFormat) {
+        LOG.info("The JsonSchemaConverter for deserialization is disabled for JSON_SR.");
```
lihaosky, line 151: Remove this log? It's possible this serde will be called many time, this log can spam log. I guess this can be deduced from the value of `KSQL_JSON_SR_CONVERTER_DESERIALIZER_ENABLED` which might be logged as config.

lihaosky (commented):
Thanks @spena ! LGTM overall just a few comments.
Can you also comment on https://github.com/confluentinc/ksql/pull/9130/files#r922437085? I don't fully understand that.

ksqldb-serde/src/main/java/io/confluent/ksql/serde/json/KsqlJsonSerdeFactory.java
spena: I removed it. I was hesitating to add it in the first place. But you're right, we can deduct this by looking at the config.

ksqldb-serde/src/test/java/io/confluent/ksql/serde/json/KsqlJsonSchemaDeserializerTest.java
```diff
@@ -0,0 +1,149 @@
...
+
+  @Before
+  public void before() throws Exception {
+    schema = (new JsonSchemaTranslator()).fromConnectSchema(ORDER_SCHEMA.schema());
+
+    schemaRegistryClient = mock(SchemaRegistryClient.class);
+    when(schemaRegistryClient.getSchemaBySubjectAndId(anyString(), anyInt())).thenReturn(schema);
+
+    converter = new JsonSchemaConverter(schemaRegistryClient);
+    converter.configure(ImmutableMap.of(
```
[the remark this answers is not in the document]
spena: I removed it

ksqldb-functional-tests/src/test/resources/historical_plans/elements_-_validate_schema_id_without_elements_OK_-_JSON_SR_SCHEMA/7.1.0_1636006774347/spec.json
spena: I figured out why QTT is behaving different. Before this fix, QTT was using the plain JSON serialiazer which does not handle SR schema. It just uses the one from the historical plan and serializes the record. 

With the fix, I had to change QTT to use the JsonSchemaConvert (which is used in Prod). In Prod, when the schema ID is present, then it does not register the schema again, but in QTT we register it 'cause the `ConnectSerdeSupplier` does not add the schema ID to the converter. That's what the commit I reverted used to do (add the schema ID to prevent register the schema).

So, I either re-revert the commit that used to do that so the historical plans are unchanged or we stick with these changes. What do you think?

ksqldb-functional-tests/src/test/resources/historical_plans/json_sr_-_ARRAY_-_key_-_inference/6.2.0_1607115574018/spec.json
lihaosky: Synced offline, why `col1` was there previously was a mystery but `FOO` is correct now. @spena also test in prod and there's no issue of `col1`. So ignore this...

ksqldb-functional-tests/src/test/resources/historical_plans/elements_-_validate_schema_id_without_elements_OK_-_JSON_SR_SCHEMA/7.1.0_1636006774347/spec.json
lihaosky: We can stick to current code then. Adding logic in test framework in order to make test work seems not right...e ven test framework logic isn't what prod code is doing. cc @suhas-satish , @colinhicks

lihaosky (approved):
lgtm

PR #9134, opened 2022-05-19
feat: cull the list of API consumable/editable properties

rodesai:
This patch culls the list of properties that can be read or edited over
the HTTP API. We want to restrict this set to only properties that affect
semantics or query performance. All other properties should be set at
the server level.

The basic framework for how how we should think about this is as follows.
Configs control one of three things (there's some overlap, but I think this is
mostly true):
- availability: these configurations control whether or not a query can start
   and do it's work. Examples would be bootstrap servers, ssl configs, connection
   timeouts, state store dir, etc.
- cost/performance: these configurations control tradeoffs in query performance.
   Examples would be cache.max.bytes.buffering, poll interval, batch sizes, etc. I think
   standby replicas also falls in this category as it lets the user make a cost vs recovery
   time tradeoff.
- semantics: these configuration control query processing semantics. Examples include
   auto.offset.reset or ksql.nested.error.set.null.

We should only support configuring performance and semantics configs over the API,
and we should require you to configure any availability impacting configs using the
properties file. This aligns with a model where there is an operator and user persona,
where the operator is responsible for operating the service and keeping it up, and the
user issues queries and tunes their performance/behavior.

I'm looking for feedback on:
- is the categorization above reasonable?
- is the proposal to omit availability-related properties from the API reasonable?
- is the assignment of configs to categories suggested in the PR reasonable?


colinhicks (approved):
Thanks for the careful reasoning and context. I agree with the framework/rationale and the changes.

ksqldb-rest-model/src/main/java/io/confluent/ksql/rest/entity/PropertiesList.java
```diff
@@ -134,67 +132,27 @@ public class PropertiesList extends KsqlEntity {
...
       MAX_WARMUP_REPLICAS_CONFIG,
       NUM_STANDBY_REPLICAS_CONFIG,
       POLL_MS_CONFIG,
       PROBING_REBALANCE_INTERVAL_MS_CONFIG,
       PROCESSING_GUARANTEE_CONFIG,
-      REPLICATION_FACTOR_CONFIG,
-      ROCKSDB_CONFIG_SETTER_CLASS_CONFIG,
-      SECURITY_PROTOCOL_CONFIG,
-      STATE_CLEANUP_DELAY_MS_CONFIG,
       STATE_DIR_CONFIG,
```
JSeb225, line 153: Does this one make sense in the cloud?

JSeb225 (commented):
Classification/principle looks good to me. I couldn't find the exact mapping with the customer-facing config name for all yet, but noticed one that may not be needed in the cloud (added comment).

ksqldb-rest-model/src/main/java/io/confluent/ksql/rest/entity/PropertiesList.java
rodesai: oh yeah we should clean that one up

ableegoldman (commented):
Thanks for the PR, I've been a proponent of limiting the available configs since this whole thing kicked off and hope we can include this in the 0.26 hotfix/whatever version we ultimately roll out to GA

Regarding the classification, those three categories make sense to me but I wonder if we may need to carve out the gray area between "availability" and "performance into its own category, let's call it "operation" for now because I can't think of anything better at the moment.

By this I mean configs that might tangentially affect the performance but generally should be configurable to ensure availability eg when environmental conditions change. This is mainly timeout and backoff configs, for example `TRANSACTION_TIMEOUT_CONFIG`, `REQUEST_TIMEOUT_MS_CONFIG`, `RETRY_BACKOFF_MS_CONFIG`, etc

I notice that those and several similar configs are being removed from the system properties list, but others that I would consider in the same boat such as `SESSION_TIMEOUT_MS_CONFIG` and `DELIVERY_TIMEOUT_MS_CONFIG` are not. Is there any particular logic to the timeout configs that were kept vs removed? If not, I would suggest leaving all of these in the system properties list so that we can use `ALTER SYSTEM` to toggle the values eg if transactions or requests begin to time out.

PR #9141, opened 2022-05-21
feat: enable new emit-final implementation

mjsax:


ksqldb-execution/src/main/java/io/confluent/ksql/execution/windows/HoppingWindowExpression.java
```diff
@@ -80,7 +92,7 @@ public String toString() {
         + "ADVANCE BY " + advanceBy
         + retention.map(w -> " , RETENTION " + w).orElse("")
         + gracePeriod.map(g -> " , GRACE PERIOD " + g).orElse("")
-        + " ) ";
+        + " ) " + emitStrategy.map(Enum::toString).orElse("");
```
mjsax, line 95: This is a nasty hack -- did not find a better way to preserve the information.

ksqldb-common/src/main/java/io/confluent/ksql/serde/WindowInfo.java
```diff
@@ -73,12 +85,14 @@ public boolean equals(final Object o) {
       return false;
     }
     final WindowInfo that = (WindowInfo) o;
+    // we omit `emitStrategy`
```
vcrfxia, line 88: Why do we omit this from the comparator?

ksqldb-engine/src/main/java/io/confluent/ksql/ddl/commands/CreateSourceFactory.java
```diff
@@ -142,14 +143,29 @@ public CreateStreamCommand createStreamCommand(
   }
 
   // This method is called by CREATE_AS statements
-  public CreateTableCommand createTableCommand(final KsqlStructuredDataOutputNode outputNode) {
+  public CreateTableCommand createTableCommand(
+      final KsqlStructuredDataOutputNode outputNode,
+      final Optional<RefinementInfo> emitStrategy
```
vcrfxia, line 148: Why is this being added here? Do we even support syntax that says that source tables should be created with EMIT FINAL?

ksqldb-parser/src/main/java/io/confluent/ksql/parser/json/WindowExpressionDeserializer.java
```diff
@@ -20,13 +20,26 @@
...
   @Override
   @SuppressWarnings("unchecked")
   public T deserialize(final JsonParser parser, final DeserializationContext ctx)
       throws IOException {
-    return (T) ExpressionParser.parseWindowExpression(parser.readValueAs(String.class));
+    final String extendedWindowExpression = parser.readValueAs(String.class);
+
+    final KsqlWindowExpression windowExpression;
+
+    if (extendedWindowExpression.endsWith("FINAL")) {
```
vcrfxia, line 35: Instead of this hack, can we update the serializer/deserializer to serialize an additional, optional field to specify the emitStrategy? Ideally there'd be a way to deserialize a window object that is either represented as a raw string (legacy) or as an object (new).

vcrfxia (commented):
Thanks @mjsax -- some initial comments/questions inline.

ksqldb-execution/src/main/java/io/confluent/ksql/execution/windows/HoppingWindowExpression.java
vcrfxia: For other readers: this "hack" is being added so that deserialization works properly here: https://github.com/confluentinc/ksql/pull/9141#discussion_r879937488

ksqldb-common/src/main/java/io/confluent/ksql/serde/WindowInfo.java
mjsax: `WindowInfo` should actually only describe the window itself and some code relies on it -- emit-strategy is not really a window property and it's just bolted on her as it's hard to get a hold on it otherwise.

Because `emit-strategy` could be set on one instance but not on another, while both instances describe the same window, we omit it here to not break the comparison.

(IIRC)

ksqldb-engine/src/main/java/io/confluent/ksql/ddl/commands/CreateSourceFactory.java
mjsax: This is for `CTAS` statements, ie, persistent queries.

ksqldb-common/src/main/java/io/confluent/ksql/config/KsqlConfigResolver.java
```diff
@@ -132,11 +133,21 @@ private static Optional<ConfigItem> resolveConfig(
...
     }
 
-    return Optional.of(ConfigItem.resolved(configKey));
+    return Optional.empty();
+  }
+
+  private static boolean isInternalStreamsConfig(final String key) {
+    // add more internal configs on demand
+    return key.equals(InternalConfig.EMIT_INTERVAL_MS_KSTREAMS_OUTER_JOIN_SPURIOUS_RESULTS_FIX)
+        || key.equals(InternalConfig.EMIT_INTERVAL_MS_KSTREAMS_WINDOWED_AGGREGATION);
```
mjsax, line 150: Need this one to run QTTs...

vcrfxia: Is this only meant for QTT or are we expecting users to want/be able to set this for their own queries too? (Guessing the latter but want to check.)

Also, the first config (outer join spurious results fix emit interval) is unrelated to this PR, right? Is that another config we think users might want to set?

ksqldb-common/src/main/java/io/confluent/ksql/parser/OutputRefinement.java
```diff
@@ -31,13 +31,22 @@ public enum OutputRefinement {
...
   /**
-   * Only final results should be materialized.
+   * Only final results should be materialized, using suppress(), ie, in-memory store.
    *
    * <p>For a table, this would mean output only the finalized result per-key.
    *
    * <p>For a stream, all events are final, so all are output.
    */
-  FINAL;
+  FINAL,
```
vcrfxia, line 40: Is this old value still needed after this PR? It should be obsolete, right? Existing queries have suppress nodes in their plans, and new queries will only ever set `FINAL_PERSISTENT` instead.

ksqldb-common/src/main/java/io/confluent/ksql/serde/WindowInfo.java
vcrfxia, line 88: This is just a "philosophical" choice (saying that conceptually a window info doesn't care about emit strategy), rather than a practical choice, right? As in, nothing breaks if we decide to include emitStrategy in the equals/hashCode methods, does it?

ksqldb-common/src/test/java/io/confluent/ksql/serde/WindowInfoTest.java
```diff
@@ -40,27 +41,32 @@ public void shouldThrowNPEs() {
...
+            WindowInfo.of(HOPPING, Optional.of(Duration.ofMillis(19)), Optional.empty())
         )
         .addEqualityGroup(
-            WindowInfo.of(TUMBLING, Optional.of(Duration.ofMillis(1010)))
+            WindowInfo.of(TUMBLING, Optional.of(Duration.ofMillis(1010)), Optional.empty())
+        )
+        .addEqualityGroup(
+            WindowInfo.of(TUMBLING, Optional.of(Duration.ofMillis(1020)), Optional.of(OutputRefinement.CHANGES)),
+            WindowInfo.of(TUMBLING, Optional.of(Duration.ofMillis(1020)), Optional.of(OutputRefinement.FINAL)),
+            WindowInfo.of(TUMBLING, Optional.of(Duration.ofMillis(1020)), Optional.of(OutputRefinement.FINAL_PERSISTENT))
```
vcrfxia, line 61: nit: empty optional should compare equal in this group too

ksqldb-engine/src/main/java/io/confluent/ksql/analyzer/RewrittenAnalysis.java
```diff
@@ -139,18 +139,20 @@ public Optional<WindowExpression> getWindowExpression() {
...
-    if (!(windowExpression.isPresent()
-        && !windowExpression.get().getKsqlWindowExpression().getGracePeriod().isPresent()
-        && refinementInfo.isPresent()
-        && refinementInfo.get().getOutputRefinement() == OutputRefinement.FINAL
-         )
-    ) {
+    // we only need to rewrite if we have a window expression and if we use emit final
+    if (!windowExpression.isPresent()
+        || !refinementInfo.isPresent()
+        || !refinementInfo.get().isFinal()) { // can we test for isFinal? or memory vs rocks?
```
vcrfxia, line 145: What's the concern here? Are you saying that users of in-memory stores wouldn't want to enable the new emit strategy? (Probably shouldn't affect ksql since ksql always uses rocks, just trying to understand.)

ksqldb-functional-tests/src/test/resources/query-validation-tests/suppress.json
```diff
@@ -3,7 +3,8 @@
     {
       "name": "should emit final result immediately at window end if grace is specified as zero",
       "properties": {
-        "ksql.suppress.enabled": true
+        "ksql.suppress.enabled": true,
```
vcrfxia, line 6: We need new historic plans for these existing QTTs with this change, right? The query plans should've changed to no longer have the suppress.

ksqldb-engine/src/main/java/io/confluent/ksql/ddl/commands/CommandFactories.java
```diff
@@ -114,11 +116,14 @@ public DdlCommand create(
   }
 
   @Override
-  public DdlCommand create(final KsqlStructuredDataOutputNode outputNode) {
+  public DdlCommand create(
+      final KsqlStructuredDataOutputNode outputNode,
+      final Optional<RefinementInfo> emitStrategy
```
vcrfxia, line 121: Should this be passed as part of the output node, rather than separately? I could see the argument for this going either way, just tossing it out for consideration.

vcrfxia (commented):
Thanks @mjsax -- LGTM at a high level. Some questions inline.

ksqldb-common/src/main/java/io/confluent/ksql/config/KsqlConfigResolver.java
mjsax: Both are internal KS configs -- they control how often we try to emit results -- the other configs is for left/outer stream-stream joins (yes, it's unrelated to this PR, but serves the exact same purpose for a different operator).

Because both are internal, we don't really expect users to set them (with the exception that there is a perf problem and we tell them to change it). But we should be able for set the configs IMHO -- otherwise, the guard they provide for KS does not work for ksqlDB.

There are a few more internal configs, and we could consider to add them here, too.

ksqldb-common/src/main/java/io/confluent/ksql/parser/OutputRefinement.java
mjsax: 🤔 -- fair point. Could be that we don't need it...

Wondering if we should remove it? Or if we should re-purpose it?

ksqldb-common/src/main/java/io/confluent/ksql/serde/WindowInfo.java
mjsax: Some stuff did break... I originally included it and only remove it later and added the comment (cannot remember from the top of my head what did break though).

ksqldb-functional-tests/src/test/resources/query-validation-tests/suppress.json
mjsax: Correct. I did not yet create historical plan on purpose, as it "pollutes" the PR (and I need to wait for the session window KS PR anyway)

ksqldb-engine/src/main/java/io/confluent/ksql/ddl/commands/CommandFactories.java
mjsax: I don't think it has anything to do with the output node? The old `suppress()` is just one more operator in the tree (and for the new emit final it's not even a new operator but a config to `aggregate()`.

`OutputNode` only contains a potential `LIMIT` clause and the result schema. It seem independent from the emit strategy?

Overall, passing in `emitStrategy` here is also not clean, but given how the code and planner is setup, it's the easiest way to do it this way. In a proper planner, we would do it differently, and make it a property of the logical aggregation node itself.

ksqldb-engine/src/main/java/io/confluent/ksql/analyzer/RewrittenAnalysis.java
mjsax: We test for `!isFinal()` that returns `!true` for `FINAL (in-memory) || FINAL_PERSISTENT (rocks)`?

The question is, if we need to test for only `!FINAL_PERSISTENT` instead?

Given you other comment about `FINAL` might be obsolete, the answer is "no". However, if we should return the unmodified window expression only for `FINAL` (but not for `FINAL_PERSISTENT`) the answer might be "yes".

ksqldb-common/src/main/java/io/confluent/ksql/parser/OutputRefinement.java
vcrfxia: The distinction would only be internal, right? As in, users will never see `FINAL` vs `FINAL_PERSISTENT`, and we also don't write anything of the sort into the command topic. If so, I think either's fine since we'd always be able to change it in the future. My vote is to repurpose `FINAL` if you feel confident that it's safe, else replace `FINAL` with `FINAL_PERSISTENT` since we know for sure that's safe.

ksqldb-common/src/main/java/io/confluent/ksql/serde/WindowInfo.java
vcrfxia: Interesting! I'd be curious to know what broke, but if it's difficult to find out again then don't worry about it.

ksqldb-engine/src/main/java/io/confluent/ksql/analyzer/RewrittenAnalysis.java
vcrfxia: Got it, I initially misunderstood what you meant by "memory vs rocks". I would hope that only new queries go through the RewrittenAnalysis so the two should be equivalent. It's been a while since I've worked with this part of the code, though.

ksqldb-engine/src/main/java/io/confluent/ksql/ddl/commands/CommandFactories.java
vcrfxia: That's fair, let's leave it out then. 

(FWIW, KsqlStructuredDataOutputNode contains a number of extra fields beyond those in OutputNode. It's still fair to argue that emitStrategy doesn't belong, though.)

ksqldb-common/src/main/java/io/confluent/ksql/config/KsqlConfigResolver.java
vcrfxia: That makes sense. Just to confirm: users _will_ be able to set these internal streams configs on their ksql queries (and have the configs take effect) after this change, right? 

I think it's fine to leave the second config in -- we're essentially saying it's safe for users to set these configs on their ksql queries if they want to.

mjsax: Technically, yes.

Of course, both configs are "internal" meaning it's not documented anywhere (on purpose) and user most likely don't know about them (and should not need to know about them in general).

ksqldb-common/src/main/java/io/confluent/ksql/parser/OutputRefinement.java
mjsax: Yes, internal only, there is just `EMIT FINAL` syntax for the user and if this is merged it should always use `FINAL_PERSISTENT` internally.

> My vote is to repurpose

Ack. Will look into it. Was not 100% sure if it would be save, but seems it should be. Better to cleanup right away.

ksqldb-common/src/main/java/io/confluent/ksql/serde/WindowInfo.java
mjsax: Could not reproduce it -- maybe something change in the middle. Added it to both `equals` and `hashCode`.

ksqldb-common/src/main/java/io/confluent/ksql/serde/WindowInfo.java
```diff
@@ -73,20 +85,35 @@ public boolean equals(final Object o) {
       return false;
     }
     final WindowInfo that = (WindowInfo) o;
+
+    final Optional<OutputRefinement> thisEmit = emitStrategy.isPresent()
```
mjsax, line 89: @vcrfxia -- I hope this make sense? The idea is to make `empty` and `changes` equal (also cf `hashCode` below)

vcrfxia: Makes sense but I have a code style question -- if empty is always the same as EMIT CHANGES, then can be change the constructor to store `OutputRefinement` rather than `Optional<OutputRefinement>` and use `emitStrategy.orElse(OutputRefinement.CHANGES)` in the constructor, in order to simplify these methods here?

ksqldb-engine/src/test/java/io/confluent/ksql/planner/LogicalPlannerTest.java
```diff
@@ -193,6 +192,7 @@ public void testSimpleRightJoinFilterLogicalPlan() {
     assertThat(rightSource.getSources().get(0), instanceOf(PreJoinRepartitionNode.class));
   }
 
+  @Ignore
```
vcrfxia, line 195: Looks like the latest commit adds this Ignore annotation to a number of unit tests. I'm guessing you have plans to address these?

ksqldb-common/src/main/java/io/confluent/ksql/serde/WindowInfo.java
mjsax: I did consider it, but was not sure -- as you bring it up, too, seems to be the better choice. Let me update it.

ksqldb-functional-tests/src/test/resources/query-validation-tests/suppress.json
```diff
@@ -20,9 +20,6 @@
     },
     {
       "name": "should not emit final result before window end if grace is specified as zero",
-      "properties": {
-        "ksql.suppress.enabled": true
```
vcrfxia, line 24: How come most of the tests have `"ksql.streams.__emit.interval.ms.kstreams.windowed.aggregation__": 0` added, but this one and the one immediately below it don't?

ksqldb-functional-tests/src/test/resources/sql-tests/query-upgrades/filters.sql
```diff
@@ -328,10 +328,9 @@ CREATE OR REPLACE TABLE b AS SELECT id, COUNT(*)
 ----------------------------------------------------------------------------------------------------
 --@test: add filter to Suppress
 --@expected.error: io.confluent.ksql.util.KsqlException
---@expected.message: Upgrades not yet supported for TableSuppress
+--@expected.message: Upgrades not yet supported for StreamWindowedAggregate
```
vcrfxia, line 331: I'm fine simply deleting this test rather than updating the expected error message. There's already another one which expects `Upgrades not yet supported for StreamWindowedAggregate`.

ksqldb-functional-tests/src/test/resources/query-validation-tests/suppress.json
mjsax: Guess it still passed because we expect no output (and thus I missed it), but I think we should set it to zero to not accidentally mask a bug.

Great catch!

ksqldb-functional-tests/src/test/resources/sql-tests/query-upgrades/filters.sql
mjsax: Maybe @agavra can chime in if this test still adds value or not?

agavra: yeah we can probably remove this test - the idea was that if you had a suppress it would fail regardless of what was being aggregated, but I don't think it adds much value since suppress only makes sense with windowed aggregates anyway (which we already test for above)

mjsax: Follow up PR to remove this test: https://github.com/confluentinc/ksql/pull/9308

PR #9145, opened 2022-05-25
fix: revert default `/query-stream` Content-Type to `application/vnd.ksqlapi.delimited.v1` from `application/vnd.ksql.v1+protobuf`

cprasad1:
### Description 
https://github.com/confluentinc/ksql/pull/9103 introduced a new protobuf Content-Type: `application/vnd.ksql.v1+protobuf`.
Upon introduction, `application/vnd.ksql.v1+protobuf` accidentally became the default content type for the `/query-stream` endpoint superseding our old default `application/vnd.ksqlapi.delimited.v1`. This happened because the `/query-stream` route had an equal weighting for all the content types including the new protobuf format. When there are equal weights on all the formats, then `routingContext.getAcceptableContentType()` https://github.com/confluentinc/ksql/blob/bd9034774cf11693f751472ef168f6fe9c99515d/ksqldb-rest-app/src/main/java/io/confluent/ksql/api/server/QueryStreamHandler.java#L123 just picks the first match (which happened to be the protobuf format in our case).

We can set a weight of 0 for the protobuf format, but as far as I can tell from the Vert.x docs, there is no way to individually specify weights for the types. The next best approach is to add a separate `Route` for the protobuf format to the endpoint. By default routes are matched in the order they are added to the router [Vert.x docs](https://vertx.io/docs/vertx-web/java/#_route_order).

This PR also exposes the protobuf format on the `/query` endpoint which is a follow-up to https://github.com/confluentinc/ksql/pull/9103.

### Testing done 

### Reviewer checklist

### DIFF
--- a/ksqldb-rest-app/src/main/java/io/confluent/ksql/api/server/ServerVerticle.java
+++ b/ksqldb-rest-app/src/main/java/io/confluent/ksql/api/server/ServerVerticle.java
@@ -142,6 +142,9 @@ public class ServerVerticle extends AbstractVerticle {
         .produces(DELIMITED_CONTENT_TYPE)
         .produces(JSON_CONTENT_TYPE)
         .produces(KsqlMediaType.KSQL_V1_JSON.mediaType())
+        .handler(BodyHandler.create(false))
+        .handler(new QueryStreamHandler(endpoints, connectionQueryManager, context, server, false));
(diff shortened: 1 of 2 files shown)


Gerrrr (approved):
Thank you for a quick fix!

PR #9147, opened 2022-05-26
feat: close websocket connections with expired tokens

jzaralim:
### Description 
Check if a websocket connection has a JWT token. If it does, then close the connection when the timeout passes.

### Testing done 
Manual + unit testing.

### Reviewer checklist

### DIFF
--- a/ksqldb-common/src/main/java/io/confluent/ksql/util/KsqlConfig.java
+++ b/ksqldb-common/src/main/java/io/confluent/ksql/util/KsqlConfig.java
@@ -678,10 +678,10 @@ public class KsqlConfig extends AbstractConfig {
       = "ksql.websocket.connection.max.timeout.ms";
   public static final long KSQL_WEBSOCKET_CONNECTION_MAX_TIMEOUT_MS_DEFAULT = 0;
   public static final String KSQL_WEBSOCKET_CONNECTION_MAX_TIMEOUT_MS_DOC
-      = "If this config is set to a positive number, then ksqlDB will terminate authenticated"
-      + "websocket connections after a timeout. The timeout will be the lower of the token's "
-      + "lifespan and the value of this config. If this config is set to 0, then ksqlDB will not"
-      + "close websockets even if the token has an expiration time.";
+      = "If this config is set to a positive number, then ksqlDB will terminate websocket"
+      + " connections after a timeout. The timeout will be the lower of the auth token's "
+      + "lifespan (if present) and the value of this config. If this config is set to 0, then "
+      + "ksqlDB will not close websockets even if the token has an expiration time.";
 
   private enum ConfigGeneration {
     LEGACY,
--- a/ksqldb-common/src/main/java/io/confluent/ksql/util/KsqlConfig.java
+++ b/ksqldb-common/src/main/java/io/confluent/ksql/util/KsqlConfig.java
@@ -674,6 +674,13 @@ public class KsqlConfig extends AbstractConfig {
       = "The default timeout for ASSERT SCHEMA statements if no timeout is specified "
       + "in the statement.";
 
+  public static final String KSQL_WEBSOCKET_TIMEOUT_ENABLE
+      = "ksql.websocket.timeout.enable";
+  public static final boolean KSQL_WEBSOCKET_TIMEOUT_ENABLE_DEFAULT = true;
+  public static final String KSQL_WEBSOCKET_TIMEOUT_ENABLE_DOC
+      = "If the websocket auth token has a timeout, then close the connection from the ksqlDB"
+      + "server when the token expires.";
+
   private enum ConfigGeneration {
     LEGACY,
     CURRENT
@@ -1453,6 +1460,13 @@ public class KsqlConfig extends AbstractConfig {
             KSQL_ASSERT_SCHEMA_DEFAULT_TIMEOUT_MS_DEFAULT,
             Importance.LOW,
             KSQL_ASSERT_SCHEMA_DEFAULT_TIMEOUT_MS_DOC
+        )
+        .define(
+            KSQL_WEBSOCKET_TIMEOUT_ENABLE,
+            Type.BOOLEAN,
+            KSQL_WEBSOCKET_TIMEOUT_ENABLE_DEFAULT,
+            Importance.LOW,
+            KSQL_WEBSOCKET_TIMEOUT_ENABLE_DOC
         )
         .withClientSslSupport();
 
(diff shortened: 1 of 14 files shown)


ksqldb-common/src/main/java/io/confluent/ksql/util/KsqlConfig.java
```diff
@@ -674,6 +674,13 @@ public class KsqlConfig extends AbstractConfig {
       = "The default timeout for ASSERT SCHEMA statements if no timeout is specified "
       + "in the statement.";
 
+  public static final String KSQL_WEBSOCKET_TIMEOUT_ENABLE
+      = "ksql.websocket.timeout.enable";
+  public static final boolean KSQL_WEBSOCKET_TIMEOUT_ENABLE_DEFAULT = true;
```
spena, line 679: What about using what KIP-368 does? Kafka has a `connections.max.reauth.ms` that if set to `0`, then the re-auth is disabled, but if set to a positive number, then timeout will use the token expiry time or this config timeout, whichever is the lowest.

We can name this config `ksql.websocket.connection.max.reauth.ms` or `...timeout.ms`. But as we're doing this on authentication only, then `...reauth.ms` makes sense 'cause the `timeout.ms` won't work on unauthenticated connections.

ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/KsqlServerEndpoints.java
```diff
@@ -296,20 +300,39 @@ public void executeWebsocketStream(final ServerWebSocket webSocket, final MultiM
...
+            getTokenTimeout(apiSecurityContext.getPrincipal())
+        );
       } finally {
         ksqlSecurityContext.getServiceContext().close();
       }
       return null;
     }, workerExecutor);
   }
 
+  private Optional<Long> getTokenTimeout(final Optional<KsqlPrincipal> principal) {
```
spena, line 321: I'm afraid you'll need this implementation as part of the `confluent-security-plugins` to access the `JwtPrincipal`. The security plugins is not open-source, so we cannot ship it with it.

One good thing is this timeout will only available on RBAC environments, so we can reuse the `KsqlSecurityExtension` to access security features. There's one impl. in  the `confluent-security-plugins`. The `KsqlSecurityExtension` is part of ksqldb as interface and dummy impls. only.

ksqldb-common/src/main/java/io/confluent/ksql/util/KsqlConfig.java
jzaralim: We're not supporting reauth. We're just terminating connections, so I set the config to `ksql.websocket.connection.max.timeout.ms`. The behavior is, if the connection has a token with an expiration, then close the websocket after `min(token expiration, config expiration)`. Otherwise, don't close the connection.

ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/KsqlServerEndpoints.java
jzaralim: I added a `KsqlAuthTokenProvider` to the extensions. The security-plugins changes are here: https://github.com/confluentinc/confluent-security-plugins/pull/483

ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/util/AuthenticationUtil.java
```diff
@@ -0,0 +1,53 @@
...
+        && authTokenProvider.isPresent()
+        && principal.isPresent()
+        && authTokenProvider.get().getLifetimeMs(principal.get()).isPresent()
+    ) {
+      final long tokenTimeout = authTokenProvider.get().getLifetimeMs(principal.get()).get()
+          - clock.millis();
+      return Optional.of(Math.min(tokenTimeout, maxTimeout));
+    } else {
+      return Optional.empty();
+    }
```
spena, line 50: If `KSQL_WEBSOCKET_CONNECTION_MAX_TIMEOUT_MS` is a positive number, but there's not an Auth provider or other principal details, shouldn't the connection timeout be set to the maximum used in the config instead of empty? Because we're not linking this config with Auth, then I think we should use the MAX config in that case.

ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/util/AuthenticationUtil.java
```diff
@@ -0,0 +1,53 @@
...
+public class AuthenticationUtil {
+
+  final Clock clock;
+
+  public AuthenticationUtil(final Clock clock) {
+    this.clock = Objects.requireNonNull(clock);
+  }
+
+  public Optional<Long> getTokenTimeout(
+      final Optional<KsqlPrincipal> principal,
```
pgaref, line 34: Would it make sense to extend the `KsqlPrincipal` interface to provide a getTimeout method?
We could return a default value based on configuration and have the other classes like JWT override the method.

jzaralim: That wouldn't work, because the `KsqlPrincipal` is implemented in the ksql library, and extracting timestamps requires a dependency to the security plugins (`JwtPrincipal` does not override `KsqlPrincipal`)

ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/util/AuthenticationUtil.java
jzaralim: Yeah that makes sense

ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/resources/streaming/WSQueryEndpoint.java
```diff
@@ -105,10 +106,19 @@ public WSQueryEndpoint(
...
 
   public void executeStreamQuery(final ServerWebSocket webSocket, final MultiMap requestParams,
-      final KsqlSecurityContext ksqlSecurityContext, final Context context) {
+      final KsqlSecurityContext ksqlSecurityContext, final Context context,
+      final Optional<Long> timeout) {
 
     try {
       final long startTimeNanos = Time.SYSTEM.nanoseconds();
+      if (timeout.isPresent()) {
+        exec.schedule(
```
spena, line 115: Could you add a LOG.info() line that specifies the timeout?

PR #9157, opened 2022-06-01
fix: remove pull query related configs from queryLevelProperties

ableegoldman:
We just merged a PR that[ moved several missing query configs](https://github.com/confluentinc/ksql/pull/9144) to the `queryLevelProperties` list, including two configs related to pull queries that were said to be query-level in the current docs and specifically mentioned using the `SET` command to. However it seems these configs may not be properties of persistent queries after all and should be removed from this list.


ksqldb-rest-model/src/main/java/io/confluent/ksql/rest/entity/PropertiesList.java
```diff
@@ -127,13 +126,12 @@ public class PropertiesList extends KsqlEntity {
       DEFAULT_DESERIALIZATION_EXCEPTION_HANDLER_CLASS_CONFIG,
       DEFAULT_TIMESTAMP_EXTRACTOR_CLASS_CONFIG,
       FAIL_ON_DESERIALIZATION_ERROR_CONFIG,
-      KSQL_STRING_CASE_CONFIG_TOGGLE,
-      KSQL_NESTED_ERROR_HANDLING_CONFIG,
-      KSQL_QUERY_ERROR_MAX_QUEUE_SIZE,
-      KSQL_QUERY_RETRY_BACKOFF_INITIAL_MS,
-      KSQL_QUERY_RETRY_BACKOFF_MAX_MS,
       KSQL_QUERY_PULL_MAX_ALLOWED_OFFSET_LAG_CONFIG,
-      KSQL_QUERY_PULL_TABLE_SCAN_ENABLED,
```
ableegoldman, line 136: The only real change is to remove this config -- everything else is just moving the ksql configs that are out of alphabetical order to maintain readability

agavra: how is this class used? these are still "per query" configurations in that you can set them per pull query. I see in the description you explicitly call out that they're not "persistent query" configurations, which is correct

ableegoldman:
Closing this as the consensus seems to be that these should indeed be query-level configs and controllable by `SET`

Issue #9140, opened 2022-05-20
COLLECT_SET should apply to tables as well as streams


vvcephei:
Currently, the `COLLECT_SET` is only available for windowed stream aggregations, whereas `COLLECT_LIST` is available for table and stream aggregations.

The reason for this discrepancy is that table and stream aggregations are computed differently in the runtime, and in particular table aggregations need to define a "subtractor" for how to remove a record from the aggregation. (Both table and stream aggregations define an "adder" for adding records to the set). The reason for this is that stream aggregations can never remove a record from a window (since historical stream events are immutable), whereas tables (obviously) support deleting and modifying existing records. Therefore, stream records can never "leave" an aggregation, whereas table records can.

The punchline for COLLECT_SET is that if I have an input table like:
```
key | color | number
a    | red    | 17
b    | red    | 17
c    | red    | 18
```

and I do a query like

```
CREATE ... AS SELECT color, COLLECT_SET(number) AS numbers FROM input GROUP_BY color;
```

Then, I should get the result

```
color | numbers
red   |  {17, 18}
```

At this point, if we delete record `a` or `b` from the input table, then the set should remain `{17, 18}`, but if we delete _both_ `a` _and_ `b`, then the set should only become `{18}`. In other words, we can't just handle the deletion of `a` (for example) by removing the corresponding element from the set.

One possible solution is to maintain a counter for each set element (internally) and decrement the counter when removing from the set.

jnh5y:
After a brief look at this and some discussion with @agavra, I think the solution is to:

1. Create a new separate CollectSet class which aggregates counts (and hence can decrement them).
2. Add a new default KSQL configuration setting to show that new queries can used the new setting.
3. Implement a `AggregateFunctionFactory` which will select the correct CollectSet class based on the configuration.

jnh5y:
For the configuration, @rodesai notes that we may want to make it a `CompatibilityBreakingConfig`.

https://github.com/confluentinc/ksql/blob/bd9034774cf11693f751472ef168f6fe9c99515d/ksqldb-common/src/main/java/io/confluent/ksql/util/KsqlConfig.java#L682

PR #9168, opened 2022-06-03
feat: introduce ATTR aggregation function

agavra:
### Description 

Introduces the `ATTR` aggregation function, modeled after https://kb.tableau.com/articles/howto/when-to-use-the-attribute-attr-function, which essentially just ensures that you only have one column value within an aggregate. A classic example of this would be:

```sql
CREATE TABLE customers (customerId INT PRIMARY KEY, name STRING);
CREATE STREAM orders (orderID INT, customerID INT);
CREATE STREAM ordersWithNames AS SELECT * 
  FROM orders 
  JOIN customers ON orders.customerID = customers.customerId;

SELECT customerId, ATTR(name), COUNT(*) from ordersWithNames GROUP BY customerId;
```

There are various ways you could solve this with existing UDAFs (e.g. `REDUCE(COLLECT_SET(foo), (a, b) -> IF a = b THEN a ELSE null))` but that's bulky 😆 

### Testing done 

QTTs and Unit Tests

### Reviewer checklist


ksqldb-engine/src/main/java/io/confluent/ksql/function/udaf/attr/Attr.java
```diff
@@ -0,0 +1,146 @@
...
+  /**
+   * This class implements the logic for checking that we only ever see a valid value. Note
+   * that there is some complexity in this class to differentiate the case from a "valid null",
+   * which is the case for when the aggregate is initialized, from an "invalid null".
+   */
+  @VisibleForTesting
+  static class Impl<T> implements TableUdaf<T, Struct, T> {
+
+    static final String INIT = "INIT";
+    static final String VALID = "VALID";
```
pgaref, line 61: SEEN seems more appropriate here than VALID right?

ksqldb-engine/src/main/java/io/confluent/ksql/function/udaf/attr/Attr.java
```diff
@@ -0,0 +1,146 @@
...
+import org.apache.kafka.connect.data.Struct;
+
+@UdafDescription(
+    name = "ATTR",
+    description = "The ATTR() aggregation indicates there are multiple values, but only one "
+        + "was expected. For example, if aggregating against a table that semantically should "
+        + "have only one value for a column given a key, this aggregation enables users to "
+        + "indicate that they expect only a single value.\n\n"
+        + "If the aggregation encounters more than a single value for the expected singular "
+        + "column, the entire aggregation will return null."
```
pgaref, line 39: Any particular reason we are returning null instead of something like a "*" as Tableau does? What if all the column values before aggregation were actually null?

ksqldb-engine/src/main/java/io/confluent/ksql/function/udaf/attr/Attr.java
```diff
@@ -0,0 +1,146 @@
+/*
+ * Copyright 2020 Confluent Inc.
```
jnh5y, line 2: 2022

ksqldb-engine/src/test/java/io/confluent/ksql/function/udaf/attr/AttrTest.java
```diff
@@ -0,0 +1,119 @@
+/*
+ * Copyright 2020 Confluent Inc.
```
jnh5y, line 2: 2022

ksqldb-engine/src/main/java/io/confluent/ksql/function/udaf/attr/Attr.java
```diff
@@ -0,0 +1,146 @@
...
+    }
+
+    @Override
+    public Optional<SqlType> getReturnSqlType() {
+      return Optional.of(inType);
+    }
+
+    @Override
+    public Struct undo(final T valueToUndo, final Struct aggregateValue) {
+      return aggregateValue;
```
jnh5y, line 93: If you are going to make this a Table function, you'd have to keep track of counts of each value and decrement them in the undo.

agavra: good catch! I had filled this out as a skeleton and then forgot to update this part.

ksqldb-engine/src/main/java/io/confluent/ksql/function/udaf/attr/Attr.java
agavra: `"*"` is a string value, which is not applicable to all value formats. if all columns are actually null then the aggregation will (correctly) return null (fwiw, this comment is applicable for all defaults, including something like `*`).

ksqldb-engine/src/main/java/io/confluent/ksql/function/udaf/attr/Attr.java
```diff
@@ -0,0 +1,154 @@
...
+        + "column, the entire aggregation will return null."
+)
+public final class Attr {
+
+  private Attr() {
+    // checkstyle complains otherwise
+  }
+
+  @UdafFactory(description = "Collect as a singleton")
+  public static <T> TableUdaf<T, List<Struct>, T> createAttr() {
```
jnh5y, line 51: As a note, Histogram needs to maintain the same internal structure and it uses a Map<T, Long>.  

Questions:

1.  Is one of these approaches "better" for some reason?
2.  If one is better, is it useful to standardize on them?

(I ask with https://github.com/confluentinc/ksql/issues/9140 in view as well since it'll use the same internal structure.  If one were really, really looking for refactoring options, the three functions may be able to share most of the code...)

blueedgenick: that makes total sense! thanks for explaining :)

ksqldb-functional-tests/src/test/resources/query-validation-tests/attr.json
```diff
@@ -0,0 +1,141 @@
...
+      "outputs": [
+        {"topic": "OUT", "key": 1, "value": {"S": "s", "I": 1, "BI": 1, "D": 1.0, "BO": true}},
+        {"topic": "OUT", "key": 1, "value": {"S": "s", "I": 1, "BI": 1, "D": 1.0, "BO": true}},
+        {"topic": "OUT", "key": 1, "value": {"S": "s", "I": 1, "BI": 1, "D": 1.0, "BO": true}}
+      ]
+    },
+    {
+      "name": "valid attr with various non-scalar fields",
+      "statements": [
+        "CREATE STREAM input (k INT KEY, li ARRAY<INT>, s STRUCT<A STRING, B INT>) WITH (kafka_topic='in', value_format='JSON');",
```
jnh5y, line 26: Does Attr work with Maps?  

Just curious.  Not sure it requires a QTT.

ksqldb-engine/src/main/java/io/confluent/ksql/function/udaf/attr/Attr.java
```diff
@@ -0,0 +1,154 @@
...
+
+      return (T) collect.get(0).get(VALUE);
+    }
+
+    private void update(final List<Struct> agg, final Object current, final int count) {
+      boolean found = false;
+      for (final Struct entry : agg) {
+        if (Objects.equals(entry.get(VALUE), current)) {
+          found = true;
+          entry.put(COUNT, Math.max(0, entry.getInt32(COUNT) + count));
```
jnh5y, line 145: Can we `break;` and stop iterating once we've found the element?

If there can be multiple copies in the list, then this would update all of them by 1!

ksqldb-engine/src/main/java/io/confluent/ksql/function/udaf/attr/Attr.java
```diff
@@ -0,0 +1,154 @@
...
+    @Override
+    public List<Struct> aggregate(final T current, final List<Struct> agg) {
+      final List<Struct> out = new ArrayList<>(agg);
+      update(out, current, 1);
+      return out;
+    }
+
+    @Override
+    public List<Struct> merge(final List<Struct> one, final List<Struct> two) {
+      // use O(n^2) algorithm here because in practice each of these lists
```
jnh5y, line 109: Based on the other discussions on this thread, I'd note that the core of this class is likely how we'd want to implement Collect_Set as a TableUDAF.

Given that, it may be amusing to think through reducing the complexity.  That said, I'm totally happy to see that deferred to another ticket.  I think we should be happy with the types as-is.  (And changing those is trickier.)

In the general case, would it be faster to sort each list and merge them together?  Naively, seems like that's `O(n log n) + O(m log m) + O(n + m)` rather than `O(n*m)`.

ksqldb-engine/src/main/java/io/confluent/ksql/function/udaf/attr/Attr.java
jnh5y, line 145: When the count goes to zero, would it make sense to remove it from the list?

ksqldb-functional-tests/src/test/resources/query-validation-tests/attr.json
agavra: It should! I tested struct (I was going to test a map nested in a struct but looks like I forgot to do that) because generally that's the most complicated one

ksqldb-engine/src/main/java/io/confluent/ksql/function/udaf/attr/Attr.java
agavra: hah! you're right 😉 don't think it should ever come up in practice but I'll go ahead and update this

ksqldb-engine/src/main/java/io/confluent/ksql/function/udaf/attr/Attr.java
agavra: Now that I'm giving this some more thought.... if we use insertion sort, this may actually amortize to `O(n)` sort because we know that we're starting with a sorted list every time. Let me noodle with that a bit in the IDE and see if my intuition is correct.

ksqldb-engine/src/main/java/io/confluent/ksql/function/udaf/attr/Attr.java
agavra: I thought about that but decided against it for code simplicity - that would cause concurrent modification or I'd need to track what to remove afterwards

even if the cardinality was higher, i don't think it's worth removing it from the list because that would only happen in `undo` operations - which I suspect are somewhat rare

jnh5y (approved):
:shipit:!

ksqldb-engine/src/main/java/io/confluent/ksql/function/udaf/attr/Attr.java
agavra: erm... just realized I'd need to implement the comparator for complex values as well, which would be rather difficult...

ksqldb-engine/src/main/java/io/confluent/ksql/function/udaf/attr/Attr.java
agavra: ah man... if only I had responded to this with "can't catch a break, can I?"

jnh5y: Next time, if you try exceptionally hard, you'll find the resources to make bad puns more readily before the close of business....

PR #9180, opened 2022-06-07
fix: add BYTES support for KAFKA format

### LINKED ISSUE #8693: KAFKA format does not support BYTES
**Describe the bug**
If you try to create a KAFKA formatted source with a BYTES column, the command will return

```
CREATE STREAM TEST (ID BYTES KEY, b BYTES) WITH (kafka_topic='test', format='DELIMITED');
The 'KAFKA' format does not support type 'BYTES'
```

This is because the BYTES type is missing [here](https://github.com/confluentinc/ksql/blob/a27e5e7501891e644196f8d164d078672e0feecd/ksqldb-serde/src/main/java/io/confluent/ksql/serde/kafka/KafkaSerdeFactory.java#L40)

nateab:
### Description 
Fixes #8693 by adding ByteBuffer serdes to the KafkaSerdeFactory. Added a new QTT to the bytes.json to check correct behaviour for the KAFKA format.


jzaralim (approved):
LGTM, thanks for the fix @nateab

PR #9182, opened 2022-06-07
chore: ignore new findbugs 4.7.0 detectors causing ksql build to fail

spena:
### Description 
There's an issue with the recent Findbugs 4.7.0 version that introduce new detectors.
https://github.com/spotbugs/spotbugs/issues/2040

The issue seems to fail only on Java 11. This PR excludes the new detectors from findbugs.

### Testing done 

### Reviewer checklist

### DIFF
--- a/findbugs/findbugs-exclude.xml
+++ b/findbugs/findbugs-exclude.xml
@@ -45,6 +45,16 @@ For a detailed description of findbugs bug categories, see http://findbugs.sourc
     <Match>
         <Bug pattern="JUA_DONT_ASSERT_INSTANCEOF_IN_TESTS"/>
     </Match>
+    <!-- false positive in Java 11, see https://github.com/spotbugs/spotbugs/issues/2040 -->
+    <Match>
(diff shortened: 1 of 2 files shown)


mjsax (approved):

pgaref:
We will just need to revert those changes in the next spotbugs release that turns off the false positive alters

spena:
Doing good @big-andy-coates. How about you? Working now in findbugs or what :P? I saw your comments in a findbugs issue and PR which I hoped it would've be ready soon, but I had to patch ksql 'cause it was blocking us. I'll remove these exclusions once there's a findbugs release.

xli1996:
I updated the common to use 4.7.1 which should include @big-andy-coates 's fix. https://github.com/confluentinc/common/pull/466

pgaref:
Hey @xli1996 -- assuming that spotbug version/plugin is inherited from common we should just bump and revert the exclusions introduced here right?

pgaref:
> yes

Awesome @xli1996  -- would you like to do a followup on this?

Issue #9185, opened 2022-06-08
After a call to CAST, results cannot be compared.


jnh5y:
**Describe the bug**
A query like 
```
SELECT *
FROM DataGen_Stream 
	WHERE CAST( FROM_UNIXTIME(REGISTERTIME) as DATE) <=  DATESUB(DAYS,365,'2018-09-23') 
EMIT CHANGES;
```

hits an error like

`Invalid Predicate: Line 19, Column 14: A method named "compareTo" is not declared in any enclosing class nor any supertype, nor through a static import. expression: (CAST(FROM_UNIXTIME(REGISTERTIME) AS DATE) <= DATESUB(DAYS, 365, '2018-09-23')), schema:REGISTERTIME BIGINT, USERID STRING, REGIONID STRING, GENDER STRING, ROWTIME BIGINT, ROWPARTITION INTEGER, ROWOFFSET BIGINT Statement: Select * FROM DataGen_Stream WHERE CAST( FROM_UNIXTIME(REGISTERTIME) as DATE) <= DATESUB(DAYS,365,'2018-09-23') EMIT CHANGES;`

**Expected behavior**
That one would be able to compare values after calling `CAST`.

PR #9186, opened 2022-06-08
fix: Allows results from CAST to compared.

jnh5y:
Address https://github.com/confluentinc/ksql/issues/9185

### Testing done 
Just a quick unit test to show that the Java code can be compiled and evaluated.

### Reviewer checklist


suhas-satish (approved):
LGTM, thanks for the quick fix.

PR #9193, opened 2022-06-09
chore: switch to pinned nanoversion strategy

vvcephei:
Switches the build to pinned nanoversions.

Depends on https://github.com/confluentinc/jenkins-common/pull/474 and https://github.com/confluentinc/ci-tools/pull/101


pom.xml
```diff
@@ -335,25 +333,25 @@
...
                 <artifactId>kafka-connect-json-schema-converter</artifactId>
-                <version>${io.confluent.schema-registry.version}</version>
+                <version>7.3.0-615</version>
             </dependency>
 
             <dependency>
                 <groupId>io.confluent</groupId>
                 <artifactId>kafka-connect-protobuf-converter</artifactId>
-                <version>${io.confluent.schema-registry.version}</version>
+                <version>7.3.0-615</version>
```
jnh5y, line 354: Just to confirm, the common pom version and the kafka.version are different, right?

jnh5y (approved):
I'm excited to try out the pinned versions!

guozhangwang (approved):
+1, excited to see this happening, finally!

pom.xml
vvcephei: Yes. The `kafka.version` is actually set in the common pom. This line is the schema registry version, though, and it's also different than the common pom.

vvcephei:
The last build failed because the agent ran out of disk space, but a prior one passed, so I'll go ahead and merge.

PR #9201, opened 2022-06-15
fix: change Object#equals back to `==` to fix RocksDBMetricsCollectorTest#shouldNotUpdateIfWithinInterval

ableegoldman:
Oddly `RocksDBMetricsCollectorTest.shouldNotUpdateIfWithinInterval` has started failing since I changed `==` into `Objects#equals` to fix the compiler warning on the side during my ThroughMetricsReporter PR. I haven't been able to reproduce it, nor did it fail on the PR build that I merged -- still, the timing and clear relation between failing test and modified code makes me pretty suspicious.

Note that the compiler warning can actually be ignored, as `==` is indeed correct in this particular case since `accumulateAndGet` will return a reference to the actual `now` object that's passed in (if condition is true)

See https://github.com/confluentinc/ksql/pull/9187/files#r897668861

### DIFF
--- a/ksqldb-streams/src/main/java/io/confluent/ksql/execution/streams/metrics/RocksDBMetricsCollector.java
+++ b/ksqldb-streams/src/main/java/io/confluent/ksql/execution/streams/metrics/RocksDBMetricsCollector.java
@@ -257,10 +257,10 @@ public class RocksDBMetricsCollector implements MetricsReporter {
 
     boolean check() {
       final Instant now = clock.get();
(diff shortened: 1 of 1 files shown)


guozhangwang:
I cannot understand why either.. to me the reference equality and value equality here should not make differences.

ableegoldman:
Seems like @rodesai merged an actual fix for this so I'll go ahead and close this -- see https://github.com/confluentinc/ksql/pull/9205

Issue #6403, opened 2020-10-10
Support pausing/resuming persistent queries


rodesai:
As a ksqldb user, I want the ability to pause execution of a query and then resume execution later on. I should be able to pause/resume using some ksql syntax. The syntax should describe the desired state of the query (e.g. running, or stopped), and should be persisted to the command topic like a ddl or dml so that ksql converges on the desired state even as nodes come and go.

PR #9203, opened 2022-06-15
feat: Support pausing/resuming persistent queries

jnh5y:
Addresses: https://github.com/confluentinc/ksql/issues/6403
Implements KLIP-63: PAUSE and RESUME for persistent queries.

### Description 
Adds the ability to pause and resume persistent queries.

### Testing done 
Unit tests have been added to cover the new syntax and basic usage.
An integration test class has been created.

### Reviewer checklist


jnh5y:
NB: I owe the following:
* Docs
* A little more work on command topic compaction
* A few more updates on the integration test.

Also, the build will fail until https://github.com/apache/kafka/pull/12161 is merged and synced to CCS Kafka.

ksqldb-engine/src/main/java/io/confluent/ksql/util/BinPackedPersistentQueryMetadataImpl.java
```diff
@@ -62,6 +64,7 @@ public class BinPackedPersistentQueryMetadataImpl implements PersistentQueryMeta
   private static final Logger LOG = LoggerFactory
       .getLogger(BinPackedPersistentQueryMetadataImpl.class);
 
+  private final AtomicBoolean isPaused = new AtomicBoolean(false);
```
vcrfxia, line 67: Are we actually expecting concurrent access to this value? ksql processing statements from the command topic should be single-threaded, so I'd expect at least updates to this value to only be from a single thread. It's possible that there are other threads which might try to read the value concurrently. (Curious what your thinking here was.)

ksqldb-engine/src/main/java/io/confluent/ksql/util/SharedKafkaStreamsRuntimeImpl.java
```diff
@@ -227,6 +227,14 @@ public synchronized void close() {
     kafkaStreams.cleanUp();
   }
 
+  @Override
+  public void pause(final QueryId queryId) {
```
vcrfxia, line 231: If this is a no-op in all implementations, does it even need to be part of the interface?

ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/execution/ListQueriesExecutor.java
```diff
@@ -113,8 +111,7 @@ private static Map<QueryId, RunningQuery> getLocalSimple(
                   ImmutableSet.of(),
                   ImmutableSet.of(),
                   q.getQueryId(),
-                  QueryStatusCount.fromStreamsStateCounts(
```
vcrfxia, line 116: This method is test-only now right? Let's remove it as dead code if so.

ksqldb-rest-app/src/test/java/io/confluent/ksql/rest/integration/PauseResumeIntegrationTest.java
```diff
@@ -0,0 +1,279 @@
...
+    // 5 more records have been produced
+    assertThatEventually(supplier, equalTo(14));
+    assertThat(getRunningCount(), equalTo(1));
+  }
+
+  @Test
+  public void shouldPauseAndResumeMultipleQueries() {
+    // Given:
+    createQuery("1");
+    createQuery("2");
```
vcrfxia, line 149: What's the purpose of having the second query in this test?

ksqldb-rest-app/src/test/java/io/confluent/ksql/rest/integration/PauseResumeIntegrationTest.java
```diff
@@ -0,0 +1,279 @@
...
+    // Then:
+    RestIntegrationTestUtil.makeKsqlRequest(REST_APP, "RESUME " + queryId + ";");
+
+    // 5 more records have been produced
+    assertThatEventually(supplier, equalTo(14));
+
+    List <KsqlEntity> showQueries3 = RestIntegrationTestUtil.makeKsqlRequest(REST_APP, "SHOW "
+        + "QUERIES;");
+    assertThat(((Queries) showQueries3.get(0)).getQueries().get(0).getStatusCount().getStatuses()
+        .get(KsqlQueryStatus.RUNNING), equalTo(1));
```
vcrfxia, line 178: Can we add a status check into this test for when the query is paused too?

ksqldb-rest-app/src/test/java/io/confluent/ksql/rest/integration/PauseResumeIntegrationTest.java
```diff
@@ -0,0 +1,279 @@
...
+    REST_APP.start();
+
+    // Verify PAUSED state -- eventually
+    assertThatEventually(this::getPausedCount, equalTo(1));
+
+    // Verify number of processed records
+    assertThatEventually(supplier, equalTo(9));
+
+    // Then:
+    String queryId2 = ((Queries) RestIntegrationTestUtil.makeKsqlRequest(REST_APP, "SHOW "
```
vcrfxia, line 215: Is this guaranteed to be the same query? (Is ordering of returned queries from show queries consistent?)

ksqldb-rest-app/src/test/java/io/confluent/ksql/rest/integration/PauseResumeIntegrationTest.java
```diff
@@ -0,0 +1,279 @@
...
+        + "QUERIES;").get(0)).getQueries().get(0).getId().toString();
+    RestIntegrationTestUtil.makeKsqlRequest(REST_APP, "RESUME " + queryId2 + ";");
+    assertThatEventually(this::getPausedCount, equalTo(0));
+    assertThatEventually(this::getRunningCount, equalTo(1));
```
vcrfxia, line 219: I thought we started two queries -- shouldn't this be 2? Alternatively, if the presence of the second query isn't relevant for the test, then let's just remove it.

ksqldb-rest-app/src/test/java/io/confluent/ksql/rest/integration/PauseResumeIntegrationTest.java
```diff
@@ -0,0 +1,279 @@
...
+    assertThatEventually(supplier, equalTo(14));
+
+    List <KsqlEntity> showQueries3 = RestIntegrationTestUtil.makeKsqlRequest(REST_APP, "SHOW "
+        + "QUERIES;");
+    assertThat(((Queries) showQueries3.get(0)).getQueries().get(0).getStatusCount().getStatuses()
+        .get(KsqlQueryStatus.RUNNING), equalTo(1));
+  }
+
+  // Parameterize this entire class to run with multiple REST instances?
+  // @Test
```
vcrfxia, line 234: I don't think this test is necessary. It'd essentially be re-testing that the command topic distributes commands as intended, which is outside the scope of pause/resume.

ksqldb-rest-model/src/main/java/io/confluent/ksql/rest/entity/PauseQueryEntity.java
```diff
@@ -0,0 +1,73 @@
...
+
+package io.confluent.ksql.rest.entity;
+
+import com.fasterxml.jackson.annotation.JsonCreator;
+import com.fasterxml.jackson.annotation.JsonIgnoreProperties;
+import com.fasterxml.jackson.annotation.JsonProperty;
+import java.util.Objects;
+
+@JsonIgnoreProperties(ignoreUnknown = true)
+public class PauseQueryEntity extends KsqlEntity {
```
vcrfxia, line 24: Is this used anywhere? I don't see this actually returned from anything. Same for ResumeQueryEntity below.

ksqldb-tools/src/test/java/io/confluent/ksql/tools/migrations/util/CommandParserTest.java
```diff
@@ -258,6 +258,28 @@ public void shouldParseTerminateStatement() {
     assertThat(commands.get(0).getCommand(), is("terminate some_query_id;"));
   }
 
+  @Test
+  public void shouldParsePauseStatement() {
```
vcrfxia, line 262: If we're planning to support these statements in the migration tool (seems reasonable to me), we need better test coverage here. At minimum, MigrationsTest.java (the integration test for the migrations tool) needs to be updated.

ksqldb-engine/src/main/java/io/confluent/ksql/util/BinPackedPersistentQueryMetadataImpl.java
```diff
@@ -416,6 +419,27 @@ public void start() {
...
+    if (isPaused.get()) {
+      return KsqlQueryStatus.PAUSED;
+    } else {
+      return KsqlConstants.fromStreamsState(getState());
+    }
+  }
+
+  @Override
+  public void pause() {
+    sharedKafkaStreamsRuntime.getKafkaStreams().pauseNamedTopology(topology.name());
```
vcrfxia, line 433: I'd like to understand more about how pausing queries works with named topologies. I saw the note about it in the KIP but am still unclear about the details. I also have some questions about how pausing a KS app affects query status and the metrics emitted from the app. Let me try to catch you offline and then summarize the interesting points here.

ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/computation/ValidatedCommandFactory.java
```diff
@@ -177,6 +187,66 @@ private static Command createForAlterSystemQuery(
...
 
+  private static Command createForPauseQuery(
+      final ConfiguredStatement<? extends Statement> statement,
+      final KsqlExecutionContext context
+  ) {
+    final PauseQuery pauseQuery = (PauseQuery) statement.getStatement();
+    final Optional<QueryId> queryId = pauseQuery.getQueryId();
+
+    if (!queryId.isPresent()) {
+      context.getPersistentQueries().forEach(PersistentQueryMetadata::pause);
```
vcrfxia, line 198: Note for myself (and other readers) since I was confused the first time through: this code here is executed against a sandbox, in order to validate that the query ID exists and is not a source table query. The analogous code in InteractiveStatementExecutor is the one that actually executes this statement from the command topic.

ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/computation/ValidatedCommandFactory.java
```diff
@@ -177,6 +187,66 @@ private static Command createForAlterSystemQuery(
...
+
+    queryMetadata.pause();
+    return Command.of(statement);
+  }
+
+  private static Command createForResumeQuery(
+      final ConfiguredStatement<? extends Statement> statement,
+      final KsqlExecutionContext context
+  ) {
+    final ResumeQuery terminateQuery = (ResumeQuery) statement.getStatement();
```
vcrfxia, line 224: nit: name
```suggestion
    final ResumeQuery resumeQuery = (ResumeQuery) statement.getStatement();
```

ksqldb-rest-app/src/test/java/io/confluent/ksql/rest/integration/PauseResumeIntegrationTest.java
```diff
@@ -0,0 +1,279 @@
...
+  }
+
+  private int getRunningCount() {
+    return getCount(KsqlQueryStatus.RUNNING);
+  }
+
+  private int getCount(KsqlQueryStatus status) {
+    try {
+
+    List <KsqlEntity> showQueries2 = RestIntegrationTestUtil.makeKsqlRequest(REST_APP, "SHOW "
```
vcrfxia, line 266: nit: name
```suggestion
    List <KsqlEntity> showQueries = RestIntegrationTestUtil.makeKsqlRequest(REST_APP, "SHOW "
```

ksqldb-rest-app/src/test/java/io/confluent/ksql/rest/integration/PauseResumeIntegrationTest.java
```diff
@@ -0,0 +1,279 @@
...
+        + "QUERIES;");
+    System.out.println("Show queries " + showQueries2.stream().map( q ->
+            ((Queries)q).getQueries().get(0).getStatusCount().toString())
+        .collect(Collectors.joining("\n")));
+
+    // This failing shows that the command topic isn't being read completely?
+    return ((Queries) showQueries2.get(0)).getQueries().get(0).getStatusCount().getStatuses()
+        .get(status);
+    } catch (Exception e) {
+      return 0;
```
vcrfxia, line 276: nit: return `-1` (or some other nonsensical value) to indicate this error, rather than a plausible value (even though this value similarly fails tests).

vcrfxia (commented):
Thanks @jnh5y -- LGTM overall. 

I agree with the need to enhance command topic compaction to not repeatedly pause/resume queries during restart. Happy to have that be a separate PR, though, if that's your preference.

ksqldb-engine/src/main/java/io/confluent/ksql/util/BinPackedPersistentQueryMetadataImpl.java
jnh5y: Cool.  Did our conversations at the KSE on-site answer these questions well enough?

ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/computation/ValidatedCommandFactory.java
jnh5y: +1.  Yeah, some of this is confusing.  Shout if there's an obvious way to improve it for this PR!

ksqldb-engine/src/main/java/io/confluent/ksql/util/SharedKafkaStreamsRuntimeImpl.java
jnh5y: It does not; removing it!

Likely, it was left-over from the original cribbing off of terminate query.

ksqldb-engine/src/main/java/io/confluent/ksql/util/BinPackedPersistentQueryMetadataImpl.java
jnh5y: Ah!  I didn't know for sure that this class is used in a single-threaded manner.  So I assumed that an AtomicBoolean was better/more sensible.  

If there's multi-threaded access, I can see that things may be a mess for other things in class.

I'll switch from an AtomicBoolean to a regular one.

ksqldb-rest-model/src/main/java/io/confluent/ksql/rest/entity/PauseQueryEntity.java
jnh5y: Ah!  It is used here: https://github.com/confluentinc/ksql/pull/9203/files#diff-fea2df4b7059566421cb038afd9ded3c367b19c5d42f59b87284c75f12160b5bR198-R201

This class helps with the formatting for things in the Console.

vcrfxia: But PauseQueryEntity is never actually returned from anywhere, is it? When I looked earlier it seemed like all the infrastructure for pretty-printing was dead code since it's never called. Very possible I missed something, though.

jnh5y: Ok....  so I was cribbing off of `TERMINATE` when I implemented this.

It looks like there a separate `Executor` for Terminate commands which returns the `TerminateQueryEntity`.  

Likely I need to find the right `Executor` and add returns for the Pause and Resume query entities?  Does that seem reasonable?

vcrfxia: I think we can actually just remove all this. It looks like the only reason there's a separate `TerminateQueryExecutor` (from which `TerminateQueryEntity` is returned) at all is to handle pull query termination. The regular persistent query terminations go through the same distributing executor as your new pause/resume statements, and there's a different KsqlEntity already being returned from there. The CLI should know how to pretty-print those already.

You can run a quick manual check to see if the CLI output from pause/resume looks acceptable with the PR as is. If so, go ahead and remove all this as dead code.

ksqldb-rest-app/src/test/java/io/confluent/ksql/rest/integration/PauseResumeIntegrationTest.java
jnh5y: The goal of this test is to show that pausing one query does not impact a separate query.  I've updated the test to reflect that a little bit.

ksqldb-rest-app/src/test/java/io/confluent/ksql/rest/integration/PauseResumeIntegrationTest.java
jnh5y: Yes, I'm trying to add a little bit for that.

ksqldb-rest-app/src/test/java/io/confluent/ksql/rest/integration/PauseResumeIntegrationTest.java
jnh5y: I don't know.  I'll just re-use `queryId` from above!

ksqldb-rest-app/src/test/java/io/confluent/ksql/rest/integration/PauseResumeIntegrationTest.java
jnh5y: Updating the test.

ksqldb-rest-app/src/test/java/io/confluent/ksql/rest/integration/PauseResumeIntegrationTest.java
```diff
@@ -149,33 +149,39 @@ public void shouldPauseAndResumeMultipleQueries() {
...
 
     // 5 more records have been produced
     assertThatEventually(supplier, equalTo(14));
+    assertThatEventually(supplier2, equalTo(14));
 
     List <KsqlEntity> showQueries3 = RestIntegrationTestUtil.makeKsqlRequest(REST_APP, "SHOW "
         + "QUERIES;");
     assertThat(((Queries) showQueries3.get(0)).getQueries().get(0).getStatusCount().getStatuses()
         .get(KsqlQueryStatus.RUNNING), equalTo(1));
+    assertThat(getRunningCount(), equalTo(2));
```
vcrfxia, line 184: This looks like it was meant to replace the line above it?

ksqldb-rest-app/src/test/java/io/confluent/ksql/rest/integration/PauseResumeIntegrationTest.java
```diff
@@ -260,17 +251,18 @@ private int getRunningCount() {
     return getCount(KsqlQueryStatus.RUNNING);
   }
 
+  // JNH: Fix this.
```
vcrfxia, line 254: Out of curiosity, is this referring to the comment below that says `This failing shows that the command topic isn't being read completely?`? I'd be curious to understand more about what the issue is.

ksqldb-rest-app/src/test/java/io/confluent/ksql/rest/server/execution/ListQueriesExecutorTest.java
```diff
@@ -169,10 +169,11 @@ public void shouldScatterGatherAndMergeShowQueries() {
     final KsqlEngine engine = mock(KsqlEngine.class);
     when(engine.getAllLiveQueries()).thenReturn(ImmutableList.of(localMetadata));
     when(engine.getPersistentQueries()).thenReturn(ImmutableList.of(localMetadata));
-    
+
+    // new QueryStatusCount(Collections.singletonMap(q.getQueryStatus(), 1))
```
vcrfxia, line 173: Looks like this might've been left in unintentionally?

vcrfxia (commented):
Thanks @jnh5y -- this is looking really good! We should be ready to merge once the open issues we've already talked about are addressed: removing Pause/ResumeQueryEntity, migrations integration test, monitoring updates (outside the ksql repo). Also docs and command topic compaction, but those can be follow-ups.

ksqldb-rest-app/src/main/java/io/confluent/ksql/rest/server/computation/InteractiveStatementExecutor.java
```diff
@@ -265,7 +267,23 @@ private void executeStatement(
       final CommandId commandId,
       final Optional<CommandStatusFuture> commandStatusFuture
   ) {
-    if (statement.getStatement() instanceof TerminateQuery) {
+    if (statement.getStatement() instanceof PauseQuery) {
+      pauseQuery((PreparedStatement<PauseQuery>) statement);
+
+      final String successMessage = "Query paused.";
```
jnh5y, line 273: @vcrfxia Ah!  Here's what is returned when a query is run.  

Thoughts on this?  It follows what happens with `terminate`...

vcrfxia: Works for me! I'm all for slimmer code. Will take a look at your changes later today.

jnh5y: Yeah... it is maybe wonky if one is using PAUSE ALL.  Also, if one repeatedly pauses or resumes a query, then there's no note that the query was already paused, etc.

vcrfxia: Hm good point. We could refactor `pauseQuery()` to return the success message, based on whether the query exists, whether it was "PAUSE ALL" etc? Could also do the same thing for terminate while we're at it.

I think that'd be a nice touch. Could do it in a follow-up instead of this PR if you prefer.

ksqldb-rest-app/src/test/java/io/confluent/ksql/rest/integration/PauseResumeIntegrationTest.java
```diff
@@ -197,8 +202,7 @@ public void pausedQueriesShouldBePausedOnRestart()  {
     // When:
     String queryId = ((Queries) RestIntegrationTestUtil.makeKsqlRequest(REST_APP, "SHOW "
```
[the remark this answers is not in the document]
jnh5y: I think I've sorted out with some refactoring.

ksqldb-rest-app/src/test/java/io/confluent/ksql/rest/integration/PauseResumeIntegrationTest.java
jnh5y: Good catch.

ksqldb-rest-app/src/test/java/io/confluent/ksql/rest/integration/PauseResumeIntegrationTest.java
jnh5y: Fixed it!

ksqldb-engine/src/test/java/io/confluent/ksql/integration/EndToEndIntegrationTest.java
```diff
@@ -255,6 +259,73 @@ public void shouldSupportDroppingAndRecreatingJoinQuery() throws Exception {
     assertThat(columns.get(2).toString(), either(is("FEMALE")).or(is("MALE")));
   }
 
+  @Test
+  public void shouldSupportPauseAndResume() throws Exception {
```
vcrfxia, line 263: What does this test offer beyond the new integration tests in PauseResumeIntegrationTest.java? If nothing, we can remove this one. It'd be nice to keep all the integration testing for the feature in one place.

ksqldb-rest-app/src/test/java/io/confluent/ksql/rest/integration/PauseResumeIntegrationTest.java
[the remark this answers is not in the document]
vcrfxia: Sorry, what was the refactoring that was done here? It looks like the test still relies on this line returning the query ID for the query that corresponds to `supplier` rather than `supplier2`.

ksqldb-rest-app/src/test/java/io/confluent/ksql/rest/server/execution/ListQueriesExecutorTest.java
vcrfxia: This still needs to be addressed.

vcrfxia (approved):
Thanks @jnh5y -- LGTM pending the minor comments/questions inline.

As discussed offline, let's run a quick manual test for the migrations tool (which also exercises the java client) for now, and then add integration test coverage for both in a follow-up PR. Also docs updates for both (just to mention that PAUSE and RESUME are officially supported).

Command topic compaction can wait, and I see you've got a separate docs PR open already which is great. Let's get this merged!

ksqldb-rest-app/src/test/java/io/confluent/ksql/rest/integration/PauseResumeIntegrationTest.java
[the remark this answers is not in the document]
jnh5y: Ok, here's try two.  I'm getting the queryId immediately after creating only that first query.  That should skip over any ordering in the return of `show queries`.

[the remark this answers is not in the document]
vcrfxia: Thanks, this makes sense. Can you apply the same update to the other test in this file too?

PR #9205, opened 2022-06-16
fix: reset collector before reconfiguring

rodesai:
the shouldNotUpdateIfWithinInterval test checks that we dont
update the metric value if the interval has not passed. It
tries to configure a very long interval. However the config
does not get applied because we need to first reset the
static configuration.


spena (approved):

spena:
I'll merge this PR.

ableegoldman:
Hey @rodesai just wondering, any thoughts on why this so suddenly started to fail, and so consistently?

PR #9211, opened 2022-06-17
fix: change group name and extend CumulativeSum in ThroughputMetricsReporter

ableegoldman:
This PR contains two fixes:
1. changes the group name for ksql's per-query topic-level metrics to `ksql-query-throughput-metrics` so that we can distinguish them from the Streams metrics they are derived from
2. have the `ThroughputTotalMetric` extend the Kafka metrics class `CumulativeSum` so that the ce-kafka metrics reporter we integrate with will know to compute the diff and only forward the throughput deltas (see [here](https://github.com/confluentinc/ce-kafka/blob/6bb7cd13e0c7ca01474c3f04f4e081bdf42d2448/ce-metrics/src/main/java/io/confluent/telemetry/collector/KafkaMetricsCollector.java#L171)). This is a bit of a hack but it's necessary to ensure that the downstream metrics processing will accumulate the totals correctly.


rodesai (approved):
LGTM

PR #9215, opened 2022-06-19
fix: convert topic tag name and add consumer group member id tag to ThroughputTotalMetrics

ableegoldman:
Adjust the throughput metric tags  used in `ThroughputMetricsReporter` to conform to the existing tag names used elsewhere (eg in the clients):

1. Convert the `"topic-name"` tag currently set by Streams to just `"topic"` -- can remove this once the upstream fix is merged & synced through to ccs-kafka (won't cause any issues if it's not removed in time, though)
2. Replace the `"thread-id"` tag with the `"consumer_group_member_id"` tag that's used in other metrics/is an existing metrics label in Druid


swist (approved):
lgtm

Issue #5148, opened 2020-04-23
Topic configuration while creating a stream/table


stefanfrehse:
**Is your feature request related to a problem? Please describe.**
For statements as CSAS and CTAS topics are usually automatically created. Currently, the user has no direct way to change the default topic configuration directly at the point where the topic is created. The topic can, of course, be changed at the cluster level.

However, in our setup, we derived certain streams to have a use-case based materialized-view and these materialized-view topics need to have unlimited retention time/bytes.

**Describe the solution you'd like**
As a KSQLDB user, it would be nice to apply a configuration of a topic directly while creating CSAS or CTAS. For example, 

```
CREATE STREAM title_cased_songs WITH('log.retention.ms' = '-1') AS
    SELECT artist, UCASE(title) AS capitalized_song
    FROM rock_songs
    EMIT CHANGES;
```
**Describe alternatives you've considered**
Indeed, configure the topic at the cluster-level is possible but leads to two open-ends. Moreover, integrating into CI/CD is even more complex.

MichaelDrogalis:
@stefanfrehse Although it doesn't address all the aspects of your issue, you can [set the retention](https://docs.ksqldb.io/en/latest/concepts/time-and-windows-in-ksqldb-queries/#window-retention) time on a windowed materialized view using `RETENTION N DAYS`.

maksymilian-gasztych:
@MichaelDrogalis  Is there any estimate on when this feature could be released, if ever ? It would be really convenient to be able to set retention period related settings and cleanup.policy=compact,delete for tables.

MichaelDrogalis:
Hey @maksymilian-gasztych. We're not working on this right now, but we'd take a patch for it if you have the bandwidth?

PR #9223, opened 2022-06-24
Add topic retention config to WITH clause

bvarghese1:
### Description 
Pass retention_ms as a config in the WITH clause.
This param will be used to set the topic retention while creating the topic if the topic is not already present. 
It does not update the retention if the topic is already present.
Additionally, validation is performed to ensure retention_ms config cannot be changed once a topic exists similar to partition count.

### Testing done 
Unit tests
Integration Tests
Manual Tests using CLI
```
ksql> CREATE STREAM s1 (k1 int key, v1 int) WITH (kafka_topic='s1', format='avro', partitions=2);

 Message
----------------
 Stream created  // Creates topic with default retention of 1 day
----------------

ksql> CREATE STREAM s2 (k1 int key, v1 int) WITH (kafka_topic='s2', format='avro', partitions=2, retention_ms=30000);

 Message
----------------
 Stream created  // Creates topic with retention.ms=30000
----------------

ksql> CREATE STREAM s3 (k1 int key, v1 int) WITH (kafka_topic='s2', format='avro', partitions=2, retention_ms=50000);
A Kafka topic with the name 's2' already exists, with different partition/replica/retention configuration than required. KSQL expects 2 partitions (topic has 2), 1 replication factor (topic has 1), and 50000 retention (topic has 30000).

ksql> CREATE STREAM ss2 AS SELECT * FROM s2;

 Message
----------------------------------
 Created query with ID CSAS_SS2_9    // Creates topic SS2 with retention.ms=30000 & partitions=2 copied from s2
----------------------------------

ksql> CREATE STREAM ss3 WITH (kafka_topic='ss3', partitions=3, format='avro', retention_ms=50000) AS SELECT * FROM s2;

 Message
 Created query with ID CSAS_SS3_11  // Creates topic SS3 with retention.ms=50000 & partitions=3 as specified in WITH

ksql> CREATE TABLE t1 (k1 int PRIMARY KEY, v1 int) WITH (kafka_topic='t1', partitions=2, format='avro');

 Message
---------------
 Table created
---------------

ksql> CREATE TABLE t2 (k1 int PRIMARY KEY, v1 int) WITH (kafka_topic='t2', partitions=2, format='avro', retention_ms=30000);
Invalid config variable in the WITH clause: RETENTION_MS. Non-windowed tables do not support retention.

ksql> CREATE TABLE t2 (k1 int PRIMARY KEY, v1 int) WITH (kafka_topic='t1', partitions=2, format='avro', retention_ms=30000);
Invalid config variable in the WITH clause: RETENTION_MS. Non-windowed tables do not support retention.

ksql> CREATE TABLE tt2 WITH (kafka_topic='tt2', format='avro', partitions=3, retention_ms=50000) AS SELECT * FROM t1;
Invalid config variable in the WITH clause: RETENTION_MS. Non-windowed tables do not support retention.

ksql> CREATE TABLE tt2 WITH (kafka_topic='tt2', format='avro', partitions=3) AS SELECT * FROM t1;

 Message
 Created query with ID CTAS_TT2_15

ksql> CREATE TABLE tt3 WITH (kafka_topic='tt3', format='avro', partitions=3, retention_ms=30000) AS SELECT k1, count(*) FROM s1 WINDOW TUMBLING (SIZE 10 SE
CONDS, RETENTION 2 DAYS) GROUP BY k1;

 Message
 Created query with ID CTAS_TT3_17  // Creates topic with retention.ms=172800000 (2 DAYS)

ksql> CREATE TABLE tt4 WITH (kafka_topic='tt4', format='avro', partitions=3, retention_ms=432000000) AS SELECT k1, count(*) FROM s1 WINDOW TUMBLING (SIZE 1
0 SECONDS, RETENTION 2 DAYS) GROUP BY k1;

 Message
 Created query with ID CTAS_TT4_19  // Creates topic with retention.ms=432000000 (5 DAYS)

ksql> CREATE TABLE tt5 WITH (kafka_topic='tt4', format='avro', partitions=3, retention_ms=50000) AS SELECT k1, count(*) FROM s1 WINDOW TUMBLING (SIZE 10 SE
CONDS, RETENTION 2 DAYS) GROUP BY k1;
A Kafka topic with the name 'tt4' already exists, with different partition/replica/retention configuration than required. KSQL expects 3 partitions (topic has 3), 1 replication factor (topic has 1), and 172800000 retention (topic has 432000000).

```

### Reviewer checklist
- Docs may need to be updated
- Fixes https://github.com/confluentinc/ksql/issues/5148
- https://confluentinc.atlassian.net/browse/FF-5549


ksqldb-common/src/main/java/io/confluent/ksql/properties/with/CommonCreateConfigs.java
```diff
@@ -86,6 +87,14 @@ public static void addToConfigDef(
                 + "Kafka cluster configuration for replicas will be used for creating a new "
                 + "topic."
         )
+        .define(
+            SOURCE_TOPIC_RETENTION_IN_MS,
+            ConfigDef.Type.LONG,
+            null,
+            Importance.MEDIUM,
+            "The retention in milliseconds in the backing topic. If this property is"
+                + "not set then the default value of 7 days will be used for creating a new topic."
```
agavra, line 96: > not set then the default value of 7 days will be used for creating a new topic.

the default for tables is infinite + compaction, which we should make sure that we maintain. Actually, we may want to consider not allowing users to specify this when they create a table altogether

ksqldb-engine/src/main/java/io/confluent/ksql/topic/TopicCreateInjector.java
```diff
@@ -206,6 +215,10 @@ private TopicProperties createTopic(
 
     final Map<String, Object> config = new HashMap<>();
     config.put(TopicConfig.CLEANUP_POLICY_CONFIG, topicCleanUpPolicy);
+    // This is to make sure WINDOW RETENTION config is honored over WITH RETENTION_MS config
```
agavra, line 218: shouldn't we chose the greater of the two? `WINDOW RETENTION` can be less than the topic retention (I think) it just can't be greater than

agavra (commented):
Implementation looks good to me, some feature level comments inline. Two other open questions:

1. Can you test to make sure that this doesn't affect the retention of internal topics? (It shouldn't, but would be good to double check)
2. In the case that the topic exists and retention is different (e.g. in the `CS` case), do we want to consider either (a) failing the statement or (b) allowing it to override? In the case of partitions, if you specify something different than the existing partitions the statement will fail (IIRC)

ksqldb-engine/src/main/java/io/confluent/ksql/topic/TopicCreateInjector.java
bvarghese1: Users might be used to the `WINDOW RETENTION` config which gets applied to the topic retention - that was the reason I chose to leave it as is. But what you said makes sense. We could choose the greater of the two.
cc @colinhicks

colinhicks: Choosing the greater value sounds good to me, @bvarghese1

ksqldb-common/src/main/java/io/confluent/ksql/properties/with/CommonCreateConfigs.java
colinhicks: > Actually, we may want to consider not allowing users to specify this when they create a table altogether

I agree with this, @bvarghese1. Let's allow retention to be customized only for streams

bvarghese1:
> Implementation looks good to me, some feature level comments inline. Two other open questions:
> 
> 1. Can you test to make sure that this doesn't affect the retention of internal topics? (It shouldn't, but would be good to double check)
> 2. In the case that the topic exists and retention is different (e.g. in the `CS` case), do we want to consider either (a) failing the statement or (b) allowing it to override? In the case of partitions, if you specify something different than the existing partitions the statement will fail (IIRC)

For 2, I think we should fail if the retention specified is different than the existing topic's retention. The same is true for partitions.

colinhicks:
> 2. In the case that the topic exists and retention is different (e.g. in the CS case), do we want to consider either (a) failing the statement or (b) allowing it to override? In the case of partitions, if you specify something different than the existing partitions the statement will fail (IIRC)

Let's make the behavior for retention consistent with the behavior for partition count.

ksqldb-engine/src/test/java/io/confluent/ksql/services/FakeKafkaTopicClient.java
```diff
@@ -103,20 +105,27 @@ public int hashCode() {
...
       final String topic,
       final int numPartitions,
       final int replicationFactor,
       final Map<String, ?> configs) {
-    topicMap.put(topic, createFakeTopic(topic, numPartitions, replicationFactor, configs));
+    final FakeTopic info = createFakeTopic(topic, numPartitions, replicationFactor, configs);
+    topicMap.put(topic, info);
+    topicMapConfig.put(topic, configs);
+    //createdTopics.put(topic, info);
+    //createdTopicsConfig.put(topic, configs);
```
agavra, line 128: nit: commented out code

ksqldb-engine/src/main/java/io/confluent/ksql/topic/TopicCreateInjector.java
agavra: did we make this change?

agavra (approved):
LGTM! Did a quick skim and have to inline comments.

ksqldb-engine/src/main/java/io/confluent/ksql/topic/TopicCreateInjector.java
bvarghese1: No not yet. Still working on this PR. I think I need to also re-evaluate some of the tests now that I understand the cleanup.policy settings impact on retention.

ksqldb-engine/src/test/java/io/confluent/ksql/services/FakeKafkaTopicClient.java
bvarghese1: Removed.

ksqldb-engine/src/main/java/io/confluent/ksql/topic/TopicCreateInjector.java
bvarghese1: Made the change to pick the greater of the two(i.e. retention config passed via WITH and retention config specified in the WINDOW config).

ksqldb-engine/src/main/java/io/confluent/ksql/topic/TopicCreateInjector.java
```diff
@@ -176,20 +185,34 @@ private <T extends CreateAsSelect> ConfiguredStatement<?> injectForCreateAsSelec
...
     return statement.withStatement(withTopicText, withTopic);
   }
 
+  private void throwIfRetentionPresentForTable(
+      final String topicCleanUpPolicy,
+      final Optional<Long> retentionInMillis
+  ) {
+    if (topicCleanUpPolicy.equals(TopicConfig.CLEANUP_POLICY_COMPACT)
+        && retentionInMillis.isPresent()) {
+      throw new KsqlException("Invalid config variable in the WITH clause: RETENTION_MS");
```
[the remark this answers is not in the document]
bvarghese1:
```
Invalid config variable in the WITH clause: RETENTION_MS. Non-windowed tables do not support retention.
```
OR
```
Invalid config variable in the WITH clause: RETENTION_MS. This config cannot be applied when the cleanup.policy setting is set to compact.
```
@agavra will any of the above error messages help?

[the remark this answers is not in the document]
colinhicks: I would go with the first option above, @bvarghese1. Some folks may not be aware of the relationship between ksql tables and Kafka compaction. Best to avoid increasing the conceptual scope.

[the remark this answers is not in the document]
agavra: +1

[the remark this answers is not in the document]
bvarghese1: Done

PR #9225, opened 2022-06-25
fix: change consumer_group_member_id tag to just member to match Druid label name

ableegoldman:
I realized that while the final tag in the Metrics API is called `consumer_group_id`, it's actually just called `member` within Druid. So we need to change the tag to match this


rodesai (approved):
LGTM

PR #9226, opened 2022-06-25
fix: Trig tests failing on master/Java 8

reneesoika:
### Description 
The checks for PR #9213 passed because they used Java 11. However, master uses Java 8, and some of the unit tests are now failing because the floating point values are not matching. (The Java library trig methods allow a 1 ulp difference from the exact result, so some implementation or optimization must have changed between versions.)

I have updated the trig unit tests to use close matches for floating point values instead of exact matches.

### Testing done 
Confirmed the tests were failing locally on Java 8 and then passing after these fixes. Also checked Java 11.

### Reviewer checklist


pgaref (approved):
Thanks for fixing this @reneesoika !

